PR #985, opened 2019-09-25
doc: vpr: Fix typo

tmichalak:
@kmurray added few tweaks to PR #983 (thanks for that!). 
There is one typo that caught my eye though, namely ``Placment`` instead of ``Placement``.
This small PR fixes it.


kmurray:
Thanks for the fix!

PR #986, opened 2019-09-25
Use FASM prefix of parent pb_type, not prefix of output pb_type.

litghost:
#### Description

The previous code used the clb prefix of the current block, but the
`interconnect`/`fasm_mux` logic operates on the parent pb_type of the
current block, not the current block itself.  As a result, if the output
of a `fasm_mux` has a prefix, it was getting lumped into the FASM mux
declaration.

#### Motivation and Context

The intention of `fasm_prefix` was to apply to metadata at the current level and deeper.  This bug caused a `fasm_prefix` to leak one level higher than intended.

1. Integration tests on Symbiflow
2. New unit test

#### Types of changes

#### Checklist:


kmurray:
Thanks for the fix!

Issue #513, opened 2019-03-22
Support Equivalent Placement Sites


kmurray:
#### Proposed Behaviour

It is possible that a single logical block type could be mapped to multiple potential physical grid tiles in the FPGA.

Examples of this include
* IO types associated with different locations (e.g. IO_LEFT, IO_RIGHT, IO_TOP, IO_BOTTOM)
* LUTRAM (CLBL/CLBM, LAB/MLAB)

In general it should be possible to list a set of equivalent placement locations/sites for each block type.

#### Current Behaviour

VPR assumes that each top-level `pb_type` can only be placed at a placement location of exactly the same type.

#### Possible Solution

We should probably decouple the packing decisions (i.e. logical block types exist and can be created), from the placer's decisions (what grid tiles the logical block types can be placed in).
Currently there is no distinction.

#### Context

This will help improve the generality of the flow.
It will also fix non-intuitive/confusing behaviour such as #268, #512, #349.

kmurray:
[Proposal](https://docs.google.com/document/d/1LQ0H2C9IFnCVtSuaniCs1dDmBoMBWwBQLKAGuo_kSc0/edit#heading=h.uo6gk3q6tsrk) from @mithro

acomodi:
Hi @kmurray. I have been thinking about this issue and how to proceed with its implementation.

**Considerations**

Currently there are three types of `pb_types` that have different properties (top level, intermediate and primitives). The idea would be to get to a situation for which the top level `pb_type` is represented as a physical `tile`, and this representation will be used by the placer to decide where the clustered blocks obtained from the `packer` have to be positioned.
As far as I understand, there already is a data structure that represents a physical type which is [`t_type_descriptor`](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d4771a1e5cb4b9c2501abf9fb6768d9a7d5dc45e/libs/libarchfpga/src/physical_types.h#L544) and is used to hold information of the top level `pb_types` of the architecture (correct me if I'm wrong).

**How to proceed**

First thing I would focus on, is to substitute the idea of the top level `pb_type` with idea of `tile`. Therefore, in the `XML`  description we will have a new top level `tiles` tag representing all the previous top level `pb_types` with all the tags and properties of a top level `pb_type`.
IMO this will help decoupling packer and placer decisions. The packer will still use information gathered from the `pb_types` structure, while the placer will work at the `tile` level.
An idea on how it will look like is the following:
[...]

PR #988, opened 2019-10-02
Equivalent sites

acomodi:
#### Description

This PR is a followup of https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/941. The main purpose is to add the equivalent sites placement. In fact, if the architecture specifies that a pb_type can be placed in two or more different physical tiles, the placer can now use those tiles.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/513

#### Motivation and Context

The placer should be able to find better solutions whenever there is a clustered block that could be placed in all the possible physical tiles that can be used.

This change neatly separates the domains of the packer and placer. This means that the placer becomes more flexible in finding solutions that are currently unfeasible and strictly dependent on the packer outcomes.

There is a WIP regression test, which needs to be fixed.

#### Types of changes

#### Checklist:


acomodi:
@kmurray @litghost FYI

I have added the WIP tag as the regression tests will fail due to the need of updating the architecture files. In fact now also the pin directs have to be provided to each equivalent site.

I believe though that the major effort is completed and this PR can go through a first review step.

acomodi:
I have just tried to manually update the archs and got the following error on all the regression tests:

```
...
Warning 24: in check_rr_node: RR node: 631 type: OPIN location: (1,1) pin: 102 pin_name: clb.cout[0] has no out-going edges.
Warning 25: in check_rr_node: RR node: 632 type: OPIN location: (1,1) pin: 103 pin_name: clb.cout[1] has no out-going edges.
Warning 26: in check_rr_node: RR node: 1573 type: OPIN location: (3,1) pin: 102 pin_name: clb.cout[0] has no out-going edges.
Warning 27: in check_rr_node: RR node: 1574 type: OPIN location: (3,1) pin: 103 pin_name: clb.cout[1] has no out-going edges.
Warning 28: in check_rr_graph: fringe node 24 IPIN at (0,1) has no fanin.                                          
         This is possible on a fringe node based on low Fc_out, N, and certain lengths.                            
## Build routing resource graph took 0.00 seconds (max_rss 23.3 MiB, delta_rss +0.0 MiB)                           
  RR Graph Nodes: 2676                                                                                             
  RR Graph Edges: 5593                                                                                             
Confirming router algorithm: TIMING_DRIVEN.                                                                        
---- ------ ------- ---- ------- ------- ------- ----------------- --------------- -------- ---------- ---------- ---------- ---------- --------
Iter   Time    pres  BBs    Heap  Re-Rtd  Re-Rtd Overused RR Nodes      Wirelength      CPD       sTNS       sWNS       hTNS       hWNS Est Succ
      (sec)     fac Updt    push    Nets   Conns                                       (ns)       (ns)       (ns)       (ns)       (ns)     Iter
---- ------ ------- ---- ------- ------- ------- ----------------- --------------- -------- ---------- ---------- ---------- ---------- --------
vtr/vpr/src/route/route_timing.cpp:1044 timing_driven_route_net: Assertion 'route_ctx.rr_node_route_inf[rt_root->inode].occ() <= device_ctx.rr_nodes[rt_root->inode].capacity()' failed (SOURCE should never be congested).
Aborted (core dumped) 
```

I guess there is something wrong with the `rr_nodes`

acomodi:
@kmurray I have solved the issue with routing. When assigning the physical pin index from the `directs` mapping I did not consider the capacity, hence there was an overusage of node causing the failure.

I have temporarily added the upgrade_arch.sh script step in Travis CI to check the regression tests (https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/988/commits/f050da86b0d14af2759ac56a3f7d9b693a28d33b).

CI should turn green, at least on the `strong_regression_tests` which I have locally tested.

acomodi:
@kmurray FYI, I have fixed the issues to let CI go green

acomodi:
@kmurray with [this commit](https://github.com/verilog-to-routing/vtr-verilo-to-routing/pull/988/commits/d0f0ed82583f6c3795d0ab1e75bd0e07a81b4622) I am trying to get rid of the necessity of calling the `pin_physical_index()` method to retrieve the physical index associated with a block.
I have thought to move the information on what logical pin is associated with a specific tile elsewhere, instead of having it embedded in the clustering information.

acomodi:
@kmurray Ping.
This PR can now use equivalent sites in more complex situations as done in this Symbiflow PR: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1072.

This still needs some work, but IMO it is ready for review.

acomodi:
TODO 
- [ ] More intelligent initial placement;
- [ ] SINK/SOURCE to IPIN/OPIN issue;
- [ ] Add documentation

mithro:
Build failure is probably caused by the build cache? If @kmurray can delete the cache it should work...

acomodi:
@kmurray @vaughnbetz Ping

kmurray:
I'm still working through the review, but one initial thought is related to the architecture file format.

As currently implemented, we require the user to explicitly specify the mapping from tile to pb_type pins. This is quite repetitive for the common case, as all of our architectures use a direct 1:1 mapping from tile to pb_type pins:
```xml
 <tile name="clb" area="53894">
      <equivalent_sites>
        <site pb_type="clb">
          <direct from="clb.I1" to="clb.I1"/>
          <direct from="clb.I2" to="clb.I2"/>
          <direct from="clb.I3" to="clb.I3"/>
          <direct from="clb.I4" to="clb.I4"/>
          <direct from="clb.cin" to="clb.cin"/>
          <direct from="clb.O" to="clb.O"/>
          <direct from="clb.cout" to="clb.cout"/>
          <direct from="clb.clk" to="clb.clk"/>
        </site>
      </equivalent_sites>
      <input name="I1" num_pins="13" equivalent="full"/>
      <input name="I2" num_pins="13" equivalent="full"/>
      <input name="I3" num_pins="13" equivalent="full"/>
      <input name="I4" num_pins="13" equivalent="full"/>
      <input name="cin" num_pins="1"/>
      <output name="O" num_pins="20" equivalent="none"/>
      <output name="cout" num_pins="1"/>
      <clock name="clk" num_pins="1"/>
      <fc in_type="frac" in_val="0.15" out_type="frac" out_val="0.10">
        <fc_override port_name="cin" fc_type="frac" fc_val="0"/>
        <fc_override port_name="cout" fc_type="frac" fc_val="0"/>
      </fc>
      <pinlocations pattern="spread"/>
    </tile>
```

I think it would be good to have short form for this. Something like:
```xml
 <tile name="clb" area="53894">
      <equivalent_sites>
        <site pb_type="clb" pin_mapping="direct"/>
      </equivalent_sites>
      <input name="I1" num_pins="13" equivalent="full"/>
      <input name="I2" num_pins="13" equivalent="full"/>
      <input name="I3" num_pins="13" equivalent="full"/>
      <input name="I4" num_pins="13" equivalent="full"/>
      <input name="cin" num_pins="1"/>
      <output name="O" num_pins="20" equivalent="none"/>
      <output name="cout" num_pins="1"/>
      <clock name="clk" num_pins="1"/>
      <fc in_type="frac" in_val="0.15" out_type="frac" out_val="0.10">
        <fc_override port_name="cin" fc_type="frac" fc_val="0"/>
        <fc_override port_name="cout" fc_type="frac" fc_val="0"/>
      </fc>
      <pinlocations pattern="spread"/>
    </tile>
```
and we could still support customization like:
```xml
 <tile name="clb" area="53894">
      <equivalent_sites>
        <site pb_type="clb" pin_mapping="custom">
          <direct from="clb.I1" to="clb.I1"/>
          <direct from="clb.I2" to="clb.I2"/>
          <direct from="clb.I3" to="clb.I3"/>
          <direct from="clb.I4" to="clb.I4"/>
          <direct from="clb.cin" to="clb.cin"/>
          <direct from="clb.O" to="clb.O"/>
          <direct from="clb.cout" to="clb.cout"/>
          <direct from="clb.clk" to="clb.clk"/>
        </site>
      </equivalent_sites>
      <input name="I1" num_pins="13" equivalent="full"/>
      <input name="I2" num_pins="13" equivalent="full"/>
      <input name="I3" num_pins="13" equivalent="full"/>
      <input name="I4" num_pins="13" equivalent="full"/>
      <input name="cin" num_pins="1"/>
      <output name="O" num_pins="20" equivalent="none"/>
      <output name="cout" num_pins="1"/>
      <clock name="clk" num_pins="1"/>
      <fc in_type="frac" in_val="0.15" out_type="frac" out_val="0.10">
        <fc_override port_name="cin" fc_type="frac" fc_val="0"/>
        <fc_override port_name="cout" fc_type="frac" fc_val="0"/>
      </fc>
      <pinlocations pattern="spread"/>
    </tile>
```

It would be best to apply this before merging this PR, as it would be much more difficult to reverse the expanded to simple from *before* modifying all the architecture files to use the expanded form.

acomodi:
@kmurray Makes sense, I'm on it

vaughnbetz:
Thanks Kevin and Alessandro.  I think that's a great convenience idea/feature.

acomodi:
> I think it would be best to apply this change before merging this PR, as it would be eaiser to modify the architecture file into the appropriate form.

@kmurray Done

acomodi:
@kmurray @vaughnbetz I am thinking also that maybe we could make the `pin_mapping` attribute optional and set it to `direct` as default. There is of course a check that makes sure that, for instance, the number of pins of the pb_type and the tile are the same and that the `direct` pin_mapping can be applied.

vaughnbetz:
That seems like a good idea to me.

acomodi:
I have double-checked and the changes with this PR do not break the drawing tool, at least for the architectures that have no multiple equivalent sites.

acomodi:
The titan benchmark comparison between `master` and `equivalent-sites` is here: [comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/3857923/comparison.xlsx)

vaughnbetz:
It looks like the placements are identical -- is that right?
It looks like there is some compile time increase (2% pack, 3% place and 5% route).  Is that just due to machine load?  Not sure why everything would slow down.
The critical path delay also went up slightly, but I assume that's noise (i.e. are placements the same)?

acomodi:
> It looks like the placements are identical -- is that right?

Indeed, packing and placement results are identical. I will re-run this PR again against vtr_titan benchmarks to check whether it was due to machine load.

vpr/src/base/clustered_netlist.h
```diff
@@ -148,9 +148,13 @@ class ClusteredNetlist : public Netlist<ClusterBlockId, ClusterPortId, ClusterPi
      */
 
     //Returns the physical pin index (i.e. pin index on the
-    //t_logical_block_type) of the specified logical pin
+    //t_physical_tile_type) of the specified logical pin
     int pin_physical_index(const ClusterPinId id) const;
 
+    //Returns the logical pin index (i.e. pin index on the
+    //t_logical_block_type) of the cluster pin
+    int pin_logical_index(const ClusterPinId pin_id) const;
```
kmurray: Do we need to store both these pieces of information in the netlist?

While knowing the pin index within the logical block type makes sense to have in the netlist, I'm less certain about the pin_physical_index (i.e. the pin index of the tile where the block is placed). That information is really dependent on placement and is not directly clustered netlist information.

It seems like pin_physical_index() would be better as a free function (i.e. not a member of ClusteredNetlist), which looked up this information directly from the placement & tile to logical pin mapping. So something like
```cpp
int pin_tile_index(const ClusterPinId pin) {
    //Get the logical pin index of pin within it's logical block type

   //Determine the physical tile type implementing the logical block (i.e. based on the current placement)

   //Use the physical/logical pin mapping to convert the logical block pin index to the physical tile pin index & return it
}
```

This would avoid duplicating this state information in the CLB netlist and avoid a whole class of potential bugs, where due to a placement change the physical tile pin index stored in the ClusteredNetlist gets out of date.

vpr/src/place/initial_placement.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+    // The map serves to place first the most constrained block ids
+    std::map<int, std::vector<ClusterBlockId>> sorted_block_map;
+
+    for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
+        auto logical_block = cluster_ctx.clb_nlist.block_type(blk_id);
+
+        size_t num_equivalent_tiles = logical_block->equivalent_tiles.size();
+        sorted_block_map[num_equivalent_tiles].push_back(blk_id);
+    }
```
kmurray: It would likely be clearer and more straight forward to extend in the future if we just had an plain vector of ClusterBlockId, which we then sort according to some criteria:

```cpp
auto blocks = cluster_ctx.clb_nlist.blocks();
std::vector<ClusterBlockId> sorted_blocks(blocks.begin(), blocks.end());

auto criteria = [](const ClusterBlockId lhs, ClusterBlockId rhs) {
  //Sort by most possible instances in placement grid
}

std::sort(sorted_blocks.begin(), sorted_blocks.end(), criteria);
```
This makes it straight forward to change the criteria in the future (rather than requiring that it be sorted by number of tiles as currently implemented).

It also makes it possible to extend this initial placement algorithm with the 'push to front heuristic' for difficult to place blocks as previously discussed (simply move problematic blocks to the front of the list and re-do initial placement).

vpr/src/place/initial_placement.cpp
```diff
@@ -0,0 +1,426 @@
...
+        for (auto blk_id : sorted_blocks.second) {
+            if (place_ctx.block_locs[blk_id].loc.x != -1) { // -1 is a sentinel for an empty block
+                // block placed.
+                continue;
+            }
+
+            auto logical_block = cluster_ctx.clb_nlist.block_type(blk_id);
+
+            /* Don't do IOs if the user specifies IOs; we'll read those locations later. */
+            if (!(is_io_type(pick_random_physical_type(logical_block)) && pad_loc_type == USER)) {
```
kmurray: Both here and below, we probably shouldn't be using 'pick_random_physical_type()', since that may return a different type each time. We should pick a single type to try and be consistent through-out.

Otherwise a different random choice of type may be made which would be inconsistent (may pass a check one time, but not the same check again later).

doc/src/arch/reference.rst
```diff
@@ -1183,14 +1183,28 @@ The following tags are common to all ``<tile>`` tags:
 
     .. arch:tag:: <site pb_type="string"/>
```
kmurray: Need to add the `pin_mapping=` attribute and description.

doc/src/arch/reference.rst
```diff
@@ -1183,14 +1183,28 @@ The following tags are common to all ``<tile>`` tags:
...
+    Each instance of site must also specify the direct connections between the physical
+    tile pins and the logical block pins.
+
+    .. arch:tag:: <direct from="string" to="string">
+
+    Attributes:
+        - ``from`` is relative to the physical tile pins
+        - ``to`` is relative to the logical block pins
+
     :req_param pb_type: Name of the corresponding pb_type.
```
kmurray: Should move `:req_param pb_type` directly under the `<site>` tag description.

doc/src/arch/reference.rst
```diff
@@ -1183,14 +1183,28 @@ The following tags are common to all ``<tile>`` tags:
...
 
+    Each instance of site must also specify the direct connections between the physical
+    tile pins and the logical block pins.
+
+    .. arch:tag:: <direct from="string" to="string">
+
+    Attributes:
+        - ``from`` is relative to the physical tile pins
+        - ``to`` is relative to the logical block pins
+
```
kmurray: As a sub-tag of `<equivalent_sites>` this should be indented, and probably put below the equivalent sites example.

libs/libarchfpga/src/physical_types.h
```diff
@@ -626,14 +610,52 @@ struct t_physical_tile_type {
...
+    }
+
+    bool operator==(const t_physical_pin o) const {
+        return pin == o.pin;
+    }
+
+    bool operator<(const t_physical_pin o) const {
+        return pin < o.pin;
+    }
+};
```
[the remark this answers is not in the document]
acomodi: I needed to find a way to use the `vtr::bimap` in order to map physical/logical pins in case we need to access one having the other and viceversa. `vtr::bimap` though does not allow (as far as I have seen), having the same types when doing the mapping. Instead of modifying the bimap library I have preferred to go with the wrapping solution

vpr/src/route/rr_graph.cpp
```diff
@@ -2654,94 +2654,103 @@ static void build_unidir_rr_opins(const int i, const int j, const e_side side, c
...
             VTR_ASSERT(start_pin_index == end_pin_index);
             start_pin_index = 0;
-            end_pin_index = pb_type->ports[j].num_pins - 1;
+            end_pin_index = tile_port.num_pins - 1;
         }
 
-        get_blk_pin_from_port_pin(clb_to_clb_directs[i].to_clb_type->index, j, start_pin_index, &clb_to_clb_directs[i].to_clb_pin_start_index);
-        get_blk_pin_from_port_pin(clb_to_clb_directs[i].to_clb_type->index, j, end_pin_index, &clb_to_clb_directs[i].to_clb_pin_end_index);
+        clb_to_clb_directs[i].to_clb_pin_start_index = tile_port.absolute_first_pin_index + start_pin_index;
+        clb_to_clb_directs[i].to_clb_pin_end_index = tile_port.absolute_first_pin_index + end_pin_index;
```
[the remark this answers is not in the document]
acomodi: IMO it would be better to leave this as it is as this operation is applied only here (this one modifies the `to_clb_pin` indices, while the one above modifies the `from_clb_pin` indices).

libs/libarchfpga/src/physical_types.h
```diff
@@ -665,20 +687,30 @@ struct t_physical_tile_port {
...
+    int index = -1; /* index of type descriptor in array (allows for index referencing) */
+
+    std::vector<t_physical_tile_type_ptr> equivalent_tiles;
 
-    // XXX Variables to hold information on mapping between site and tile
-    // XXX as well as references to the belonging pb_type and tile_type
-    //t_logical_block_type* block_type;
+    /* Map holding the priority for which this logical block needs to be placed.
+     * physical_tiles_priority[priority] -> vector holding the physical tile indices */
+    std::map<int, std::vector<int>> physical_tiles_priority;
```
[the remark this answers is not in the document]
acomodi: Yes indeed, I think `priority` is a leftover as it wasn't actually used anywhere. Removed

vpr/src/base/clustered_netlist.cpp
```diff
@@ -52,7 +52,7 @@ int ClusteredNetlist::block_pin_net_index(const ClusterBlockId blk_id, const int
 
 ClusterPinId ClusteredNetlist::block_pin(const ClusterBlockId blk, const int phys_pin_index) const {
     VTR_ASSERT_SAFE(valid_block_id(blk));
-    VTR_ASSERT_SAFE_MSG(phys_pin_index >= 0 && phys_pin_index < physical_tile_type(block_type(blk))->num_pins, "Physical pin index must be in range");
+    VTR_ASSERT_SAFE_MSG(phys_pin_index >= 0 && phys_pin_index < pick_random_physical_type(block_type(blk))->num_pins, "Physical pin index must be in range");
```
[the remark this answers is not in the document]
acomodi: Fixed

vpr/src/place/initial_placement.cpp
acomodi: I have added the sorting step instead of the `map` solution. The only problem with this, as far as I have locally experienced, is that the regression tests will complete, but some of the QoR results will be different. I have checked the placer outputs and they are actually different and the run-time has slightly increased (for what regards strong tests).

Would it be ok, in your opinion, to add a check to see whether the architecture does have equivalent sites or not. In the case it has some tiles with multiple sites, the sorting is triggered, otherwise not.

vpr/src/place/initial_placement.cpp
```diff
@@ -0,0 +1,426 @@
...
+                       const char* pad_loc_file) {
+    /* Randomly places the blocks to create an initial placement. We rely on
+     * the legal_pos array already being loaded.  That legal_pos[itype] is an
+     * array that gives every legal value of (x,y,z) that can accommodate a block.
+     * The number of such locations is given by num_legal_pos[itype].
+     */
+
+    // Loading legal placement locations
+    alloc_legal_placements();
+    load_legal_placements();
```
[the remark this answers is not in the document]
acomodi: Done

vpr/src/route/rr_graph.cpp
```diff
@@ -2654,94 +2654,103 @@ static void build_unidir_rr_opins(const int i, const int j, const e_side side, c
...
             VTR_ASSERT(start_pin_index == end_pin_index);
             start_pin_index = 0;
-            end_pin_index = pb_type->ports[j].num_pins - 1;
+            end_pin_index = tile_port.num_pins - 1;
         }
-        get_blk_pin_from_port_pin(clb_to_clb_directs[i].from_clb_type->index, j, start_pin_index, &clb_to_clb_directs[i].from_clb_pin_start_index);
-        get_blk_pin_from_port_pin(clb_to_clb_directs[i].from_clb_type->index, j, end_pin_index, &clb_to_clb_directs[i].from_clb_pin_end_index);
+
+        clb_to_clb_directs[i].from_clb_pin_start_index = tile_port.absolute_first_pin_index + start_pin_index;
+        clb_to_clb_directs[i].from_clb_pin_end_index = tile_port.absolute_first_pin_index + end_pin_index;
```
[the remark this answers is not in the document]
acomodi: Done

vpr/src/util/vpr_utils.cpp
```diff
@@ -330,36 +312,42 @@ void swap(IntraLbPbPinLookup& lhs, IntraLbPbPinLookup& rhs) {
...
-        atom_pins = find_clb_pin_sink_atom_pins(clb, clb_pin, pb_gpin_lookup);
+        atom_pins = find_clb_pin_sink_atom_pins(clb, log_pin, pb_gpin_lookup);
     }
 
     return atom_pins;
 }
 
 //Returns the atom pin which drives the top level clb output pin
-AtomPinId find_clb_pin_driver_atom_pin(ClusterBlockId clb, int clb_pin, const IntraLbPbPinLookup& pb_gpin_lookup) {
+AtomPinId find_clb_pin_driver_atom_pin(ClusterBlockId clb, int log_pin, const IntraLbPbPinLookup& pb_gpin_lookup) {
```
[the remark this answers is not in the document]
acomodi: I have actually changed all the `log_pin` and `phy_pin` with a more extensive naming convention, namely `logical_pin` and `physical_pin`. I agree it is much clearer.

vpr/src/util/vpr_utils.cpp
```diff
@@ -2241,6 +2146,75 @@ int max_pins_per_grid_tile() {
...
+    if (result == direct_map.end()) {
+        VTR_LOG_WARN(
+            "Couldn't find the corresponding physical pin of the logical pin %d."
+            "Physical Tile: %s, Logical Block: %s.\n",
+            pin, physical_tile->name, logical_block->name);
+        return OPEN;
+    }
+
+    return result->second.pin;
+}
```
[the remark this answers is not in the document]
acomodi: I have added comments in the `physical_types.h` header file, Where the `t_logical_pin` and `t_physical_pin` structures are defined.

I believe that is the best place where to add their description. I think I will also add them somewhere in the documentation.

vtr_flow/arch/common/arch.xml
```diff
@@ -14,7 +14,12 @@
   <tiles>
     <tile name="ff_tile">
       <equivalent_sites>
-        <site pb_type="ff_tile"/>
+        <site pb_type="ff_tile">
+          <direct from="ff_tile.in" to="ff_tile.in"/>
+          <direct from="ff_tile.out" to="ff_tile.out"/>
+          <direct from="ff_tile.clk" to="ff_tile.clk"/>
+          <direct from="ff_tile.cen" to="ff_tile.cen"/>
+        </site>
```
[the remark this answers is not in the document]
acomodi: Done

libs/libarchfpga/src/physical_types.h
```diff
@@ -626,14 +610,52 @@ struct t_physical_tile_type {
 
     int index = -1; /* index of type descriptor in array (allows for index referencing) */
 
-    int logical_block_index = -1; /* index of the corresponding logical block type */
+    std::vector<std::string> equivalent_sites_names;
+    std::vector<t_logical_block_type_ptr> equivalent_sites;
```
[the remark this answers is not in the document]
acomodi: It could actually be eliminated. `equivalent_sites_names` was necessary to link physical tiles to logical blocks, but I have found a better way to do that which could let me avoid using the extra `equivalent_sites_names`

libs/libarchfpga/src/physical_types.h
```diff
@@ -626,14 +610,52 @@ struct t_physical_tile_type {
...
+
+    /* Map holding the priority for which this logical block needs to be placed.
+     * logical_blocks_priority[priority] -> vector holding the logical_block indices */
+    std::map<int, std::vector<int>> logical_blocks_priority;
```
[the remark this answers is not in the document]
acomodi: Fixed

vpr/src/place/initial_placement.cpp
kmurray: Since this would change the initial placement it means the placer would take a different walk through the space of possible placements, so I would expect the final placements (and hence QoR) to change somewhat.

So long as the new QoR is reasonable (e.g. on the VTR and Titan benchmarks) I'm OK if it changes.

acomodi: Ok, once I have performed all the changes I will go through another Titan benchmarks run and see the outcome.

There is one failure though of a strong regression test in CI which cannot route within 50 iterations: https://travis-ci.org/verilog-to-routing/vtr-verilog-to-routing/jobs/614569674?utm_medium=notification&utm_source=github_status. This seems odd, but I need to investigate what is happening there

vpr/src/util/vpr_utils.cpp
```diff
@@ -2241,6 +2146,75 @@ int max_pins_per_grid_tile() {
...
+
+t_physical_tile_type_ptr pick_random_physical_type(t_logical_block_type_ptr logical_block) {
+    auto equivalent_tiles = logical_block->equivalent_tiles;
+
+    size_t num_equivalent_tiles = equivalent_tiles.size();
+    int index = 0;
+
+    if (num_equivalent_tiles > 1) {
+        index = vtr::irand((int)equivalent_tiles.size() - 1);
+    }
```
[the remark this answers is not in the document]
acomodi: With https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/988/commits/e68d23eee42d3b59e0fa2e7a214a3498549f070b I have enabled the sorting of the equivalent tiles/sites list of physical and logical types based on the differences in the number of pins between two equivalent tiles of a certain logical block and viceversa.

Then, with the new `pick_best_physical/logical_type` the first element of the list is returned, corresponding to the best equivalent type/site. This way we can avoid the extensive usage of `pick_random_type`.

vpr/src/base/ShowSetup.cpp
```diff
@@ -74,15 +74,20 @@ void printClusteredNetlistStats() {
     L_num_p_outputs = 0;
 
     for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
-        num_blocks_type[cluster_ctx.clb_nlist.block_type(blk_id)->index]++;
-        auto type = physical_tile_type(blk_id);
-        if (is_io_type(type)) {
-            for (j = 0; j < type->num_pins; j++) {
+        auto logical_block = cluster_ctx.clb_nlist.block_type(blk_id);
+        auto physical_tile = pick_random_physical_type(logical_block);
```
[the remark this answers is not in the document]
acomodi: Done

vpr/src/base/check_netlist.cpp
```diff
@@ -91,14 +91,18 @@ static int check_connections_to_global_clb_pins(ClusterNetId net_id, int verbosi
     int global_to_non_global_connection_count = 0;
     for (auto pin_id : cluster_ctx.clb_nlist.net_pins(net_id)) {
         ClusterBlockId blk_id = cluster_ctx.clb_nlist.pin_block(pin_id);
-        int pin_index = cluster_ctx.clb_nlist.pin_physical_index(pin_id);
+        auto logical_type = cluster_ctx.clb_nlist.block_type(blk_id);
+        auto physical_type = pick_random_physical_type(logical_type);
```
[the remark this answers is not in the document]
acomodi: Done

vpr/src/place/initial_placement.cpp
acomodi: Done

vpr/src/place/uniform_move_generator.cpp
```diff
@@ -14,10 +14,13 @@ e_create_move UniformMoveGenerator::propose_move(t_pl_blocks_to_be_moved& blocks
...
     auto grid_from_type = g_vpr_ctx.device().grid[from.x][from.y].type;
-    VTR_ASSERT(physical_tile_type(cluster_from_type) == grid_from_type);
+    VTR_ASSERT(is_tile_compatible(grid_from_type, cluster_from_type));
 
     t_pl_loc to;
-    if (!find_to_loc_uniform(physical_tile_type(b_from), rlim, from, to)) {
+
+    auto type = pick_random_physical_type(cluster_from_type);
+
+    if (!find_to_loc_uniform(type, rlim, from, to)) {
```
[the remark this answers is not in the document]
acomodi: Agreed, I have swapped the `compress_grid_location` to use the `logical_block_types` indices, and, for each of them, I have added all the locations of the `equivalent_tiles`.

vpr/src/base/clustered_netlist.h
acomodi: I believe that, from a theoretical point of view you are completely right, the pin physical index should not be part of the clustered netlist, as it shouldn't be aware of the location chosen for a specific clb, hence its physical pins. 

From what I have learned, this information (physical pin related to a `ClusterPinId`) is stored in a fast-accessible structure. If we extract this in an external function, I believe that run-time will be affected. This because for each `ClusteredPinId` we would need to get many different information (adopted physical tile, logical pin, and physical pin from the physical/logical pin mapping).

I may be wrong and run-time will not be affected at all, therefore I will try to get rid of both `pin_physical_index`, `net_pin_physical_index` and `set_pin_physical_index` and extract them from the clustered netlist.

acomodi: Let me rephrase. Actually I think that after all this is doable as a lot of `pin_physical_index` can be swapped to `pin_logical_index`. Pin physical index will be needed only after placement only and there are not many places where it is used, so run-time should not be affected by this.

vpr/src/place/move_utils.cpp
```diff
@@ -497,27 +497,27 @@ bool find_to_loc_uniform(t_physical_tile_type_ptr type,
...
-    //Determin the valid compressed grid location ranges
+    //Determine the valid compressed grid location ranges
     int min_cx = std::max(0, cx_from - rlim_x);
-    int max_cx = std::min<int>(compressed_block_grid.compressed_to_grid_x.size() - 1, cx_from + rlim_x);
+    int max_cx = std::min<int>(to_compressed_block_grid.compressed_to_grid_x.size() - 1, cx_from + rlim_x);
     int delta_cx = max_cx - min_cx;
 
     int min_cy = std::max(0, cy_from - rlim_y);
-    int max_cy = std::min<int>(compressed_block_grid.compressed_to_grid_y.size() - 1, cy_from + rlim_y);
+    int max_cy = std::min<int>(to_compressed_block_grid.compressed_to_grid_y.size() - 1, cy_from + rlim_y);
```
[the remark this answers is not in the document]
acomodi: Done

vpr/src/place/move_utils.cpp
```diff
@@ -588,7 +588,33 @@ bool find_to_loc_uniform(t_physical_tile_type_ptr type,
...
+                    auto block_type = cluster_ctx.clb_nlist.block_type(blk);
+                    if (!is_tile_compatible(from_type, block_type)) {
+                        impossible_swap = true;
+                        break;
+                    }
+                }
+
+                if (!impossible_swap) {
+                    legal = true;
+                }
```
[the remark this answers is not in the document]
acomodi: Done, now the check is done when assessing the swap legality. This will prevent situations for which a block in the `to` location gets swapped in an illegal location.

vpr/src/base/clustered_netlist.h
acomodi: As predicted repeating the lookup on the direct map several times for each pin increases run-time of a non-negligible amount.

I have opted to store the `ClusterPinId <-> physical_pin_mapping` in the placement context. The mapping gets computed after the initial placement and modified at the end of the whole placement process, taking into account also the `z` location in tiles having `capacity > 1`.

I think placement context is a better place to store this kind of information.

acomodi:
@kmurray Except for the documentation and the initial placement issue for which I get different QoR (and on CI one test fails to route) when sorting the initial blocks, all the other comments should have been addressed.

I will deal now with updating the documentation, but I believe that the PR is ready for another review-pass.

acomodi:
@kmurray @vaughnbetz I have added documentation and a small tutorial on how to use the equivalent sites as well.

vpr/src/base/clustered_netlist.h
kmurray: Yes that seems like a better place for it!

vpr/src/place/place.cpp
```diff
@@ -1385,7 +1390,7 @@ static void update_net_bb(const ClusterNetId net,
         }
     } else {
         //For large nets, update bounding box incrementally
-        int iblk_pin = cluster_ctx.clb_nlist.pin_physical_index(blk_pin);
+        int iblk_pin = pin_tile_index(blk_pin);
```
kmurray: Since this is a free function (which is good), we probably want to give it a somewhat more descriptive name. `pin_tile_index` could be miss-understood to mean the tile index (within the grid) which contains the specified pin.

I think it really means the pin index of the tile corresponding to the blk_pin. Perhaps rename it to something like `tile_pin_index(ClusterPinId pin)` where the fact that it takes a ClusterPinId in combination with its name make sit clear it returns the pin index with respect to the tile.

vpr/src/place/place.cpp
```diff
@@ -1836,7 +1841,7 @@ static void get_bb_from_scratch(ClusterNetId net_id, t_bb* coords, t_bb* num_on_
     auto& grid = device_ctx.grid;
 
     ClusterBlockId bnum = cluster_ctx.clb_nlist.net_driver_block(net_id);
-    pnum = cluster_ctx.clb_nlist.net_pin_physical_index(net_id, 0);
+    pnum = net_pin_tile_index(net_id, 0);
```
kmurray: Similarly, maybe call this `net_pin_to_tile_pin_index()`?

vpr/src/route/route_common.cpp
```diff
@@ -1024,27 +1022,19 @@ static vtr::vector<ClusterNetId, std::vector<int>> load_net_rr_terminals(const t
...
              * are offset to get their actual rr_node */
-            node_block_pin = cluster_ctx.clb_nlist.pin_logical_index(pin_id);
-
-            int orig_phys_pin = get_physical_pin(type, logical_block, node_block_pin);
-
-            VTR_ASSERT(type->num_pins % type->capacity == 0);
-            int max_num_block_pins = type->num_pins / type->capacity;
-
-            int phys_pin = orig_phys_pin + place_ctx.block_locs[block_id].loc.z * max_num_block_pins;
+            int phys_pin = pin_tile_index(pin_id);
```
kmurray: It's great to see this cleaned-up nicely behind a a function!

vpr/src/util/vpr_utils.cpp
```diff
@@ -2105,24 +2105,34 @@ void print_switch_usage() {
...
+        if (result != place_ctx.physical_pins.end()) {
+            place_ctx.physical_pins[pin] = new_physical_pin_index;
+        } else {
+            place_ctx.physical_pins.insert(pin, new_physical_pin_index);
+        }
     }
 
     //Mark the block as synced
-    place_ctx.block_locs[iblk].nets_and_pins_synced_to_z_coordinate = true;
+    //place_ctx.block_locs[iblk].nets_and_pins_synced_to_z_coordinate = true;
```
kmurray: If this is unused, perhaps `place_ctx.block_locs[iblk].nets_and_pins_synced_to_z_coordinate` should be removed (along with this line)?

vpr/src/util/vpr_utils.cpp
```diff
@@ -2200,6 +2210,21 @@ int get_physical_pin(t_physical_tile_type_ptr physical_tile,
...
+
+    return pin_tile_index(pin_id);
+}
+
+int pin_tile_index(const ClusterPinId pin) {
+    auto& place_ctx = g_vpr_ctx.placement();
+
+    return place_ctx.physical_pins[pin];
+}
+
```
kmurray: We probably want some comments describing what these are doing (probably at the relevant header declaration)

vpr/src/util/vpr_utils.h
```diff
@@ -180,6 +180,9 @@ int get_physical_pin(t_physical_tile_type_ptr physical_tile,
                      t_logical_block_type_ptr logical_block,
                      int pin);
 
+int net_pin_tile_index(const ClusterNetId net_id, int net_pin_index);
+int pin_tile_index(const ClusterPinId pin);
```
kmurray: Since these are commonly used, should include give some comments describing what they do.

acomodi:
@kmurray Thanks for the review.

One note about the new changes:
I have deleted the usage of the following utility functions as their functionality was not required anymore:
    - `int find_clb_pb_pin(ClusterBlockId clb, int clb_pin);`
    - `int find_pb_pin_clb_pin(ClusterBlockId clb, int pb_pin);`

In fact they were returning different values depending on the `nets_and_pins_synced_to_z_coordinate` boolean which is currently unused.

vpr/src/util/vpr_utils.h
acomodi: Done

vpr/src/util/vpr_utils.cpp
acomodi: Done

vpr/src/util/vpr_utils.cpp
acomodi: Done

vpr/src/place/place.cpp
acomodi: Done

vpr/src/place/place.cpp
acomodi: Done

kmurray:
Great, thanks @acomodi !

PR #709, opened 2019-07-09
Add  convenience constructors for flat_map2

litghost:
Also convert forwarding version of constructors to require r-value
reference (and avoid potential copy).

PR #707, opened 2019-07-08
Provide structure for Router lookahead and placement delay lookup SerDes

litghost:
#### Description

This PR restructures the router lookahead and placement delay objects to enable several goals:
 - Properly encapsulate existing functionality behind virtual interfaces
 - Provide command line arguments and virtual methods to read and write the router lookahead and placement delay objects
 - Avoid reconstruction of the router lookahead object during placement delay lookup

Once this PR is merged, a followup PR will be submitted to implement the read/write methods with a proposed serialization format.

#### Motivation and Context

The router lookahead computation and placement delay matrix computations can take a fairly significant amount of time.  For example (take from one of the titan bench marks, ``stratixiv_arch.timing.xml/gsm_switch_stratixiv_arch_timing.blif``):

```
## Computing router lookahead map
## Computing router lookahead map took 862.39 seconds (max_rss 7998.8 MiB, delta_rss +0.0 MiB)
# Create Device took 1146.56 seconds (max_rss 7998.8 MiB, delta_rss +1742.6 MiB)

# Placement
## Computing placement delta delay look-up
RR graph channel widths unchanged, skipping RR graph rebuild
### Computing delta delays
### Computing delta delays took 5395.12 seconds (max_rss 17405.8 MiB, delta_rss +9407.0 MiB)
## Computing placement delta delay look-up took 5396.87 seconds (max_rss 17405.8 MiB, delta_rss +9407.0 MiB)

...

VPR suceeded
The entire flow of VPR took 16719.29 seconds (max_rss 19437.3 MiB)
```

In this example, the router lookahead map computation took ~5% of the total runtime, and the delay matrix lookup took ~33% of the total runtime.  In terms of wall clock time, taking ~2 hrs before starting placement is less than ideal, especially if parameter tuning is being attempted.

#### How Has This Been Tested?

In theory this change is simply a refactor, so existing tests should pass with zero or minimal runtime impact.

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #989, opened 2019-10-02
Initial place matrix serialization/deserialization support.

litghost:
#### Description

Bind ``--read_placement_delay_lookup`` and ``--write_placement_delay_lookup`` to SerDes (implemented using Cap’n Proto).

#### Motivation and Context

See justification in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/707

This PR has predecessor PRs:
 - #707 
 - #708 
 - #709 

Unit testing on SerDes, and some integration testing with Symbiflow.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libvtrcapnproto/README.md
+++ b/libs/libvtrcapnproto/README.md
@@ -1,0 +1,84 @@
+Capnproto usage in VTR
+======================
+
+Capnproto is a data serialization framework designed for portabliity and speed.
+In VPR, capnproto is used to provide binary formats for internal data
+structures that can be computed once, and used many times.  Specific examples:
+ - rrgraph
+ - Router lookahead data
+ - Place matrix delay estimates
+
+What is capnproto?
+==================
+
+capnproto can be broken down into 3 parts:
+ - A schema language
+ - A code generator
+ - A library
+
+The schema language is used to define messages.  Each message must have an
+explcit capnproto schema, which are stored in files suffixed with ".capnp".
+The capnproto documentation for how to write these schema files can be found
+here: https://capnproto.org/language.html
+
+The schema by itself is not especially useful.  In order to read and write
+messages defined by the schema in a target language (e.g. C++), a code
+generation step is required.  Capnproto provides a cmake function for this
+purpose, `capnp_generate_cpp`.  This generates C++ source and header files.
+These source and header files combined with the capnproto C++ library, enables
+C++ code to read and write the messages matching a particular schema.  The C++
+library API can be found here: https://capnproto.org/cxx.html
+
+Contents of libvtrcapnproto
+===========================
+
+libvtrcapnproto should contain two elements:
+ - Utilities for working capnproto messages in VTR
+ - Generate source and header files of all capnproto messages used in VTR
+
+I/O Utilities
+-------------
+
+Capnproto does not provide IO support, instead it works from arrays (or file
+descriptors).  To avoid re-writing this code, libvtrcapnproto provides two
+utilities that should be used whenever reading or writing capnproto message to
+disk:
+ - `serdes_utils.h` provides the writeMessageToFile function - Writes a
+   capnproto message to disk.
+ - `mmap_file.h` provides MmapFile object - Maps a capnproto message from the
+   disk as a flat array.
+
+NdMatrix Utilities
+------------------
+
+A common datatype which appears in many data structures that VPR might want to
+serialize is the generic type `vtr::NdMatrix`.  `ndmatrix_serdes.h` provides
+generic functions ToNdMatrix and FromNdMatrix, which can be used to generically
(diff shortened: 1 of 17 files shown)


libs/libvtrcapnproto/README.md
```diff
@@ -0,0 +1,84 @@
+Capnproto usage in VTR
```
kmurray: This is a great overview of the tool, how it fits together, and how to extend it in the future. Thanks!

libs/libvtrcapnproto/matrix.capnp
```diff
@@ -0,0 +1,9 @@
+@0xafffc9c1f309bc00;
+
```
kmurray: I assume we can add comments to the schema files?

It would be good to give a comment which tells the reader this corresponds to serializing a vtr::NdMatrix. It would also be good to comment what the various fields represent so someone less familiar with capnp could get the jist of it.

vpr/src/place/place_delay_model.h
```diff
@@ -53,16 +52,14 @@ class DeltaDelayModel : public PlaceDelayModel {
...
     void dump_echo(std::string filepath) const override;
 
-    void read(const std::string& /*file*/) override {
-        VPR_THROW(VPR_ERROR_ROUTE, "DeltaDelayModel::read unimplemented");
-    }
-    void write(const std::string& /*file*/) const override {
-        VPR_THROW(VPR_ERROR_ROUTE, "DeltaDelayModel::write unimplemented");
+    void read(const std::string& file) override;
+    void write(const std::string& file) const override;
+    const vtr::Matrix<float>& delays() const {
```
kmurray: Is it necessary to expose this implementation detail? Part of the motivation for making the PlaceDelayModel interface was to hide details like this.

vpr/src/place/place_delay_model.h
```diff
@@ -75,19 +72,20 @@ class OverrideDelayModel : public PlaceDelayModel {
...
 
   public: //Mutators
+    void set_base_delay_model(std::unique_ptr<DeltaDelayModel> base_delay_model);
+    const DeltaDelayModel* base_delay_model() const;
+    float get_delay_override(int from_type, int from_class, int to_type, int to_class, int delta_x, int delta_y) const;
+    void set_delay_override(int from_type, int from_class, int to_type, int to_class, int delta_x, int delta_y, float delay);
+
   private:
-    std::unique_ptr<PlaceDelayModel> base_delay_model_;
+    std::unique_ptr<DeltaDelayModel> base_delay_model_;
```
kmurray: Is it necessary to change this from the abstract PlaceDelayModel to DeltaDelayModel? The idea was that any (potentially non-delta delay model) could be combined with this class to provide fix-up overrides.

libs/libvtrcapnproto/ndmatrix_serdes.h
```diff
@@ -0,0 +1,87 @@
...
+//  N = Number of matrix dimensions, must be fixed.
+//  CapType = Source capnproto message type that is a single element the
+//            Matrix capnproto message.
+//  CType = Target C++ type that is a single element of vtr::NdMatrix.
+//
+// Arguments:
+//  m_out = Target vtr::NdMatrix.
+//  m_in = Source capnproto message reader.
+//  copy_fun = Function to convert from CapType to CType.
+//
```
kmurray: Good comment which makes the code much easier to follow.

It might be helpful to have examples of how someone would call ToNdMatrix and FromNdMatrix included in the comments. This would make it clearer how someone would use these functions.

vpr/src/place/place_delay_model.h
litghost: Serialization would require a mirroring of the inner model type.  We could create a string registry of types, and then serialize the string.  However, I didn't want to take that additional complexity right now given how few models we have.

Do we want to add such a registry now?

vpr/src/place/place_delay_model.h
litghost: The reason was ease of implementation of OverrideDelayModel serialization, see https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/989#discussion_r330758264

I know how to remove this method if needed, but it is really about what to do about https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/989#discussion_r330758264 that changes what the "solution" is here.

libs/libvtrcapnproto/matrix.capnp
litghost: Sure

libs/libvtrcapnproto/ndmatrix_serdes.h
litghost: Done

kmurray:
@litghost Thanks for addressing all the comments!

PR #992, opened 2019-10-07
vpr: pass t_placer_opts in binary_search_place_and_route by reference

renrichard:
#### Description

Coverity has reported a defect in the build.

#### Related Issue

** CID 202040:  Performance inefficiencies  (PASS_BY_VALUE)
/home/travis/build/verilog-to-routing/vtr-verilog-to-routing/vpr/src/base/place_and_route.cpp: 46 in binary_search_place_and_route(t_placer_opts, const t_annealing_sched &, const t_router_opts &, const t_analysis_opts &, const t_file_name_opts &, const t_arch *, bool, int, t_det_routing_arch *, std::vector<t_segment_inf, std::allocator<t_segment_inf>> &, vtr::vector<vtr::StrongId<cluster_net_id_tag, int, (int)-1>, float *> &, std::shared_ptr<SetupHoldTimingInfo>, std::shared_ptr<PostClusterDelayCalculator>)()

#### Motivation and Context

We do not need to pass the entire t_placer_opts into the function, its reference shall suffice.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/place_and_route.cpp
+++ b/vpr/src/base/place_and_route.cpp
@@ -43,7 +43,7 @@ static float comp_width(t_chan* chan, float x, float separation);
 
 /************************* Subroutine Definitions ****************************/
 
-int binary_search_place_and_route(t_placer_opts placer_opts,
+int binary_search_place_and_route(const t_placer_opts& placer_opts_ref,
                                   const t_annealing_sched& annealing_sched,
                                   const t_router_opts& router_opts,
                                   const t_analysis_opts& analysis_opts,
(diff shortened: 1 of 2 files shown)


kmurray:
@renrichard Thanks for taking a look at this!

My only comment is that we should prefer passing by const reference (e.g. `const t_placer_opts&`), since it enforces that `binary_search_place_and_route()` can't change the overall placement settings.

renrichard:
In response to your suggestion, I now realize that because t_placer_opts was being passed by value, the channel width was only being modified within the scope of the function binary_search_place_and_route  (see line 166).

166 placer_opts.place_chan_width = current;

Thus, I have taken your suggestion into account.

kmurray:
@renrichard Looks like since the placer opts are mutated by the function (channel width as you pointed out) passing by const reference causes a compilation error.

We could either leave the code as-is (pass by value), or make a copy of it right-before modification (around like 166) and pass the modified copy to try_place() (which should make coverity happy).

renrichard:
@kmurray, I have updated the branch.

kmurray:
Thanks @renrichard!

PR #993, opened 2019-10-07
genfasm placeholders

mkurc-ant:
This PR adds support for placeholders (or tags) that can be specified in fasm features/prefixes. Those placeholders are then substituted with given values during fasm generation.

#### Description
Values of the placeholders can be defined at the topmost level only - in the `<layout>` section of the architecture XML file via the `fasm_placeholders` metadata entry.

Inside the metadata entry, a placeholder is defined in assigned with a value in the following manner. One placeholder per line:
```
<placeholder name> = <placeholder value>
```

For example, top level tiles in the grid can now be specified as:
```
<single type="clb" priority="1" x="10" y="20">
  <metadata>
    <meta fasm_placeholders>
      LOC = X10Y20
      LR = R
    </meta>
  </metadata>
</single>
<single type="clb" priority="1" x="10" y="21">
  <metadata>
    <meta fasm_placeholders>
      LOC = X10Y21
      LR = R
    </meta>
  </metadata>
</single>
```
where `LOC` and `LR` are placeholder names with assigned values.

A common pb_type for both tiles can be defined as:
```
<pb_type>
  ...
  <metadata>
    <meta name="fasm_prefix">
    {LR}IOB_{LOC}
    </meta>
  </metadata>
</pb_type>
```

Then, during fasm output generation those placeholders will be replaced with values defined at the grid definition level.

#### Motivation and Context
The change allows to use the same pb_type definition for different tile types by allowing to control content of fasm features/prefixes at the pb_type level. In another words: a pb_type can have multiple fasm features defined that correspond to tiles with different prefixes.

There is a test suite for genfasm along with an example arch XML. This test was augmented to verify operation of the implemented placeholder mechanism.

#### Types of changes

#### Checklist:


mkurc-ant:
@kmurray @litghost Please take a look.

utils/fasm/test/test_fasm.cpp
```diff
@@ -4,16 +4,173 @@
...
+        {"site", "SLICE"}
+    };
+
+    REQUIRE_THROWS(fasm::substitute_tags(feature, tags));
+}
+
+// ============================================================================
+
+/*
+VPR may rearrange LUT inputs and modify its content accordingly. This function
```
litghost: You shouldn't need to brute force the permutations, as VPR explicitly indicates how the LUTs are rotated.  See https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/utils/fasm/src/fasm.cpp#L316

utils/fasm/test/test_fasm.cpp
```diff
@@ -4,16 +4,173 @@
...
+#include <regex>
+#include <cmath>
+#include <algorithm>
 
 static constexpr const char kArchFile[] = "test_fasm_arch.xml";
 static constexpr const char kRrGraphFile[] = "test_fasm_rrgraph.xml";
 
+// Prototype here to be able to test it
+namespace fasm {
+std::vector<std::string> find_tags_in_feature (const std::string& a_String);
```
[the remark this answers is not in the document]
mkurc-ant: Fixed.

utils/fasm/test/test_fasm.cpp
mkurc-ant: I know that the permutation can be extracted from VPR and this is what the genfasm does. But I see no point in duplicating the LUT extraction code (being tested) just for the purpose of the test. As in the end we might not know where a potential bug is (in the code or in the test).

litghost:
@kmurray This is good to merge.

utils/fasm/test/test_fasm.cpp
kmurray: We've relied on our logical equivalency check integration tests (strong/strong_func*) to verify that the LUT permutations VTR has stored are correct. These tests use ABC to verify the input netlist to VPR and the post-synthesis netlist (generated by VPR which has the input rotations applied) are equivalent.

kmurray:
Thanks for the enhancement @mkurc-ant, and for the code review @litghost!

PR #994, opened 2019-10-08
Add debug logging if there is a non-finite value from the lookahead.

litghost:
#### Description

When debugging lookaheads, a common problem early on is simply the lookahead failing to find valid values (e.g. finite).  This adds a small debug log message if a non-finite lookahead is detected and router debugging is enabled.

#### Motivation and Context

This is simply a debugging aid.

Compiled and checked the output on a failure.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -2033,6 +2033,11 @@ static void evaluate_timing_driven_node_costs(t_heap* to,
 
     //Update total cost
     float expected_cost = router_lookahead.get_expected_cost(to_node, target_node, cost_params, to->R_upstream);
+    VTR_LOGV_DEBUG(f_router_debug && !std::isfinite(expected_cost),
+                   "        Lookahead from %s (%s) to %s (%s) is non-finite, expected_cost = %f, to->R_upstream = %f\n",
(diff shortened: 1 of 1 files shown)


vpr/src/route/route_timing.cpp
```diff
@@ -2033,6 +2033,15 @@ static void evaluate_timing_driven_node_costs(t_heap* to,
 
     //Update total cost
     float expected_cost = router_lookahead.get_expected_cost(to_node, target_node, cost_params, to->R_upstream);
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+    if (f_router_debug && !std::isfinite(expected_cost)) {
+        VTR_LOGV_DEBUG(f_router_debug,
```
kmurray: I don't think we need the VTR_ENABLE_DEBUG_LOGGING define guard, since the guard is already included in the [VTR_LOGV_DEBUG macro](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/libvtrutil/src/vtr_log.h#L48-L55) (which is intentional to avoid #ifdefs all over the place).

It should be equivalent (and cleaner) to just do:
```
VTR_LOGV_DEBUG(f_router_debug && !std::isfininte(expected_cost),
```
and drop the include guard

litghost: Done

PR #996, opened 2019-10-09
simulation mem leak - latch names array in create latch

sdamghan:
tested by make test odin

#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/SRC/read_blif.cpp
+++ b/ODIN_II/SRC/read_blif.cpp
@@ -368,12 +368,12 @@ void create_latch_node_and_driver(FILE *file, Hashtable *output_nets_hash)
 
 	/* add names and type information to the created input pins */
 	npin_t *new_pin = allocate_npin();
-	new_pin->name = names[0];
+	new_pin->name = vtr::strdup(names[0]);
 	new_pin->type = INPUT;
 	add_input_pin_to_node(new_node, new_pin,0);
 
 	new_pin = allocate_npin();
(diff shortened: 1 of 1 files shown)


ODIN_II/SRC/read_blif.cpp
```diff
@@ -399,6 +399,11 @@ void create_latch_node_and_driver(FILE *file, Hashtable *output_nets_hash)
 	output_nets_hash->add(new_node->name, new_net);
 
 	/* Free the char** names */
+	for (i = 0; i < input_token_count; i++)
+		// Except these two since they were assigned to new_node
+		if (i!=0 && i!=3)
+			vtr::free(names[i]);
+
```
jeanlego: do strdup at line 371 and line 376 rather and here simply free them all
avoid complicated logic like this at all cost, change may occur in the code base and this would produce some unexpected side effect down the line

ODIN_II/SRC/read_blif.cpp
```diff
@@ -399,6 +399,10 @@ void create_latch_node_and_driver(FILE *file, Hashtable *output_nets_hash)
 	output_nets_hash->add(new_node->name, new_net);
 
 	/* Free the char** names */
+	for (i = 0; i < input_token_count; i++)
+		// Except these two since they were assigned to new_node
```
jeanlego: remove that comment since it is not relevant anymore

PR #997, opened 2019-10-09
Simulation mem leak - create hard block mappings

sdamghan:
tested by make test odin

#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/SRC/read_blif.cpp
+++ b/ODIN_II/SRC/read_blif.cpp
@@ -536,14 +536,13 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
 
 	for (i = 0; i < count; i++)
 	{
-		if (mappings[i] != NULL)
-			mappings[i] = NULL;
-
 		vtr::free(mappings[i]);
+		mappings[i] = NULL;
 	}
 	
-	mappings = NULL;
 	vtr::free(mappings);
+	mappings = NULL;
+
 
 	// Look up the model in the models cache.
  	hard_block_model *model = NULL;
--- a/ODIN_II/SRC/read_blif.cpp
+++ b/ODIN_II/SRC/read_blif.cpp
@@ -640,6 +640,9 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
   	free_hard_block_ports(ports);
   	mapping_index->destroy_free_items();
 	delete mapping_index;
+	for (i = 0; i < count; i++)
+		vtr::free(mappings[i]);
+		
   	vtr::free(mappings);
   	vtr::free(names);
 
--- a/ODIN_II/SRC/read_blif.cpp
+++ b/ODIN_II/SRC/read_blif.cpp
@@ -649,10 +649,7 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
 
   	free_hard_block_ports(ports);
   	mapping_index->destroy_free_items();
-	delete mapping_index;
-	for (i = 0; i < count; i++)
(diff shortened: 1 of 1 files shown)


ODIN_II/SRC/read_blif.cpp
```diff
@@ -640,6 +640,9 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
   	free_hard_block_ports(ports);
   	mapping_index->destroy_free_items();
 	delete mapping_index;
+	for (i = 0; i < count; i++)
+		vtr::free(mappings[i]);
+
```
jeanlego: move up to line 536

sdamghan: the line 540 uses mapping array for "get_hard_block_ports", we can't free them before this line.

jeanlego: get_hard_block_ports is at line 535, hence why i think the freeing should be moved ate 536

ODIN_II/SRC/read_blif.cpp
```diff
@@ -534,6 +534,9 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
 	// Index the mappings in a hard_block_ports struct.
 	hard_block_ports *ports = get_hard_block_ports(mappings, count);
 
+	for (i = 0; i < count; i++)
+		vtr::free(mappings[i]);
+
```
jeanlego: make sure you set mappings[i] = NULL in the for loop.

jeanlego: also free the mappings array here and set to NULL

ODIN_II/SRC/read_blif.cpp
jeanlego: once moved, please remove all the occurence of mappings down here

ODIN_II/SRC/read_blif.cpp
```diff
@@ -535,7 +535,15 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
 	hard_block_ports *ports = get_hard_block_ports(mappings, count);
 
 	for (i = 0; i < count; i++)
+	{
+		if (mappings[i] != NULL)
```
jeanlego: you just lost your pointer there.

set to NULL after free,
don't check if its NULL before setting to NULL, its a bit redundant

ODIN_II/SRC/read_blif.cpp
```diff
@@ -535,7 +535,15 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
...
+			mappings[i] = NULL;
+
 		vtr::free(mappings[i]);
+	}
+	
+	mappings = NULL;
```
jeanlego: same here

ODIN_II/SRC/read_blif.cpp
jeanlego: https://www.eskimo.com/~scs/cclass/notes/sx11b.html

ODIN_II/SRC/read_blif.cpp
```diff
@@ -640,7 +650,9 @@ void create_hard_block_nodes(hard_block_models *models, FILE *file, Hashtable *o
   	free_hard_block_ports(ports);
   	mapping_index->destroy_free_items();
 	delete mapping_index;
-  	vtr::free(mappings);
+	for (i = 0; i < count; i++)
```
jeanlego: Remove these line (mapping is set to NULL, you can't access anything inside it. This is why I've made you free above and set to NULL)

PR #998, opened 2019-10-09
Add tile filtering option for placer delay matrix sampling.

litghost:
#### Description

This is useful if the placer delay sampling method tries to sample clock
network tiles (e.g. PLLs, global/local clock buffers) which will result
in a poor delay matrix which is unsuitable for LUTs/FFs/RAMs.

#### Motivation and Context

If the delay matrix sampling algorithm chooses a specialized tile (e.g. PLL, global/local clock buffers) it effectively builds the delay matrix for the specialized network, instead of the general routing fabric.  This option prevents this from being possible.

Need to finish testing with variant, but a simpler version of this PR was tested with Symbiflow.

Testing remaining to be done:
- [ ] Make sure CI is green
- [ ] Add test for new flag

#### Types of changes

#### Checklist:


kmurray:
Seems like a reasonable enhancement!

PR #1007, opened 2019-10-16
Fix bug in add_subtree_to_route_tree that can result in double free.

litghost:
#### Description

Previous code potentially added a rt_node to two linked_rt_edge's,
resulting in a double free when freeing the tree.

#### Motivation and Context

Previous code would segfault in some circumstances.

1. Segfaulting case no longer segfaults
2. CI is green

#### Types of changes

#### Checklist:


kmurray:
Thanks for the fix!

PR #1008, opened 2019-10-16
Use Dijkstra to generate place delay matrix

litghost:
#### Description

`calculate_all_path_delays_from_rr_node` and `timing_driven_find_all_shortest_paths_from_route_tree` were intended to be used in the place delay matrix computation, but were never implemented (see https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/99f91ffbba6d9f5ef40b00cb83d8a126b3a1ad15).  This PR attempts to replace `route_connection_delay` with `calculate_all_path_delays_from_rr_node`.

#### Related Issue

- https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1071

#### Motivation and Context

Place matrix delay computation for larger grids is a large term.  On the `A50T` 7-series graph, it was taking an extremely long time (>24 hrs) and on the Titan benchmarks it was taking 4202.33 seconds (see https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1071#issuecomment-542398926).  The graph expansion method should return better data, in less time, than the `width*height` A* search that was present before.

- [x] CI is green
- [ ] Check Titan QoR

#### Types of changes

#### Checklist:


litghost:
Note that #1007 is included in this PR, as https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1008/commits/128693c4b0f40237c91f7c59236a5e628f9e15a9 exposed it.

litghost:
I've kicked off Titan QoR, results will likely show up by Friday.

litghost:
CI failure is a segfault, investigating.

litghost:
> CI failure is a segfault, investigating.

There appears to have been a long standing bug in the place delay code, fixed now with https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1008/commits/8b1a6757dd8580bbf878a3f18d600dfde5f7f22e

litghost:
CI is now passing except for 2 strong tests that have golden violations:
```
regression_tests/vtr_reg_strong/strong_custom_grid...[Fail] 
 non_column_tall_aspect_ratio.xml/raygentop.v/common min_chan_width: golden = 44 result = 58
[Fail] 
 non_column_tall_aspect_ratio.xml/raygentop.v/common min_chan_width_routing_area_total: golden = 2.76853e+06 result = 3.61431e+06
[Fail] 
 non_column_tall_aspect_ratio.xml/raygentop.v/common min_chan_width_routing_area_per_tile: golden = 2616.76 result = 3416.17
[Fail] 
 multiple_io_types.xml/raygentop.v/common place_time: golden = 10.79 result = 119.76

regression_tests/vtr_reg_strong/strong_multiclock...[Fail] 
 k6_frac_N10_mem32K_40nm.xml/multiclock.blif/common crit_path_delay_mcw: golden = 1.06954 result = 1.46003
[Fail] 
 k6_frac_N10_mem32K_40nm.xml/multiclock.blif/common clk2_to_output_cpd: golden = 1.06954 result = 1.46003
[Fail] 
 k6_frac_N10_mem32K_40nm.xml/multiclock.blif/common clk2_to_output_hold_slack: golden = 0.918173 result = 1.30866
```

The QoR results from the Titan benchmarks will show whether this is noise, or a regression.  @kmurray / @vaughnbetz Agreed?

vaughnbetz:
I agree; the Titan results are a better indicator of qor.  The multiclock design is very small and isn't really intended as a QoR test.
raygentop.v's placement time went up by 10x though on the multiple I/O types regtest -- any idea why?  May indicate some cpu time issue with the new code on an architecture with multiple I/O types.

Which Titan benchmark took 4200 s to generate the placement delay matrix?  That certainly sounds high; I think we get significantly lower numbers than that (machine configuration difference?).

litghost:
> Which Titan benchmark took 4200 s to generate the placement delay matrix?

Gaussian blur, which is the largest of the Titan benchmarks.

litghost:
@vaughnbetz So for example, the placement/lookahead cost from the baseline commit (51c5513da5925df425f6f6ff1fb5b2c74df21d4b) is:
```
# Placement
## Computing placement delta delay look-up
RR graph channel widths unchanged, skipping RR graph rebuild
### Computing router lookahead map
### Computing router lookahead map took 1662.57 seconds (max_rss 22892.1 MiB, delta_rss +0.0 MiB)
### Computing delta delays
### Computing delta delays took 8281.94 seconds (max_rss 67141.9 MiB, delta_rss +44249.8 MiB)
## Computing placement delta delay look-up took 9946.57 seconds (max_rss 67141.9 MiB, delta_rss +44249.8 MiB)
```

Whereas the new code (a208c2b5912892d5d0b0b7c4bf17b4f4bd2d71bd), same test:
```
## Computing placement delta delay look-up
RR graph channel widths unchanged, skipping RR graph rebuild
### Computing router lookahead map
### Computing router lookahead map took 1629.68 seconds (max_rss 22890.7 MiB, delta_rss +0.0 MiB)
### Computing delta delays
### Computing delta delays took 3758.11 seconds (max_rss 22890.7 MiB, delta_rss +0.0 MiB)
## Computing placement delta delay look-up took 5389.49 seconds (max_rss 22890.7 MiB, delta_rss +0.0 MiB)
```

The tests are still running, so we don't know if the new code preserves QoR, but you can see that the new code takes less time and less memory!

vaughnbetz:
Yes, that is promising.  Are you running with astar_fac = 1 (Kevin mentioned this was how he was running high-effort Titan, and it slows down the place delay cache creator).  3758 s is still a lot, so maybe there's more room still for improvement ...

litghost:
> Yes, that is promising. Are you running with astar_fac = 1 (Kevin mentioned this was how he was running high-effort Titan, and it slows down the place delay cache creator). 3758 s is still a lot, so maybe there's more room still for improvement ...

Yes, I'm just using the default titan regression test settings, which does include astar_fac = 1.

litghost:
QoR are very mixed, likely need to investigate some edges:
[comp.xlsx.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/3752479/comp.xlsx.zip)

Overall, the memory wins seem weird, I wonder if there is a memory leak in the old code where a vector needs to be destroyed?

The overall runtime is variable. which is unexpected.  One thing I know is that if the place delay expansion code has a persistent x/y -> x/y pair that cannot be found, it may result in an excessive number of expansions (see https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1008/files#diff-154b820343b39f67db4328801f31dbbcR398 ).  It might be worth while to fall back to the "route a net" method after the first expansion if the number of dead x/y coordinates is low, because it will be faster to prove no connection than the force expansion all source ptcs.

arch | circuit | baseline | expand_place_delay | ratio
-- | -- | -- | -- | --
stratixiv_arch.timing.xml | gaussianblur_stratixiv_arch_timing.blif | 152482.33 | 142279.07 | 93.31%
stratixiv_arch.timing.xml | bitcoin_miner_stratixiv_arch_timing.blif | 38495.8 | 37875.08 | 98.39%
stratixiv_arch.timing.xml | directrf_stratixiv_arch_timing.blif | 56747.62 | 50617.24 | 89.20%
stratixiv_arch.timing.xml | neuron_stratixiv_arch_timing.blif | 936.75 | 801.76 | 85.59%
stratixiv_arch.timing.xml | sparcT1_core_stratixiv_arch_timing.blif | 1053.75 | 1217.89 | 115.58%
stratixiv_arch.timing.xml | stereo_vision_stratixiv_arch_timing.blif | 873.37 | 780.99 | 89.42%
stratixiv_arch.timing.xml | cholesky_mc_stratixiv_arch_timing.blif | 1174.35 | 1182.25 | 100.67%
stratixiv_arch.timing.xml | LU230_stratixiv_arch_timing.blif | 43609.43 | 52887.29 | 121.27%
stratixiv_arch.timing.xml | gsm_switch_stratixiv_arch_timing.blif | 9814.92 | 9496.95 | 96.76%
stratixiv_arch.timing.xml | mes_noc_stratixiv_arch_timing.blif | 13847.06 | 14719.79 | 106.30%
stratixiv_arch.timing.xml | dart_stratixiv_arch_timing.blif | 2319.49 | 2633.32 | 113.53%
stratixiv_arch.timing.xml | denoise_stratixiv_arch_timing.blif | 7581.77 | 8977.48 | 118.41%
stratixiv_arch.timing.xml | sparcT2_core_stratixiv_arch_timing.blif | 6710.8 | 7843.97 | 116.89%
stratixiv_arch.timing.xml | cholesky_bdti_stratixiv_arch_timing.blif | 3471.44 | 3397.22 | 97.86%
stratixiv_arch.timing.xml | minres_stratixiv_arch_timing.blif | 4671.11 | 4024.19 | 86.15%
stratixiv_arch.timing.xml | stap_qrd_stratixiv_arch_timing.blif | 3929 | 4540.49 | 115.56%
stratixiv_arch.timing.xml | openCV_stratixiv_arch_timing.blif | 3893.46 | 3555.04 | 91.31%
stratixiv_arch.timing.xml | sparcT1_chip2_stratixiv_arch_timing.blif | 47007.75 | 49880.08 | 106.11%
stratixiv_arch.timing.xml | bitonic_mesh_stratixiv_arch_timing.blif | 6038.87 | 4969.32 | 82.29%
stratixiv_arch.timing.xml | segmentation_stratixiv_arch_timing.blif | 2373.68 | 2712.15 | 114.26%
stratixiv_arch.timing.xml | LU_Network_stratixiv_arch_timing.blif | 15180.82 | 15529.25 | 102.30%
stratixiv_arch.timing.xml | SLAM_spheric_stratixiv_arch_timing.blif | 1416.18 | 1930.63 | 136.33%
stratixiv_arch.timing.xml | des90_stratixiv_arch_timing.blif | 2449.83 | 2442.01 | 99.68%

vaughnbetz:
Those are pretty mixed.  critical path delay and wirelength are about the same (good) but place time is mostly unchanged and route time is up 16%.  The placer thinks the circuits are a little (4%) faster too.

Can you also parse out the actual place_delay_matrix and lookahead builder times?  Maybe they'll shed some light on what's going on.

Two possible theories to check, without a lot of evidence for each (just brainstorming ideas here):

One theory: is the router lookahead builder getting built twice with the new code, or there is less re-used data?
Another theory: is the router lookahead built using the new code as well and it's a little more optimistic about delay (suggestive data: placer thinks the circuits are faster) and that's costing CPU as it would make the router a bit more prone to back-tracking?

kmurray:
As a side note, regardless of the outcome of this PR, it looks like the fix in 8b1a675 should be merged.

litghost:
> As a side note, regardless of the outcome of this PR, it looks like the fix in [8b1a675](https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/8b1a6757dd8580bbf878a3f18d600dfde5f7f22e) should be merged.

Agreed.

vpr/src/base/read_options.cpp
```diff
@@ -1378,6 +1378,10 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .help("Displays delay statistics even if placement is not timing driven")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    place_grp.add_argument<bool, ParseOnOff>(args.use_expansion_for_delay_matrix, "--use_expansion_for_delay_matrix")
```
kmurray, line 1381: I'd suggest changing the definition of this option to something like: '--place_delta_delay_matrix_calculation_method {astar, dijkstra}'.

That seems to better capture what this option is trying to do (and leaves the door open for other methods to be added in the future).

vpr/src/place/timing_place_lookup.cpp
```diff
@@ -77,7 +77,20 @@ static float route_connection_delay(
...
+    int source_y,
+    int start_x,
+    int start_y,
+    int end_x,
+    int end_y,
+    const t_router_opts& router_opts,
+    bool measure_directconnect,
+    const std::set<std::string>& allowed_types);
+
+static void generic_compute_matrix_with_expansion(
```
kmurray, line 93: I think 'with_router' and 'with_expansion' don't really clearly describe the differences between these two approaches.

I'd suggest something like: `generic_compute_matrix_iterative_astar()` and `generic_compute_matrix_dijkstra_flood()`.

vpr/src/place/timing_place_lookup.cpp
```diff
@@ -358,7 +371,152 @@ static float route_connection_delay(
...
+                            delta_x, delta_y,
+                            source_x, source_y,
+                            sink_x, sink_y);
+#endif
+                }
+            }
+        }
+
+        return;
+    }
```
kmurray, line 425: I think we need some comments discussion what this code block is doing. It looks like it's trying to initialize only a subset of deltas with EMPTY_DELTA, but I'm not sure beyond that.

vpr/src/place/timing_place_lookup.cpp
```diff
@@ -358,7 +371,152 @@ static float route_connection_delay(
...
+
+    vtr::Matrix<bool> found_matrix({matrix.dim_size(0), matrix.dim_size(1)}, false);
+
+    auto best_driver_ptcs = get_best_classes(DRIVER, device_ctx.grid[source_x][source_y].type);
+    for (int driver_ptc : best_driver_ptcs) {
```
kmurray, line 430: Similarly here, we should try to capture in a comment what this code block is doing.

For instance, it isn't immediately clear why we loop through all the potential driver PTCs.

vpr/src/place/timing_place_lookup.cpp
```diff
@@ -505,6 +669,25 @@ static vtr::Matrix<float> compute_delta_delays(
...
+        int,
+        int,
+        int,
+        int,
+        int,
+        int,
+        const t_router_opts&,
+        bool,
+        const std::set<std::string>&)>
+        generic_compute_matrix;
```
kmurray, line 684: We don't use std::function very much (or at all?) in VPR. I think this should have a comment outlining what it is for and what its doing.

I think its so you can alias generic_compute_matrix to the relevant underlying function.  It might be clearer if we created a typedef (e.g. t_generic_compute_matrix_func).

Any notes on why you chose this approach (e.g. vs defining a generic compute matrix class)?

vpr/src/base/read_options.cpp
litghost: Done

vpr/src/place/timing_place_lookup.cpp
litghost: This code is actually a directly translation of the behavior of generic_compute_matrix_iterative_astar, but what happens when you use a dijstra expansion.  I wanted to minimize the behaviorial differences between the iterative astar and dijstra expansion methods.

vpr/src/place/timing_place_lookup.cpp
litghost: This code is actually a directly translation of the behavior of generic_compute_matrix_iterative_astar, but what happens when you use a dijstra expansion.  I wanted to minimize the behaviorial differences between the iterative astar and dijstra expansion methods.

vpr/src/place/timing_place_lookup.cpp
litghost: `defining a generic compute matrix class` - You mean an interface class?  That seems overkill for an interface with 1 method and little state?

litghost: > We don't use std::function very much (or at all?) in VPR

Sorry,  are you requesting a comment to explain what std::function is doing?  I want to be clear.

kmurray: After reflecting on this more, I'm leaning towards just typedef'ing the std::function at the top of the file, and instantiating it here. Thoughts?

litghost: Done

kmurray:
Thanks the code looks good to merge now.

I think the only item left would be to add a regression test to the `--place_delta_delay_matrix_calculation_method` options. The goal of such a test would be code coverage to detect if something related to this broke (rather than run-time QoR), so I'd probably make a `strong_place_delay_calc_method` test (base it off something quick like `strong_place_delay_model`).

litghost:
> I think the only item left would be to add a regression test to the `--place_delta_delay_matrix_calculation_method` options

Yep, done!

kmurray:
Actually, would it be possible to make it a separate test (i.e. a `strong_place_delay_calc_method`) rather than combining it with `strong_place_delay_model`?

Unless there is a good reason (like internal coupling/interactions between options), its generally easier to debug if the tests check features orthogonally. Since the `--place_delta_delay_matrix_calculation_method` and `--place_delay_model` are orthogonal in the code keep the tests orthogonal seems like a good idea.

litghost:
> Actually, would it be possible to make it a separate test (i.e. a `strong_place_delay_calc_method`) rather than combining it with `strong_place_delay_model`?

Sure

PR #1010, opened 2019-10-18
Avoid global routing messing up FASM generation

leon575777642:
Added a null-pointer test in the main loop of the `FasmWriterVisitor::walk_routing()` method, so an empty trace won't cause segment fault during FASM generation.

#### Description
When calling the `genfasm` util with routing results containing global routes (for example a clock), an empty trace will appear in the `g_vpr_ctx.mutable_routing().trace` vector. The head of such trace is a null pointer, and the `FasmWriterVisitor::walk_routing()` method does not check this before walking the route tree converted from it, eventually causing a segment fault.

#### Types of changes

#### Checklist:


kmurray:
@leon575777642 Thanks for the fix!

Issue #506, opened 2019-03-14
Eliminate vtr_free and vtr_malloc/vtr_calloc/vtr_realloc


vaughnbetz:
We have a mix of new/delete and malloc/free in the vpr code base.  This is dangerous as someone might free something created with new, or delete something created with malloc.

#### Proposed Behaviour & Solution
Remove all free/malloc/calloc/realloc calls and replace with new/delete equivalents.

Take some care when dealing with the vtr_chunk_alloc routines, as we still want those efficient memory pooling allocators (if they call new/delete when they actually get the memory they're managing in chunks, that would be sufficient for this clean up).

Holding this for now and will assign to a summer student.

litghost:
I wonder if all instances of raw malloc/free/new/delete should be using a managed container?  There may be a handful of places where the cost std::vector is expensive, but that should be replaced by vtr::light_vector or something.

kmurray:
I agree using a container to automatically manage memory makes sense.

vaughnbetz:
We did most of this.  VPR no longer has vtr_malloc or vtr::malloc. Odin II, parmys, libarchfpga and vqm2blif do have vtr::malloc.  
It might be worth eliminating vtr::malloc from libarchfpga. @AlexandreSinger @AmirhosseinPoolad : any thoughts?  It could be a useful summer student task.

AlexandreSinger:
I agree with this completely. I am on the side of Kevin on this. Using a container to manage memory is safer and easier to work with. Ideally we should create something similar to LLVM's SmallVector / DenseMap / etc. which have the same interface as std::vector / std::map / etc. but with added memory optimizations for the size of elements expected for compilers.

I agree it would be a useful summer task to have them change all vtr::mallocs into `new`, and then if they are keen on it potentially abstracting some of them into containers. We can explore creating our own custom containers down the road. I am not sure how much gain that can get, but they were certainly worth it in LLVM!

AmirhosseinPoolad:
I also agree with this. Changing all malloc/frees to new/delete is definitely doable without spending a lot of time, but the end goal should be using containers everywhere or in case of some super optimized code encapsulating all manual memory management in constructor/destructors (Basically RAII). I'll try to make a list of all places we do malloc/free and also all of the manual memory management.

vaughnbetz:
Let's use this as a first task for a summer student.

vaughnbetz:
[...]

PR #1011, opened 2019-10-18
Release ref count on timing information when freeing draw state.

litghost:


kmurray:
@litghost Could you add a comment describing why this is needed?

I think its because the draw state will otherwise hold onto the timing analyzer forever (shared_ptr semantics). But this shouldn't cause any real memory leaks in the standard tool flow, since whenever the graphics were passed a new timing analyzer its assigned to draw_sate->setup_timing_info, which would decrement the ref count on the old (previously held analyzer, and increment the new ref count). Correct?

litghost:
> @litghost Could you add a comment describing why this is needed?

Sure

> I think its because the draw state will otherwise hold onto the timing analyzer forever (shared_ptr semantics). But this shouldn't cause any real memory leaks in the standard tool flow, since whenever the graphics were passed a new timing analyzer its assigned to draw_sate->setup_timing_info, which would decrement the ref count on the old (previously held analyzer, and increment the new ref count). Correct?

Your analysis doesn't account for what happens VPR is terminating.  `draw_sate->setup_timing_info` is never cleared, then it's refcount is always 1 at termination.  As a result it looks like a leak of the setup_timing_info at termination.

litghost:
It is interesting though, the shared_ptr should be cleared after main exits, but asan/leak sanitizer did find a leak related to this shared_ptr.  Adding this reset resolved the leak sanitizer report, but I wonder if it might be an error in leak sanitizer?

kmurray:
> It is interesting though, the shared_ptr should be cleared after main exits, but asan/leak sanitizer did find a leak related to this shared_ptr.

Yes, I'd expect it to be cleaned-up then as well. 

Just speculating, but I've seen cases in the past in VPR, where constructors/destructors weren't being called correctly since the containing struct was allocated with malloc instead of new (motivating #506). Perhaps that could be the case with draw_state?

litghost:
> Just speculating

Good idea, but nope: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/draw/draw_global.cpp#L22 . I'll need to dig into this later.

PR #1012, opened 2019-10-21
Free remained portname's memory

sdamghan:
tested by make test Odin

#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/SRC/read_blif.cpp
+++ b/ODIN_II/SRC/read_blif.cpp
@@ -1542,7 +1542,7 @@ hard_block_ports *get_hard_block_ports(char **pins, int count)
 	ports->count = 0;
 	ports->sizes = NULL;
 	ports->names = NULL;
-	char *prev_portname = 0;
+	char *prev_portname = NULL;
 	int i;
 	for (i = 0; i < count; i++)
 	{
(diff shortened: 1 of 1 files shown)


ODIN_II/SRC/read_blif.cpp
```diff
@@ -1556,10 +1556,13 @@ hard_block_ports *get_hard_block_ports(char **pins, int count)
 			ports->sizes[ports->count] = 0;
 			ports->names[ports->count] = portname;
 			ports->count++;
+			prev_portname = portname;
+
```
jeanlego: This changes the behavior. Only gets assigned on I == 0 or when previous is equal to current

sdamghan:
I changed the code based on your opinion, and it's working correctly. I also checked the process of this part of the code step by step, and I've found that we don't need actually to set the value of the prev_portname in the all status except I == 0 and previous equal to current. BTW, both work correctly.

ODIN_II/SRC/read_blif.cpp
jeanlego: Can you initialize prev_portname at the beginning of the function to NULL rather than 0

PR #1015, opened 2019-10-23
Inline and make const ref some args in the inner loop of the router.

litghost:
#### Description

- Avoid copying cost_parameters structure.
- Request inline of inner loop functions as they are invoked at few (1-3)
  callsites, but get called many many times.
- Don't get congestion if it won't be used.  Useful for accelerating
  pure expansion.

#### Related Issue

#1008 

#### Motivation and Context

When integrating the router for expansions in the placer and lookahead, I noted that a non-trival fraction of profiler CPU samples were being spent on function preambles.  This PR avoids this time via inlining or making const ref some arguments.  The cost is an increased binary size, however the number of callsites for the inlined functions are low (1-3), and the number of function calls is very very high, so this should provide a benefit to both pure expansions and timing driven expansions.

1. [x] Tested expansions with and without this change
2. [ ] Titan QoR run

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -168,13 +168,13 @@ static std::vector<t_heap> timing_driven_find_all_shortest_paths_from_heap(const
                                                                            std::vector<int>& modified_rr_node_inf,
                                                                            RouterStats& router_stats);
 
-static void timing_driven_expand_cheapest(t_heap* cheapest,
-                                          int target_node,
-                                          const t_conn_cost_params cost_params,
-                                          t_bb bounding_box,
-                                          const RouterLookahead& router_lookahead,
-                                          std::vector<int>& modified_rr_node_inf,
-                                          RouterStats& router_stats);
+static inline void timing_driven_expand_cheapest(t_heap* cheapest,
+                                                 int target_node,
+                                                 const t_conn_cost_params cost_params,
+                                                 t_bb bounding_box,
+                                                 const RouterLookahead& router_lookahead,
+                                                 std::vector<int>& modified_rr_node_inf,
+                                                 RouterStats& router_stats);
 
 static t_rt_node* setup_routing_resources(int itry, ClusterNetId net_id, unsigned num_sinks, float pres_fac, int min_incremental_reroute_fanout, CBRR& incremental_rerouting_res, t_rt_node** rt_node_of_sink);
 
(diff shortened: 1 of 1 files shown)


kmurray:
> When integrating the router for expansions in the placer and lookahead, I noted that a non-trival fraction of profiler CPU samples were being spent on function preambles.

That is somewhat unexpected, I'd expect those functions to be inlined by the compiler. What compiler version are you using?

Does this happen even if VPR is built with profile guided optimization: `make BUILD_TYPE=release_pgo`?


>     * void copying cost_parameters structure.

Agreed, this one makes sense.

>     * Request inline of inner loop functions as they are invoked at few (1-3)
>       callsites, but get called many many times.

My understanding was that putting 'inline' in the delcaration doesn't have any effect on modern compilers (since they consider functions for inlining by default, its not a hint). Do we have any measurements to show it makes a difference? I'd prefer to minimize the code churn otherwise.



>     * Don't get congestion if it won't be used.  Useful for accelerating
>       pure expansion.

Can you clarify the motivation here? I expect that all nets (even the critical path since we set max_criticality < 1.0) would need this information, and my preference would be to keep the control flow as straight forward as possible.

litghost:
> Can you clarify the motivation here? I expect that all nets (even the critical path since we set max_criticality < 1.0) would need this information, and my preference would be to keep the control flow as straight forward as possible.

When doing straight expansions (criticality = 1), not getting congestion has a significant effect.  #1008  and the attempts to have the map lookahead use the main router are the places where this is a win.  I'd expect the branch predictor in the router case to have 100% success, so it should be okay.

litghost:
> My understanding was that putting 'inline' in the delcaration doesn't have any effect on modern compilers (since they consider functions for inlining by default, its not a hint). Do we have any measurements to show it makes a difference? I'd prefer to minimize the code churn otherwise.

It modified the inlining behavior on clang v7 at `-O3`, so it did something.  It is true that the `inline` attribute is not demand, just a request, but in this case clang v7 was not doing the inline

kmurray:
> When doing straight expansions (criticality = 1), not getting congestion has a significant effect. #1008

We should put a comment describing this motivation. Otherwise someone might remove it assuming it was a harmless change in the future :/

litghost:
> > When doing straight expansions (criticality = 1), not getting congestion has a significant effect. #1008
> 
> We should put a comment describing this motivation. Otherwise someone might remove it assuming it was a harmless change in the future :/

Done

litghost:
Superceeded

PR #1025, opened 2019-10-30
Add some bounding box methods to vtr::Rect with unit tests

HackerFoo:
Add some bounding box methods to vtr::Rect with unit tests.

#### Description
Adds a couple of useful constructors, `empty()` and `bounding_box` and `expand_bounding_box` to compute bounding rectangles.

#### Motivation and Context
I added a bounding box class while working on the connection box based lookahead, and vtr::Rect had most of what I implemented in the new class, so this allows me to use vtr::Rect instead.

I've added unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libvtrutil/src/vtr_geometry.h
+++ b/libs/libvtrutil/src/vtr_geometry.h
@@ -75,9 +75,10 @@ class Point {
 template<class T>
 class Rect {
   public: //Constructors
+    Rect();
     Rect(T left_val, T bottom_val, T right_val, T top_val);
     Rect(Point<T> bottom_left_val, Point<T> top_right_val);
-    Rect();
+    Rect(Point<T> bottom_left_val, T size);
 
   public: //Accessors
     //Co-ordinates
@@ -101,8 +102,16 @@ class Rect {
     //Returns true if the point is coincident with the rectangle (including the top-right edges)
     bool coincident(Point<T> point) const;
 
+    //Returns true if no points are contained in the rectangle
+    bool empty() const;
+
     friend bool operator== <>(const Rect<T>& lhs, const Rect<T>& rhs);
     friend bool operator!= <>(const Rect<T>& lhs, const Rect<T>& rhs);
+
+    template<class U>
+    friend Rect<U> operator|(const Rect<U>& lhs, const Rect<U>& rhs);
+    template<class U>
+    friend Rect<U>& operator|=(Rect<U>& lhs, const Rect<U>& rhs);
 
   public: //Mutators
     //Co-ordinates
--- a/libs/libvtrutil/src/vtr_geometry.h
+++ b/libs/libvtrutil/src/vtr_geometry.h
@@ -5,6 +5,7 @@
 #include <vector>
 #include <tuple>
 #include <limits>
+#include <type_traits>
 
 namespace vtr {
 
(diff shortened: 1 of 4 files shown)


libs/libvtrutil/src/vtr_geometry.h
```diff
@@ -101,9 +102,17 @@ class Rect {
...
+    //Returns true if no points are contained in the rectangle
+    bool empty() const;
+
     friend bool operator== <>(const Rect<T>& lhs, const Rect<T>& rhs);
     friend bool operator!= <>(const Rect<T>& lhs, const Rect<T>& rhs);
 
+    template<class U>
+    friend Rect<U> operator|(const Rect<U>& lhs, const Rect<U>& rhs);
+    template<class U>
+    friend Rect<U>& operator|=(Rect<U>& lhs, const Rect<U>& rhs);
```
kmurray: Can you clarify what these operators are supposed to do? It looks like it returns the minimal enclosing rectangle of lhs and rhs. To me this isn't an immediately obvious use of the 'or' operator (one might expect it to return a RectUnion instead).

For clarity, I'd tend to shy away from defining overloaded operators unless its very obvious what the should do.

Perhaps this should just be a free function like:
```cpp
Rect<U> bounding_box(const Rect<U>& a, const Rect<U>& b);
```
which would seem to capture the intent more explicitly.

libs/libvtrutil/src/vtr_geometry.h
```diff
@@ -101,9 +102,17 @@ class Rect {
     //Returns true if the point is coincident with the rectangle (including the top-right edges)
     bool coincident(Point<T> point) const;
 
+    //Returns true if no points are contained in the rectangle
+    bool empty() const;
```
kmurray: Likely want to clarify this comment. Its a little unclear what 'no points are contained' means. I think it means the Rect has zero area (i.e. width or height = 0).

libs/libvtrutil/src/vtr_geometry.h
```diff
@@ -75,9 +75,10 @@ class Point {
 template<class T>
 class Rect {
   public: //Constructors
+    Rect();
     Rect(T left_val, T bottom_val, T right_val, T top_val);
     Rect(Point<T> bottom_left_val, Point<T> top_right_val);
-    Rect();
+    Rect(Point<T> bottom_left_val, T size);
```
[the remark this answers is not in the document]
HackerFoo: This is useful to add points on an integer grid (size = 1), so that `Rect(p, 1).contains(p) == true`, so `size` is the size of the point.

I'm okay with making it width/height, defaulting the size to 1, or doing all of these with overloads.

Just using the smallest size that satisfies the above property (1 for integers, and epsilon for floating point) is probably the most reasonable, and stating the property in comments and adding tests.

libs/libvtrutil/src/vtr_geometry.h
HackerFoo: Agreed. The operators look nice, but `bounding_box` and `expand_bounding_box` are more readable.

libs/libvtrutil/src/vtr_geometry.h
HackerFoo: It means `rect.empty() => forall p. rect.contains(p) == false`. I'll add this property to the comments, but it probably wouldn't make a good unit test.

HackerFoo: Or closer to the comment:  `rect.empty() => not exists p. rect.contains(p)`

libs/libvtrutil/src/vtr_geometry.h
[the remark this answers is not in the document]
HackerFoo: Using epsilon for floats won't always work, so I think this constructor should just be enabled for integers.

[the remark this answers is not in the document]
HackerFoo: I've created a constructor from a single vtr::Point<T> where T is an integral type.

libs/libvtrutil/src/vtr_geometry.h
HackerFoo: Done.

libs/libvtrutil/src/vtr_geometry.h
HackerFoo: Done.

mithro:
@kmurray or @vaughnbetz could you please take another look at this pull request? It looks good to merge?

kmurray:
Thanks!

PR #1026, opened 2019-10-30
Make it easier to use GDB

HackerFoo:
Turn off optimization and add debug flags for the "debug" build type.

This avoids the dreaded `<value optimized out>` in GDB.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,16 @@ else()
 endif()
 
 #
+# Build type flags
+#
+
+if(NOT MSVC)
+    # for GCC and Clang
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0 -g3")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -g3")
+endif()
+
+#
 # Warning flags
 #
 set(WARN_FLAGS "") #The actual warning flags to be applied


CMakeLists.txt
```diff
@@ -110,6 +110,13 @@ else()
     message(STATUS "Building with IPO: off")
 endif()
 
+#
+# Build type flags
+#
+
+set(CMAKE_C_FLAGS_DEBUG "-O0 -g -ggdb")
```
mithro: `-g3` gets you **a lot** better debugging information (at the cost of increased binary size) -- this allows debuggers to understand macros and similar.

I highly recommend adding `-fvar-tracking` and `-fvar-tracking-assignments` for better info even with optimization turned on.

Both of the above probably need  `-gdwarf-4` to really work properly.

Lastly, `-gsplit-dwarf` might be a good idea to speed up the link time on debug builds too and `-gz` would be good if the file sizes get very large.

mithro:
It would be really good to have `-gdwarf-4` and `-g3` enabled (and probably `-gsplit-dwarf` and `-gz` will be needed then).

mithro:
Two random interesting links that are semi-related to this topic;
 - https://sourceware.org/gdb/onlinedocs/gdb/dotdebug_005fgdb_005fscripts-section.html#dotdebug_005fgdb_005fscripts-section
 - https://rethinkdb.com/blog/make-debugging-easier-with-custom-pretty-printers/

vaughnbetz:
Agreed; if we're not using the best options for symbol tables we should definitely set them.  I agree with Kevin that doing so in a way that doesn't cause issues with non-gcc compilers is the way to go.

HackerFoo:
This PR now enables `-g3`, `-O0`, and `gdwarf` (which should choose the default version.) I've enabled these flags for only non-MSVC compilers, and appended the flags instead of replacing them.

vaughnbetz:
Thanks, sounds good to me.

MohamedElgammal (commented):
Thanks! I think that would be useful for debugging process

mithro:
Can you check we are getting at least DWARF version 4 so we get macro debugging?

HackerFoo:
According to [the GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html):

> the default version for most targets is 4. DWARF Version 5 is only experimental.

HackerFoo:
The debug executable is 109MB on my machine, which I think doesn't warrant `-gsplit-dwarf` or `-gz`.
Link time for `vpr` is 6 seconds.

Macro debugging seems to work. To test:
```
make BUILD_TYPE=debug
gdb build/vpr/vpr
(gdb) b vpr_print_title
(gdb) run
(gdb) macro expand VTR_LOG("VPR FPGA Placement and Routing.\n")
expands to: do { if (true) vtr::printf("VPR FPGA Placement and Routing.\n"); } while (false)
```

I removed the `-gdwarf` flag. It wasn't needed.

vaughnbetz:
Sounds good -- thanks!

HackerFoo:
@vaughnbetz I think one of the checks is stuck again. There are two "Travis CI - Pull Request" checks and the queued one links to the other. Could you please override it? Thanks.

mithro:
@HackerFoo - What happened to `-fvar-tracking` and `-fvar-tracking-assignments`? Did you determine they are on by default?

vaughnbetz:
Fat finger (close & reopen by accident).  Merging ...

HackerFoo:
@mithro They are [enabled by default](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html) when optimization is used with `-g`.

PR #1034, opened 2019-11-04
Convert the vtr_flow/tasks/regression_tests/README to Markdown.

mithro:
This file should probably be merged / replaced with something from the VtR docs, but for now, I converted it to Markdown.

See https://github.com/mithro/vtr-verilog-to-routing/tree/vtr-readme-md/vtr_flow/tasks/regression_tests for a preview of the output.


kmurray:
> This file should probably be merged / replaced with something from the VtR docs

Yep, most of this is already covered in http://docs.verilogtorouting.org/en/latest/dev/developing/#running-tests

Issue #1038, opened 2019-11-06
Assertion failure with PLLs potentially due to propigated clocks


litghost:
#### Expected Behaviour

No assertion is fired on PLLs.

#### Current Behaviour

Following assertion is firing, I've added more to the assertion message:

```
/usr/local/google/home/keithrothman/cat_x/vtr-verilog-to-routing/vpr/src/timing/timing_graph_builder.cpp:207 add_block_to_timing_graph: Assertion '!model_port->is_clock' failed (Primitive data output (CLKFBOUT) can not be a clock generator, attached to plle2_test.pll.CLKFBOUT.plle2_test.pll.CLKFBOUT[0]).
```

> "Primitive data output can not be a clock generator"
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blame/master/vpr/src/timing/timing_graph_builder.cpp#L200

Assertion was introduced in https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/53510d94baf7691e1f15b8e5a857edec140688ed#diff-fdc393e2374fe736fe08a9f1fc7fbdf9R194, however I believe the bug is related to https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/440e6b1baa04e889f11edd4259471a88291e1fa8#diff-cb6d86211fc138d5f2a87ba2ac745365 removing some clock sources from `find_netlist_clocks`.

#### Possible Solution

It seems like `find_netlist_clocks` clocks needs two outputs:
 - The list of `SOURCE` pins that are clocks
 - The map of `SOURCE` that are unconstrained and are implicitly defined as coming from a parent clock

#### Steps to Reproduce

1. See https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1041

#### Your Environment

* VTR revision used:
* Operating System and version:
* Compiler version:

litghost:
Sorry I think the bug was here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/5007035d3389c84b7506d050e7b2b01bd9017a26#diff-31ec6c191fb261c72ad6804a23bf1a26

litghost:
[...]

PR #1041, opened 2019-11-06
Add test that demostrates #1038

litghost:
#### Description

Demonstrates assertion failure from https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1038

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1038

#### Motivation and Context

Adds a failing test.

#### Types of changes

#### Checklist:


litghost:
@kmurray

litghost:
This PR (c382f3c8a5c54603c9a637ba5d255cde68f82d89) actually demonstrates two additional bugs:
 - If the size of a FPGA never allows it to satisfy the usage constraints, then it will enter an infinite loop
 - The current equivilant tiles implement is broken if the `pb_type`'s are defined in a different order than the `tile`'s

I believe the second bug should be fixed in #988

PR #1042, opened 2019-11-07
Add missing C_internal to is_valid_route_tree.

litghost:
#### Description

I believe when C_internal support was added, VPR was not tested at `-DVTR_ASSERT_LEVEL=3`, and as a result it was not noticed that `is_valid_route_tree` was failing.

#### Motivation and Context

Assertion violation on 7-series tests when running at `-DVTR_ASSERT_LEVEL=3`.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/route_tree_timing.cpp
+++ b/vpr/src/route/route_tree_timing.cpp
@@ -1370,6 +1370,8 @@ bool is_valid_route_tree(const t_rt_node* root) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


kmurray:
Thanks for the fix!

This seems to indicate a hole in our regression test coverage. We run vtr_reg_basic and vtr_reg_strong with VTR_ASSERT_LEVEL=3, so it would seem we don't have c_internal tests in those tests.

PR #1044, opened 2019-11-13
Enable Cinternal on buffer.

litghost:
#### Description

Even though buffer's are non-configurable, so `Cinternal` will always be
applied, but that is a modelling choice, not a bug.

#### Motivation and Context

While it is possible to push `Cinternal` into the preceding node, but it isn't a bug to have to have an ambiguous modelling choice.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/read_xml_arch_file.cpp
+++ b/libs/libarchfpga/src/read_xml_arch_file.cpp
@@ -3637,7 +3637,7 @@ static void ProcessSwitches(pugi::xml_node Parent,
 
         } else if (0 == strcmp(type_name, "buffer")) {
             type = SwitchType::BUFFER;
-            expect_only_attributes(Node, {"type", "name", "R", "Cin", "Cout", "Tdel", "buf_size", "power_buf_size"}, " with type '"s + type_name + "'"s, loc_data);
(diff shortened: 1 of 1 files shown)


kmurray:
My understanding of C_internal was to capture the internal capacitance which is exposed when a switch is turned on.

I don't follow how it would be possible for a non-configurable buffer to have an internal capacitance, since the buffer's input diffusion capacitance should be constant.

litghost:
> I don't follow how it would be possible for a non-configurable buffer to have an internal capacitance, since the buffer's input diffusion capacitance should be constant.

Think of it like a modelling artifact.  A switch can be configurable or not (which is a function of the bitstream interface), but the underlying switch has either a internal capacitance or capacitance depending on whether the switch is configurable or not.  The rr graph builder assigns the capacitance of switches onto upstream nodes during graph construction, but the 7-series graph doesn't use the graph builder.  It supplies its own graph.  In this case, a non-configurable switch with internal capacitance is the same as lumping the capacitance during graph construction.  They are equivalent.  It is easier for symbiflow to not worry about this artifact, and treat switch capacitance as always internal, and for buffers that are always on, that internal capacitance is always added.

PR #990, opened 2019-10-02
initial step to add RRGraph object for refactoring rr_graph

tangxifan:
Initial Effort on refactoring the Routing Resource Graph (RRG) to a unified RRGraph object,

#### Description
This commit focuses on adding
1. the class definition and member functions for the new object RRGraph;
2. a basic checking function check_rr_graph_obj(), which guarantee that the RRGraph meets the demand of VPR routers;
2. light modification on vtr_geometry.h|tpp to support for manipulations.

#### Related Issue
This is a starting point for VPR code base refactoring. 
We try to create standard data structures for modeling FPGA devices.

#### Motivation and Context
Motivation
-----------
With an aim to group the discrete routing-related data structures into a unified object, which can ease router development and improve runtime and memory usage.
RRGraph object is designed to in a general way how routing resources are connected in a FPGA fabric, replacing the classical rr_node and rr_edge data structures.

Vision
------
The RRGraph is designed to be a read-only database/graph, once created.
To be generic and easy to maintain, any builder for RRGraph should be a free function outside the object, which uses the public accessors and mutators.
RRGraph is supposed to a constant database for developers/users to access information about the FPGA routing architecture.
Placement and routing should NOT change any attributes of RRGraph.
Any placement and routing results should be stored in other data structures, such as PlaceContext and RoutingContext.

Object Organization
--------------------
A Routing Resource Graph (RRGraph or RRG) is a directed graph (has many cycles), which consists of a number of nodes and edges.
Each node represents a routing resource, which could be 
1. a routing track in X-direction or Y-direction (CHANX or CHANY) 
2. an input or an output of a logic block (IPIN or OPIN) 
3. a virtual source or sink node (SOURCE or SINK), which are starting/ending points of routing trees.

Each edge represents a switch between routing resources, which could be
1. a multiplexer
2. a tri-state buffer
3. a pass gate
4. a non-configurable (can not be turned off) buffer
5. a short (metal connection that can not be turned off

RRGraph is designed to a self-contained data structure as much as possible.
It contains the segment- and switch-related information which are used to describe node-level and edge-level attributes.

The rr-graph is the single largest data structure in VPR, so avoid adding unnecessary information per node or per edge to it, as it will impact memory footprint.
Instead, using indices to point to the outside data source instead of embedding to RRGraph

To avoid the misuse of data indices when accessing node/edge/switch/segment, we introduce specialized data types RRNodeId, RREdgeId, RRSwitchId, RRSegmentId.

[...]

PR #1048, opened 2019-11-15
Converter from rr_nodes to RRGraph object

tangxifan:
Continued Effort on refactoring the Routing Resource Graph (RRG) to a unified RRGraph object.
This change brings the RRGraph to the Device Context of VPR engine, and it will create an RRGraph object by porting the rr_nodes information (classic routing resource graph) to the refactored object.
This change has not impact any routers, analyzers and drawers which are major clients of routing resource graph. 
But later, changes will be made to these downstream functions.
Detailed design document: 
https://docs.google.com/document/d/1LMIlpYoppFtuSi_OZZJNLZ7edWFuuC0js4hm1Ah0n2M/edit?usp=sharing

#### Description

1. An RRGraph object has been added to DeviceContext data structure, in parrallel to the classic rr_nodes. This object when replace the legacy data structures, i.e., rr_nodes, rr_node_indices, rr_switches, rr_segments etc. when refactoring is done.

2. A loading function `create_rr_graph()` is introduced, which loads the rr_node information to RRGraph object.  This function duplicates the rr_node information. 

3. Multiple bug fixing to checking codes for RRGraph. 

4. Added default constructors in `vtr_geometry.h` to resolve compatibility issues in Clang compilers.

5. Changes in golden_results for vtr_strong regression tests, which are mainly for relaxing the memory usage.  This is actually temporary! Since we duplicate the routing resource graph, it is expected that the memory usage will explode. 
On my local machine test, four tests in vtr_strong fails:
[a] regression_tests/vtr_reg_strong/strong_titan
 stratixiv_arch.timing.xml/ucsb_152_tap_fir_stratixiv_arch_timing.blif/common max_vpr_mem: previous_golden = 1011288 current_golden = 1368492
[b] regression_tests/vtr_reg_strong/strong_custom_grid
 non_column.xml/raygentop.v/common max_vpr_mem: previous_golden = 91316 current_golden = 193908
 [c] regression_tests/vtr_reg_strong/strong_custom_grid
non_column_tall_aspect_ratio.xml/raygentop.v/common max_vpr_mem: previous_golden = 87860 current_golden = 191944
 [d] regression_tests/vtr_reg_strong/strong_custom_grid
multiple_io_types.xml/raygentop.v/common max_vpr_mem: previous_golden = 417392 current_golden  = 1258700

To ease the integration, I have done a run to rewrite the golden results. Feel free to bug this if you want only a few to be relaxed. 
For sure, the golden results will be improved when refactoring is done. The duplication on routing resource graph will not exist.

#### Related Issue

Continued Effort on refactoring the Routing Resource Graph (RRG) to a unified RRGraph object.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/990

#### Motivation and Context

With an aim to group the discrete routing-related data structures into a unified object, which can ease router development and improve runtime and memory usage.
RRGraph object is designed to in a general way how routing resources are connected in a FPGA fabric, replacing the classical rr_node and rr_edge data structures.

Passed basic and strong regression tests with relaxation on memory usage. (See details in Description section 5)
 

CentOS, g++-8.2
TravisCI 

#### Types of changes

#### Checklist:


tangxifan:
> Why do the golden files change?

This is mainly temporary during our code construction process. To enforce incremental pull requests, we will first adapt downstream routers to use RRGraph objects and then refactor the builders for RRGraph (You can see more details in our design document https://docs.google.com/document/d/1LMIlpYoppFtuSi_OZZJNLZ7edWFuuC0js4hm1Ah0n2M/edit?usp=sharing). 

Therefore, at this stage, I will create a `RRGraph` object, which duplicates the routing resource graph modeled by legacy data structure `rr_node`. This will cause memory overhead considered by current golden results. However, this duplication will be eliminated when refactoring is done. 
The `RRGraph` object is a unified data structure to contain `rr_node`, `rr_node_indices`, `rr_segment` as well as `rr_switches`, without adding more information other than those already in these legacy data structures. So, there would not be memory overhead in the refactored version.

Actually, I put this issue open for discussion. Feel free to shot how you want to keep the current golden results.

kmurray:
I quickly ran this PR on the Titan23 design `neuron` and got the following:
```
## Build routing resource graph took 32.35 seconds (max_rss 2405.3 MiB, delta_rss +612.6 MiB)
  RR Graph Nodes: 5149691
  RR Graph Edges: 40724744
## Build routing resource graph object
## Build routing resource graph object took 30.41 seconds (max_rss 5911.4 MiB, delta_rss +3506.2 MiB)
```
which implies that the new object is using substantially more memory (5.7x) than the previous data structure. We probably need to look into why this is. I would expect the new RR graph object to use somewhat more (maybe 20-30%) but not this much more. My best guess is that there is some memory fragmentation occurring which is the problem. (I've previously seen these types of effects with the old RR graph as well).

Looking at the creation code, it seems like you aren't doing any reservations for the per-node edge lists. While you do reserve the edge elements themselves, we don't reserve the RRGraph::node_in_edges_ or RRGraph::node_out_edges_ sub vectors. That is probably the first thing I would try.

After that I've had good success using the Valgrind's 'Massif' heap profiler to figure out memory usage hot spots.

tangxifan:
> I quickly ran this PR on the Titan23 design `neuron` (which is a moderate sized design; the vtr_reg_basic vtr_reg_strong tests only include small designs so that they run fast) and got the following:
> 
> ```
> ## Build routing resource graph took 32.35 seconds (max_rss 2405.3 MiB, delta_rss +612.6 MiB)
>   RR Graph Nodes: 5149691
>   RR Graph Edges: 40724744
> ## Build routing resource graph object
> ## Build routing resource graph object took 30.41 seconds (max_rss 5911.4 MiB, delta_rss +3506.2 MiB)
> ```
> 
> which implies that the new object is using substantially more memory (5.7x) than the previous data structure. We probably need to look into why this is. I would expect the new RR graph object to use somewhat more (maybe 20-30%) but not this much more. My best guess is that there is some memory fragmentation occurring which is the problem. (I've previously seen these types of effects with the old RR graph as well).
> 
> Looking at the creation code, it seems like you aren't doing any reservations for the per-node edge lists. While you do reserve the edge elements themselves, we don't reserve the RRGraph::node_in_edges_ or RRGraph::node_out_edges_ sub vectors. That is probably the first thing I would try, as they will otherwise be constantly reallocated whenever we do RRGraph::add_edge().
> 
> After that I've had good success using the Valgrind's 'Massif' heap profiler to figure out memory usage hot spots.

Thanks for the insights. Very helpful! I will follow this idea and optimize the `RRGraph` object. 
So I will revert back to the old golden results and see if I can fit again.
By the way, I am thinking about other sources of the problem. 
* Does the current statistics include memory usage for `rr_node_indices`? Current `RRGraph` contains it besides `rr_nodes`.
* We have two big mappings `old_to_new_rr_node` and `old_to_new_rr_edge`. Will these create memory overhead?

kmurray:
>     * Does the current statistics include memory usage for `rr_node_indices`? Current `RRGraph` contains it besides `rr_nodes`.

Yes, I believe it does. Since rr_node_indices is created at the same time as the RR graph (in the traditional RR graph builder) it should be included in the +612MiB

>     * We have two big mappings `old_to_new_rr_node` and `old_to_new_rr_edge`. Will these create memory overhead?

Yes, they will have some overhead. I'd focus on the edge lists first since I expect they will be the bigger issue.

You currently are using:
```cpp
     std::map<int, RRNodeId> old_to_new_rr_node; 
```
Since both the key and value are small (2x32-bits) the overhead for a std::map BST node will be large (at least 2x64 bit pointers per key-value pair). Since we know the old RR node indices are contiguous integers you could achieve the same look-up effect with:
```cpp
std::vector<RRNodeId> old_to_new_rr_node(rr_nodes.size(), -1);
```
which would be much lower overhead (no pointer overhead per node). Note that we sized it exactly to avoid dynamically growing its size which can cause memory fragmentation.

The edge look-up is more challenging since the key is not a contiguous ID range:
```cpp
     std::map<std::pair<int, int>, RREdgeId> old_to_new_rr_edge; // Key: 
```
Since we know precisely how many edges there are in the graph, you could do something like:
```cpp
std::vector<std::pair<std::pair<int,short>,RREdgeId> old_to_new_rr_edges_vec(num_edges);

for (int inode = 0; inode < rr_nodes.size(); ++i) {
    for (short iedge : rr_nodes[inode].edges()) {
        RREdgeId edge_id = rr_graph.add_node();
        old_to_new_rr_edges_vec.emplace_back({{inode,iedge},edge_id});
    }
}
vtr::flat_map<std::pair<int,short>,RREdgeId> old_to_new_rr_edges(std::move(old_to_new_rr_edges_vec));
```
Which uses a vtr::flat_map (a map implemented as a sorted vector, rather than a BST), constructed directly from an exactly sized vector. That would again avoid the 2x64bit pointer overhead for each BST node used by std::map.

kmurray:
Re-running on `neuron` with the latest changes does show improvement:
```
## Build routing resource graph took 29.99 seconds (max_rss 2505.3 MiB, delta_rss +656.2 MiB)
  RR Graph Nodes: 5149691
  RR Graph Edges: 40724744
## Build routing resource graph object
## Build routing resource graph object took 3.80 seconds (max_rss 3787.9 MiB, delta_rss +1282.6 MiB)
```
The object data structure only uses 2x the baseline RR graph (down from 5.7x).

LNIS-Projects:
> Re-running on `neuron` with the latest changes does show improvement:
> 
> ```
> ## Build routing resource graph took 29.99 seconds (max_rss 2505.3 MiB, delta_rss +656.2 MiB)
>   RR Graph Nodes: 5149691
>   RR Graph Edges: 40724744
> ## Build routing resource graph object
> ## Build routing resource graph object took 3.80 seconds (max_rss 3787.9 MiB, delta_rss +1282.6 MiB)
> ```
> 
> The object data structure only uses 2x the baseline RR graph (down from 5.7x).
Indeed, I found that the `node_in_edge` and `node_out_edge` reservation do help in reducing memory fragments. 
After applying this, the number of failed test cases in regression tests drops from 4 to 2.
Moreover, after I removed the big edge mapping `old_to_new_rr_edge` in `convert_rr_graph`, the memory footprint drops significantly. 
After applying this, the number of failed test cases in regression tests further drops from 2 to 1.
The memory footprint of the test case `multiple_io_types.xml/raygentop.v/common` drops from 1GB to 600MB, while the golden result is 400MB.
Now I keep looking into the memory hotspots in `convert_rr_graph`.
I have also tried to reserve memory in the fast `node_lookup` but it appears to have no impact on the memory footprint.

However, we have to admit that `RRGraph` object indeed consumes more memory than the legacy `rr_node`. This is because:
* The input edges for each node are now stored in `RRGraph` object, while they are not in `rr_node`. Note that `rr_node` only keeps the number of input edges, which is `fan_in`. This is the major source of memory overhead. But in practice, `node_in_edges` is very important and useful when you traverse the graph and apply back-annotation. Memory investment is worthwhile.
*  The ids of `node` and `edge` cost additional memory.
*  Segment id for each `node` costs additional memory. I may try to remove it and see memory benefits.

Any thoughts are warmly welcomed.

kmurray:
> However, we have to admit that `RRGraph` object indeed consumes more memory than the legacy `rr_node`.

Yes this is true. 

Previously when I've looked at this I've found it helpful to work out on a spread sheet what the memory usage is. Its much easier to experiment with a spreadsheet than re-coding everything.Take a look at [this spreadsheet](https://docs.google.com/spreadsheets/d/11tA6je9clKlTFo3L5ikT8BgMgMgEwMkMZmHYH_xhZfY/edit?usp=sharing) which works out roughly where we spend memory.

The major culprits are things which are O(edges) as there are far more edges than anything else.

Optimizations which seem to be no-brainers:
* Use uint16_t for num_config in/out edges (size_t is hugely overkill, and the old RR graph also used shorts for this).
* We can re-define the NodeLookup from:
```cpp
    /* Fast look-up to search a node by its type, coordinator and ptc_num 
     * Indexing of fast look-up: [0..xmax][0..ymax][0..NUM_TYPES-1][0..ptc_max][0..NUM_SIDES-1] 
     */
    typedef std::vector<std::vector<std::vector<std::vector<std::vector<RRNodeId>>>>> NodeLookup;
    mutable NodeLookup node_lookup_;

```
to
```cpp
    /* Fast look-up to search a node by its type, coordinator and ptc_num 
     * Indexing of fast look-up: [0..xmax][0..ymax][0..NUM_TYPES-1][0..ptc_max][0..NUM_SIDES-1] 
     */
    typedef vtr::NdMatrix<std::vector<std::vector<RRNodeId>>,3> NodeLookup;
    mutable NodeLookup node_lookup_;
```
Since the width, height and number of RR types are fixed (i.e. known) at RR graph construction time. This would avoid the std::vector overhead (24 bytes for each instance). We'd want to keep the inner two dimensions as std::vector, since they are not fully dense.

Potential functionality preserving optimizations:
* Instead of storing the each valid Id (node_ids_, edge_ids_), simply store the delimeters of the valid range in a vector and do binary search to check whether a given Id is in the valid range or not. Since we usually only have valid IDs this should be effectively constant time, but will save ~175MiB (mostly from eliminating edge_ids).

* In various places where we're storing a small number of elements we could try using vtr::small_vector, which trades-off maximum vector size for more efficient storage (For small sizes it can also store elements in place within the object). For instance this could be used for the inner dimensions of the NodeLookup:
```
    /* Fast look-up to search a node by its type, coordinator and ptc_num 
     * Indexing of fast look-up: [0..xmax][0..ymax][0..NUM_TYPES-1][0..ptc_max][0..NUM_SIDES-1] 
     */
    typedef vtr::NdMatrix<vtr::small_vector<vtr::small_vector<RRNodeId,uint8_t>,uint16_t>,3> NodeLookup;
    mutable NodeLookup node_lookup_;
```
Since NUM_SIDES=4 we can get away with using 1 byte to store the size of the inner dimension, and since the number of PTCs is at most a few hundred we can use 2 bytes for the size.

Potential Optimizations with more trade-offs:
* Don't store backward edges (i.e. node_in_edges). We loose some functionality (but match the functionality of the old RR graph). Saves ~260MiB.

* Use a unified array per node to store all edges (in, out, config, non config) along with counts to calculate the start/end of each range of edges.

Small optimizations which probably don't offer a big pay off:
* Use a union for node side/direction
* Store only unique RC values

kmurray:
If I had to push for the simplest path forward it would be to just apply the obvious optimizations, followed by:
* Storing only the delimters of valid ID ranges

and if that is insufficient:
* Drop storing the node_in_edges

vaughnbetz:
Discussing this with Kevin ...
I think the edge_ids vector can be removed.
Current use: storing whether or not an id is valid. This can be more efficiently stored as:
unordered_map of the invalid edges (as usually this hash table will be empty).

Checking if an edge is valid: checks if the id is in range (0 < id < edge_ids.size) and probes the unordered_map to see if it is invalid.

Compressing rr-graph to remove invalid edges: You can walk through the hash table and put the invalid edge ids in a vector, sort it, use that to walk (once) through the edge_ids and reindex them as we can see immediately by how much to reduce the value of every edge id.

Should do the same thing for node_ids as it gives savings too.

Expected savings from the spreadsheet Kevin attached:  11.8% from edges + 1.5% from nodes = 13.3% memory footprint.  Should have no impact on runtime.

vaughnbetz:
From the spreadsheet, bidirectional edges are costing about 21.8% memory footprint.  Convenient though.  Suggest optimizing everything else and then we can think about that one, as deleting them is a real trade-off.

vaughnbetz:
I like the single node edge array + count delimiter option.   13% expected memory reduction, no impact on typical users of the rr-graph.  We lose constant time insertion of an arbitrary type of edge as we don't have unique vectors for each category, but only the rr-graph builder would care about that, and it could just set up the edge_* arrays, and then the end of the rr-graph builder could walk them to set up the node_edges* data.

tangxifan:
Thanks a lot for the advice. 
To follow up, I have modified the `RRGraph` object by 
* remove `node_ids_` and replace with a number and an unordered map for invalid ids 
* remove `edge_ids_` and replace with a number and an unordered map for invalid ids 
* use short type for `node_num_non_config_in/out_edges`
* use `NdMatrix` in fast node lookup

These optimizations indeed reduce the peak memory usage of the failing test case `multiple_io_types` from 600MB to 550MB (while the golden is 400MB).
In particular, I notice that `RRGraph` object does not contribute to major memory usage: 
## Build routing resource graph took 2.66 seconds (max_rss 211.9 MiB, delta_rss +171.1 MiB)
## Conversion to routing resource graph object
## Conversion to routing resource graph object took 0.69 seconds (max_rss 295.2 MiB, delta_rss +83.3 MiB)
  RR Graph Nodes: 478281
  RR Graph Edges: 3520986

That being said, I will probably try the following optimizations:
* use `vtr::small_vector` in `build_fast_node_lookup()`
*  do not fill vectors in `node_in_edges` when building `RRGraph`. I will create a method `build_node_in_edges()`, which can be called on-demand. By doing so, we may save some memory.

Then, we will see what we can do next.

kmurray:
@tangxifan Can you allow me [to push to this branch](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)? I have some refactorings which should help on this.

kmurray:
>     * do not fill vectors in `node_in_edges` when building `RRGraph`. I will create a method `build_node_in_edges()`, which can be called on-demand. By doing so, we may save some memory.

Based on discussions with @vaughnbetz it would be nice to maintain this by default if possible.

I'd suggest focusing on the single edge array per node (with counts of to delimit in/out/config/non-config nodes).

tangxifan:
> > ```
> > * do not fill vectors in `node_in_edges` when building `RRGraph`. I will create a method `build_node_in_edges()`, which can be called on-demand. By doing so, we may save some memory.
> > ```
> 
> Based on discussions with @vaughnbetz it would be nice to maintain this by default if possible.
> 
> I'd suggest focusing on the single edge array per node (with counts of to delimit in/out/config/non-config nodes).

Sure. I have added you to the contributor list of my branch. Feel free to push.
Thanks for the reminder on the `node_in_edges`. Let me do this.

tangxifan:
I have done another optimization by merging the two vectors`node_in_edges` and `node_out_edges` to one vector. And we use delimiters to identify different groups of edges.
The effect is not trivial! For `multiple_io_types` test case, the peak memory usage drops from 550MB to 520MB (while the golden is 400MB).
``` c++
## Build routing resource graph took 2.53 seconds (max_rss 211.8 MiB, delta_rss +171.0 MiB)
## Conversion to routing resource graph object
## Conversion to routing resource graph object took 0.66 seconds (max_rss 278.6 MiB, delta_rss +66.8 MiB)
  RR Graph Nodes: 478281
  RR Graph Edges: 3520986
# Create Device took 3.23 seconds (max_rss 278.6 MiB, delta_rss +237.8 MiB)
```

kmurray:
I've added the lazy iteration support, and also re-worked the merged in/out edges to use a single pointer instead of a vector (saves more memory).

Now when I run on `neuron` I get:
```
## Build routing resource graph took 26.85 seconds (max_rss 2542.8 MiB, delta_rss +916.0 MiB)
## Conversion to routing resource graph object
## Conversion to routing resource graph object took 5.49 seconds (max_rss 3448.6 MiB, delta_rss +905.8 MiB)
```
(using the glibc memory allocator), and
```
## Build routing resource graph took 29.01 seconds (max_rss 2345.3 MiB, delta_rss +877.4 MiB)
## Conversion to routing resource graph object
## Conversion to routing resource graph object took 5.90 seconds (max_rss 3213.6 MiB, delta_rss +868.3 MiB)
```
(using the TBB memory allocator).

Which indicates we use comparable (perhaps a little less) memory than the old RR graph!

kmurray:
@tangxifan It would be good to run a wider set of benchmarks to see if we've really got all the memory usage in check.

In particular, can you run:
1. The VTR benchmark nightly regression test (`vtr_flow/tasks/regression_tests/vtr_reg_nightly/vtr_reg_qor_chain`), and
2. The Titan benchmark weekly regression test (`vtr_flow/tasks/regression_tests/vtr_reg_weekly/vtr_reg_titan`)

And parse out the result for old and new RR graph memory usage (probably want a spreadsheet of memory ratios per benchmark)? That will tell us if there are any other poor behaviours we need to fix before merging this.

tangxifan:
> @tangxifan It would be good to run a wider set of benchmarks to see if we've really got all the memory usage in check.
> 
> In particular, can you run:
> 
>     1. The VTR benchmark nightly regression test (`vtr_flow/tasks/regression_tests/vtr_reg_nightly/vtr_reg_qor_chain`), and
> 
>     2. The Titan benchmark weekly regression test (`vtr_flow/tasks/regression_tests/vtr_reg_weekly/vtr_reg_titan`)
> 
> 
> And parse out the result for old and new RR graph memory usage (probably want a spreadsheet of memory ratios per benchmark)? That will tell us if there are any other poor behaviours we need to fix before merging this.

I have tried `vtr_flow/tasks/regression_tests/vtr_reg_nightly/vtr_reg_qor_chain`. 
All benchmarks passed functional test, among which 5 benchmarks failed in QoR.
I attached a spreadsheet to compare memory usage between `RRGraph` and `rr_nodes` 
https://docs.google.com/spreadsheets/d/1aDcFZ3eo5SwTaVwxuGgf_fo0QTU8XieI0Uw0W_bvTm8/edit?usp=sharing
In general, the new object is smaller or close to the legacy data structure.

I am trying titan now. But it seems that something is broken due to out-of-date architecture file `stratixiv_arch.timing.xml`
``` c++
Architecture file: stratixiv_arch.timing.xml
Circuit name: sparcT1_core_stratixiv_arch_timing

# Loading Architecture Description
# Loading Architecture Description took 1.39 seconds (max_rss 14.9 MiB, delta_rss +7.6 MiB)
Error 1: stratixiv_arch.timing.xml:1 Missing required child node 'tiles' in parent node 'architecture'
The entire flow of VPR took 2.44 seconds (max_rss 14.9 MiB)
```
Can you let me what changes are needed for the architecture file?
Then, I can continue this test.

vaughnbetz:
@acomodi is this possibly an architecture that needs to be updated to match the equivalent tiles code?

acomodi:
@vaughnbetz @tangxifan I believe the stratix IV arch was not yet updated after the physical/logical block split that happened ~1 month ago. In `vtr_flow/scripts` there is the upgrade_vtr_arch.py (or something similar) that updates an input architecture with the new specifications.

By running the script with the architecture that does not contain the `tiles` tag will solve the issue

acomodi:
Being Titan benchmarks (and also architecture) external to the repo they weren't modified accordingly to satisfy the new `tiles` tag addition

tangxifan:
> Being Titan benchmarks (and also architecture) external to the repo they weren't modified accordingly to satisfy the new `tiles` tag addition

Thanks! It works!

kmurray:
> > Being Titan benchmarks (and also architecture) external to the repo they weren't modified accordingly to satisfy the new `tiles` tag addition

FYI, I've just upgraded the download_titan.py script to automatically apply upgrade_arch.py to the architectures it creates. Should avoid that issue in the future.

tangxifan:
> @tangxifan It would be good to run a wider set of benchmarks to see if we've really got all the memory usage in check.
> 
> In particular, can you run:
> 
>     1. The VTR benchmark nightly regression test (`vtr_flow/tasks/regression_tests/vtr_reg_nightly/vtr_reg_qor_chain`), and
> 
>     2. The Titan benchmark weekly regression test (`vtr_flow/tasks/regression_tests/vtr_reg_weekly/vtr_reg_titan`)
> 
> 
> And parse out the result for old and new RR graph memory usage (probably want a spreadsheet of memory ratios per benchmark)? That will tell us if there are any other poor behaviours we need to fix before merging this.

I have finished running the two set of benchmarks. 
Detailed statistics can be found at: 
\url{https://docs.google.com/spreadsheets/d/1aDcFZ3eo5SwTaVwxuGgf_fo0QTU8XieI0Uw0W_bvTm8/edit?usp=sharing}

For VTR benchmarks, we have only a few cases that fail in QoR. This is mainly due to the RRGraph is quite small (the design is not huge) and memory overhead is not significant.
For Titan benchmarks, we have most of them failed in QoR. I think these cases are worth to study because their RRGraph is large (we use a fixed routing channel width). This can give us a direct comparison between `rr_node` and  `RRGraph`.

Meanwhile, I also notice that delta memory for `rr_node` is not accurate. A few Titan benchmarks have zero memory increase when creating `rr_node`. I believe this is mainly due to packing reach the peak memory usage and then a lot of memory get freed when packing is finished.
Creation of `rr_node` just reuse these freed memory space and thus do not increase peak memory usage. 
Do you have any idea about how to capture the memory footprint of `rr_node` more precisely?

vaughnbetz:
The inaccurate rr_node memory is a good point; the delta memory is a lower bound on the size of the rr_graph.
Possibly you could have a loop that allocates memory and constantly checks the resident set size until it increases.  You could allocate big chunks (1 kB) until memory increases, then free the last one, then allocate smaller chunks (32 bytes?) until it increases, then free the last one, etc.  That way you make sure all the holes in memory are filled before going on to the next phase of memory allocation, and you will capture the whole increase in memory resident size increase.  
If you do this, it is likely a useful utility function (ensure_all_memory_holes_filled) or some such, for cases like this where we'd like to precisely measure the size of some data structure.

vaughnbetz:
Are the titan failures due to the regtest flagging the resident memory set as being out of bounds?  Or did the compile fail (out of memory or anything else)?
Can you produce a spreadsheet that gives before and after peak memory footprint (with and without the new rr-graph object being created) as that would be another useful view of the memory footprint impact (it'll be higher than the real impact as both rr-graphs are loaded, but it would let us lock for unusually large increases).

kmurray:
@tangxifan I'd suggest temporarily inserting calls to `vtr::malloc_trim(0)` (a wrapper around GLIBC's malloc_trim) before building the old RR graph, and before creating the new RR graph. That should try to free any unused top-of heap memory which *should* make the delta rss values more accurate. See `man malloc_trim` for details.

tangxifan:
It seems that we have some errors in getting titan benchmarks.
```
Building target(s): get_titan_benchmarks
Downloading (~1GB) and extracting Titan benchmarks (~10GB) into VTR source tree.
  File "./vtr_flow/scripts/download_titan.py", line 86
    print "Found existing {} with matching checksum (skipping download and extraction)".format(tar_gz_filename)
                                                                                      ^
SyntaxError: invalid syntax
```

tangxifan:
> Are the titan failures due to the regtest flagging the resident memory set as being out of bounds? Or did the compile fail (out of memory or anything else)?
> Can you produce a spreadsheet that gives before and after peak memory footprint (with and without the new rr-graph object being created) as that would be another useful view of the memory footprint impact (it'll be higher than the real impact as both rr-graphs are loaded, but it would let us lock for unusually large increases).

All the benchmarks passed the `run_vtr_flow`, meaning that packing, placement and routing are successful. Errors occurred only when `parse_vtr_flow` checks the memory usage and complains they exceed the golden results too much.

I am reworking the spreadsheet to record golden memory results and current memory usage.
Actually, I have another idea to capture the memory size of `rr_node`.
I would like to call only VPR router for each benchmark. In this case, the packer and placer will only read-in previous results, which should not create too much holes in memory.
Then router will build `rr_graph` and we can find the actual memory size of `rr_node`.
What do you think?

kmurray:
> I am reworking the spreadsheet to record golden memory results and current memory usage.
> Actually, I have another idea to capture the memory size of `rr_node`.
> I would like to call only VPR router for each benchmark. In this case, the packer and placer will only read-in previous results, which should not create too much holes in memory.
> Then router will build `rr_graph` and we can find the actual memory size of `rr_node`.
> What do you think?

Yes, that seems reasonable.

Also, your changes in 11f719a put the `vtr::malloc_trim()` calls after the `vtr::ScopedStartFinishTimer`s are constructed. You'll actually want to do them the other way around (since the timers record the current memory usage when constructed in order to calculate the delta, we want to ensure memory is trimmed first).

tangxifan:
> > I am reworking the spreadsheet to record golden memory results and current memory usage.
> > Actually, I have another idea to capture the memory size of `rr_node`.
> > I would like to call only VPR router for each benchmark. In this case, the packer and placer will only read-in previous results, which should not create too much holes in memory.
> > Then router will build `rr_graph` and we can find the actual memory size of `rr_node`.
> > What do you think?
> 
> Yes, that seems reasonable.
> 
> Also, your changes in [11f719a](https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/11f719a4ec150182c6795541dde127585c2d67ba) put the `vtr::malloc_trim()` calls after the `vtr::ScopedStartFinishTimer`s are constructed. You'll actually want to do them the other way around (since the timers record the current memory usage when constructed in order to calculate the delta, we want to ensure memory is trimmed first).

Thanks for the advice. Just made the modification.
From what I observed so far, peak memory is reached most of time when `create_device()` function is executed. After looking into that, I realize that you create a rr_graph before placement and routing. But it seems that this rr_graph is a detailed rr_graph? Are we still using global rr_graphs for placer?

vaughnbetz:
We create a detailed routing (wires, switches) rr-graph for the placer, so we can profile / search the routing graph to produce delay lookups that the placer can quickly access.

tangxifan:
I have summarized in the [spreadsheet](https://docs.google.com/spreadsheets/d/1aDcFZ3eo5SwTaVwxuGgf_fo0QTU8XieI0Uw0W_bvTm8/edit?usp=sharing) about the peak memory usage of VPR flow and rr_graph.
It also compares the memory usage of golden results,  before refactoring (tested on our local server) and after refactoring  (tested on our local server).
I have tested over both VTR benchmarks and Titan benchmarks.
I noticed that on my local server, it seems that all the Titan benchmarks failed in QoR (too much overhead in memory usage）even before refactoring.
Can you check quickly on your local computers and tell me if it goes wrong at your side as well? 

In the spreadsheet, you may notice that even when I turn on router only, the memory statistics on `rr_node` is not accurate. You can see some benchmarks give zero usage which does not make any sense. But when you compare to the memory statistics of running vpr flow, it is more accurate.

vaughnbetz:
For the Titan benchmarks, what is the difference between the two router-only runs?  I see two columns, with and without refactoring, for the router only runs.

tangxifan:
> For the Titan benchmarks, what is the difference between the two router-only runs? I see two columns, with and without refactoring, for the router only runs.

Sorry. That is a typo. The left part is full VPR flow-run while the right part is router-only run.
I have corrected it in the spreadsheet.

tangxifan:
Hi all, I understand that the memory footprint has become a critical concern here. To address that, I suggest unit tests before deployment. My plan is as follows:
1. Move data structure `rr_node` and `RRGraph` out from `vpr` to a library `librrgraph` in `libs`. 
    And I remove the conversion function from VPR codes.
2. Build readers and writers for `RRGraph` as we did for `rr_node`. We make sure that the `RRGraph` can read the same XML and output the same XML as `rr_node` did.
3. Compare the memory footprint of both objects with a set of routing resource graphs (*.xml) 
4. Optimize the internal data organization of `RRGraph` if needed.

As such, we can be confident that before deployment
1. RRGraph object is functional 
2. RRGraph object will not cause overheads 

After this, we can continue our incremental refactoring.
Your advice is warmly welcomed.

tangxifan:
Hi all, 
   I have experimentally deployed the RRGraph object in VPR8 version (commit 2780988d58da6be454364e4a06cff8da8caf09e8).
   My implementation is now under testing in [the OpenFPGA framework](https://github.com/LNIS-Projects/OpenFPGA/commit/6dc2126ce611820960b55c7a73c0ae2fb4c61a36).
   My modification covers 
   1. routing resource graph builder
   2. Routers
   3. Routing results storage.
   4. Routing stats print-out
   5. Drawer
   Currently, I have not seen any memory overhead reported by vtr-basic and vtr-strong regression tests. It means that it really worths a try to land this refactoring effort.
    During this effort, I have seen that current rr_node_indicies have a lot of exceptions, which are not allowed by the new data structures. For example, the indexing on SOURCE/SINK/OPIN/IPIN of grid whose width and height is >1. These all bring a lot of difficulties, i.e., many hard-to-fix bugs, when adapting the rr_graph builders. 
    In terms of QoR, I have seen some QoR shift: 
    1. minimum routable channel width is reduced for some benchmarks. I am still investigating why.
    2. critical path delay is slightly reduced. I suspect there is something wrong in the rc-tree annotation.
    However, these problems should be addressed very soon.

PR #1052, opened 2019-11-21
Add base cost option that is bounded.

litghost:
#### Description

This adds a base cost option that is length scaling, but where the length scaling is bounded to the range 1 to 4, where 1 is the segments with the shortest length, and 4 is the longest length.  This was discovered to be helpful in the congestion routing of the 7-series graph because it lowered the backtracking effect of a lookahead mispredict, but still enabled the router to prefer shorter wires when performing decongestion.

In particular, the 7-series arch has a large min to max length range than that found in the Stratix 4 arch.  The Stratix 4 arch only has length 4 and length 16 segments, resulting in a 4-16 scaling range. To mirror that for the 7-series arch (which has length 1 up to length 44 routing segments), the base cost scaling was limited artifically to the range 1 to 4.

#### Related Issue

#1047 

So far this has only been tested on the 7-series arch.

#### Types of changes

#### Checklist:


kmurray:
The code for this looks fine. However I'm less sure about how necessary this approach is.

> In particular, the 7-series arch has a large min to max length range than that found in the Stratix 4 arch.

This is true.

However, one of the nice features of the length-scaled base costs is that it makes the basic cost of using resources of different lengths consistent. In particular, it makes the cost of using a wire per unit distance the same regardless of length.

For instance, with the (unbounded) length scaling:
 * L44:
   * cost: 44
   * cost / distance: 44 / 44 = 1
* L1:
   * cost: 1
   * cost / distance: 1 / 1 = 1

So for intermediate distances (e.g. 6) it is cheaper to use the short L1 wires (6\*1 = 6) than the L44 wires (44\*1 = 44) which is likely what is desired.

In contrast, with the proposed bounded method:
 * L44:
   * cost: 4
   * cost / distance: 4 / 44 = 0.09
* L1:
   * cost: 1
   * cost / distance: 1 / 1 = 1

So for intermediate distances (e.g. 6) it is _more expensive_ to use the short L1 wires (6\*1 = 6) than the L44 (1\*4 = 4).

This means that many connections will want to use the longer wires since they are cheapest per-unit distance travelled. Since longer wires are usually much less plentiful than short wires, this can lead them to become unnecessarily congested (e.g. with non-timing critical connections which would be fine on the shorter wires), significantly slowing down routing convergence.

kmurray:
As discussed, we'll merge this for now, but eventually this will be superseded by the customized base cost scaling alternative discussed in #1207.

Issue #1063, opened 2019-12-11
Allow for heterogenous placements


litghost:
#### Proposed Behaviour

Enable a tile definition that has the following properties:
 - Heterogeneous placement types
 - Non-uniform physical to logical pin mappings

#### Current Behaviour

Tile capacity causes a full duplication of the tile ports and each instance in the capacity stack must be identical.

#### Possible Solution

I believe this feature requires two steps:
1. Decouple how physical tile ports are defined from physical tile capacity.  Currently, capacity causes all ports to be duplicated N times.  If the tile has heterogenous logical tiles, this duplication is incorrect.   
2. Allow the logical tile mapping to be flexible.  This would modify:
   - Where valid placements exist?
   - Resource summaries

Because in step #1 the relationship between physical tile ports and logical tiles was already made flexible, there should not be a problem once step #2 is implemented at the placer.

#### Context

The 7-series fabric has several instance of heterogeneous tiles, where multiple logical tiles that should be placed independently are co-located at the same routing location.  While it is possible to split the tiles, it will further throw off the wirelength and lookahead calculations, and in the long term heterogeneous placement is likely desired.

In addition, there is some need to flexibility when it comes to homogeneous tiles.  The current "duplicate the port" strategy is simple, but unrealistic.  For example, in the 7-series fabric, the global clock tile has 16 instances.  To take a single port example, the clock enable (CE) port on instance 1-4 should have a height offset of 0, 5-8 should have a height offset of 1, etc.  The current strategy of duplicating ports does not accommodate this level of flexibility.

#### Note on simplicity and backwards compatibility

The plan is to add new flexible into the arch parser, rather than break backwards campatiplity.  My current thinking is to add a capacity_type attribute that defaults to the current duplicate strategy.  If the atttibute is missing, the capacity_type will be "duplicate" (e.g. duplicate the ports for each instance of capacity).  The new capacity_type of "explicit" will not duplicate the ports.  Instead, a custom pin mapping will be **required** to map the explicit ports to the logical instances.

litghost:
@acomodi @vaughnbetz @kmurray

kmurray:
Thanks for filing the issue @litghost.

I do not quite grasping what 'heterogeneous placement types' and 'Non-uniform physical to logical pin mappings' mean. Can you perhaps provide some examples and diagrams of what you want to model, and how these ideas would enable that?

[...]

PR #1065, opened 2019-12-12
Initial implementation of explicit ports.

litghost:
This implements part one of #1063

### DIFF
--- a/libs/libarchfpga/src/physical_types.cpp
+++ b/libs/libarchfpga/src/physical_types.cpp
@@ -110,7 +110,12 @@ std::vector<int> t_physical_tile_type::get_clock_pins_indices() const {
     int clock_pins_start_idx = 0;
     int clock_pins_stop_idx = 0;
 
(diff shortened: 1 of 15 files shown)


litghost:
@vaughnbetz @kmurray I know we talked about a different direction (e.g. stacks of physical tiles) but this change was significantly easier to implement in the short term, and unblocked progress on the Symbiflow front.

Take a look. Some of the changes here I think are better than the previous code (e.g. more robust / less assumptions), in particular I'm thinking about how the new code translates IPIN/OPIN rr nodes back to clusters.  I can isolate those changes if you want to accept them.

I also think adding heterogeneous logical tiles is fairly straight-forward from this PR, where-as stacks of physical tiles is going to be a lot more work (e.g. code changes), but I agree with the reasoning why to take one approach over another.   In the new year, I may finish the heterogeneous logical tiles before circling back and making the larger changes required to support stacks of physical tiles.

PR #1067, opened 2019-12-16
WIP: Generate profile of netlist characteristics from BLIF or Verilog Input in Odin-II

mflawn:
Squashed commit of project to generate netlist statistics including of
node types, levels, fanout and path length. Parser file to process the
output. Additional arguments added to Odin for Verbose and non-verbose
versions.

#### Description
This is the project aimed at gathering netlist characteristics in Odin like node type, counts, levels, fanout and path length.
The mechanism of collection for basic node information is just a recursive dfs traversal based on output_blif. I ended up separating verbose and non-verbose versions of the stat generator in part because some of the verbose data is irrelevant to basic profiling (# nodes, edges etc.) and because collecting path statistics is costly and possibly not as reliable and/or desirable. The non-verbose version still does a separate dfs of all COs to get the max level. 

Adds to odin.cpp "--stat" and --stat_verbose" arguments.

The parser parse_odin_stats.py processes the output of generate_blif_stats.cpp and dumps it into a csv/tsv. It also can compare outputs and flag major changes etc as part of a set of basic QoR features. The parser could probably be made a bit cleaner in terms of readability but adding/removing metrics is fairly simple.

I left this as a WIP because there were a couple things I wasn't sure about and I wasn't sure to what degree this should be pared down or added to. For example: I left a commented out reversed get_max_levels function in but I don't have any use for it presently, it can be used as a different way of getting path length data but I assume should be removed? I have questions w/r/t the way the max path is calculated. _I will include these questions as well as highlighted code shortly._
In particular the increased overhead of adding data members to the node and net structs should be discussed.

#### Motivation and Context
The intention of the project was to collect data on the netlist struct generated after using -V or -b in Odin-II to quantify possible changes between netlists either as a result of changing the Odin-II logic synthesis process or between Odin and ABC BLIF outputs or other reasons. Basic count information (e.g. #latches) is known or obtained easily from the netlist struct and can be found with the simulator or ABC but other metrics are not as easily obtained, this is a different non-simulator means of collecting basic or verbose stats and at the least provides the framework with which to add other metrics for collection. It is my understanding that there was a netlist_stat struct that was never fully used or implemented. Perhaps it would be easier to start with the non-verbose version, which provides basic desirable information with low overhead.

Passed running "make test" on previous commit, will run again. Unclear what counts as full test coverage. Failed one large memory benchmark because of limitations in testing environment. Works with verilog input and BLIF input although Verilog input will yield very slightly different results. Does not substantially alter existing code aside from odin_types.h and odin_ii.cpp. 

#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/SRC/generate_blif_stats.cpp
+++ b/ODIN_II/SRC/generate_blif_stats.cpp
@@ -1,0 +1,1128 @@
+/*
+Permission is hereby granted, free of charge, to any person
+obtaining a copy of this software and associated documentation
+files (the "Software"), to deal in the Software without
+restriction, including without limitation the rights to use,
+copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following
+conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+#include "odin_types.h"
+#include "odin_globals.h"
+
+#include "netlist_utils.h"
+#include "odin_util.h"
+#include "output_blif.h"
+#include "generate_blif_stats.h"
+
+#include "node_creation_library.h"
+
+#include "multipliers.h"
+#include "hard_blocks.h"
+#include "adders.h"
+#include "subtractions.h"
+#include "vtr_util.h"
+#include "vtr_memory.h"
+
+void depth_first_traversal_to_output_stats(short marker_value, netlist_t *netlist, FILE *out);
+void depth_traverse_output_blif_stats(nnode_t *node, int traverse_mark_number);
+void gather_vcc_gnd_stats(short marker_value, netlist_t *netlist);
+void output_node_stats(nnode_t *node, short traverse_number);
+void depth_traverse_gnd_vcc(nnode_t *node, int traverse_mark_number);
+void depth_traverse_clear_stat(nnode_t *node, int traverse_mark_number , bool hardClear);
+int get_max_levels_rec(nnode_t *node, int traverse_mark_number, int cycleId, netlist_t *netlist);
+int get_max_levels(short marker_value, netlist_t *netlist);
+void single_node_lookahead(nnode_t *node, short travId, short marker);
+//int reverse_lvl(short marker_value, netlist_t *netlist);
+//int reverse_lvl_rec(nnode_t *node, int traverse_mark_number, int cycleId, netlist_t *netlist);
+void node_profile(nnode_t *node, bool verbose, FILE *out);
+int get_node_fanout(nnode_t *node);
+void set_path_lengths(nnode_t *node, short marker);
+int pathlen_heuristic(nnode_t *node, short marker);
+
+static unsigned int num_latches = 0;
+static unsigned int vcc_fanout_live = 0;
+static unsigned int gnd_fanout_live = 0;
+static unsigned int pad_fanout_live = 0;
+static unsigned int vcc_fanout_all = 0;
+static unsigned int gnd_fanout_all = 0;
+static unsigned int pad_fanout_all = 0;
+static unsigned int max_input_pathrange = 0; // this is like reverse net_range bc it is dist to output not depth, abs(min-max pathlength)
+static unsigned int min_input_pathrange = 0-1;
+static float avg_pathrange = 0;
+static unsigned int num_nodes = 0;
+static unsigned int num_carry_func = 0;
+static unsigned int num_names = 0; // consider renaming this to num_gates bc that's what it is
+static unsigned int num_nets = 0; // same as output pin total it seems
+static unsigned int num_edges_fo = 0; // fan out total for all nets (includes vcc and gnd, unclear if it should, abc does not)
+static unsigned int num_edges_fi = 0; // fan in total, will equal num_total_input_pins and thus is redundant
+static unsigned int num_logic_nodes = 0;
+static unsigned int num_total_input_pins = 0; // reflects number of edges in graph that are live, should equal num_edges_fo but may not if there are fo pins w/out node 
+static unsigned int output_pin_total = 0; // this (should) always be the same as num_nets, maybe redundant
+static unsigned int num_subckts = 0;
+static unsigned int num_muxes = 0;
+static unsigned int num_generic_nodes = 0;
+static unsigned int num_input_nodes = 0;
+static unsigned int top_input_fanout = 0;
+static unsigned int num_unconn = 0; // Unreliable, maybe count pad traversals
+static unsigned int num_clock_nodes = 0;
+static unsigned int num_adders = 0;
+
+// This is all metadata on the number of traversals it took to calculate these figures, not passed to parser
+static unsigned int num_traversals_stat = 0; // how many depth first traversals have we done? Internal stat metric to measure cost of collection
+static unsigned int num_traversals_input = 0; // only counts traversals from top inputs (derived value not counter)
+static unsigned int gnd_vcc_pad_traverals = 0; // metadata on measuring gnd and vcc traversals, should be minimal if non-verbose and after all other traversals
+static unsigned int num_clearing_traversals = 0; // node traversals from clearing stat structs
+
+//static unsigned int num_unique_paths = 0; // Experimental, number of unique paths, this would be hard to calculate
+static unsigned int fanin_outputs_total = 0; // this is possibly not a useful stat
+//static unsigned int longest_chain = 1; // we're looking at latches but one might want adders or carries etc.
+//static unsigned int longest_consecutive_chain = 1;
+static unsigned int fanin_node_to_output = 0; // not very interesting
+static unsigned int fanout_node_to_output = 0; // experimental
+static unsigned int num_nodes_leading_to_output = 0; // unless we have fanout in net this will equal fanin_outputs_total, isn't v interesting but needed to derive
+static unsigned int num_deadends = 0; // experimental, every time we hit a deadEnd in path not including already visited nodes
+static unsigned int num_already_visited = 0; // experimental, every time we hit a previously visited node, prob not interesting but maybe
+static unsigned int num_deadend_inputs = 0; // number of inputs that go nowhere, dist values of 0, not including vcc and gnd for now
+static unsigned int max_fanout_node = 0; // highest # of fanout pins of a node including all of its associated nets
+static unsigned int max_outpins_node = 0; // highest num_output_pins of a node, tends to be low
+static unsigned int max_fanout_net = 0; // num_output_pins of net with highest value
+static unsigned int max_fanin_node = 0;
+static int output_node_count = 0; // count outputs so we can figure out if any outputs are unconn
+// our averages do not include zero length paths
+static unsigned int max_dist_out = 0;
+static unsigned int min_dist_out = 0-1;
+static unsigned int gnd_min_path = 0;
+static unsigned int vcc_min_path = 0;
+static unsigned int gnd_max_path = 0;
+static unsigned int vcc_max_path = 0;
+static unsigned int shortest_maxpath = 0-1; 
+static unsigned int longest_minpath = 0;
+static float top_input_avg_dist_max = 0; // avg longest path of top level inputs
+static float top_input_avg_dist_min = 0;
+static float top_lvl_avg_max = 0; // this name is confusing, sum of node->stat->avg_max_dist_to_out of top input nodes, experimental, possibly not useful (average of averages...)
+static unsigned int num_memories = 0;
+static unsigned int max_level = 0; // max number of levels in a CIs logic cone
+static bool sequentialCycles = false; // not for parser
+
+static bool verbose_mode = false; // only set once, hoping to avoid parameter bloat
+
+void output_blif_stats(std::string file_name, netlist_t *netlist, bool verbose)
+{
+	if(verbose)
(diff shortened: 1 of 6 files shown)


mflawn:
Questions for @jeanlego 
1) Some latches have cycles that make tracking the max path length difficult. The minimum distance input to out is more likely to be accurate but the max path length is more valuable for the profile. Are path length metrics worth collecting? At the very least the max_level figure is consistent with what ABC outputs. If longest path is worth collecting, is there a specific way it should be collected? I had a few different approaches that gave different results.

[set_path_lengths function](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d81ac63caf3e30e69deb656fa9649b015bd33c0d/ODIN_II/SRC/generate_blif_stats.cpp#L953)

1b) The function single_node_lookahead is meant to fill in gaps in path information when a node has been visited but has unset path length values because of a loop. It employs a method similar to what would be used to detect a combinational loop, using a different marker than traverse_visited to indicate whether a node is on the current traversal path. The lookahead recursively searches for more visited nodes not on the current path to see if the node's outputs have since had path lengths set. This will get caught in a loop if all fanout pins are tested and both nets and nodes are reset as not on path upon function return. It is unclear which configuration is the correct one as changing whether nodes unmark themselves on return results in different path lengths. This whole process is costly and I'm not sure if it is correct.

[single_node_lookahead resetting next_net traversal_id](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d81ac63caf3e30e69deb656fa9649b015bd33c0d/ODIN_II/SRC/generate_blif_stats.cpp#L1124)

1c) Is adding to the net struct acceptable? This is used in single_node_lookahead but could be used for combinational loop detection though that functionality seems to already exist in netlist_check using net->data.
[net->traversal_id](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d81ac63caf3e30e69deb656fa9649b015bd33c0d/ODIN_II/SRC/include/odin_types.h#L530)

2) Should the clock node be included in stats related to fanout etc? e.g. The max fanout of a primary input.

3) I left in some unused functions that had been part of previous metric calculations, should those be removed?
[path_length_heuristic](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d81ac63caf3e30e69deb656fa9649b015bd33c0d/ODIN_II/SRC/generate_blif_stats.cpp#L1030)
This basically takes a random path from a PI to PO and returns the maximum value.
[reverse_lvl](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d81ac63caf3e30e69deb656fa9649b015bd33c0d/ODIN_II/SRC/generate_blif_stats.cpp#L852)
This is just get_max_level but in reverse, I'm guessing this should be removed.

ODIN_II/SRC/odin_ii.cpp
```diff
@@ -278,34 +280,48 @@ netlist_t *start_odin_ii(int argc,char **argv)
...
-	|| global_args.sim_num_test_vectors 
-	|| global_args.sim_vector_input_file.provenance() == argparse::Provenance::SPECIFIED)
+	/* read the blif */
+	if(global_args.blif_file.provenance() == argparse::Provenance::SPECIFIED || global_args.sim_vector_input_file.provenance() == argparse::Provenance::SPECIFIED)
 	{
 		// if we started with a verilog file read the output that was made since
 		// the simulator can only simulate blifs
-		if(global_args.blif_file.provenance() != argparse::Provenance::SPECIFIED)
-		{
-			configuration.list_of_file_names = { global_args.output_file };
```
jeanlego: this should be kept and maybe moved above this block, but nonetheless it is needed unless you change the read blif function.

ODIN_II/SRC/odin_ii.cpp
```diff
@@ -278,34 +280,48 @@ netlist_t *start_odin_ii(int argc,char **argv)
...
-			current_parse_file =0;
-		}
-
-		try
```
jeanlego: this should be kept and maybe moved above this block, but nonetheless it is needed unless you change the read blif function.

ODIN_II/SRC/odin_ii.cpp
```diff
@@ -278,34 +280,48 @@ netlist_t *start_odin_ii(int argc,char **argv)
...
+	if(global_args.blif_file.provenance() == argparse::Provenance::SPECIFIED || global_args.sim_vector_input_file.provenance() == argparse::Provenance::SPECIFIED)
 	{
 		// if we started with a verilog file read the output that was made since
 		// the simulator can only simulate blifs
-		if(global_args.blif_file.provenance() != argparse::Provenance::SPECIFIED)
-		{
-			configuration.list_of_file_names = { global_args.output_file };
-			current_parse_file =0;
-		}
-
```
jeanlego: this should be kept and maybe moved above this block, but nonetheless it is needed unless you change the read blif function.

ODIN_II/parse_odin_stats.py
```diff
@@ -0,0 +1,948 @@
...
+    elif(code==4):
+        print("No input specified")
+    elif(code==5):
+        print("Formatting error: check log file")
+    elif(code==6):
+        print("Wrong number of arguments, expecting [./exec -c <input.log> <input2.log> (opt)<output.csv>]")
+    elif(code==7):
+        # this is our only mechanism against the possibility of overwriting the log file, important that input files are .log
+        print("\nImproper output format, output file must not begin with '-' or end in .log, use no args or -h/--help for proper format")
+
```
jeanlego: rather than multiple print statement use
```
print '''
your multiline
text
goes here
'''
```

jeanlego: move the help text as close to the top of this file ( maybe make an array of multiline string at the top and use the code as the index) this way it documents the usage of this python script. It will be the first thing you see when you open the file

ODIN_II/parse_odin_stats.py
```diff
@@ -0,0 +1,948 @@
...
+        return True
+    else:
+        return False
+
+# many different ways to approach this, unclear which is right way, just proof of concept, for now it flags anything +/- threshold limit
+# Change to not be abs value if only increases should be flagged
+def within_threshold(thresholds, category, diff, percentChange):
+    ret_str = ""
+    limit = thresholds.get(category, "N/A") #this doesn't work as hoped
+    if limit == "N/A":
```
jeanlego: use None rather than "" and "N/A" to keep it more consistent
i.e
ret_str = None
limit == None

ODIN_II/parse_odin_stats.py
```diff
@@ -0,0 +1,948 @@
...
+    # TODO Streamline all this args stuff into something more elegant
+    # Minimum of 3 inputs required
+    # TODO (maybe) Could do diffs on these as well as long as they are bunched in twos but that is perhaps overkill
+    # TODO Maybe add percent changes as a fourth line
+    # this hinges on the x.abc.blif a.odin.blif format generated by run_vtr_flow and supplied in alphabetical order
+    # This is kind of the inverse of how we presently use compare mode, which assumes the 'reduced' log will be compared against input1
+    # NOTE This now by default does not produce flag files for -mc mode (but still flag warnings) unless "--flags explicitly specified 
+    # Also --concise flag disables printing of summary to stdout
+    # TODO Multicompare works fine with 2 inputs and so compareMode is somewhat redundant, only different in defaulting to flags on
+    # Another way to do multicompare would be to test every line against every previous line, not sure if that is better or worse than using pairs"""
```
jeanlego: keep your multiline """ markers location consistent

jeanlego: I would add the TODO to the top bellow the help text since it is insightful to know ahead of time

ODIN_II/parse_odin_stats.py
```diff
@@ -0,0 +1,948 @@
...
+    elif 'M' in size_in or 'm' in size_in:
+        size_in = str(assert_float(size_in.replace('M','').replace('m','')) / 1024.0 / base_size)
+
+    elif 'K' in size_in or 'k' in size_in:
+        size_in = str(assert_float(size_in.replace('K','').replace('k','')) / 1024.0 / base_size)
+
+    #we are now in byte
+    return str(assert_float(size_in) * 1024.0 * 1024.0)
+
+def contains(line_in, on_tokens):
```
jeanlego: I don't know that this will ever come up as an issue but this matches the on_tokens in any order on the line, If this is not intended I would revise it such that they stay in order, match first token, match second token on substring after first, etc...

ODIN_II/parse_odin_stats.py
```diff
@@ -0,0 +1,948 @@
...
+    for header in coreStats:
+        wrt += header + ": "+str(percentage)+"\n"
+    ft.write(wrt)
+
+# TODO the parser really doesn't like newline after word w/out space, which could cause problems for people creating their own threshold files
+def parse_line(benchmarks, line):
+
+    line.strip()
+    line.strip("\n")
+    line = " ".join(line.split())
```
jeanlego: I think that simply doing ```line = " ".join(line.split()``` should achieve the same thing and fix your TODO above

ODIN_II/SRC/generate_blif_stats.cpp
```diff
@@ -0,0 +1,1128 @@
...
+	return;
+}
+
+/*-------------------------------------------------------------------
+ * (function: pathlen_heuristic)
+ * The idea of this is to do a dfs starting at random fanout node and run many times to see if this gives us any kind of indication of what mean path lengths should be.
+ * Ideally would probably iterate over the precreding fanout nodes after as well. Over enough iterations the results are consistent.
+ * Very very costly to just repeat this many times.	
+ *------------------------------------------------------------------*/
+int pathlen_heuristic(nnode_t *node, short marker)
```
jeanlego: I would simply call it "aproximate_longest_path"  and adjust the comment to reflect that it returns an approximated max

ODIN_II/SRC/generate_blif_stats.cpp
```diff
@@ -0,0 +1,1128 @@
...
+static unsigned int num_clearing_traversals = 0; // node traversals from clearing stat structs
+
+//static unsigned int num_unique_paths = 0; // Experimental, number of unique paths, this would be hard to calculate
+static unsigned int fanin_outputs_total = 0; // this is possibly not a useful stat
+//static unsigned int longest_chain = 1; // we're looking at latches but one might want adders or carries etc.
+//static unsigned int longest_consecutive_chain = 1;
+static unsigned int fanin_node_to_output = 0; // not very interesting
+static unsigned int fanout_node_to_output = 0; // experimental
+static unsigned int num_nodes_leading_to_output = 0; // unless we have fanout in net this will equal fanin_outputs_total, isn't v interesting but needed to derive
+static unsigned int num_deadends = 0; // experimental, every time we hit a deadEnd in path not including already visited nodes
```
jeanlego: these are actually very bad to have since anything that does not drive an output node should be cleaned.   we should warn to do a cleanup phase beforehand

ODIN_II/SRC/generate_blif_stats.cpp
```diff
@@ -0,0 +1,1128 @@
...
+		fprintf(out, "(W/ clock %d top inputs)\n", netlist->num_top_input_nodes); // includes clock, not in parser
+		fprintf(out, "num_deadend_inputs: %u\n", num_deadend_inputs); // number of inputs that go nowhere, dist values of 0
+		fprintf(out, "num_clock_nodes: %u\n", num_clock_nodes); // this doesn't include global sim base clock which is confusingly of type input
+		fprintf(out, "num_logic_nodes: %u\n", num_logic_nodes);
+		fprintf(out, "num_carry_func: %u\n", num_carry_func);
+		fprintf(out, "num_generic_nodes: %u\n", num_generic_nodes);
+		fprintf(out, "num_unconn: %u\n", num_unconn); // unreliable, how would we find an unconn node?
+		fprintf(out, "num_memories: %u\n", num_memories);
+		fprintf(out, "num_adders: %u\n", num_adders);
+		fprintf(out, "num_deadends: %u\n", num_deadends); // unclear if fully reliable
```
jeanlego: I would make sure it is reliable and later down the road a sanitize phase should clean this up

jeanlego:
to calculate the path lengths I would go about it using recursion starting from output pins to input pins and if the next node is already traversed don't recurse on it. 

Also after consideration, I believe keeping stats in the nnet_t may be preferable since it uses a one-to-many relationship (one driver, many outputs) and would then be the stats for a single driver.

For a node to get the stats, it would iterate through its driver every time.
The reasoning behind this is that we may use this dynamically throughout the lifespan of the netlist and even when it changes.
Doing so would allow to check if the nnet_t stats_traversal_nb != driver_pin->node->stats_traversal_nb to recalculate the driver bottom up recursively and keep recalculating everytime theres a mismatch.
Once this is in place, it would allow to get stats from any node throughout the circuit and gather insight on optimization done without recomputing everything, only what is needed.

I would be careful about using flags already there unless you know all the side-effects from using it. It is preferable to reuse if you can, but this could be work done down the line.

mflawn:
> to calculate the path lengths I would go about it using recursion starting from output pins to input pins and if the next node is already traversed don't recurse on it.
> 
> Also after consideration, I believe keeping stats in the nnet_t may be preferable since it uses a one-to-many relationship (one driver, many outputs) and would then be the stats for a single driver.
> 
> For a node to get the stats, it would iterate through its driver every time.
> The reasoning behind this is that we may use this dynamically throughout the lifespan of the netlist and even when it changes.
> Doing so would allow to check if the nnet_t stats_traversal_nb != driver_pin->node->stats_traversal_nb to recalculate the driver bottom up recursively and keep recalculating everytime theres a mismatch.
> Once this is in place, it would allow to get stats from any node throughout the circuit and gather insight on optimization done without recomputing everything, only what is needed.
> 
> I would be careful about using flags already there unless you know all the side-effects from using it. It is preferable to reuse if you can, but this could be work done down the line.

The thing is, I don't know that the path lengths currently calculated are wrong, I just have no way to verify whether they are right or wrong. They're right for non-looping netlists as far as I can tell at least. I've tried output-to-input recursive traversal and the results aren't especially different. As for storing the stats in the net struct, that has also occurred to me as possibly making more sense but as I recall the results weren't especially different. I'll try out the bottom up net-centric approach though.

jeanlego:
> The thing is, I don't know that the path lengths currently calculated are wrong, 
You can safely assume there is no loop in the circuit so long as there are no latches, if you encounter a latch it should close its own loop or have no loop.

With this in mind, you can easily make simple circuits enough to traverse them pen and paper and compute the statistics you are looking to compute with the tool and compare your results. We ould make these part of the regression test and it would assert basic functionality of your implementation. Once we have these, we can be a little bit more confident that the statistics are correct on larger corcuit

ODIN_II/SRC/generate_blif_stats.cpp
mflawn: Deadends only seem to be encountered when using a .v file as input instead of a BLIF, these deadends then seem to get cleaned up as part of the process of writing out to a BLIF (and thus is not reflected in this stat profiling). I'm not sure if it's in the spirit of profiling a netlist to then alter it however, although I can have it throw warnings on encountering a deadend.

Also it counts top input nodes (excluding pad) with no outgoing edges as a form of deadend and thus always considers the global_sim_base_clock as a deadend, I'm not sure if this should be counted as such.

jeanlego:
superseded by #1174
closing

PR #1069, opened 2019-12-20
Replace some uses of malloc/free with std::unique_ptr

HackerFoo:
#### Description
`malloc()` and `free()` can be error-prone, so I'm working on replacing them with safer alternatives, but without adding overhead.

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/506

#### Motivation and Context
See issue above.

I've run `vtr_reg_strong` tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/check_route.cpp
+++ b/vpr/src/route/check_route.cpp
@@ -42,9 +42,7 @@ void check_route(enum e_route_type route_type) {
     int max_pins, inode, prev_node;
     unsigned int ipin;
     bool valid, connects;
-    bool* connected_to_route; /* [0 .. device_ctx.rr_nodes.size()-1] */
     t_trace* tptr;
-    bool* pin_done;
 
     auto& device_ctx = g_vpr_ctx.device();
     auto& cluster_ctx = g_vpr_ctx.clustering();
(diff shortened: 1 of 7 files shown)


HackerFoo:
Maybe `std::vector` can be used if the overhead of the bounds checks are acceptable.

mithro:
Looks like this is failing on valgrind test?

HackerFoo:
I couldn't convert the heap to use `new[]`/`delete[]` without larger changes:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/4ff2a673884f6fda134a42ce3d9ac46753ad547d/vpr/src/route/route_common.cpp#L470-L472

The right thing to do is probably to use a `std::vector`.

kmurray:
> Maybe `std::vector` can be used if the overhead of the bounds checks are acceptable.

I agree, unless we have a particular reason otherwise using a std::vector makes sense. There should not be any overhead with the regular operator[] on a vector, as it doesn't do bounds checking (only `at()` does).

HackerFoo:
I'll try replacing the malloc'ed heap with a std::vector, but I'd like to do that in another PR.

kmurray:
> I'll try replacing the malloc'ed heap with a std::vector, but I'd like to do that in another PR.

As in, merge this PR and do the vector conversion in a follow up PR? Or just a new PR? Either way seems reasonable to me.

HackerFoo:
The first option. I'd like to keep the PRs small.

kmurray:
Sounds good, thanks!

PR #1074, opened 2020-01-22
Refactor vtr::flat_map operator []

litghost:
#### Description

Previous implementation of the flat_map `operator []` did a `find` followed by an `insert`.  This effectively did a binary search twice (`find` calls `std::lower_bound`, and so does `insert`).  The correct optimal solution is to use `std::lower_bound`, and avoid the second lookup.

#### Motivation and Context

This makes using `operator []` faster by avoiding a second call to `std::lower_bound`.

- [x] CI is green
- [x] Change was profiled before and after

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libvtrutil/src/vtr_flat_map.h
+++ b/libs/libvtrutil/src/vtr_flat_map.h
@@ -122,13 +122,22 @@ class flat_map {
     }
 
     mapped_type& operator[](const key_type& key) {
-        auto iter = find(key);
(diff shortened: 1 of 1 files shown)


kmurray:
Yep looks like a good enhancement.

@litghost Can you fix the formatting?  After that it should be good to merge.

litghost:
> @litghost Can you fix the formatting? After that it should be good to merge.

Fixed

kmurray:
Thanks!

PR #1075, opened 2020-01-22
Use 2 boolean's per physical types instead of std::set to mark IO types.

litghost:
#### Description

`std::set` is a fairly heavy-weight data structure, and consumes more
memory and is slower than a boolean.

#### Motivation and Context

Using a `std::set` is both slower and more memory intensive than simply using two booleans.  The new code no longer accommodates iterating over input or output tiles, but that was never done in the old code.

- CI is green
- Profiled before and after

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -620,7 +620,13 @@ struct t_physical_tile_type {
     /* Returns the indices of pins that contain a clock for this physical logic block */
     std::vector<int> get_clock_pins_indices() const;
 
+    // TODO: Remove is_input_type / is_output_type as part of
+    // https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1193
+
+    // Does this t_physical_tile_type contain an inpad?
     bool is_input_type;
+
+    // Does this t_physical_tile_type contain an outpad?
     bool is_output_type;
 };
 
@@ -1347,11 +1353,11 @@ enum class BufferSize {
  *            we would expect an additional "internal capacitance"           *
  *            to arise when the pass transistor is enabled and the signal    *
  *            must propogate to the buffer. See diagram of one stream below: *
- *                                                                           *   
+ *                                                                           *
  *                  Pass Transistor                                          *
  *                       |                                                   *
  *                     -----                                                 *
- *                     -----      Buffer                                     *   
+ *                     -----      Buffer                                     *
  *                    |     |       |\                                       *
  *              ------       -------| \--------                              *
  *                |             |   | /    |                                 *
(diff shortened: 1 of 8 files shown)


kmurray:
Can you comment on how this effects the run-time behaviour?

It seems to me that this is fairly cold-path code so I wouldn't expect too much of a difference.

litghost:
> Can you comment on how this effects the run-time behaviour?
> 
> It seems to me that this is fairly cold-path code so I wouldn't expect too much of a difference.

This path is hot during initial placement:
```
rc/place/initial_placement.cpp
286:        if (!(is_io_type(pick_best_physical_type(logical_block)) && pad_loc_type == USER)) {
317:            if (is_io_type(pick_best_physical_type(logical_block)) && pad_loc_type == RANDOM) {
```

`is_io_type` is `is_input_type || is_output_type`.

kmurray:
Initial placement is usually extremely fast, so I don't expect this to be a big issue. Do you have any examples/measurements which show an issue?

litghost:
So when I original made this change, it was driven based on profiler data showing that at least 1-2% of a P&R run was spent here.  Can I ask why this PR is getting such scrutiny?  A `std::set<t_physical_tile*>` requires roughly 2 pointers and a dynamic allocation per tile instance and querying a `std::set` requires pointer chasing.  This in comparison with 2 bool, which consumes two bytes if packed.

The only cost to this change is that the size of t_physical_tile increases by ~2-8 bytes per type, but we say ~16 bytes per IO tile.  In a simple arch with 1 IO type, there is small memory footprint increase.  But complicated arches have multiple IO types, 7-series has 8 IO types, so not only is the overahead of `std::set` removed, but the overall memory footprint decreases.

libs/libarchfpga/src/physical_types.h
```diff
@@ -619,6 +619,9 @@ struct t_physical_tile_type {
 
     /* Returns the indices of pins that contain a clock for this physical logic block */
     std::vector<int> get_clock_pins_indices() const;
+
+    bool is_input_type;
```
vaughnbetz: Should at least comment these (what do they mean)?
Also say most code should try not to use and be generic for all block types if possible (could deprecate at that point).

vaughnbetz: Keith can you do a grep to figure out how many call sites that use this now.
Could we get rid of these variables?

vaughnbetz: I have no strong opinion about changing these, but would rather get rid of them.

litghost: Comments in and issue filed

PR #1076, opened 2020-01-22
Convert 5-level nested vectors to 3-level array/NdMatrix/vector.

litghost:
#### Description

By converting the first 4 layers into array+NdMatrix, the memory
locality is improved and the number of pointer chases changes from 5 to
2ish.

#### Motivation and Context

There is almost never a time where 5 nested std::vector's are the right data structure.  Changing the middle 3 std::vector's to an `NdMatrix` isn't ideal, but it is better.  Given that root vector is fixed size, using std::array instead of std::vector saves one layer of indirection.

- CI is green
- Profiled change before and after

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libvtrutil/src/vtr_ndmatrix.h
+++ b/libs/libvtrutil/src/vtr_ndmatrix.h
@@ -33,6 +33,8 @@ class NdMatrixProxy {
         , dim_strides_(dim_strides)
         , start_(start) {}
 
+    NdMatrixProxy<T, N>& operator=(const NdMatrixProxy<T, N>& other) = delete;
+
     const NdMatrixProxy<T, N - 1> operator[](size_t index) const {
         VTR_ASSERT_SAFE_MSG(index >= 0, "Index out of range (below dimension minimum)");
         VTR_ASSERT_SAFE_MSG(index < dim_sizes_[0], "Index out of range (above dimension maximum)");
@@ -64,6 +66,8 @@ class NdMatrixProxy<T, 1> {
         : dim_sizes_(dim_sizes)
         , dim_strides_(dim_stride)
         , start_(start) {}
+
+    NdMatrixProxy<T, 1>& operator=(const NdMatrixProxy<T, 1>& other) = delete;
 
     const T& operator[](size_t index) const {
         VTR_ASSERT_SAFE_MSG(dim_strides_[0] == 1, "Final dimension must have stride 1");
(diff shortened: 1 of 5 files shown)


kmurray:
> By converting the first 4 layers into array+NdMatrix, the memory
> locality is improved and the number of pointer chases changes from 5 to
> 2ish.

Yep, I've contemplated similar changes in the past, but never got round to them.

One subtlety of this data structure is the usage of the [side] dimension. It is actually only relevant for IPIN/OPIN types. The other types (CHANX/CHANY/SOURCE/SINK) don't actually have a side, and [just use the 0th side](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph2.cpp#L1285). Sub-optimal but gave a uniform-ish interface.

So one drawback of the proposed approach is that we no longer treat the [side] dimension as sparse,  potentially paying an additional NUM_SIDES overhead on 4 of 6 RR types.

We should probably measure this, and if its an issue we could do something like:

```
typedef std::array<vtr::NdMatrix<std::vector<std::vector<int>, 2>, NUM_RR_TYPES> t_rr_node_indices;
```
Where we use:
 * the std::array for the RR type dimension
 * NdMatrix<2> for the grid x & y dimension
 * vector for the side dimension (single entry for most RR types)
 * vector for the number of elements

A more drastic change would be to refactor the storage of this info separately for PINs and non-PINs to avoid the SIDE overhead entirely for non-PINs.

> Given that root vector is fixed size, using std::array instead of std::vector saves one layer of indirection.

This definitely makes sense.

litghost:
> So one drawback of the proposed approach is that we no longer treat the [side] dimension as sparse, potentially paying an additional NUM_SIDES overhead on 4 of 6 RR types.

The old code did _not_ take advantage of that fact, this change completely preserves the previous dimensionality.

If we really wanted to go crazy and make this memory ideal, it would actually be a structure.  But I think that is a follow up enhancement.  This is still an improvement, with no downside, and the same dimensions are the previous code.

kmurray:
> The old code did _not_ take advantage of that fact, this change completely preserves the previous dimensionality.

Yep, you're right!

If the formatting is fixed this should be good to merge.

litghost:
The formatting error was in master, but I've rebased now so it should go green.

kmurray:
Thanks!

PR #1077, opened 2020-01-22
Support for string internment and use interned strings for metadata

litghost:
#### Description

String internment is basically the idea that if you know you have a small pool of strings relative to the number of total strings, you can save memory and maybe save CPU time by interning the strings.  Then the leafs of the code have intern ids rather than the full strings.

This PR adds a simple string internment using `std::unordered_map` for use when interning strings, and a `std::vector` for intern id to string lookup. This simple string internment does use 2x the memory for string storage, but is straight forward to implement and verify.

This PR also converts the metadata structures to use interned string ids instead of `std::string`.  This is not critical for metadata on objects in physical types, but is critical when there is metadata that is on a per edge or per node basis.  The CPU overhead of build the string internship is small (order 2.5 seconds for 22 million internments), but the memory reduction is significant if the strings and/or substrings are repeated, which is very typical.

This PR also adds a `MetadataStorage` object, which performs delay'd map generation.  Because node and edge metadata is always insert all data, followed by query, a simple two-phase data structure can be used.  This saves CPU time by avoiding the needed to maintain a live map during initial insertion.

#### Motivation and Context

This is a CPU - memory trade-off.  On the artix 50T, the cost of storing edge metadata is roughly 5 GiB of memory without internment, and that memory usage scales linearly with the number of edges (22 million on the 50T).  By using string internment, the memory used still scales linearly, but the constant factor and number of allocations drops dramatically.  If `absl` is added to VTR, the number of allocations can potentially be brought to 0 using `absl::InlinedVector`.

- CI passes
- CPU and memory usage profiled before and after change.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -107,6 +107,89 @@ struct hash<std::tuple<int, int, short>> {
 };
 } // namespace std
 
+// MetadataStorage is a two phase data structure.  In the first phase,
+// metadata is added cheaply by simply pushing onto a vector.  This is
+// unsuitable for lookup, but is fast, and because strings are interned, not
+// too expensive memory wise.
+//
+// The second phase occurs after all data has been inserted.  When/if a lookup
+// is needed, the vector is sorted and the final metadata lookup is generated.
+// In the event that the lookup is never needed, this saves the time spent
+// building the lookup, and reduces the number of outstanding memory
+// allocations dramatically.
+template<typename LookupKey>
+class MetadataStorage {
+  public:
+    void add_metadata(const LookupKey& lookup_key, vtr::interned_string meta_key, vtr::interned_string meta_value) {
+        // Can only add metadata prior to building the map.
+        VTR_ASSERT(map_.empty());
+        data_.push_back(std::make_tuple(lookup_key, meta_key, meta_value));
+    }
+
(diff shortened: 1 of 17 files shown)


vpr/src/base/vpr_context.h
```diff
@@ -107,6 +107,105 @@ struct hash<std::tuple<int, int, short>> {
...
+// unsuitable for lookup, but is fast, and because strings are interned, not
+// too expensive memory wise.
+//
+// The second phase occurs after all data has been inserted.  When/if a lookup
+// is needed, the vector is sorted and the final metadata lookup is generated.
+// In the event that the lookup is never needed, this saves the time spent
+// building the lookup, and reduces the number of outstanding memory
+// allocations dramatically.
+template<typename LookupKey>
+class MetadataStorage {
```
kmurray: Generally I've tried to keep the lower-level data structure definitions/details/implementations out of vpr_context.h. Since that header is really key to tracing/finding data in VTR I think its very important to keep it as short and straightforward to traverse as possible.

As a result, I'd suggest moving this class somewhere else. Traditionally this has been `vpr_types.h` but that file is too big. Perhaps move it into a 'metadata_storage.h' in either `vpr/src/base` or `vpr/src/util`?

litghost: Fixed

kmurray:
Thanks!

Issue #1079, opened 2020-01-23
RR graph edge storage refactoring


litghost:
#### Proposed Behaviour

The edge storage should be optimized for router behavior, which is driven by two concerns:
 - Ability to iterate quickly (see https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/route_timing.cpp#L1984-L1986)
 - Memory locality

The current allocation strategy has poor memory locality, and requires a pointer per node.

#### Current Behaviour

Currently rr graph edges are stored in per node allocations.  This is a reasonable solution when edges are being mutated, but once edges are fixed (either after rr graph construction or if the rr graph is read from file) having one allocation be node is overly flexible.

#### Possible Solution

Assuming that during writing, node edges are only written and never read, and afterwards node edges are only read and never written, a two phase data structure can be used.

litghost:
FYI @vaughnbetz / @kmurray this is the issue we were talking about

@tangxifan This is issue is relevant to your refactoring.  @tangxifan in particular, your first iteration (https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/device/rr_graph_obj.h#L804-L808) preserves this same 1 allocation per rr node).

I'm going to be making several further posts fleshing out my notes on this change.

litghost:
Current usage of the edge interfaces
============================

All rr node edge access is controlled by the `t_rr_node` object.  

There are 4 classes of methods:
 - Get node to edge relationships (6 methods)
 - Get data for an edge (2 methods)
 - Node to edge mutation (4 methods)

Methods: Get node to edge relationships
------------------------------------------------------
There are 6 methods in this category:
 - `t_rr_node::edges`
 - `t_rr_node::configurable_edges`
 - `t_rr_node::non_configurable_edges`
 - `t_rr_node::num_edges`
 - `t_rr_node::num_configurable_edges`
 - `t_rr_node::num_non_configurable_edges`

Usage:
- `num_edges` is used heavily (57 callsites).
- Both `num_non_configurable_edges` and `num_configurable_edges` is never called outside of `t_rr_node` itself.
- `configurable_edges` and `non_configurable_edges` are used infrequently:

```
src/route/route_tree_timing.cpp
388:        for (int iedge : device_ctx.rr_nodes[rr_node].non_configurable_edges()) {

src/route/check_rr_graph.cpp
[...]

PR #1082, opened 2020-01-24
Adding the kokoro CI configuration.

litghost:
Kokoro is a Google internal service which allows significantly bigger
and longer CI builds using Google Cloud Platform.

This enables us to run all the vtr_reg_strong, vtr_reg_nightly and
vtr_reg_nightly on every pull request. This will happen once the
@kokoro-team has been given access to the repo and the web hook set up.

Examples of the outputs from Kokoro can be seen at SymbiFlow/symbiflow-arch-defs#1170 and https://source.cloud.google.com/results/invocations/810121e5-fd2a-41d4-bad1-ac0859d597a6/log

Copy of #1054

PR #1084, opened 2020-01-24
Flyweight rr node

litghost:
#### Description

This is laying the ground work for refactoring rr node and rr edge memory layout and storage.  This replaces the t_rr_node object with a fly-weight object.  The fly-weight then references a backing array that is identical to the current storage method.

By using a fly-weight object, anything that was taking references to the rr_nodes array stopped working.  This PR changed those to use values instead of pointers/references.

This PR builds on #1081

#### Related Issue

#1079 
#1048

#### Motivation and Context

This is change should be a no-op, which should be reflected in the QoR.

 - [x] Basic and strong CI passes
 - [ ] Run and check nightly and weekly QoR

#### Types of changes

#### Checklist:


litghost:
@tangxifan FYI, this pattern could be useful for your implementation as well.

litghost:
So preliminary examination of the nightly results show that this PR has basically no change in QoR.  I'll post comparisons later today.

Weekly QoR data is running now against this PR.  I don't have baseline against https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/2780988d58da6be454364e4a06cff8da8caf09e8, but if the weekly in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1082 completes, we could probably use that data.

vaughnbetz:
Keith, doesn't this cost us an additional indirection on all rr_node accesses?  
If we fly-weight the whole rr_node, it doesn't seem like we'll have any nodes that are the same in all respects, and hence the fly-weight would devolve to one unique entry per rr_node, but with an extra indirection to get at the data.  Is that right, or am I missing something?

litghost:
> Keith, doesn't this cost us an additional indirection on all rr_node accesses?
> If we fly-weight the whole rr_node, it doesn't seem like we'll have any nodes that are the same in all respects, and hence the fly-weight would devolve to one unique entry per rr_node, but with an extra indirection to get at the data. Is that right, or am I missing something?

There is no extra indirection, because `t_rr_node` is a value object.  For example, where-as before we would write:
```
t_rr_node &node = rr_nodes[idx];
```

Which is really just:
```
t_rr_node *node = rr_nodes + idx
```

The new code returns the base pointer + offset, so the same code becomes:
```
(t_rr_node *first_node, size_t offset) = (rr_nodes, idx)
```

The compiler can then inline the rest, which is what we see.

litghost:
[titan_small_comp.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4118044/titan_small_comp.xlsx)
[comp.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4118045/comp.xlsx)

litghost:
@kmurray / @vaughnbetz It actually looks like the flyweight RR node is slight win on the nightly QoR metrics!  Please review when you get a chance.  Weekly reg tests are still running,

vpr/src/route/rr_node.h
```diff
@@ -157,43 +122,21 @@ class t_rr_node {
...
-
-    union {
-        int16_t pin_num;
-        int16_t track_num;
-        int16_t class_num;
-    } ptc_;
-    t_edge_size fan_in_ = 0;
-    uint16_t capacity_ = 0;
+    t_rr_node_storage* storage_;
+    RRNodeId id_;
```
kmurray: A thought on a potential optimization.

VPR only has a single RRGraph at any point in time. Instead of paying the overhead for:
```
t_rr_node_storage* storage_;
```
in each t_rr_node proxy object, you could just replace it with a call to:
```
g_vpr_ctx.device().rr_nodes;
```
where needed, which would make `sizeof(t_rr_node) == sizeof(RRNodeId)`.

While in the long-term I think we should really move to better encapsulating the RR graph as a whole this would be a temporary optimization.

litghost:
> I think we'll need to see the results on the full Tita23 set to really evaluate this from a QoR perspective however.

I've hit a hickup in the QoR comparisions, in that the quality is different on this PR versus baseline.  This is hard to explain right now, so I'm looking for answers.  I suspect using unstable sort's and partitions are part of the issue, and I'm working on demostrating that this is the case.  Either way, we should hold off on merge this PR until I figure out what is going on.

kmurray:
OK.

Although if unstable sorting/ordering is the cause I'm not too worried (assuming the QoR change is small of course!). Good to verify that is the case.

litghost:
> I've hit a hickup in the QoR comparisions, in that the quality is different on this PR versus baseline. This is unexpected, as there should be zero impact on quality (e.g. CPD).

I believe I've isolate the QoR change, which was due to an old baseline.  The new baseline is running, and I have full QoR from #1084 and #1085.  #1096 is running right now, and is showing reasonable performance with good memory behavior.  Once I have QoR results on all 4 (baseline, rr proxy, refactor edges, memory clean), we can talk about how to proceed.

litghost:
[refactor_comp.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4148984/refactor_comp.xlsx)

Baseline QoR is in.  The compares a baseline commit with #1084 (proxy rr node) and #1085 (refactor edges).

Summary:
 - Quality is unchanged
 - Memory used for #1084 is unchanged as expected
 - #1085 causes an expected 4% memory increase
 - #1084 causes a 20% overall CPU time increase
 - #1085 causes a 5% overall CPU time increase

Preliminary results from #1096 are (bitcoin miner, and gaussian blur up through router lookahead and placer delay matrix):
 - #1096 causes a 5% overall CPU time increase
 - #1096 reduces memory usage by 35% - 60%

This week I'm working on recovering some of that 5% CPU time increase, I believe reordering some memory loads to make the new memory pattern prefetchable will close the gap.

litghost:
@kmurray / @vaughnbetz FYI

PR #1098, opened 2020-01-30
Remove empty statement in switch/case. Add missing break in another.

hzeller:
o The semicolon in the empty fallthrough statement in netlist.tpp
  triggers fallthrough warnings, so removing the semicolon better
  reflects the intentions.
o check_rr_graph: there is a missing break in the switch/case which
  happens to not introduce a bug as the following cases also break.
  Add 'break' to reflect intention of code.


kmurray:
Thanks for the fix!

PR #1104, opened 2020-02-03
Docs: Adding clock architecture documentation

mustafabbas:
Documentation for the clock architecture spec

#### Related Issue
Addresses part of https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/928#issuecomment-555432218

#### Motivation and Context
Adds documentation for an implemented architecture option

N/A

#### Types of changes
Documentation 

#### Checklist:


mustafabbas:
@HackerFoo cant seem to add you as a reviewer for some reason. But mentioning you here in case you find this helpful

mithro:
FYI - @litghost

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+            <!-- connections from inter-block routing to central spine -->
+            <tap from="ROUTING" to="spine1.drive" locationx="W/2" locationy="H/2" switch="general_routing_switch" fc_val="1.0"/>
+            <!-- connections from spine to rib -->
+            <tap from="spine1.tap" to="rib1.drive" switch="general_routing_switch" fc_val="0.5"/>
+            <!-- connections from rib to clock pins -->
+            <tap from="rib1.tap" to="CLOCK" switch="ipin_cblock" fc_val="1.0"/>
+        </clock_routing >
+    </clocknetworks >
+
+The ``<metal_layers>`` element describes the per unit length electrical parameters used to implement the clock distribution  wires. There can be one or more wiring implementation options (metal layer, width and spacing) that are used by the later  clock network specification and each is described in a separate ``<metal_layer>`` sub-element. The syntax of the wiring electrical information is:
```
kmurray: I'd likely add that the metal layer width and spacing are the physical implementation characteristics which result in the Rmetal/Cmetal but that we only model them as those characteristics (as currently written someone might expect width/spacing options for the metal layers).

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+    :req_pram Cmetal: The capacitance in Farads of the wire per unit block.
+
+The ``<clock_network>`` element contains sub-elements that describe the clock distribution wires for the clock architecture.  There could be more than one ``<clock_network>`` element to describe separate types of distribution wires. The high-level start tag for a clock network is as follows:
+
+.. arch:tag:: <clock_network name="string" num_inst="integer">
+
+    :req_param name: A unique string for reference.
+    :req_param num_inst: which describes the number of parallel instances of the clock distribution types described in the ``<clock_network>`` sub-elements.
+
+The supported clock distribution types are ``<spine>`` and ``<rib>``. ``Spine`` is used to describe vertical clock distribution wires. Whereas, ``Rib`` is used to describe a horizontal clock distribution wire.
```
kmurray: I think we should have diagrams illustrating what these options are (c.f. https://docs.verilogtorouting.org/en/latest/arch/reference/#grid-location-tags).

Probably one each for spine and rib. It would also be good to label the various options (e.g. x, startx, starty, endy, repeatx, repeaty).

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+
+.. arch:tag:: <spine metal_layer="string" x="integer" starty="integer" endy="integer" repeatx="integer" repeaty="integer"/>
```
kmurray: Many of these parameters aren't actually integers but expressions (e.g. `H`, `H/2`, `W`). Should probably change the type to something like 'expr'.

We should also have a section which just describes what the expressions are and what the variables mean (e.g. W, H). If I recall correctly these are the same variables/expressions as used in grid construction, so we can likely just cross-ref to https://docs.verilogtorouting.org/en/latest/arch/reference/#grid-location-expressions. Although we should remind readers here that W and H are the device grid width/height and that `w` and `h` aren't available.

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+    :req_param metal_layer: A referenced metal layer that sets the unit resistance and capacitance of the distribution wire over the length of the wire.
+    :req_param starty: The start y grid location, of the wire which runs parallel to the y-axis from starty and ends at endy, inclusive. Value can be relative to the device size.
+    :req_param endy: The end of y grid location of the wire. Value can be relative to the device size.
+    :req_param x: The location of the spine with respect to the x-axis. Value can be relative to the device size.
+    :opt_param repeatx: The horizontal repeat factor of the spine along the device. Value can be relative to the device size.
+    :opt_param repeaty: The vertical repeat factor of the spine along the device. Value can be relative to the device size.
+
+The provided example clock network defines two spines, and neither repeats as each spans the entire height of the device and is locally at the horizontal midpoint of the device.
```
kmurray: It would be good to add a cross-reference back to the example you are referring to. Likely means you'll need to label the example code block so you can `:ref:` it.

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+
+.. arch:tag:: <switch_point type="{drive | tap}" name="string" yoffset="integer" xoffset="integer" xinc="integer" yinc="integer" buffer="string">
+    :req_param type:
+        * ``drive`` -- Drive points are where the clock distribution wire can be driven by a routing switch or buffer.
+        * ``tap`` --  Tap points are where it can drive a routing switch or buffer to send a signal to a different ``clock_network`` or logicblock.
+    :opt_param xoffset: (Only for ``rib`` network) Offset from the ``startx`` of a rib network.
+    :opt_param yoffset: (Only for ``spine`` network) Offset from the ``starty`` of a spine network.
+    :opt_param xinc: (Only for rib ``tap`` points) Descibes the repeat factor of a series of evenly spaced tap points.
+    :opt_param yinc: (Only for spine ``tap`` points) Descibes the repeat factor of a series of evenly spaced tap points.
+    :req_param buffer: (Required only for ``drive`` points) A reference to a pre-defined routing switch. This switch will be used at the drive point. The clock generator uses two buffers to drive the two portions of this ``clock_network`` wire when it is split at the drive point.
```
kmurray: Probably want a cross-ref to the `<switch>` section so leaders know how these buffers are defined.

The part about two buffers is a bit confusing. I'd suggest adding a diagram to illustrate this. Also in general there may be >2 sub-portions of the clock network, so you may want to rephrase in terms of 'N' buffers where the clock network splits into N parts.

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+    :req_param metal_layer: A referenced metal layer that sets the unit resistance and capacitance of the distribution wire over the length of the wire.
+    :req_param startx: The start x grid location, of the wire which runs parallel to the x-axis from startx and ends at endx, inclusive. Value can be relative to the device size.
+    :req_param endx: The end of x grid location of the wire. Value can be relative to the device size.
+    :req_param y: The location of the rib with respect to the y-axis. Value can be relative to the device size.
+    :opt_param repeatx: The horizontal repeat factor of the rib along the device. Value can be relative to the device size.
+    :opt_param repeaty: The vertical repeat factor of the rib along the device. Value can be relative to the device size.
+
+Along each spine and rib is a group of switch points. Switch points are used to describe drive or tap locations along the clock distribution wire:
+
+.. arch:tag:: <switch_point type="{drive | tap}" name="string" yoffset="integer" xoffset="integer" xinc="integer" yinc="integer" buffer="string">
```
kmurray: I think we'll need diagrams to illustrate this, other wise it's likely unclear what a switchpoint is.

If the earlier rib-spine diagrams aren't too busy we could annotate the switchpoints on them. However I expect it may be better to create new figures using the same illustrative base diagrams with switchpoint annotations instead.

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+
+        <clock_routing>
+            <!-- connections from inter-block routing to central spine -->
+            <tap from="ROUTING" to="spine1.drive" locationx="W/2" locationy="H/2" switch="general_routing_switch" fc_val="1.0"/>
+            <!-- connections from spine to rib -->
+            <tap from="spine1.tap" to="rib1.drive" switch="general_routing_switch" fc_val="0.5"/>
+            <!-- connections from rib to clock pins -->
+            <tap from="rib1.tap" to="CLOCK" switch="ipin_cblock" fc_val="1.0"/>
+        </clock_routing >
+    </clocknetworks >
```
kmurray: I like the example, but I think we'll need a diagram illustrating what this clock network looks like. Otherwise only people very familiar with how this is implemented in VPR will be able to follow.

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+    :opt_param yinc: (Only for spine ``tap`` points) Descibes the repeat factor of a series of evenly spaced tap points.
+    :req_param buffer: (Required only for ``drive`` points) A reference to a pre-defined routing switch. This switch will be used at the drive point. The clock generator uses two buffers to drive the two portions of this ``clock_network`` wire when it is split at the drive point.
+
+Lastly the ``<clock_routing>`` element consists of a group of ``tap`` elements that each separately describe the connectivity from one routing resource (pin or wire) to another. The tap element and its attribute sare as follows:
+
+.. arch:tag:: <tap from="string" to="string" locationx="integer" locationy="integer" switch="string" fc_val="float">
+    :req_param from: The routing resource to make a connection from. The referecing schema can be seen in the clock network exambple where the unique clock network name is forllowe by a period and then the switch name. In addition a special literal ``"ROUTING"`` can be used to reference a connection from the general inter-block routing.
```
kmurray: Probably want to expand this a bit more formally illustrating the two cases:
 * `clock_name.tap_name` format, and
 * `ROUTING` special value

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+    :req_param to: The routing resource to make a connection to. A special literal ``"CLOCK"`` describes connections from clock network tap points that supply the clock to clock pins on blocks at the tap locations; these clock inputs are already  specified on blocks in the VTR architecture file.
```
kmurray: Similarly here

doc/src/arch/reference.rst
kmurray: Also small typo: `forllowe` -> `followed`

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
...
+    :req_param locationx: (Only when using the special literal "ROUTING") The x grid location of inter-block routing.
+    :req_param locationy: (Only when using the special literal "ROUTING") The y grid location of inter-block routing.
```
kmurray: Probably need a bit more detail about how this works.

What does it do at the specific X/Y location? Does it pick all wires passing that location (i.e. both CHANX and CHANY)? Does it only connect to wires starting/ending at that X/Y, or wires which have switchpoints, or any wire?

doc/src/arch/reference.rst
```diff
@@ -1897,6 +1902,98 @@ The clocking configuration is specified with ``<clock>`` tags within the ``<cloc
     :opt_param C_wire_per_m: The wire capacitance, in Farads per Meter.
     :opt_param buffer_size: The size of each clock buffer.
 
+**Specifing a Clock Architecture**
```
litghost: This doesn't talk about the virtual clock sink, and how that is placed within the grid?

HackerFoo: My understanding is that it's specified in the `clocknetworks/clock_routing/tap[@from="ROUTING"]` element. See `setup_clock_connections()` in `setup_clocks.cpp`.

HackerFoo: Note that the current code seems to only support one virtual clock network root, with the index stored in the `DeviceContext`.

doc/src/arch/reference.rst
HackerFoo: The diagram on page 38 of [Mustafa's thesis](https://tspace.library.utoronto.ca/bitstream/1807/97807/3/Abbas_Mustafa_S_201911_MSc_thesis.pdf) shows why there can only be two buffers with the spine/rib model.

HackerFoo:
@mustafabbas It might be good to also include your thesis somewhere, or at least a reference.

kmurray:
Thanks @mustafabbas, those figures look great!

I've fixed up some of the formatting and clarified some of the descriptions. Once CI passes this should be good to merge.

vaughnbetz:
Thanks a lot Mustafa!

Issue #479, opened 2019-03-05
High fanout routing logic causes routing failure


litghost:
#### Expected Behaviour

Router should route all nets.

#### Current Behaviour

Outputs the warning:
Warning 815554: No routing path found in high-fanout mode for net connection (to sink_rr 125107), retrying with full route tree

and eventually fails to route after a while.

#### Possible Solution

Disabling the high fanout mode with "--router_high_fanout_threshold -1" resolves the issue in the short term.

#### Steps to Reproduce

1. Build VPR from https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/478
2. Unzip [failing_test.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/2933647/failing_test.zip) to a directory
3. Run:

```
/usr/local/google/home/keithrothman/cat_x/vtr-verilog-to-routing/vpr/vpr arch.unique_pack.xml top.eblif --device xc7a50t-test --read_rr_graph rr_graph_xc7a50t_test.rr_graph.real.xml --min_route_chan_width_hint 100 --max_criticality 0.0 --max_router_iterations 500 --routing_failure_predictor off --constant_net_method route --route_chan_width 500 --clock_modeling route --place_algorithm bounding_box --enable_timing_computations off --allow_unrelated_clustering on --route
```

and it will fail.

run:

```
/usr/local/google/home/keithrothman/cat_x/vtr-verilog-to-routing/vpr/vpr arch.unique_pack.xml top.eblif --device xc7a50t-test --read_rr_graph rr_graph_xc7a50t_test.rr_graph.real.xml --min_route_chan_width_hint 100 --max_criticality 0.0 --max_router_iterations 500 --routing_failure_predictor off --constant_net_method route --route_chan_width 500 --clock_modeling route --place_algorithm bounding_box --enable_timing_computations off --allow_unrelated_clustering on --route --router_high_fanout_threshold -1
```
and routing will succeed.

litghost:
Failing router output:
```
---- ------ ------- ---- ------- ------- ------- ----------------- --------------- -------- ---------- ---------- ---------- ---------- --------
Iter   Time    pres  BBs    Heap  Re-Rtd  Re-Rtd Overused RR Nodes      Wirelength      CPD       sTNS       sWNS       hTNS       hWNS Est Succ
[...]

PR #1113, opened 2020-02-06
Prevent non-termination with high fanout routing logic

HackerFoo:
#### Description
This change uses the routing predictor to detect when congestion resolution slows down, and disables the high fanout routing logic.

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/479

#### Motivation and Context
High fanout routing logic has been disabled in Symbiflow due to non-termination in some tests. This fixes that.

This change has been tested in Symbiflow, and also strong and nightly tests have been run locally.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -1986,9 +1986,10 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
 
     route_timing_grp.add_argument<float>(args.router_high_fanout_max_slope, "--router_high_fanout_max_slope")
         .help(
-            "Maximum routing predictor slope where high fanout routing is enabled.\n"
-            " -1.0 is normal progress, 0 is no progress.")
-        .default_value("-0.5")
+            "Minimum routing progress where high fanout routing is enabled."
+            " This is a ratio of the actual congestion reduction to what is expected based in the history.\n"
+            " 1.0 is normal progress, 0 is no progress.")
+        .default_value("0.1")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
     route_timing_grp.add_argument<e_router_lookahead, ParseRouterLookahead>(args.router_lookahead_type, "--router_lookahead")
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -1984,6 +1984,13 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("64")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument<float>(args.router_high_fanout_max_slope, "--router_high_fanout_max_slope")
+        .help(
+            "Maximum routing predictor slope where high fanout routing is enabled.\n"
+            " -1.0 is normal progress, 0 is no progress.")
+        .default_value("-0.5")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
     route_timing_grp.add_argument<e_router_lookahead, ParseRouterLookahead>(args.router_lookahead_type, "--router_lookahead")
         .help(
             "Controls what lookahead the router uses to calculate cost of completing a connection.\n"
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -1223,7 +1223,7 @@ static bool timing_driven_route_sink(
     //We normally route high fanout nets by only adding spatially close-by routing to the heap (reduces run-time).
     //However, if the current sink is 'critical' from a timing perspective, we put the entire route tree back onto
     //the heap to ensure it has more flexibility to find the best path.
-    if (high_fanout && !sink_critical && !net_is_global && !net_is_clock && !(routing_predictor.get_slope() > router_opts.high_fanout_max_slope)) {
+    if (high_fanout && !sink_critical && !net_is_global && !net_is_clock && -routing_predictor.get_slope() > router_opts.high_fanout_max_slope) {
         std::tie(found_path, cheapest) = router.timing_driven_route_connection_from_route_tree_high_fanout(rt_root,
                                                                                                            sink_node,
                                                                                                            cost_params,
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -82,7 +82,8 @@ static bool timing_driven_route_sink(
     t_rt_node* rt_root,
     t_rt_node** rt_node_of_sink,
     SpatialRouteTreeLookup& spatial_rt_lookup,
-    RouterStats& router_stats);
+    RouterStats& router_stats,
+    const RoutingPredictor& routing_predictor);
 
 template<typename ConnectionRouter>
 static bool timing_driven_pre_route_to_clock_root(
@@ -414,7 +415,8 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
                                                            route_timing_info,
                                                            pin_timing_invalidator.get(),
                                                            budgeting_inf,
-                                                           was_rerouted);
+                                                           was_rerouted,
+                                                           routing_predictor);
             if (!is_routable) {
                 return (false); //Impossible to route
             }
@@ -756,7 +758,8 @@ bool try_timing_driven_route_net(ConnectionRouter& router,
                                  std::shared_ptr<SetupTimingInfo> timing_info,
                                  ClusteredPinTimingInvalidator* pin_timing_invalidator,
                                  route_budgets& budgeting_inf,
-                                 bool& was_rerouted) {
+                                 bool& was_rerouted,
+                                 const RoutingPredictor& routing_predictor) {
     auto& cluster_ctx = g_vpr_ctx.clustering();
     auto& route_ctx = g_vpr_ctx.mutable_routing();
 
@@ -787,7 +790,8 @@ bool try_timing_driven_route_net(ConnectionRouter& router,
                                             netlist_pin_lookup,
                                             timing_info,
                                             pin_timing_invalidator,
-                                            budgeting_inf);
+                                            budgeting_inf,
+                                            routing_predictor);
 
         profiling::net_fanout_end(cluster_ctx.clb_nlist.net_sinks(net_id).size());
 
@@ -911,7 +915,8 @@ bool timing_driven_route_net(ConnectionRouter& router,
                              const ClusteredPinAtomPinsLookup& netlist_pin_lookup,
                              std::shared_ptr<SetupTimingInfo> timing_info,
                              ClusteredPinTimingInvalidator* pin_timing_invalidator,
-                             route_budgets& budgeting_inf) {
+                             route_budgets& budgeting_inf,
+                             const RoutingPredictor& routing_predictor) {
     /* Returns true as long as found some way to hook up this net, even if that *
      * way resulted in overuse of resources (congestion).  If there is no way   *
      * to route this net, even ignoring congestion, it returns false.  In this  *
@@ -1047,7 +1052,8 @@ bool timing_driven_route_net(ConnectionRouter& router,
                                       router_opts.high_fanout_threshold,
                                       rt_root, rt_node_of_sink,
                                       spatial_route_tree_lookup,
-                                      router_stats))
+                                      router_stats,
+                                      routing_predictor))
             return false;
 
         profiling::conn_finish(route_ctx.net_rr_terminals[net_id][0],
@@ -1187,7 +1193,8 @@ static bool timing_driven_route_sink(
     t_rt_node* rt_root,
     t_rt_node** rt_node_of_sink,
     SpatialRouteTreeLookup& spatial_rt_lookup,
-    RouterStats& router_stats) {
+    RouterStats& router_stats,
+    const RoutingPredictor& routing_predictor) {
     /* Build a path from the existing route tree rooted at rt_root to the target_node
      * add this branch to the existing route tree and update pathfinder costs and rr_node_route_inf to reflect this */
     auto& route_ctx = g_vpr_ctx.mutable_routing();
@@ -1216,7 +1223,7 @@ static bool timing_driven_route_sink(
     //We normally route high fanout nets by only adding spatially close-by routing to the heap (reduces run-time).
     //However, if the current sink is 'critical' from a timing perspective, we put the entire route tree back onto
     //the heap to ensure it has more flexibility to find the best path.
-    if (high_fanout && !sink_critical && !net_is_global && !net_is_clock) {
+    if (high_fanout && !sink_critical && !net_is_global && !net_is_clock && !(routing_predictor.get_slope() > -0.5)) {
         std::tie(found_path, cheapest) = router.timing_driven_route_connection_from_route_tree_high_fanout(rt_root,
                                                                                                            sink_node,
                                                                                                            cost_params,
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -78,7 +78,7 @@ static bool timing_driven_route_sink(
     unsigned itarget,
     int target_pin,
     const t_conn_cost_params cost_params,
-    int high_fanout_threshold,
+    const t_router_opts& router_opts,
     t_rt_node* rt_root,
     t_rt_node** rt_node_of_sink,
     SpatialRouteTreeLookup& spatial_rt_lookup,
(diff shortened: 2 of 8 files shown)


HackerFoo:
A number of nightly tests failed on master, so I'll have to compare to make sure there are no new failures.

HackerFoo:
I got the same failures on master, so this should be okay.

vpr/src/route/routing_predictor.cpp
```diff
@@ -63,12 +63,8 @@ float covariance(std::vector<size_t> x_values, std::vector<float> y_values, floa
     return cov;
 }
 
-float RoutingPredictor::get_slope() {
-    if (iterations_.size() > min_history_) {
-        auto model = fit_model(iterations_, iteration_overused_rr_node_counts_, history_factor_);
-        return model.get_slope();
-    }
-    return -1;
+float RoutingPredictor::get_slope() const {
```
vaughnbetz: Should comment the caching of slope (you mentioned it n the meeting).

HackerFoo: Added comment.

vpr/src/base/read_options.cpp
```diff
@@ -1591,6 +1591,13 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("64")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument<float>(args.router_high_fanout_max_slope, "--router_high_fanout_max_slope")
+        .help(
+            "Maximum routing predictor slope where high fanout routing is enabled.\n"
+            " -1.0 is normal progress, 0 is no progress.")
+        .default_value("-0.5")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
```
kmurray: I think this description is fairly opaque and needs some elaboration.

What is the slope being considered? What are its units? Why is -1 normal progress, and the default set to -0.5? What value would disable this feature (e.g. may want to disable it during minW search if the run-time hit is high)?

Ideally we want a tunable which is somewhat intuitive for those who don't know the internals of the routing predictor.

Perhaps something like an expected routing convergence iteration threshold would be more intuitive. For example, disable HF routing if: `predicted_success_iteration / max_router_iterations > alpha` (where alpha is the parameter this option controls).

vpr/src/route/route_timing.cpp
```diff
@@ -1230,14 +1237,14 @@ static bool timing_driven_route_sink(ClusterNetId net_id,
...
-    bool high_fanout = is_high_fanout(cluster_ctx.clb_nlist.net_sinks(net_id).size(), high_fanout_threshold);
+    bool high_fanout = is_high_fanout(cluster_ctx.clb_nlist.net_sinks(net_id).size(), router_opts.high_fanout_threshold);
     constexpr float HIGH_FANOUT_CRITICALITY_THRESHOLD = 0.9;
     bool sink_critical = (cost_params.criticality > HIGH_FANOUT_CRITICALITY_THRESHOLD);
 
     //We normally route high fanout nets by only adding spatially close-by routing to the heap (reduces run-time).
     //However, if the current sink is 'critical' from a timing perspective, we put the entire route tree back onto
     //the heap to ensure it has more flexibility to find the best path.
-    if (high_fanout && !sink_critical && !net_is_global) {
+    if (high_fanout && !sink_critical && !net_is_global && !(routing_predictor.get_slope() > router_opts.high_fanout_max_slope)) {
```
kmurray: As this logic gets more complicated (and it depends on more things like the RoutingPredictor), I wonder if we should hoist it out of timing_driven_route_sink and simply have a parameter which controls whether we do HF routing or not. I'm not sure which is better, so any thoughts are welcome.

HackerFoo:
| parse_results_1113.txt
-- | --
vtr_flow_elapsed_time | 1.05
odin_synth_time | 1.03
abc_depth | 1.00
abc_synth_time | 1.01
num_clb | 1.00
num_memories | 1.00
num_mult | 1.00
max_vpr_mem | 0.95
num_pre_packed_blocks | 1.00
num_post_packed_blocks | 1.00
device_grid_tiles | 1.00
pack_time | 1.02
placed_wirelength_est | 1.00
place_time | 1.03
placed_CPD_est | 1.00
min_chan_width | 1.01
routed_wirelength | 1.00
min_chan_width_route_time | 1.04
crit_path_routed_wirelength | 1.00
critical_path_delay | 1.00
crit_path_route_time | 1.03

HackerFoo:
Eh, I should probably look closer. `min_chan_width_route_time` for `LU32PEEng.v` is ~3x, 2986.94s vs 981.12s

parse_results_1113.txt |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
arch | circuit | vtr_flow_elapsed_time | odin_synth_time | abc_depth | abc_synth_time | num_clb | num_memories | num_mult | max_vpr_mem | num_pre_packed_blocks | num_post_packed_blocks | device_grid_tiles | pack_time | placed_wirelength_est | place_time | placed_CPD_est | min_chan_width | routed_wirelength | min_chan_width_route_time | crit_path_routed_wirelength | critical_path_delay | crit_path_route_time
k6_frac_N10_frac_chain_mem32K_40nm.xml | arm_core.v | 0.95 | 0.97 | 1.00 | 1.00 | 1.00 | 1.00 |   | 1.13 | 1.00 | 1.00 | 1.00 | 0.99 | 1.00 | 1.02 | 1.00 | 1.02 | 0.99 | 0.89 | 1.00 | 1.00 | 1.04
k6_frac_N10_frac_chain_mem32K_40nm.xml | bgm.v | 1.01 | 1.14 | 1.00 | 0.98 | 1.00 |   | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.03 | 1.00 | 1.02 | 1.00 | 1.00 | 1.00 | 1.01 | 1.00 | 1.00 | 0.99
k6_frac_N10_frac_chain_mem32K_40nm.xml | blob_merge.v | 1.00 | 0.94 | 1.00 | 0.99 | 1.00 |   |   | 0.99 | 1.00 | 1.00 | 1.00 | 0.97 | 1.00 | 0.95 | 1.00 | 1.00 | 1.00 | 1.09 | 1.00 | 1.00 | 1.01
k6_frac_N10_frac_chain_mem32K_40nm.xml | boundtop.v | 1.01 | 1.00 | 1.00 | 1.04 | 1.00 |   |   | 0.73 | 1.00 | 1.00 | 1.00 | 1.01 | 1.00 | 1.03 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00
k6_frac_N10_frac_chain_mem32K_40nm.xml | ch_intrinsics.v | 0.99 | 1.00 | 1.00 | 0.87 | 1.00 | 1.00 |   | 0.73 | 1.00 | 1.00 | 1.00 | 0.94 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00
k6_frac_N10_frac_chain_mem32K_40nm.xml | diffeq1.v | 1.00 | 1.00 | 1.00 | 0.95 | 1.00 |   | 1.00 | 0.74 | 1.00 | 1.00 | 1.00 | 0.98 | 1.00 | 1.02 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.06
k6_frac_N10_frac_chain_mem32K_40nm.xml | diffeq2.v | 0.97 | 1.00 | 1.00 | 0.86 | 1.00 |   | 1.00 | 0.94 | 1.00 | 1.00 | 1.00 | 1.02 | 1.00 | 1.01 | 1.00 | 1.00 | 1.00 | 0.96 | 1.00 | 1.00 | 1.02
k6_frac_N10_frac_chain_mem32K_40nm.xml | LU8PEEng.v | 1.09 | 0.87 | 1.00 | 1.01 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.03 | 1.00 | 1.03 | 1.00 | 1.00 | 1.00 | 1.21 | 1.00 | 1.00 | 1.08
k6_frac_N10_frac_chain_mem32K_40nm.xml | LU32PEEng.v | 1.77 | 1.05 | 1.00 | 1.03 | 1.00 | 1.00 | 1.00 | 0.89 | 1.00 | 1.00 | 1.00 | 1.04 | 1.00 | 1.04 | 1.00 | 0.98 | 1.01 | 3.04 | 1.00 | 1.00 | 1.03
k6_frac_N10_frac_chain_mem32K_40nm.xml | mcml.v | 0.94 | 1.02 | 1.00 | 1.10 | 1.00 | 1.00 | 1.00 | 0.99 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.05 | 1.00 | 1.14 | 0.98 | 0.39 | 0.99 | 1.00 | 1.03
k6_frac_N10_frac_chain_mem32K_40nm.xml | mkDelayWorker32B.v | 1.03 | 1.04 | 1.00 | 1.01 | 1.00 | 1.00 |   | 0.98 | 1.00 | 1.00 | 1.00 | 1.02 | 1.00 | 1.02 | 1.00 | 1.00 | 1.00 | 1.05 | 1.00 | 1.00 | 1.06
k6_frac_N10_frac_chain_mem32K_40nm.xml | mkPktMerge.v | 1.01 | 1.00 | 1.00 | 1.14 | 1.00 | 1.00 |   | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 0.98 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.07
k6_frac_N10_frac_chain_mem32K_40nm.xml | mkSMAdapter4B.v | 1.08 | 1.19 | 1.00 | 1.04 | 1.00 | 1.00 |   | 0.98 | 1.00 | 1.00 | 1.00 | 1.04 | 1.00 | 1.04 | 1.00 | 1.00 | 1.00 | 1.13 | 1.00 | 1.00 | 1.00
k6_frac_N10_frac_chain_mem32K_40nm.xml | or1200.v | 1.04 | 1.15 | 1.00 | 1.06 | 1.00 | 1.00 | 1.00 | 0.99 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.02 | 1.00 | 1.00 | 1.00 | 1.08 | 1.00 | 1.00 | 1.20
k6_frac_N10_frac_chain_mem32K_40nm.xml | raygentop.v | 0.98 | 0.97 | 1.00 | 1.04 | 1.00 |   | 1.00 | 0.98 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.03 | 1.00 | 0.97 | 1.05 | 0.88 | 1.05 | 1.00 | 1.25
k6_frac_N10_frac_chain_mem32K_40nm.xml | sha.v | 1.03 | 1.05 | 1.00 | 1.02 | 1.00 |   |   | 1.00 | 1.00 | 1.00 | 1.00 | 1.07 | 1.00 | 1.05 | 1.00 | 1.00 | 1.00 | 1.03 | 1.00 | 1.00 | 1.03
k6_frac_N10_frac_chain_mem32K_40nm.xml | spree.v | 1.06 | 1.17 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.05 | 1.00 | 1.07 | 1.00 | 1.00 | 1.00 | 1.06 | 1.00 | 1.00 | 1.03
k6_frac_N10_frac_chain_mem32K_40nm.xml | stereovision0.v | 1.11 | 1.06 | 1.00 | 1.06 | 1.00 |   |   | 1.00 | 1.00 | 1.00 | 1.00 | 1.07 | 1.00 | 1.10 | 1.00 | 1.00 | 1.00 | 1.24 | 1.00 | 1.00 | 1.07
k6_frac_N10_frac_chain_mem32K_40nm.xml | stereovision1.v | 1.07 | 1.05 | 1.00 | 1.01 | 1.00 |   | 1.00 | 1.06 | 1.00 | 1.00 | 1.00 | 1.04 | 1.00 | 1.05 | 1.00 | 1.05 | 0.93 | 1.15 | 1.00 | 0.98 | 0.97
k6_frac_N10_frac_chain_mem32K_40nm.xml | stereovision2.v | 1.03 | 1.03 | 1.00 | 1.02 | 1.00 |   | 1.00 | 0.99 | 1.00 | 1.00 | 1.00 | 1.04 | 1.00 | 1.02 | 1.00 | 1.00 | 1.00 | 1.02 | 1.00 | 1.00 | 1.08
k6_frac_N10_frac_chain_mem32K_40nm.xml | stereovision3.v | 1.06 | 1.00 | 1.00 | 1.00 | 1.00 |   |   | 0.97 | 1.00 | 1.00 | 1.00 | 1.18 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 0.75
#NUM! | #NUM! | 1.05 | 1.03 | 1.00 | 1.01 | 1.00 | 1.00 | 1.00 | 0.95 | 1.00 | 1.00 | 1.00 | 1.02 | 1.00 | 1.03 | 1.00 | 1.01 | 1.00 | 1.04 | 1.00 | 1.00 | 1.03

kmurray:
Thinking about this some more, there may be another approach to consider.

As proposed this PR disable the HF routing optimization for all nets when congestion is detected. In particular this could have a notable CPU time impact if only a few nets were running into issues (since the HF routing optimization would be disabled on **all** nets).

Another idea would be to detect which nets (or ideally net connections) were being routed with the HF optimization, but were repeatedly becoming congested and preventing convergence. The HF routing optimization could then be disabled _only_ for those connections. That would ensure that those connections which route fine with the HF optimization continue to do so, while those for which it is problematic would avoid it.

HackerFoo:
I think selectively disabling the optimization could be worth considering if a good threshold can't be found, but it's much more complicated. This flag allows the HF logic to be used at all for XC7 where it is currently disabled.
I'm collecting data to see how much re-enabling this helps.

vpr/src/route/route_timing.cpp
HackerFoo: I prefer moving in the other direction, and letting optimized methods decide if they apply, and then several methods can be tried in order of priority.

HackerFoo:
This flag still has little effect on Titan tests:

Geomeans for `titan_quick_qor` tests:
 flag setting | crit_path_route_time | routed_wirelength | critical_path_delay | vtr_flow_elapsed_time
 -- | -- | -- | -- | --
0.1 | 1.002046 | 1.000000 | 1.000007 | 1.012878
0.2 | 1.040168 | 1.000000 | 1.000007 | 1.093631
0.5 | 1.013354 | 1.000112 | 1.000000 | 1.012346
not set | 1.000000 | 1.000000 | 1.000007 | 1.000000

vpr/src/base/read_options.cpp
HackerFoo: I've updated the description.

HackerFoo:
@litghost  I did a parameter sweep on Ibex for the Arty, and it showed that regardless of the slope threshold, re-enabling the high fanout logic doubles routing time: https://colab.research.google.com/drive/1wNF5lNLXQIVx5djih4qwGgBzzCOGJG2c

So I chose the most conservative slope threshold of 0.1.

The optimal fanout threshold seems to be 128 for that particular design.

litghost:
> @litghost I did a parameter sweep on Ibex for the Arty, and it showed that regardless of the slope threshold, re-enabling the high fanout logic triples routing time: https://colab.research.google.com/drive/1wNF5lNLXQIVx5djih4qwGgBzzCOGJG2c
> 
> So I chose the most conservative slope threshold of 0.1.
> 
> The optimal fanout threshold seems to be 128 for that particular design.

Is it possible to add final CPD to that table?  Anyways, it looks like from a router runtime perspective, high fanout routing logic is always bad.  If there is no CPD benefit, it seems like we should continue to disable that code pathway, even with #1113 merged?

vaughnbetz:
Good to add cpd as a generally good thing to track, but as the router high-fanout code is space pruning code to reduce CPU it's unlikely to give a better cpd.  So it definitely does seem like it is not useful for Arty.

litghost:
> Good to add cpd as a generally good thing to track, but as the router high-fanout code is space pruning code to reduce CPU it's unlikely to give a better cpd. So it definitely does seem like it is not useful for Arty.

I suspect there is some underlying issue present here that is make the high-fanout code less useful, but that is a much less obvious analysis than "does it converge at all", which was the previous issue.

I think this code is still worth merging as it breaks a quasi-infinite loop, but it does seem we should leave the path disabled for now.

HackerFoo:
`fpga-tool-perf` reports the clock frequency rather than CPD, so I've added that to the Colab. The maximum clock frequency was achieved with the high-fanout logic disabled.

PR #1118, opened 2020-02-06
Add vtr_reg_weekly_no_he that drops the high effort Titan tests and gaussian blur

litghost:
#### Description

This PR is attempting to create a weekly that can provide the vast majority of the signal, without the excessive runtime.

#### Motivation and Context

This PR is attempting to do two things:
 - Return QoR / wallclock / memory usage in ~12 - 24 for purposes of quality comparisons

#### Types of changes

#### Checklist:


litghost:
@kmurray Dropping the high effort titan tests and running with 3 cores means that the "weekly" test set returns in ~40 hrs, which isn't bad!  I'm trying with 4 cores to see if:
 - It completes faster
 - Doesn't blow the memory budget

I've added a CPU and memory load log every 5 minutes in this PR, which will hopefully identify why the CI takes "forever" if a large memory regression is introduced.

litghost:
@kmurray : Is there a better way to compose the `vtr_reg_weekly` and `vtr_reg_weekly_no_he` with less duplication?

kmurray:
> Is there a better way to compose the `vtr_reg_weekly` and `vtr_reg_weekly_no_he` with less duplication?

I've considered that, and it's potentially possible but would need some scripting work. 

How I'd approach it would be to extend the `task_list.txt` format to support `#include` like behaviour, similar to the parse/[pass results files](https://docs.verilogtorouting.org/en/latest/vtr/pass_requirements/#file-format). We could the define vtr_reg_weekly_he task_list.txt something like:
```
%include vtr_reg_weekly/task_lists.basic.txt
%include vtr_reg_weekly/task_lists.he.txt
```

This would support selectively not running the HE titan test by using the appropriate task list. Selectively dropping a particular benchmark (e.g. gaussianblur) would be more difficult.

kmurray:
> Is there a better way to compose the `vtr_reg_weekly` and `vtr_reg_weekly_no_he` with less duplication?

I've considered that, and it's potentially possible but would need some scripting work. 

How I'd approach it would be to extend the `task_list.txt` format to support `#include` like behaviour, similar to the parse/[pass results files](https://docs.verilogtorouting.org/en/latest/vtr/pass_requirements/#file-format). We could the define vtr_reg_weekly_he task_list.txt something like:
```
%include vtr_reg_weekly/task_lists.basic.txt
%include vtr_reg_weekly/task_lists.he.txt
```

This would support selectively not running the HE titan test by using the appropriate task list. Selectively dropping a particular benchmark (e.g. gaussianblur) would be more difficult.

litghost:
> This would support selectively not running the HE titan test by using the appropriate task list. Selectively dropping a particular benchmark (e.g. gaussianblur) would be more difficult.

Once we have the include feature, it is straightforward to make 3 tasks.

Do you need the deduplication effort solved right now, or is this PR acceptable?

vaughnbetz:
Discussed in meeting. Don't need this PR, and instead just stop doing weekly as a pre-submit (do as post-merge only).

PR #1123, opened 2020-02-07
Add default effort titan test to weekly regression

kmurray:
#### Description

Adds a default effort level titan23 regression test to the weekly regression tests.

Currently a WIP as regeneration of the golden results is pending.


kmurray:
Runs ~38 hours with 4 cores

Issue #1087, opened 2020-01-28
Overuse of heap in initialization


litghost:
#### Expected Behaviour

Dynamic allocation should be done deliberately, and avoided when possible.  1 allocation is better than N allocations, and N allocations is better than N*N allocations.

#### Current Behaviour

Dynamic allocation is being used in several places that is simply generating garbage.  Concrete examples:

- [my_atof_2D](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/libvtrutil/src/vtr_token.cpp#L151) duplicates a const char *, simply to delimit a substring and call `vtr::atoi` on it.  Solution: have `vtr::atoi` use a `string_view`.
- [load_delay_annotations](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/pack/pb_type_graph_annotations.cpp#L263-L267) allocates a bunch of arrays, including a 2D array with 1+N arrays, and then simply frees them.
- [load_delay_annotations](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/pack/pb_type_graph_annotations.cpp#L315-L321) uses a std::set to check for pin usage, which generates 1 allocation per output pin.  A sorted vector would only require 1 allocation, and be faster!

As a concrete example, the function `load_delay_annotations` (and its descendants) generates around 1/3 of all allocation requests during `SetupVPR`.  Number of allocations done within `load_delay_annotations` is in excess of 21 million when loading the `stratixiv_arch.timing.xml` architecture.

To be clear, when not running with the heap profiler, this doesn't take too much time:
```
# Loading Architecture Description
# Loading Architecture Description took 0.80 seconds (max_rss 78.4 MiB, delta_rss +64.0 MiB)
# Building complex block graph
# Building complex block graph took 7.55 seconds (max_rss 620.5 MiB, delta_rss +542.1 MiB)
```

but given that the vast majority of those allocations are simply unneeded.

#### Possible Solution

[...]

PR #1126, opened 2020-02-07
Memory and CPU refactoring for rr graph build 

litghost:
#### Description

This patch set was written with the following goals:
 - Make allocations patterns more heap friendly, resulting in lower max RSS
 - Leave CPU runtime around the same or better
 - Enable future refactoring around the router

#### Related Issue

#1079 
#1087

#### Motivation and Context

Baseline for comparison was 6ea11c1757e1e43a1770c36608907c943cf814a2

QoR:
[opt6_results.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4172348/opt6_results.zip)
[baseline_results.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4172349/baseline_results.zip)
[titan_comp.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4172350/titan_comp.xlsx)

Top level ratios (for Titan, minus gaussian blur):
 - 40% reduction in max RSS
 - 4% reduction in critical path route time
 - 1% reduction in overall VPR flow time

- [x] CI is green
- [x] Titan weekly (minus gaussion blur) QoR gathered

#### Types of changes

#### Checklist:


litghost:
QoR:
[opt6_results.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4172348/opt6_results.zip)
[baseline_results.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4172349/baseline_results.zip)
[titan_comp.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4172350/titan_comp.xlsx)

litghost:
Top level ratios (for Titan):
 - 40% reduction in max RSS
 - 4% reduction in critical path route time
 - 1% reduction in overall VPR flow time

litghost:
Baseline for comparison was 6ea11c1757e1e43a1770c36608907c943cf814a2

litghost:
@kmurray FYI

vpr/src/route/rr_graph_storage.h
```diff
@@ -216,6 +216,10 @@ class t_rr_graph_storage {
     short node_track_num(RRNodeId id) const; //Same as ptc_num() but checks that type() is consistent
     short node_class_num(RRNodeId id) const; //Same as ptc_num() but checks that type() is consistent
 
+    inline void prefetch_node(RRNodeId id) const {
```
vaughnbetz: Should comment what these prefetchs are.

vpr/src/route/route_timing.cpp
```diff
@@ -1960,6 +1960,16 @@ void Router::timing_driven_expand_neighbours(t_heap* current,
     RREdgeId first_edge = rr_nodes_->first_edge(from_node);
     RREdgeId last_edge = rr_nodes_->last_edge(from_node);
     int num_edges = size_t(last_edge) - size_t(first_edge);
+    for (int iconn = 0; iconn < num_edges; iconn++) {
+        RREdgeId from_edge(size_t(first_edge) + iconn);
+
+        RRNodeId to_node = rr_nodes_->edge_sink_node(from_edge);
```
vaughnbetz: Explain why you're prefetching (to_node and rr_switch_inf will be used soon) and what speed up you got.

vpr/src/route/route_timing.cpp
```diff
@@ -1960,6 +1960,16 @@ void Router::timing_driven_expand_neighbours(t_heap* current,
...
+        rr_nodes_->prefetch_node(to_node);
+
+        int switch_idx = rr_nodes_->edge_switch(from_edge);
+        __builtin_prefetch(&rr_switch_inf_[switch_idx], 0, 0);
+    }
+
```
HackerFoo: Is it worth wrapping this in an `#ifdef` since it uses an intrinsic that may not perform the same or be present on all CPU architectures?

vaughnbetz: Good question. Will this intrinsic compile on a non-x86 architeture?  OK if it is ignored, but not good to have a compiler error.  If it won't compile on other platforms, it should be wrapped in an #ifdef.

litghost: > Will this intrinsic compile on a non-x86 architeture?

Yes

vaughnbetz: Thanks.  Follow-up: will it compile in compilers other than g++ (e.g. if you used IBM's compiler to target Power8, would this line error out)?

litghost: Both GCC and clang have support for `__builtin_prefetch`, other targets will vary.

kmurray: If we're going to put stuff like this in (which I'm not sure is worth it), I'd like to see it wrapped up in a function (e.g. something like `vtr::prefetch()`). That way we can #ifdef it to a no-op in one location on platforms which don't support it.

litghost: > which I'm not sure is worth it

What's your claim here?  I have evidence that it is worthwhile, so I'm not sure what your statement means.

litghost: > I'd like to see it wrapped up in a function (e.g. something like `vtr::prefetch()`).

Done, in a macro (because it has to be as far as I can tell).

kmurray: > What's your claim here? I have evidence that it is worthwhile, so I'm not sure what your statement means.

To me there is a tension here, between code generality/maintainability and run-time performance.

My concern is that including these types of actions (e.g. SW prefetching) are generally pretty fragile, mysterious, and tend not to age well. The motivation/reasoning/measurements of why they were added fade overtime, it becomes unclear what assumptions make it necessary, and is unclear when changes in those assumptions (which could occur far away from this code) would invalidate using it (ie when it should be removed). Another concern is that these types of optimizations may be over-fitting the code base to the specific platform on which they are developed/tested.

Now, if the performance benefits are substantial and the above considerations are well accounted for and documented then it may be acceptable, but it depends.

litghost: > Now, if the performance benefits are substantial and the above considerations are well accounted for and documented then it may be acceptable, but it depends.

I was showing around a 6-8% wallclock runtime difference on directrf and gsm that was repeatable and consistent.  It also makes sense given the access pattern of the core expansion loop.

litghost: > My concern is that including these types of actions (e.g. SW prefetching) are generally pretty fragile, mysterious, and tend not to age well.

The ideal case would be to write the inner most expansion loop such that the hardware prefetcher can handle the access pattern.  I tried 3-4 refactoring to achieve similiar performance to the simple prefetch loop, and the prefetch loop outperformed all of them.

libs/libvtrutil/src/vtr_memory.h
```diff
@@ -46,6 +46,11 @@ void chunk_delete(T* obj, t_chunk* /*chunk_info*/) {
 //Cross platform wrapper around GNU's malloc_trim()
 // TODO: This is only used in one place within VPR, consider removing it
 int malloc_trim(size_t pad);
+
+// This is a macro because it has to be.  rw and locality must be constants,
```
vaughnbetz: Thanks for wrapping in a macro.  I agree with Kevin it's important to minimize and modularize code like this that may not port or age well.

I'd also like a comment describing what it does.  Calls the x86 prefetch instruction to preload data into the cache, which should only be done for performance critical code where it shows a gain.  If in doubt, leave it to the HW prefetcher.

litghost: Done

vpr/src/route/route_timing.h
```diff
@@ -10,9 +10,131 @@
 #include "route_common.h"
 #include "router_stats.h"
 #include "router_lookahead.h"
+#include "spatial_route_tree_lookup.h"
 
 int get_max_pins_per_net();
 
+class Router {
```
litghost: rename to `ConnectionRouter`
Add description

vpr/src/route/rr_node_impl.h
```diff
@@ -0,0 +1,166 @@
+#ifndef _RR_NODE_IMPL_H_
+#define _RR_NODE_IMPL_H_
+
+#include "rr_node.h"
```
litghost: Who should be reading this, how to update, and when to update?

vpr/src/route/rr_node.h
```diff
@@ -1,96 +1,34 @@
...
- *                                                                           *
- * direction: if the node represents a track, this field                     *
- *            indicates the direction of the track. Otherwise                *
- *            the value contained in the field should be                     *
- *            ignored.                                                       *
- * side: The side of a grid location where an IPIN or OPIN is located.       *
- *       This field is valid only for IPINs and OPINs and should be ignored  *
- *       otherwise.                                                          */
 
 class t_rr_node {
```
litghost: This is a proxy

litghost: Done

vpr/src/route/rr_node_impl.h
litghost: done

vpr/src/route/route_timing.h
litghost: done

litghost:
@vaughnbetz / @kmurray I've added an initial set of comments, please re-review.

vpr/src/route/rr_graph_storage.h
```diff
@@ -72,6 +75,11 @@ struct t_rr_node_ptc_data {
 static_assert(sizeof(t_rr_node_data) == 16, "Check t_rr_node_data size");
 static_assert(alignof(t_rr_node_data) == 16, "Check t_rr_node_data size");
 
+// aligned_allocator is a STL allocator that allocates memory in an aligned
+// fashion (if supported by the platform).
+//
+// It is worth noting the C++20 std::allocator does aligned allocations, but
+// C++20 has poor support.
 template<class T>
 struct aligned_allocator {
```
kmurray: It seems like this allocator should probably be in vtr_memory.h

vpr/src/route/route_timing.h
```diff
@@ -78,20 +109,34 @@ class ConnectionRouter {
...
     }
 
+    // Common logic from timing_driven_route_connection_from_route_tree and
+    // timing_driven_route_connection_from_route_tree_high_fanout for running
+    // connection router.
     t_heap* timing_driven_route_connection_from_route_tree(
         t_rt_node* rt_root,
         int sink_node,
         const t_conn_cost_params cost_params,
         t_bb bounding_box);
```
kmurray: It looks like the only difference between this and the public method is passing RouterStats. I feel like this should be named differently, as its confusing to overload these two functions.

vpr/src/route/route_timing.h
```diff
@@ -10,9 +10,196 @@
...
+    }
+
+    // Common logic from timing_driven_route_connection_from_route_tree and
+    // timing_driven_route_connection_from_route_tree_high_fanout for running
+    // connection router.
+    t_heap* timing_driven_route_connection_from_route_tree(
+        t_rt_node* rt_root,
+        int sink_node,
+        const t_conn_cost_params cost_params,
+        t_bb bounding_box);
```
kmurray: I think this should have a different name (its currently overloaded with the public method). Perhaps something like `timing_driven_route_connection_common_setup()`?

libs/libvtrutil/src/vtr_memory.h
```diff
@@ -46,6 +47,18 @@ void chunk_delete(T* obj, t_chunk* /*chunk_info*/) {
...
+
+inline int memalign(void** ptr_out, size_t align, size_t size) {
+    return posix_memalign(ptr_out, align, size);
+}
+
+// This is a macro because it has to be.  rw and locality must be constants,
+// not just constexpr.
+//
+// This generates a prefetch instruction on all architectures that include it.
+// This is all modern x86 and ARM64 platforms.
```
vaughnbetz: Comment what rw and locality mean, and recommended value.

vpr/src/route/route_timing.h
litghost: Renamed

vpr/src/route/rr_graph_storage.h
litghost: Yep

vpr/src/route/route_timing.h
litghost: Done

vpr/src/route/rr_graph_storage.h
litghost: Done

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+ *       length 1.  These values are used to decide whether or not this      *
+ *       node should be added to the expansion heap, based on things         *
+ *       like whether it's outside the net bounding box or is moving         *
+ *       further away from the target, etc.                                  *
+ * type:  What is this routing resource?                                     *
+ * cost_index: An integer index into the table of routing resource indexed   *
+ *             data t_rr_index_data (this indirection allows quick dynamic   *
+ *             changes of rr base costs, and some memory storage savings for *
+ *             fields that have only a few distinct values).                 *
+ * capacity:   Capacity of this node (number of routes that can use it).     *
```
vaughnbetz: Add rc_index comment: flyweight index for R and C values for this node.

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+ *           Needed to properly draw.                                        */
+struct t_rr_node_ptc_data {
+    union {
+        int16_t pin_num;
+        int16_t track_num;
+        int16_t class_num;
+    } ptc_;
+};
+
+static_assert(sizeof(t_rr_node_data) == 16, "Check t_rr_node_data size");
```
vaughnbetz: This is a key data structure, so fail an assertion if it gets bigger than expected (16 bytes right now), so developers only expand it after careful consideration.

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+//
+// Within VPR, there are two locations that the rr graph is expected to be
+// modified, either:
+//  - During the building of the rr graph in rr_graph.cpp
+//  - During the reading of a static rr graph from a file in rr_graph_reader
+//  / rr_graph_uxsdcxx_serializer.
+//
+// It is expected and assume that once the graph is completed, the graph is
+// fixed until the entire graph is cleared.  This object enforces this
+// assumption with state flags.  In particular RR graph edges are assumpted
```
vaughnbetz: assumed

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+        edge_dest_node_.clear();
+        edge_switch_.clear();
+        edges_read_ = false;
+        partitioned_ = false;
+        remapped_edges_ = false;
+    }
+
+    // Shrink memory usage of the RR graph storage.
+    //
+    // Note that this will temporary increase the amount of storage required
```
vaughnbetz: temporarily

vpr/src/route/rr_graph_storage.h
litghost: Done

vpr/src/route/rr_graph_storage.h
litghost: Done

vpr/src/route/rr_graph_storage.h
litghost: Fixed

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+
+    // Edge initialization ordering:
+    //  1. Use reserve_edges/emplace_back_edge/alloc_and_load_edges to
+    //     initialize edges.  All edges must be added prior to calling any
+    //     methods that read edge data.
+    //
+    //     Note: Either arch_switch_inf indicies or rr_switch_inf should be
+    //     used with emplace_back_edge and alloc_and_load_edges.  Do not mix
+    //     indicies, otherwise things will be break.
+    //
```
vaughnbetz: Note: rr_switch_inf switches are remapped versions of the arch_switch_inf switch indices that are used when we have different delays and hence different indices based on the fanout of a switch.

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+    //      - edge_id
+    //      - edge_sink_node
+    //      - edge_switch
+
+    /* Edge mutators */
+
+    // Reserve at least num_edges in the edge backing arrays.
+    void reserve_edges(size_t num_edges);
+
+    // Adds ones edge.  This method is efficient if reserve_edges was called with
```
vaughnbetz: one edge

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+    // amortized O(1), like std::vector::emplace_back, but both runtime and
+    // peak memory usage will be higher if reallocation is required.
+    void emplace_back_edge(RRNodeId src, RRNodeId dest, short edge_switch);
+
+    // Adds a batch of edges.
+    void alloc_and_load_edges(const t_rr_edge_info_set* rr_edges_to_create);
+
+    /* Edge finalization methods */
+
+    // Counts the number of rr switches needed based on fan in.
```
vaughnbetz: to support mux size dependent switch delays.

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+    // Get the first and last RREdgeId for the specified RRNodeId.
+    //
+    // The edges belonging to RRNodeId is [first_edge, last_edge), excluding
+    // last_edge.
+    //
+    // If first_edge == last_edge, then a RRNodeId has no edges.
+    RREdgeId first_edge(const RRNodeId& id) const {
+        return first_edge_[id];
+    }
+    RREdgeId last_edge(const RRNodeId& id) const {
```
vaughnbetz: Return the first_edge of the next rr_node, which is one past the edge id range for the node we care about. This implies we have one dummy rr_node at the end of first_edge_, and we always allocate that dummy node. We also assume that the edges have been sorted by rr_node, which is true after partition_edges().

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+    short edge_switch(const RRNodeId& id, t_edge_size iedge) const {
+        return edge_switch(edge_id(id, iedge));
+    }
+
+    /******************
+     * Fan-in methods *
+     ******************/
+
+    /* Init per node fan-in data.  Should only be called after all edges have
+     * been allocated */
```
vaughnbetz: This is an expensive, O(N), operation so it should be called once you have a complete rr-graph and not called often.

vpr/src/route/rr_graph_storage.h
```diff
@@ -0,0 +1,578 @@
...
+     *
+     *****************/
+
+    // storage_ stores the core RR node data used by the router and is **very**
+    // hot.
+    vtr::vector<RRNodeId, t_rr_node_data, vtr::aligned_allocator<t_rr_node_data>> storage_;
+
+    // The PTC data is cold data, and is generally not used during the inner
+    // loop of either the placer or router.
+    vtr::vector<RRNodeId, t_rr_node_ptc_data> ptc_;
```
vaughnbetz: prefix these with node_ for consistency.

vpr/src/route/route_timing.h
```diff
@@ -10,9 +10,196 @@
 #include "route_common.h"
 #include "router_stats.h"
 #include "router_lookahead.h"
+#include "spatial_route_tree_lookup.h"
 
 int get_max_pins_per_net();
 
+// This class encapsolates the timing driven connection router. This class
+// routes from some initial set of sources (via the input rt tree to a
```
vaughnbetz: add closing brace after "rt tree)"

vpr/src/route/route_timing.h
```diff
@@ -10,9 +10,196 @@
...
+    // empty).  When using cost_params.astar_fac = 0, for efficiency the
+    // RouterLookahead used should be the NoOpLookahead.
+    std::vector<t_heap> timing_driven_find_all_shortest_paths_from_route_tree(
+        t_rt_node* rt_root,
+        const t_conn_cost_params cost_params,
+        t_bb bounding_box,
+        RouterStats& router_stats);
+
+  private:
+    // Mark that inode has been modified, and needs to be reset in
```
vaughnbetz: Mark that the data associated with rr_node "inode" has been ...

vpr/src/route/route_timing.h
```diff
@@ -10,9 +10,196 @@
...
+    // Conditionally adds to_node to the router heap (via path from from_node
+    // via from_edge).
+    //
+    // RR nodes outside bounding box specified in bounding_box are not added
+    // to the heap.
+    void timing_driven_expand_neighbour(
+        t_heap* current,
+        const int from_node,
+        const RREdgeId from_edge,
+        const t_edge_size from_edge_int,
```
vaughnbetz: Suggest calling this one "from_node_edge_idx"

litghost: done

vpr/src/route/route_timing.h
litghost: done

vpr/src/route/route_timing.h
litghost: Done

vpr/src/route/rr_graph_storage.h
litghost: done

vpr/src/route/rr_graph_storage.h
litghost: done

vpr/src/route/rr_graph_storage.h
litghost: done

vpr/src/route/rr_graph_storage.h
litghost: done

vpr/src/route/rr_graph_storage.h
litghost: done

vpr/src/route/rr_graph_storage.h
litghost: Done

vpr/src/route/rr_graph_storage.h
litghost: Done

libs/libvtrutil/src/vtr_memory.h
litghost: done

litghost:
@vaughnbetz / @kmurray All comments addressed

litghost:
Travis CI is green now.

kmurray:
Thanks for all the work on this @litghost!

PR #1139, opened 2020-02-12
Detect missing name attribute on model port declaration.

litghost:
#### Description

Previously this would cause a std::logic_error to be thrown from
std::string.

This is a good example of something that would be caught using the uxsdcxx parser generator, rather than hand writing parsing.

#### Motivation and Context

Previous logic just reported:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

#### Types of changes

#### Checklist:


kmurray:
@litghost Can you fix the formatting? Otherwise it looks good.

Issue #537, opened 2019-04-19
check_route appears to be fairly slow


litghost:
#### Expected Behaviour

Given that ``check_route`` should just be a sanity check, I would expect it to be a fast check.  

#### Current Behaviour

Currently ``check_route`` appears to take a long as the routing step!

Runtime of routing with calls ``check_route`` commented out:
```
The entire flow of VPR took 374.07 seconds (max_rss 5237.5 MiB)
```
Runtime of routing with ``check_route`` not commented out:
```
The entire flow of VPR took 635.03 seconds (max_rss 5162.6 MiB)
```

That's approximately 261 seconds dedicated to ``check_route``, and that seems unreasonably high.

#### Possible Solution

1. Provide command line option to disable ``check_route``
2. Profile and optimize ``check_route``.  I suspect there is a O(n^2) or O(n) with large time constant algorithm lurking in ``check_route``.

#### Steps to Reproduce

1. Example from https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/526 ([vpr_bug3.zip](https://ufile.io/a6jpf)) demonstrates this issue.  Run with VPR with ``check_route`` commented out and run again without ``check_route`` commented out.
```
vpr \
 arch.unique_pack.xml \
 top.eblif \
 --device xc7a50t-basys3-test --read_rr_graph rr_graph_xc7a50t-basys3_test.rr_graph.real.xml \
 --min_route_chan_width_hint 100 --max_criticality 0.0 --max_router_iterations 500 --routing_failure_predictor off --router_high_fanout_threshold -1 \
 --constant_net_method route --route_chan_width 500 --clock_modeling route --place_algorithm bounding_box --enable_timing_computations off \
 --allow_unrelated_clustering on --route
```

With ``check_route``: ``The entire flow of VPR took 147.16 seconds (max_rss 6301.9 MiB)``
Without ``check_route``: ``The entire flow of VPR took 74.74 seconds (max_rss 6302.2 MiB)``

#### Context

[...]

PR #1147, opened 2020-02-14
Add --quick_check_route to disable slow non-configurable edge check

HackerFoo:
#### Description
The check in `check_non_configurable_edges()` is non-linear and takes a long time for the XC7 architecture in SymbiFlow.

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/537

#### Motivation and Context
This increases performance by making it possible to skip an expensive check.

This flag has been tested and is currently being used in [symbiflow-arch-defs](https://github.com/SymbiFlow/symbiflow-arch-defs)

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/vpr_api.cpp
+++ b/vpr/src/base/vpr_api.cpp
@@ -688,7 +688,9 @@ RouteStatus vpr_route_flow(t_vpr_setup& vpr_setup, const t_arch& arch) {
         std::string graphics_msg;
         if (route_status.success()) {
             //Sanity check the routing
-            check_route(router_opts.route_type);
+            if (!router_opts.disable_check_route) {
+                check_route(router_opts.route_type, router_opts.quick_check_route);
+            }
             get_serial_num();
 
             //Update status
--- a/vpr/src/route/check_route.cpp
+++ b/vpr/src/route/check_route.cpp
@@ -3,6 +3,7 @@
 #include "vtr_assert.h"
 #include "vtr_log.h"
 #include "vtr_memory.h"
+#include "vtr_time.h"
 
 #include "vpr_types.h"
 #include "vpr_error.h"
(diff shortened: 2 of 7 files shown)


vpr/src/base/vpr_api.cpp
```diff
@@ -688,7 +688,9 @@ RouteStatus vpr_route_flow(t_vpr_setup& vpr_setup, const t_arch& arch) {
         std::string graphics_msg;
         if (route_status.success()) {
             //Sanity check the routing
-            check_route(router_opts.route_type);
+            if (!router_opts.disable_check_route) {
+                check_route(router_opts.route_type, router_opts.quick_check_route);
+            }
```
kmurray: Looks like this also adds an option to disable check_route? Not sure where `router_opts.disable_check_route` is coming from (don't see it in this PR).

vpr/src/route/check_route.cpp
```diff
@@ -622,6 +626,18 @@ static void check_node_and_range(int inode, enum e_route_type route_type) {
     check_rr_node(inode, route_type, device_ctx);
 }
 
+//Checks that all non-configurable edges are in a legal configuration
+//This check is slow, so it has been moved out of check_route()
+static void check_all_non_configurable_edges() {
+    vtr::ScopedStartFinishTimer timer("Checking to ensure non-configurable edges are legal");
+    auto non_configurable_rr_sets = identify_non_configurable_rr_sets();
```
kmurray: I believe the non-config rr sets are now stored at a RR graph creation in the device_ctx, so we likely don't need to re-calculate it here.

litghost:
@HackerFoo Are you planning on updating this PR address comments?

HackerFoo:
@litghost Yes, I'll work on this.

Issue #1161, opened 2020-02-19
RR graphs with reordered nodes don't load correctly


HackerFoo:
#### Expected Behaviour
Remapping node IDs should not change the behavior of VPR, other than affecting cache locality.

#### Current Behaviour
Doing so will crash VPR if `grid_loc` ptc attributes are not in increasing order with respect to the enclosing node ID.

#### Possible Solution
Fix code in the serializer.

#### Steps to Reproduce
Reorder node IDs, using a program such as https://github.com/HackerFoo/Gorder.

#### Context
I'm experimenting with graph reordering algorithms to see if there is any benefit.

#### Your Environment
* VTR revision used: SymbiFlow/master
* Operating System and version: NixOS
* Compiler version: GCC 8.3

PR #1162, opened 2020-02-19
Fix Serialization of ptcs

HackerFoo:
#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1161
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1079

#### Motivation and Context
This bug prevents testing different graph ordering algorithms, which could benefit VPR runtime.

It has been tested with random and specific orderings using https://github.com/HackerFoo/Gorder in symbiflow-arch-defs.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/rr_graph_uxsdcxx_serializer.h
+++ b/vpr/src/route/rr_graph_uxsdcxx_serializer.h
@@ -1569,27 +1569,25 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
             if (node.type() == SOURCE || node.type() == SINK) {
                 for (int ix = node.xlow(); ix <= node.xhigh(); ix++) {
                     for (int iy = node.ylow(); iy <= node.yhigh(); iy++) {
-                        if (node.type() == SOURCE) {
-                            indices[SOURCE][ix][iy][0].push_back(inode);
-                            indices[SINK][ix][iy][0].push_back(OPEN);
-                        } else {
-                            VTR_ASSERT(node.type() == SINK);
-                            indices[SINK][ix][iy][0].push_back(inode);
-                            indices[SOURCE][ix][iy][0].push_back(OPEN);
+                        if (node.ptc_num() >= (int)indices[SOURCE][ix][iy][0].size()) {
+                            indices[SOURCE][ix][iy][0].resize(node.ptc_num() + 1, OPEN);
                         }
+                        if (node.ptc_num() >= (int)indices[SINK][ix][iy][0].size()) {
+                            indices[SINK][ix][iy][0].resize(node.ptc_num() + 1, OPEN);
+                        }
+                        indices[node.type()][ix][iy][0][node.ptc_num()] = inode;
                     }
                 }
             } else if (node.type() == IPIN || node.type() == OPIN) {
                 for (int ix = node.xlow(); ix <= node.xhigh(); ix++) {
                     for (int iy = node.ylow(); iy <= node.yhigh(); iy++) {
-                        if (node.type() == OPIN) {
-                            indices[OPIN][ix][iy][node.side()].push_back(inode);
-                            indices[IPIN][ix][iy][node.side()].push_back(OPEN);
-                        } else {
-                            VTR_ASSERT(node.type() == IPIN);
-                            indices[IPIN][ix][iy][node.side()].push_back(inode);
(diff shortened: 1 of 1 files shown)


vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1569,27 +1569,25 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
                     for (int iy = node.ylow(); iy <= node.yhigh(); iy++) {
-                        if (node.type() == SOURCE) {
-                            indices[SOURCE][ix][iy][0].push_back(inode);
-                            indices[SINK][ix][iy][0].push_back(OPEN);
-                        } else {
-                            VTR_ASSERT(node.type() == SINK);
-                            indices[SINK][ix][iy][0].push_back(inode);
-                            indices[SOURCE][ix][iy][0].push_back(OPEN);
+                        if(node.ptc_num() >= indices[SOURCE][ix][iy][0].size()) {
+                            indices[SOURCE][ix][iy][0].resize(node.ptc_num() + 1, OPEN);
```
litghost: Rather than resizing multiple times, compute the required size and allocate once.

HackerFoo: I don't think the effort involved to do this is worthwhile, and will not necessarily be faster. Note that this change requires no more allocations in the worst case than the code it replaced.

litghost: > I don't think the effort involved to do this is worthwhile, and will not necessarily be faster. Note that this change requires no more allocations in the worst case than the code it replaced.

That is not true, because `std::vector::push_back` is worst case amortized `O(1)`, and your code is worst case `O(N)`.

HackerFoo: Sure, but N `push_back`s are `O(N)`, while M upward `resize`s to N are `O(M + N)`, and M < N for `O(N)`.

HackerFoo: The average case is about half that, though.

litghost: > while M upward `resize`s to N are `O(M + N)`

But you also have the data copy implications, which becomes `O(N^2)`.  So not only are you generating `N` allocations, instead of `log(N)` allocations, but you are generating potentially significantly more copying.

HackerFoo: At any rate, it takes 0.08 seconds.

litghost: To be clear, `ptc` is roughly equal to # of IPINs + # OPINs, which on a BRAM is ~360.  So worst case your code will generate ~360 allocations and ~64980 copies, versus 9 allocations and ~3200 copies with the old code that used `push_back`.

litghost: Oh, and the PCIE block has 2249 pins, so worst case your code will generated 2249 allocations, and ~2530125 copies, versus 11 allocations and ~26988 copies.

HackerFoo: For the loop at 1567-1606 (https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d809ad6b618821eecee5ef04edf683b872805a04/vpr/src/route/rr_graph_uxsdcxx_serializer.h#L1567-L1606)

HackerFoo: `v.push_back(x);` is equivalent to `v.resize(v.size() + 1); v[v.size()-1] = x;`.

`std::vector` will take care of growing capacity in order to avoid excessive reallocations, that's the point of it.

HackerFoo: Anyway, it's not worth wasting any more time over 80 milliseconds.

litghost: > `v.push_back(x);` is equivalent to `v.resize(v.size() + 1); v[v.size()-1] = x;`.
> 
> `std::vector` will take care of growing capacity in order to avoid excessive reallocations, that's the point of it.

You are correct, I was mistaken.

PR #1166, opened 2020-02-20
Treat clock nets specially for routing purposes.

litghost:
Signed-off-by: Keith Rothman <537074+litghost@users.noreply.github.com>

#### Description

- Use max_criticality for clock nets.  This will cause them to use min
  delay paths rather than avoid congestion.  As a future enchancement,
  the clock nets should likely route for min slew, but that is a larger
  change.
- Clocks should use a bounding box that includes the entire fabric.
  This is when a clock spine extends from a global buffer point to avoid
  the net target, the default bounding box behavior will prevent the
  router from finding a path using the clock network.  This is not
  catastrophic if the only path to the clock sink is via the clock
  network, because eventually the heap will empty and the router will
  use the full part bounding box anyways.

  Where the problems start is when there is a path from the clock network to
  the general interconnect and vise-versa.  Even if this path has a high
  cost, because it is within the bounding box, the router will find the
  clock network -> general interconnect -> clock network path, rather
  than staying on the clock network the whole time.  If the bounding box
  is the entire part, this will not happen because the router will
  quickly explore the cheap path (the clock network) well before
  exploring the significantly more expensive clock network -> general
  interconnect -> clock network path.

#### Related Issue

https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1328

#### Motivation and Context

swkb6Xt-th5RhgTc7sTBpIA

Consider the example clock network above. The clock row and clock column buffers are represented by edges in the graph.  The clock buffer is represented by a sub-circuit.  So in order for the router to stay on the clock network, it must go through both a row and column buffer to reach the clock sinks within the CLB's.

In the happy case, consider the green net bounding box from the clock buffer to the CLB 0,1.  In this case the entire clock network is encompassed, so the router will behave well.

Now consider that sad case, consider the red net bounding box from the clock buffer to the CLB 2,0.  In this case neither the clock row buffer nor the clock column buffer is included in the net bounding box.  In this case, the router will fully expand the graph within the red net bounding box.  In the event there is any (even a bad) path, the router will choose it.  This may result in the router finding a bad path avoiding the clock network, because the bounding box excludes the "good" route through the clock network.

Tested on symbiflow designs where the router was avoiding the clock network because of either criticality (e.g. avoiding long wires typically found in the clock network) or because the router was prevented from exploring the spine by the default route bb.

#### Types of changes

#### Checklist:


litghost:
@vaughnbetz @kmurray @acomodi FYI

litghost:
Strong failure is actually a CPD improvement:
```
regression_tests/vtr_reg_strong/strong_dedicated_clock...[Fail] 
 timing/k6_frac_N10_frac_chain_mem32K_htree0_routedCLK_40nm.xml/verilog/mkPktMerge.v/common_--clock_modeling_dedicated_network critical_path_delay: golden = 5.30223 result = 4.11344
[Fail] 
 timing/k6_frac_N10_frac_chain_mem32K_htree0_routedCLK_40nm.xml/verilog/mkPktMerge.v/common_--clock_modeling_dedicated_network setup_WNS: golden = -5.30223 result = -4.11344
```

vpr/src/route/route_common.cpp
```diff
@@ -1114,15 +1117,46 @@ static vtr::vector<ClusterBlockId, std::vector<int>> load_rr_clb_sources(const t
...
+        full_device_bounding_box.ymin = 0;
+        full_device_bounding_box.xmax = device_ctx.grid.width() - 1;
+        full_device_bounding_box.ymax = device_ctx.grid.height() - 1;
+    }
 
     auto nets = cluster_ctx.clb_nlist.nets();
     route_bb.resize(nets.size());
     for (auto net_id : nets) {
-        route_bb[net_id] = load_net_route_bb(net_id, bb_factor);
+        if (!route_ctx.is_clock_net[net_id]) {
```
vaughnbetz: Please add a comment on why you're doing this (well explained in the PR, but put an overview here too).
Is is_clock_net true for only globally routed clocks (special network) or all clocks including locally routed clocks?  Ideally we'd turn off the bounding box only for clocks that are being routed on special global networks.

vpr/src/route/route_timing.cpp
```diff
@@ -1007,7 +1007,12 @@ bool timing_driven_route_net(ClusterNetId net_id,
     for (int ipin : remaining_targets) {
         if (timing_info) {
             auto clb_pin = cluster_ctx.clb_nlist.net_pin(net_id, ipin);
-            pin_criticality[ipin] = calculate_clb_net_pin_criticality(*timing_info, netlist_pin_lookup, clb_pin);
+            if (!route_ctx.is_clock_net[net_id]) {
+                pin_criticality[ipin] = calculate_clb_net_pin_criticality(*timing_info, netlist_pin_lookup, clb_pin);
```
vaughnbetz: Add comment on why you're using max_criticality (don't get reasonable criticalities for clocks; really want low skew which is approximated by routing them for min delay, particularly if they are on a dedicated network).

vpr/src/route/route_common.cpp
litghost: > Is is_clock_net true for only globally routed clocks (special network) or all clocks including locally routed clocks? Ideally we'd turn off the bounding box only for clocks that are being routed on special global networks.

Any net that sinks at a `<clock>` is consider a clock net.

vpr/src/route/route_timing.cpp
litghost: Add comment

litghost:
@vaughnbetz / @kmurray I believe this is ready for review.

kmurray:
Looks good to me.

PR #1171, opened 2020-02-27
Fix missing <sstream> include for use of std::istringstream

hzeller:
in id_or_pin_name_to_tn()


kmurray:
Thanks for the fix!

PR #1173, opened 2020-02-27
vpr: use alias name mapping for clock nets when removing buffer luts

acomodi:

#### Description

This patch disables the lut buffer removal for clock nets during netlist cleanup.

#### Motivation and Context

VPR modifies the input circuit unless `--absorb_buffer_luts` is turned off. 
This is ok in general, except for clock nets. In fact, when the buffer between two clock nets is removed, the resulting net is renamed, using the combination of the two previous nets.
This behavior makes it hard to write correct `sdc` files, as the name of a clock in the original netlist may be changed by VPR in case the signal is propagated through submodules or `assign` statements are used.

#### Types of changes

#### Checklist:


litghost:
This seems like the wrong solution?  It means the clock nets will have one name that is valid, rather than any of the valid names.  It also means some lut1 will exist on clock nets.

acomodi:
@litghost Another solution I thought of and testing now is to allow the removal of the buffer, but rename it correctly. 

What I am not sure of is how to correctly rename it though, if it is ok to use only the input_net name

acomodi:
One of the issues I am having is that there are some nets which are identifyied as clock nets, but shouldn't actually be clock nets. These non-clock nets are the `$true` and `$false` ones.

litghost:
> @litghost Another solution I thought of and testing now is to allow the removal of the buffer, but rename it correctly.
> 
> What I am not sure of is how to correctly rename it though, if it is ok to use only the input_net name

Maintain a map of original names -> final nets, and use that mapping in the SDC parser stage.

litghost:
> @litghost Another solution I thought of and testing now is to allow the removal of the buffer, but rename it correctly.
> 
> What I am not sure of is how to correctly rename it though, if it is ok to use only the input_net name

Maintain a map of original names -> final nets, and use that mapping in the SDC parser stage.

acomodi:
@litghost I have actually adopted another solution that needs less changes. The buffer is removed and the new driver pin of the merged net is always the input one, so now there is a check whether the new driver pin is a clock pin, and, if yes, the input_net name is taken as new name of the merged net.

acomodi:
One thing that I am checking is the final name given to the clock net.

Apparently, the clock name is relative to the location where the clock net is generated. This means that, if a clock is generated in the top module, the net name has no prefixes. E.g.:

```verilog
module top(
    input clk,
    output led
);

wire sys_clk;

BUFG BUFG_1(.I(clk), .O(sys_clk));

CPU CPU_instance(
    .clk(sys_clk),
    ...
)
...

endmodule
```

The two clock names in this case are: `sys_clk` and `clk`.

Instead, if BUFG was instantated in the CPU module, its name would be: `CPU_instance.sys_clk`

acomodi:
With further analysis, the clock name that is now used within VPR after lut buffer removal matches the name relative to the clock primitives clock input pin of the `.eblif`.

acomodi:
@litghost I have adopted your suggestion, as it is more suitable. This way we can use all the names from the netlist that are relative to a specific clock net in the sdc.

E.g.:

**Before**: 
- clock net name: `sys_clk__module1.clk__module3.another_clk__moduleN.clk`
- sdc clock name: `sys_clk__module1.clk__module3.another_clk__moduleN.clk`

**Now**:
- clock net name: `sys_clk`
- possible sdc clock names: `sys_clk`, `module1.clk`, `module3.another_clk`, `moduleN.clk`

acomodi:
CI went green, I have also added some in-code comments.

vpr/src/base/atom_netlist_utils.cpp
```diff
@@ -38,7 +38,7 @@ bool is_removable_input(const AtomNetlist& netlist, const AtomBlockId blk, std::
 bool is_removable_output(const AtomNetlist& netlist, const AtomBlockId blk, std::string* reason = nullptr);
 
 //Attempts to remove the specified buffer LUT blk from the netlist. Returns true if successful.
-bool remove_buffer_lut(AtomNetlist& netlist, AtomBlockId blk, int verbosity);
+bool remove_buffer_lut(AtomNetlist& netlist, std::set<AtomPinId> clock_drivers, AtomBlockId blk, int verbosity);
```
kmurray: Since I suggest generalizing this to capture *all* merged nets we don't need to pass in the clock drivers specifically.

vpr/src/base/atom_netlist.cpp
```diff
@@ -188,6 +200,10 @@ AtomNetId AtomNetlist::add_net(const std::string name, AtomPinId driver, std::ve
     return Netlist::add_net(name, driver, sinks);
 }
 
+void AtomNetlist::add_clock_net_alias(const std::string alias_net_name, const std::string assigned_net_name) {
```
kmurray: This would be come `add_net_alias()`

vpr/src/base/atom_netlist_utils.cpp
```diff
@@ -870,6 +879,11 @@ bool remove_buffer_lut(AtomNetlist& netlist, AtomBlockId blk, int verbosity) {
     } else if (!(driver_is_pi || po_in_input_sinks) && po_in_output_sinks) {
         //Must use the output name to perserve primary-output name
         new_net_name = netlist.net_name(output_net);
+    } else if (new_driver_is_clock) {
+        auto input_net_name = netlist.net_name(input_net);
+        auto output_net_name = netlist.net_name(output_net);
+        new_net_name = input_net_name;
+        netlist.add_clock_net_alias(output_net_name, input_net_name);
```
kmurray: This branch would no longer be needed, and calls to:
```cpp
netlist.add_net_alias(new_net_name, input_net_name);
netlist.add_net_alias(new_net_name, output_net_name);
```
would be always added.

vpr/src/base/atom_netlist.cpp
```diff
@@ -102,6 +102,18 @@ AtomBlockId AtomNetlist::find_atom_pin_driver(const AtomBlockId blk_id, const t_
...
+std::string AtomNetlist::get_assigned_net_name(const std::string alias_net_name) const {
+    std::string assigned_net_name;
+
+    auto result = clock_net_names_map_.find(alias_net_name);
+
+    if (result != clock_net_names_map_.end()) {
+        assigned_net_name = result->second;
+    }
+
+    return assigned_net_name;
```
kmurray: It looks like this will return the empty string for a net name which is not in the alias list?

This should check for an alias first (via `clock_net_names_map_`), using the re-mapped name if found or otherwise the passed argument (`alias_net_name`). It should then check that the 'resolved' net name is actually valid (to handle the case of an invalid `alias_net_name` being passed).

acomodi: One of the original reasons I did leave the empty string is that it is possible that the SDC target clock has a wildcard in the name.
If this happens, the alias map would not contain the alias net, and I didn't think it was a good idea to output the alias name in case the assigned net name was not found.

What I am doing now is to check if the net alias is a valid clock net, and, if not, it returns an empty string.

vpr/src/base/atom_netlist_utils.cpp
acomodi: Done, also added a check during insertion to see if the alias key is already present and if so, the content needs to match to the net that was to be assigned.

vpr/src/base/atom_netlist.cpp
acomodi: Done

vpr/src/base/atom_netlist_utils.cpp
acomodi: Done

vpr/src/base/atom_netlist.cpp
kmurray: > One of the original reasons I did leave the empty string is that it is possible that the SDC target clock has a wildcard in the name.

Hmm, I see your point.

I think it would be best to leave any wildcard handling to the SDC processing rather than the netlist data structures (which are more fundamental).

I'd suggest we do the following:
1. Add a method to the netlist which allows iterating through the net alias names (e.g. `AtomNetlist::net_aliases(AtomNetId net_id)`).
2. Have the SDC code iterate through both the current netlist net names (i.e. `AtomNetlist::nets()`), and the aliases (`AtomNetlist::net_aliases()`) to do wild-card name matching (and thereby capturing both current and aliased net names).
3. Have the SDC code resolve the wild card matches back to the actual netlist AtomNetId (via  `AtomNetlist::get_assigned_net_name()` and `AtomNetlist::find_net()`) to which the constraints actually apply.

acomodi: I have done something similar:

1. `AtomNetlist::net_aliases(AtomNetId)` which return the set of aliases for the specified net.
2. during buffer lut removal add both input net and output net to the aliases (also for the IPAD and OPAD related nets)
3. During SDC, for each net, iterate through the set of aliases (which contains also the assigned net name during buffer_removal) and check the regex match.

If buffer lut removal is disabled, the `net_aliases` function returns a set with the original net name. 

Each setter and getter for the aliases set has an assertion to verify the net is valid.

vpr/src/base/atom_netlist.cpp
```diff
@@ -188,6 +205,21 @@ AtomNetId AtomNetlist::add_net(const std::string name, AtomPinId driver, std::ve
...
+    auto result = net_aliases_map_.find(net_id);
+
+    if (result != net_aliases_map_.end()) {
+        aliases = result->second;
+        aliases.insert(alias_net_name);
+    } else {
+        aliases.insert(alias_net_name);
+        net_aliases_map_.insert({net_id, aliases});
+    }
+}
```
kmurray: I'm not sure this is correct. Since `aliases` is a local variable, the first branch of the if will insert `alias_net_name` into the local variable copy of the set, and is never added to `net_aliases_map_`.

I'd suggest something like:
```
net_aliases_map_[net_id].insert(alias_net_name);
```
which is much simpler and handles both cases (where net_id has been seen/not seen) directly.

vpr/src/base/atom_netlist.h
```diff
@@ -206,6 +217,9 @@ class AtomNetlist : public Netlist<AtomBlockId, AtomPortId, AtomPinId, AtomNetId
 
     //Port data
     vtr::vector_map<AtomPortId, const t_model_ports*> port_models_; //Architecture port models of each port
+
+    //Clock net aliases
```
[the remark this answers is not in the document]
acomodi: Sure, done

vpr/src/base/atom_netlist.cpp
acomodi: Yep, much nicer as you suggested. Done

kmurray:
Thanks @acomodi for working through this, the code looks good to merge!

I just had one more thought, it would be good to have some regression tests to cover this behaviour (i.e. interaction between buffer absorption and SDC name mapping) to ensure we have consistent behaviour, and don't break it in the future. Such tests would probably go in vtr_reg_strong.

acomodi:
>  it would be good to have some regression tests to cover this behaviour

@kmurray Sounds like a good idea

acomodi:
@kmurray I had to change the addressing to the `net_alias_map_`. In fact, if the key is an AtomNetId, the aliases set could result to be empty. This because it is possible (I didn't track it down, though) that the net could be modified once again in a following step, therefore, the clock_net_id that is used to retrieve the aliases set in the read_sdc step, may be invalid, in the sense that it does not contain the aliases.

Instead, by using the name of the net as the key of the map, this behavior does not show up. There is of course a check in the net_aliases function to see whether the corresponding net_id is valid.

Regarding the regression test, I have generated a .blif file and than used three different .sdc files  that do create clocks using different aliases of the same clock net.

kmurray:
> Instead, by using the name of the net as the key of the map, this behavior does not show up. There is of course a check in the net_aliases function to see whether the corresponding net_id is valid.

That actually does make sense, you could have multiple buffers being absorbed which could cause a previously merged net to be merged with another. We also re-number the various IDs at the end of netlist cleaning, which if not accounted (c.f. remap_ids) for could also cause issues. Using the net names is likely easier and more robust as you noted.


> Regarding the regression test, I have generated a .blif file and than used three different .sdc files that do create clocks using different aliases of the same clock net.

That sounds good. I guess adding the tests was a good exercise since it caught the above issue you noted!

acomodi:
> That sounds good. I guess adding the tests was a good exercise since it caught the above issue you noted!

Yep, that was a very good call. Always better to add tests over a change like this.

kmurray:
Thanks for putting this together!

PR #1175, opened 2020-03-02
Add vtr::array_view object.

litghost:
This enables creating views into allocated data, without direct
ownership.


kmurray:
Looks good!

PR #1178, opened 2020-03-02
Add CMake option to enable/disable building of ODIN.

litghost:
#### Description

This PR adds a CMake option to control the building of ODIN.  This option is useful for cutting down compile times.

#### Types of changes

#### Checklist:


kmurray:
How would you contrast this with just doing `make vpr`?

litghost:
> How would you contrast this with just doing `make vpr`?

This modifies `make all` and `make install`, which Symbiflow uses during packaging.  Specifically, the compile runtime on Travis of packaging was hitting a limit when we included all of VTR, so we needed to cut out ODIN/ADC/etc.  We have a local patch to do this, but it goes stalled, hence this PR.

kmurray:
OK that makes more sense.

@jeanlego Any thoughts on this?

CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ option(VTR_ENABLE_CAPNPROTO "Enable capnproto binary serialization support in VP
 option(WITH_BLIFEXPLORER "Enable build with blifexplorer" OFF)
 option(WITH_LIBRTLNUMBER "Enable build with librtlnumber" OFF)
 
+option(ODIN_BUILD "Enable building odin" ON)
```
jeanlego: I would change the name to option(WITH_ODIN"Enable build with odin" ON)

CMakeLists.txt
```diff
@@ -341,16 +342,20 @@ endif()
 
 
 #Add the various sub-projects
-add_subdirectory(abc)
+if(${ODIN_BUILD})
```
jeanlego: it would make more sense for it all to be under the same if to keep things together as much as possible.

Also librtlnumber is only consumed by odin, so it should only be enabled if building with odin. 
blifexplorer depends on odin and this should also be reflected here.

I think ABC should have a separate flag i.e. WITH_ABC since odin and abc dont depend on each other and are discreet part of the flow

litghost: > odin and abc

ODIN links against abc?

litghost: I've updated the CMakeLists with you suggestions.

CMakeLists.txt
litghost: Done

jeanlego:
> OK that makes more sense.
> 
> @jeanlego Any thoughts on this?

functionally, this looks good to me now on my side.
Odin does not actually use anything from abc anymore. 
I'll raise an issue to remove the stale link and replace it with pthread.

thanks for this!

kmurray:
> functionally, this looks good to me now on my side.

Great, thanks for the review!

PR #1177, opened 2020-03-02
Encapsulate binary heap.

litghost:
#### Description

This work is in preparation for integrating the heap approximation.

#### Related Issue

#952

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #952, opened 2019-09-06
Refactor router heap to use heap approximation.

litghost:
### Description

The heap approximation doesn't precisely follow the heap property, but
offers an approximation that is sufficient for the router's purpose.
This new data structure is faster to clear, faster during route time,
and results in better router behavior because it provides some
randomness on elements with costs that are same within 1 ps.

#### Motivation and Context

Profiler data showed a significant amount of time being spent managing the heap, especially the heap pop operation.  This refactoring has a couple gains for a couple reasons:
 - The new data structure does not provide an exact heap.  Instead it is an approximation based on grouping of cost.  This cost grouping is currently being done on a [0, 1) picosecond (1e-12) basis.
 - The new data structure provides some randomness on nodes with costs in the same grouping.  For items with identical cost, this avoids sending the router down identical paths, and resolves some congestion without explicitly costing it.
 - The new data structure and memory management clear very quickly by avoiding rebuilding the free list.

In limited functional testing these changes resulted in a 30-40% runtime decrease with a ~3% CPD increase.

#### How Has This Been Tested?

Some smaller functional tests (44k connections) show better convergence properties, and better per iteration runtime, and better overall runtime.

Titan benchmarks are being run now to see if the change holds on a variety of designs (especially larger ones).

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1179, opened 2020-03-03
Move heap ownership to the ConnectionRouter.

litghost:
#### Description

This PR combines #1176 and #1177 , and completely migrates the encapsulated heap to be owned by the `ConnectionRouter` (or the breadth first search router).

Once this change is merged, heap to the `ConnectionRouter` can be templated over to enable switch of heaps.  This change also allows the 3 places the heap is used to be controlled independently.

#### Related Issue

#952
#1176 
#1177 

#### Motivation and Context

By moving ownership of the heap to the `ConnectionRouter`, different heap implementations will easily be swapable.

- [ ] CI is green
- [ ] QoR comparisions against Titan benchmarks

#### Types of changes

#### Checklist:


vpr/src/route/connection_router.cpp
```diff
@@ -8,7 +8,7 @@
 //as valid start locations for the current connection.
 //
 //Returns either the last element of the path, or nullptr if no path is found
-t_heap* ConnectionRouter::timing_driven_route_connection_from_route_tree(
+std::pair<bool, t_heap> ConnectionRouter::timing_driven_route_connection_from_route_tree(
```
kmurray: Can you clarify what's the motivation for changing the return value?

vpr/src/route/connection_router.h
```diff
@@ -64,14 +69,18 @@ class ConnectionRouter {
         ::reset_path_costs(modified_rr_node_inf_);
     }
 
+    void empty_heap() {
+        heap_.empty_heap();
+    }
```
kmurray: Can you clarify why this needs to be exposed in the public interface?

vpr/src/route/heap_type.h
```diff
@@ -159,4 +160,66 @@ class HeapInterface {
...
+    hptr->cost = total_cost;
+    VTR_ASSERT_SAFE(hptr->u.prev.node == NO_PREVIOUS);
+    VTR_ASSERT_SAFE(hptr->u.prev.edge == NO_PREVIOUS);
+    hptr->u.prev.node = prev_node;
+    hptr->u.prev.edge = prev_edge;
+    hptr->backward_path_cost = backward_path_cost;
+    hptr->R_upstream = R_upstream;
+
+    heap->add_to_heap(hptr);
+}
```
kmurray: I'd tend to classify these as router (rather than heap type) constructs. So they're probably better in router files (perhaps route_common.h or route_timing.h).

Additionally, I think the only difference between these two operations is whether we call add_to_heap() or push_back() at the end, so cost check and hptr setup should really be factored out in to a common function.

vpr/src/route/route_common.cpp
```diff
@@ -1383,6 +1372,8 @@ void reserve_locally_used_opins(float pres_fac, float acc_fac, bool rip_up_local
         }
     }
 
+    heap->empty_heap();
```
kmurray: It's not clear why this is needed here, so we should have a comment describing why.

vpr/src/route/connection_router.cpp
litghost: It ensures that all memory allocated by the `ConnectionRouter` stays within the connection router.  The previous code transfers ownership to the caller, who would read some fields and then immediately free it.

vpr/src/route/route_common.cpp
litghost: Done

vpr/src/route/connection_router.h
litghost: Done

vpr/src/route/route_common.cpp
litghost: Done

vpr/src/route/heap_type.h
litghost: Agreed, done

vpr/src/route/connection_router.h
litghost: Done

PR #1185, opened 2020-03-04
Add heap approximation

litghost:
#### Description

This PR builds on #1177 and #1179.

This is follow-up to #952 that adds a second heap implemented.  See #952 for previous QoR and comments.

This PR changes the `ConnectionRouter` to be template'd over the heap implementation, to enable swap of heap implementations as needed.  If this approach is good, I can add a command line option to control the heap implementation, or simply leave it as a template parameter.

#### Related Issue

#952

#### Motivation and Context

Profiler data showed a significant amount of time being spent managing the heap, especially the heap pop operation. This refactoring has a couple gains for a couple reasons:

- The new data structure does not provide an exact heap. Instead it is an approximation based on grouping of cost. This cost grouping is currently being done on a [0, 1) picosecond (1e-12) basis.
- The new data structure provides some randomness on nodes with costs in the same grouping. For items with identical cost, this avoids sending the router down identical paths, and resolves some congestion without explicitly costing it.
- The new data structure and memory management clear very quickly by avoiding rebuilding the free list.

In limited functional testing these changes resulted in a 30-40% runtime decrease with a ~3% CPD increase.

#### Types of changes

#### Checklist:


vpr/src/route/connection_router.h
```diff
@@ -38,20 +39,25 @@ struct t_conn_cost_params {
 // rr_node_route_inf.  The routed path can be found by tracing from the sink
 // node (which is returned) through the rr_node_route_inf.  See
 // update_traceback as an example of this tracing.
+template<typename HeapImplementation>
```
kmurray: I think its worth discussing whether using a template (i.e. compile time polymorphism) or relying on the virtual functions of the HeapInterface (i.e. run-time polymorphism) is the right approach here.

I quite like the idea of having a command-line option to control which heap is in use (e.g. `--router_heap { binary | bucket }`), as it makes it easier to fall-back to the binary heap if necessary (e.g. if due to bugs or other poor behavior).

Compile time polymorphism is theoretically more run-time efficient, but I expect a the virtual function call to easily predictable in this case (it would always go the same way during a run of VPR). So I think this is something worth considering, and most importantly measuring.

kmurray:
@vaughnbetz See my comments above on virtual vs templating the heap type. Any thoughts?

vpr/src/route/connection_router.h
litghost: We can still have a command-line option to switch between heaps with the template, it just costs the size of the code increase.  I expect template performance to be significantly faster than run-time polymorphism.

litghost: > Compile time polymorphism is theoretically more run-time efficient, but I expect a the virtual function call to easily predictable in this case (it would always go the same way during a run of VPR). So I think this is something worth considering, and most importantly measuring.

The cost is the inability to inline across the vtable call.  The overhead of pushing and popping the stack for that call is likely significant.  I agree measurements are required.

kmurray: > We can still have a command-line option to switch between heaps with the template, it just costs the size of the code increase.

Yep, that may make the most sense. Are you envisioning a run-time option to choose between ConnectionRouter<BinaryHeap> or ConnectionRouter<BucketHeap>? If we can do that reasonably cleanly I'd be happy with it.

I'm not really concerned about the mechanism (template vs virtual), just that it would be good to be able to switch the heap with a command-line option.

litghost: > I'm not really concerned about the mechanism (template vs virtual), just that it would be good to be able to switch the heap with a command-line option.

The PR has been updated with a command line flag already (~2 minutes ago), so take a look.

vpr/src/route/route_common.cpp
```diff
@@ -307,14 +307,33 @@ bool try_route(int width_fac,
...
+                    net_delay,
+                    netlist_pin_lookup,
+                    timing_info,
+                    delay_calc,
+                    first_iteration_priority);
+                break;
+            default:
+                VPR_ERROR(VPR_ERROR_ROUTE, "Unknown heap_type %d",
+                          router_opts.router_heap);
+        }
```
kmurray: This is a bit awkward, and requires templating all the timing_driven_route_* functions by the heap type.

What are you're thoughts about making a `ConnectionRouterInterface`, and having `ConnectionRouter<BinaryHeap>` and `ConnectionRouter<Bucket>` implement it?

That would ensure the core loop (path search) is fully templated, while avoiding templating the rest of the router. 

Since the number of invocations of the Connection router is  small compared to the number of heap operations any overhead of the virtual call should be negligible (e.g. on the large Titan bitcoin_miner design, there are 178x fewer invocations of the connection router than heap operations (3M vs 680M)).

litghost: I'll post the virtual version soon and once the nightly CI completes we can compare https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1185/commits/87577fd21f8865477607afd85428702d6334667c (fully templated) with the new interface-at-ConnectionRouter.

If the performance hit is small, I think having the `ConnectionRouterInterface` interface is superior to fully templating the router!

litghost: I've kicked off two kokoro runs:
 - Nightly with fully templated router: https://source.cloud.google.com/results/invocations/4ceccd64-1170-42f7-be59-b816c53300fe
 - Nightly with `ConnectionRouterInterface`: https://source.cloud.google.com/results/invocations/0ed2f9b6-f74a-4824-af6d-ef9c25e10c62/targets

In about 12-16 hrs we'll be able to compare the relative impacts.

kmurray: I like the virtualized interface, it keeps the code much cleaner.

We'll see how the results come back, but hopefully the impacts will be small!

litghost:
Initial results are in: 
[template_vs_virtual.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4299731/template_vs_virtual.xlsx)

There is an overall 3% slowness from the virtual run, which is likely baseline noise, but that still leaves a 3% slowdown on critical path route time.  Let's wait and see what the results from https://source.cloud.google.com/results/invocations/9e3a79a1-cf95-40d3-9e9e-de9495c5afd8 come back with, as it possible this is just noise.

If needed I will run on dedicate machines to get a clearer signal, but I wanted to try using the data form kokoro CI.

litghost:
I have run QoR on dedicate machines for the baseline version and full template version, and the QoR looks good: 
[baseline_vs_heap.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4299773/baseline_vs_heap.xlsx)

I'll kick off the virtual version now, which should give a clear answer about whether the virtual version has a negative affect on performance.

litghost:
QoR results are in for the virtual interface, and they are not pretty: 
[baseline_vs_heap.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4308332/baseline_vs_heap.xlsx)

It is roughly a 5% runtime increase to use the virtual interface versus the fully templated version.  I'll try to make a cleaner templated version.

litghost:
@kmurray Take a look now.  I don't believe the templating is less painful now.  I'm re-running the QoR at https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1185/commits/afc346f26381f7726259cc22d18e79beb25f9806

litghost:
> 5% runtime

I do need to clarify:
 - Overall VPR runtime change 1.4% higher
 - Critical path route time change 5% higher

litghost:
> I think there should be a comment somewhere explaining that we're templating ConnectionRouter so you can get different heaps (another template) with no run-time overhead.

Sure, added

vpr/src/route/bucket.cpp
```diff
@@ -0,0 +1,234 @@
...
+    float max_cost = max_cost_;
+    VTR_ASSERT(max_cost != std::numeric_limits<float>::min());
+    if (min_cost == std::numeric_limits<float>::max()) {
+        min_cost = max_cost;
+    }
+    auto min_bucket = cost_to_int(min_cost);
+    auto max_bucket = cost_to_int(max_cost);
+
+    if (min_bucket < 0 || max_bucket < 0 || max_bucket > 1000000) {
+        // If min and max are close to each other, assume 3 orders of
```
vaughnbetz: This comment doesn't seem to exactly line up with the code.  Please check the math (it doesn't seem to do what the comment says).

vpr/src/route/bucket.cpp
```diff
@@ -0,0 +1,234 @@
...
+    if (cost < min_cost_ && cost > 0) {
+        min_cost_ = cost;
+        check_scale = true;
+    }
+    if (cost > max_cost_) {
+        max_cost_ = cost;
+        check_scale = true;
+    }
+
+    if (check_scale) {
```
vaughnbetz: Explain that you check_scaling (number of buckets) if the min_cost,max_cost range has changed, to ensure you've got a reasonable number of buckets.

vpr/src/route/bucket.h
```diff
@@ -0,0 +1,229 @@
...
+        return (int)(cost * conv_factor_);
+    }
+
+    // Simple fast random function used for randomizing item selection on pop.
+    size_t fast_rand() {
+        seed_ = (0x234ab32a1 * seed_) ^ (0x12acbade);
+        return seed_;
+    }
+
+    void check_scaling();
```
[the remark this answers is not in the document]
vaughnbetz: Also good to give a general overview of your strategy for heap sizing.  Scale to 1 ps per bucket, with device_grid initial buckets.  You double the heap size whenever you get a cost that is out of range of the buckets.  And you call check_scaling to change the bucket size (but I don't fully follow the philosophy of when).

litghost:
> 2\. --router_algorithm breadth_first (--base_cost_type demand_only should give a similar check, if that's easier)

The breadth_first router cannot currently use the anything other than the `BinaryHeap`

vaughnbetz:
OK, that's fine.  I think it's fine for non-default routers to not use the bucket router. If 1 & 3 could call the bucket heap they should be tested with it though.
The breadth-first router is unloved, but should still have a comment saying it must use (has only been tested with) the binary heap wherever it instantiates it.

vpr/src/route/bucket.cpp
litghost: The comment is valid?  Can you provide a counter example?

vpr/src/route/bucket.h
```diff
@@ -0,0 +1,229 @@
...
+
+  private:
+    /* Vector of all items ever allocated. Used for full item iteration and
+     * for reuse after a `clear` invocation. */
+    std::vector<BucketItem*> heap_items_;
+
+    /* Tracks how many items from heap_items_ are in use. */
+    size_t alloced_items_;
+
+    /* Number of outstanding allocated items. */
```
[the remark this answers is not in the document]
litghost: This does **not** include those on the free list?

vpr/src/route/bucket.h
```diff
@@ -0,0 +1,229 @@
...
+// nets.
+class BucketItems {
+  public:
+    BucketItems() noexcept;
+
+    // Returns all allocated items to be available for allocation.
+    //
+    // This operation is only safe if all outstanding references are discarded.
+    // This is true when the router is starting on a new net, as all outstanding
+    // items should be in the bucket, which is cleared at the start of routing.
```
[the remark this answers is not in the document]
litghost: It does exactly what is says: "Returns all allocated items to be available for allocation"

It then explains why doing such a thing might be a bad idea.  However it is very cheap to do, so if you can meet the requirements, it is significantly faster than managing the free list when preparing for a new route.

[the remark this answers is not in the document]
litghost: Also the BucketItems class comment directly addresses this:
```
  // Supports fast clearing of the items, under the assumption that when clear      
  // is invoked, all outstanding references can be dropped.  This should be true    
  // between net routing, and avoids the need to rebuild the free list between      
  // nets. 
```

vpr/src/route/bucket.cpp
litghost: Done

vpr/src/route/bucket.h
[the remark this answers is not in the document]
litghost: Done

vpr/src/route/bucket.h
```diff
@@ -0,0 +1,229 @@
...
+    size_t fast_rand() {
+        seed_ = (0x234ab32a1 * seed_) ^ (0x12acbade);
+        return seed_;
+    }
+
+    void check_scaling();
+
+    // Expand the number of buckets.
+    //
+    // Only call if insufficient bucets exist.
```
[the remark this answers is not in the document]
litghost: Fixed

litghost:
> All 3 of these cases should be tested in the regtests, so doing a run with the approximate heap turned on for all regtests should also test them.

Done

> It's good to look at their results closely if you just run the regtest though to make sure things didn't blow up.

Golden results are in, take a look.

kmurray:
> I have run QoR on a semi-dedicated VM machines for the baseline version and full template version, and the QoR looks good:
> [baseline_vs_heap.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4299773/baseline_vs_heap.xlsx)

If I understand correctly, this compares the baseline version of VPR with the templated bucket-based heap. However it shows effectively no run-time performance difference (total time: 0.996x, route time 0.9989x). That is much smaller than originally reported in #952, to the point where it may just be machine bounce. 

@litghost Any thoughts about why this is the case?

litghost:
> If I understand correctly, this compares the baseline version of VPR with the templated bucket-based heap.

Sorry, I'm comparing baseline VPR with templated **binary heap** based heap.  For now I'm trying to get the runtime configurable heap stuff landed.  I believe testing and tuning of the bucket heap should be done as follow on work.  Having no change means that changes required to support runtime heap selection had no adverse affects.

kmurray:
> For now I'm trying to get the runtime configurable heap stuff landed. I believe testing and tuning of the bucket heap should be done as follow on work. Having no change means that changes required to support runtime heap selection had no adverse affects.

OK that makes more sense. Thanks for clarifying!

litghost:
QoR results are in showing PR has small or no impact on performance: 
[baseline_vs_template_12.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4325248/baseline_vs_template_12.xlsx)

Per https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1185#issuecomment-597222846, I'd like to merge this as-in, and have it left as future work to tune the bucket heap.  Sounds good?

vpr/src/route/bucket.h
```diff
@@ -0,0 +1,226 @@
+#ifndef _BUCKET_H
+#define _BUCKET_H
+
+#include <vector>
+
+#include "heap_type.h"
+
```
vaughnbetz: Please add a high-level comment on how the bucket sort works, particularly the two levels of allocation (large, reusable memory blocks are pre-allocated and managed, and individually allocated and released out to callers heap items).  Define terms for the two types of references/storage being tracked (suggest pre-allocated and allocated, or maybe group-allocated and fully-allocated).

litghost: Done

vpr/src/route/route_timing.cpp
```diff
@@ -333,6 +390,10 @@ bool try_timing_driven_route(const t_router_opts& router_opts,
         device_ctx.rr_switch_inf,
         route_ctx.rr_node_route_inf);
 
+    // Make sure template type ConnectionRouter is a ConnectionRouterInterface.
+    ConnectionRouterInterface* router_interface = &router;
+    (void)router_interface;
```
kmurray: Looks like this is unused, and can be removed?

litghost: This is effectively "static_assert(is_router drived from ConnectionRouterInterface)"

kmurray: Could we do something more explicit like:
```
static_assert(std::is_base_of<ConnectionRouterInterface, ConnectionRouter>::value, "ConnectionRouter must implement the ConnectionRouterInterface");
```

litghost: Yep, that works.  For some reason I thought it was only available in later C++'s

kmurray:
@litghost Looks good. Once CI is fixed should be good to merge.

PR #1186, opened 2020-03-04
Options to enable faster rr graph reading

litghost:
#### Description

This PR adds two options that enable faster reading of RR graphs from disk.

Option 1 enables the ability to disable calling `check_rr_graph`.  This is takes ~12 seconds on the A50T graph, and if the entire flow is otherwise taking ~60 seconds, this is a significant fraction of runtime.  The idea here is that when the RR graph was first created, `check_rr_graph` can be called, and then once the graph is written to disk, future instance of VPR do not need to call it again.

Option 2 is a two part change.  By default VPR doesn't use edge metadata, so by default the RR graph reader will no longer read edge metadata.  However there are reasons to read the edge metadata (e.g. format conversion, profiling, etc), so an option is being added to re-enable reading of edge metadata from disk.  The default is reasonable, delivering the performance boost, and the option exists to retain a full read if needed.

#### Motivation and Context

This enables a faster RR graph read, and is fully controllable via flags.

Tests have been updated, and symbiflow has been using these flags for a while.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -1673,6 +1673,11 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument<bool, ParseOnOff>(args.read_rr_edge_metadata, "--read_rr_edge_metadata")
+        .help("Read RR edge metadata from --read_rr_graph.  RR edge metadata is not used in core VPR algorithms, and is typically not read to save runtime and memory. (Default: off).")
+        .default_value("off")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
     route_timing_grp.add_argument(args.router_debug_net, "--router_debug_net")
         .help(
             "Controls when router debugging is enabled for nets.\n"
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -1705,6 +1705,11 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("-2")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument<bool, ParseOnOff>(args.check_rr_graph, "--check_rr_graph")
+        .help("Controls whether to check the rr graph when reading from disk.")
+        .default_value("on")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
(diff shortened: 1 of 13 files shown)


vpr/src/base/read_options.cpp
```diff
@@ -1705,6 +1710,11 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("-2")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument<bool, ParseOnOff>(args.disable_check_rr_graph, "--disable_check_rr_graph")
+        .help("Disables checking rr graph when reading from disk.")
+        .default_value("off")
+        .show_in(argparse::ShowIn::HELP_ONLY);
```
kmurray: I'd probably invert this to be an option `--check_rr_graph` which defaults to `on` and applies whenever an RR graph is created (e.g. algorithmically or loaded from disk). That seems more intuitive and easier to explain. End users who don't want to pay the check time when loading from disk can then disable it on the command line.

vpr/src/route/route_common.cpp
```diff
@@ -281,7 +283,9 @@ bool try_route(int width_fac,
                     router_opts.trim_obs_channels,
                     router_opts.clock_modeling,
                     directs, num_directs,
-                    &warning_count);
+                    &warning_count,
+                    router_opts.read_edge_metadata,
+                    router_opts.do_check_rr_graph);
```
kmurray: The argument list of create_rr_graph is getting long. Since many of those options are coming from router_opts, it may make sense to just pass the whole router_opts in directly.

vpr/src/base/read_options.cpp
litghost: Sure, done

vpr/src/route/route_common.cpp
litghost: I'm a fan of keeping things explicit.  If we really need to do this, I believe we should split the rr generator options from the router options, so we can more easily track which of the options are actually used by `create_rr_graph`.

vpr/src/base/read_options.cpp
```diff
@@ -1673,6 +1673,11 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument<bool, ParseOnOff>(args.read_edge_metadata, "--read_edge_metadata")
```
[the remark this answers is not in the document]
litghost: Fixed

kmurray:
Looks good, thanks!

Issue #1045, opened 2019-11-13
Placer awareness of dedicate routing resources


litghost:
#### Proposed Behaviour

Placer should be place instances to take advantage of dedicated routing resources

#### Current Behaviour

Placer only uses delay matrix to choose placement decisions

#### Possible Solution

Open to suggestions

#### Context

This is in the context of the 7-series routing fabric.  There are two classes of placements that are relevant here:
 - Placements where the nets are unroutable
 - Placements where the nets will have to use sub-optimal routing

Some placements will result in unroutable solutions.  Examples:

- BUFIO have exactly one IO pair that can be used.  If an unconstrained BUFIO is instanced, it must be placed on the instance that connects to its IO pad.
- BUFH (horizontal clock buffers) drive into a specific CMT (clock management tiles), and all clocked instances that use that BUFH must be in that CMT, otherwise there is no routing path

Some placements will result in sub-optimal routing (e.g. using general interconnect).  Examples:

 - Some IO pads are CCIO pins, and can connect to BUFG instances (global clock buffers) directly.  If IO pads are not fixed (e.g. prior to PCB design), then some placements will prevent the usage of this dedicate clock path to the BUFG.
- Even if clock IO pads are fixed, PLL/MMCM instances must be located within the same CMT to use the dedicated path from the IO pad to the PLL.

I can give more examples if needed.

FYI @vaughnbetz @mkurc-ant

mithro:
Does marking these as carry chain pack patterns have an effect?

litghost:
> Does marking these as carry chain pack patterns have an effect?

They do not qualify, because they actually have routing.  For example, the CCIO -> PLL connections can route to either CLKIN1 or CLKIN2.

mithro:
Well, you mentioned two types of issues. The first type might be solved in a similar style to dedicate carry chain outputs.

The second is more complicated.

litghost:
> Well, you mentioned two types of issues. The first type might be solved in a similar style to dedicate carry chain outputs.
> 
> The second is more complicated.

Nah, they both are really the same issue.  BUFH to clock in also has routing choices.

mithro:
Can the routing just be moved into a tile and then a pack pattern used to keep these things together?

litghost:
> Can the routing just be moved into a tile and then a pack pattern used to keep these things together?

For the BUFIO, maybe?  For the BUFH and it's ilk, absolutely not.

[...]

Issue #932, opened 2019-08-20
Add Placement Constraints to VPR


vaughnbetz:
Currently VPR can only lock down the placement of I/Os (in the pads file), and there isn't a direct way to control the clusterer with constraints.  

1. We should have a method to lock any primitives to a specific (x,y,z) location on the device (not just I/O pads).  

2. This constraint should affect not only placement, but also packing.  Primitives with the same (x,y,z) should be packed together.  This allows end users to use primitive names (which are more stable) to control placement, rather than post-packing names.

3. We would also like to be able to constrain primitives to a portion of the chip.  Specifying a set of (xmin,ymin) to (xmax,ymax) regions that are legal places to place a primitive accomplishes this.  In the common case there will be 0 rectangles (no constraint) or 1 rectangle, but we should support the union of multiple rectangles as the placement constraint so that more complex floorplans can be created.

#### Constraint options:
The above constraints could be implemented as two separate constraints:  (x,y.z) and region.  
Or they could be one constraint: a set of (xmin,ymin,zmin) to (xmax,ymax,zmax) rectangles.  Placing at a specific (x,y,z) would then be achieved by having xmin=xmax,ymin=ymax,zmin=zmax in the constraint.  I suspect one more general constraint is easier to document, so probably that is best.

#### Possible Solution
1. Constraint file: we could have a constraints.xml file that has a section for placement constraints and specifies these constraints.  An alternative we've discussed is to re-use the .pad file, but that may be confusing as currently that file constrains only the placer, so I think a new constraints.xml file is more clear.

The names of primitives used in this file should support wildcards, so you can match all the primitives in some part of your design easily (key for floorplanning); e.g. my_processor|alu32|*.  There is code in the sdc parser to do this already (vpr/src/timing/read_sdc.cpp).

2. Packer: 

- do not pack things with region constraints that don't intersect together. 
- after packing, check if any regions are too fill (more clusters of a certain type assigned to them than exist in the region).  If there are any such regions, re-try clustering with all primitives in those regions having a high attraction to each other to force a more dense packing of that region.

3. Placer:
- Compute a placement constraint by intersecting the region constraints of all primitives in a cluster together.  Also propagate any resulting placement constraint to other member of a place_macro (carry chain, etc.).
- Initial placement must place only with the legal region for each cluster
- The move generator must not propose (or must immediately reject) any move that violates a region constraint.

4. Placement checking:
[...]

PR #1189, opened 2020-03-04
Fixed non iopads

litghost:
#### Description

This PR was developed by @acomodi.

This PR allows VPR to place clusters other than IO pads, which is useful for placing clock elements explicitly.

#### Related Issue

#1045

#### Motivation and Context

In fabrics with specialized clocking networks, clock elements within the network may have specific requirements.  See #1045 for discussion.  For example, in the 7-series fabric, there are top global buffers and bottom global buffers.  In order to use the dedicated clock path between an input pad and the global clock buffer, if the input pad is in the bottom half of the fabric, it needs to use a "bottom" global buffer, top half of the fabric, it needs to use a "top" global buffer.

Rather than finding a general solution to this problem, for now these elements are given fixed placements *prior* to running VPR via the IO fixing code.  This PR enables that pathway to be more general, and fix any cluster, rather than just IO pads.

Symbiflow has been using this code since December to place non-IO cluster instances.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/read_place.cpp
+++ b/vpr/src/base/read_place.cpp
@@ -146,6 +146,7 @@ void read_user_pad_loc(const char* pad_loc_file) {
     int xtmp, ytmp;
     FILE* fp;
     char buf[vtr::bufsize], bname[vtr::bufsize], *ptr;
+    std::unordered_set<ClusterBlockId> constrained_blocks;
 
     auto& cluster_ctx = g_vpr_ctx.clustering();
     auto& device_ctx = g_vpr_ctx.device();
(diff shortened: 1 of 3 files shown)


kmurray:
As I understand it, this extends the current PAD file support to work for arbitrary block types. Is that correct?

Seems like the long term fix is proper constraint support as noted in #932, and this is a small extension for the interim. Am I following correctly?

litghost:
> As I understand it, this extends the current PAD file support to work for arbitrary block types. Is that correct?

Yes

> Seems like the long term fix is proper constraint support as noted in #932, and this is a small extension for the interim. Am I following correctly?

Sure, but given that I don't believe there is a design for placement constraint, nor an initial prototype, this feature seems like a natural step.  Once a constraint system is in place, this feature will simply be one of the inputs.

PR #1201, opened 2020-03-10
Use RREdgeId in t_heap

litghost:
#### Description

When reviewing https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1197#discussion_r389216334 @kmurray suggested removing `iconn` from the `ConnectionRouter` completely here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1197

I've made that change, but done it here, because removing `iconn` is more involved because of its usage in `t_heap`.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/vpr_types.h
+++ b/vpr/src/base/vpr_types.h
@@ -41,6 +41,7 @@
 #include "vtr_cache.h"
 #include "vtr_string_view.h"
 #include "vtr_dynamic_bitset.h"
+#include "rr_graph_fwd.h"
 
 /*******************************************************************************
  * Global data types and constants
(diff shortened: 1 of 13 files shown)


vpr/src/route/connection_router.cpp
```diff
@@ -431,7 +431,7 @@ void ConnectionRouter::timing_driven_expand_neighbour(t_heap* current,
                        "      Pruned expansion of node %d edge %d -> %d"
                        " (to node location %d,%dx%d,%d outside of expanded"
                        " net bounding box %d,%dx%d,%d)\n",
-                       from_node, from_node_edge_idx, to_node_int,
+                       from_node, from_edge, to_node_int,
```
kmurray: I think here, and a couple of other places we need to be calling `size_t(from_edge)` and using `%zu` in the printf pattern.

vpr/src/base/vpr_types.h
```diff
@@ -1205,7 +1233,7 @@ struct t_trace {
  * occ:        The current occupancy of the associated rr node              */
 struct t_rr_node_route_inf {
     int prev_node;
```
kmurray: In the same vein, we should replace various references to nodes by int index with RRNodeId. Although in some cases like this, since RREdgeIds are unique we may even be able to drop storing the previous node (provided it can be looked up efficiently via the RR graph). Probably best to land this, and do the RRNodeId conversion in another PR.

vpr/src/route/heap_type.h
```diff
@@ -43,11 +43,36 @@ struct t_heap {
 
     struct t_prev {
         int node;
-        int edge;
+        unsigned int edge;
```
kmurray: Ultimately this should be replaceable with just RREdgeId (as noted earlier). Thoughts?

kmurray:
@litghost after resolving the conflicts this should be good to merge!

vpr/src/route/heap_type.h
litghost: Having non-trival types in a union is hard(-ish).  It is possible, but it is really intended for things like a variant type.  I used this approach because it avoided needing to handle non-trival types under the union

vpr/src/base/vpr_types.h
litghost: Ya, a follow on PR to convert all the inode to RRNodeId would be good.  But smaller PR's -> faster to review,  faster to merge is probably better.

vpr/src/route/connection_router.cpp
litghost: Yep, fixed.

litghost:
> @litghost after resolving the conflicts this should be good to merge!

Conflicts resolved, waiting for green CI.

litghost:
> > @litghost after resolving the conflicts this should be good to merge!
> 
> Conflicts resolved, waiting for green CI.

CI is green, merge if acceptable!

PR #931, opened 2019-08-16
Do not use unordered_map over flat rr node vector lookup.

litghost:
#### Description

Use of unordered_map consumes ~10% of routing time on designs with
non-configurable edges.

#### Motivation and Context

Profiled design with non-configurable edges, and found unordered_map lookahead was taking ~10% of time.

#### Types of changes

- [x] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1206, opened 2020-03-11
Encapsulate non-configurable node sets

litghost:
#### Description

This PR builds on #1204 and encapsulates non-configurable node sets behind `t_rr_graph_storage`. 

#### Motivation and Context

This PR encapsulates non-configurable node sets behind `t_rr_graph_storage` for a couple reasons:
 - Construction of non-configurable node sets is not optional, and should be handled near `t_rr_graph_storage::partition_edges`
 - Raw access to the underlying non-configurable node set storage made some PR's (e.g. #931) harder to do because of distributed logic
 - Formalized the idea of the non-configurable node set id, which already existed, but was just a `size_t` in the previous iteration

 - [ ] Existing tests that use non-configurable nodes pass
 - [ ] Adds non-configurable node set verification

#### Types of changes

#### Checklist:


vpr/src/route/rr_graph_storage.cpp
```diff
@@ -626,3 +631,432 @@ void t_rr_graph_storage::set_node_side(RRNodeId id, e_side new_side) {
...
+RRNonConfigurableSetId t_rr_graph_view::non_configurable_set_id(RRNodeId id) const {
+    auto itr = rr_node_to_non_config_node_set_.find(id);
+    if (itr != rr_node_to_non_config_node_set_.end()) {
+        return itr->second;
+    } else {
+        return RRNonConfigurableSetId::INVALID();
+    }
+}
+
+vtr::array_view<const RRNodeId> t_rr_graph_view::get_non_configurable_set(RRNonConfigurableSetId id) const {
```
vaughnbetz: Kevin's suggestion: use free functions that both rr_graph_storage and rr_graph_view call to reduce code duplication here (and similar functions).

vpr/src/route/rr_graph_storage.h
```diff
@@ -584,6 +617,14 @@ class t_rr_graph_storage {
     vtr::vector<RREdgeId, RRNodeId> edge_dest_node_;
     vtr::vector<RREdgeId, short> edge_switch_;
 
+    // Sets of non-configurably connected nodes
```
vaughnbetz: Need a comment on what these non-configurable edge sets are.  May go here, or at top of rr_graph_storage (with other top-level rr_graph comments).
Non-configurable edges are switches in the rr-graph that cannot be turned off. They may be metal shorts, or non-tri-stateable buffers: in both cases the switch always connects two nodes, and cannot be turned off no matter what the router does.
These non-configurable edges are useful for describing L and T-shaped wires as the rr_graph can only store X- or Y- directed (straight) wires.  Using non-configurable edges lets us model metal shorts to describe the L or T-shaped wire. Non-configurable edges can also occur with non-tristatable buffers, which can be used in clock networks or on wires so long that they have internal buffering that can't be turned off.
If the router uses one member of a non-configurable node set (i.e. rr_nodes connected with non-configurable edges) then it has implicitly used them all, as the others are directly wired to the one used.

... Keith: add explanation on exactly what is in non_config_set_nodes_ and non_config_first_node_idx_

litghost: Done

PR #1207, opened 2020-03-11
Add penalty cost attribute to switches

acomodi:
#### Description

This PR adds an additional attribute to the switches, namely the `penalty_cost`. There are some situations for which the router chooses some paths that are allowed, but only for special reasons, and shouldn't be generally chosen.

To give an example, regarding a known SymbiFlow issue, the clock nets should always be routed through the dedicated clock network.
The architecture, though, is designed in a way that the clock nets can "escape" from the clock network through a switch and enter in the general interconnect, resulting in a possible increase in the clock skew and timing violations.
It is necessary though to have the switch, as some logic designs still should have the freedom (even though it is highly discouraged), to use the clock net within logic cells. If the problematic switch that allows clock nets to enter general interconnect were removed, such designs would not be implementable.

To limit the issue, a `penalty_cost` is added with this PR. This cost is not related to any physical property, but is a way for the architecture, to instruct the router to avoid using the problematic switch if there is no need to use it.
This has the benefit of not compromising the timing analysis, as the penalty cost is not taken into account when running STA.

vtr_reg_basic
vtr_reg_strong

#### Types of changes

#### Checklist:

### DIFF
--- a/doc/src/arch/reference.rst
+++ b/doc/src/arch/reference.rst
@@ -631,7 +631,7 @@ The tags within the ``<switchlist>`` tag specifies the switches used to connect 
 
     .. code-block:: xml
 
-        <switch type="mux" name="my_awesome_mux" R="551" Cin=".77e-15" Cout="4e-15" Cinternal="5e-15" Tdel="58e-12" mux_trans_size="2.630740" buf_size="27.645901"/>
+        <switch type="mux" name="my_awesome_mux" R="551" Cin=".77e-15" Cout="4e-15" Cinternal="5e-15" Tdel="58e-12" penalty_cost="1e-8" mux_trans_size="2.630740" buf_size="27.645901"/>
 
 
     :req_param type:
@@ -698,6 +698,14 @@ The tags within the ``<switchlist>`` tag specifies the switches used to connect 
         .. note:: Required if no ``<Tdel>`` tags are specified
 
         .. note:: A ``<switch>``'s resistance (``R``) and output capacitance (``Cout``) have no effect on delay when used for the input connection block, since VPR does not model the resistance/capacitance of block internal wires.
+
+    :opt_param penalty_cost:
+
+        Penalty cost added to the switch.
+        This cost is not related to the physical properties of the switch. It adds an additional cost to it for the router to avoid its usage.
+        This can be useful for switches that should not be used in general, but that are necessary for certain types of designs that need to route through them.
+
+        .. note:: The addition of this attribute to the switch does not affect Timing Analysis
 
     :opt_param buf_size:
 
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1384,6 +1384,7 @@ struct t_arch_switch_inf {
     float Cin = 0.;
     float Cout = 0.;
     float Cinternal = 0.;
+    float penalty_cost = 0.;
     float mux_trans_size = 1.;
     BufferSize buf_size_type = BufferSize::AUTO;
     float buf_size = 0.;
@@ -1448,6 +1449,7 @@ struct t_rr_switch_inf {
     float Cout = 0.;
     float Cinternal = 0.;
     float Tdel = 0.;
+    float penalty_cost = 0.;
     float mux_trans_size = 0.;
     float buf_size = 0.;
     const char* name = nullptr;
--- a/libs/libarchfpga/src/read_xml_arch_file.cpp
+++ b/libs/libarchfpga/src/read_xml_arch_file.cpp
@@ -3779,23 +3779,23 @@ static void ProcessSwitches(pugi::xml_node Parent,
         SwitchType type = SwitchType::MUX;
         if (0 == strcmp(type_name, "mux")) {
             type = SwitchType::MUX;
-            expect_only_attributes(Node, {"type", "name", "R", "Cin", "Cout", "Cinternal", "Tdel", "buf_size", "power_buf_size", "mux_trans_size"}, " with type '"s + type_name + "'"s, loc_data);
+            expect_only_attributes(Node, {"type", "name", "R", "Cin", "Cout", "Cinternal", "Tdel", "penalty_cost", "buf_size", "power_buf_size", "mux_trans_size"}, " with type '"s + type_name + "'"s, loc_data);
 
         } else if (0 == strcmp(type_name, "tristate")) {
             type = SwitchType::TRISTATE;
-            expect_only_attributes(Node, {"type", "name", "R", "Cin", "Cout", "Cinternal", "Tdel", "buf_size", "power_buf_size"}, " with type '"s + type_name + "'"s, loc_data);
+            expect_only_attributes(Node, {"type", "name", "R", "Cin", "Cout", "Cinternal", "Tdel", "penalty_cost", "buf_size", "power_buf_size"}, " with type '"s + type_name + "'"s, loc_data);
 
         } else if (0 == strcmp(type_name, "buffer")) {
             type = SwitchType::BUFFER;
-            expect_only_attributes(Node, {"type", "name", "R", "Cin", "Cout", "Tdel", "buf_size", "power_buf_size"}, " with type '"s + type_name + "'"s, loc_data);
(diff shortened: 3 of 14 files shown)


kmurray:
It seems to me, the fundamental issue is the router is not choosing the desirable path (i.e. staying on the clock network).

While this is one approach for tackling this (presumably by making clock network -> general routing connection expensive). I wonder what other approaches have been considered? For instance, what about improving the router lookahead, so it knows staying on the clock network is the most efficient (i.e. lowest cost) approach?

An additional concern is that adding this option starts to mix CAD concerns into the architecture description which is somewhat undesirable. (Ideally the architecture description would focus only on the architecture description and leave it up to the tool to do the right thing based upon it). Furthermore, how should the value of the cost penalty be calculated? Leaving it up to the user to set a feature which impacts the deep inner workings of the tool is likely to be highly non-intuitive. Is there perhaps some way which an appropriate cost penalty could be determined automatically by VPR?

acomodi:
> I wonder what other approaches have been considered? For instance, what about improving the router lookahead, so it knows staying on the clock network is the most efficient (i.e. lowest cost) approach?

@kmurray Actually this same problem is currently managed in SymbiFlow with the penalty cost addition and this already merged PR: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1166.
One thing that is to be tested still is whether the penalty cost is necessary. I will run tests to verify that https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1166 is actually enough to solve the issue.

I failed to mention also other cases in which the penalty cost is necessary. For instance, there are situations for which the clock net gets out from a BUFH, traverses the whole clock region just to come back at same node on a different BUFH of the same tile.

As regards the second comment, I believe that the penalty cost is still something that adds an additional information about the architecture itself, namely that some connections are definitely undesirable w.r.t to others, but still needed. This is something that, IMO, is left to the architect designer to choose, not the CAD tool itself. This is why, the cost value, can be defined by the user only, after some experiments. The values that are applied in SymbiFlow have been tuned after some testing.

Still, I will need to test whether this behavior can be avoided only by using https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1166.

kmurray:
> Still, I will need to test whether this behavior can be avoided only by using #1166.

Makes sense. Lets see if that is sufficient.

acomodi:
The following error is generated when we have situation in which the clock nets are being routed through the general interconnect:
```
ERROR: [Place 30-176] Unroutable Placement! The following clock source instance is driving the following locked load instances. The clock source instance is placed too far away from its load instance to be routable.
        CLK_HROW_TOP_R_X60Y78_BUFHCE_X1Y18_BUFHCE (BUFHCE.O) is locked to BUFHCE_X1Y18
        The loads are distributed to 1 user pblock constraints. In addition, there are 20 loads not in user pblock constraints.

        Displaying only the first 20 or fewer instances under each constraint as list of loads is too long
```

Unfortunately, this means that the base cost addition is necessary and https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1166 is not enough to mitigate the problem.

I am still running the QoR tests to check what the impact of these changes is.

acomodi:
Some QoR came back and do not look too promising. Here is the comparisons file: [comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4330350/comparison.xlsx)

litghost:
> Some QoR came back and do not look too promising. Here is the comparisons file: [comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4330350/comparison.xlsx)

This all looks to short running examples.  Compare the `titan_quick_qor` from vtr_reg_nightly for a better comparision.

acomodi:
@litghost Yes, I started those builds earlier today, should have results soon

acomodi:
Got quick titan benchmark results: [titan_quick_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4342549/titan_quick_qor_comparison.xlsx)

kmurray:
Run-time penalty looks to be fairly high: +5% route time, +2% total time

We should consider possible ways to mitigate that if possible.

acomodi:
@kmurray Yeah, we should definitely cut run-time down. Any idea on how to tackle this?

vaughnbetz:
Hacky solution:  increase the delay of that switch type.  That will mess up timing analysis which is not great.  You could change the delay back at the end of the router (a bit better, but still not great).

Better solution: add a zero-length wire (mux node) that breaks this edge into two pieces.  Put a high base cost on that node.  Requires changing the rr-graph generation code in symbiflow.

kmurray:
> Better solution: add a zero-length wire (mux node) that breaks this edge into two pieces. Put a high base cost on that node. Requires changing the rr-graph generation code in symbiflow.

Yep, that sounds like the right approach. It would eliminate the run-time cost for every edge in the RR graph, and instead we would only pay to explore the extra node when it was explored (which would be rarely).

acomodi:
@kmurray @vaughnbetz As I understand it, this change would not require to merge this PR, correct?

The fact is that I need to check how complex could it turn out to be modifying the rr_graph generation mechanism in SymbiFlow.

Furthermore, this is my understanding of the solution, correct me if that's wrong:

Currently there is a situation like this: 
```
Node --> Edge (relative to a penalty cost switch) --> Node (that lets into general interconnect)
```

Solution:
```
Node --> Edge --> Node (zero-length mux wire with high base cost) --> Edge --> Node (that lets into general interconnect)
```

My questions are: 
- Where to define the node's base cost?
- How to set it to have zero length?

vaughnbetz:
The node length would be defined in the rr_graph.xml.  It doesn't have to be 0 length; just makes it more obvious that it's not a real wire.  What you currently do for "bounce" paths etc. is probably what you want to do for this one.
I don't think we expose the base_cost in the rr_graph.xml, so we'd need to add the optional ability to specify a base_cost_multipier (I would do it as a multiplier) in the rr_graph file & import.

kmurray:
> I don't think we expose the base_cost in the rr_graph.xml, so we'd need to add the optional ability to specify a base_cost_multipier (I would do it as a multiplier) in the rr_graph file & import.

This would show up in the arch.xml as a new segment type. So I'd suggest adding a base_cost_multiplier parameter to the architecture segment definition, instead of in the rr_graph.xml.

In order to work with the existing base cost command line options to VPR the base_cost_multiplier option should default to an 'auto' setting which lets VPR decide the base cost (i.e. for regular wires).

acomodi:
@kmurray @vaughnbetz I think I get what needs to be done and I will summarize as follows:

- Add a node to break the "problematic" edge;
- Add a segment to the architecture that has a `base_cost_multiplier`;
- Assign the segment to the new node;
- Add another `base_cost_type` parameter (e.g. `delay_base_cost_multiplier`) that will work as the `default` and, additionally multiplies the base_cost for nodes with segments that have a base_cost_multiplier attribute set.

Does this sound correct?

kmurray:
>     * Add another `base_cost_type` parameter (e.g. `delay_base_cost_multiplier`) that will work as the `default` and, additionally multiplies the base_cost for nodes with segments that have a base_cost_multiplier attribute set.

It all sounds good except the above. The way I'd envision this working is something like this:

```xml
    <!-- Implicitly using auto base costs -->
    <segment name="L4" freq="260" length="4" type="unidir" Rmetal="201.7" Cmetal="18.0e-15"> 

    <!-- Explicitly using auto base costs (equivalent to above) -->
    <segment name="L4" freq="260" length="4" type="unidir" Rmetal="201.7" Cmetal="18.0e-15" base_cost_multiplier="auto">

    <!-- Explicit non-auto base cost -->
    <segment name="clk2route" freq="1" length="0" type="unidir" Rmetal="0" Cmetal="0" base_cost_multiplier="10">
```
where the first two methods would result in the current behavior, where VPR will determine their base costs based on the `--base_cost_type` command line option.

In the last case the architect is overriding the base cost to be some other multiple of the unit base cost (where the value of a unit base cost is still determined by VPR).

vaughnbetz:
Yes, this is what I was thinking as well (I'm in line with Kevin's modification).

acomodi:
Penalty cost is not required anymore. Closing

PR #1208, opened 2020-03-11
Add Heterogeneous Tiles capability

acomodi:
#### Description

This PR is an initial development towards the solution to have heterogeneous tiles capability.
There are currently two types of situations that VPR cannot handle:

1. Capacity tiles that need specific pin_locations for each of the `z` location in which the tile is placed. As a concrete example, the BUFG tile in the Series7 comprehends 16 BUFGCTRL sites, each with the same I/Os and functionality. They could be expressed with a single tile, with capacity of 16. The issue though is that the pin_locations of some of the BUFGCTRL is not equal to the others (as visible in the image below).
This PR is an initial step towards the solution of this issue.

Untitled Diagram (1)

2. Heterogeneous tiles: there are a few kind of special tiles (still in the 7series) that comprehend different kinds of sites with different types of I/Os and different functionalities. A simple example of tiles composed by heterogeneous resources is the following:

Untitled Diagram

The final solution would ideally solve both of the above mentioned situations, as we could treat the BUFGCTRL sites not homogeneously as it happens now, but as an heterogeneous set, composed, with each element of the set having different pin_locations

#### Related Issue

This PR is to resume the discussion started here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1063

#### Types of changes

#### Checklist:


acomodi:
@litghost @kmurray @vaughnbetz FYI

kmurray:
Thanks @acomodi for the diagrams. They are helpful for grasping the issue, although I still don't think I fully comprehend the current limitations.

It seems to me like these options should simply be encoded as options within the pb_type at that X/Y location.

For example, in case 1, can you comment on why putting the 16 bufg intances within a single pb_type and using custom pin locations is insufficient?
```xml
<pb_type name="bufg_16">
    <input name="bufg1_in"/>
    <input name="bufg2_in"/>
    ....
    <input name="bufg16_in"/>
    <!--similarly for outputs-->


    <pb_type name="bufg" num_pb="16">
        ...
    </pb_type>
    ....
</pb_type>

<tile name="bufg_tile"  width="1" height="4">
...
        <loc side="left" y_offset="0"> bufg_16.bufg1_in bufg_16.bufg2_in bufg_16.bufg3_in bufg_16.bufg4_in</loc>
        <loc side="left" y_offset="1"> bufg_16.bufg5_in bufg_16.bufg6_in bufg_16.bufg7_in bufg_16.bufg8_in</loc>
        ...
        <loc side="left" y_offset="3"> bufg_16.bufg13_in bufg_16.bufg14_in bufg_16.bufg15_in bufg_16.bufg16_in</loc>
    </pinlocations>
</tile>
```

acomodi:
@kmurray Sure, probably I should have explained a bit more on the BUFG issue.

The fact is that we need to have the freedom to choose where a certain BUFG instance is going to be placed. This because the 7series for instance require special treatment of clock resources (e.g. BUFG and PLL must be in the same clock region).
To achieve this, each BUFG is a separate entity that can be placed in one of the 16 placement locations.

Untitled Diagram (2)

litghost:
> For example, in case 1, can you comment on why putting the 16 bufg intances within a single pb_type and using custom pin locations is insufficient?

This is because that defines **one** cluster, rather than **sixteen** clusters.  In the 7-series arch, there are two BUFGCTRL tiles, a **top** and a **bottom**.  Some BUFG's **must** be in the top or bottom, some BUFG can be in either.  In order to allow for the placer (or fixed placement constraints) to work, each BUFGCTRL must be in its own cluster, so they can be placed independently.

kmurray:
OK, so the desire is to support different pb_types (which may possibly be equivalent with other pb_types) at each 'capacity' location within a specific tile (i.e. physical routing interface). Is that correct?

litghost:
> OK, so the desire is to support different pb_types (which may possibly be equivalent with other pb_types) at each 'capacity' location within a specific tile (i.e. physical routing interface). Is that correct?

Yes.

vaughnbetz:
Discussion March 12:
- General agreement that we want to be able to store heterogenous blocks at an (x,y).  Alessandro to make a document proposing the architecture file representation and the grid representation.  Vaughn's preference: 
- grid has a type / physical data for (x,y,index) values, where index at a certain x,y goes from 0 to capacity-1.  This lets us have arbitrary heterogeneous blocks.
- We also want to rename z to something else (blk_index? sub_block?) but that can be a separate update or merged into this discussion as appropriate.

acomodi:
@litghost @kmurray @vaughnbetz @HackerFoo FYI, document is here: https://docs.google.com/document/d/1UUfPxmzcroBoBToi-s5xJqJK9YUIgeinU8ZPStR0r9M/edit#

acomodi:
@kmurray @vaughnbetz Ping

acomodi:
@litghost @kmurray @vaughnbetz FYI, I have started implementing the Heterogeneous tiles following the [document's](https://docs.google.com/document/d/1UUfPxmzcroBoBToi-s5xJqJK9YUIgeinU8ZPStR0r9M) proposal.

DONE:
- Pin Locations and Pin Classes have been refactored to have a more C++ style. I have changed it  them to use vectors instead of plain arrays. This results in fixing things in multiple places.
- Added SubTile parsing capability from the XML

TODO:
- [ ] Fix compilation errors due to the pin locations and pin classes refactor
- [ ] Adjust placement to use sub tiles locations
- [ ] Change regression tests architecture to reflect the sub tile addition
- [ ] Test against symbiflow architectures

libs/libarchfpga/src/physical_types.h
```diff
@@ -582,21 +583,15 @@ struct t_physical_tile_type {
     int width = 0;
     int height = 0;
 
-    bool**** pinloc = nullptr; /* [0..width-1][0..height-1][0..3][0..num_pins-1] */
+    std::vector<std::vector<std::vector<std::vector<bool>>>> pinloc; /* [0..width-1][0..height-1][0..3][0..num_pins-1] */
```
kmurray: Might consider using vtr::Matrix<std::vector<bool>> here (since the block width/height should be known when building the tile).

libs/libarchfpga/src/physical_types.h
```diff
@@ -611,6 +606,8 @@ struct t_physical_tile_type {
 
     int index = -1; /* index of type descriptor in array (allows for index referencing) */
 
+    std::vector<t_sub_tile> sub_tiles;
+
     std::vector<t_logical_block_type_ptr> equivalent_sites;
```
kmurray: Probably on the TODO list, but it would seem that this should be removed since it's now in the sub-tiles.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -72,26 +72,56 @@ struct t_fc_override {
...
 
+struct t_pin_counts {
+    int total = 0;
+    int input = 0;
+    int output = 0;
+    int clock = 0;
+
+    void sum() {
+        total = input + output + clock;
+    }
```
kmurray: I'd probably just have:
```
    void total() {
        return input + output + clock;
    }
```
which reduces the amount of state we carry around (e.g. don't need to worry about whether sum() has been called)

vpr/src/place/timing_place_lookup.cpp
```diff
@@ -220,22 +220,24 @@ std::vector<int> get_best_classes(enum e_pin_type pintype, t_physical_tile_type_
      */
 
     std::vector<int> best_classes;
-
+    int best_class_index = 0;
     //Collect all classes of matching type which do not have all their pins ignored
-    for (int i = 0; i < type->num_class; i++) {
-        if (type->class_inf[i].type == pintype) {
+    for (auto class_inf : type->class_inf) {
```
kmurray: Looks like this will be copying the class_inf, probably want to use `const auto&` to keep the same reference only behavior of the previous code.

vpr/src/util/vpr_utils.cpp
```diff
@@ -221,17 +223,18 @@ std::string block_type_pin_index_to_name(t_physical_tile_type_ptr type, int pin_
...
     std::string pin_name = type->name;
 
-    if (type->capacity > 1) {
-        int pins_per_inst = type->num_pins / type->capacity;
-        int inst_num = pin_index / pins_per_inst;
-        pin_index %= pins_per_inst;
+    int sub_tile_index = 0;
+    auto logical_pin = type->tile_block_pin_directs_map(physical_pin(pin_index));
+    sub_tile_index = logical_pin.sub_tile_index;
+    pin_index = logical_pin.pin
```
kmurray: This code shows up in a couple places. We should probably put it into a re-usable function which does this lookup.

acomodi: I just realized that this piece of code actually won't work. The `directs_map` needs a logical block, which is not provided.

The problem here is to calculate the correct sub_tile based on a physical pin index. I think an additional member to the sub_tile defining the pin ranges so to easily calculate in which sub_tile the pin_index is located (and also change the math needed to get the pin index in the capacity instance.

vpr/src/place/timing_place_lookup.cpp
acomodi: Done

libs/libarchfpga/src/read_xml_arch_file.cpp
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done. I have actually used an `NdMatrix` of dimension 3 as also the number of sides is known.

acomodi:
@kmurray I should have solved all the strong and basic regression tests issues (they do pass on my local machine). I believe this PR is now ready for a first review pass.

What is still needed though is correctly picking a sub tile coordinate that is compatible with the logic block type during placement.

acomodi:
@kmurray I believe all the issues have been addressed and solved.

Currently what is left to do to have a complete PR is to add the relative documentation for the sub tiles usage (maybe even a tutorial similar to the `equivalent sites` one).

acomodi:
@kmurray I have also temporarily reverted the change on the architecture XMLs so to make code review easier. I will remove the revert commit once all the future review comments have been addressed.

vpr/src/place/move_utils.cpp
```diff
@@ -632,9 +637,17 @@ bool find_to_loc_uniform(t_logical_block_type_ptr type,
...
+    //a z location within a compatible sub tile.
+    std::vector<int> compatible_sub_tiles;
+    for (int capacity = 0; capacity < to_type->capacity; capacity++) {
+        t_pl_loc loc(to.x, to.y, capacity);
+
+        if (is_tile_compatible(to_type, type, loc)) {
+            compatible_sub_tiles.push_back(capacity);
+        }
+    }
+    to.sub_tile = compatible_sub_tiles[vtr::irand((int)compatible_sub_tiles.size() - 1)];
```
kmurray, line 650: This is allocating a a std::vector deep in the placer's inner loop, which is likely to be a significant performance issue. 

Probably we'll need to pull this out and pre-determine what sub-tile locations are compatible with each logical type. I expect this will end-up in the compressed grid data structures.

vpr/src/route/rr_graph.cpp
```diff
@@ -1396,10 +1396,10 @@ static void build_rr_sinks_sources(const int i,
         return;
 
     auto type = grid[i][j].type;
-    int num_class = type->num_class;
-    t_class* class_inf = type->class_inf;
+    int num_class = (int)type->class_inf.size();
+    std::vector<t_class> class_inf = type->class_inf;
     int num_pins = type->num_pins;
-    int* pin_class = type->pin_class;
+    std::vector<int> pin_class = type->pin_class;
```
kmurray, line 1402: Minor, but we don't need to copy these. Should be able to take as `const &`

vpr/src/route/rr_graph2.cpp
```diff
@@ -1454,7 +1454,7 @@ int get_rr_node_index(const t_rr_node_indices& L_rr_node_indices,
      * and ptc gives the number of this resource.  ptc is the class number,
      * pin number or track number, depending on what type of resource this
      * is.  All ptcs start at 0 and go up to pins_per_clb-1 or the equivalent.
-     * There are type->num_class SOURCEs + SINKs, type->num_pins IPINs + OPINs,
+     * There are (int)type->class_inf.size() SOURCEs + SINKs, type->num_pins IPINs + OPINs,
```
kmurray, line 1457: Looks like search and replace, we don't the cast in the comments :)

vpr/src/util/vpr_utils.h
```diff
@@ -30,9 +30,14 @@ bool is_empty_type(t_logical_block_type_ptr type);
...
 
+//Returns the sub tile corresponding to the logical block location within a physical type
+int get_sub_tile_index(ClusterBlockId blk);
+
 int get_unique_pb_graph_node_id(const t_pb_graph_node* pb_graph_node);
 
-void get_class_range_for_block(const ClusterBlockId blk_id, int* class_low, int* class_high);
+void get_class_range_for_block(const ClusterBlockId blk_id,
+                               int* class_low,
+                               int* class_high);
```
kmurray, line 40: I saw there was a t_class_range. Should that be used here?

vpr/src/util/vpr_utils.h
```diff
@@ -146,6 +148,8 @@ int get_max_num_pins(t_logical_block_type_ptr logical_block);
 
 bool is_tile_compatible(t_physical_tile_type_ptr physical_tile, t_logical_block_type_ptr logical_block);
 
+bool is_tile_compatible(t_physical_tile_type_ptr physical_tile, t_logical_block_type_ptr logical_block, t_pl_loc loc);
```
kmurray, line 151: I expect the t_pl_loc is needed to see which capacity location the logical block is in.

If that is the case, we probably don't need the physical_tile as well (as it is also determined by the loc).

It may make sense to factor this out so there is a:
```cpp
bool is_subtile_compatible(t_sub_tile sub_tile, t_logical_block_type_ptr logical_block);
```
which is_tile_compatible() calls internally.

Thoughts?

vpr/src/util/vpr_utils.cpp
```diff
@@ -2184,10 +2217,41 @@ int get_logical_pin(t_physical_tile_type_ptr physical_tile,
     return result->second.pin;
 }
 
+int get_physical_pin(int sub_tile_index,
+                     t_physical_tile_type_ptr physical_tile,
+                     t_logical_block_type_ptr logical_block,
+                     int pin) {
```
kmurray, line 2223: What is the difference between these two versions of get_physical_pins()?

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_sub_tiles/config/config.txt
```diff
@@ -0,0 +1,28 @@
+##############################################
```
kmurray, line 1: Thanks for adding the test case. However I don't see the associated golden_results.txt file in the PR. It should be included (you can generate it with parse_vtr_task.pl -create_golden after running the task).

vpr/src/place/move_utils.cpp
acomodi: As suggested I have moved the creation of the vector during create_compressed_block_grids.

To do so an unordered map is necessary, as it may be possible that a logical block is can be placed in different tiles with different capacities, hence, the sub tile index depends not only on the logical block, but also on the physical tile chosen for the swap.

This should reduce the impact on run time, and I will run the qor benchmarks to verify this.
If the performance hit is too much, we could just let the move generator produce a random sub tile index which is checked at a later stage out of the inner loop.

vpr/src/route/rr_graph.cpp
acomodi: Done

vpr/src/util/vpr_utils.h
acomodi: Done

vpr/src/util/vpr_utils.h
acomodi: This function was actually thought to be used in cases where the placement is not yet completed, to check whether specific locations were compatible (e.g. read_place.cpp, move_utils.cpp).

It works the other way around, meaning that, first, there is a check to test whether a specific sub tile index is compatible within a physical tile, and then, the logic block itself is checked as well, to see if that is compatible with the physical tile (by calling the original `is_tile_compatible` function.

vpr/src/util/vpr_utils.cpp
acomodi: Basically, one is the superset of the other.

I have modified as follows;
- The one with the sub_tile_index parameter is the one that performs the lookup to get the physical pin.
- The other homonym function, as first thing gets the sub tile index based on which sub tile is compatible with the logical block, then it calls the `get_physical_pin` with the sub_tile_index parameter.

This was done mainly because now, in the `logical pin <-> physical pin` mapping there is the sub tile concept introduced. The logical pin now has a sub tile index.

During clustering there is the need to access the physical pin related to a logical one, but it is not important which exact sub tile is needed, as the physical pin is needed only to check the type of pin for instance.

The only important place where the sub tile index needs to be the correct one, is during the block sync at the end of placement, where the "real" physical pins are mapped against the logical pins.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_sub_tiles/config/config.txt
acomodi: Done

vpr/src/route/rr_graph2.cpp
acomodi: Done

acomodi:
@kmurray @vaughnbetz Titan quick QoR runs finished. Results seem to reasonably good. There is an increase of:
- `0.7%` in the whole flow;
- `0.6%` in place time.

[heterogeneous_qor_results.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4434488/heterogeneous_qor_results.xlsx)

kmurray:
@acomodi Those QoR results look good. No quality change with very small overall run-time changes (likely in the measurement noise).

One thing that might be worth looking into is the route time on sparcT2_core which is +45% vs the baseline (an outlier). It would be good to understand what changed here (since the appears identical). My usual approach for this kind of thing is to diff the two log file.

acomodi:
@kmurray All right, I will give a look to the log files and, in case I will re-run the QoR tests.

In addition, I have updated the PR with the tutorial and I have adjusted some function names as well.

acomodi:
@kmurray By looking at the routing difference between the two runs on the SparcT2 I noticed that the values are the same in both the runs, only thing that changes is runtime for each iteration:

```
-## Initializing router criticalities took 0.04 seconds (max_rss 4633.8 MiB, delta_rss +0.0 MiB)
+## Initializing router criticalities took 0.06 seconds (max_rss 4637.8 MiB, delta_rss +0.0 MiB)
 ---- ------ ------- ---- ------- ------- ------- ----------------- --------------- -------- ---------- ---------- ---------- ---------- --------
 Iter   Time    pres  BBs    Heap  Re-Rtd  Re-Rtd Overused RR Nodes      Wirelength      CPD       sTNS       sWNS       hTNS       hWNS Est Succ
       (sec)     fac Updt    push    Nets   Conns                                       (ns)       (ns)       (ns)       (ns)       (ns)     Iter
 ---- ------ ------- ---- ------- ------- ------- ----------------- --------------- -------- ---------- ---------- ---------- ---------- --------
-   1   48.2     0.0    0 5.6e+08  184710  578353  394846 ( 5.498%) 5745015 (56.7%)   11.235 -8.036e+05    -10.235      0.000      0.000      N/A
-   2   32.9     0.5  960 3.9e+08  166480  552340  114049 ( 1.588%) 4446552 (43.9%)   11.232 -1.043e+06    -10.232      0.000      0.000      N/A
-   3   25.5     0.6  249 2.6e+08   80355  354612   88016 ( 1.226%) 4457032 (44.0%)   11.245 -1.006e+06    -10.245      0.000      0.000      N/A
-   4   26.4     0.8  166 2.4e+08   66903  304138   70788 ( 0.986%) 4464751 (44.1%)   11.231 -1.032e+06    -10.231      0.000      0.000      N/A
-   5   26.9     1.1  189 2.1e+08   54632  254941   49503 ( 0.689%) 4487157 (44.3%)   11.231 -1.033e+06    -10.231      0.000      0.000      N/A
-   6   24.5     1.4  164 1.8e+08   41359  196479   33618 ( 0.468%) 4512706 (44.5%)   11.231 -1.041e+06    -10.231      0.000      0.000      N/A
-   7   21.6     1.9  187 1.4e+08   29419  143285   21080 ( 0.294%) 4542463 (44.8%)   11.231 -1.050e+06    -10.231      0.000      0.000      N/A
-   8   18.4     2.4  146 9.8e+07   18965   92562   11597 ( 0.161%) 4568790 (45.1%)   11.231 -1.055e+06    -10.231      0.000      0.000      N/A
-   9   14.0     3.1  122 6.3e+07   11086   49752    5664 ( 0.079%) 4590949 (45.3%)   11.231 -1.066e+06    -10.231      0.000      0.000      N/A
-  10   10.9     4.1   95 3.9e+07    5956   24818    2523 ( 0.035%) 4605736 (45.4%)   11.231 -1.057e+06    -10.231      0.000      0.000       25
-  11    8.5     5.3   97 2.3e+07    3035   12314    1010 ( 0.014%) 4614739 (45.5%)   11.231 -1.057e+06    -10.231      0.000      0.000       22
-  12    6.3     6.9   49 1.2e+07    1331    5279     404 ( 0.006%) 4620422 (45.6%)   11.231 -1.057e+06    -10.231      0.000      0.000       21
-  13    5.0     9.0   33 6030341     542    2107     155 ( 0.002%) 4622679 (45.6%)   11.231 -1.058e+06    -10.231      0.000      0.000       20
+   1   71.8     0.0    0 5.6e+08  184710  578353  394846 ( 5.498%) 5745015 (56.7%)   11.235 -8.036e+05    -10.235      0.000      0.000      N/A
+   2   49.4     0.5  960 3.9e+08  166480  552340  114049 ( 1.588%) 4446552 (43.9%)   11.232 -1.043e+06    -10.232      0.000      0.000      N/A
+   3   38.1     0.6  249 2.6e+08   80355  354612   88016 ( 1.226%) 4457032 (44.0%)   11.245 -1.006e+06    -10.245      0.000      0.000      N/A
+   4   39.7     0.8  166 2.4e+08   66903  304138   70788 ( 0.986%) 4464751 (44.1%)   11.231 -1.032e+06    -10.231      0.000      0.000      N/A
+   5   40.2     1.1  189 2.1e+08   54632  254941   49503 ( 0.689%) 4487157 (44.3%)   11.231 -1.033e+06    -10.231      0.000      0.000      N/A
+   6   36.7     1.4  164 1.8e+08   41359  196479   33618 ( 0.468%) 4512706 (44.5%)   11.231 -1.041e+06    -10.231      0.000      0.000      N/A
+   7   32.4     1.9  187 1.4e+08   29419  143285   21080 ( 0.294%) 4542463 (44.8%)   11.231 -1.050e+06    -10.231      0.000      0.000      N/A
+   8   27.6     2.4  146 9.8e+07   18965   92562   11597 ( 0.161%) 4568790 (45.1%)   11.231 -1.055e+06    -10.231      0.000      0.000      N/A
+   9   20.9     3.1  122 6.3e+07   11086   49752    5664 ( 0.079%) 4590949 (45.3%)   11.231 -1.066e+06    -10.231      0.000      0.000      N/A
+  10   16.4     4.1   95 3.9e+07    5956   24818    2523 ( 0.035%) 4605736 (45.4%)   11.231 -1.057e+06    -10.231      0.000      0.000       25
+  11   12.7     5.3   97 2.3e+07    3035   12314    1010 ( 0.014%) 4614739 (45.5%)   11.231 -1.057e+06    -10.231      0.000      0.000       22
+  12    9.6     6.9   49 1.2e+07    1331    5279     404 ( 0.006%) 4620422 (45.6%)   11.231 -1.057e+06    -10.231      0.000      0.000       21
+  13    7.4     9.0   33 6030341     542    2107     155 ( 0.002%) 4622679 (45.6%)   11.231 -1.058e+06    -10.231      0.000      0.000       20
   14    4.0    11.6   15 2076366     218     946      49 ( 0.001%) 4623856 (45.6%)   11.231 -1.058e+06    -10.231      0.000      0.000       19
   15    3.6    15.1    6  581543      94     423      20 ( 0.000%) 4624228 (45.6%)   11.231 -1.058e+06    -10.231      0.000      0.000       19
   16    3.5    19.7    2  204681      65     313      10 ( 0.000%) 4624852 (45.6%)   11.231 -1.058e+06    -10.231      0.000      0.000       18

```

This may mean that something happened in the system while the tests were running on the heterogeneous tiles build. I believe that, if placement was different, also each routing iteration would be different as well, correct (in terms of overused nodes, CPD, etc.)?

kmurray:
> This may mean that something happened in the system while the tests were running on the heterogeneous tiles build. I believe that, if placement was different, also each routing iteration would be different as well, correct (in terms of overused nodes, CPD, etc.)?

Yep if all the other metrics are identical this likely indicates a system load issue. Another metric to check would be number of heap operations (push, pop etc.), if those are the same then the router did the same amount of work, so we wouldn't expect any run-time difference aside from machine load.

acomodi:
> Another metric to check would be number of heap operations (push, pop etc.)

@kmurray Ok, so in this case they are exactly the same. 

The same applies for the other long-long taking test (stap_qrd_stratixiv_arch_timing):
```
-## Initializing router criticalities took 0.02 seconds (max_rss 4498.5 MiB, delta_rss +0.0 MiB)
+## Initializing router criticalities took 0.03 seconds (max_rss 4498.7 MiB, delta_rss +0.0 MiB)
 ---- ------ ------- ---- ------- ------- ------- ----------------- --------------- -------- ---------- ---------- ---------- ---------- --------
 Iter   Time    pres  BBs    Heap  Re-Rtd  Re-Rtd Overused RR Nodes      Wirelength      CPD       sTNS       sWNS       hTNS       hWNS Est Succ
       (sec)     fac Updt    push    Nets   Conns                                       (ns)       (ns)       (ns)       (ns)       (ns)     Iter
 ---- ------ ------- ---- ------- ------- ------- ----------------- --------------- -------- ---------- ---------- ---------- ---------- --------
-   1   56.5     0.0    0 4.0e+08  144290  285425  178147 ( 2.321%) 3203646 (29.5%)    8.236 -4.506e+05     -7.236      0.000      0.000      N/A
+   1   76.2     0.0    0 4.0e+08  144290  285425  178147 ( 2.321%) 3203646 (29.5%)    8.236 -4.506e+05     -7.236      0.000      0.000      N/A
    2   44.4     0.5  873 2.8e+08  126139  262763   26081 ( 0.340%) 2647211 (24.4%)    9.095 -6.218e+05     -8.095      0.000      0.000      N/A
    3   29.4     0.6  171 1.4e+08   26259   99064   12942 ( 0.169%) 2638576 (24.3%)    8.512 -6.050e+05     -7.512      0.000      0.000      N/A
-   4   23.3     0.8  131 9.5e+07   14907   60953    7983 ( 0.104%) 2646497 (24.4%)    8.490 -6.091e+05     -7.490      0.000      0.000      N/A
-   5   13.0     1.1   83 7.2e+07    9828   43027    4978 ( 0.065%) 2650873 (24.4%)    8.512 -6.127e+05     -7.512      0.000      0.000      N/A
-   6    9.1     1.4   50 4.5e+07    5720   26495    2772 ( 0.036%) 2655921 (24.5%)    8.512 -6.145e+05     -7.512      0.000      0.000      N/A
-   7    5.9     1.9   25 2.3e+07    2922   14327    1342 ( 0.017%) 2660424 (24.5%)    8.373 -6.165e+05     -7.373      0.000      0.000      N/A
-   8    4.2     2.4   25 1.1e+07    1354    6835     577 ( 0.008%) 2664937 (24.6%)    8.373 -6.169e+05     -7.373      0.000      0.000      N/A
-   9    3.1     3.1   12 3784807     483    2355     195 ( 0.003%) 2666299 (24.6%)    8.373 -6.174e+05     -7.373      0.000      0.000      N/A
-  10    2.6     4.1    0 1028778     173     699      46 ( 0.001%) 2667021 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       16
-  11    2.4     5.3    1  257655      56     181      11 ( 0.000%) 2667177 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       14
-  12    2.4     6.9    0  140267      22      65       1 ( 0.000%) 2667213 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       13
-  13    2.3     9.0    0   98985      11      34       0 ( 0.000%) 2667213 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       12
+   4   23.5     0.8  131 9.5e+07   14907   60953    7983 ( 0.104%) 2646497 (24.4%)    8.490 -6.091e+05     -7.490      0.000      0.000      N/A
+   5   20.0     1.1   83 7.2e+07    9828   43027    4978 ( 0.065%) 2650873 (24.4%)    8.512 -6.127e+05     -7.512      0.000      0.000      N/A
+   6   13.9     1.4   50 4.5e+07    5720   26495    2772 ( 0.036%) 2655921 (24.5%)    8.512 -6.145e+05     -7.512      0.000      0.000      N/A
+   7    9.2     1.9   25 2.3e+07    2922   14327    1342 ( 0.017%) 2660424 (24.5%)    8.373 -6.165e+05     -7.373      0.000      0.000      N/A
+   8    6.6     2.4   25 1.1e+07    1354    6835     577 ( 0.008%) 2664937 (24.6%)    8.373 -6.169e+05     -7.373      0.000      0.000      N/A
+   9    4.9     3.1   12 3784807     483    2355     195 ( 0.003%) 2666299 (24.6%)    8.373 -6.174e+05     -7.373      0.000      0.000      N/A
+  10    4.1     4.1    0 1028778     173     699      46 ( 0.001%) 2667021 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       16
+  11    3.9     5.3    1  257655      56     181      11 ( 0.000%) 2667177 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       14
+  12    3.8     6.9    0  140267      22      65       1 ( 0.000%) 2667213 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       13
+  13    3.8     9.0    0   98985      11      34       0 ( 0.000%) 2667213 (24.6%)    8.373 -6.176e+05     -7.373      0.000      0.000       12

```

If everything else looks ok in the PR, I will proceed and re-add the XML changes.

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+A visual representation of the new VPR grid is the following:
+
+.. image:: sub_tiles_grid.png
+
+Heterogeneous tiles examples
+----------------------------
+
+1. Different pin locations among sub tiles.
+
+   The Xilinx Series 7 Clock tile is composed of 16 BUFGCTRL sites. Even though they are equivalent regarding the ports and Fc
```
kmurray, line 32: Maybe  we can add a link to the Xilinx Series 7 clock tile documentation, so that people can look up what the hardware this is modelling actually looks like? Perhaps we could also have a diagram illustrating what this is modelling?

Otherwise this example will be hard to follow for those who don't know the low-level details of 7 series clocking tiles.

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+            <input name="IGNORE0" num_pins="1"/>
+            <input name="IGNORE1" num_pins="1"/>
+            <input name="S0" num_pins="1"/>
+            <input name="S1" num_pins="1"/>
+            <output name="O" num_pins="1"/>
+            <mode ...>
+            ...
+        </pb_type>
+    </complexblocklist>
+
```
kmurray, line 116: I think that after giving the XML, you need to have text which walks through what it represents. So something like:

    The above `BUFG_TILE` contains three types of sub-tiles (`BUFGCTRL_0`, `BUFGCTRL_1` and `BUFGCTRL_2`).
    While each sub-tile type can contain the same pb_type (equivalent_sites of `BUFGCTRL`), the differ in two ways:
        1) Each sub-tile has different pin locations. For example `BUFGCTRL_0` has the `I1` pins on the top side of the tile, while `BUFGCTRL_1` and `BUGCTRL_2` have them on the right and left sides respectively.
        2) Each sub-tile has a different 'capacity' (i.e. a different number of sites). `BUFGCTRL_1` and `BUFGCTRL_2` have capacity 4, while `BUFGCTRL_1` has capacity 8. As a result the `BUFG_TILE` can implement a total of 16 `BUFGCTRL` blocks.

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+2. Different sub tile types.
```
kmurray, line 117: May be re-word:

    Sub-tiles containing different block types

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+
+   As another example taken from the Xilinx Series 7 fabric, the HCLK_IOI tile is composed of three different site types, namely BUFIO, BUFR and IDELAYCTRL.
```
kmurray, line 119: Again, a link to the Xilinx documentation, and a diagram would be helpful here.

doc/src/tutorials/arch/heterogeneous_tiles.rst
kmurray: As an aside, it may help to come up with a more distinct name for the sub-tiles in the example. `BUFGCTRL_1` sub-tile is very similar to `BUFGCTRL` pb-type. Maybe use `BUFGSUBTILE` instead for clarity?

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+
+   Each one of these site types has different IO pins as well as pin locations.
+
+   A visual representation of this case is the following:
+
+.. image:: hclk_ioi.png
```
kmurray, line 125: Again I'd probably use `.. figure::`

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+
+.. code-block:: xml
```
kmurray, line 127: Again there appear to be some XML highlighting issues.

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+                <site name="BUFR_SITE" pin_mapping="direct"/>
+            </equivalent_sites>
+            <fc ...>
+            <pinlocations ...>
+        </sub_tile>
+        <sub_tile name="IDELAYCTRL" capacity="1">
+            <clock name="REFCLK" num_pins="1"/>
+            <output name="RDY" num_pins="1"/>
+            <equivalent_sites>
+                <site name="IDELAYCTRL_SITE" pin_mapping="direct"/>
```
kmurray, line 153: This example uses `<site name=` while the previous example uses `<site pb_type=` we should make sure they are consistent with the actual XML format.

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+            </equivalent_sites>
+            <fc ...>
+            <pinlocations ...>
+        </sub_tile>
+    </tile>
```
kmurray, line 158: We will also want text waling through the XML. For example something like:

    Each ``HCLK_IOI`` tile contains three sub-tiles, each containing a different type of pb_type. The ``BUFIO`` sub-tile supports 4 instances (capacity=4) of pb_type `BUFIO_SITE`, the ``BUFR`` sub-tile supports 4 instances of `BUF_SITE` pb_types, and the `IDELAYCTRL` sub-tile supports 1 instances of the `IDELAYCTRL_SITE`.

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
...
+- ``Fc definition``
+- ``pinlocations``
+- ``IO ports definition``
+
+With this new capability, the device grid of a given architecture does include a new ``depth`` coordinate that identifies
+the type of sub tile used and its actual location, in case the capacity is greater than 1.
+
+A visual representation of the new VPR grid is the following:
+
+.. image:: sub_tiles_grid.png
```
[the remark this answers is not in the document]
kmurray: I think there should also be some discussion of the placer's co-ordinate system. Something along the lines of:

    The placer moves things in x/y/sub-block dimensions. Where the x/y location corresponds to the routing interface location, and sub-block the type-dependent coordinate for multiple things at that location.

doc/src/arch/reference.rst
```diff
@@ -826,6 +826,8 @@ The following tags are common to all ``<tile>`` tags:
 
 .. arch:tag:: <sub_tile name"string" capacity="{int}">
 
+    .. seealso:: For a tutorial on describing the usage of sub tiles for ``heterogeneous tiles`` definition see :ref:`heterogeneous_tiles_tutorial`.
```
kmurray, line 829: I feel like we need a definition of heterogeneous tiles here (novice users wont know what that means). Something along the lines of: heterogeneous tiles (tiles which support multiple instances of the same or different pb_types).

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,158 @@
+.. _heterogeneous_tiles_tutorial:
+
+Heterogeneous tiles tutorial
+============================
+
+This tutorial aims at providing information to the user on how to model sub tiles to enable ``heterogeneous tiles`` in VPR.
```
kmurray, line 6: It seems like you're using \`\` text \`\` (fixed-width text) for emphasis. Generally we've tried to use that when explicitly referring to a something from code (e.g. ``BUFIO`` below).

If you want to emphasize something I'd suggest using italics (`*heterogeneous tiles*`) or bold (`**heterogeneous tiles**`), for more details see: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html.

vpr/src/place/place.cpp
```diff
@@ -2446,7 +2443,7 @@ static int check_block_placement_consistency() {
                 }
```
kmurray, line 2443: One code related item I thought of, and don't see any changes related to. Should there be updates to the placement consistency checks given that tiles can now be of heterogeneous types?

It looks like `check_block_placement_consistency()` is unchanged with respect to this. I think the above block type check will need to be generalized to verify that the sub-tile location (not just the outer tile) is type consistent.

I should add, that if we haven't hit this issue in the add `strong_sub_tile` regression test it likely means we need more test cases which would exercise this case.

acomodi:
@kmurray I have updated the tutorial following your comments. Some notes about that:

- The two examples are actually a simplified version of the real architectures (I have added a note specifying that in the tutorial)
- The HCLK_IOI tile is not described in the documentation. The docs for the Series7 refer to the primitives (e.g. BUFR, BUFIO and IDELAYCTRL) separately, but it happens that those primitives are collected in the HCLK_IOI tile.
- The XML not highlighting was due to syntax errors in the xml itself. I have spotted the same issue in the equivalent sites tutorial and fixed that as well.

I have also added the sub tile consistency check during placement and also increased the netlist size so to have the placer being more stressed o the correct sub tile location to choose.

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,189 @@
...
+                <input name="CE1" num_pins="1"/>
+                <input name="IGNORE0" num_pins="1"/>
+                <input name="IGNORE1" num_pins="1"/>
+                <input name="S0" num_pins="1"/>
+                <input name="S1" num_pins="1"/>
+                <output name="O" num_pins="1"/>
+                <fc in_type="abs" in_val="2" out_type="abs" out_val="2"/>
+                <pinlocations pattern="custom">
+                    <loc side="top">BUFG_SUB_TILE_1.S1 BUFG_SUB_TILE_1.CLK_BUFG_I0 BUFG_SUB_TILE_1.CE1 BUFG_SUB_TILE_1.CLK_BUFG_I1 BUFG_SUB_TILE_1.IGNORE1 BUFG_SUB_TILE_1.IGNORE0 BUFG_SUB_TILE_1.CE0 BUFG_SUB_TILE_1.S0</loc>
+                    <loc side="right">BUFG_SUB_TILE_1.CLK_BUFG_I0 BUFG_SUB_TILE_1.CLK_BUFG_I1 BUFG_SUB_TILE_1.CLK_BUFG_O</loc>
```
kmurray, line 88: Looks like pin names mixed up ( have 'CLKB_BUFG_' prefixed).

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,189 @@
...
+
+    Device grid, with x, y and sub block coordinates. Each block can be moved by the placer in all the three spatial dimensions.
+
+To correctly model an architecture, each :ref:`arch_tiles` requires at least one sub tile definition. This represents a default
+homogeneous architecture, composed of one or many capacity instances of the sub tile within the physical tile.
+
+To enhance the expressivity of VPR architecture, additional sub tiles can be inserted alongside with the default sub tile.
+This enables the definition of the *heterogeneous tiles*.
+
+With this new capability, the device grid of a given architecture does include a new ``depth`` coordinate that identifies
```
kmurray, line 29: Should be ``subblock`` instead of depth for consistency?

vpr/src/place/compressed_grid.cpp
```diff
@@ -20,8 +23,22 @@ std::vector<t_compressed_block_grid> create_compressed_block_grids() {
...
+            std::vector<int> compatible_sub_tiles;
+
+            for (int capacity = 0; capacity < physical_tile->capacity; capacity++) {
+                if (is_sub_tile_compatible(physical_tile, &logical_block, capacity)) {
+                    compatible_sub_tiles.push_back(capacity);
+                }
+            }
+
+            compressed_block_grid.compatible_sub_tiles_map.insert({physical_tile->index, compatible_sub_tiles});
+        }
```
kmurray, line 39: There should be some comments describing what this is calculating.

For instance, it isn't immediately clear what this is doing. It appears to be storing the capacities of compatible sub-tiles, but it isn't clear why this is needed, or how it will be used.

vpr/src/place/compressed_grid.h
```diff
@@ -32,6 +32,8 @@ struct t_compressed_block_grid {
     //stored sparsely, since we may not have full columns of blocks.
     //This makes it easy to check whether there exist
     std::vector<vtr::flat_map2<int, t_type_loc>> grid;
+
+    std::unordered_map<int, std::vector<int>> compatible_sub_tiles_map;
```
kmurray, line 36: We'll need a comment here noting what this is storing, what the key/value are, and its overall purpose. May also need a more descriptive name (compatible_sub_tiles_map gets across some of the idea, but not all).

doc/src/tutorials/arch/heterogeneous_tiles.rst
acomodi: Done

doc/src/tutorials/arch/heterogeneous_tiles.rst
acomodi: Yeah, find/replace gone wrong :). Done

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,189 @@
...
+                <input name="CE1" num_pins="1"/>
+                <input name="IGNORE0" num_pins="1"/>
+                <input name="IGNORE1" num_pins="1"/>
+                <input name="S0" num_pins="1"/>
+                <input name="S1" num_pins="1"/>
+                <output name="O" num_pins="1"/>
+                <fc in_type="abs" in_val="2" out_type="abs" out_val="2"/>
+                <pinlocations pattern="custom">
+                    <loc side="right">BUFGCTRL_1.S1 BUFGCTRL_1.CLK_BUFG_I0 BUFGCTRL_1.CE1 BUFGCTRL_1.CLK_BUFG_I1 BUFGCTRL_1.IGNORE1 BUFGCTRL_1.IGNORE0 BUFGCTRL_1.CE0 BUFGCTRL_1.S0</loc>
+                    <loc side="left">BUFGCTRL_1.CLK_BUFG_I0 BUFGCTRL_1.CLK_BUFG_I1 BUFGCTRL_1.CLK_BUFG_O</loc>
```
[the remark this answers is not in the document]
acomodi: Done

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,189 @@
...
+This comes with the introduction of a third spatial coordinate that identifies the placement of the block type within the x and y grid coordinate.
+
+Moreover, the placer can choose and assign different locations for each block type within the same coordinates as well.
+
+.. figure:: sub_tiles_grid.png
+
+    Device grid, with x, y and sub block coordinates. Each block can be moved by the placer in all the three spatial dimensions.
+
+To correctly model an architecture, each :ref:`arch_tiles` requires at least one sub tile definition. This represents a default
+homogeneous architecture, composed of one or many capacity instances of the sub tile within the physical tile.
```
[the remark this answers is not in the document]
acomodi: Done

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,189 @@
...
+An *heterogeneous tile* is a tile that includes two or more site types that may differ in the following aspects:
+
+- *Fc* definition
+- *Pin locations* definition
+- *IO ports* definition
+
+As a result, an *heterogeneous tile* has the possibility of having multiple block types at the same *x* and *y* location in the grid.
+This comes with the introduction of a third spatial coordinate that identifies the placement of the block type within the x and y grid coordinate.
+
+Moreover, the placer can choose and assign different locations for each block type within the same coordinates as well.
```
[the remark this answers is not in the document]
acomodi: Done

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,189 @@
...
+Heterogeneous tiles tutorial
+============================
+
+This tutorial aims at providing information to the user on how to model sub tiles to enable *heterogeneous tiles* in VPR.
+
+An *heterogeneous tile* is a tile that includes two or more site types that may differ in the following aspects:
+
+- *Fc* definition
+- *Pin locations* definition
+- *IO ports* definition
```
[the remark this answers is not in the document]
acomodi: Done

doc/src/tutorials/arch/heterogeneous_tiles.rst
```diff
@@ -0,0 +1,189 @@
...
+
+As a result, an *heterogeneous tile* has the possibility of having multiple block types at the same *x* and *y* location in the grid.
+This comes with the introduction of a third spatial coordinate that identifies the placement of the block type within the x and y grid coordinate.
```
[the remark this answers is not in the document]
acomodi: Done

vpr/src/place/compressed_grid.h
acomodi: Done. I am unable to find a better name for this, I replaced it with `compatible_sib_tiles_for_tile_map` which sounds maybe even worse than before. Do you have any suggestions?

vpr/src/place/compressed_grid.cpp
```diff
@@ -29,13 +29,21 @@ std::vector<t_compressed_block_grid> create_compressed_block_grids() {
...
 
+            // Build a vector to store all the sub tile indices that are compatible with the
+            // (physical_tile, logical_block) pair.
+            //
+            // Given that a logical block can potentially be compatible with only a subset of the
+            // sub tiles of a physical tile, we need to ensure which sub tile locations are part of
+            // this subset.
             for (int capacity = 0; capacity < physical_tile->capacity; capacity++) {
                 if (is_sub_tile_compatible(physical_tile, &logical_block, capacity)) {
                     compatible_sub_tiles.push_back(capacity);
```
kmurray: I'd suggest renaming `capacity` to `isubblk` (or similar). Calling the variable capacity is kind of confusing, since that is usually the term for the 'count' of sub-tiles not their index.

vpr/src/place/compressed_grid.h
```diff
@@ -33,7 +33,11 @@ struct t_compressed_block_grid {
     //This makes it easy to check whether there exist
     std::vector<vtr::flat_map2<int, t_type_loc>> grid;
 
-    std::unordered_map<int, std::vector<int>> compatible_sub_tiles_map;
+    //The sub type compatibility for a given physical tile and a compressed block grid
+    //corresponding to the possible placement location for a given logical block
+    //  - key: physical tile index
+    //  - value: vector of compatible sub tiles for the physical tile/logical block pair
+    std::unordered_map<int, std::vector<int>> compatible_sub_tiles_for_tile_map;
```
kmurray, line 40: The name makes more sense now.

However I'd suggest dropping `_map` from the name (as its redundant with the type), so: `compatible_sub_tiles_for_tile`.

vpr/src/place/compressed_grid.cpp
acomodi: Sure, done

vpr/src/place/compressed_grid.h
acomodi: Done

kmurray:
Thanks @acomodi for putting this together! This should be a great enhancement to the placer and our modelling capabilities.

PR #1211, opened 2020-03-11
Verify RR Node Indicies are Consistent with RR Graph

kmurray:
#### Description

While debugging other issues I found inconsistencies with the rr_node_indices reverse RR graph look-up.

This PR adds:
 * A verification routine which sanity checks RR node indices against the current RR graph (this also has the added benefit of codifying exactly what is stored in rr_node_indicies which hasn't always been clear).
 * Fixes several bugs (mostly in the clock RR graph construction code) which were causing the inconsistencies

### DIFF
--- a/vpr/src/route/rr_graph2.h
+++ b/vpr/src/route/rr_graph2.h
@@ -3,11 +3,10 @@
 #include <vector>
 
 #include "build_switchblocks.h"
+#include "rr_graph_fwd.h"
 #include "rr_graph_util.h"
 #include "rr_types.h"
 #include "device_grid.h"
-
-struct t_rr_graph_storage; //Forward declaration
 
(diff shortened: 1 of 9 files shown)


vpr/src/route/rr_graph2.h
```diff
@@ -7,6 +7,8 @@
 #include "rr_types.h"
 #include "device_grid.h"
 
+struct t_rr_graph_storage; //Forward declaration
```
litghost: t_rr_graph_storage is a class, not a struct

kmurray: I've fixed this for consistency, although I believe there is no real distinction between struct/class (other than default members being public vs private).

litghost: It generates warnings from the compile, which we avoid.

PR #1109, opened 2020-02-04
Use improved criticality estimates for the first routing iteration

kmurray:
#### Description

Traditionally VPR has routed all connections during the 1st router iteration as timing critical (i.e. criticality = 1). While this helps ensure good initial delay estimates it has a CPU-time cost.

In particular, since the spatial-lookup for High Fanout (HF) nets is only active for less timing critical nets (i.e. criticality < 0.9) using high criticality can have a significant run-time cost on designs with large high-fanout nets.

Currently this PR only implements this functionality, which is off by default.

#### Related Issue
#1073 

#### Motivation and Context

If we can use more realistic criticality estimates we can ensure that less-timing critical high-fanout  nets use the more efficient spatial-lookup based approach.

Also, for designs which aren't heavily congested the 1st routing iteration is usually the longest, so reducing it's run-time is desirable.

#### How Has This Been Tested?

Currently tested locally on vtr_reg_basic, vtr_reg_strong. Also on the Titan benchmarks with `--router_lookahead map --router_initial_timing lookahead --router_update_lower_bound_delays on`, showing ~6% route time reduction on average, with larger reductions on large designs with high-fanout nets (e.g. 35% route-time reduction on gaussianblur).

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1216, opened 2020-03-12
WIP: Default map lookahead

kmurray:
#### Description
Previously the 'map' router lookahead was not used by default since it was not always robust (e.g. on small devices, or with non-unifrom RR graph structures like clock networks).

This PR addresses these issues, and switches the default lookahead to be the map lookahead.

It also fixes incorrect initialization of RR index data for non-uniform RR graph structures (e.g. clock networks) which were previously not correctly initialized.

With a robust map lookahead it will be possible to flip the `--router_initial_timing` option to `lookahead` by default to gain the benefits of #1109.


kmurray:
I am in the process of collecting updated QoR comparison data.
The nightly/weekly golden results will likely also need to be updated.

Previously (before rebasing around recent RR graph changes) this achieved: 
 * Run-time Reductions: ~13% place time (mostly faster delta calculation), ~20% route time, resulting in ~11% total time
 * QoR: -4% WL, 0% CPD (no change)

The improved run-time/quality is likely due to a more accurate lookahead caused by picking more/better sampling locations while profiling the RR graph.

kmurray:
Another item to do before merging:
 * Currently the code uses the legacy t_rr_node interface. This should be updated to use the new interface, particularly in the lookahead querying code which is called in the core router loop.

kmurray:
I have now addressed some run-time issues with the original version of this PR (original values posted above were incorrect). Here are the new results: [comparison_output_alt_more_sample.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4393783/comparison_output_alt_more_sample.xlsx)

Summary: Equivalent QoR, 5% reduction in route time.

PR #1235, opened 2020-03-25
Updates to Kokoro artifacts and docs

kmurray:
Include '*.out' files (e.g. vpr.out, vpr.crit_path.out) in kokoro artifacts. These are the file which are useful for debugging flow issues and so should be included.

This PR will also update the kokoro documentation based on my experience debugging issues in kokoro runs.

### DIFF
--- a/.github/kokoro/presubmit/strong.cfg
+++ b/.github/kokoro/presubmit/strong.cfg
@@ -8,6 +8,7 @@ timeout_mins: 60
 action {
   define_artifacts {
     # File types
+    regex: "**/*.out"
     regex: "**/vpr_stdout.log"
     regex: "**/parse_results.txt"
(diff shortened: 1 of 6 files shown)


mithro:
FYI -- @litghost

.github/kokoro/presubmit/strong.cfg
```diff
@@ -8,6 +8,7 @@ timeout_mins: 60
 action {
   define_artifacts {
     # File types
+    regex: "**/*.out"
```
litghost: Probably want this for continuous/weeky and continuous/strong too?

kmurray: Yep good point, I'll put those in too.

Issue #1244, opened 2020-03-28
MacOS support


HackerFoo:
#### Proposed Behaviour
VtR should build successfully on MacOS. This will allow people to work on VtR using a Mac machine without using a Linux VM.

#### Current Behaviour
It doesn't.

#### Possible Solution
Add a `default.nix` file and fix any bugs preventing a successful build.

#### Context
I was working on improving my [Nix](https://nixos.org/nix/) setup, and realized Nix works on MacOS too.

This could also be useful to port to [Windows 10](https://nathan.gs/2019/04/12/nix-on-windows/).

PR #1246, opened 2020-03-28
Fix minor bugs to get VtR building on MacOS with Nix

HackerFoo:
### Description
Fixes bugs and adds a working `shell.nix` for developers using MacOS.

#### Related Issue
#1244

#### Motivation and Context
I have a Mac. I like Nix.

I ran `vtr_reg_strong` on my Mac. There was 1 test failure and 13 QoR failures, which is interesting because I didn't change anything significant.

#### Types of changes

#### Checklist:

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -94,6 +94,24 @@ apt-get install \
 	ctags
 ```
 
+#### Using Nix ####
+
+Although the recommended platform is Debian or Ubuntu, Nix can be used to build VTR on other platforms, such as MacOS.
+
+If you don't have [Nix](https://nixos.org/nix/), you can [get it](https://nixos.org/nix/download.html) with:
+
+```shell
+$ curl -L https://nixos.org/nix/install | sh
+```
+
+These commands will set up dependencies for Linux and MacOS and build VTR:
+
+```shell
+#In the VTR root
+$ nix-shell dev/nix/shell.nix
+$ make
+```
+
 ### Building using the Makefile wrapper ###
 Run `make` from the root of the VTR source tree
 
--- a/libs/libvtrutil/src/vtr_logic.h
+++ b/libs/libvtrutil/src/vtr_logic.h
@@ -1,5 +1,20 @@
+// Put this above guard so that TRUE/FALSE are undef'ed
+// even if this file was already included earlier.
 #ifndef VTR_LOGIC_H
 #define VTR_LOGIC_H
+
+#ifdef FALSE
+#    undef FALSE
+#endif
+#define FALSE FALSE
+
+#ifdef TRUE
+#    undef TRUE
+#endif
+#define TRUE TRUE
+
+constexpr int FALSE = 0;
+constexpr int TRUE = 1;
 
 namespace vtr {
 
--- a/vpr/src/route/gen/rr_graph_uxsdcxx.h
+++ b/vpr/src/route/gen/rr_graph_uxsdcxx.h
@@ -20,7 +20,6 @@
 #include <string>
 #include <vector>
 
-#include <error.h>
 #include <stddef.h>
 #include <stdint.h>
 #include "pugixml.hpp"
--- a/ODIN_II/SRC/include/Hashtable.hpp
+++ b/ODIN_II/SRC/include/Hashtable.hpp
@@ -25,6 +25,7 @@
 
 #include <stdlib.h>
 #include <stdint.h>
+#include <string>
 #include <unordered_map>
 #include <string>
 
--- a/ODIN_II/SRC/odin_ii.cpp
+++ b/ODIN_II/SRC/odin_ii.cpp
@@ -250,7 +250,7 @@ netlist_t* start_odin_ii(int argc, char** argv) {
 
     /* read the FPGA architecture file */
     if (global_args.arch_file.provenance() == argparse::Provenance::SPECIFIED) {
-        printf("Architecture: %s\n", basename(global_args.arch_file.value().c_str()));
+        printf("Architecture: %s\n", vtr::basename(global_args.arch_file.value()).c_str());
         fflush(stdout);
 
         printf("Reading FPGA Architecture file\n");
@@ -267,7 +267,7 @@ netlist_t* start_odin_ii(int argc, char** argv) {
     /* do High level Synthesis */
     if (!configuration.list_of_file_names.empty() && configuration.is_verilog_input) {
         for (std::string v_file : global_args.verilog_files.value()) {
-            printf("Verilog: %s\n", basename(v_file.c_str()));
+            printf("Verilog: %s\n", vtr::basename(v_file).c_str());
         }
         fflush(stdout);
 
@@ -294,7 +294,7 @@ netlist_t* start_odin_ii(int argc, char** argv) {
             configuration.list_of_file_names = {global_args.output_file};
             current_parse_file = 0;
         } else {
-            printf("Blif: %s\n", basename(global_args.blif_file.value().c_str()));
+            printf("Blif: %s\n", vtr::basename(global_args.blif_file.value()).c_str());
             fflush(stdout);
         }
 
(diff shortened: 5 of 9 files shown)


libs/libvtrutil/src/vtr_logic.h
```diff
@@ -1,5 +1,15 @@
+// Put this above guard so that TRUE/FALSE are undef'ed
+// even if this file was already included earlier.
+#ifdef FALSE
+#    undef FALSE
+#endif
+
+#ifdef TRUE
+#    undef TRUE
+#endif
```
kmurray: Is there a potential issue with code including this header but relying on TRUE/FALSE to be defined (e.g. other header files included after this one)?

The TRUE/FALSE in the enum are scoped (vtr::LogicValue::TRUE/FALSE), but any usage of TRUE/FALSE defines may not be. Thoughts?

shell.nix
```diff
@@ -0,0 +1,34 @@
+with import <nixpkgs> {};
```
kmurray: Can you clarify what this files is for?

Generally we try to avoid checking OS/IDE specific configuration files into the VTR source tree.

libs/libvtrutil/src/vtr_logic.h
HackerFoo: The best way to resolve this is to rename `vtr::LogicValue::TRUE/FALSE`, because these are defined as [macros on MacOS](https://opensource.apple.com/source/xnu/xnu-6153.11.26/osfmk/mach/boolean.h.auto.html).

I found a workaround defining the macros to themselves, preventing redefinition by `boolean.h`, and then defining constexprs for `TRUE` and `FALSE`.

shell.nix
HackerFoo: This is so that you can type `nix-shell` and enter an [environment where all dependencies are installed](https://nixos.org/nixos/nix-pills/developing-with-nix-shell.html), ready for `make`. It's not OS or IDE specific, but does use the [Nix package manager](https://nixos.org/nix/), which runs on Linux, MacOS, and maybe Windows 10 with WSL.

I think it's worth including so that other developers can get started quickly. It's also good documentation, because `nix-shell --pure` enforces that every dependency must be listed in the .nix file by hiding all system libraries and headers.

For non-developers, I could submit a PR to [nixpkgs](https://nixos.org/nixpkgs/) at some point, but it's not clear to me that there are any non-developer users yet that would benefit.

HackerFoo: I've added documentation to `BUILDING.md`.

PiotrZierhoffer:
@HackerFoo I have a minor issue when trying to use your code. Not knowing of your PR I was working on the same thing and came up with a similar set of fixes.

For some reason I also have to have this commit: https://github.com/antmicro/vtr-verilog-to-routing/commit/238189cadafcfb08a2a5b6a5ff0d8181106aff7b

It was needed because isnan required a float and tatum::Time, despite having a cast operator, didn't pass the `is_floating_point` filter.

I will try to build it on Travis to see how it goes without your patch.

I wonder where the difference might come from. Different xcode version?

HackerFoo:
Are you using Nix? The entire environment is specified in the `shell.nix` file. The compiler supplied by `nixpkgs` is clang 7.1.0.

If you follow the directions added to `BUILDING.md` it should just work.

PiotrZierhoffer:
Indeed, I missed that part. I'm trying to build on a plain Mac and, eventually, in Travis with Conda. I'll verify my deps, thanks.

BUILDING.md
```diff
@@ -94,6 +94,22 @@ apt-get install \
 	ctags
 ```
 
+#### Using Nix ####
+
+If you don't have [Nix](https://nixos.org/nix/), you can [get it](https://nixos.org/nix/download.html) with:
```
kmurray, line 99: We probably want to give a little more context here. In particular that:
 - This is an alternate way to install dependencies (the recommended/supported way is with apt)
 - This is focused on Mac support(?)

libs/libvtrutil/src/vtr_vector.h
```diff
@@ -61,7 +61,6 @@ class vector : private std::vector<V, Allocator> {
     using storage::size;
 
     using storage::back;
-    using storage::data;
```
kmurray, line 64: Why is the data() member removed? It is part of C++11.

HackerFoo:
> Indeed, I missed that part. I'm trying to build on a plain Mac and, eventually, in Travis with Conda. I'll verify my deps, thanks.

Travis [supports Nix](https://docs.travis-ci.com/user/languages/nix/), so it might be a [lot easier than Conda](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/use-conda-with-travis-ci.html).

libs/libvtrutil/src/vtr_vector.h
HackerFoo: I removed it to fix this error:
```
In file included from /Users/deweese/src/vtr-verilog-to-routing/libs/libvtrutil/src/vtr_string_interning.h:54:
/Users/deweese/src/vtr-verilog-to-routing/libs/libvtrutil/src/vtr_vector.h:64:20: error: no member named 'data' in 'std::__1::vector<bool, std::__1::allocator<bool> >'
    using storage::data;
          ~~~~~~~~~^
/Users/deweese/src/vtr-verilog-to-routing/vpr/src/draw/draw_types.h:205:39: note: in instantiation of template class 'vtr::vector<vtr::StrongId<cluster_block_id_tag, int, -1>, bool,
      std::__1::allocator<bool> >' requested here
    vtr::vector<ClusterBlockId, bool> use_default_block_color_;
                                      ^
```
Maybe the problem is in `draw_types.h` instead, since it's that same error repeated. I'll look at it some more.

BUILDING.md
HackerFoo: I've added this to clarify: "Although the recommended platform is Debian or Ubuntu, Nix can be used to build VTR on other platforms."

libs/libvtrutil/src/vtr_vector.h
kmurray: Hmm, `std::vector<bool>` is [has a proxied implementation in the C++ standard](https://stackoverflow.com/questions/17794569/why-is-vectorbool-not-a-stl-container) which means it isn't really a container. Comparing the definitions of [std::vector\<T\>](http://www.cplusplus.com/reference/vector/vector/) and [std::vector\<bool\>](http://www.cplusplus.com/reference/vector/vector-bool/) show significant differences in the availability of different member functions.

Probably the fix is to only enable the:
```cpp
using storage::data;
```
in vtr::vector only if `V` is not a `bool` (perhaps with `std::enable_if`?).

hzeller, line 64: Fixed that in #1282 (it was also needed to then actually provide an implementation of data() for all the places that actually use `data()` for non-`vector<bool>` specialized vectors)

HackerFoo: This is now fixed by #1282. Thanks @hzeller!

HackerFoo: Thanks!

vpr/src/route/gen/rr_graph_uxsdcxx.h
```diff
@@ -20,7 +20,6 @@
 #include <string>
 #include <vector>
 
-#include <error.h>
```
kmurray, line 23: This file is generated by uxsdcxx, we really should be fixing this upstream (c.f. #1286) rather than here in VTR. Otherwise, we'll loose this change when we next re-generate this code.

HackerFoo: I've submitted a PR upstream: https://github.com/SymbiFlow/uxsdcxx/pull/51

HackerFoo:
@kowalewskijan Fixed, thanks.

PR #1252, opened 2020-03-31
Add infrastructure for saving VPR graphics to generate animations

kmurray:
#### Description
Adds `--graphics_commands` to VPR and the `dev/vpr_animate.py` script used to generate the optimization animations on www.verilogtorouting.org and #1251.

These features should be generally useful for inspecting/visualizing/understanding what VPR is doing on particular benchmarks.


mithro:
@kmurray - Let the animation party begin!

PR #1262, opened 2020-04-03
Improve Timing Graph Build Comments

### LINKED ISSUE #1040: Code review comments from tracing timing graph build
This issue tracks various code review comments on the timing graph code.

kmurray:
#### Description

Add high-level overview and more comments to timing graph construction code. To address comments from #1040

Closes #1040

### DIFF
--- a/vpr/src/timing/timing_graph_builder.cpp
+++ b/vpr/src/timing/timing_graph_builder.cpp
@@ -351,7 +351,7 @@ void TimingGraphBuilder::add_io_to_timing_graph(const AtomBlockId blk) {
 
     NodeId tnode = tg_->add_node(node_type);
 
-    netlist_lookup_.set_atom_pin_tnode(pin, tnode);
+    netlist_lookup_.set_atom_pin_tnode(pin, tnode, BlockTnode::EXTERNAL);
 }
 
 //Creates the timing graph nodes and internal edges for a netlist block
@@ -427,7 +427,7 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
                                                         model_port->combinational_sink_ports.end());
 
         //Save the pin to external tnode mapping
-        netlist_lookup_.set_atom_pin_tnode(input_pin, tnode);
+        netlist_lookup_.set_atom_pin_tnode(input_pin, tnode, BlockTnode::EXTERNAL);
     }
 
     //Create the clock pins
@@ -442,7 +442,7 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
 
         NodeId tnode = tg_->add_node(NodeType::CPIN);
 
-        netlist_lookup_.set_atom_pin_tnode(clock_pin, tnode);
+        netlist_lookup_.set_atom_pin_tnode(clock_pin, tnode, BlockTnode::EXTERNAL);
     }
 
     //Create the output pins
@@ -498,7 +498,7 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
         }
 
         //Record as external tnode
-        netlist_lookup_.set_atom_pin_tnode(output_pin, tnode);
+        netlist_lookup_.set_atom_pin_tnode(output_pin, tnode, BlockTnode::EXTERNAL);
     }
 
     //Connect the clock pins to the sources and sinks
@@ -724,7 +724,7 @@ void TimingGraphBuilder::remap_ids(const tatum::GraphIdMaps& id_mapping) {
     for (auto kv : new_tnode_atom_pin) {
         tatum::NodeId tnode = kv.first;
         AtomPinId pin = kv.second;
-        netlist_lookup_.set_atom_pin_tnode(pin, tnode);
+        netlist_lookup_.set_atom_pin_tnode(pin, tnode, BlockTnode::EXTERNAL);
     }
 }
 
--- a/vpr/src/timing/timing_graph_builder.cpp
+++ b/vpr/src/timing/timing_graph_builder.cpp
@@ -302,7 +302,7 @@ void TimingGraphBuilder::build(bool allow_dangling_combinational_nodes) {
     //Break any combinational loops (i.e. if the graph is not a DAG)
     fix_comb_loops();
 
-    //Levelize the graph (i.e. determine its topological ordering and record 
+    //Levelize the graph (i.e. determine its topological ordering and record
     //the level of each node) for the timing analyzer
     tg_->levelize();
 }
@@ -388,6 +388,16 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
      * bad practise, but it happens). We detect such cases and also create a
      * SOURCE (and leave any combinational inputs to that node disconnected).
      */
+
+    auto clock_generator_tnodes = create_block_timing_nodes(blk);
+    create_block_internal_data_timing_edges(blk, clock_generator_tnodes);
+    create_block_internal_clock_timing_edges(blk, clock_generator_tnodes);
+}
+
+//Constructs the timing graph nodes for the specified block
+//
+//Returns the set of created tnodese which are clock generators
+std::set<tatum::NodeId> TimingGraphBuilder::create_block_timing_nodes(const AtomBlockId blk) {
     std::set<std::string> output_ports_used_as_combinational_sinks;
 
     //Create the tnodes corresponding to input pins
@@ -501,6 +511,10 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
         netlist_lookup_.set_atom_pin_tnode(output_pin, tnode, BlockTnode::EXTERNAL);
     }
 
+    return clock_generator_tnodes;
+}
+
+void TimingGraphBuilder::create_block_internal_clock_timing_edges(const AtomBlockId blk, const std::set<tatum::NodeId>& clock_generator_tnodes) {
     //Connect the clock pins to the sources and sinks
     for (AtomPinId pin : netlist_.block_pins(blk)) {
         for (auto blk_tnode_type : {BlockTnode::EXTERNAL, BlockTnode::INTERNAL}) {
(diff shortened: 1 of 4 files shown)


kmurray:
@litghost Please take a look. 

I think the high-level comment provides more context as to the why behind the code. However if things still aren't clear let me know!

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -1,3 +1,196 @@
...
+ *      These correspond to net connections, combinational paths inside primitives, 
+ *      clock-to-q, and setup/hold constraints for sequential elements.
+ *
+ * The timing graph is constructed by first creating timing sub-graphs corresponding to 
+ * each primitive block in the AtomNetlist. Each sub-graph contains the timing graph 
+ * nodes required to represent the primitive, and all internal timing graph edges (i.e.
+ * those completely contained within the primitive).
+ *
+ * Next timing graph edges corresponding to nets in the netlist (i.e. connections between 
+ * netlist primitives) are created to "stich" the sub-graph together. This results in a
```
litghost, line 21: `"stich"` - Spelling?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -1,3 +1,196 @@
...
+ * timing graph corresponding to the netlist.
+ *
+ *
+ * Modelling Primitive Combinational and Sequetial Logic as a Timing Graph
```
litghost, line 25: `"Sequetial"` - SP?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -1,3 +1,196 @@
...
+ * pin 'clk'), and two clouds of combinational logic 'C' and 'D'.
+ *
+ * The combinational logic 'D' is driven by primtive input 'e' and 
+ * drives primitive output pin 'g'.
+ *
+ * The combinational logic 'C' is driven by the sequential element 'A' and
+ * primitive input pin 'e'; it then drives the input of the sequential 
+ * element 'B'.
+ *
+ * Sequentail element 'A' is driven by primitive input 'f', and sequential
```
litghost, line 39: SP

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -1,3 +1,196 @@
...
+ *           |   +-^-+                 +-^-+   |
+ *           |     |                     |     |
+ *           |     |                     |     |
+ *   clk --->|---------------------------+     |
+ *           |                                 |
+ *           +---------------------------------+
+ *
+ *  As a result there are the following "timing sub-paths" within this primitive:
+ *
+ *  1) 'e' -> 'D' -> 'g'
```
litghost, line 64: I think it would be useful to expand the types of edges here?  If I understand e -> and D -> g, those are both simple propigation delay?  clk -> A is a clk-to-Q edge (e.g. launch delay).

Should `'f' -> 'A'` be a `'f' -> 'A.D', and `'A' -> 'C'` be `'A.Q' -> 'C'`, etc.?  Because A is actually 3 ports, A.D, A.clk, and A.Q.  There is not edge from A.D to A.Q at all.  Only `A.D` captured from `A.clk`, and `A.Q` launched from `A.clk`.  Yes?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -1,3 +1,196 @@
...
+ * tagged on the pins of the primitive which indicate:
+ *     - whether a pin is sequential, combinational or a clock
+ *     - whether the pin is combinationally connected to another pin 
+ *       within the primitive.
+ * 
+ * Mostly there is a one-to-one correspondance between netlist pins and tnodes,
+ * the only exception is for sequential-sequential connections within a primtive
+ * (e.g. the fully internal 'A' to 'B' timing path above).
+ *
+ * As a result we make a distinction between tnodes which are strictly "internal"
```
litghost, line 156: What is the modelling affect of internal vs external?  Or is simply a label, and has no affect on the timing analysis results?  Please add answer to comment.

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -1,3 +1,196 @@
...
+ * to a primitive, and those which are "external" (i.e. connect to tnodes outside
+ * the primitive). This is a reasonable choice, since the rest of VPR doesn't care
+ * about the internal timing paths, as they are typically uneffected by any of 
+ * VPR's optimization choices.
+ *
+ * As a result the timing graph we build (for the primitive example above) is 
+ * described within VPR as:
+ *
+ *       +------------+                                            +------------+
+ *   --->|   IPIN e   |------------------------------------------->|   OPIN g   |--->
```
litghost, line 166: What does it mean that these are both internal and external?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -69,21 +265,27 @@ void TimingGraphBuilder::build(bool allow_dangling_combinational_nodes) {
...
+    //Walk through the netlist nets adding the edges representing each net to
+    //the timiing graph
     for (AtomNetId net : netlist_.nets()) {
         add_net_to_timing_graph(net);
     }
 
+    //Break any combinational loops (i.e. if the graph is not a DAG)
     fix_comb_loops();
 
+    //Levelize the graph for the timing analyzer
```
litghost, line 277: What is meant by levelize?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -50,13 +243,16 @@ std::unique_ptr<TimingGraph> TimingGraphBuilder::timing_graph(bool allow_danglin
...
 
+//Constructs the timing graph from the netlist
 void TimingGraphBuilder::build(bool allow_dangling_combinational_nodes) {
     tg_ = std::make_unique<tatum::TimingGraph>();
 
     // Optionally allow dangling combinational nodes.
     // Set by `--allow_dangling_combinational_nodes on`. Default value is false
     tg_->set_allow_dangling_combinational_nodes(allow_dangling_combinational_nodes);
 
+    //Walk through the netlist blocks adding the associated nodes and internal edges
```
litghost, line 254: Maybe this comment is better as a little more explicit.  If I follow the top-level comments, this look is constructing the portion of the timing graph that are within each pb-type, leaving a set of of external nodes unconnected until add_net_to_timing_graph?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -148,11 +383,13 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
         }
```
litghost, line 383: I think it is worth adding a comment on line "tnode = tg_->add_node(NodeType::SINK);" that this is a sink because it a capture point (e.g. T setup / hold to clock).

litghost: I don't believe the comment on line 377 is correct ("//There is an internal combinational connection starting at this sequential input")  makes sense to me.  If there is a combination connection from the sequence input to a part of this block, should simply be an edge from SINK node to the combination destinations?

I believe you need to add to your top-level comment computational + sequence examples.

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -273,56 +518,56 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
...
-            //Look-up the combinationally connected sink ports name on the port model
-            AtomPortId src_port = netlist_.pin_port(src_pin);
-            const t_model_ports* model_port = netlist_.port_model(src_port);
+        auto src_type = tg_->node_type(src_tnode);
 
-            for (const std::string& sink_port_name : model_port->combinational_sink_ports) {
-                AtomPortId sink_port = netlist_.find_port(blk, sink_port_name);
-                if (!sink_port) continue; //Port may not be connected
+        //Look-up the combinationally connected sink ports name on the port model
+        AtomPortId src_port = netlist_.pin_port(src_pin);
```
litghost, line 529: Need `VTR_ASSERT_SAFE(bool(src_port))`?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -273,56 +518,56 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
...
+        const t_model_ports* model_port = netlist_.port_model(src_port);
```
litghost, line 530: `VTR_ASSERT_SAFE(model_port != nullptr)`?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -273,56 +518,56 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
...
 
-                //We now need to create edges between the source pin, and all the pins in the
-                //output port
-                for (AtomPinId sink_pin : netlist_.port_pins(sink_port)) {
-                    //Get the tnode of the sink
-                    NodeId sink_tnode = netlist_lookup_.atom_pin_tnode(sink_pin, BlockTnode::INTERNAL);
+        for (const std::string& sink_port_name : model_port->combinational_sink_ports) {
```
litghost, line 532: This nesting is pretty deep.  Does it make sense to break down functionality into a named function?  Maybe "find_primative_connections_from_IPIN" or something?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -273,56 +518,56 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
         //and between the internal SOURCEs and SINKS for sequential blocks
```
litghost, line 518: Did I missing something, or is there no loop that constructs sequential to sequential internal delays?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -1,3 +1,196 @@
+/*
+ * Timing Graph Builder
+ *
+ * This file implements the TimingGraphBuilder class which constructs the timing graph
+ * from the provided AtomNetlist.
+ *
+ * The Timing Graph is a directed acyclic graph (DAG) consisting of nodes and edges:
+ *    - Nodes: represent netlist pins (tatum::IPIN, tatum::OPIN) and logical sources/
+ *      sinks ike primary inputs/outpus, flip-flops and clock generators (tatum::SOURCE,
```
mkurc-ant, line 9: `"ike"` -> "like" ?

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -69,21 +265,27 @@ void TimingGraphBuilder::build(bool allow_dangling_combinational_nodes) {
...
     }
 
+    //Break any combinational loops (i.e. if the graph is not a DAG)
     fix_comb_loops();
 
+    //Levelize the graph for the timing analyzer
     tg_->levelize();
 }
 
+//Re-order the timing graph to be ordered in a traversal friendly manner
```
mkurc-ant, line 281: I'd like a more detailed description how the optimization is done and why there is a need to remap ids.

vpr/src/timing/timing_graph_builder.cpp
```diff
@@ -351,7 +596,7 @@ void TimingGraphBuilder::add_block_to_timing_graph(const AtomBlockId blk) {
                 NodeId sink_tnode = netlist_lookup_.atom_pin_tnode(sink_pin, BlockTnode::EXTERNAL);
 
                 tg_->add_edge(tatum::EdgeType::PRIMITIVE_COMBINATIONAL, src_tnode, sink_tnode);
-                VTR_LOG("Adding edge from '%s' (%zu) -> '%s' (%zu)\n", netlist_.pin_name(src_clock_pin).c_str(), size_t(src_tnode), netlist_.pin_name(sink_pin).c_str(), size_t(sink_tnode));
+                VTR_LOG("Adding edge from '%s' (tnode: %zu) -> '%s' (tnode: %zu) to allow clocks to propagate\n", netlist_.pin_name(src_clock_pin).c_str(), size_t(src_tnode), netlist_.pin_name(sink_pin).c_str(), size_t(sink_tnode));
```
mkurc-ant, line 599: It would be good to add messages referencing pin names to all assert messages wherever possible. A simple message that there is something wrong with the timing graph does not tell much without the information where the problem is in the architecture / netlist.

kmurray: > A simple message that there is something wrong with the timing graph does not tell much without the information where the problem is in the architecture / netlist.

I agree.

> It would be good to add messages referencing pin names to all assert messages wherever possible.

However, I'm less concerned about having end-user friendly assert messages (which ideally should only be relevant to developers), and more concerned that we should produce good error messages.

If an assertion is regularly firing due to bad user input, that means it should be converted to an error with an appropriate message (including user-friendly information).

vpr/src/timing/timing_graph_builder.cpp
kmurray: Done

vpr/src/timing/timing_graph_builder.cpp
kmurray: Done

Issue #1225, opened 2020-03-20
std::tie is slower than expected for symbiflow benchmark tests


xuqinziyue:
comparison using std::tie in **operator<(const t_override& lhs, const t_override& rhs)** for struct **t_override** in **place_delay_model.h** seems to be slower than expected

#### Expected Behaviour
The function is supposed to only do several comparison between short numbers, so it is not expected to consume too much execution time.

#### Current Behaviour
gprof shows the function is consuming more than 8% of place time on picosoc benchmark test for symbiflow

#### Possible Solution
I replaced the std::tie with following code and there seems to be around 9% reduction in place time for the xc7 tests I measured on:

            unsigned long long temp_left = (unsigned long long)lhs.from_type<<48 | (unsigned long long)lhs.to_type<<32 | (unsigned long long)lhs.from_class << 16 | (unsigned long long)lhs.to_class;
            unsigned long long temp_right = (unsigned long long)rhs.from_type<<48 | (unsigned long long)rhs.to_type<<32 | (unsigned long long)rhs.from_class << 16 | (unsigned long long)rhs.to_class;
            if (temp_left < temp_right) {
                return true;
            }
            else if (temp_left == temp_right) {
                temp_left = ((unsigned)lhs.delta_x << 16 | (unsigned)lhs.delta_y);
                temp_right = ((unsigned)rhs.delta_x << 16 | (unsigned)rhs.delta_y);
                return temp_left < temp_right;
            }
            else {
                return false;
            }

  | Original place time | New place time
top_bram_n1 | 2.39 | 2.15
top_bram_n2 | 3.97 | 3.68
top_bram_n3 | 6.09 | 5.51
murax_basys | 6.3 | 5.69
picosoc_basys3 | 38.7 | 35.11

#### Your Environment
* VTR revision used: symbiflow/vtr with commit number 7351b7cfcdea5807e58971cf463249101a87d444
* Operating System and version: linux

vaughnbetz:
Adding @kmurray to comment.
Easton, are other QoR metrics for Symbiflow unaffected (should be, but best to check).
The next step is to turn on the placement_delay override for the VTR and Titan benchmarks (see https://docs.verilogtorouting.org/en/latest/vpr/command_line_usage/) as it is a non-default option, and see what impact there is on place time (reduction hopefully) and other metrics (should be none).

[...]

PR #1265, opened 2020-04-05
Force operator< to be inlined to reduce place time

xuqinziyue:
#### Description
The change forces operator< to be inlined properly.

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1225

#### Motivation and Context
Original operator< is not inlined by compiler. Since the function is timing critical, force the function to be inlined can reduce place time by ~10% with xc7 tests and ~5% with QoR and Titan benchmark tests

Run through xc7 tests like picosoc and murax and passed QoR and Titan Benchmarks.
Comparison shows a place time reduction.
Comparison also proves there is no impact on quality of results.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/place_delay_model.h
+++ b/vpr/src/place/place_delay_model.h
@@ -6,6 +6,20 @@
 #include "vpr_types.h"
 #include "router_delay_profiling.h"
 
+#ifndef __has_attribute
+#    define __has_attribute(x) 0 // Compatibility with non-clang compilers.
+#endif
+
+#if defined(COMPILER_GCC) && defined(NDEBUG)
+#    define ALWAYS_INLINE inline __attribute__((__always_inline__))
+#elif defined(COMPILER_MSVC) && defined(NDEBUG)
+#    define ALWAYS_INLINE __forceinline
+#elif __has_attribute(always_inline)
+#    define ALWAYS_INLINE __attribute__((always_inline)) // clang
+#else
+#    define ALWAYS_INLINE inline
+#endif
+
 //Abstract interface to a placement delay model
 class PlaceDelayModel {
   public:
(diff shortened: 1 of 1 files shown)


vpr/src/place/place_delay_model.h
```diff
@@ -6,6 +6,20 @@
...
+#if defined(COMPILER_GCC) && defined(NDEBUG)
+#    define ALWAYS_INLINE inline __attribute__((__always_inline__))
+#elif defined(COMPILER_MSVC) && defined(NDEBUG)
+#    define ALWAYS_INLINE __forceinline
+#elif __has_attribute(always_inline)
+#    define ALWAYS_INLINE __attribute__((always_inline)) // clang
+#else
+#    define ALWAYS_INLINE inline
+#endif
+
```
kmurray, line 22: I wonder if it would be better to use [C++ attributes](https://en.cppreference.com/w/cpp/language/attributes) (available since C++11) as a cleaner way to specify this.

For example:
```cpp
[[gnu::always_inline]] friend bool operator<(const t_override& lhs, const t_override& rhs) {
```

That would avoid having to define the ALWAYS_INLINE macro in a cleaner cross-compiler way.

@mithro Any thoughts on this?

vpr/src/place/place_delay_model.h
```diff
@@ -95,9 +109,14 @@ class OverrideDelayModel : public PlaceDelayModel {
...
-            return std::tie(lhs.from_type, lhs.to_type, lhs.from_class, lhs.to_class, lhs.delta_x, lhs.delta_y)
-                   < std::tie(rhs.from_type, rhs.to_type, rhs.from_class, rhs.to_class, rhs.delta_x, rhs.delta_y);
+        //A combination of ALWAYS_INLINE attribute and std::lexicographical_compare
+        //is required for operator< to be inlined by compiler.
+        //Proper inlining of the function reduces place time by around 5%.
+        //For more information: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1225
+        friend ALWAYS_INLINE bool operator<(const t_override& lhs, const t_override& rhs) {
+            const short* left = reinterpret_cast<const short*>(&lhs);
+            const short* right = reinterpret_cast<const short*>(&rhs);
+            return std::lexicographical_compare(left, left + 6, right, right + 6);
```
kmurray, line 119: I think we also need to explicitly state what assumptions this code is making. In particular, I think it is assuming:

- All data members of t_override are shorts
- There is no padding in t_override
- Based on the above, treating t_override's memory layout as an array of shorts is the same as a member-wise comparison

We should also encode these assumptions into some compile-time checks, otherwise we are setting ourselves up for some very challenging to find bugs in the future (when t_override changes).

I'd suggest something like:
```cpp
static_assert(sizeof(t_override) = sizeof(t_override::from_type)
                                   + sizeof(t_override::to_type)
                                    + sizeof(t_override::from_class)
                                    + sizeof(t_override::to_class)
                                    + sizeof(t_override::delta_x)
                                    + sizeof(t_override::delta_y),
              "Expect t_override to have a memory layout equivalent to an array of short (no padding)");
static_assert(sizeof(t_override::from_type) == sizeof(short), "Expect all t_override data members to be shorts");
static_assert(sizeof(t_override::to_type) == sizeof(short), "Expect all t_override data members to be shorts");
static_assert(sizeof(t_override::from_class) == sizeof(short), "Expect all t_override data members to be shorts");
static_assert(sizeof(t_override::to_class) == sizeof(short), "Expect all t_override data members to be shorts");
static_assert(sizeof(t_override::delta_x) == sizeof(short), "Expect all t_override data members to be shorts");
static_assert(sizeof(t_override::delta_y) == sizeof(short), "Expect all t_override data members to be shorts");
```
which would verify there is no padding in the t_override memory layout and that each data member is a short. This would catch potential bugs caused by changing the number of data members in t_override, or the data type of members in t_override.

We could then also modify the comparison to not be explicitly hard coded for 6 data members:
```cpp
constexpr size_t NUM_T_OVERRIDE_MEMBERS = sizeof(t_override) / sizeof(short);
return std::lexicographical_compare(left, left + NUM_T_OVERRIDE_MEMBERS, right, right + NUM_T_OVERRIDE_MEMBERS);
```

xuqinziyue:
> Thanks for putting this together @xuqinziyue!
> 
> One additional thought. Have you tried the `std::tie` based comparison operator in combination with `always_inline`? If that produces similar performance, I think it would be preferable since the std::tie approach would be less brittle to future code changes (since it makes fewer assumptions).
> 
> If that isn't the case I've described how we can have the compiler check that the assumptions the proposed method depends are valid. I've also provided more detailed comments below.

I have tried the combination of std::tie and always_inline but based on my experiment, operator < is not inlined in the case. Thanks for the comments below,  and I will update the pull request with the revisions you suggest shortly.

kmurray:
@xuqinziyue The updates look good. Can you fix the code formatting (`make format`) and update so it passes the CI tests? Once that is done it should be good to merge.

xuqinziyue:
@kmurray I have fixed the formatting issue and have done a rebase. Let me know if there is anything else I should change

kmurray:
Thanks!

PR #1266, opened 2020-04-07
Update parse pass nonvirt geomean cpd

kmurray:
#### Description

* Record geomean non-virtual intra-domain critical path delay metric in parse/golden regression test results (this is the metric which we should be tracking for multi-clock benchmark sets like Titan)
* Report geomean non-virtual intra-domain critical path delay from VPR even if only one clock (for consistency)
* Skip zero cpd clocks when calculating geomean non-virtual intra-domain critical path delay (happens if the clock actually has intra-domain timing paths)
* Update vtr_reg_titan_he to use proper SDC files


mithro:
@HackerFoo

kmurray:
Given the recent changes to master, I'm in the process of regenerating the golden results and will update this PR when they are complete.

PR #1269, opened 2020-04-08
edge_sort_iterator: also add operator> and >=.

hzeller:
Some system implementations of sort are not really good in
only using the equality and less-than operators on iterators; add
operators observed to be needed.

CC: @litghost

### DIFF
--- a/vpr/src/route/rr_graph_storage.cpp
+++ b/vpr/src/route/rr_graph_storage.cpp
@@ -208,6 +208,10 @@ class edge_sort_iterator {
         return lhs.swapper_.idx_ >= rhs.swapper_.idx_;
     }
 
+    friend bool operator<=(const edge_sort_iterator& lhs, const edge_sort_iterator& rhs) {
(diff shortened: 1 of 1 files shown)


vpr/src/route/rr_graph_storage.cpp
```diff
@@ -200,6 +200,14 @@ class edge_sort_iterator {
         return lhs.swapper_.idx_ < rhs.swapper_.idx_;
     }
 
+    friend bool operator>(const edge_sort_iterator& lhs, const edge_sort_iterator& rhs) {
+        return lhs.swapper_.idx_ > rhs.swapper_.idx_;
+    }
+
+    friend bool operator>=(const edge_sort_iterator& lhs, const edge_sort_iterator& rhs) {
```
litghost, line 207: While we are here, does it make sense to define <=?

hzeller: Can't harm. Added.

kmurray:
Thanks!

Issue #1270, opened 2020-04-10
Reorder RR graph nodes to improve caching


HackerFoo:
#### Proposed Behaviour
Reordering RR graph nodes can improve the spatial locality of the data structure.

#### Current Behaviour
No explicit optimization exists.

#### Possible Solution
I have [tested a number of algorithms](https://github.com/HackerFoo/GraphReorderAndConverter) and found that the [`hybrid_bfsdeg`](https://github.com/HackerFoo/GraphReorderAndConverter/blob/master/graph-reordering/reordering/porder.cpp#L93-L140) algorithm works well for some tests in Symbiflow (picosoc and litex_ddr).

#### Context
Hilbert curve ordering has been effective for Symbiflow, but it is achieved by preprocessing the RR graph input. Implementing such an optimization within VPR may accelerate non-Symbiflow use, such as VTR regression tests.

PR #1271, opened 2020-04-10
Add flags to reorder RR graph nodes

HackerFoo:
#### Description
This adds two flags to control RR graph node reordering:

* `--reorder_rr_graph_nodes_threshold`: minimum node count to enable reordering
* `--reorder_rr_graph_nodes_algorithm`:
  * `none` disables reordering
  * `degree_bfs` chooses an ordering based on BFS traversal order and node degree
  * `random_shuffle` chooses a random ordering
* `--reorder_rr_graph_nodes_seed`: PRNG seed for `random_shuffle`

#### Related Issue
#1270 
#1289

#### Motivation and Context
This change was motivated to reduce run time, but it has also been useful to expose a few bugs.

In addition to CI, I have compared QoR using nightly tests with the `degree_bfs` algorithm.
I've added a test using `random_shuffle` to `vtr_reg_basic` tests to catch implicit node order assumptions in the future.

#### Types of changes

#### Checklist:


mithro:
@HackerFoo So what are the QoR results?

HackerFoo:
@mithro While initial tests were positive, the `titan_quick_qor` test has shown a ~30% *increase* in run time with the `bfs_degree` algorithm, so I need to investigate that.

vaughnbetz:
I'd like to get the check_route for non-configurable edges speedup into the code base so it would be good to merge that in (can separate out).

HackerFoo:
These changes do not improve performance for VTR tests, but they do allow experimentation, and improve testing by breaking ordering assumptions.

HackerFoo:
Depends on #1289

vpr/src/base/read_options.cpp
```diff
@@ -163,6 +163,37 @@ struct ParseRouterAlgorithm {
     }
 };
 
+struct ParseNodeReorderAlgorithm {
+    ConvertedValue<e_node_reorder_algorithm> from_str(std::string str) {
+        ConvertedValue<e_node_reorder_algorithm> conv_value;
+        if (str == "degree_bfs")
+            conv_value.set_value(DEGREE_BFS);
```
litghost, line 170: Shouldn't there be a "don't reorder nodes" option?

vpr/src/base/read_options.cpp
```diff
@@ -1735,6 +1766,21 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
...
+    route_grp.add_argument<e_node_reorder_algorithm, ParseNodeReorderAlgorithm>(args.reorder_rr_graph_nodes_algorithm, "--reorder_rr_graph_nodes_algorithm")
+        .help(
+            "Specifies the node reordering algorithm to use.\n"
+            " * degree_bfs: sort by degree and then by BFS\n"
+            " * random_shuffle: a random shuffle\n")
+        .default_value("degree_bfs")
+        .choices({"degree_bfs", "random_shuffle"})
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
+    route_grp.add_argument(args.reorder_rr_graph_nodes_threshold, "--reorder_rr_graph_nodes_threshold")
```
litghost, line 1778: Unclear why this option exists?  Why would you want to only reorder above a certain number?

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -709,3 +709,68 @@ t_rr_graph_view t_rr_graph_storage::view() const {
         vtr::make_const_array_view_id(edge_dest_node_),
         vtr::make_const_array_view_id(edge_switch_));
 }
+
+void t_rr_graph_storage::reorder(const vtr::vector<RRNodeId, RRNodeId>& order,
+                                 const vtr::vector<RRNodeId, RRNodeId>& inverse_order) {
+    {
+        auto old_node_storage = node_storage_;
```
litghost, line 716: Do you need to have 2x copies of the data for the re-order?

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -709,3 +709,68 @@ t_rr_graph_view t_rr_graph_storage::view() const {
...
+    }
+    {
+        auto old_node_first_edge = node_first_edge_;
+        auto old_edge_src_node = edge_src_node_;
+        auto old_edge_dest_node = edge_dest_node_;
+        auto old_edge_switch = edge_switch_;
+        RREdgeId cur_edge(0);
+
+        // Reorder edges by source node
+        for (size_t i = 0; i < node_storage_.size(); i++) {
```
litghost, line 733: Why is this loop required?  Why not simply map src and dest nodes from old -> new and re-sort?

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -709,3 +709,68 @@ t_rr_graph_view t_rr_graph_storage::view() const {
...
+    }
+    {
+        auto old_node_fan_in = node_fan_in_;
+        for (size_t i = 0; i < node_fan_in_.size(); i++) {
+            node_fan_in_[order[RRNodeId(i)]] = old_node_fan_in[RRNodeId(i)];
+        }
+    }
+
+    // Check that edges are still partitioned
+    if (partitioned_) {
```
litghost, line 760: Why not call the method that checks the partitioning instead of re-implementing it?

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -709,3 +709,68 @@ t_rr_graph_view t_rr_graph_storage::view() const {
...
+        }
+    }
+    {
+        auto old_node_fan_in = node_fan_in_;
+        for (size_t i = 0; i < node_fan_in_.size(); i++) {
+            node_fan_in_[order[RRNodeId(i)]] = old_node_fan_in[RRNodeId(i)];
+        }
+    }
+
+    // Check that edges are still partitioned
```
litghost, line 759: This entire method **only** works as written if partitioning is completed.  It could be written in another fashion, however it is not.

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -709,3 +709,68 @@ t_rr_graph_view t_rr_graph_storage::view() const {
         vtr::make_const_array_view_id(edge_dest_node_),
         vtr::make_const_array_view_id(edge_switch_));
 }
+
+void t_rr_graph_storage::reorder(const vtr::vector<RRNodeId, RRNodeId>& order,
+                                 const vtr::vector<RRNodeId, RRNodeId>& inverse_order) {
```
litghost, line 714: I believe the inverse_order argument is not required.  Also, if it is required, why not just compute the inverse_order in the method instead of taking it as an argument?

HackerFoo:
I'm working on #1289 before I can address comments here.

vpr/src/base/read_options.cpp
HackerFoo: Setting the threshold to `-1` achieves this.

vpr/src/base/read_options.cpp
HackerFoo: This is because the reordering algorithm can be expensive, so this allows disabling it for smaller designs that fit in cache. Also to allow disabling reordering entirely.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: Yes. Note that `node_storage_` is written in a different pattern than `old_node_storage` is read. I believe I made this error initially before realizing I needed two copies. The anonymous block scope is used to allow the copy to be discarded as soon as possible.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: Why use an O(N log N) algorithm (sort) instead of an O(N) one?

vpr/src/route/rr_graph_storage.cpp
HackerFoo: I'm not sure I understand this statement. The check is to make sure reordering preserves partitioning, but it does not depend on the edges to be partitioned. It preserves the same relative edge ordering for each source node e.g. edges with the same source node are moved as a block.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: Also, this would not preserve partitioning.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: The `order` is computed as the input, but the `inverse_order` is also used outside this method, so passing in just one of them would require computing the other twice. I could overload the method, but it's only used in one place.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: I've replaced this with an assertion on `validate()`.

vpr/src/route/rr_graph_storage.cpp
litghost: Let me rephrase.  Could you write a version of this loop that doesn't required 2x data?  And instead only needs a temporary?

I agree that the obvious reordering algorithm needs 2x memory.  Could you instead write something that consumes more runtime, but doesn't have the 2x memory hit?  As a concrete example of a solution, a non-stable sorting of the `zip(order, node_storage),sort(key=lambda x: x[0])` sorting on the order would achieve this.  There are other options here.

vpr/src/route/rr_graph_storage.cpp
litghost: It's been a while since I wrote this comment, so I'm not totally confident what I meant either.  However what I believe what I was thinking is that this entire method requires that it already be partitioned_, so really, it should be `VTR_ASSERT(partioned_)` at the top of the method, because it will explode horribly that if that isn't true.

vpr/src/base/read_options.cpp
litghost: I believe that needs to be part of the parameter documentation then

vpr/src/base/place_and_route.cpp
```diff
@@ -353,16 +353,17 @@ int binary_search_place_and_route(const t_placer_opts& placer_opts_ref,
...
                     arch->Directs, arch->num_directs,
-                    &warnings,
-                    router_opts.read_rr_edge_metadata,
-                    router_opts.do_check_rr_graph);
+                    &warnings);
 
     init_draw_coords(final);
+
+    /* Allocate and load additional rr_graph information needed only by the router. */
+    alloc_and_load_rr_node_route_structs();
```
litghost, line 363: It seems weird this got added here, but nothing was removed.  I believe you added this because it is required, but can you explain why a call to these methods didn't get removed?

litghost: A different way of asking this is "why did this work before?"

vpr/src/route/rr_graph_storage.cpp
HackerFoo: Why would it explode? It doesn't care. The only reason the check is there is to make sure it's safe to not re-partition.

litghost: It's internal state variable.  If RrGraphStorage has not been partitioned, then some of the data you are working on won't be populated, and so your code will run off array boundries, etc.

vpr/src/base/place_and_route.cpp
HackerFoo: I believe this worked before because the graph would be rebuilt with a graph similar enough on each iteration that it wasn't noticed before. With `random_shuffle`, it gets shuffled on each iteration.

vpr/src/base/read_options.cpp
HackerFoo: I've reconsidered, and added a `none` option, since it's easier to understand, and actually setting the threshold to `-1` is difficult since the `-1` seems to get interpreted as another option because of the leading dash.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: It might be possible to trade memory usage for runtime, but I don't think it's worthwhile.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: Is this documented somewhere? The reordering algorithm only requires the edge data to be correct. It doesn't care if it shuffles garbage otherwise.

Do you mean that `rr_graph_storage` is unsafe to use without calling `validate()` first?

vpr/src/route/rr_graph_storage.cpp
litghost: We have rr-graphs that consume 22-44 GBs, and doubling that overhead doesn't seem worth avoiding?

vpr/src/base/place_and_route.cpp
HackerFoo: This discussion did help me resolve a bug where the graph isn't restored correctly to output the best route, because the PRNG seed was not fixed for `random_shuffle`. Thanks!

vpr/src/route/rr_graph_storage.cpp
litghost: `rr_graph_storage` is fine to use without first partitioning the edges, in the sense that a failure to conform to the interface will result in an assertion violation.  However you are adding a method, extending the interface, and it possible for there to be a call order where you add nodes and edge data, and then reorder where the assumptions **you** have made about the state of the object are wrong.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: This is an optional feature that can consume more memory, but is not enabled by default.

At ~44GB and a `t_rr_node_data` size of 16 bytes, that means ~2.75 billion nodes, so ~2^(31.3). So, it might take 31x more time to compute. But thats ignoring the cache. Sorting will likely be less cache-friendly.

I have not analyzed this trade-off, because that would take significant time to perform, requiring me to write and test the other way of shuffling nodes, which will likely not pan out. I choose not to do this, unless justified.

Nonetheless, my quick calculations show that this would probably not be a good tradeoff.

Also note that effort was taken to copy things in stages, so the memory requirement is less than 2x the entire graph.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: I've added `validate()` assertions at the beginning and end of `reorder()`.

HackerFoo:
Running `vtr_reg_nightly.titan_quick_qor`, the [maximum memory usage ratios](https://colab.research.google.com/drive/14J_xFbz6SS60nfjsxlnNMlZbiBjhnH_w?usp=sharing) are:

```
degree_bfs        1.000199
none              1.000000
random_shuffle    1.000174
```

Based on this, the memory impact of both algorithms is insignificant.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: I've provided data below.

litghost:
Can you please rebase on master so this PR only has the new changes?

HackerFoo:
> Can you please rebase on master so this PR only has the new changes?

Done.

vpr/src/base/read_options.cpp
```diff
@@ -1897,6 +1932,28 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
...
+        .choices({"none", "degree_bfs", "random_shuffle"})
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
+    route_grp.add_argument(args.reorder_rr_graph_nodes_threshold, "--reorder_rr_graph_nodes_threshold")
+        .help(
+            "Reorder rr_graph nodes to optimize memory layout above this number of nodes.")
+        .default_value("0")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
+    route_grp.add_argument(args.reorder_rr_graph_nodes_seed, "--reorder_rr_graph_nodes_seed")
```
vaughnbetz, line 1951: Should comment why the rr_graph reordering algorithm needs its own seed.  Is it so you can get a deterministically re-ordered routing graph and hence load in a prior routing generated with the same reordering algorithm?  If there isn't a strong reason for the algorithm to have its own seed I'd just use the regular random number generator (and implicitly, its seed).

vpr/src/base/vpr_types.h
```diff
@@ -977,6 +977,12 @@ enum e_router_algorithm {
     TIMING_DRIVEN,
 };
 
+enum e_node_reorder_algorithm {
```
vaughnbetz, line 980: Should put an rr_ in the name of this type -- node_reorder is a bit vague.  Should also comment what this enumerated type is for.

vpr/src/base/vpr_types.h
```diff
@@ -1029,6 +1035,9 @@ enum class e_incr_reroute_delay_ripup {
 constexpr int NO_FIXED_CHANNEL_WIDTH = -1;
 
 struct t_router_opts {
+    e_node_reorder_algorithm reorder_rr_graph_nodes_algorithm = DONT_REORDER;
```
vaughnbetz, line 1038: I'd move these to the end of the router_opts as they're not going to be the most commonly used options, and add a comment in front of the group of 3 saying they control rr-graph node reordering, which can be useful for testing and could optimize memory layout (but hasn't so far).

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -739,3 +739,53 @@ t_rr_graph_view t_rr_graph_storage::view() const {
         vtr::make_const_array_view_id(edge_dest_node_),
         vtr::make_const_array_view_id(edge_switch_));
 }
+
+void t_rr_graph_storage::reorder(const vtr::vector<RRNodeId, RRNodeId>& order,
```
vaughnbetz, line 743: Should comment what this routine does, and exactly what order and inverse_order are (entry
order[old_rr_id] = new_rr_id and so on).

vpr/src/route/rr_graph_util.cpp
```diff
@@ -75,3 +80,76 @@ int seg_index_of_sblock(int from_node, int to_node) {
         return OPEN; //Should not reach here once thrown
     }
 }
+
+void reorder_rr_graph_nodes(const t_router_opts& router_opts) {
```
vaughnbetz, line 84: Comment what the routine does, and any invariants expected (e.g. I assume that the rr-graph should be built before calling this, but there should not be a routing created or loaded as it would depend on the rr_node indices in the current graph).  The rr-graph nodes are renumbered in place. Any other conditions where this shouldn't be called?

vpr/src/route/rr_graph_util.cpp
```diff
@@ -75,3 +80,76 @@ int seg_index_of_sblock(int from_node, int to_node) {
...
+
+    vtr::ScopedStartFinishTimer timer("Reordering rr_graph nodes");
+
+    vtr::vector<RRNodeId, RRNodeId> src_order(v_num); // new id -> old id
+    size_t cur_idx = 0;
+    for (RRNodeId& n : src_order) { // Initialize to [0, 1, 2 ...]
+        n = RRNodeId(cur_idx++);
+    }
+
+    if (router_opts.reorder_rr_graph_nodes_algorithm == DEGREE_BFS) {
```
vaughnbetz, line 100: Should explain what this algorithm is.  Sort first by node out-degree, breaking ties by bfs level from an arbitrarily selected (starting at node 0)?  Or something else?  Also explain the intuition behind the algorithm (why might this be good) and that it currently seems to get the same runtime and memory footprint as the original rr-graph creation order for generated rr-graphs.  Not sure for Artix7 graphs (if you know that worth saying that too).

vpr/src/route/rr_graph_util.cpp
```diff
@@ -75,3 +80,76 @@ int seg_index_of_sblock(int from_node, int to_node) {
...
+        std::shuffle(src_order.begin(), src_order.end(), g);
+    }
+    vtr::vector<RRNodeId, RRNodeId> dest_order(v_num);
+    cur_idx = 0;
+    for (auto u : src_order)
+        dest_order[u] = RRNodeId(cur_idx++);
+
+    graph.reorder(dest_order, src_order);
+
+    // update rr_node_indices
```
vaughnbetz, line 141: I'd explain what this is (update rr_node_indices, which is a mapping structure for fast rr_index lookups).

vpr/src/base/read_options.cpp
HackerFoo: The `--seed` flag help says "Placement random number generator seed" and is in the "placement options" group. I considered making that just a general seed, but I can imagine wanting to try the same reordering with different SA seeds, and vice versa, so I just added another argument.

If this isn't a good reason, I can use the other flag and change the documentation and grouping.

vaughnbetz: I'm OK with leaving it as is then.  Didn't realize we called seed just placement seed; I don't have a really strong opinion about one seed vs. several so if you like this it's fine with me.

vpr/src/route/rr_graph_storage.h
```diff
@@ -435,6 +435,8 @@ class t_rr_graph_storage {
         node_storage_.emplace_back();
         node_ptc_.emplace_back();
     }
+    void reorder(const vtr::vector<RRNodeId, RRNodeId>& order,
```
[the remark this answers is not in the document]
HackerFoo: Done.

vpr/src/route/rr_graph_util.cpp
HackerFoo: Done.

vpr/src/route/rr_graph_util.cpp
HackerFoo: Done.

I mentioned that the `DEGREE_BFS` algorithm could likely be improved. It's just not a priority for me right now, and I'd rather provide a good starting point for whoever has time to do the research, maybe later me.

vpr/src/route/rr_graph_util.cpp
HackerFoo: Done.

vpr/src/route/rr_graph_storage.cpp
HackerFoo: Done.

vpr/src/base/vpr_types.h
HackerFoo: Done.

vpr/src/base/vpr_types.h
HackerFoo: Done.

HackerFoo:
I've added all requested documentation. I'll merge once CI passes.

PR #1275, opened 2020-04-14
Updated RTL tests to cover X and Z

djns99:
Added test coverage for Z into the RTL library

#### Related Issue
TODO

#### Types of changes
- [x] Test coverage
- [x] Bug fixes
#### Checklist:

### DIFF
--- a/libs/librtlnumber/CMakeLists.txt
+++ b/libs/librtlnumber/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.9)
 
 project("librtlnumber")
 
+option(RTL_ALLOW_UNKNOWN_COMPARE "Skips initial check for unknowns in comparison and compares MSB-to-LSB" OFF)
+
+if(RTL_ALLOW_UNKNOWN_COMPARE)
+    add_definitions(-DRTL_ALLOW_UNKNOWN_COMPARE)
+endif()
+
 file(GLOB_RECURSE EXEC_SOURCES main.cpp)
 file(GLOB_RECURSE LIB_SOURCES src/*.cpp)
 file(GLOB_RECURSE LIB_HEADERS src/include/*.hpp)
--- a/libs/librtlnumber/regression_tests/basic_regression_tests.csv
+++ b/libs/librtlnumber/regression_tests/basic_regression_tests.csv
@@ -12,18 +12,24 @@ simple_true,        is_true, 1'b1, pass
 simple_fail,        is_true, 1'b0, fail
 decimal_true,       is_true, 1, pass
 decimal_fail,       is_true, 0, fail
-complex_true,       is_true, 2'b1x, pass
-complex_fail,       is_true, 2'b0x, fail
+complex_true,       is_true, 3'b1xz, pass
+complex_fail,       is_true, 3'b0xz, fail
 unknown_x,          is_true, 1'bx, fail
+unknown_z,          is_true, 1'bz, fail
 large_number_pass,  is_true, 128'h8000_0000_0000_0000, pass
 
 # type test
-simple_x_pass,      is_x,   1'bx,   pass
-simple_x_fail,      is_x,   1'b1,   fail
-simple_z_pass,      is_z,   1'bz,   pass
-simple_z_fail,      is_z,   1'b1,   fail
-simple_unk_pass,    is_unk, 2'b1x,  pass
-simple_unk_fail,    is_unk, 2'b11,  fail
+simple_x_pass,      is_x,   1'bx,    pass
+simple_x_fail,      is_x,   1'b1,    fail
+complex_x_pass,     is_x,   4'bxxxx, pass
+complex_x_fail,     is_x,   4'bzxxx, fail
+simple_z_pass,      is_z,   1'bz,    pass
+simple_z_fail,      is_z,   1'b1,    fail
+complex_z_pass,     is_z,   4'bzzzz, pass
+complex_z_fail,     is_z,   4'bxzzz, fail
+simple_unk_x_pass,  is_unk, 2'bx1,   pass
+simple_unk_z_pass,  is_unk, 2'bz1,   pass
+simple_unk_fail,    is_unk, 2'b10,   fail
 
 # sign test
 simple_is_unsigned_pass,     is_unsigned,    2'b11,  pass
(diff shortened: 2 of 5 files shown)


libs/librtlnumber/src/rtl_int.cpp
```diff
@@ -44,6 +46,9 @@ static compare_bit eval_op(VNumber& a_in, VNumber& b_in) {
     assert_Werr(b_in.size(),
                 "empty 2nd bit string");
 
+    if (a_in.is_dont_care_string() || b_in.is_dont_care_string())
```
djns99, line 49: @jeanlego @aarongraham9 I would be interested in hearing your guys opinion here. Technically the spec say that this is correct. However, this has a performance impact since now we have to loop through the string an extra time.
My questions are:
1) Does the performance matter here, since we are usually only going to be up to 32 bits. Though I imagine this code might be quite hot
2) Do we care about exact conformance to the spec
3) What do you think about having a has_unknown bool in VNumber so we do not have to loop.

(Apologies if this is a duplicate, I lost my first comment on my screen)

libs/librtlnumber/verify_librtlnumber.sh
```diff
@@ -98,7 +98,7 @@ for INPUT in ${0%/*}/regression_tests/*.csv; do
 		then
 			echo "--- PASSED == $TEST_LABEL ( ${OUTPUT_AND_RESULT} ) "
 
-		elif [ "pass" == "$(${0%/*}/rtl_number is_true $(${0%/*}/rtl_number ${OUTPUT_AND_RESULT} == ${EXPECTED_RESULT}))" ]
+		elif [ "pass" == "$(${0%/*}/rtl_number is_true $(${0%/*}/rtl_number ${OUTPUT_AND_RESULT} === ${EXPECTED_RESULT}))" ]
```
jeanlego, line 101: I think at this time, we would be better served with just comparing the output of the === operation.
You could, in turn, take out the is_true (originally temporary) that is only used for the tests.

libs/librtlnumber/src/rtl_int.cpp
aarongraham9: Yeah I think this is reasonable. Go for it! :)

libs/librtlnumber/verify_librtlnumber.sh
aarongraham9: Agreed! :smiley:

libs/librtlnumber/src/rtl_int.cpp
```diff
@@ -261,17 +269,17 @@ VNumber V_BITWISE_XOR(VNumber& a) {
 }
 
 VNumber V_BITWISE_NAND(VNumber& a) {
-    VNumber to_return = a.bitwise_reduce(l_nand);
+    VNumber to_return = a.bitwise_reduce(l_and).invert();
```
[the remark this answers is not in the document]
djns99: l_nand is used for the non-reduction bitwise operators. This behaviour is identical to inverting after so I would keep that

libs/librtlnumber/verify_librtlnumber.sh
djns99: Surely you need to test is_true? I assume that is what you use to evaluate ternary conditionals?

djns99:
```                                
VNumber eval = V_CASE_EQUAL(*(child_condition->types.vnumber), *(item->children[0]->types.vnumber));
if (V_TRUE(eval)) 
```
In CASE statements as well as IF and IF_Q

libs/librtlnumber/regression_tests/basic_regression_tests.csv
```diff
@@ -89,64 +103,89 @@ Bitwise-Nand,   5'b1010,	~&, 5'b1000,	5'b10111
...
 Logical-less,           4'b0000,    <,  4'b0000,    1'b0	
 Logical-less-1,			4'b0000,	<,	4'b0001,	1'b1
 Logical-less-2,         4'b0001,    <,  4'b0000,    1'b0
 Logical-less-3,         4'bxxxx,    <,  4'b0001,    1'bx
 Logical-less-4,         4'b0001,    <,  4'bxxxx,    1'bx
-Logical-less-5,         4'b0xxx,    <,  4'b1001,    1'b1
-Logical-less-6,         4'b1001,    <,  4'b0xxx,    1'b0
-Logical-less-7,         4'sb1001,   <,  4'sb0xxx,   1'b1
-Logical-less-8,         4'sb0xxx,   <,  4'sb1001,   1'b0
+Logical-less-5,         4'b0xxx,    <,  4'b1001,    1'bx
```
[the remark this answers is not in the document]
djns99: Guess that can be something for my research to cover

libs/librtlnumber/CMakeLists.txt
```diff
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.9)
 
 project("librtlnumber")
 
+option(RTL_ALLOW_UNKNOWN_COMPARE "Skips initial check for unknowns in comparison and compares MSB-to-LSB" OFF)
+
+if(RTL_ALLOW_UNKNOWN_COMPARE)
+    add_definitions(-DRTL_ALLOW_UNKNOWN_COMPARE)
+endif()
+
```
jeanlego, line 10: I think it is better off being a runtime variable (or at least template)
Although it will change a few of the call spot in Odin, if we are to consume it with the simulator and the synthesizer, they have to use it differently. Since the simulator is an integral part of the tool (than can and should be changed eventually) We cannot compile this both way.

djns99: If you think that would be valuable I can. Personally, I can't see a use case where we would want it compiled in both situations.

djns99: I'm not sure I understand what you mean by needing to consume it differently in synthesizer and simulator?

jeanlego: I don't think it will be an issue for the near future so I can merge.

libs/librtlnumber/verify_librtlnumber.sh
jeanlego: It does not hurt to keep it so may as well leave it, but I was talking about `is_true` in the driver for the test harness, I don't know that we use this any more:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/librtlnumber/main.cpp#L147-L155

PR #1277, opened 2020-04-14
flow: Record additional build info in parse results

kmurray:
To aid in debugging and understanding performance/QoR differences across different infrastructure/platforms, parse out additional metrics about how VPR was built.


mithro:
@kmurray Can we get the git hash into that output somehow?

mithro:
@kmurray Compiler version and flags used for compilation might be useful too...

kmurray:
> Can we get the git hash into that output somehow?

That's already included in the vpr_revision field

> Compiler version and flags used for compilation might be useful too...

Yep that's included in the vpr_compiler, vpr_compiled, vpr_build_info fields added by this PR.

PR #1278, opened 2020-04-14
Travis & gcc10 fixes

### LINKED ISSUE #1263: Make travis.yml file configuration valid.
Currently the `.travis.yml` file has a bunch of issues. See https://travis-ci.com/github/verilog-to-routing/vtr-verilog-to-routing/builds/158302571/config for details. See the travis documentation at https://docs.travis-ci.com/user/build-config-validation/ for more information.

```
jobs.include: duplicate key: name
jobs.include: duplicate key: script
jobs.include: duplicate key: env
jobs.include: duplicate key: script
jobs.include: duplicate key: env
jobs.include: duplicate key: name
```

Once these errors are fixed, we should enable the `version: ~> 1.0` so that it stays valid.

kmurray:
* Fixes for building with gcc-10
* Fixes for Travis infrastructure (should close #1263)
* ~~Add Travis gcc-10 and clang-9/10 build tests~~
  * Turns out gcc-10 and clang-9/clang-10  aren't available with the OS version we're using in Travis

### DIFF
--- a/libs/libvtrutil/src/vtr_ndmatrix.h
+++ b/libs/libvtrutil/src/vtr_ndmatrix.h
@@ -36,7 +36,6 @@ class NdMatrixProxy {
     NdMatrixProxy<T, N>& operator=(const NdMatrixProxy<T, N>& other) = delete;
 
     const NdMatrixProxy<T, N - 1> operator[](size_t index) const {
-        VTR_ASSERT_SAFE_MSG(index >= 0, "Index out of range (below dimension minimum)");
         VTR_ASSERT_SAFE_MSG(index < dim_sizes_[0], "Index out of range (above dimension maximum)");
         VTR_ASSERT_SAFE_MSG(dim_sizes_[1] > 0, "Can not index into zero-sized dimension");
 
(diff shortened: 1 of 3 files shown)


libs/libvtrutil/src/vtr_ndmatrix.h
```diff
@@ -36,7 +36,6 @@ class NdMatrixProxy {
     NdMatrixProxy<T, N>& operator=(const NdMatrixProxy<T, N>& other) = delete;
 
     const NdMatrixProxy<T, N - 1> operator[](size_t index) const {
-        VTR_ASSERT_SAFE_MSG(index >= 0, "Index out of range (below dimension minimum)");
```
mithro, line 39: Why does this assert go away?

kmurray: GCC warns (correctly) that size_t can't be less than zero as it is an unsigned type, so the assertion must always be true (and the assertion is redundant).

(That assertion actually came as legacy from an earlier version of the NdMatrix which allowed indicies to start from non-zero values.)

mithro: I was sure that size_t was signed but it seems like that is false -- https://stackoverflow.com/questions/1089176/is-size-t-always-unsigned
> Yes, size_t is guaranteed to be an unsigned type. According to the standard it is unsigned, however I recall that some older implementations used a signed type for the typedef.

Depending on how paranoid we want to be -- maybe we just need a compiler assert somewhere that `size_t` is unsigned?

kmurray: > I was sure that size_t was signed but it seems like that is false

Perhaps you were thinking of `ssize_t` from POSIX?

> maybe we just need a compiler assert somewhere that `size_t` is unsigned?

I believe that is guaranteed by the C++ standard so it seems redundant (see: https://en.cppreference.com/w/cpp/types/size_t).

mithro: Looks like you are right! Thanks.

PR #1276, opened 2020-04-14
infra: Fix Kokoro nightly run-time failures

kmurray:
I've observed a number of run-time failures on the Kokoro nightly regression test, despite having fully updated the nightly regression test results, and the same tests passing when run locally.

For example see: https://source.cloud.google.com/results/invocations/354130ae-411d-454b-bdb0-6d66751a702e/targets/foss-fpga-tools%2Fverilog-to-routing%2Fupstream%2Fpresubmit%2Fnightly/log
where we see everything passes run-time and quality except for route time on some tests:
```
regression_tests/vtr_reg_nightly/complex_switch...[Fail]
 k4_N8_topology-0.85sL2-0.15gL4-on-cb-off-sb_22nm_22nm.xml/mcml.v/common crit_path_route_time: golden = 81.87 result = 2124.87
```

Diffing the offending [VPR on Kokoro log file](https://storage.cloud.google.com/vtr-verilog-to-routing/artifacts/prod/foss-fpga-tools/verilog-to-routing/upstream/presubmit/nightly/577/20200413-122350/vtr_flow/tasks/regression_tests/vtr_reg_nightly/complex_switch/run001/k4_N8_topology-0.85sL2-0.15gL4-on-cb-off-sb_22nm_22nm.xml/mcml.v/common/vpr.out) with one run locally 
([vpr.out.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4475704/vpr.out.txt)) shows identical results and router operations, but significantly longer run-times on Kokoro.

Checking the amount of free memory as the tests run shows:
```
$ grep 'Mem:' kokoro-sisyphus_resultstore_prod_foss-fpga-tools_verilog-to-routing_upstream_presubmit_nightly_354130ae-411d-454b-bdb0-6d66751a702e_build.log | awk '{print $4}' | sort -h | uniq -c
      1 512M
      1 516M
      1 547M
      1 592M
      2 595M
      1 596M
      1 599M
      3 600M
      1 602M
      1 641M
      1 836M
      1 837M
      1 838M
      3 1.1G
      1 1.2G
[...]
     12 35G
     15 36G
      2 37G
     16 38G
[...]
      1 85G
      2 86G
      1 88G
```
Indicating that at points the machine is nearing memory exhaustion (<1G free).

And begins swapping to disk, which shows up as a non-zero amount of swap usage:
```
[...]

PR #1281, opened 2020-04-15
Add vtr_reg_strong with sanitizers to Kokoro

kmurray:
#### Description
Building on PR #1276, this adds an additional Kokoro test, which runs vtr_reg_strong with a [sanitizer enabled](https://docs.verilogtorouting.org/en/latest/dev/developing/#sanitizers) build of VTR.

Since vtr_reg_strong is out main 'feature check' test it would be good to have it checked with sanitizers for memory issues/leaks etc.

We already have a sanitized version of the vtr_reg_basic test running in TravisCI, but the combination of longer compilation times with sanitizers enabled, and longer test time make it impossible to run vtr_reg_strong with sanitizers on TravisCI.

### DIFF
--- a/.github/kokoro/continuous/nightly.cfg
+++ b/.github/kokoro/continuous/nightly.cfg
@@ -40,6 +40,13 @@ env_vars {
   value: "vtr_reg_nightly"
 }
 
+#Options for run_reg_test.pl
+# -show_failures: show tool failures in main log output
+env_vars {
+  key: "VTR_TEST_OPTIONS"
+  value: "-show_failures"
+}
+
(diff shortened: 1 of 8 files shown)


litghost:
New env vars and new configs both need configuration changes on the kokoro side.

kmurray:
> New env vars and new configs both need configuration changes on the kokoro side.

Yep that makes sense. If you're happy with these changes I think it'd be good to get this in.

litghost:
In 20 minutes the new config should be prepare.  Just add a "kokoro:force-run" and it should go.

PR #1282, opened 2020-04-15
Fix compilation of vtr_vector for specializations without data().

hzeller:
The using storage::data method forwarding only works if there is such
a method defined in the base class, which might not be true for all
specializations (I ran into this with a `vector<bool>` specialization).

This can be fixed by forwarding the method in a traditional way with
an inline implementation; then, the compiler will defer compiling it
to when actually needed.


hzeller:
cc @litghost

kmurray:
Thanks for the fix!

@HackerFoo I think this fixes the last outstanding item in #1246

PR #1284, opened 2020-04-20
Placement bb cost computation fix

mkurc-ant:
#### Description
If it happens that a routing graph contains zero-capacity channels then the computation of bounding-box placement cost yields in +infs. This may happen when VPR operates on a rr graph provided by a user. This makes the placement optimizer not optimize anything.

The cost is computed as inverse of a channel capacity. This PR adds a condition which catches zeros, logs a warnings and substitutes them with ones. This way channels of capacity 0 and 1 will have the same placement cost.

#### Motivation and Context
Solves poor timing-driven placement results in SymbiFlow on the Artix 200T architecture.

The placement algorithm was run twice: first in its unmodified version, second with the proposed modification. Each time using the same packed netlist. The results were assessed through the worst critical path delay estimated after placement:

| Design | Original | Proposed |
| --- | --- | --- |
| picosoc | 62.61ns | 22.75ns |
| murax | 36.90ns | 15.53ns |
| LiteX SoC (VexRiscv + DDR) | 41.83ns | 19.49ns |
| LiteX SoC (Linux capable VexRiscv + DDR) | 60.66ns | 20.41ns |

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/place.cpp
+++ b/vpr/src/place/place.cpp
@@ -2348,6 +2348,11 @@ static void alloc_and_load_for_fast_cost_update(float place_cost_exp) {
 
     for (size_t high = 0; high < device_ctx.grid.height(); high++)
         for (size_t low = 0; low <= high; low++) {
+            if (chanx_place_cost_fac[high][low] == 0.0f) {
+                VTR_LOG_WARN("CHANX place cost fac is 0 at %d %d\n", high, low);
+                chanx_place_cost_fac[high][low] = 1.0f;
+            }
+
             chanx_place_cost_fac[high][low] = (high - low + 1.)
                                               / chanx_place_cost_fac[high][low];
             chanx_place_cost_fac[high][low] = pow((double)chanx_place_cost_fac[high][low], (double)place_cost_exp);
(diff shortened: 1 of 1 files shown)


kmurray:
@mkurc-ant Nice find! This would certainly mess-up the placer's cost function of there are empty channels.

kmurray:
@vaughnbetz Any thoughts about this? Is using 1.0 the correct choice for empty channels?

vpr/src/place/place.cpp
```diff
@@ -2348,6 +2348,11 @@ static void alloc_and_load_for_fast_cost_update(float place_cost_exp) {
 
     for (size_t high = 0; high < device_ctx.grid.height(); high++)
         for (size_t low = 0; low <= high; low++) {
+            if (chanx_place_cost_fac[high][low] == 0.0f) {
+                VTR_LOG_WARN("CHANX place cost fac is 0 at %d %d\n", high, low);
+                chanx_place_cost_fac[high][low] = 1.0f;
```
vaughnbetz, line 2353: Looks good.  I agree 1.0 is a good value.
Please add a comment here though: since we will divide the wiring cost by the average channel capacity between high and low, having only 0 width channels will result in infinite wiring capacity normalization factor, and extremely bad placer behaviour. Hence we change this to a small (1 track) channel capacity instead.

vaughnbetz:
Looks good; I added one review request for a comment so this is understandable for posterity though. I agree 1 is a good value to use here (small compared to regular channel widths).

vpr/src/place/place.cpp
mkurc-ant: Ok, I've added the comment.

kmurray:
Thanks @mkurc-ant for finding this and putting a fix together!

mkurc-ant:
@kmurray No problem, thanks for quick review & merge.

vaughnbetz:
Thanks for finding and fixing!

PR #1287, opened 2020-04-22
Adding placer debug user interface

MohamedElgammal:
#### Description
This PR adds a debugging user interface that enables some features to visualize move generation for placement as follows:
 * Placement pauses at each proposed move
 * The UI highlight the old location of the block in Gold, the new location of the block in GREEN, the fanin and the fanout
 * Print out the move type, move outcome and delta cost in the status bar

The implementation is added under "VTR_ENABLE_DEBUG_LOGGING" compilation flag to avoid any overhead in case of release mode.
Moreover, another flag is defined in the code (move_utils.cpp) to enable this debugging feature as the compilation flag is used for many other debugging logging. The flag is called ""f_placer_debug".

It would be better to group the debugging utilities of each stage (packing-placement-routing) under a flag for this stage. All these flags would be checked only if VTR_ENABLE_DEBUG_LOGGING is on.

#### Motivation and Context

This change enables more visualization of how the placer is working and giving more insights into the move generation process

- run_vtr_reg_nightly to make sure that no overhead is added if the flag is disabled
- test that the UI is working for multiple VPR benchmarks
- test that the feature is disabled if the compilation or the code flag is disabled

#### Types of changes

#### Checklist:


mithro:
FYI - I think this will be of interest to you -- @mkurc-ant @HackerFoo

vpr/src/draw/draw.cpp
```diff
@@ -976,16 +980,56 @@ static void drawplace(ezgl::renderer* g) {
...
                     logical_block_type = cluster_ctx.clb_nlist.block_type(bnum);
                 } else {
                     block_color = get_block_type_color(device_ctx.grid[i][j].type);
                     block_color = lighten_color(block_color, EMPTY_BLOCK_LIGHTEN_FACTOR);
-
+    
                     auto tile_type = device_ctx.grid[i][j].type;
                     logical_block_type = pick_best_logical_type(tile_type);
                 }
+#endif
```
kmurray, line 1032: Generally the drawing code isn't performance critical, so I think it'd be better to take this out of the #ifdef to ensure it is always compiled.

The run-time check for f_placer_debug shouldn't be a problem.

vpr/src/draw/draw.cpp
```diff
@@ -976,16 +980,56 @@ static void drawplace(ezgl::renderer* g) {
...
                     block_color = draw_state->block_color(bnum);
                     logical_block_type = cluster_ctx.clb_nlist.block_type(bnum);
                 } else {
                     block_color = get_block_type_color(device_ctx.grid[i][j].type);
                     block_color = lighten_color(block_color, EMPTY_BLOCK_LIGHTEN_FACTOR);
-
+    
                     auto tile_type = device_ctx.grid[i][j].type;
                     logical_block_type = pick_best_logical_type(tile_type);
                 }
```
kmurray, line 1031: This code is duplicated three times in the above code block.

We should try to restructure the code to avoid duplicating it.

vpr/src/place/move_utils.h
```diff
@@ -4,6 +4,13 @@
 #include "move_transactions.h"
 #include "compressed_grid.h"
 
+//Run-time flag to control when placer debug information is printed
+//Note : only enables debug output if compiled with VTR_ENABLE_DEBUG_LOGGING is defined
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+extern bool f_placer_debug;
+#endif
```
kmurray, line 11: Rather than exposing this as a global variable in the header, I'd suggest wrapping it in an accessor function. something like: 
```cpp
bool placer_debug_enabled(); //In .h
bool placer_debug_enabled() { return f_placer_debug; } //In .cpp
```

vpr/src/draw/draw.cpp
```diff
@@ -976,16 +980,56 @@ static void drawplace(ezgl::renderer* g) {
                 //Determine the block color
                 ezgl::color block_color;
                 t_logical_block_type_ptr logical_block_type = nullptr;
+#ifdef  VTR_ENABLE_DEBUG_LOGGING
+                if(f_placer_debug){ 
+                    t_pl_loc curr_loc;
+                    curr_loc.x = i;
+                    curr_loc.y = j;
+                    auto it = std::find_if(draw_state->colored_blocks.begin(),draw_state->colored_blocks.end(),[&curr_loc](const std::pair<t_pl_loc, ezgl::color>& a){return (a.first.x==curr_loc.x && a.first.y==curr_loc.y);});
```
kmurray, line 988: This is a bit awkward, and requires a linear search through the colored block.

Perhaps we should re-consider how this is structured.

Why not store a color for *all* blocks
```cpp
vtr::vector<ClusteredBlockId,ezgl::color> colored_blocks; //Color of each block in the netlist
```
And use a sentinel color (e.g. perhaps define ``ezgl::color DEFAULT_BLOCK_COLOR = ezgl::WHITE``) to mark blocks which aren't highlighted.

For instance the setup code would then be:
```cpp
//All blocks initially use default color (which will be determined by the block-type)
colored_blocks = vtr::vector<ClusteredBlockId,ezgl::color> (clb_nlist.blocks().size(), DEFAULT_BLOCK_COLOR);
```

And the check here would become:
```cpp
if (f_placer_debug && draw_state->colored_blocks[bnum] != DEFAULT_BLOCK_COLOR) {
    block_color = draw_state->colored_blocks[bnum];
} else { //Draw with default block type based-color
    block_color = get_block_type_color(device_ctx.grid[i][j].type);
    block_color = lighten_color(block_color, EMPTY_BLOCK_LIGHTEN_FACTOR);
}
//Draw the block in the specified color
```
which would be much simpler and also more flexible (other code, such as the placer, only needs to mark the block colors).

vpr/src/place/move_utils.cpp
```diff
@@ -5,6 +5,12 @@
 
 #include "vtr_random.h"
 
+
+//Run-time flag to control when placer debug information is printed
+//Note : only enables debug output if compiled with VTR_ENABLE_DEBUG_LOGGING is defined
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+bool f_placer_debug = true;
+#endif
```
kmurray, line 13: This doesn't need to be in the `#ifdef` if we just default it to `false`.

We also probably want to make a command-line option to enable placer debugging (which only does something if compiled with VTR_ENABLE_DEBUG_LOGGING, and gives a warning if specified but not compiled with the option).

vpr/src/place/place.cpp
```diff
@@ -1350,6 +1359,32 @@ static e_move_result try_swap(float t,
...
+        msg += vtr::string_fmt(", Delta_cost: %1.6f (bb_delta_cost= %1.5f , timing_delta_c= %6.1e)", delta_c, bb_delta_c, timing_delta_c);
+        
+        auto& cluster_ctx = g_vpr_ctx.clustering();
+        
+        deselect_all();
+        draw_highlight_blocks_color(cluster_ctx.clb_nlist.block_type(blocks_affected.moved_blocks[0].block_num), blocks_affected.moved_blocks[0].block_num); 
+        draw_state->colored_blocks.clear();
+        
+        draw_state->colored_blocks.push_back(std::make_pair(blocks_affected.moved_blocks[0].old_loc, blk_GOLD));
+        draw_state->colored_blocks.push_back(std::make_pair(blocks_affected.moved_blocks[0].new_loc, blk_GREEN));
```
kmurray, line 1383: Rather than reaching into  the drawing code and manipulating `draw_state` directly, I'd suggest wrapping these operations in an API. 

Generally we try to avoid coupling distince parts of the code base together to help maintainability. For instance, in this case, if someone ever refactors the drawing code and changes `draw_state`, they'll also need to update the placer (and vice versa). It also makes this the only code outside of `vpr/draw` which includes anything other than `draw.h`.

I'd suggest re-working this to put these operations behind and API in `draw.h` and removing the dependency on the other `draw_*.h` header files. For instance, by defining a function like:
```cpp
void set_draw_block_color(ClusterBlockId blk, ezgl::color color); //In draw.h
```

```cpp
void set_draw_block_color(ClusterBlockId blk, ezgl::color color) { //In draw.cpp
    draw_state->block_colors[blk] = color;
}
```

```cpp
//In place.cpp

ezgl::color OLD_BLOCK_LOCATION_COLOR = ezgl::GOLD;
ezgl::color NEW_BLOCK_LOCATION_COLOR = ezgl::GREEN;

set_block_color(blk, OLD_BLOCK_LOCATION);
set_block_color(blk, NEW_BLOCK_LOCATION);
//...
```

MahshadJ:
Added all modified files that are related to breakpoints. Placer breakpoints are mostly done, router breakpoints haven't been implemented yet.

vpr/src/draw/breakpoint.h
```diff
@@ -0,0 +1,81 @@
+#ifndef BREAKPOINT_H
+#define BREAKPOINT_H
+
+#include <vector>
+#include <string>
+
+#include "move_transactions.h"
+#include "expr_eval.h"
+
+class breakpoint {
```
kmurray, line 10: The convention we follow is that classes should be capitalized (i.e. `Breakpoint`).

libs/libarchfpga/src/expr_eval.cpp
```diff
@@ -2,13 +2,17 @@
...
+
 #include <string>
 #include <sstream>
 
 using std::stack;
 using std::string;
 using std::stringstream;
 using std::vector;
+current_information current_info_e;
+bool is_breakpoint = false;
```
kmurray, line 14: The expression evaluator is general purpose code, so it likely shouldn't have any context specific behavior (such as whether it is evaluating a breakpoint or not).

vpr/src/draw/draw.cpp
```diff
@@ -154,13 +158,38 @@ static void set_block_text(GtkWidget* widget, gint /*response_id*/, gpointer /*d
 static void clip_routing_util(GtkWidget* widget, gint /*response_id*/, gpointer /*data*/);
 static void run_graphics_commands(std::string commands);
```
vaughnbetz, line 160: draw.cpp is pretty big.  Could these functions be moved to another file (e.g. draw_debug.cpp)?

vpr/src/draw/draw.cpp
```diff
@@ -154,13 +158,38 @@ static void set_block_text(GtkWidget* widget, gint /*response_id*/, gpointer /*d
...
+bool valid_expression(std::string exp);
+
 /************************** File Scope Variables ****************************/
 
 //The arrow head position for turning/straight-thru connections in a switch box
 constexpr float SB_EDGE_TURN_ARROW_POSITION = 0.2;
 constexpr float SB_EDGE_STRAIGHT_ARROW_POSITION = 0.95;
 constexpr float EMPTY_BLOCK_LIGHTEN_FACTOR = 0.20;
 
+//debugger global variables
```
vaughnbetz, line 187: Better to group in a struct or class.

vpr/src/draw/draw.cpp
```diff
@@ -990,6 +1019,41 @@ static void drawplace(ezgl::renderer* g) {
                 //Determine the block color
                 ezgl::color block_color;
                 t_logical_block_type_ptr logical_block_type = nullptr;
+#    ifdef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 1022: Needs a comment on what this does and should be moved into a subroutine.

vpr/src/draw/draw.cpp
```diff
@@ -3997,6 +4066,577 @@ void net_max_fanout(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data
     application.refresh_drawing();
 }
 
+//DEBUGGER SETTINGS
```
vaughnbetz, line 4069: This new code all looks weakly coupled to draw.cpp, which is good, but it means that it shouldn't be hard to move it into a new draw_debug.cpp file and that should be done.

vpr/src/draw/draw.cpp
```diff
@@ -3997,6 +4066,577 @@ void net_max_fanout(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data
...
+    if (valid_expression(expr)) {
+        draw_state->list_of_breakpoints.push_back(breakpoint("bt_expression", expr));
+        std::string bpDescription = "Breakpoint at " + expr;
+        add_to_bpList(bpDescription);
+    } else
+        invalid_entry_window("Invalid expression");
+}
+
+//window that pops up when an entry is not valid
+void invalid_entry_window(std::string error) {
```
vaughnbetz, line 4470: Might want to add _bp_ somewhere in this and similar function names (best would probably be a bp_ prefix) to make them more distinct.  We may have multiple invalid entries types eventually in the gui.

vpr/src/draw/draw.cpp
```diff
@@ -3997,6 +4066,577 @@ void net_max_fanout(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data
...
+    gtk_grid_attach((GtkGrid*)grid, curr_info, 0, 1, 1, 1);
+
+    //info grid
+    GtkWidget* info_grid = gtk_grid_new();
+    gtk_widget_set_margin_left(info_grid, 30);
+    gtk_widget_set_margin_right(info_grid, 30);
+    gtk_widget_set_margin_bottom(info_grid, 20);
+
+    //images
+    GtkWidget* m = gtk_image_new_from_file("src/draw/m.png");
```
vaughnbetz, line 4533: Pretty short names.  Should comment what they are and more descriptive file names would be better.

vpr/src/draw/draw.cpp
```diff
@@ -3997,6 +4066,577 @@ void net_max_fanout(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data
...
+    gtk_widget_set_halign(button, GTK_ALIGN_CENTER);
+
+    gtk_grid_attach((GtkGrid*)grid, button, 0, 3, 1, 1);
+    g_signal_connect(button, "clicked", G_CALLBACK(ok_close_window), window);
+
+    gtk_container_add(GTK_CONTAINER(window), grid);
+    gtk_widget_show_all(window);
+}
+
+//closess the "invalid entry" window
```
vaughnbetz, line 4593: closess (typo)

vpr/src/draw/draw.cpp
```diff
@@ -3997,6 +4066,577 @@ void net_max_fanout(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data
...
+void ok_close_window(GtkWidget* /*widget*/, GtkWidget* window) {
+    gtk_window_close((GtkWindow*)window);
+}
+
+//checks if an expression is valid by checking the order of operators
```
vaughnbetz, line 4598: Explain what the exp is (+,-, && etc. I think).  Seems like it would be a better fit and more generally useful in the expr_eval.cpp file, and should be moved there.

vpr/src/draw/draw.cpp
```diff
@@ -3997,6 +4066,577 @@ void net_max_fanout(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data
...
+bool valid_expression(std::string exp) {
```
vaughnbetz, line 4599: Need to comment what this does and what it is used for.  It's not clear to me what ops being 0 or 1 (magic numbers) means. Might also be better with enumerated constants instead of 1 or 0, but I can't really say since I don't know what this routine is doing.

vpr/src/draw/draw.h
```diff
@@ -83,6 +83,13 @@ float get_net_alpha();
 
 ezgl::color get_block_type_color(t_physical_tile_type_ptr type);
 
+void breakpoint_info_window(std::string bpDescription, current_information current_info_d);
+
+struct open_windows {
```
vaughnbetz, line 88: Again, I think this could be moved into a debug_breakpoints.h. If open_windows is intended only for the breakpoints code, it should probably be moved inside a breakpoints state class or struct.

vpr/src/place/move_utils.h
vaughnbetz: Agreed.

vpr/src/place/place.cpp
```diff
@@ -44,6 +44,16 @@
...
 
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+#    include "draw_types.h"
+#    include "draw_global.h"
+#    include "draw_color.h"
+#    include "breakpoint.h"
+//map of the available move types and their corresponding type number
+std::map<int, std::string> available_move_types = {
+    {0, "Uniform"}};
+#endif
```
vaughnbetz, line 55: Enumerated type for the move types would be better than magic numbers like 0.

vpr/src/place/place.cpp
```diff
@@ -1218,6 +1234,13 @@ static bool update_annealing_state(t_annealing_state* state,
                                    const t_placer_costs& costs,
                                    const t_placer_opts& placer_opts,
                                    const t_annealing_sched& annealing_sched) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    if (draw_state->list_of_breakpoints.size() != 0) {
+        //update temperature in the current information variable
+        current_info_p.temp_count++;
```
vaughnbetz, line 1240: Need to move current_info into a breakpoint state class with a better name.  That class should be accessible from draw_state, or you should have a different get_bp_state() call to get a pointer to the top-level data class (either could work, and either avoids creating a new global variable that is accessible from anywhere which is bad).

vpr/src/place/place.cpp
vaughnbetz: Agreed, that makes sense.
I also think the code for DEBUG_LOGGING here is big enough that the whole straight-line code segment should be moved into one or a few subroutines and called.  That makes it easier for someone reading the code in the future to see high-level code like:
check_breakpoints (...);
draw_move (...);
That immediately gives a better idea of what's going on, and lets someone decide whether to read further into those routines or skip over them, depending on what he/she is trying to figure out.

vpr/src/place/place.cpp
```diff
@@ -3004,3 +3072,10 @@ static void init_annealing_state(t_annealing_state* state,
 bool placer_needs_lookahead(const t_vpr_setup& vpr_setup) {
     return (vpr_setup.PlacerOpts.place_algorithm == PATH_TIMING_DRIVEN_PLACE);
 }
+
+//transforms the vector moved_blocks to a vector of ints and adds it in current_info_p
+void transform_blocks_affected(t_pl_blocks_to_be_moved blocksAffected) {
+    current_info_p.blocks_affected_vector.clear();
+    for (size_t i = 0; i < blocksAffected.moved_blocks.size(); i++)
+        current_info_p.blocks_affected_vector.push_back(size_t(blocksAffected.moved_blocks[i].block_num));
```
vaughnbetz, line 3080: Why do you need the size_t conversion?  Whatever type block_num is should be OK (I think); if not comment why you need to convert.

vpr/src/route/route_timing.cpp
```diff
@@ -159,6 +163,10 @@ static void prune_unused_non_configurable_nets(CBRR& connections_inf);
 static void init_net_delay_from_lookahead(const RouterLookahead& router_lookahead,
                                           ClbNetPinsMatrix<float>& net_delay);
 
+//debug related functions
+void update_router_info_and_check(char type, int net_id);
```
vaughnbetz, line 167: and_check part seems vague.  update_router_info_and_check_bp would be better.  Or route_check_bp ().

vpr/src/route/route_timing.cpp
```diff
@@ -1982,3 +1995,26 @@ static void init_net_delay_from_lookahead(const RouterLookahead& router_lookahea
         }
     }
 }
+
+//updates router iteration information and checks for breakpoints
+//stops if a breakpoint is encountered
+void update_router_info_and_check(char type, int net_id) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    if (draw_state->list_of_breakpoints.size() != 0) {
+        if (type == 'r')
```
vaughnbetz, line 2004: enumerated type (BP_ROUTE_ITER and BP_NET_ID) would be more clear than 'r' and 'n' as flags.

vpr/src/route/route_timing.cpp
```diff
@@ -1982,3 +1995,26 @@ static void init_net_delay_from_lookahead(const RouterLookahead& router_lookahea
         }
     }
 }
+
+//updates router iteration information and checks for breakpoints
```
vaughnbetz, line 1999: I like that you wrapped the router code up in a single routine.
Comment should say more precisely what router information (iteration count or net id being re-routed) is updated.  Should also say if this should be updated at the start of an iteration and before a net is re-routed, or after a net is re-routed.  (I.e. what is the convention?)

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -2,8 +2,14 @@
 #include "vtr_error.h"
 #include "vtr_util.h"
 #include "vtr_math.h"
+
 #include <string>
 #include <sstream>
+#include <iostream>
+
+current_information current_info_e;
```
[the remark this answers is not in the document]
vaughnbetz: After looking through the code, I think you can remove these global variables and just pass in a struct or class to the expression evaluator with this additional info.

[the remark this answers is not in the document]
vaughnbetz: Add a high-level comment of what this file is for: expression evaluation, with following operators (list them) supported; this can reduce expressions to numbers and is useful in architecture files. It can also check the state (passed in) of some variables in the placer and router for breakpoint evaluation.

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
+#include "breakpoint.h"
+#include "draw_global.h"
+
+#include <iostream>
+
+//global variables
+int tempCount = 0;
+current_information current_info_b;
```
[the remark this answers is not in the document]
MahshadJ: the issue with defining it in the breakpoint class is that it needs to be accessed in the expr_eval class as well which would make things more complicated as expr_eval is included in the breakpoint class.

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -47,15 +53,24 @@ static bool is_char_number(const char ch);
...
+static bool is_variable(std::string var);
+
 // returns the length of any identifier (e.g. name, function) starting at the beginning of str
 static int identifier_length(const char* str);
 
 /* increments str_ind until it reaches specified char is formula. returns true if character was found, false otherwise */
 static bool goto_next_char(int* str_ind, const string& pw_formula, char ch);
 
+bool same_string(std::string str1, std::string str2);
+bool additional_assignemnt_op(int arg1, int arg2);
```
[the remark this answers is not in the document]
MahshadJ: Resolved

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -47,15 +53,24 @@ static bool is_char_number(const char ch);
 // returns true if ch is an operator
```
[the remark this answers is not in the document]
MahshadJ: resolved

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -47,15 +53,24 @@ static bool is_char_number(const char ch);
...
 static bool is_operator(const char ch);
 
-// returns true if the specified name is an known function operator
+// returns true if the specified name is a known function operator
 static bool is_function(std::string name);
 
+// returns true if the specified name is a known compound operator
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/breakpoint.h
MahshadJ: resolved

vpr/src/draw/draw.cpp
MahshadJ: the images are icons of the letters. So it's literally an image of the letter m. It's just something I used to make the information pop up look nicer and less wordy. Should I still change the names?

vpr/src/draw/draw.cpp
MahshadJ: resolved

vpr/src/place/place.cpp
MahshadJ: resolved

MahshadJ: the conversion is needed because block_num is of type ClusterBlockId and that type isn't recognized in expr_eval so we need a vector of ints. Added this explanation in the comments. So resolved.

vpr/src/route/route_timing.cpp
MahshadJ: resolved

vpr/src/route/route_timing.cpp
MahshadJ: resolved

vpr/src/route/route_timing.cpp
MahshadJ: resolved

vpr/src/draw/breakpoint.cpp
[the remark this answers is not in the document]
vaughnbetz: I think you're saying that you're trying to avoid a circular dependency between headers (breakpoint.h #includes vtr_eval.h and vtr_eval.h #includes breakpoint.h)?
Some general ideas for avoiding that are here: https://stackoverflow.com/questions/4816698/avoiding-circular-dependencies-of-header-files
Probably the most general one is to factor the header files into multiple smaller ones, with some just giving limited interfaces (used by more distant code) and some giving more detailed interfaces (used by other code, perhaps more tightly joined).

vpr/src/draw/draw.cpp
vaughnbetz: OK, in that case the names are OK.  Should still comment what they are (these are just letters).  
Could keep the current names, or change them to letter_m.png and such.

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -312,6 +343,22 @@ static void get_formula_object(const char* ch, int& ichar, const t_formula_data&
         ichar--;
         fobj->type = E_FML_NUMBER;
         fobj->data.num = vtr::atoi(ss.str().c_str());
+    } else if (is_compound_op(ch) != 0) {
+        fobj->type = E_FML_OPERATOR;
+        int comp_op_code = is_compound_op(ch);
+        if (comp_op_code == 1)
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw.cpp
MahshadJ: added the comment and resolved

vpr/src/draw/draw.cpp
MahshadJ: I had assigned 0 to comparing operators and 1 to boolean operators and in a valid expression, they appear in a certain manner. So first I would go through the expression and store the operator codes in a vector, then go through the vector and check for the correct pattern. I changed 0 and 1 to enum types of COMP_OP and BOOL_OP and added more comments. So resolved.

vpr/src/draw/draw.cpp
MahshadJ: the function, however, is called in draw.cpp when the user's input is retrieved and the error window pops up if the expression is invalid. The error window function is also in draw.cpp so I believe draw is still the best place for it. I will move all these functions in a new draw_breakpoint.cpp file to further clarify.

vpr/src/draw/draw.cpp
MahshadJ: resolved

vpr/src/draw/draw.cpp
MahshadJ: resolved

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -693,6 +769,32 @@ static bool is_function(std::string name) {
     return false;
 }
 
+int is_compound_op(const char* ch) {
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw.cpp
MahshadJ: resolved

vpr/src/draw/draw.cpp
MahshadJ: resolved

vpr/src/draw/breakpoint.h
```diff
@@ -0,0 +1,89 @@
...
+                return true;
+            else if (bp.type.compare("bt_expression") == 0 && bp.bt_expression.compare(this->bt_expression) == 0)
+                return true;
+            else
+                return false;
+        }
+    }
+};
+
+//functions
```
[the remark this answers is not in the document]
MahshadJ: They aren't part of the breakpoint class but they are all breakpoint related functions that get called in place, draw, and route_timing.cpp. Added comments for all of them. So resolved

vpr/src/draw/breakpoint.h
```diff
@@ -0,0 +1,89 @@
...
+    breakpoint() {
+        bt_moves = 0;
+        bt_temps = 0;
+        bt_from_block = -1;
+        bt_router_iter = 0;
+        bt_route_net_id = -1;
+        active = true;
+    }
+
+    breakpoint(std::string ty, int mtb) {
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+//deletes a breakpoint from the breakpoints vector given its index
+void delete_breakpoint_by_index(int index) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    draw_state->list_of_breakpoints.erase(draw_state->list_of_breakpoints.begin() + index);
+}
+
+//prints the current placer information to the terminal
+void print_current_info() {
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
...
+        else if (draw_state->list_of_breakpoints[i].type.compare("bt_expression") == 0 && draw_state->list_of_breakpoints[i].active)
+            return check_for_expression_breakpoints(draw_state->list_of_breakpoints[i].bt_expression);
+    }
+    return false;
+}
+
+//activates or deactivates a breakpoint given its index
+void activate_breakpoint_by_index(int index, bool enabled) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    if (enabled)
```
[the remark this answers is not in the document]
MahshadJ: resolved

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -728,4 +830,69 @@ bool FormulaParser::is_piecewise_formula(const char* formula) {
...
+        before_addition = arg1;
+    result = (arg1 == (before_addition + arg2));
+    if (result)
+        before_addition = 0;
+    return result;
+}
 } //namespace vtr
+
+//gets current information from breakpoint.cpp
+void get_current_info_e(current_information ci) {
```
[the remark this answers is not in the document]
MahshadJ: resolved

libs/libvtrutil/src/vtr_expr_eval.h
```diff
@@ -153,6 +174,23 @@ class FormulaParser {
     std::vector<Formula_Object> rpn_output_;
     std::stack<Formula_Object> op_stack_; /* stack for handling operators and brackets in formula */
 };
+} // namespace vtr
+//struct that holds all necessary current information about the placer
+struct current_information {
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
+#include "breakpoint.h"
+#include "draw_global.h"
+
+#include <iostream>
+
+//global variables
+int tempCount = 0;
```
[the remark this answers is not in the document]
MahshadJ: I had already eliminated that variable but forgot to take out the declaration :) Resolved

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
...
+    get_current_info_e(ci);
+    current_info_b = ci;
+}
+
+//for accessing get_current_info_b. specifically in place.cpp
+current_information get_current_info_b() {
+    return current_info_b;
+}
+
+//checks if there are any move breakpoints
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
...
+void send_current_info_b(current_information ci) {
+    get_current_info_e(ci);
+    current_info_b = ci;
+}
+
+//for accessing get_current_info_b. specifically in place.cpp
+current_information get_current_info_b() {
+    return current_info_b;
+}
+
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
...
+        if (draw_state->list_of_breakpoints[i].type.compare("bt_moves") == 0 && draw_state->list_of_breakpoints[i].active)
+            return check_for_moves_breakpoints(draw_state->list_of_breakpoints[i].bt_moves);
+        else if (draw_state->list_of_breakpoints[i].type.compare("bt_from_block") == 0 && draw_state->list_of_breakpoints[i].active)
+            return check_for_block_breakpoints(draw_state->list_of_breakpoints[i].bt_from_block);
+        else if (draw_state->list_of_breakpoints[i].type.compare("bt_temps") == 0 && draw_state->list_of_breakpoints[i].active)
+            return check_for_temperature_breakpoints(draw_state->list_of_breakpoints[i].bt_temps);
+        else if (draw_state->list_of_breakpoints[i].type.compare("bt_router_iter") == 0 && draw_state->list_of_breakpoints[i].active)
+            return check_for_router_iter_breakpoints(draw_state->list_of_breakpoints[i].bt_router_iter);
+        else if (draw_state->list_of_breakpoints[i].type.compare("bt_route_net_id") == 0 && draw_state->list_of_breakpoints[i].active)
+            return check_for_route_net_id_iter_breakpoints(draw_state->list_of_breakpoints[i].bt_route_net_id);
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/breakpoint.h
```diff
@@ -0,0 +1,89 @@
...
+            bt_temps = mtb;
+        else if (ty.compare("bt_from_block") == 0)
+            bt_from_block = mtb;
+        else if (ty.compare("bt_router_iter") == 0)
+            bt_router_iter = mtb;
+        else if (ty.compare("bt_route_net_id") == 0)
+            bt_route_net_id = mtb;
+    }
+
+    breakpoint(std::string ty, std::string expr) {
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw.h
MahshadJ: resolved

vpr/src/place/place.cpp
MahshadJ: resolved

vpr/src/draw/breakpoint.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "draw_global.h"
+
+#include <iostream>
+
+//global variables
+int tempCount = 0;
+current_information current_info_b;
+
+//gets current information from place.cpp
+void send_current_info_b(current_information ci) {
```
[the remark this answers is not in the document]
MahshadJ: resolved

libs/libvtrutil/src/vtr_expr_eval.cpp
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw.h
```diff
@@ -83,6 +84,19 @@ float get_net_alpha();
...
 ezgl::color get_block_type_color(t_physical_tile_type_ptr type);
 
+//void breakpoint_info_window(std::string bpDescription, current_information current_info_d);
+
+/*struct open_windows {
+ * bool debug_window = false;
+ * bool advanced_window = false;
+ * };*/
+
+typedef enum operator_type_in_expression {
```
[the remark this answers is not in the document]
vaughnbetz: I think this is enum op_type_in_expr is a duplicate of draw_debug.h and should be deleted?

vpr/src/draw/draw_debug.h
```diff
@@ -0,0 +1,35 @@
+/*This file contains all functions reagrding the graphics of setting and handling breakpoints*/
```
vaughnbetz, line 1: typo (regarding)
Use Doxygen /** so we can build doxygen in future.
I'd say "related to the setting of place and route breakpoints"

vpr/src/place/move_utils.cpp
```diff
@@ -5,6 +5,11 @@
 
 #include "vtr_random.h"
 
+//Run-time flag to control when placer debug information is printed
+//Note : only enables debug output if compiled with VTR_ENABLE_DEBUG_LOGGING is defined
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+bool f_placer_debug = false;
```
vaughnbetz, line 11: Should measure f_placer_debug performance impact.  Does it control more than manual moves?  If so, probably separating manual moves from it would be a good idea, and if manual moves has no performance impact (when off) then I'd turn it on by default.

vpr/src/place/place.cpp
```diff
@@ -1518,6 +1535,45 @@ static e_move_result try_swap(float t,
...
+
+    t_draw_state* draw_state = get_draw_state_vars();
+    if (draw_state->list_of_breakpoints.size() != 0) {
+        //update current information
+        transform_blocks_affected(blocks_affected);
+        get_bp_state_globals()->get_glob_breakpoint_state()->move_num++;
+        get_bp_state_globals()->get_glob_breakpoint_state()->from_block = size_t(blocks_affected.moved_blocks[0].block_num);
+
+        //check for breakpoints
+        f_placer_debug = check_for_breakpoints(true);
```
vaughnbetz, line 1548: I don't think I understand how f_placer_debug is being used.  Is it something that a person needs to edit in the source code to turn on debugging, or a variable that's being used for communication inside the program?

vpr/src/place/move_utils.h
```diff
@@ -4,6 +4,12 @@
 #include "move_transactions.h"
 #include "compressed_grid.h"
 
+//Run-time flag to control when placer debug information is printed
+//Note : only enables debug output if compiled with VTR_ENABLE_DEBUG_LOGGING is defined
+#ifdef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 9: This f_placer_debug variable seems to be used as some kind of read/write control variable later on.  Needs a better comment as I'm not sure what it's really being used for.  If you don't need a global variable, better to get rid of it (not clear from its use that it actually needs to be global).

vpr/src/route/route_timing.cpp
```diff
@@ -1982,3 +1998,20 @@ static void init_net_delay_from_lookahead(const RouterLookahead& router_lookahea
...
+//updates router iteration information and checks for router iteration and net id breakpoints
+//stops after the specified router iteration or net id is encountered
+void update_router_info_and_check_bp(bp_router_type type, int net_id) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    if (draw_state->list_of_breakpoints.size() != 0) {
+        if (type == BP_ROUTE_ITER)
+            get_bp_state_globals()->get_glob_breakpoint_state()->router_iter++;
+        else if (type == BP_NET_ID)
+            get_bp_state_globals()->get_glob_breakpoint_state()->net_id = net_id;
+        f_router_debug = check_for_breakpoints(false);
```
vaughnbetz, line 2011: same comment as f_placer_debug:  is f_router_debug global and does it need to be?

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -2,8 +2,18 @@
...
 #include "vtr_util.h"
 #include "vtr_math.h"
+
 #include <string>
 #include <sstream>
+#include <iostream>
+
+//global variables
+
+//bp_state_globals is a variable of type BreakpointStateGlobals that holds a member of type BreakpointState. This member is altered by the breakpoint class, the placer, and router and holds the most updated values for variables that can trigger breakpoints (e.g move_num, temp_num etc.)
```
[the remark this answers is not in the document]
vaughnbetz: Don't need to say "is of type BreakpointStateGlobals" as the line below makes that clear.  The rest of the comment is good.

libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -2,8 +2,18 @@
...
+BreakpointStateGlobals bp_state_globals;
+
+//this variables is used for the += operator and holds the initial value of the variable that is to be added to. after every addition, the related function compares with initial value to ensure correct incrementation
+int before_addition = 0;
```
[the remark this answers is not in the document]
vaughnbetz: If this variable is not used outside this file, putting the keyword static in front:
static int before_addition = 0;
localizes its definition to this file, which is better than making it global.

vaughnbetz:
Hi Mahshad,
Thanks for all the updates; it looks much improved.  I have some more comments and questions though.  If you can address these quickly and/or explain why there shouldn't be a change that would be good so we can get this merged.
Vaughn

doc/src/vpr/debug_aids.rst
```diff
@@ -11,6 +11,9 @@ To access detailed echo files from VPR’s operation, use the command-line optio
 After parsing the netlist and architecture files, VPR dumps out an image of its internal data structures into echo files (typically ending in ``.echo``).
 These files can be examined to be sure that VPR is parsing the input files as you expect.
 
+To display and debug the placement move generator, enable the f_placer_flag in move_utils.cpp file. The new location of the moving block for each proposed move will be highlighted with GREEN and the old location will be highlighted with GOLD. The fanin and fanout blocks will also be highlighted. The move type, move outcome and delta cost will be printed in the status bar.
```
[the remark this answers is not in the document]
MahshadJ: f_placer_debug is a runtime flag that will be enabled if the compile-time flag VTR_ENABLE_DEBUG_LOGGING is enabled. The compile-time flag is off by default. Also, how can I measure its impact on compile time?

libs/libvtrutil/src/vtr_expr_eval.cpp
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw.h
```diff
@@ -83,6 +84,19 @@ float get_net_alpha();
 
 ezgl::color get_block_type_color(t_physical_tile_type_ptr type);
 
+//void breakpoint_info_window(std::string bpDescription, current_information current_info_d);
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw_debug.h
MahshadJ: resolved

libs/libvtrutil/src/vtr_expr_eval.cpp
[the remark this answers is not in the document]
MahshadJ: moved inside vtr namespace and made static. Resolved

vpr/src/draw/draw.h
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/place/move_utils.cpp
MahshadJ: I think by manual move you mean breakpoints. If so, yes f_placer_debug is only used to stop the program whenever a breakpoint is reached. But it shouldn't be true by default because setting it true would stop the program after each perturbation.

vpr/src/place/place.cpp
MahshadJ: Basically, after every placement perturbation, the program checks if f_placer_debug is true, and if true, it stops the program until proceed is pressed again. Breakpoints stop the program by setting f_placer_debug to true.

vpr/src/route/route_timing.cpp
MahshadJ: f_router_debug is used for more than just breakpoints and was there before I started editing the code.

vpr/src/draw/draw.cpp
MahshadJ: I'm not sure what you mean by moving it in a subroutine. VTR_ENABLE_DEBUG_LOGGING is a compile-time flag that is enabled allows debugging

vpr/src/route/route_timing.cpp
vaughnbetz: OK, I think I understand it better now.  It needs a comment as it's an important variable (anything file scope or global should be important and commented).  You should comment that if this variable is true, the placer will stop and let the user examine the state of the placement in the GUI.

doc/src/vpr/debug_aids.rst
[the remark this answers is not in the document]
vaughnbetz: OK, if you already measured the compile time impact of breakpoints with this #define enabled and it was negligible then it's OK.  Given your description above I assume checking the runtime has already been done (when you checked runtime with and without your breakpoint code).  If that's correct, then this is checked already.

vpr/src/place/move_utils.cpp
vaughnbetz: OK, just comment the variable use then.

README.md
```diff
@@ -84,7 +84,7 @@ Developers checkout a local copy of the code at the start of development, then d
 When a developer has a tested, working change to put back into the trunk, he/she performs a `git push` operation.
 Unstable code should remain in the developer's local copy.
 
-We do automated testing of the trunk using BuildBot to verify functionality and Quality of Results (QoR).
+We do automated testing of the trunk using BuildBot to verify functionality and Quality of Results (QoR). The QoR tests indicate insignificant change in vtr_flow_elapsed_time.
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/route/route_timing.cpp
MahshadJ: resolved.

vpr/src/place/place.cpp
vaughnbetz: Mahshad I think you asked me what I meant by this.  I mean that the code guarded by DEBUG_LOGGING is large enough that it should be separated into a different subroutine.

doc/src/vpr/debug_aids.rst
vaughnbetz, line 14: Now that I understand f_placer_debug, I think this documentation is not correct/appropriate.  

"You can visualize and control the placement move generator whenever the placement engine is paused in the UI. 
Run with graphics enabled and set a breakpoint to stop placement at an appropriate point and then ... (now put in the info on move colors etc.)"

(Or else f_place_debug does something other than control breakpoints.  But if f_place_debug is being set automatically by placement breakpoints it doesn't make sense to tell users to go edit a file and set it to true.).

vaughnbetz:
Mahshad, were Kevin's comments on draw.cpp (moving code outside the VTR_ENABLE_DEBUG_LOGGING, some refactoring) ever addressed?  If not, I think we could land this PR, but right afterward it would be good to try those refactorings and moving the VTR_ENABLE_DEBUG_LOGGING guard.  Compile time guards make it harder to keep code working since most people won't be compiling that code, so if they don't save runtime for the average user or turn off excessive output etc. it's better to not have them.

doc/src/vpr/debug_aids.rst
MahshadJ: resolved

vaughnbetz:
Mahshad, is this ready to go at this point?  I can't rebase and merge, but could do a merge commit (harder to track the commit history, but I would like to get this in at this point and I'm not sure what it would take to manually rebase).

If this is ready to go I'll merge it.  Then Bill can apply his changes on top of it, and make some fine-tuning like moving more of the breakpoint logic out of place.cpp (since he moved file scope variables out, he can now split code out of place.cpp more easily).

vaughnbetz:
Mahshad confirmed by email this is ready to go.  Merging (rebase has conflicts, so doing a merge commit which isn't quite as nice).

PR #1293, opened 2020-05-04
Odin: use json for regression expected results and add expected warnings

jeanlego:
#### Description
This PR switches from csv to json to allow more complex content to be stored.  

#### Motivation and Context
It was neccessary since multiline values in csv would not work properly, json will allow us to parse expected errors and support finer grain regression test and prevent changes that would break the flow.

Ran throught the whole odin suite

#### Types of changes

#### Checklist:


jeanlego:
@djns99 as discussed this should solve the problem

PR #1295, opened 2020-05-06
vpr: Convert 2D nested vectors storing net-pin data to a (sparse) flat matrix

kmurray:
vtr::FlatSparseMatrix takes into account that some rows (nets) are
smaller than others and lays them out in adjacent memory. This should
reduce memory fragmentation and improve locality.


vpr/src/route/route_budgets.cpp
```diff
@@ -66,11 +66,11 @@ void route_budgets::free_budgets() {
     /*Free associated budget memory if set
      * if not set, only free the chunk memory that wasn't used*/
     if (set) {
-        free_net_delay(delay_min_budget, &min_budget_delay_ch);
-        free_net_delay(delay_max_budget, &max_budget_delay_ch);
-        free_net_delay(delay_target, &target_budget_delay_ch);
-        free_net_delay(delay_lower_bound, &lower_bound_delay_ch);
-        free_net_delay(delay_upper_bound, &upper_bound_delay_ch);
+        delay_min_budget.clear();
```
litghost, line 69: clear does not free the memory, only sets the size to 0.  Is this the intention?.  shrink_to_fit increases the chance that the memory is free'd, and the capacity is set to 0.

libs/libvtrutil/src/vtr_ragged_matrix.h
```diff
@@ -0,0 +1,188 @@
+#ifndef VTR_RAGGED_MATRIX_H
```
litghost, line 1: This needs unit tests in libs/libvtrutils/test

libs/libvtrutil/src/vtr_ragged_matrix.h
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+      private:
+        size_t irow_;
+        Callback& callback_;
+    };
+
+  private:
+    std::vector<T> data_;
+    std::vector<int> first_elem_;
```
litghost, line 183: Consider make a method to sanity check that first_elem_ is sorted?

vpr/src/route/route_budgets.cpp
kmurray: Good point, I've added vtr::release_memory(), which will force container memory to be released.

libs/libvtrutil/src/vtr_ragged_matrix.h
kmurray: Done

libs/libvtrutil/src/vtr_ragged_matrix.h
kmurray: The structure is enforced by the constructor and there is no method to modify it after construction, so no need to check for sorting.

libs/libvtrutil/src/vtr_ragged_matrix.h
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    friend void swap(FlatRaggedMatrix<T, Index0, Index1>& lhs, FlatRaggedMatrix<T, Index0, Index1>& rhs) {
+        lhs.swap(rhs);
+    }
+
+  public: //Types
+    //Proxy class used to represent a 'row' in the matrix
+    template<typename U>
+    class ProxyRow {
```
litghost, line 142: Can you just use array_view here, rather than defining a new thing?

kmurray: Yep, makes sense. I've updated it.

PR #1296, opened 2020-05-06
Add support for incremental STA

kmurray:
#### Motivation and Context

Traditionally VPR has performed (full) timing updates sporadically (e.g. once per placement temperature, or once per routing iteration) since they are relatively expensive (linear in the size of the timing graph).

Since VPR mostly operates by incrementally modifying the circuit it would be very helpful to use more up-to-date timing information to ensure the optimizers have a more accurate view of the current circuit timing.

#### Description

This PR adds support to VPR for incremental timing updates based on the new incremental STA support which has been added to [Tatum](https://github.com/kmurray/tatum/pull/5).

By recording what edges in the timing graph have had their delays changed, it is possible for the STA engine to incrementally update timing information, usually in much less than linear time. While support for this is now included in Tatum, it needs additional support to work effectively in VPR.

In particular this PR adds:
 * Support for invalidating edges who's delay has changed to allow the STA engine to incrementally re-calculate timing information
 * Re-works VPR's slack, crticality and timing costs to update incrementally based on the nodes/pins who's timing information has changed (as reported by the STA engine). This is actually a significant amount of work, but also means the various algorithms are more efficient even when no incremental STA is used.

At the moment, incremental STA is not enabled by default (would need to set `--timing_update_type incremental`) and the traditional full STA updates are used (although the incremental placer cost updates are always enabled). Currently only the placer has been updated to support incremental timing updates, but it should be straightforward to extend this to the router as well.

#### Future work (beyond scope of this PR)
 * Extend incremental timing updates to router
 * Improve placer algorithm to exploit less stale timing information
 * Parallelize incremental analyzer in Tatum

Tested on vtr_reg_basic/vtr_reg_strong and a subset of nightly (vtr_req_qor_chain, titan_other, titan_quick_qor).

#### Types of changes

#### Checklist:


kmurray:
### Outstanding work before merging

 * Track down inconsistencies between full and incremental STA
   * Most Titan circuits produce identical results, but 7 do not (LU_Network, stap_qrd, gsm_switch, neuron, sparcT1_core, openCV, des90, dart)
 * Detailed performance analysis and comparison

kmurray:
Preliminary Experimental Results:
image

When run with frequent STA during quench (~62,000x more STA), incremental TA is only 8x slower. Reducing overall placement time by ~77x. However the standard placer (one STA per temperature) is a little bit slower.

kmurray:
More detailed experimental results: [main.pdf](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4610390/main.pdf)

* When doing frequent STA during quench, incremental STA is >100x faster than non-incremental
* The overhead of the additional logic to support incremental STA is currently ~6% placement run-time when doing full timing updates

kmurray:
I've worked to clean-up the remaining issues with this PR:

* Different results in full vs incremental modes: 
  * Fixed. Full and incremental modes now always produce identical results
* Run-time overhead (in combination with #1295) compared to baseline with no incremental support (6c14352f2ad3ae):
  * Full (i.e. incremental placer cost updates, but _non-incremental_ STA): ~1% slower
  * Incremental (i.e incremental placer cost updates, and _incremental_ STA): ~2% slower
  * Incremental + Frequent Timing Analysis (during quench): > 100x faster

Experimental Results: [titan_results.pdf](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4629209/incr_sta_titan_incratomcritfix.pdf)

kmurray:
@litghost @HackerFoo @vaughnbetz This should now be ready for review

@Bill-hbrhbr please also feel free to review and give feedback.

litghost:
I believe I understand the motivation behind this change, but shouldn't we have results showing that the effort lead to better placement results?  This appears to add around ~3k lines of code, but does having the more accurate placer timing data actually improve the output?  If anything, these changes appear to make the WL and CPD worse?

kmurray:
> shouldn't we have results showing that the effort lead to better placement results?
> does having the more accurate placer timing data actually improve the output?

This PR is a fundamental infrastructure enhancement that unlocks potential new optimization approaches which were not previously tenable due to the run-time cost of frequent STA. 

It is future work to improve the placer to exploit this capability. A variety of previous research work has shown that increasing STA frequency is helpful (see for instance: [Section IX of the Tatum paper](http://www.eecg.utoronto.ca/~kmurray/tatum/fpt2018_tatum.pdf), [Chapter 5 of Ken Eguro's thesis](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/eguro-dissertation.pdf])).

>  If anything, these changes appear to make the WL and CPD worse?

I do not believe so. As shown in the [titan_results.pdf](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4629209/incr_sta_titan_incratomcritfix.pdf) (e.g. Table 2), this PR produces equivalent wirelength (1.00x) and CPD (1.00x) as master.

vaughnbetz:
It should also allow an improvement in routing time, as we can employ incremental STA there with no algorithmic changes. But that will take some extra code for the router to mark the appropriate edges (re-routed) as dirty so they get re-timing analyzed.  I think it's best to land this capability without waiting for full exploitation by the router or placer, as it's already a big PR.  But I am confident we'll be able to exploit it for a near-term router runtime reduction, and it will allow new placer algorithms for fine-tuning that should yield quality gains in the medium term.

vpr/src/place/timing_place.cpp
```diff
@@ -14,86 +14,93 @@
...
-            float clb_pin_crit = calculate_clb_net_pin_criticality(timing_info, pin_lookup, clb_pin);
-
-            /* The placer likes a great deal of contrast between criticalities.
-             * Since path criticality varies much more than timing, we "sharpen" timing
-             * criticality by taking it to some power, crit_exponent (between 1 and 8 by default). */
-            f_timing_place_crit[net_id][ipin] = pow(clb_pin_crit, crit_exponent);
+     * in that pin), timing_place_crit_ = criticality^(criticality exponent) */
+
+    //Determine what pins need updating
+#ifdef INCR_UPDATE_CRITICALITIES
```
litghost, line 37: Does this really need to be a #ifdef?  I'd be suprised if this being a runtime option is that expensive?

vpr/src/timing/slack_evaluation.cpp
```diff
@@ -19,7 +31,24 @@ SetupSlackCrit::SetupSlackCrit(const AtomNetlist& netlist, const AtomLookup& net
     , netlist_lookup_(netlist_lookup)
     , pin_slacks_(netlist_.pins().size(), NAN)
     , pin_criticalities_(netlist_.pins().size(), NAN) {
-    //pass
+#if !defined(INCR_SLACK_UPDATE) || !defined(INCR_UPDATE_CRIT)
```
litghost, line 34: Again, but shouldn't these be runtime modifications, not compile time?  How often is this constructor used?

vpr/src/place/timing_place.cpp
kmurray: I agree making this a run-time option is unlikely to be a performance issue. 

However these are fairly low-level configuration options which I do not think end-users should be concerned about/controlling. As a result, I don't see much point in exposing them at the command-line.

litghost: I'm actually more objecting to the use of #ifdef's.  Consider that you have two ifdef's that control the same behavior in timing_place and slack_evaluation, I think the use of #ifdef's in this PR seems confusing and unclear to it's goal.

kmurray: The main motivations for having these options is to:

- Allow measurement of the performance characteristics of full vs incremental updates
- Support debugging by allowing various aspects of full vs incremental updates to be enabled/disabled

>  Consider that you have two ifdef's that control the same behavior in timing_place and slack_evaluation

To clarify, these defines actually all control different parts of the incremental update. There are multiple data structures/costs which are incrementally updated. Each has it's own define controlling whether it is fully or incrementally updated.

> A runtime bool, even if not exposed as a command line option, seems to make more sense to me.

Yes, this is reasonable and would avoid some potential code rot issues since all paths would always  be compiled.

vpr/src/timing/timing_util.h
```diff
@@ -77,6 +81,108 @@ tatum::NodeId find_origin_node_for_hold_slack(const tatum::TimingTags::tag_range
 //Returns the a map of domain's and their clock fanout (i.e. logical outputs at which the clock captures)
 std::map<tatum::DomainId, size_t> count_clock_fanouts(const tatum::TimingGraph& timing_graph, const tatum::SetupTimingAnalyzer& setup_analyzer);
 
+//Helper class for iterating through the timing edges associated with a particular
+//clustered netlist pin, and invalidating them.
+//
+//For efficiency, it stores pre-calculates and stores the mapping from ClusterPinId -> tatum::EdgeIds,
```
litghost, line 87: > it stores pre-calculates and stores the mapping from ClusterPinId -> tatum::EdgeIds,

Wording?

vpr/src/timing/timing_util.h
```diff
@@ -77,6 +81,108 @@ tatum::NodeId find_origin_node_for_hold_slack(const tatum::TimingTags::tag_range
...
+
+        for (ClusterPinId clb_pin : clb_nlist.pins()) {
+            pin_first_edge_.push_back(timing_edges_.size());
+
+            for (const AtomPinId atom_pin : clb_atom_pin_lookup.connected_atom_pins(clb_pin)) {
+                tatum::EdgeId tedge = atom_pin_to_timing_edge(timing_graph, atom_nlist, atom_lookup, atom_pin);
+
+                if (!tedge) {
+                    /*
+                     *VTR_LOG_WARN("Found no timing edge corresponding to atom pin '%s' (%zu), connected to cluster pin '%s' (%zu)\n",
```
litghost, line 111: Why have this as a comment?

libs/EXTERNAL/libtatum/tatum_test/main.cpp
```diff
@@ -94,6 +104,34 @@ Args parse_args(int argc, char** argv);
 double median(std::vector<double> values);
 double arithmean(std::vector<double> values);
 
+template<class I>
+double median(I begin, I end) {
+    std::sort(begin, end);
```
HackerFoo, line 109: There are faster (linear time) [algorithms](https://en.wikipedia.org/wiki/Selection_algorithm) for this.

kmurray: Agreed, but as this is non-critical test verification code not used by VTR I don't think its a major issue.

libs/EXTERNAL/libtatum/tatum_test/main.cpp
```diff
@@ -94,6 +104,34 @@ Args parse_args(int argc, char** argv);
...
+
+    size_t size = std::distance(begin, end);
```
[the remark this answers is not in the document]
kmurray: This is test code in tatum and so is not critical.

vpr/src/place/timing_place.cpp
kmurray: I've converted these define to bools.

vpr/src/timing/timing_util.h
kmurray: Removed

vpr/src/timing/timing_util.h
kmurray: Fixed

vpr/src/timing/slack_evaluation.cpp
kmurray: Done

PR #1298, opened 2020-05-11
Fix bug in sub tiles

acomodi:
#### Description

This PR fixes some bugs related to the recent sub tiles addition.

#### Motivation and Context

For SymbiFlow purposes, the sub tiles were tested on regular tiles, such as the BUFG tile which has the same kind of sub tiles, which differ only for the pinlocations.

When it comes to more heterogeneous tiles, such as the CLBM, the sub tiles differ also in the pin definition.

What went undiscovered in the sub tile addition were two things:

- correct pin mappings checks between logical and physical tiles that (fixed with https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/2d5cb9979cbd49eeb2797699927d19b1741a7f77)
- fix initial placement that cannot have incompatible `logical_blocks <-> sub_tiles` locations. Also fixed the check for the sub tile compatibility (fixed with https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/b30e26b6df791cce0770f805ededf67a1920e873).

strong and basic regression tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -352,12 +352,6 @@ static void initial_placement_blocks(std::vector<std::vector<int>> free_location
 
         to = legal_pos[itype][isub_tile][ipos];
 
-        if (to.sub_tile > 10000) {
-            VTR_LOG("OUT OF BOUNDS");
-        }
-
-        VTR_LOG("TYPE: %s, SUBTILE: %d\n", type->name, isub_tile);
-        VTR_LOG("%d, %d, %d\n", to.x, to.y, to.sub_tile);
         // Make sure that the position is EMPTY_BLOCK before placing the block down
         VTR_ASSERT(place_ctx.grid_blocks[to.x][to.y].blocks[to.sub_tile] == EMPTY_BLOCK_ID);
 
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -336,12 +336,15 @@ static void initial_placement_blocks(std::vector<std::vector<int>> free_location
         auto possible_sub_tiles = get_possible_sub_tile_indices(type, logical_block);
 
         int isub_tile = OPEN;
-        while (isub_tile == OPEN) {
-            isub_tile = possible_sub_tiles[vtr::irand(possible_sub_tiles.size() - 1)];
-            if (free_locations[itype][isub_tile] == 0) {
-                std::remove(possible_sub_tiles.begin(), possible_sub_tiles.end(), isub_tile);
-                isub_tile = OPEN;
+        for (int sub_tile : possible_sub_tiles) {
+            if (free_locations[itype][sub_tile] > 0) {
+                isub_tile = sub_tile;
+                break;
             }
+        }
+
+        if (isub_tile == OPEN) {
+            VPR_THROW(VPR_ERROR_PLACE, "No more sub tiles available for initial placement.");
         }
 
         t_pl_loc to;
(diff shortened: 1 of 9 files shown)


vpr/src/place/initial_placement.cpp
```diff
@@ -161,9 +189,9 @@ static int try_place_macro(int itype, int ipos, t_pl_macro pl_macro) {
     return (macro_placed);
 }
 
-static void initial_placement_pl_macros(int macros_max_num_tries, int* free_locations) {
+static void initial_placement_pl_macros(int macros_max_num_tries, std::vector<std::vector<int>> free_locations) {
```
kmurray, line 192: free_locations should be passed by reference to avoid copying (and match previous behaviour).

vpr/src/place/initial_placement.cpp
```diff
@@ -298,11 +331,33 @@ static void initial_placement_blocks(int* free_locations, enum e_pad_loc_type pa
...
+
+        auto possible_sub_tiles = get_possible_sub_tile_indices(type, logical_block);
+
+        int isub_tile = OPEN;
+        for (int sub_tile : possible_sub_tiles) {
+            if (free_locations[itype][sub_tile] > 0) {
+                isub_tile = sub_tile;
+                break;
+            }
+        }
```
kmurray, line 344: Need comments here, or perhaps make into a named sub-routine (e.g. find_empty_subtile())?

vpr/src/place/initial_placement.cpp
```diff
@@ -298,11 +331,33 @@ static void initial_placement_blocks(int* free_locations, enum e_pad_loc_type pa
...
+        int ipos = vtr::irand(free_locations[itype][isub_tile] - 1);
+
+        to = legal_pos[itype][isub_tile][ipos];
+
+        if (to.sub_tile > 10000) {
+            VTR_LOG("OUT OF BOUNDS");
+        }
 
+        VTR_LOG("TYPE: %s, SUBTILE: %d\n", type->name, isub_tile);
+        VTR_LOG("%d, %d, %d\n", to.x, to.y, to.sub_tile);
```
kmurray, line 360: Looks like debug code, remove?

vpr/src/place/initial_placement.cpp
```diff
@@ -320,23 +375,20 @@ static void initial_placement_blocks(int* free_locations, enum e_pad_loc_type pa
...
 
-static void initial_placement_location(const int* free_locations, int& ipos, int itype, t_pl_loc& to) {
-    ipos = vtr::irand(free_locations[itype] - 1);
-    to = legal_pos[itype][ipos];
-}
-
 static t_physical_tile_type_ptr pick_placement_type(t_logical_block_type_ptr logical_block,
                                                     int num_needed_types,
-                                                    int* free_locations) {
+                                                    std::vector<std::vector<int>> free_locations) {
```
kmurray, line 385: Pass by reference

vpr/src/place/initial_placement.cpp
acomodi: Sure, removed

acomodi:
> We should make sure we have regression tests which cover the cases which this PR fixes

I have enhanced the sub tiles architecture for the sub tiles reg test actually. Would it be better to have a separate test instead?

> would be good to re-evaluate the impact of this on larger benchmarks, particularly since there are changes to the valid placement locations data structure used during initial placement.

Once all the issues are solved I'll run a vtr_quick_qor test. 
Currently I have spotted yet another issue in the clb directs.

kmurray:
> I have enhanced the sub tiles architecture for the sub tiles reg test actually. Would it be better to have a separate test instead?

That should be fine, so long as we have some test which covers these issues that should be sufficient.

vpr/src/place/initial_placement.cpp
acomodi: Done

vpr/src/place/initial_placement.cpp
acomodi: Done (called the sub routing get_free_subtile)

vpr/src/place/initial_placement.cpp
acomodi: Done

acomodi:
@kmurray I am currently running the titan_quick_qor tests as the outstanding implementation issues have been solved.

Strong tests should now turn green.

Once I'll have the QoR results, I will paste them here.

acomodi:
@kmurray I have the results from the `titan_quick_qor` tests back. Seems that there is no regression with baseline.

[qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4623599/qor_comparison.xlsx)

kmurray:
Looks great! Thanks @acomodi.

PR #1301, opened 2020-05-15
Updated error message reporting format

djns99:
Improves the error reporting syntax

#### Description
Improves the error reporting to print the failure location using _file:line_ format so IDEs can recognise the error location and provide a quick jump.
It also tweaks some of the other formatting to be closer to the normal C _assert()_ function output

#### Types of changes

#### Checklist:


ODIN_II/SRC/odin_error.cpp
```diff
@@ -107,8 +106,29 @@ void _log_message(odin_error error_type, long column, long line_number, long fil
...
+    if (!condition) {
+        fprintf(stderr, "%s:%ld: %s: ", odin_file_name, odin_line_number, odin_function_name);
+        if (condition_str) {
+            // We are an assertion, print the condition that failed and which line it occurred on
+            fprintf(stderr, "Assertion %s failed\n", condition_str);
+            // -1 since __LINE__ starts from 1
+            print_culprit_line(-1, odin_line_number - 1, odin_file_name);
+        } else {
+            // We are a parsing error, dont print the culprit line
+            fprintf(stderr, "Fatal error\n");
```
djns99, line 129: Is there any value to printing the line in odin for this case? Seems like the user shouldn't care and devs should be able to just look at the code.

ODIN_II/SRC/odin_error.cpp
```diff
@@ -107,8 +106,29 @@ void _log_message(odin_error error_type, long column, long line_number, long fil
...
+
+void _verbose_assert(bool condition, const char* condition_str, const char* odin_file_name, long odin_line_number, const char* odin_function_name) {
+    fflush(stdout);
+    if (!condition) {
+        fprintf(stderr, "%s:%ld: %s: ", odin_file_name, odin_line_number, odin_function_name);
+        if (condition_str) {
+            // We are an assertion, print the condition that failed and which line it occurred on
+            fprintf(stderr, "Assertion %s failed\n", condition_str);
+            // -1 since __LINE__ starts from 1
+            print_culprit_line(-1, odin_line_number - 1, odin_file_name);
```
djns99, line 126: Would this be more valuable if it printed a couple of lines of context as well?

ODIN_II/SRC/include/odin_error.h
```diff
@@ -28,32 +28,25 @@ extern const char* odin_error_STR[];
...
-}
+void _verbose_assert(bool condition, const char* condition_str, const char* odin_file_name, long odin_line_number, const char* odin_function_name);
 
 #define oassert(condition) _verbose_assert(condition, #condition, __FILE__, __LINE__, __func__)
 
 void _log_message(odin_error error_type, long column, long line_number, long file, bool soft_error, const char* function_file_name, long function_line, const char* function_name, const char* message, ...);
 
 #define error_message(error_type, line_number, file, message, ...) \
-    _log_message(error_type, -1, line_number, file, true, __FILE__, __LINE__, __func__, message, __VA_ARGS__)
+    _log_message(error_type, -1, line_number, file, true, __FILE__, __LINE__, __PRETTY_FUNCTION__, message, __VA_ARGS__)
```
jeanlego, line 38: is that supported by clang and gcc down to 5? `__PRETTY_FUNCTION__`

ODIN_II/SRC/odin_error.cpp
jeanlego: yes it would! would you mind adding that?

ODIN_II/SRC/odin_error.cpp
jeanlego: I don't disagree, the biggest advantage here is when a user report an error, now we can actually see what failed until a better issue is reported. More often than not all that is sent is just the error message, no test file etc... so it's hard to reproduce and this gives us some bread crumbs. While I agree in the long term this should go away, right now I don't know.

ODIN_II/SRC/odin_error.cpp
```diff
@@ -94,7 +93,7 @@ void _log_message(odin_error error_type, long column, long line_number, long fil
             file_name = file_name.substr(slash_location + 1);
```
jeanlego, line 93: I've done filename only here since otherwise I would need a more complex regex in the parse_conf, but if you see some value to it, I'm not against switching the format.

jeanlego:
Since it touches the regression results, I wouldn't mind merging as is since further modification shouldn't affect it. you can continue work on the comment afterwards

ODIN_II/SRC/odin_error.cpp
djns99: My IDE (CLion) worked if you printed the file name by itself or the entire path. I cant speak for other IDEs though. So this was fine

ODIN_II/SRC/odin_error.cpp
djns99: `/home/dast/vtr-verilog-to-routing/ODIN_II/SRC/odin_util.cpp:106: open_file: Assertion false failed
104: FILE* open_file(const char* file_name, const char* open_type) {
105:     FILE* opened_file = fopen(file_name, open_type);
106:     oassert(false);
         ^~~~
107:     if (opened_file == NULL) {
108:         error_message(UTIL, -1, -1, "cannot open file: %s\n", file_name);`

ODIN_II/SRC/odin_error.cpp
djns99: What I meant is should we be printing the culprit line instead of just the line number. I definitely see the value of keeping the ODIN line in for the foreseeable future. But I think printing the actual code in ODIN will be confusing for the user and should not be relevant to them, since (in theory) its a bug in their code not ours.
So for now we will continue to print the ODIN line number and no extra information

ODIN_II/SRC/odin_error.cpp
jeanlego: Looks awesome! Thanks for that!

ODIN_II/SRC/odin_error.cpp
jeanlego: Then I guess it's good as is!

ODIN_II/SRC/odin_error.cpp
jeanlego: My bad. I thought you meant line number :face-palm:.
Then no, line number is all the info we want here. I agree that it would just be confusing.

jeanlego:
Just confirming that's its good to go before merging, did it pass all the odin suite?

PR #1305, opened 2020-05-20
Reserve VPR delayless switch name

tangxifan:
This pull request aims to solve the following problem: 

  - Keep the VPR delayless switch name in a constant string and can be used easily in later functions
  - Avoid users to define a switch in the reserved names, which may crash the software.
 

#### Description

Modify the readarch library to add a constant string for delayless switches.
Check codes are added to error out `read_arch`  when users try to define a switch in the same name as the delayless switch.
VPR setup code has been adapted to use the delayless switch constant.

#### Related Issue

This issue has been discussed during the VTR developer's meeting at FPGA'2020

#### Motivation and Context

Delayless switch is a special switch created by VPR internally.
It is the switch that is used by edges between SOURCE and OPIN, IPIN and SINK, or direct connection between CLBs in VPR's routing resource graphs.
It has a special name and users are not allowed to define it.

Passed the regression tests in CI without QoR changes.
Tested with XML sample, to ensure that read_arch will error out when reserved names are used.

No impact on the rest of VPR codes. 

#### Types of changes

- [ X ] Bug fix (change which fixes an issue)

#### Checklist:

- [X ] All new and existing tests passed

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1390,6 +1390,17 @@ enum class SwitchType {
 };
 constexpr std::array<const char*, size_t(SwitchType::NUM_SWITCH_TYPES)> SWITCH_TYPE_STRINGS = {{"MUX", "TRISTATE", "PASS_GATE", "SHORT", "BUFFER", "INVALID"}};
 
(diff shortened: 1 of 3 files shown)


kmurray:
Thanks @tangxifan this looks like a good improvement!

PR #1306, opened 2020-05-20
Fixed post-synthesis netlist writing

mkurc-ant:
#### Description

VPR crashes when run with `--analysis` and `--gen_post_synthesis_netlist`. It turned out that the crash was caused by unconnected pins not having assigned nets in the `t_pb->pb_route` map.

I fixed that by checking whether a key is present in the map instead of retrieving its value and checking if it corresponds to a valid net. 

#### Related Issue
https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1467

#### Motivation and Context
This is a bugfix that allows to get a post-synthesis netlist from VPR.

Locally by enabling the `--gen_post_synthesis_netlist` and seeing it not crash.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/netlist_writer.cpp
+++ b/vpr/src/base/netlist_writer.cpp
@@ -1603,14 +1603,16 @@ class NetlistWriterVisitor : public NetlistVisitor {
                 const t_port* port = pin->port;
 
                 int cluster_pin_idx = pin->pin_count_in_cluster;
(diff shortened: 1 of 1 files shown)


kmurray:
Thanks @mkurc-ant! The fix looks good to merge once CI passes.

mkurc-ant:
@kmurray Great! thanks!

PR #1309, opened 2020-05-21
Packable mode support 

tangxifan:
This pull request aims to enhance users' experience in VPR packer.
It allows users to specify if a ``<mode>`` of a ``<pb_type>`` in the architecture XML to be visible/invisible to VPR packer.

#### Description

Changes are made to ``libarchfpga`` and VPR packer to support the feature.
- New XML syntax ``packable=<bool>`` has been added to ``libarchfpga`` as a property of XML node ``<mode>``. 
- Data structure ``t_mode`` has a new variable of ``packable`` to store the property.
- VPR pack is modified to flag infeasible packing when 
    - Packer tries to consider pack patterns that are defined in an unpackable ``<mode>``
    - Packer tries to map resource to a ``<pb_type>`` under an unpackable ``<mode>``
    - Logic block router tries to expand edges that belong to an unpackable ``<mode>``
- Documentation has been updated for the new syntax

#### Related Issue

This issues has been presented and discussed during the VTR developers' meeting at FPGA'2020.
Currently, it is hard to debug a ``<complexblock>`` with a deep hierarchy, especially due to multi-mode ``<pb_type>``. 
Packer may drop the ball due to a dangling net/connection or other minor issues in the ``<pb_type>``, which VPR's architecture checker will not flag fatal errors.
When packing fails, it is hard for users to spot which ``<mode>`` or which  ``<pb_type>`` causes the problems.

#### Motivation and Context

Now, users can use XML syntax ``packable`` to specify if a ``<mode>`` of a ``<pb_type>`` in the architecture XML to be visible/invisible to VPR packer.
By default, every ``<mode>`` is packable.

For example, 

    `<mode name="lut6" packable="false">`

When users define an unpackable ``<mode>``, any child ``<pb_type>`` and any mode under the children will be unpackable.

As such, users can turn on/off any ``<mode>`` for packer and quickly spot which ``<mode>`` causes packing to fail.

It passed CI regression tests without QoR changes.
Passed artificial XML with unpackable mode definition.

No impact on other codes except packer.

#### Types of changes

- [ X] New feature (change which adds functionality)

#### Checklist:

- [ X] I have updated the documentation accordingly
- [ X] All new and existing tests passed


mithro:
@tangxifan -- I think this needs a "howto doc" which demonstrates how to debug an issue with the packer and thus explains why this feature exists. It seems like you could just use the verbose packing output?

doc/src/arch/reference.rst
```diff
@@ -1454,7 +1462,7 @@ The following tags are common to all <pb_type> tags:
             </mode>
         </pb_type>
 
-    specifies the ``lut`` pb_type can be used as either a single 6-input LUT, or as two 5-input LUTs (but not both).
+    specifies the ``lut`` pb_type can be used as either a single 6-input LUT, or as two 5-input LUTs (but not both). Mode ``lut5x2`` is set to be unpackable, and therefore will not be considered for mapping resources during VPR packing stage. Mode ``lut6`` is by default set to be packable, where VPR packer will try to map resources.
```
kmurray, line 1465: I'm not sure putting packable=false in the example is a great idea, since it is really a power-user debugging option and hence not typical usage.

I could anticipate a new user thinking: 'Why would you define this lut5x2 mode but make it unpackable?'. Probably better to leave it out here, and make a packer debugging tutorial like @mithro suggested, which can be cross-refed here (e.g. ``.. seealso::`).

doc/src/arch/reference.rst
```diff
@@ -1416,7 +1416,7 @@ The following tags are common to all <pb_type> tags:
     Multple clock ports are described using multiple ``<clock>`` tags.
     *See above descriptions on inputs*
 
-.. arch:tag:: <mode name="string">
+.. arch:tag:: <mode name="string" packable="bool">
```
kmurray, line 1419: Since we expect typical usage to not specify packable, perhaps we should give it a more exceptional name.

Perhaps something like `disable_pack` or `disable_packing`. Not sure

doc/src/arch/reference.rst
tangxifan: Indeed, this is a feature for power users to debug the packer and architecture file.
I think a tutorial for packer debugging where bugs in architecture file can be quickly spotted by using the ``disable_packing`` switch and the verbose output of packer.

doc/src/arch/reference.rst
tangxifan: This is a good point. I will change the name.

tangxifan:
Hi @mithro @kmurray ,
   I have added a short tutorial about how-to-debug with the ``disable_packing`` switch.
   I notice that actually it can be part of how to build a multi-mode logic block, where we show how to use architecture description language to model a multi-mode logic block and then use the ``disable_packing`` switch to validate it correctness.
   Therefore, I added a tutorial to the architecture part.
   Feel free to advise.

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+This tutorial aims to introduce how to build a representative multi-mode logic block by exploiting VPR architecture description language, as well as debugging tips to guarantee each mode of a logic block is functional.
+
+Definition
+----------
+
+Modern FPGA logic blocks are designed to operate in various modes, so as to provide best performance for different applications.
+VPR offers enriched syntax to support highly flexible multi-mode logic block architecture.
+
+:numref:`fig_frac_lut_le` shows the physical implemenation of a Fracturable Logic Element (FLE), which conssists of a fracturable 6-input Look-Up Table (LUT), two Flip-flops (FFs) and routing multiplexers to select between combinational and sequential outputs.
```
kmurray, line 14: sp: conssists -> consists

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+.. code-block:: xml
+
+   <!-- Dual 5-input LUT mode definition begin -->
+   <mode name="n2_lut5">
+     <pb_type name="lut5inter" num_pb="1">
+       <input name="in" num_pins="5"/>
+       <output name="out" num_pins="2"/>
+       <clock name="clk" num_pins="1"/>
+       <pb_type name="ble5" num_pb="2">
```
kmurray, line 79: Should probably have a comment here, calling out `num_pb=2` as why this models two 5-LUT + FF pairs

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+   <!-- 6-LUT mode definition begin -->
+   <mode name="n1_lut6" disable_packing="false">
+     <!-- Define 6-LUT mode -->
+     <pb_type name="ble6" num_pb="1">
+       <input name="in" num_pins="6"/>
+       <output name="out" num_pins="1"/>
+       <clock name="clk" num_pins="1"/>
+       <!-- Define LUT -->
+       <pb_type name="lut6" blif_model=".names" num_pb="1" class="lut">
```
kmurray, line 153: Similarly, it would be helpful to have a comment noting `num_pb="1"` here.

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+Validation in packer
+--------------------
+After finishing the architecture description, the next step is to validate that VPR can map logics to each operating modes.
+Since VPR packer will exhaustively try each operating mode and finally map logics to one of it.
+As long as there is an operating mode that is feasible for mapping, VPR will complete packing without errors.
+However, this may shadow the problems for other operating modes.
+It is entirely possible that an operating mode is not defined correctly and is always dropped by VPR during packing.
+Therefore, it is necessary to validate the correctness of each operating mode.
+Here, we will use the syntax ``disable_packing`` which can efficiently reach the goal.
```
kmurray, line 216: Probably should clarify that this is being done temporarily to validate the correctness of the mode definitions.

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+   <mode name="n2_lut5">
+
+to 
+
+.. code-block:: xml
+
+   <mode name="n2_lut5" disable_packing="true">
+
+As a result, VPR packer will consider the 6-input LUT mode during packing.
```
kmurray, line 230: 'will consider' -> 'will only consider'

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+We can try a benchmark `mult_2x2.blif
+<https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/benchmarks/microbenchmarks/mult_2x2.blif>`_
+by following the design flow tutorial :ref:`basic_design_flow_tutorial`.
+If the flow-run succeed, it means that the 6-input LUT mode is functional correct.
```
kmurray, line 234: 'functional correct' -> 'being successfully used by the packer'

Functional correctness is a somewhat stronger statement than what is being tested here (e.g. could be interpreted as being formally equivalent to the hardware). It is likely better to just say the packer was able to use it correctly.

doc/src/arch/reference.rst
```diff
@@ -1430,6 +1430,16 @@ The following tags are common to all <pb_type> tags:
 
     .. note:: If a ``<pb_type>`` has only one mode of operation the mode tag can be omitted.
 
+    :opt_param disable_packing:
+        Specify if a mode is disabled or not for VPR packer.
+        When a mode is defined to be disabled for packing (``disable_packing="true"``), packer will not map any logics to the mode.
```
kmurray, line 1435: 'logics' -> 'logic'

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+   </mode>
+   <!-- 6-LUT mode definition end -->
+
+Full example can be found at `link
+<https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/timing/k6_frac_N10_40nm.xml>`_.
+
+Validation in packer
+--------------------
+After finishing the architecture description, the next step is to validate that VPR can map logics to each operating modes.
+Since VPR packer will exhaustively try each operating mode and finally map logics to one of it.
```
kmurray, line 211: 'logics' -> 'logic'

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+   <mode name="n2_lut5">
+
+   <mode name="n1_lut6" disable_packing="true">
+
+In this case, VPR packer will consider the dual 5-input LUT mode during packing.
+We can again try the same benchmark `mult_2x2.blif
+<https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/benchmarks/microbenchmarks/mult_2x2.blif>`_
+by following the design flow tutorial :ref:`basic_design_flow_tutorial`.
+If the flow-run succeed, it means that the dual 5-input LUT mode is functional correct.
```
kmurray, line 256: Same as above

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+Finally, we validated that both operating modes are functional correct.
```
kmurray, line 258: 'Finally, after having validated' ... 'that both modes are being successfully used by the packer.'

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+We can enable both operating modes by changing to
```
kmurray, line 259: 'enable' -> 're-enable'

doc/src/tutorials/arch/multi_mode_ble.rst
```diff
@@ -0,0 +1,303 @@
...
+
+Finally, we validated that both operating modes are functional correct.
+We can enable both operating modes by changing to 
+
+.. code-block:: xml
+
+   <mode name="n2_lut5">
+   <mode name="n1_lut6">
+
+Now, VPR packer will consider best-fit operating mode to mapping logics.
```
kmurray, line 266: 'Now, VPR's packer will try to choose the best operating modes to use.

mithro:
@litghost Could you do a quick look see here?

tangxifan:
@kmurray Thanks for the comments. It helps a lot in improving the quality of this tutorial. I have applied these modifications and added the test case ``strong_pack_disable`` to the strong regression tests. I used the same architecture and benchmark as the tutorial.

litghost:
I'll take a look at this next week.

tangxifan:
Hi all, can I have your latest opinion? It has been quite a while since the last update.

kmurray:
Looks good to me.

mithro:
@litghost is out of office until Friday, so I think we can proceed with @kmurray's approval.

Issue #912, opened 2019-08-12
Odin II adders don't use initial carry-in value


ademmings:
#### Expected Behaviour
When instantiating an adder block, the carry-in value that's initially passed in should be used in the addition.

#### Current Behaviour
The initial carry-in value has no effect on the addition.

#### Example code:

``` Verilog
module top_module
(
    
    input [1:0] a1, b1, a2, b2,
    input cin1, cin2,
    output [1:0] sumout1, sumout2,
    output cout1, cout2
);

    adder a1 (.a(a1), .b(b1), .cin(cin1), .sumout(sumout1), .cout(cout1));
    adder a2 (a2, b2, cin2, sumout2, cout2);

endmodule
```
** the values of cin1 and cin2 have no effect on the addition of a1 + b1 or a2 + b2.

PR #1316, opened 2020-05-25
Carry-in effect issue in hard adders 

sdamghan:
#### Description
NOTE: The environment that I will clarify in the following is related to using hard adders.

The first dummy adder in the split adder chain is responsible for the general carry-in of the circuit. The inputs of the first dummy adder are connected to the ground and the cin is connected to the carry-in of the entire hard adder. Currently, the cout of the first dummy adder is connected to the cin of the next adder. However, the sumout of the first dummy adder, which contains the value of the general carry-in, should connect to the next cin. This is come from "ground+ground+cin=sumout+cout."

#### Related Issue
#912

#### Related files
1-bit hard adder
[test1.v.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686804/test1.v.txt)
[test1_input.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686805/test1_input.txt)
[test1_output.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686806/test1_output.txt)
[out1.blif.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686803/out1.blif.txt)

3-bit hard adder
[test2.v.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686811/test2.v.txt)
[test2_input.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686812/test2_input.txt)
[test2_output.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686813/test2_output.txt)
[out2.blif.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4686810/out2.blif.txt)

#### Command
```
./odin_II -a ../vtr_flow/arch/timing/k6_frac_N10_frac_chain_mem32K_40nm.xml -V test1.v -t test1_input -T test1_output -o out1.blif
```
```
./odin_II -a ../vtr_flow/arch/timing/k6_frac_N10_frac_chain_mem32K_40nm.xml -V test2.v -t test2_input -T test2_output -o out2.blif
```

#### Types of changes

#### Checklist:


sdamghan:
@kmurray Hi Kevin, According to Issue #912, the general carry-in did not affect in hard adders. However, in the first commit, by changing the pin connections, the problem has been solved. Based on @jeanlego 's request, we decided to remove dummy adders. So, Do you mind to keep dummy adders at the beginning and the end of the adders chain? or to keep only dummy carry-in?

jeanlego:
More specifically -> 
this PR remove the global cin capability in odin.
Due to a bug, the first carry in was completely disregarded and, to resolve it, the simplest solution is simply disregard the dummy adders that pad a hard-block adders. I have tried digging through literature but it didn't turn anything substantial.
 I assume this was added during the carry-chain paper and we are trying to figure out what is the purpose of a soft logic 1 bit full adder padding the start and the tail of a chain of adder. There seems to be no side effect noticable by the removal but we wanted to make sure. Do you have any knowledge of this? 
I see one of you guys added a global_cin flag last summer (I'll add the exact commit once I'm on my PC) so I presume it has some specific functionality I am not aware of.

Thanks

kmurray:
I think this needs more careful consideration. My understanding was that the dummy adder was a requirement for correct functionality on some architectures, so I'd be hesitant to just remove it. (if it were to be removed this would require very detailed verification across both ODIN, ABC and VPR to ensure we were not breaking any assumptions).

Adding @MohamedEldafrawy since he put in the --adder_cin_global option and likely has more insight. Also adding @vaughnbetz.

kmurray:
> This is come from "ground+ground+cin=sumout+cout."

I agree it makes sense that this would only cause sumout to toggle based on cin.

However wouldn't a fix simply be to set "vcc+gnd+cin=sumout+cout"? So that toggling cin would cause cout to toggle accordingly?

kmurray:
Additionally, I think this may indicate a deeper issue. If we have been generating cin to adders incorrectly (i.e. functionally incorrect output), how come this hasn't been caught by any regression tests?

I know we do formal logical equivalence tests between ODIN's generated BLIF and the post VPR BLIF using ABC. Doesn't ODIN do simulation based verification between the input RTL and it's generated BLIF? Shouldn't that have caught this?

jeanlego:
> Additionally, I think this may indicate a deeper issue. If we have been generating cin to adders incorrectly (i.e. functionally incorrect output), how come this hasn't been caught by any regression tests?

Lack of test there... the first cin is rarely ever really used unless a hard-adder is specifically requested and the cin is attached, wich we have no test that does that.

kmurray:
> Lack of test there... the first cin is rarely ever really used unless a hard-adder is specifically requested and the cin is attached, wich we have no test that does that.

Hmm, it would be good to cover this case.

A common scenario is a combined adder/subtractor which usually requires feeding a different cin based on whether addition or subtraction is being performed.

jeanlego:
> Lack of test there... the first cin is rarely ever really used unless a hard-adder is specifically requested and the cin is attached, wich we have no test that does that.

correction: rarely if ever used in the test we have specifically.

@sdamghan would you mind adding more tests along with the bug fix to cover these cases in this PR?

MohamedEldafrawy:
@kmurray The dummy adder feature in ODIN II was there before I added any of my changes. I might have fixed a bug or added a feature to that but the core feature was already there.

Let me give some details on why ODIN II adds a dummy adder at the start and the end of an addition:

To start and addition/subtraction we need to drive the Cin of the first adder with 0/1. To do that ODIN II have two options:

1. Connect the first Cin of this adder/subtractor chain directly to gnd/vcc (which is the global cin feature I added)
2. Add a dummy adder with both inputs connected to gnd to implement an addition or both connected to vcc to implement a subtraction. By doing so regardless of the value of the Cin of this dummy adder (which is left unconnected in this case) we are forcing the Cout of this dummy adder to be either 0 or 1 for addition or subtraction. 

Option 2 is very useful to give the packer a much higher degree of freedom when packing adder chains (more specifically small adder chains with num_adders < total_adders_per_lab). Since in option 1, the packer is forced to pack the adder chain at the top of the LAB so that Cin is connect to the Cin port of LAB.

I have added a feature to VPR before to support architectures with more than one Cin entry points for instance before the first ALM and the middle ALM (there is a reg test for that). However, all the architectures except the one in the regtest don't actually make use of this feature and that's fine since it's an architecture feature that could be added to only a subset of the architectures.

Finally, the dummy adder at the end of an adder/subtraction chain is always needed since this is how we route the last Cout to the out of the ALM (connecting both adder inputs gnd causing Cout to be passed to Sout of this dummy adder)

To add support for the issue you are talking about you can do the following:

1. For the global Cin mode: just connect the Cin coming from the rtl to the Cin of the first adder
2. For the dummy adder mode: instead of connecting both adder inputs to 0 or 1 connect both of them to the Cin signal defined in the rtl.

I hope this answers your question.

sdamghan:
@MohamedEldafrawy Seems reasonable having dummy adders. Thank you for the clarification. 
Would you please explain more about the offset feature which you implemented? (Specifically, for offset = 0) What was the purpose of connecting the Cin of the first adder/subtractor chain to GND/VCC? 
I think the dummy adder at the beginning of the chain, which has two inputs connected to GND and the Cin connected to the entire hard adder Cin, can handle both situations. I mean, the Cin will propagate to the adder/subtractor chain and so on. So, it can handle the addition/subtraction either the Cin equals to 0 or 1, without any need of specific GND connection.

jeanlego:
from @sdamghan 
> So, if I want to add more benchmarks, they should cover subtraction beside addition. How can I find out do we have any subtractor hard blocks in arch file. I mean is it possible to define like this: `subtraction s1 (a,b,cin,subout,cout)`

I am unaware of a subtractor hard-block, I believe to do that you twos complement the term you are negating with. The simplest way here would be to bitwise not the second term and drive a one in the first cin.

MohamedEldafrawy:
@sdamghan So if you are doing an addition the first Cin will be gnd (0) Sum = A + B. While to do a subtraction in hardware you will do Sub = A + ~B + 1 where the ~B + 1 is the two's complement of B. Then in this case you need to connect the first Cin to vcc (1).

If you connect both inputs of the dummy adder to GND the cout of the dummy adder will always be 0. Also, the main reason we use dummy adders, is that we don't want to have an input connect to the Cin of the first adder which will add constraints on packing as I mentioned in my previous comment. So, the solution I mentioned before is what solves this issue by keeping the dummy adder and connecting both of it's inputs to Cin so that this Cin propagates the carry in input of the first adder.

sdamghan:
@jeanlego, please check and verify the changes.

ODIN_II/regression_test/benchmark/verilog/micro/hard_adder_cin_propagation.v
```diff
@@ -0,0 +1,15 @@
...
+    input cin,
+    output [1:0] sumout,
+    output cout
+);
+
+	wire  cout1;
+	wire [1:0] sumout1;	
+
+	adder add1 (a, b, cin, sumout1, cout1);
+	adder add2 (.a(c), .b(sumout1), .cin(cout1), .sumout(sumout), .cout(cout));
```
jeanlego, line 13: can you make create two 2 bit adders using 4 bit wide input

```
input [3:0] a, b;
output [4:0] out;

wire cout;

adder add1(.a(a[1:0]), .b(b[1:0]), .cin(1'b0), .sumout(out[1:0]), .cout(cout))
adder add2(.a(a[3:2]), .b(b[3:2]), .cin(cout), .sumout(out[3:2]), .cout(out[4]))
```

this will makes the vector more readable, 
also avoid mix and match instance port assignment, please stick to named and ordered since this is not testing port declaration syntax

jeanlego:
@sdamghan, you are changing QoR results, please follow the procedure outlined here https://docs.verilogtorouting.org/en/latest/dev/developing/#evaluating-quality-of-result-qor-changes

jeanlego:
@sdamghan if you squash your commits and resolve the conflicts this looks fine to me
@kmurray this affect some of the QoR, I would rather have your eye on this to make sure this seems reasonable to you

kmurray:
> @kmurray this affect some of the QoR, I would rather have your eye on this to make sure this seems reasonable to you

Sure, it would be good to see a detailed QoR comparison on the nightly vtr_reg_qor_chain test. For methodology on how to do this see [here](https://docs.verilogtorouting.org/en/latest/dev/developing/#evaluating-quality-of-result-qor-changes). The output of this should be a spreadsheet (see link for how to generate it using a script) comparing master to this PR.

sdamghan:
> > 
> 
> Sure, it would be good to see a detailed QoR comparison on the nightly vtr_reg_qor_chain test. For methodology on how to do this see [here](https://docs.verilogtorouting.org/en/latest/dev/developing/#evaluating-quality-of-result-qor-changes). The output of this should be a spreadsheet (see link for how to generate it using a script) comparing master to this PR.

@kmurray, I've generated the new QoR measurements for vtr_reg_nightly/vtr_req_qor_chain according to the link you mentioned. The results didn't change in comparison with the previous state.

sdamghan:
@kmurray, Here are the comparison spreadsheet plus the master and PR related parse_result files.

[comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4725628/comparison.xlsx)
[Master_parse_results.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4725629/Master_parse_results.txt)
[PR_parse_results.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4725630/PR_parse_results.txt)

kmurray:
@sdamghan QoR mostly looks good. However critical path delays seem to be missing, we would need those metrics as well to fully verify. Can you look into it?

I did make changes to the printing parsing of QoR critical path delay data, so if you're parsing previously generated results with the latest scripts that might be why the critical path data didn't show up. If so, either try re-running, or use the parsing scripts from the same revision as the code which was tests.

sdamghan:
@kmurray I ran it again, here are the new results:

[master_parse_results.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4747626/master_parse_results.txt)
[pr_parse_results.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4747628/pr_parse_results.txt)
[new_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4747627/new_comparison.xlsx)

kmurray:
Looks good from a QoR perspective.

I'll leave it to @jeanlego to check over the rest.

Issue #1307, opened 2020-05-20
Routing Traceback Order Issue


andrewboutros:
I am working on implementing a SAT-based router in VPR. My router creates routing traces that are supposed to be legal according to the data structure description, however, they cause a crash.

My explanation is that there is an underlying assumption that routing branches are supposed to be listed in order (when I fixed the order of my generated traces, the problem was resolved). The figure below shows a (synthetic) example of a trace that is legal but causes a crash.

![image](https://user-images.githubusercontent.com/39923017/82495871-cc927900-9ab9-11ea-8ba0-fa64f85bc7e0.png)

After I generate the routing trace and call the "load_net_delay_from_routing" routine> My first order investigation shows that VPR crashes in the load_new_subtree_R_upstream. My guess is that there might be a problem in the code that converts traces to routing trees before calculating delays for it.

dpbaines:
The problem is a result of dangling branches I've decided, a few tests I've ran have resulted in successful parsing no matter the ordering of branches. I only get the seg fault in the functions listed above if there is a dangling branch, I will fix it so it produces an error message instead of crashing.

PR #1318, opened 2020-05-25
Adding improved error checking for the route file parser

dpbaines:
#### Description

I added a simple checker that checks for continuity between traces in read_route.cpp, if a weird discontinuity is detected it throws an exception, and tells the user which net and line threw the error. This is an improvement over the segmentation fault that happened before hand.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1307

#### Motivation and Context

Following the issue described by Andrew; where the program would crash if VPR is given an externally produced route file, it was determined that the bug was caused by the introduction of a 'dangling branch', which is when the route trace has a discontinuity between traces.

I ran the regression tests and as expected there is no impact on the normal flow. I also ran several flows, and played around with the generated route files, checking to see if random moving or deleting of nodes caused errors, which they did. Furthermore I checked to make sure it wouldn't cause issues if the route file was not modified. 

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/read_route.cpp
+++ b/vpr/src/base/read_route.cpp
@@ -50,6 +50,7 @@ static void process_global_blocks(std::ifstream& fp, ClusterNetId inet, const ch
 static void format_coordinates(int& x, int& y, std::string coord, ClusterNetId net, const char* filename, const int lineno);
 static void format_pin_info(std::string& pb_name, std::string& port_name, int& pb_pin_num, std::string input);
 static std::string format_name(std::string name);
+static bool check_rr_graph_connectivity(RRNodeId prev_node, RRNodeId node);
 
 /*************Global Functions****************************/
 bool read_route(const char* route_file, const t_router_opts& router_opts, bool verify_file_digests) {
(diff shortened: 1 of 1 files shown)


vpr/src/base/read_route.cpp
```diff
@@ -445,3 +469,78 @@ static std::string format_name(std::string name) {
         return nullptr;
     }
 }
+
+/* Check if there is a discontinuity in the trace
+ * There is more to this but I'll test thoroughly, my priority is not to break it */
+static bool check_dangling_branch(int x1, int y1, int x2, int y2, int prev_x1, int prev_y1, int prev_x2, int prev_y2, std::string curr_type, std::string prev_type) {
+    
+    if(prev_x1 == -1) return false;
```
kmurray, line 477: It looks like this code is checking for connectivity by comparing the RR node co-ordinates. This won't be robust in practice since it is possible to have wires which are adjacent but not connected.

I think a more robust way to do this would be to check explicitly for connectivity based on the Routing Resource (RR) graph. I'd suggest reading through the RR graph header files to learn about it and the interface.

A rough sketch of what this could look like is:
```cpp
bool rr_nodes_are_connected(RRNodeId prev_node, RRNodeId node) {
    const auto& rr_graph = device_ctx.rr_nodes;

    for (RRedgeId edge : rr_graph.edge_range(prev_node)) {
        if (rr_graph.edge_sink_node(edge) == node) {
            return true;
        }
    }
    return false;
}
```

kmurray:
@dpbaines Another item to be aware of when writing this check is that we do expect some dangling nodes, if they are part of a 'non-configurable node set'.

kmurray:
@dpbaines Looks like this commit includes a large number of other updates which have already been comited to master. Perhaps it should be rebased onto the latest master?

PR #1319, opened 2020-05-25
Commandline to gui

MahshadJ:
added command line options: draw_block_outline, draw_block_text, clip_routing_util, net_max_fanout, set_net_alpha, to the user interface.

#### Description

Added the 3 of the 5 mentioned options in the main.ui file. Added their callback functions in draw.cpp. Added the other 2 widgets and buttons in buttons.cpp. The alpha value in the intra_block.cpp can be adjusted with the gui now and the command line options are all in the gui as well.

#### Motivation and Context

This way the user can make all changes to the circuit through the gui only, which is easier and more intuitive.

I ran vpr with disp on, on multiple circuits and made sure the buttons and widget make their corresponding change.

#### Types of changes

#### Checklist:


kmurray:
Looks like these commits are trying to add the modified files into the root of the VTR source tree, instead of modifying the relevant files in the source tree (e.g. under vpr/src/draw).

@MahshadJ Where you uploading these via the github web browser interface? Usually these kinds of change should be commited to a branch within your local git repo and then pushed via `git push` to a branch on Github, from which you open the PR. (This ensures that the modifications are tracked within the existing files.).

kmurray:
@MahshadJ 

Structure looks much better now! You'll notice that the 'Travic CI - Pull Request' is failing. Can you take a look at that? I expect its the code auto-formatting, see [here](https://docs.verilogtorouting.org/en/latest/dev/developing/#code-formatting) for details.

vpr/src/draw/buttons.cpp
```diff
@@ -53,6 +91,36 @@ void button_for_toggle_nets() {
...
+
+    //find maximum fanout
+    auto& cluster_ctx = g_vpr_ctx.clustering();
+    auto& clb_nlist = cluster_ctx.clb_nlist;
+    size_t max_fanout = 0;
+    for (ClusterNetId net_id : clb_nlist.nets())
+        max_fanout = std::max(max_fanout, clb_nlist.net_sinks(net_id).size());
+
+    //spin box for net_max_fanout, set the range and increment step
+    GtkWidget* net_max_fanout_widget = gtk_spin_button_new_with_range(0, std::numeric_limits<int>::max(), 1.);
```
kmurray, line 106: Shouldn't the high-end of this range be `max_fanout`, rather than std::numeric_limits<int>::max()?

Currently, if the user enters a large number (e.g. `999999999999`which is much greater than max_fanout), it is replaced with `2147483647` (int max). Would be better to replace it with the value of `max_fanout`.

vpr/src/draw/buttons.cpp
```diff
@@ -26,6 +27,43 @@ gint label_left_start_col = 0;
...
+
+    //text entry for apha value
+    GtkWidget* entry = gtk_entry_new();
+    std::string initialValue = std::to_string(get_net_alpha());
+    gtk_entry_set_text((GtkEntry*)entry, initialValue.c_str());
+    gtk_entry_set_input_purpose((GtkEntry*)entry, GTK_INPUT_PURPOSE_NUMBER);
+    gtk_widget_set_name(entry, "alphaValue");
+
+    //label
+    GtkWidget* alpha_label = gtk_label_new("Set alpha value (<1)");
```
kmurray, line 46: Change text to `Set net transparency [0., 1.]` ?

Also, perhaps this button should also be next to the other net-related buttons (e.g. after 'Net Max Fanout').

vpr/src/draw/draw.cpp
```diff
@@ -152,6 +155,7 @@ void run_graphics_commands(std::string commands);
 constexpr float SB_EDGE_TURN_ARROW_POSITION = 0.2;
 constexpr float SB_EDGE_STRAIGHT_ARROW_POSITION = 0.95;
 constexpr float EMPTY_BLOCK_LIGHTEN_FACTOR = 0.20;
+float net_alpha = 0.0275;
```
kmurray, line 158: We should move `net_alpha` to be a member of t_draw_state, rather than a file-scope variable.

That should ensure that if the draw state is reset, so is the net alpha.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -547,7 +547,8 @@ void draw_logical_connections(ezgl::renderer* g) {
 
     g->set_line_dash(ezgl::line_dash::none);
 
-    constexpr float NET_ALPHA = 0.0275;
+    //constexpr float NET_ALPHA = 0.0275;
+    float NET_ALPHA = get_net_alpha();
```
kmurray, line 551: It seems like alpha only controls the drawn net alpha when 'Toggle Nets:' is set to 'Logical Connections'.

We should also update drawnets (responsible for drawing nets with 'Toggle Nets:' set to 'Nets, I think?) also uses this alpha value.

vpr/src/draw/buttons.cpp
```diff
@@ -26,6 +27,43 @@ gint label_left_start_col = 0;
...
+
+    //button
+    GtkWidget* button = gtk_button_new_with_label("set");
```
kmurray, line 49: It seems like the user must press the 'set' button to change the alpha value. It doesn't seem to work if the user enters a value and just presses enter. 

It would be good to support pressing enter.

vpr/src/draw/draw.cpp
kmurray: We should also probably modify the default to be a little less transparent (e.g. 0.1), as with 0.0275 single connections are very faint.

vpr/src/draw/buttons.cpp
```diff
@@ -96,14 +101,24 @@ void button_for_net_max_fanout() {
...
     for (ClusterNetId net_id : clb_nlist.nets())
         max_fanout = std::max(max_fanout, clb_nlist.net_sinks(net_id).size());
 
+    auto& atom_ctx = g_vpr_ctx.atom();
+    auto& atom_nlist = atom_ctx.nlist;
+    size_t max_fanout2 = 0;
+    for (AtomNetId net_id : atom_nlist.nets())
+        max_fanout2 = std::max(max_fanout2, atom_nlist.net_sinks(net_id).size());
+
+    max_fanout2 > max_fanout ? max = max_fanout2 : max = max_fanout;
```
kmurray, line 118: It's probably clearer to do this as:
```cpp
size_t max = std::max(max_fanout, max_fanout2);
```

vpr/src/draw/draw.cpp
```diff
@@ -3767,12 +3772,21 @@ static void highlight_blocks(double x, double y) {
...
+    draw_state->net_alpha = std::stof(fa);
+    application.refresh_drawing();
+}
+
+void set_net_alpha_value_with_enter(GtkWidget* widget, gint /*response_id*/, gpointer /*data*/) {
+    std::string fa(gtk_entry_get_text((GtkEntry*)widget));
+    t_draw_state* draw_state = get_draw_state_vars();
+    draw_state->net_alpha = std::stof(fa);
     application.refresh_drawing();
 }
```
kmurray: This looks identical to `set_net_alpha_value`. Could just use the same function rather than duplicating the code?

vpr/src/draw/draw.cpp
```diff
@@ -155,7 +156,7 @@ void run_graphics_commands(std::string commands);
 constexpr float SB_EDGE_TURN_ARROW_POSITION = 0.2;
 constexpr float SB_EDGE_STRAIGHT_ARROW_POSITION = 0.95;
 constexpr float EMPTY_BLOCK_LIGHTEN_FACTOR = 0.20;
-float net_alpha = 0.0275;
+//float net_alpha = 0.1;
```
kmurray, line 159: Should probably remove this since it's now unused.

kmurray:
Thanks for putting this all together @MahshadJ!

Issue #1291, opened 2020-04-30
Odin: Comparing simulation output file with generated output (header specifically)


emacdo12:
#### Expected Behaviour
Odin should compare the titles by name and make sure they match. Also be case sensitive (which I think odin is). 

#### Current Behaviour
Odin uses strcmp thus comparing the names and blank spaces not the tokens. 

#### Possible Solution
Compare the tokens instead of using strcmp 

#### Steps to Reproduce

1. create three files, a verilog file, an input file and an expected output file.
2. build odin then run "./odin_II -V <test.v> -t <input_vector>  -T  <output_vector> 

In this example this verilog code was used: 

```Verilog
module simple_op(out,a,b);
input a;
input b;
output out;

nand(out,a,b);
endmodule
``` 
This input file was used: 
```
GLOBAL_SIM_BASE_CLK a b
1 0 1 
0 1 1 
1 0 1 
0 0 1 
1 1 1 
0 0 0 
1 0 1 
0 1 1 
1 0 0 
0 1 1
```
This output file was used: 
```
out  
0 
1 
0 
0 
1 
0 
0 
1 
0 
1 
```

```terminal 
emacdo12@CASA37:~/workspace/vtr-verilog-to-routing/ODIN_II$ ./odin_II -V regression_test/benchmark/verilog/Keywords/unary_bitwise_andnode.v -t regression_test/benchmark/verilog/Keywords/unary_bitwise_andnode_input -T regression_test/benchmark/verilog/Keywords/unary_bitwise_andnode_output 
--------------------------------------------------------------------
Welcome to ODIN II version 0.1 - the better High level synthesis tools++ targetting FPGAs (mainly VPR)
Email: jamieson.peter@gmail.com and ken@unb.ca for support issues

Using Lut input width of: -1
Verilog: unary_bitwise_andnode.v
--------------------------------------------------------------------
High-level synthesis Begin
Parser starting - we'll create an abstract syntax tree. Note this tree can be viewed using Grap Viz (see documentation)
Adding file regression_test/benchmark/verilog/Keywords/unary_bitwise_andnode.v to parse list
Optimizing module by AST based optimizations
Converting AST into a Netlist. Note this netlist can be viewed using GraphViz (see documentation)
==========================
Detected Top Level Module: 	simple_op
==========================
Performing Optimizations of the Netlist
Performing Partial Map to target device
Outputting the netlist to the specified output format
Successful High-level synthesis by Odin
[...]

PR #1320, opened 2020-05-26
Output vector Header bug fix

emacdo12:
#### Description

Created output_vector_headers_equal which compares the tokens of each vector header and ignores white spaces. This function calls on string_to_token_array which tokenizes each string and returns as an array as well as initializes a pointer passed to the size of the array. Before a strcmp was used.
#### Related Issue

Issue #1291, a bug that says the vector headers are mistmatched if the blank spaces differ.

#### Motivation and Context

There was a minor bug and this remedies it.

I ran through make test. As well, I created a seperate c file where I could test different headers easily to ensure they passed and failed when they should.

#### Types of changes

#### Checklist:


jeanlego:
lgtm :)

PR #1330, opened 2020-06-02
Produce more detailed post-placement timing results.

kmurray:
#### Description

Provides the same timing metrics after placement, as we do after routing.


kmurray:
@MohamedElgammal FYI as we discussed, with this PR we will now report the geomean non-virtual critical path delay (a good metric for multi-clock circuits).

PR #1341, opened 2020-06-04
Add -check_incremental_sta_consistency option to regression tests

Bill-hbrhbr:
#### New Features
A regression tests' command-line option: **-check_incremental_sta_consistency**

#### Motivation and Context
Sanity check that full and incremental STA produces the same packing, placements, routing and timing reports (7 files in total).
**my $packing_file = "$benchmark_name" . ".net";
my $placement_file = "$benchmark_name" . ".place";
my $routing_file = "$benchmark_name" . ".route";
my $setup_timing_report = "report_timing.setup.rpt";
my $hold_timing_report = "report_timing.hold.rpt";
my $unconstrained_setup_timing_report = "report_unconstrained_timing.setup.rpt";
my $unconstrained_hold_timing_report = "report_unconstrained_timing.hold.rpt";**

Tested on Arch_Flagship + Stereovision3.v, Arch_Flagship + LU8PEEng.v, and Arch_Titan + Neuron Benchmark.

#### Types of changes

#### Checklist:


vpr/src/base/read_options.cpp
```diff
@@ -1296,6 +1296,10 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .metavar("RR_GRAPH_FILE")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    file_grp.add_argument(args.TimingReportPrefix, "--timing_report_prefix")
+            .help("Prefix for timing report files")
```
kmurray, line 1300: We could probably just add:
```
    .default("")
```
Instead of manually setting it below.

vpr/src/base/read_options.cpp
```diff
@@ -2133,6 +2137,11 @@ void set_conditional_defaults(t_options& args) {
         args.RouteFile.set(route_file, Provenance::INFERRED);
     }
 
+    if (args.TimingReportPrefix.provenance() != Provenance::SPECIFIED) {
+        std::string timing_report_prefix = "";
+        args.TimingReportPrefix.set(timing_report_prefix, Provenance::INFERRED);
+    }
+
```
kmurray, line 2144: Probably preferred to use .default() as described above.

kmurray:
Also seems like its failing the [code formatting](https://docs.verilogtorouting.org/en/latest/dev/developing/#code-formatting) check (see 'Travis CI -Pull Request'). Could you take a look?

vpr/src/base/vpr_api.cpp
```diff
@@ -1207,9 +1208,9 @@ void vpr_analysis(t_vpr_setup& vpr_setup, const t_arch& Arch, const RouteStatus&
 
         //Timing stats
         VTR_LOG("\n");
-        generate_hold_timing_stats(/*prefix=*/"", *timing_info,
+        generate_hold_timing_stats(/*prefix=*/vpr_setup.AnalysisOpts.timing_report_prefix, *timing_info,
                                    *analysis_delay_calc, vpr_setup.AnalysisOpts);
-        generate_setup_timing_stats(/*prefix=*/"", *timing_info,
+        generate_setup_timing_stats(/*prefix=*/vpr_setup.AnalysisOpts.timing_report_prefix, *timing_info,
```
kmurray, line 1213: This should also be removed.

The prefix is used when printing out information to stdout/vpr log, since this is called in multiple places (e.g. post-place, post-route). It should not have anything set here.

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -118,6 +120,7 @@
 my $check_route             = 0;
 my $check_place             = 0;
 my $use_old_abc_script      = 0;
+my $check_incremental       = 0;
```
kmurray, line 123: `check_incremental` is a bit generic. I'd suggest something a bit more explicit, like `check_incremental_timing_consistency`. Similarly for the actual command-line option below.

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -1017,6 +1023,168 @@
                 }
             }
         }
+	} elsif ($check_incremental) { # specified channel width && check incremental
```
kmurray, line 1026: This should probably be a dedicated if statement, independent of the `if(!route_fixed_W) ... else ...`.

As currently written this will actually run VPR 3 times:
1. Initial (pack, place)
2. Check full (pack, place, route)
3. Check incremental (pack, place, route)

I'd suggest re-working it so that when 'check_incremental` is set, it is run separately. So something like:
1. Initial (pack, place, route at fixed W)
2. Check incremental (pack, place, route)

This also keeps the code-paths clearer. At the moment the check_incremental path is embedded in the middle of the 'normal' flow.

I'd also suggest putting the check_incremental into a separate function to further help organize things.

Bill-hbrhbr:
@kmurray Updated the PR according to your comments. Please review it again when you have time. Thanks!

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -467,6 +473,16 @@
...
 my $q         = "not_run";
 
+# Default output file names
+my $packing_file = "$benchmark_name" . ".net";
+my $placement_file = "$benchmark_name" . ".place";
+my $routing_file = "$benchmark_name" . ".route";
+my $setup_timing_report = "report_timing.setup.rpt";
+my $hold_timing_report = "report_timing.hold.rpt";
+my $unconstrained_setup_timing_report = "report_unconstrained_timing.setup.rpt";
+my $unconstrained_hold_timing_report = "report_unconstrained_timing.hold.rpt";
```
kmurray, line 483: Seems like these are only used in cmp_full_vs_incr_STA(), we should probably define them within that function scope to avoid polluting the file-scope namespace (which has too many variables already).

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -1173,6 +1189,14 @@
             }
         }
 
+        # Do a second-run of the incremental analysis to compare the result files
+        if ($check_incremental_sta_consistency) {
+            if (&cmp_full_vs_incr_STA ne "success") {
```
kmurray, line 1194: I'd re-structure this as something like:

```
my $result = &cmp_full_vs_incr_STA();

if ($result ne "success") {
    $error_status = $result'
    $error_code = 1;
```

That both makes it clear that we are doing a function call, and preserves the value returned (which should provide details about the error as noted below)

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -1687,3 +1711,74 @@ sub calculate_relaxed_W {
 
     return $relaxed_W;
 }
+
+sub cmp_full_vs_incr_STA {
+	# loop indices
+	my @indices = (0..6);
```
kmurray, line 1717: I think using an array for the indices is unneccesary, since you have an array of the actual strings below.

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -1687,3 +1711,74 @@ sub calculate_relaxed_W {
...
+
+	# default output file names
+	my @default_output_names = (
+			$packing_file,
+			$placement_file,
+			$routing_file,
+			$setup_timing_report,
+			$hold_timing_report,
+			$unconstrained_setup_timing_report,
+			$unconstrained_hold_timing_report
```
kmurray, line 1727: I'd probably just replace this with the strings explicitly:
```
my @default_output_names= (
       $benchmark_name . ".net",
       $benchmark_name . ".place",
       ....
);
```

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -1687,3 +1711,74 @@ sub calculate_relaxed_W {
...
+		);
+
+	# The full STA flow should have already been run
+    # directly rename the output files
+    foreach my $index (@indices) {
```
kmurray, line 1732: I think this and the similar loops below can just be:
```
foreach my $filename (@default_output_names) {
    ...
}
```
which is clearer.

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -1687,3 +1711,74 @@ sub calculate_relaxed_W {
...
+            $default_output_names[$index], 
+            "full_" . $default_output_names[$index]
+        );
+    }
+
+    my @incremental_vpr_args = @forwarded_vpr_args;
+    push(@incremental_vpr_args, ("--timing_update_type", "incremental"));
+
+    # run incremental STA flow
+    my $fixed_W_log_file = "vpr.out";
```
kmurray, line 1745: We should use a unique log file name here, as this will overwrite the previous VPR run's log message in vpr.out. Perhaps something like `vpr.incr_sta.out`.

We generally try to keep all main tool log files around, since this is helpful when debugging a failure.

vtr_flow/scripts/run_vtr_flow.pl
```diff
@@ -1687,3 +1711,74 @@ sub calculate_relaxed_W {
...
+    foreach my $index (@indices) {
+    	my $diff_result = &system_with_timeout(
+            $diff_exec, "diff.out",
+            $timeout, $temp_dir,
+            "full_" . $default_output_names[$index],
+            "incremental_" . $default_output_names[$index]
+        );
+        if ($diff_result ne "success") {
+        	print $index;
+        	return "fail";
```
kmurray, line 1778: I'd expand this error message to show what files failed the diff.

For instance:
```
return "failed: $full_filename $incr_filename differ"
```

kmurray:
Thanks @Bill-hbrhbr this looks good to merge from a code perspective.

The only thing I'm noticing is that the golden reference results seem to be missing (i.e. config/golden_results.txt for the added tests). If you can add those (use parse_vtr_flow.pl -create_golden), and fix the merge/rebase issue this should be good to go!

PR #1343, opened 2020-06-04
Add support for drawing RR SOURCE and SINK nodes

kmurray:
#### Description
Adds support for drawing SOURCE and SINK RR nodes in VPR graphics.

Also cleans-up RR drawing combobox options, and avoid expensive copying of RR nodes in pin drawing.


kmurray:
*Before:*
image


*After:*
image
where the RR source/sink nodes are drawn vertically in the centre of the logic block. Their input/output edges are also drawn.

Issue #1334, opened 2020-06-03
Memory leak when packing disable code causes failures


kmurray:
#### Expected Behaviour

Code should be free of memory errors.

#### Current Behaviour

When with sanitizers enabled there is an [error](https://source.cloud.google.com/results/invocations/a90d3085-ffd1-42fe-a51d-907094009e6f/targets/foss-fpga-tools%2Fverilog-to-routing%2Fupstream%2Fcontinuous%2Fstrong_sanitized/log):
```
pack_disable:             k6_frac_N10_40nm_disable_packing.xml/mult_5x6.blif/common                                      failed: expected to fail with 'exited with return code 1' but was 'exited with return code 23' (took 1.74 seconds)
```
where 23 usually indicates a memory leak detected by ASAN.

The resulting [log file](
https://storage.cloud.google.com/vtr-verilog-to-routing/artifacts/prod/foss-fpga-tools/verilog-to-routing/upstream/continuous/strong_sanitized/30/20200603-055813/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_pack_disable/latest/k6_frac_N10_40nm_disable_packing.xml/mult_5x6.blif/common/vpr.out
) (obtained following [this](https://docs.verilogtorouting.org/en/latest/dev/developing/#downloading-logs-from-google-cloud-storage-gcs)) shows:
```
# Load Timing Constraints took 0.00 seconds (max_rss 228.9 MiB, delta_rss +0.0 MiB)
Error 1: 
Type: Packing
/tmpfs/src/github/vtr-verilog-to-routing/vpr/src/pack/prepack.cpp
Line: 1205
Message: Failed to find any location to pack primitive of type '.names' in architecture

=================================================================
==16776==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 80 byte(s) in 1 object(s) allocated from:
    #0 0x7efc22bcb532 in operator new(unsigned long) (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x99532)
[...]

PR #1344, opened 2020-06-04
bug fix for memory leakage in prepacking stage for disable packing 

tangxifan:
Fix the bug of memory leaking when disable packing is enabled

#### Description

Free allocated memory when the packer fails at pre-packing stage. 

#### Related Issue

This aims to address the issue [https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1334](url)

#### Motivation and Context

My take on this memory leak is that the linked list of molecules is not freed automatically when VPR packer throws a fatal error.
Therefore, codes were added to  free the linked list before throwing a fatal error.

Currently CI looks fine. But I would like to see the problem is solved in kokoro.

#### Types of changes

#### Checklist:

- [] All new and existing tests passed

### DIFF
--- a/vpr/src/pack/prepack.cpp
+++ b/vpr/src/pack/prepack.cpp
@@ -849,7 +849,23 @@ t_pack_molecule* alloc_and_load_pack_molecules(t_pack_patterns* list_of_pack_pat
      * more difficult because now it needs to consider splitting molecules.
      */
     for (auto blk_id : atom_ctx.nlist.blocks()) {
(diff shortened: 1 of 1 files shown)


kmurray:
@tangxifan FYI, I added you as an external collaborator, so Kokoro runs should automatically run for you whenever you open a PR.

tangxifan:
> @tangxifan FYI, I added you as an external collaborator, so Kokoro runs should automatically run for you whenever you open a PR.

Thanks Kevin. I see Kokoro is running now. I will keep patching until it is green.

kmurray:
Thanks for the fix @tangxifan!

PR #1345, opened 2020-06-04
doc: Add notes on how to speed-up VTR edit-compile-debug cycle

kmurray:
In particular document different ways of speeding up compilation.

#### Description

Document ways of speeding up VTR compilation for faster edit-compile-debug cycles (particularly compilation time).

Thanks for @dpbaines for asking questions about this which prompted writing this down.

### DIFF
--- a/README.developers.md
+++ b/README.developers.md
@@ -930,6 +930,69 @@ where ``$VTR_ROOT`` should be replaced with the root of the VTR source tree on y
 [RR](https://rr-project.org/) extends GDB with the ability to to record a run of a tool and then re-run it to reproduce any observed issues.
 RR also enables efficient reverse execution (!) which can be *extremely helpful* when tracking down the source of a bug.
 
+# Speeding up the edit-compile-test cycle
+Rapid iteration through the edit-compile-test/debug cycle is very helpful when making code changes to VTR.
+
+The following is some guidance on techniques to reduce the time required.
+
+# Speeding Compilation
+
+1. Parallel compilation
+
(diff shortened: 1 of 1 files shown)


kmurray:
@MahshadJ @Bill-hbrhbr @MohamedElgammal FYI

MohamedElgammal:
Personally, I found the most beneficial tricks is to:

- compile on multi-thread using -j(n) for n thread
- Turn off VTR_IPO_BUILD ... this can be done easily using `ccmake build/`
- caching compilation data using `cache make  <all make options you want>`

kmurray:
@MohamedElgammal I think most of those are included in this PR? Feel free to comment directly on the diff if there is anything else we should be putting in!

MohamedElgammal:
Great! Thanks for this PR Kevin! It would be so helpful.

README.developers.md
```diff
@@ -930,6 +930,69 @@ where ``$VTR_ROOT`` should be replaced with the root of the VTR source tree on y
...
+    $ sudo apt install ccache
+    ```
+    This only needs to be done once on your development system.
+
+4. Disable Interprocedural Optimizatiaons (IPO)
+
+    IPO re-optimizes an entire executable at link time, and is automatically enabled by VTR if a supporting compiler is found.
+    This can notably improve performance (e.g. ~10-20% faster), but can significantly increase compilation time (e.g. >2x in some cases).
+    When frequently re-compiling and debugging the extra execution speed may not be worth the longer compilation times.
+    In such cases you can manually disable IPO by setting the cmake parmeter `VTR_IPO_BUILD=off`.
```
vaughnbetz, line 979: parameter

Issue #1325, opened 2020-06-01
Enhance VPR lookahead map to account for CHAN to IPIN costs


acomodi:
#### Expected Behaviour

Router correctly predicts the expected cost with correct lookahead map values

#### Current Behaviour

Router mispredicts expected costs to reach a target node from an intermediate one due to incorrectness of the lookahead map.

#### Possible Solution

This issue is related to the Series 7 fabrics.
The Lookahead map consists of expected costs that guide the router in finding the best path to a target node. In case the lookahead is incorrect, the router efficiency is drastically reduced, as it would incur in many backtracking steps due to wrong paths explorations.

Regarding SymbiFlow, a new kind of lookahead map has been used, that makes use of the so called connection boxes. These are additional data structures that hold information on the connection between IPINs and the routing nodes.
In particular, they expose the `site pin delays`, which are not taken into account in the current lookahead map. These delays should be added to the expected cost, as they differ for the various IPINs of a tile.
In fact, a LUT in a CLB tile does not have all the IPINs with the same delay values, but, depending on the IPIN selected, the delay changes.

This delay can be found on the edge connecting a CHAN to an IPIN of the RR graph.

Currently, the lookahead map has a mechanism to capture the delay between the SOURCE/OPIN nodes to the first CHAN[XY] nodes, and this delay is added to the expected cost.

A possible solution for the CHAN --> IPIN delay, is to adopt a similar mechanism:
1. An additional lookahead is to be generated, which captures the delay and congestion cost to get from the last CHAN[XY] node of the wire to the IPIN/SINK node
1. Modify the expected cost function to lookup to the target node, and, if it is a SINK node, apply the extra cost to the final expected cost. This would be equivalent to adding the site_pin_delay of the connection box map.

There is an additional potential issue with the Series 7 fabric which can be summarized as follows.
Before reaching the routing infrastructure, the OPIN/IPIN are connected to a series of two nodes with a single edge each between them. This can be visualized in the image below:
![Untitled Diagram (2)](https://user-images.githubusercontent.com/44773360/83409738-4bba7200-a415-11ea-912f-d2d0bf54b187.png).
It is yet to understand whether this could be an issue in generating a correct lookahead map for the Series 7 family.

#### Context

Symbiflow and Series 7 RR graph and lookahead computation.

acomodi:
@vaughnbetz @kmurray @litghost FYI

acomodi:
[...]

PR #1351, opened 2020-06-08
WIP: Enhancing VPR lookahead map

acomodi:

#### Description

This PR is an initial Draft with the effort of getting the lookahead map work well with the Series 7 architectures.

#### Related Issue

The issue tracking what needs to be done and all suggestions is here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1325

#### Motivation and Context

Currently, the SymbiFlow VTR fork implements the [connection box lookahead map](https://github.com/SymbiFlow/vtr-verilog-to-routing/tree/wip/lookahead_sampling2) which has been tested to work well with Series 7 devices.
The idea is to get the lookahead map code produce similar results, so that eventually, there can be only one lookahead code being used.

In case the two different approaches cannot be merged, an alternative would be to have a common utility library for the different lookaheads.

#### Types of changes

#### Checklist:


acomodi:
@litghost, @kmurray, @HackerFoo FYI.

I have completed the merge of the connection box lookahead map and the upstream lookahead map.

From the connection box lookahead, the following has been taken:
- Sampling method: this has been extracted in a separate file
- Paths expansion and lookahead filling
- Lookahead map format (different capnp schema)

From the upstream lookahead map:
- OPIN --> CHAN information
- CHAN --> IPIN information
- Delta locations calculations
- Map addressability: e.g. `cost_map[CHAN][SEG_ID][DELTA_X][DELTA_Y]`

There are two assertions that needed to be skipped as they were triggered during the place delay matrix calculations:
- https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/1b9065116a6939eea16204885d9d53997a646186/vpr/src/place/timing_place.h#L196
- https://github.com/SymbiFlow/vtr-verilog-to-routing/blob/8980e46218542888fac879961b13aa7b0fba8432/vpr/src/route/connection_box_lookahead_map.cpp#L488

The last one is due to the fact that there are some non valid values for the expected cost which is set to infinite.

I also had to temporarily remove the lookahead generation test under `vpr/test` which still needs to be adapted to the new lookahead generation mechanism.

With this implementation, the routing step has the following results:
- 238.30 seconds
- 18.796 ns (CPD)
- Works on HW

With the connection box lookahead map:
- 146.26 seconds
- 17.381 ns (CPD)
- Works on HW

There is an increase in the CPD which I still need to analyse, as well as an increased run-time. Both of them are probably due to mispredictions, which I still need to analyse in details.

litghost:
> There are two assertions that needed to be skipped as they were triggered during the place delay matrix calculations:

Both of these assertions should be in place, and violating them likely indicates a bug?

libs/libvtrcapnproto/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ endif()
 # Each schema used should appear here.
 set(CAPNP_DEFS
     place_delay_model.capnp
+    connection_map.capnp
```
litghost, line 22: connection_map.capnp shouldn't be required, because it isn't used?

vpr/src/route/route_profiling.cpp
```diff
@@ -29,6 +30,10 @@ void time_on_fanout_analysis() {}
 
 void profiling_initialization(unsigned /*max_net_fanout*/) {}
 
+void conn_start() {}
```
litghost, line 33: The changes to route_profiling.cpp should likely be in another PR.

vpr/src/route/route_timing.cpp
```diff
@@ -1096,6 +1096,8 @@ bool timing_driven_route_net(ConnectionRouter& router,
             conn_delay_budget.short_path_criticality = budgeting_inf.get_crit_short_path(net_id, target_pin);
         }
 
+        profiling::conn_start();
```
litghost, line 1099: Should move out of this PR with the route_profiling.h/.cpp changes

libs/libvtrcapnproto/CMakeLists.txt
litghost: Or if we are going to use this, it should move to map_lookahead.capnp

vpr/test/test_map_lookahead_serdes.cpp
```diff
@@ -1,62 +0,0 @@
-#include "catch.hpp"
```
litghost, line 1: Why did this test get deleted?

vpr/src/route/rr_graph_storage.h
```diff
@@ -48,7 +48,7 @@
  *       This field is valid only for IPINs and OPINs and should be ignored  *
  *       otherwise.                                                          */
 struct alignas(16) t_rr_node_data {
-    int8_t cost_index_ = -1;
+    int16_t cost_index_ = -1;
```
litghost, line 51: This should be put in another PR, to examine changing this independently of the other changes.

vpr/src/route/router_lookahead_map.h
```diff
@@ -1,48 +1,84 @@
-#pragma once
+#ifndef CONNECTION_BOX_LOOKAHEAD_H_
+#define CONNECTION_BOX_LOOKAHEAD_H_
```
litghost, line 2: Either leave this as #pragma once, or fix the #define

vpr/test/test_map_lookahead_serdes.cpp
acomodi: This was based on the previous router_lookahead_map code, I would need to readapt the test with the code merged from the connection box lookahead map and the upstream lookahead map. It still is in the TODO list.

litghost:
This PR likely combines too many changes at once.  Things unrelated to the lookahead change should go in their own PR, and the initial change to the map lookahead should be extensions that do not change behavior significantly.  For example, the CHAN to IPIN PR should be separated from the sampling / parallelism PR.

acomodi:
@litghost All right, I'll split this into multiple ones in the following order to have an incremental merge of the changes:

- Route profiling (this should be independent from the other PRs)
- Sampling method
- Lookahead generation
- OPIN --> CHAN and CHAN --> IPIN

kmurray:
>  I'll split this into multiple ones in the following order to have an incremental merge of the changes

I agree this seems like the right approach.

acomodi:
Closing, superseded by https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1449

PR #1332, opened 2020-06-03
Adding log notes for all user specifiable command line arguments

dpbaines:
#### Description

Simply added logging for all user specifiable command line arguments, which will be output to the vpr log file. If no argument is specified it will display the default value of said option. 

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1326

#### Motivation and Context

Before, only some of the options were output, meaning that users looking back on previous runs of VPR may not know the internal values for some options that were specified. Thus this commit adds those logs. 

#### How Has This Been Tested?

I ran a flow ch_intrinsics.v with a random architecture and different command line arguments, and saw that it output significantly more options.

#### Types of changes

- [x] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1356, opened 2020-06-11
Improve valgrind vtr flow

kmurray:
#### Description

Improves valgrind integration in run_vtr_flow.pl, specifically:

- Uses a unique exit code for valgrind failures (22) which is distinct from normal exit codes
- Includes the valgrind log in the regular tool log files, which ensures they show up when a failure log is printed (e.g. on TravisCI). Previously the vaglrind output was placed in a separate valgrind.log file which was rather mysterious.
- Treats detected memory leaks as errors, since most VTR tools are now leak-free in regular usage. This should detect new leak regressions.


kmurray:
@dpbaines FYI, this PR would make the failures you are seeing in #1332 more visible and explicit.

PR #1367, opened 2020-06-16
WIP: New Lookahead sampling method

acomodi:
#### Description

This PR adds a new node sampling method, taken from https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1351.

It adds a region-based sampling, with each region of the grid containing a set of representative grid points, which on their behalf have a set of nodes to be expanded using dijkstra.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1325

#### Motivation and Context

The current sampling method in the `router_lookahead_map.cpp` code does not provide enough entries in the cost map for Series 7 devices.

The sampling method proposed in this PR has been extensively tested in SymbiFlow, and proved to have the lookahead generator to produce a correct and efficient lookahead.

The main issue is performance. In fact, the number of samples to expand is much greater than the previous method, hence the lookahead computation takes longer time, hence the addition of the per-region parallelization.
Unfortunately, even with a parallel lookahead computation, all the regression tests take longer time to compute.
However, the idea behind the lookahead is to have it computed once only per device. While this is ok for already designed devices (such as Series 7), this may be impossible for devices that are being generated using VtR, where the lookahead cannot be precomputed.

It may be possible to switch the sampling method and the lookahead computation with a command line argument, so that the faster original lookahead generation can still be used.

Basic regression tests

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/router_lookahead_map.cpp
+++ b/vpr/src/route/router_lookahead_map.cpp
@@ -35,6 +35,7 @@
 #include "vtr_time.h"
 #include "vtr_geometry.h"
 #include "router_lookahead_map.h"
+#include "router_lookahead_sampling.h"
 #include "rr_graph2.h"
 #include "rr_graph.h"
 #include "route_common.h"
@@ -47,6 +48,11 @@
 #    include "mmap_file.h"
 #    include "serdes_utils.h"
 #endif /* VTR_ENABLE_CAPNPROTO */
+
+#if defined(VPR_USE_TBB)
+#    include <tbb/parallel_for_each.h>
+#    include <tbb/mutex.h>
+#endif
 
 /* we will profile delay/congestion using this many tracks for each wire type */
 #define MAX_TRACK_OFFSET 16
@@ -193,7 +199,7 @@ struct t_reachable_wire_inf {
 
 /* used during Dijkstra expansion to store delay/congestion info lists for each relative coordinate for a given segment and channel type.
  * the list at each coordinate is later boiled down to a single representative cost entry to be stored in the final cost map */
-typedef vtr::Matrix<Expansion_Cost_Entry> t_routing_cost_map; //[0..device_ctx.grid.width()-1][0..device_ctx.grid.height()-1]
+typedef vtr::NdMatrix<Expansion_Cost_Entry, 4> t_routing_cost_map; //[0..1][0..num_segments-1][0..device_ctx.grid.width()-1][0..device_ctx.grid.height()-1]
 
 typedef std::vector<std::vector<std::map<int, t_reachable_wire_inf>>> t_src_opin_reachable_wires; //[0..device_ctx.physical_tile_types.size()-1][0..max_ptc-1][wire_seg_index]
                                                                                                   // ^                                           ^             ^
@@ -229,17 +235,15 @@ static void compute_router_src_opin_lookahead();
 static vtr::Point<int> pick_sample_tile(t_physical_tile_type_ptr tile_type, vtr::Point<int> start);
 void dijkstra_flood_to_wires(int itile, RRNodeId inode, t_src_opin_reachable_wires& src_opin_reachable_wires);
 
-/* returns index of a node from which to start routing */
-static RRNodeId get_start_node(int start_x, int start_y, int target_x, int target_y, t_rr_type rr_type, int seg_index, int track_offset);
 /* runs Dijkstra's algorithm from specified node until all nodes have been visited. Each time a pin is visited, the delay/congestion information
  * to that pin is stored is added to an entry in the routing_cost_map */
 static void run_dijkstra(RRNodeId start_node, int start_x, int start_y, t_routing_cost_map& routing_cost_map, t_dijkstra_data* data);
 /* iterates over the children of the specified node and selectively pushes them onto the priority queue */
 static void expand_dijkstra_neighbours(PQ_Entry parent_entry, vtr::vector<RRNodeId, float>& node_visited_costs, vtr::vector<RRNodeId, bool>& node_expanded, std::priority_queue<PQ_Entry>& pq);
 /* sets the lookahead cost map entries based on representative cost entries from routing_cost_map */
-static void set_lookahead_map_costs(int segment_index, e_rr_type chan_type, t_routing_cost_map& routing_cost_map);
+static void set_lookahead_map_costs(t_routing_cost_map& routing_cost_map);
 /* fills in missing lookahead map entries by copying the cost of the closest valid entry */
-static void fill_in_missing_lookahead_entries(int segment_index, e_rr_type chan_type);
+static void fill_in_missing_lookahead_entries(int num_segments);
 /* returns a cost entry in the f_wire_cost_map that is near the specified coordinates (and preferably towards (0,0)) */
 static Cost_Entry get_nearby_cost_entry(int x, int y, int segment_index, int chan_index);
 /* returns the absolute delta_x and delta_y offset required to reach to_node from from_node */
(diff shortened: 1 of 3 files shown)


vpr/src/route/router_lookahead_map.cpp
```diff
@@ -193,7 +199,7 @@ struct t_reachable_wire_inf {
 
 /* used during Dijkstra expansion to store delay/congestion info lists for each relative coordinate for a given segment and channel type.
  * the list at each coordinate is later boiled down to a single representative cost entry to be stored in the final cost map */
-typedef vtr::Matrix<Expansion_Cost_Entry> t_routing_cost_map; //[0..device_ctx.grid.width()-1][0..device_ctx.grid.height()-1]
+typedef vtr::NdMatrix<Expansion_Cost_Entry, 4> t_routing_cost_map; //[0..1][0..num_segments-1][0..device_ctx.grid.width()-1][0..device_ctx.grid.height()-1]
```
litghost, line 202: What is the first index modeling?

acomodi: CHANX/Y

litghost: Please update the comment to be clearer.

vpr/src/route/router_lookahead_map.cpp
```diff
@@ -424,135 +428,66 @@ static void compute_router_wire_lookahead(const std::vector<t_segment_inf>& segm
...
-                    }
-
-                    run_dijkstra(sample_node,
-                                 sample_x,
-                                 sample_y,
-                                 routing_cost_map,
-                                 &dijkstra_data);
-                }
+#if defined(VPR_USE_TBB)
+        all_costs_mutex.lock();
```
kmurray, line 472: Rather than manually locking/unlocking the mutex which is error prone. We should probably be using a [lock guard](https://www.threadingbuildingblocks.org/docs/doxygen/a00136.html) to ensure it is automatically unlocked in all scenarios.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
+#include "router_lookahead_sampling.h"
+
+#include <vector>
+
+#include "globals.h"
+#include "vtr_math.h"
+#include "vtr_geometry.h"
+#include "vtr_time.h"
+
```
kmurray, line 9: I think we need a file-scope comment here which outlines at a high-level how this sampling method works, as it is currently not clear.

For instance, I get that there is some kind of spatial partitioning and statistics collection about number of segment instances going on, which guides the selection of sample points. But it is not clear how these different ideas fit together, or what the motivation/intuition is behind this approach.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+// because it's just interleaving binary bits, so this
+// function interleaves with 0's so that the X and Y
+// dimensions can then be OR'ed together.
+static uint64_t interleave(uint32_t x) {
+    uint64_t i = x;
+    i = (i ^ (i << 16)) & 0x0000ffff0000ffff;
+    i = (i ^ (i << 8)) & 0x00ff00ff00ff00ff;
+    i = (i ^ (i << 4)) & 0x0f0f0f0f0f0f0f0f;
+    i = (i ^ (i << 2)) & 0x3333333333333333;
+    i = (i ^ (i << 1)) & 0x5555555555555555;
```
kmurray, line 110: This is pretty mysterious and so needs some detailed comments about how it works, and where the magic numbers come from.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+// the smallest bounding box containing a node
+static vtr::Rect<int> bounding_box_for_node(int node_ind) {
+    auto& device_ctx = g_vpr_ctx.device();
+    auto& rr_graph = device_ctx.rr_nodes;
+    int x = rr_graph.node_xlow(RRNodeId(node_ind));
+    int y = rr_graph.node_ylow(RRNodeId(node_ind));
+
+    return vtr::Rect<int>(vtr::Point<int>(x, y));
+}
```
kmurray, line 31: The naming of this isn't clear. This appears to just be the start point of the node, not its actual bounding box.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+static vtr::Rect<int> sample_window(const vtr::Rect<int>& bounding_box, int sx, int sy, int n) {
+    return vtr::Rect<int>(sample(bounding_box, sx, sy, n),
+                          sample(bounding_box, sx + 1, sy + 1, n));
+}
```
kmurray, line 36: Not clear what this is doing, needs comments.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+static std::vector<SamplePoint> choose_points(const vtr::Matrix<int>& counts,
+                                              const vtr::Rect<int>& window,
+                                              int min_count,
+                                              int max_count) {
```
kmurray, line 41: Needs a comment

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+    std::sort(points.begin(), points.end(),
+              [&](const SamplePoint& a, const SamplePoint& b) {
+                  return manhattan_distance(a.location, center) < manhattan_distance(b.location, center);
+              });
+
+    return points;
+}
+
+// histogram is a map from segment count to number of locations having that count
+static int quantile(const std::map<int, int>& histogram, float ratio) {
```
kmurray, line 65: Comment should also describe what this function does (e.g. Given the histogram and ratio this calculates....)

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+        limit -= entry.second;
+        if (limit <= 0) {
+            return entry.first;
+        }
+    }
+    return 0;
+}
+
+// select a good number of segments to find
+static std::map<int, int> count_histogram(const vtr::Rect<int>& box, const vtr::Matrix<int>& counts) {
```
kmurray, line 84: Comment likely needs more detail here. This seems to be creating a histogram, it isn't clear how that corresponds to selecting segments.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+            for (int x = 0; x < SAMPLE_GRID_SIZE; x++) {
+                vtr::Rect<int> window = sample_window(bounding_box, x, y, SAMPLE_GRID_SIZE);
+                if (window.empty()) continue;
+
+                auto histogram = count_histogram(window, segment_counts[i]);
+                SampleRegion region = {
+                    /* .segment_type = */ i,
+                    /* .grid_location = */ vtr::Point<int>(x, y),
+                    /* .points = */ choose_points(counts, window, quantile(histogram, kSamplingCountLowerQuantile), quantile(histogram, kSamplingCountUpperQuantile)),
+                    /* .order = */ 0};
```
kmurray, line 173: Probably clearer and less error prone to do this with either member-wise initialization (which is effectively what the comments are showing), or a constructor.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+                if (!region.points.empty()) {
+                    /* In order to improve caching, the list of sample points are
+                     * sorted to keep points that are nearby on the Euclidean plane also
+                     * nearby in the vector of sample points.
+                     *
+                     * This means subsequent expansions on the same thread are likely
+                     * to cover a similar set of nodes, so they are more likely to be
+                     * cached. This improves performance by about 7%, which isn't a lot,
+                     * but not a bad improvement for a few lines of code. */
```
kmurray, line 182: If we're going to put performance numbers in the comments, we likely need to give more context: what benchmarks, what FPGA architecture, what system were they run on?

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+                    vtr::Point<int> location = region.points[0].location;
+
+                    // interleave bits of X and Y for a Z-curve ordering.
+                    region.order = interleave(location.x()) | (interleave(location.y()) << 1);
```
kmurray, line 186: Again, this is somewhat mysterious.

I'd suggest wrapping the bit-wise operations inside `interleave` (or perhaps a new function like `get_region_order(location)`).

vpr/src/route/router_lookahead_map.cpp
```diff
@@ -424,135 +428,66 @@ static void compute_router_wire_lookahead(const std::vector<t_segment_inf>& segm
...
 
-            //Try an exhaustive search to find a suitable sample point
-            for (int inode = 0; inode < int(device_ctx.rr_nodes.size()); ++inode) {
-                auto rr_node = RRNodeId(inode);
-                auto rr_type = rr_graph.node_type(rr_node);
-                if (rr_type != chan_type) continue;
+    /* run Dijkstra's algorithm for each segment type & channel type combination */
+#if defined(VPR_USE_TBB)
+    tbb::mutex all_costs_mutex;
+    tbb::parallel_for_each(sample_regions, [&](const SampleRegion& region) {
```
kmurray, line 442: This appears to parallelize over the sample regions. What is a sample region? 

How many are there? Is it a fixed number, or does it scale with device size? These are important considerations for how scalable this parallelization will be.

vpr/src/route/router_lookahead_sampling.h
```diff
@@ -0,0 +1,35 @@
...
+    // location on the sample grid
+    vtr::Point<int> grid_location;
+
+    // locations to try
+    // The computation will keep expanding each of the points
+    // until a number of paths (segment -> connection box) are found.
+    std::vector<SamplePoint> points;
+
+    // used to sort the regions to improve caching
+    uint64_t order;
```
kmurray, line 30: Since the ordering is an implementation detail only used by `find_sample_regions()`, would it be possible to just store it as a local vector in `find_sample_regions()`? That would avoid exposing it to outside code through this struct.

vpr/src/route/router_lookahead_sampling.cpp
```diff
@@ -0,0 +1,229 @@
...
+    i = (i ^ (i << 16)) & 0x0000ffff0000ffff;
+    i = (i ^ (i << 8)) & 0x00ff00ff00ff00ff;
+    i = (i ^ (i << 4)) & 0x0f0f0f0f0f0f0f0f;
+    i = (i ^ (i << 2)) & 0x3333333333333333;
+    i = (i ^ (i << 1)) & 0x5555555555555555;
+    return i;
+}
+
+// for each segment type, find the nearest nodes to an equally spaced grid of points
+// within the bounding box for that segment type
```
kmurray, line 115: This seems to be the key routine which ties everything about this sampling method together. This needs higher level comments which describe the overall approach it takes (i.e. the why), and how it all fits together.

kmurray:
> The main issue is performance. In fact, the number of samples to expand is much greater than the previous method, hence the lookahead computation takes longer time, hence the addition of the per-region parallelization.

Its not clear to me from reading the code exactly how many samples this new method produces. What are the parameters which control that (both in the code and with respect to architectural parameters like the number of segment types)? (This would also be good to include in the code comments).

How much benefit is there from the additional samples? Has anyone done a study to figure out how lookahead quality varies with the number of samples? If not that might be worth investigating to see if it could be reduced.

vpr/src/route/router_lookahead_sampling.cpp
HackerFoo: Since I wrote this, here's a high level overview:

For each segment type:
1. Compute bounding box for that segment type. Within this bounding box:
2. Count the number of nodes at each location.
3. For each of NxN subsections:
   a. Compute a histogram of counts for each location.
   b. Choose a specific point within a quantile, e.g. 50-70%. This is to select a representative location while avoiding extremities.
4. Sort the points. I'm using a Z-curve for spatial locality, see the comments for details.
5. Find all the nodes that fall within each sample point of the given type.

Maybe the inline comments should be expanded, but the code in `find_sample_regions()` seems pretty readable, although reading the datatypes in the header first might help.

vpr/src/route/router_lookahead_sampling.cpp
HackerFoo: Those aren't magic numbers, they are just a bit pattern like this:
```
11110000
11001100
10101010
````

This is a method to interleave bits, so given `ABCD` and `abcd` you get `AaBbCcDd`.

https://lemire.me/blog/2018/01/08/how-fast-can-you-bit-interleave-32-bit-integers/

HackerFoo: This is used to sort points on a Morton (Z)-order curve - It interleaves the two spatial dimensions by literally interleaving their bits.

vpr/src/route/router_lookahead_sampling.cpp
HackerFoo: It might be best to consistently use low or high coordinates when sampling - the original code used annotations that were just points instead of rectangles.

vpr/src/route/router_lookahead_sampling.cpp
HackerFoo: This returns a sub-rectangle from `bounding_box` split into `n` by `n` sub-rectangles, where `sx` and `sy` select the specific sub-rectangle e.g. for the bounding box (0,0) - (2,2) and `n = 2`:

|    | sx = 0 | sx = 1 |
|--|--|--|
| sy = 0 | (0,0) - (1,1) | (1,0) - (2,1) |
| sy = 1 | (0,1) - (1,2) | (1,1) - (2,2) |

vpr/src/route/router_lookahead_map.cpp
HackerFoo: There are a (mostly) fixed number of sample regions: `SAMPLE_GRID_SIZE^2 * num_segments`.

vpr/src/route/router_lookahead_sampling.h
HackerFoo: These structs shouldn't be exposed, and I doubt it will be faster or save memory.

acomodi:
Closing, superseeded by https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1449

PR #1376, opened 2020-06-23
Speed up Router's calculation of OveruseInfo

Bill-hbrhbr:
#### Motivation and Context
Trying to speed up the VPR router by detecting its various performance bottlenecks and remove them by optimizations. This PR should not change any of the VPR router's functionalities.

This PR is still undergoing changes. I will update more commits if I manage to optimize other areas of the code.

#### What Has Already Been Optimized?
1. Optimized overused info calculation. Noticeable speed up when running very large titan benchmarks, where the number of nets is large and many router iterations are performed.

- Tested on the LU8+Flagship and Stereovision+Tian benchmarks. No significant speedup.
- Tested on the Bitcoin+Titan benchmark. Appreciable speedup.

#### Types of changes

- [X] Optimization (change which speeds up code)

#### Checklist:

### DIFF
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -192,7 +192,7 @@ struct more_sinks_than {
 
 static size_t calculate_wirelength_available();
 static WirelengthInfo calculate_wirelength_info(size_t available_wirelength);
-static OveruseInfo calculate_overuse_info(std::vector<ClusterNetId>& rerouted_nets);
+static OveruseInfo calculate_overuse_info(const std::vector<ClusterNetId>& rerouted_nets);
 
 static void print_route_status_header();
 static void print_route_status(int itry,
@@ -1608,9 +1608,8 @@ static bool early_exit_heuristic(const t_router_opts& router_opts, const Wirelen
     return false;
 }
 
-static OveruseInfo calculate_overuse_info(std::vector<ClusterNetId>& rerouted_nets) {
+static OveruseInfo calculate_overuse_info(const std::vector<ClusterNetId>& rerouted_nets) {
     auto& device_ctx = g_vpr_ctx.device();
-    auto& cluster_ctx = g_vpr_ctx.clustering();
     auto& route_ctx = g_vpr_ctx.routing();
 
     std::unordered_set<int> checked_nodes;
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -192,7 +192,7 @@ struct more_sinks_than {
 
 static size_t calculate_wirelength_available();
 static WirelengthInfo calculate_wirelength_info(size_t available_wirelength);
-static OveruseInfo calculate_overuse_info();
+static OveruseInfo calculate_overuse_info(std::vector<ClusterNetId>& rerouted_nets);
 
 static void print_route_status_header();
 static void print_route_status(int itry,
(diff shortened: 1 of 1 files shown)


vaughnbetz:
Thanks Bill. The code looks good.

1. Please attach QoR data on the vtr and Titan benchmarks (see the developer guide in the documentation for how to get that, and ask Mohamed for anything that's unclear.)

2. The same technique might be possible (with extensions) in
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/0a6ce4d388ecee8aa6564307df71c3a5c26107ec/vpr/src/route/route_common.cpp#L439 (pathfinder_update_cost)
Right now that routine goes through all routing resource nodes.  If only a few nets (or even better, connections) are rerouted, you could instead go through only the nets that were rerouted, but you'd need to update the costs for all nodes in either the old routing of those nets or the new routings.  If more than a few nets are rerouted then just going through all rr_nodes is probably faster though.

vaughnbetz:
I was thinking more about #2 above and I think it is worth investigating.  It would require:
- remove pres_cost from rr_node_route_inf.  Instead compute it on the fly using the same code as in pathfinder_update_cost
- Now pres_cost will always be up to date based on occupancy.
- Change pathfinder_update_cost to only go through the congested nets and update acc_cost

Nowadays bringing more data in from storage is usually more expensive than doing a small amount of computation, so shrinking rr_node_route_inf with this on-the-fly computation may itself be a good idea (or at least not a very bad idea).  And that refactoring allows pathfinder_update_cost to become incremental in the same way as you made the overused_info computation.

This should all be done in any issue / PR though -- once you have the QoR data for this PR you can merge and close it.

kmurray:
> This should all be done in any issue / PR though -- once you have the QoR data for this PR you can merge and close it.

I agree, getting QoR data for each change individually, and merging them as separate PRs seems like the right approach.

vpr/src/route/route_timing.cpp
```diff
@@ -1608,7 +1608,7 @@ static bool early_exit_heuristic(const t_router_opts& router_opts, const Wirelen
     return false;
 }
 
-static OveruseInfo calculate_overuse_info() {
+static OveruseInfo calculate_overuse_info(std::vector<ClusterNetId>& rerouted_nets) {
```
kmurray, line 1611: Minor, but we should probably be passing this as `const std::vector<ClusterNetId>&` since it isn't modified in this function

Bill-hbrhbr:
> shrinking rr_node_route_inf with this on-the-fly computation may itself be a good idea

@vaughnbetz After some investigations, I found out that **pres_cost** is referenced in many other places, so it might not be possible to eliminate this field entirely (requires a lot of refactoring). But it does seem to me that there is an abundance of updates to this field, and some of them are probably unnecessary. What I think I'll try to do is to pick out all the seemingly unnecessary ones and comment them out to see how that will affect the results.

Bill-hbrhbr:
VTR benchmarks: https://drive.google.com/file/d/1PVIgyMGYtgoIytFgNkA99aUwlZncWt6P/view?usp=sharing

Bill-hbrhbr:
Titan benchmarks (still running): https://drive.google.com/file/d/1rimt3V6yAHtjyqKRxFJXzGDq6dUG97WI/view?usp=sharing

vaughnbetz:
QoR data is looking good so far; looks like a 3% speedup or so.
For pres_cost: yes, you'd need to refactor.  I'd write an inline function (well inline is a hint, but may help) in a header file to create pres_cost on the fly.  Then every use of pres_cost would become a call to it, and every update of pres_cost should be deleted.

Issue #1379, opened 2020-06-25
Net alias not properly considered when creating set_{min,max}_delay constrainst


acomodi:
Issue from https://github.com/SymbiFlow/vtr-verilog-to-routing/issues/524.
Alias are not considered when processing the `get_clocks` call when doing min/max delay constrainsts in the SDC.

#### Expected Behaviour

In the input netlist where you have theses aliases :

```
.names usb_I.uc_clk clk_wb
1 1
.names usb_I.tx_pkt_I.crc_16_I.clk clk_usb
1 1
```

Then the constraints : 

```
set_max_delay 40 -from [get_clocks {clk_wb}] -to [get_clocks {clk_usb}]
```

and

```
set_max_delay 40 -from [get_clocks {usb_I.tx_pkt_I.crc_16_I.clk}] -to [get_clocks {usb_I.uc_clk}]
```

should be equivalent. But in reality, only the latter works because aliases are not considered.

#### Current Behaviour

The SDC parsing fails with `set_max_path must specify at least one -from or -to clock` because it doesn't find any clocks.

#### Possible Solution

Fix https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/timing/read_sdc.cpp#L896 to also process net aliases like it was done for `create_clock`.

#### Steps to Reproduce

Try to setup any max/min delay constrainst using the non-primary clock name from the EBLIF.

#### Context

I'd like to setup constrainst with a stable clock net name and not some random name that was randomly picked among all the possible aliases by the synthesis tool.

#### Your Environment
* VTR revision used: 8980e46218542888fac879961b13aa7b0fba8432
* Operating System and version: Linux
* Compiler version: gcc 9.2.0

kmurray:
Thanks for the report @acomodi this is something that should be fixed.

PR #1380, opened 2020-06-25
clock aliases: add check through all clock aliases when getting clocks

acomodi:

#### Description

This PR fixes the clock net aliasing for SDC commands other than `create_clock`

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1379

#### Motivation and Context

Clock aliases need to be used for other SDC commands and not only create_clock.

I have added a regression test that uses set_max_delay sdc constraint

#### Types of changes

#### Checklist:


vpr/src/timing/read_sdc.cpp
```diff
@@ -911,18 +912,33 @@ class SdcParseCallback : public sdcparse::Callback {
 
             bool found = false;
             for (tatum::DomainId domain : tc_.clock_domains()) {
-                const std::string& clock_name = tc_.clock_domain_name(domain);
-                if (std::regex_match(clock_name, clock_regex)) {
-                    found = true;
+                const auto& clock_name = tc_.clock_domain_name(domain);
+
+                if (tc_.is_virtual_clock(domain)) {
```
vaughnbetz, line 917: Need a comment here I think saying what the overall idea is.  

If virtual_clock must match the clock_regex.  Otherwise, look for all aliases of the clock_name (nets that passed the clock signal through?) and see if any of them matches the clock_regex.  A comment saying why this is the right logic (and likely an example) would be helpful.

kmurray:
Looks good to me too.

Issue #1378, opened 2020-06-24
Placer crashing


litghost:
#### Expected Behaviour

Placer should not crash when running.

#### Current Behaviour

```
/tmpfs/src/github/vtr-verilog-to-routing/vpr/src/place/timing_place.h:196 PlacerTimingCosts: Assertion 'num_connections == 0 || num_level_before_leaves < num_connections' failed (Level before should have fewer nodes than connections (to ensure using the smallest binary tree)).
```

#### Possible Solution

Still investigating, recent changes in the placer (near https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/7f39b64da4a3c6f301d5bd4bb3ab92f45f9bfb72) appear to be the changes that caused the issue.

#### Steps to Reproduce

Working on reduced tested case, TBD.

1. 
2. 

#### Context

At a minimum the symbiflow-arch-defs ice40 arch started generating this exception when updating to the lastest VTR master.

vaughnbetz:
Bill, can you take a look?  It appears to be in the new code that adds up timing costs in a hierarchical way to minimize round off.  Keith is working on a small test case, but says making a circuit with a single input connected to a single output appears to cause the problem.  So maybe there is a degenerate case where very small circuits break this code (maybe an out of bounds error for such a case?).

vaughnbetz:
So this may not be debuggable without more info, but if you inspect the code perhaps you'll see it.
Also may indicate a missing test case (extremely small, one net circuit).

litghost:
When I tried to replicate this outside of CI, the issue did not appear.  I'm confused about why, but for now I'm going to close this issue.  I will re-open once I get a replication :/

litghost:
I've been able to replicate this issue!  GCC 5 and GCC 6 generate the exception, and clang-7 does not!  This hints at possibly some kind of undefined behavior being present?

litghost:
Ya, I believe the assertion fires when `num_connections = 1`.  Let me add a vtr_reg_basic wire test, and see if it I can get a replication.

Bill-hbrhbr:
It seems to me that this for loop in timing_place.cpp:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/place/timing_place.h#L165-L180 

is skipped entirely when there is only one unignored net within the netlist. This problem disappears when there are two or more unignored nets.

[...]

PR #1381, opened 2020-06-25
Add basic wire.v test.

litghost:
#### Description

This demonstrates #1378

#### Related Issue

#1378

#### Motivation and Context

This adds the most basic test, a wire between an in port and an out port.

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1382, opened 2020-06-26
PlacerTimingCosts # of heap (BST) nodes==1 (root only)

Bill-hbrhbr:
#### Related Issue
#1378

#### Types of changes

#### Checklist:


litghost:
You should add test from #1381 so this is caught in CI.

Bill-hbrhbr:
> You should add the test from #1381 so this is caught in CI.

Tests are added. I'll merge if you are ok with this solution.

Turns out my original analysis was off. I think Kevin's idea is that the root will never be assigned a sink node (even if there is only one connection, the heap will have two nodes). The function that checks # of nodes on a tree-level fails to account for this fact, and only a corner case detection is needed.

Issue #1387, opened 2020-06-30
Calculation of RR node congestion cost


Bill-hbrhbr:
#### Description
The RR node congestion cost should be calculated as a product of individual cost components:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/route_common.cpp#L730-#L732

however, in draw.cpp, the cost is calculated as a sum:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/draw/draw.cpp#L1202-#L1204

Not sure if this is a bug or an intended modification to make the graphics look nicer. We should add comments if it is the latter case.

@kmurray @vaughnbetz

vaughnbetz:
The route_common code is for sure correct.  I suspect the draw code is just a bug.  Just to double check: Kevin, was this intentional to make drawing look nicer?

kmurray:
>  I suspect the draw code is just a bug.

Yep that looks like a bug. To fix it, I'd suggest using `get_single_rr_cong_cost()` in draw.cpp so we have only a single definition of how costs are calculated.

Bill-hbrhbr:
Ok, I will update this in a future pull request.

vaughnbetz:
I'm pretty sure @Bill-hbrhbr fixed this.

PR #1396, opened 2020-07-03
Optimize pres_cost storage/calculation

Bill-hbrhbr:
#### Motivation and Context
**pres_cost**, aka the present congestion cost, is one of the factors used by VPR routing to measure the congestion of a node on the rr_graph. Originally, it exists as a data field stored within a hot data structure that is being referenced frequently during the routing process. The optimization removes this data field and calculates the cost on the fly whenever it is actually needed. This tradeoff is between 
       a more cache-friendly data structure + removing unnecessary updates
and
       slightly more computations.

#### Related Issue
#1387

The only tests that had failed during the refactoring process were the analysis only and breadth-first routing regression tests (vtr_reg_strong). Right now these issues have been resolved.

I will do a QoR testing soon.

#### Types of changes

- [X] Optimization (change which speeds up code)

#### Checklist:


vpr/src/draw/draw.cpp
```diff
@@ -1194,27 +1195,27 @@ static void draw_routing_costs(ezgl::renderer* g) {
     float min_cost = std::numeric_limits<float>::infinity();
     float max_cost = -min_cost;
     std::vector<float> rr_node_costs(device_ctx.rr_nodes.size(), 0.);
+
+    /* For display purpose only, not an accurate value of the present congestion factor */
+    float approx_pres_fac = 1.;
```
vaughnbetz, line 1200: Could you get the accurate value of pres_fac?  Using 1 here could get confusing.  You could put a copy of pres_fac in the draw_state if necessary and update it when or before calling update_screen; not sure if there's a better option.

If you use an approximate value, a large value (say 10) would be better than 1 to ensure the present congestion significantly impacts the display.  But I really think it would be best to get the real value.

vpr/src/route/route_common.cpp
```diff
@@ -93,11 +93,10 @@ static vtr::vector<ClusterNetId, std::vector<int>> load_net_rr_terminals(const t
 static vtr::vector<ClusterBlockId, std::vector<int>> load_rr_clb_sources(const t_rr_node_indices& L_rr_node_indices);
 
 static t_clb_opins_used alloc_and_load_clb_opins_used_locally();
-static void adjust_one_rr_occ_and_apcost(int inode, int add_or_sub, float pres_fac, float acc_fac);
+static void adjust_one_rr_occ_and_apcost(int inode, int add_or_sub, float acc_fac);
```
vaughnbetz, line 96: Should rename to adjust_one_rr_occ_and_acc_cost (no longer a present congestion cost update, so apcost doesn't make sense anymore).

vpr/src/route/route_common.cpp
```diff
@@ -713,24 +690,49 @@ float get_rr_cong_cost(int inode) {
...
     }
     return (cost);
 }
 
-/* Returns the congestion cost of using this rr_node, *ignoring*
- * non-configurable edges */
-static float get_single_rr_cong_cost(int inode) {
+/* Returns the congestion cost of using this rr_node,
+ * *ignoring* non-configurable edges */
+float get_single_rr_cong_cost(int inode, float pres_fac) {
```
vaughnbetz, line 702: Not necessary at this point, but you may want to explore putting this function in a header file with an inline in front of it to see if it affects speed (if you're calling it across file boundaries).
With interprocedural optimization on (the default in our makefile) it may not make any difference though.

vpr/src/route/route_common.cpp
```diff
@@ -713,24 +690,49 @@ float get_rr_cong_cost(int inode) {
...
+/* Returns the base cost of using this rr_node */
+float get_single_rr_cong_base_cost(int inode) {
     auto& device_ctx = g_vpr_ctx.device();
+    auto cost_index = device_ctx.rr_nodes[inode].cost_index();
+
+    return device_ctx.rr_indexed_data[cost_index].base_cost;
+}
+
+/* Returns the accumulated congestion cost of using this rr_node */
+float get_single_rr_cong_acc_cost(int inode) {
```
vaughnbetz, line 717: Explore if putting inline on these helps, or pasting the function body into the calling function.  get_single_rr_cong_cost is going to be a very hot function, and it calls these other short functions.  Good coding style as long as they are inlined, but worth checking speed (or you could look at the assembly or run the optimized code through the debugger to see if they are inlined).

vpr/src/route/route_common.cpp
```diff
@@ -959,14 +961,15 @@ void reset_rr_node_route_structs() {
...
-        route_ctx.rr_node_route_inf[inode].prev_node = NO_PREVIOUS;
-        route_ctx.rr_node_route_inf[inode].prev_edge = RREdgeId::INVALID();
-        route_ctx.rr_node_route_inf[inode].pres_cost = 1.0;
-        route_ctx.rr_node_route_inf[inode].acc_cost = 1.0;
-        route_ctx.rr_node_route_inf[inode].path_cost = std::numeric_limits<float>::infinity();
-        route_ctx.rr_node_route_inf[inode].backward_path_cost = std::numeric_limits<float>::infinity();
-        route_ctx.rr_node_route_inf[inode].target_flag = 0;
-        route_ctx.rr_node_route_inf[inode].set_occ(0);
+        auto& node_inf = route_ctx.rr_node_route_inf[inode];
+
```
vaughnbetz, line 965: Nice cleanup.

vpr/src/route/route_common.cpp
```diff
@@ -698,11 +675,11 @@ void reset_path_costs(const std::vector<int>& visited_rr_nodes) {
 }
 
 /* Returns the *congestion* cost of using this rr_node. */
```
vaughnbetz, line 677: Update this comment:
/* Returns the congestion cost of using this rr-node plus that of any non-configurably 
 * connected rr_nodes that *must* be used when it is used.
 */

vpr/src/route/route_common.cpp
vaughnbetz, line 702: This is a very hot function that is calling 3 other short functions.  Check to be sure it is inlined (can look at asm, or use the debugger on the optimized code).  Could put inline on the functions to give a hint. 
If they aren't inlined, I would also measure the speed of manually inlining the function bodies (simple) of the 3 short functions called (or at least the very simple base and acc_cost ones).  With code this hot and functions this short, the function call overhead can be significant.

vpr/src/draw/draw.cpp
Bill-hbrhbr: a subsystem scope variable **pres_fac** added in draw_global.h and draw_global.cpp

Bill-hbrhbr: Actually nvm. I'll put them in draw states like you said

Bill-hbrhbr:
Titan benchmarks:
https://drive.google.com/file/d/1GkTBtJoDVdeciaRyjK1bkKkFrGh2PCLy/view?usp=sharing

Bill-hbrhbr:
VTR benchmarks:
https://drive.google.com/file/d/1MZSSiVT-WRRqqkRkVdILlDv0SR71tIZl/view?usp=sharing

Bill-hbrhbr:
Updated Titan benchmarks after adding incremental acc_cost calculation:
https://drive.google.com/file/d/1sC6LKS2eucCxTKCxbaHe739Q0nKq-ONY/view?usp=sharing

Bill-hbrhbr:
Incremental acc_cost update (Commit 068d8dd) comparison:
VTR benchmarks: https://drive.google.com/file/d/1ReCWqE0gmAAvVdnqLs8-Sox2Rj25wbbo/view?usp=sharing
Titan benchmarks: https://drive.google.com/file/d/1aJ_R3mO_dA37fR42q37WTYDZtKDieq7y/view?usp=sharing

Runs are now done simultaneously to reduce the bias factors on the Wintermute system load.

vaughnbetz:
Bill, lets commit the pres_cost storage removal, as that was a net benefit.  Can you update to have a PR with only that change?

Bill-hbrhbr:
> Bill, lets commit the pres_cost storage removal, as that was a net benefit. Can you update to have a PR with only that change?
@vaughnbetz Yes, I already did that, and Kevin has reviewed that PR for me. See: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1438 for details

vaughnbetz:
Thanks.  In that case I assume we can close this one?

Bill-hbrhbr:
Yeah. I think we can close this one. I'll keep the link on my wiki for future references.

PR #1398, opened 2020-07-04
Added support for multiple driver pins

djns99:
#### Description

This adds multiple driver pins on a net to enable support for multiple assignments to a single wire.

#### Motivation and Context

This is a precursor step to implementing _inout_s as they require at least two drivers to be useful.

A new test based on the example from the Verilog 2005 standard has been added to the odin_ii suite.

#### Types of changes

#### Checklist:


ODIN_II/SRC/multipliers.cpp
```diff
@@ -491,18 +491,21 @@ void define_mult_function(nnode_t* node, FILE* out) {
 
     for (i = 0; i < node->num_input_pins; i++) {
         if (i < node->input_port_sizes[flip ? 1 : 0]) {
-            npin_t* driver_pin = flip
-                                     ? node->input_pins[i + node->input_port_sizes[0]]->net->driver_pin
-                                     : node->input_pins[i]->net->driver_pin;
+            oassert((flip ? node->input_pins[i + node->input_port_sizes[0]]->net->num_driver_pins : node->input_pins[i]->net->num_driver_pins) == 1);
```
jeanlego, line 494: store the index in a variable, otherwise, the ternary is unreadable ; especially if done twice

ODIN_II/SRC/multipliers.cpp
```diff
@@ -491,18 +491,21 @@ void define_mult_function(nnode_t* node, FILE* out) {
...
+            npin_t* driver_pin
+                = flip
+                      ? node->input_pins[i + node->input_port_sizes[0]]->net->driver_pins[0]
```
jeanlego, line 497: use the variable mentionned above

ODIN_II/SRC/multipliers.cpp
```diff
@@ -491,18 +491,21 @@ void define_mult_function(nnode_t* node, FILE* out) {
...
+                      : node->input_pins[i]->net->driver_pins[0];
 
             if (!driver_pin->name)
                 j = odin_sprintf(buffer, " %s[%ld]=%s", hard_multipliers->inputs->next->name, i, driver_pin->node->name);
             else
                 j = odin_sprintf(buffer, " %s[%ld]=%s", hard_multipliers->inputs->next->name, i, driver_pin->name);
         } else {
+            oassert((flip ? node->input_pins[i + node->input_port_sizes[1]]->net->num_driver_pins : node->input_pins[i]->net->num_driver_pins) == 1);
```
jeanlego, line 505: idem

ODIN_II/SRC/netlist_check.cpp
```diff
@@ -685,35 +686,47 @@ void levelize_backwards_clean_checking_for_liveness(short ast_based, netlist_t*
...
  *-------------------------------------------------------------------------------------------*/
 nnode_t* find_node_at_top_of_combo_loop(nnode_t* start_node) {
-    int i;
-    int idx_missed;
-    nnode_t* next_node = start_node;
-    int* fanouts_visited;
-    short all_visited;
+
+    int stack_size = 1;
+    int stack_cap = 1;
```
jeanlego, line 691: stack_size alone seems to be sufficient here.

ODIN_II/SRC/netlist_check.cpp
```diff
@@ -685,35 +686,47 @@ void levelize_backwards_clean_checking_for_liveness(short ast_based, netlist_t*
...
+            for(int i = 0; i < next_node->input_pins[idx_missed]->net->num_driver_pins; i++) {
+                if (next_node->input_pins[idx_missed]->net->driver_pins[i]->node->backward_level < next_node->backward_level) {
+                    /* IF - the next node has a lower backward level than this node suggests that it is
+                     * closer to primary outputs and not in the combo loop */
+                    vtr::free(stack);
+                    return next_node;
+                }
 
-            next_node = next_node->input_pins[idx_missed]->net->driver_pin->node;
+                if(stack_size >= stack_cap-1) {
```
jeanlego, line 721: you increment both at the same rate, just use stack size

ODIN_II/SRC/netlist_create_from_ast.cpp
```diff
@@ -3128,17 +3129,27 @@ void terminate_registered_assignment(ast_node_t* always_node, signal_list_t* ass
         }
     }
 
+#warning "Oops this be broke come fix"
```
jeanlego, line 3132: ?

ODIN_II/SRC/netlist_create_from_ast.cpp
```diff
@@ -3128,17 +3129,27 @@ void terminate_registered_assignment(ast_node_t* always_node, signal_list_t* ass
...
+            error_message( NETLIST, pin->node->line_number, pin->node->file_number, "Multiple drivers not supported for assignment to %s", pin->name);
+        }
+        if (pin->net->num_driver_pins) {
+            ref_string = pin->net->driver_pins[0]->node->name;
 
             for (j = i - 1; j >= 0; j--) {
-                if (list_dependence_pin[j] && list_dependence_pin[j]->net->driver_pin && list_dependence_type[j] && list_dependence_type[j] == BLOCKING_STATEMENT && strcmp(ref_string, assignment->pins[j]->node->name) == 0) {
+                if(list_dependence_pin[j] && list_dependence_pin[j]->net->num_driver_pins > 1)
+                {
+                    // TODO Come make this warning actually correct when I know what is happening
```
jeanlego, line 3147: do you have an example bit of Verilog code that can make this reachable? I'm just wondering if you could not just attach all the drivers anyways

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -1175,6 +1181,17 @@ int get_clock_ratio(nnode_t* node) {
     return node->ratio;
 }
 
+/**
+ * Checks if the pin is one of the drivers of the provided net
+ */
+static bool pin_is_driver( npin_t* pin, nnet_t* net )
```
jeanlego, line 1187: you can just check that the pin->net == net without looping

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -1331,8 +1374,9 @@ nnode_t** get_children_of_nodepin(nnode_t* node, int* num_children, int output_p
  */
 static void initialize_pin(npin_t* pin) {
     // Initialise the driver pin if this pin is not the driver.
```
jeanlego, line 1376: isn't it just sufficient to always just assign it no matter what? if you are going to check, just set and forget

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -1649,20 +1693,23 @@ static void compute_add_node(nnode_t* node, int cycle, int type) {
...
     int* b = (int*)vtr::malloc(sizeof(int) * node->input_port_sizes[1]);
     int* c = (int*)vtr::malloc(sizeof(int) * node->input_port_sizes[2]);
 
-    num = node->input_port_sizes[0] + node->input_port_sizes[1];
+    int num = node->input_port_sizes[0] + node->input_port_sizes[1];
     //if cin connect to unconn(PAD_NODE), a[0] connect to ground(GND_NODE) and b[0] connect to ground, flag = 0 the initial adder for addition
     //if cin connect to unconn(PAD_NODE), a[0] connect to ground(GND_NODE) and b[0] connect to vcc, flag = 1 the initial adder for subtraction
-    if (node->input_pins[num]->net->driver_pin->node->type == PAD_NODE) {
-        if (node->input_pins[0]->net->driver_pin->node->type == GND_NODE && node->input_pins[node->input_port_sizes[0]]->net->driver_pin->node->type == GND_NODE)
+#warning "Oops this be broke come fix"
```
jeanlego, line 1706: I believe you could resolve most of those to unknown automatically and throw a warning about wrongly formed circuit based on the blif standard (can't have a wire driven by more than one thing)... and given this is a blif simulator

ODIN_II/exec_wrapper.sh
```diff
@@ -175,12 +175,13 @@ function display() {
 	esac
 
 	# check for uncaught errors
-
 	ERROR_CATCH="$(cat "${LOG_FILE}" | grep 'Odin exited with code: ' | awk '{print $5}' | grep -E '^\-?[0-9]+$')"
 	[ "_${ERROR_CATCH}" != "_" ] && EXIT_CODE="${ERROR_CATCH}"
 
 	EXIT_ERROR_TYPE=$( print_exit_type "${EXIT_CODE}" )
 
+	MISMATCH_CHECK="$(cat "${LOG_FILE}" | grep "Error::OUTPUT_BLIF Vector files differ.")"
```
jeanlego, line 183: can you put that as a separate PR? this might some effect on the json and blow up the number of file changes otherwise. I'd rather see this self contained

ODIN_II/verify_odin.sh
```diff
@@ -47,14 +47,14 @@ global_failure="test_failures.log"
 ##############################################
```
jeanlego, line 47: this should be in a separate PR

ODIN_II/SRC/adders.cpp
```diff
@@ -1121,27 +1122,25 @@ void free_op_nodes(nnode_t* node) {
...
-                flag = 1;
-                break;
-            }
-            if (j == next_node->num_input_pins - 1) {
-                flag = -1;
-                break;
+    for (int i = 0; i < node->num_input_pins; i++) {
+        for (int j = 0; j < node->input_pins[i]->net->num_driver_pins; j++) {
+            bool found = false;
+            id = node->input_pins[i]->net->driver_pins[i]->unique_id;
```
jeanlego, line 1129: can you declare id here rather than outside the loop (while you are changing it)

ODIN_II/SRC/multipliers.cpp
```diff
@@ -491,18 +491,18 @@ void define_mult_function(nnode_t* node, FILE* out) {
 
     for (i = 0; i < node->num_input_pins; i++) {
         if (i < node->input_port_sizes[flip ? 1 : 0]) {
-            npin_t* driver_pin = flip
-                                     ? node->input_pins[i + node->input_port_sizes[0]]->net->driver_pin
-                                     : node->input_pins[i]->net->driver_pin;
+            nnet_t* net = flip ? node->input_pins[i + node->input_port_sizes[0]]->net : node->input_pins[i]->net;
```
jeanlego, line 494: can you just grab the index rather

ODIN_II/SRC/multipliers.cpp
```diff
@@ -491,18 +491,18 @@ void define_mult_function(nnode_t* node, FILE* out) {
...
 
             if (!driver_pin->name)
                 j = odin_sprintf(buffer, " %s[%ld]=%s", hard_multipliers->inputs->next->name, i, driver_pin->node->name);
             else
                 j = odin_sprintf(buffer, " %s[%ld]=%s", hard_multipliers->inputs->next->name, i, driver_pin->name);
         } else {
-            npin_t* driver_pin = flip
-                                     ? node->input_pins[i - node->input_port_sizes[1]]->net->driver_pin
-                                     : node->input_pins[i]->net->driver_pin;
+            nnet_t* net = flip ? node->input_pins[i + node->input_port_sizes[1]]->net : node->input_pins[i]->net;
```
jeanlego, line 503: idem

ODIN_II/SRC/simulate_blif.cpp
djns99: Im not sure I understand

ODIN_II/SRC/netlist_utils.cpp
```diff
@@ -277,7 +277,8 @@ nnet_t* allocate_nnet() {
     nnet_t* new_net = (nnet_t*)my_malloc_struct(sizeof(nnet_t));
 
     new_net->name = NULL;
-    new_net->driver_pin = NULL;
+    new_net->driver_pins = NULL;
```
jeanlego, line 280: make sure to adjust the free_nnet function

ODIN_II/SRC/simulate_blif.cpp
djns99: I thought the whole point of this was to support having a driver driven by more than one thing

ODIN_II/SRC/netlist_utils.cpp
```diff
@@ -443,10 +446,13 @@ void combine_nets(nnet_t* output_net, nnet_t* input_net, netlist_t* netlist) {
  *-------------------------------------------------------------------------------------------*/
 void join_nets(nnet_t* join_to_net, nnet_t* other_net) {
     if (join_to_net == other_net) {
-        if ((join_to_net->driver_pin) && (join_to_net->driver_pin->node != NULL) && join_to_net->driver_pin->node->related_ast_node != NULL)
-            error_message(NETLIST, join_to_net->driver_pin->node->related_ast_node->line_number, join_to_net->driver_pin->node->related_ast_node->file_number, "%s", "This is a combinational loop\n");
+#warning "Oops this be broke come fix"
+        oassert(join_to_net->num_driver_pins <= 1);
+        if ((join_to_net->num_driver_pins) && (join_to_net->driver_pins[0]->node != NULL) && join_to_net->driver_pins[0]->node->related_ast_node != NULL)
```
jeanlego, line 451: could you for loop to check?

ODIN_II/SRC/simulate_blif.cpp
jeanlego: if pin is not driver, make it driver == set pin as driver

ODIN_II/SRC/simulate_blif.cpp
jeanlego: agreed... this is the blif simulator, I'm stating this should NOT error, just that the user is threading in invalid land and theres no specification for that so we are not gonna make up one

ODIN_II/SRC/adders.cpp
djns99: Happily

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -1189,25 +1202,27 @@ nnode_t** get_children_of(nnode_t* node, int* num_children) {
         nnet_t* net = pin->net;
         if (net) {
             /*
-             *  Detects a net that may be being driven by two
-             *  or more pins or has an incorrect driver pin assignment.
+             *  Detects a net that has an incorrect driver pin assignment.
```
jeanlego, line 1205: in this perticular case you do want to forloop otherwise you are doing the check twice, this one here is to sanitize the back and forward pointers are both aligned

ODIN_II/exec_wrapper.sh
djns99: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1399

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -1175,6 +1181,13 @@ int get_clock_ratio(nnode_t* node) {
     return node->ratio;
 }
 
+/**
+ * Checks if the pin is one of the drivers of the provided net
+ */
+static bool pin_is_driver(npin_t* pin, nnet_t* net) {
```
jeanlego, line 1187: appologies, after careful review, the two places you use this, the loop actually serves to verify that the foward and back pointers are alligned

ODIN_II/SRC/simulate_blif.cpp
djns99: And if there are multiple drivers?

ODIN_II/SRC/simulate_blif.cpp
jeanlego: also, printing unknown allows loosening the requirements on the vectors. if the expected input was unknown but the result is known, we only warn, we don't error unlike any other mismatch

ODIN_II/SRC/netlist_utils.cpp
djns99: Good catch

ODIN_II/SRC/simulate_blif.cpp
jeanlego: this check is there to make sure its an input_pin to a node, if it is but its not initialized go initialize the parent (output_pin of a node) which will not be initialized since its not an input pin.
with that in mind, the best way to acheive that would be to check the pin dire

ODIN_II/SRC/netlist_utils.cpp
djns99: Yeah no idea why I didnt in the first place. Also those error messages look pretty like they could do with a cleanup? Or does `This is a combinational loop with more info` have some special meaning

ODIN_II/SRC/simulate_blif.cpp
djns99: Ok np

ODIN_II/SRC/netlist_utils.cpp
djns99: Also should I be looping over the fanout pins while I am at it?

ODIN_II/SRC/simulate_blif.cpp
jeanlego: > if pin is not driver, make it driver == set pin as driver

misread there, appologies

ODIN_II/SRC/netlist_create_from_ast.cpp
djns99: Im not sure how do go about doing that. Hence the obnoxious warning message. I will look into it some more.

ODIN_II/SRC/simulate_blif.cpp
jeanlego: good as it is :) sorry I thought this was in another bit of code

ODIN_II/SRC/simulate_blif.cpp
djns99: Im still quite confused. So just print a warning rather than assert and just assume 0?
Also, I haven't had a chance to test this thoroughly, but it seems to me that this would imply we cant use a variable that has been assigned to multiple times in an addition? - in which case do I need to rewrite this to support multiple drivers

jeanlego: unless there is a better alternative?

jeanlego: warn and assign -1 (unkown value) to the input_pin, this function evaluates the result based on input, if there is more than one, I can't pick, so just collapse into an unknown. technically everything should do that

djns99: In theory you should be able to pick yes assuming one of the drivers is Z? I think there is probably some work here for managing that correctly. Currently I just fixed all the syntax errors to make it compile

jeanlego: well yes, if the driver is disconnected it should, which is a problem, the simulator only has 0 1 and unknown, there is no disconnected value... we talked about this, but this will be a problem here.

jeanlego: The only way to know wether its disconnected or not is to propagate the disconnected value. I think for know unknown is a temporary fix, but the simulator needs to change to quad value internally

ODIN_II/SRC/netlist_utils.cpp
```diff
@@ -443,14 +449,20 @@ void combine_nets(nnet_t* output_net, nnet_t* input_net, netlist_t* netlist) {
...
     if (join_to_net == other_net) {
-        if ((join_to_net->driver_pin) && (join_to_net->driver_pin->node != NULL) && join_to_net->driver_pin->node->related_ast_node != NULL)
-            error_message(NETLIST, join_to_net->driver_pin->node->related_ast_node->line_number, join_to_net->driver_pin->node->related_ast_node->file_number, "%s", "This is a combinational loop\n");
-        else
-            error_message(NETLIST, -1, -1, "%s", "This is a combinational loop\n");
-        if ((join_to_net->fanout_pins[0] != NULL) && (join_to_net->fanout_pins[0]->node != NULL) && join_to_net->fanout_pins[0]->node->related_ast_node != NULL)
-            error_message(NETLIST, join_to_net->fanout_pins[0]->node->related_ast_node->line_number, join_to_net->fanout_pins[0]->node->related_ast_node->file_number, "%s", "This is a combinational loop with more info\n");
-        else
-            error_message(NETLIST, -1, -1, "%s", "Same error - This is a combinational loop\n");
+        oassert(join_to_net->num_driver_pins <= 1);
```
jeanlego, line 452: I don't believe this is needed, It will error out lower, can you turn all the errors bellow into warnings and finally error so we can trace the pins

djns99: Will do. Just double check looping over fanouts is actually correct

djns99: Also should I print the driver pin names?
If so would it be `driver_pin[i]->name` or `driver_pin[i]->node->name`?

jeanlego: driver_pin[i]->name i believe

ODIN_II/SRC/multipliers.cpp
```diff
@@ -491,18 +491,20 @@ void define_mult_function(nnode_t* node, FILE* out) {
...
             if (!driver_pin->name)
                 j = odin_sprintf(buffer, " %s[%ld]=%s", hard_multipliers->inputs->next->name, i, driver_pin->node->name);
             else
                 j = odin_sprintf(buffer, " %s[%ld]=%s", hard_multipliers->inputs->next->name, i, driver_pin->name);
         } else {
-            npin_t* driver_pin = flip
-                                     ? node->input_pins[i - node->input_port_sizes[1]]->net->driver_pin
-                                     : node->input_pins[i]->net->driver_pin;
+            int input_index = flip ? i + node->input_port_sizes[1] : i;
+            nnet_t* net = node->input_pins[input_index]->net;
```
jeanlego, line 505: this is throwing an error with ASAN 
```bash
Architecture: k6_frac_N10_frac_chain_mem32K_40nm.xml
Reading FPGA Architecture file
Using Lut input width of: 4
Verilog: twobits_arithmetic_multiply.v
--------------------------------------------------------------------
High-level synthesis Begin
Parser starting - we'll create an abstract syntax tree. Note this tree can be viewed using Grap Viz (see documentation)
Adding file /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/regression_test/benchmark/verilog//operators/twobits_arithmetic_multiply.v to parse list
Optimizing module by AST based optimizations
Converting AST into a Netlist. Note this netlist can be viewed using GraphViz (see documentation)
==========================
Detected Top Level Module: 	simple_op
==========================
Performing Optimizations of the Netlist
Performing Partial Map to target device
Outputting the netlist to the specified output format
=================================================================
==17580==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x603000078810 at pc 0x00000059c74c bp 0x7ffeaad83c20 sp 0x7ffeaad83c18
READ of size 8 at 0x603000078810 thread T0
    #0 0x59c74b in define_mult_function(nnode_t*, _IO_FILE*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/multipliers.cpp:505
    #1 0x46c6d6 in output_node(nnode_t*, short, _IO_FILE*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/output_blif.cpp:313
    #2 0x46bd65 in depth_traverse_output_blif(nnode_t*, unsigned long, _IO_FILE*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/output_blif.cpp:239
    #3 0x46c226 in depth_traverse_output_blif(nnode_t*, unsigned long, _IO_FILE*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/output_blif.cpp:258
    #4 0x46bcaa in depth_first_traversal_to_output(short, _IO_FILE*, netlist_t*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/output_blif.cpp:220
    #5 0x46b52d in output_blif(_IO_FILE*, netlist_t*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/output_blif.cpp:175
    #6 0x40ba3b in synthesize_verilog /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/odin_ii.cpp:177
    #7 0x40c268 in start_odin_ii(int, char**) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/odin_ii.cpp:274
    #8 0x409f09 in main /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/main.cpp:7
    #9 0x7f05b098df44 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21f44)
    #10 0x4097d8  (/home/travis/vtr-build/ODIN_II/odin_II+0x4097d8)
0x603000078810 is located 0 bytes to the right of 32-byte region [0x6030000787f0,0x603000078810)
allocated by thread T0 here:
    #0 0x7f05b2773859 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x94859)
    #1 0x7f3dfd in vtr::realloc(void*, unsigned long) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/libs/libvtrutil/src/vtr_memory.cpp:63
    #2 0x61b4e6 in allocate_more_input_pins(nnode_t*, int) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_utils.cpp:146
    #3 0x5ffd57 in create_operation_node(ast_node_t*, signal_list_t**, int, char*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:3573
    #4 0x5cec08 in netlist_expand_ast_of_module(ast_node_t**, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:640
    #5 0x5f3a68 in assignment_alias(ast_node_t*, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:2775
    #6 0x5cd9ec in netlist_expand_ast_of_module(ast_node_t**, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:514
    #7 0x602f9d in create_case_mux_statements(ast_node_t*, nnode_t*, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:3962
    #8 0x602192 in create_case(ast_node_t*, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:3877
    #9 0x5cdb5f in netlist_expand_ast_of_module(ast_node_t**, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:525
    #10 0x5ce04e in netlist_expand_ast_of_module(ast_node_t**, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:566
    #11 0x5ce04e in netlist_expand_ast_of_module(ast_node_t**, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:566
    #12 0x5ce04e in netlist_expand_ast_of_module(ast_node_t**, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:566
    #13 0x5ce04e in netlist_expand_ast_of_module(ast_node_t**, char*, sc_hierarchy*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:566
    #14 0x5c9c07 in convert_ast_to_netlist_recursing_via_modules(ast_node_t**, char*, sc_hierarchy*, int) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:221
    #15 0x5c908e in create_netlist(ast_t*) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/netlist_create_from_ast.cpp:185
    #16 0x40b653 in synthesize_verilog /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/odin_ii.cpp:125
    #17 0x40c268 in start_odin_ii(int, char**) /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/SRC/odin_ii.cpp:274
    #18 0x409f09 in main /home/travis/build/verilog-to-routing/vtr-verilog-to-routing/ODIN_II/main.cpp:7
    #19 0x7f05b098df44 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21f44)
```

jeanlego:
also, it seems the output blif could be wrong.

from vtr_reg_strong
```
k6_N10_mem32K_40nm.xml/mkPktMerge.v/common num_primary_inputs: golden = 311 result = 5
```

ODIN_II/SRC/netlist_utils.cpp
```diff
@@ -443,14 +449,29 @@ void combine_nets(nnet_t* output_net, nnet_t* input_net, netlist_t* netlist) {
...
     if (join_to_net == other_net) {
-        if ((join_to_net->driver_pin) && (join_to_net->driver_pin->node != NULL) && join_to_net->driver_pin->node->related_ast_node != NULL)
-            error_message(NETLIST, join_to_net->driver_pin->node->related_ast_node->line_number, join_to_net->driver_pin->node->related_ast_node->file_number, "%s", "This is a combinational loop\n");
-        else
-            error_message(NETLIST, -1, -1, "%s", "This is a combinational loop\n");
-        if ((join_to_net->fanout_pins[0] != NULL) && (join_to_net->fanout_pins[0]->node != NULL) && join_to_net->fanout_pins[0]->node->related_ast_node != NULL)
-            error_message(NETLIST, join_to_net->fanout_pins[0]->node->related_ast_node->line_number, join_to_net->fanout_pins[0]->node->related_ast_node->file_number, "%s", "This is a combinational loop with more info\n");
-        else
-            error_message(NETLIST, -1, -1, "%s", "Same error - This is a combinational loop\n");
+        bool has_error = false;
```
jeanlego, line 452: I believe that join_to_net == other_net is the error, the fanout would never be reached in the previous version anyway, we are simply printing to allow the user to know whats going on, at which point we can just exit.

all that to say that the boolean is unecessary, this whole section is simply to tell the user

ODIN_II/SRC/multipliers.cpp
djns99: Gosh I love analysis tools

djns99: Saved me more times than I can count

djns99:
> also, it seems the output blif could be wrong.
> 
> from vtr_reg_strong
> 
> ```
> k6_N10_mem32K_40nm.xml/mkPktMerge.v/common num_primary_inputs: golden = 311 result = 5
> ```

Quite likely I havent had a chance to test any of it

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -3198,7 +3246,8 @@ static nnode_t* print_update_trace(nnode_t* bottom_node, int cycle) {
             for (i = 0; i < node->num_input_pins; i++) {
                 npin_t* pin = node->input_pins[i];
                 nnet_t* net = pin->net;
-                nnode_t* node2 = net->driver_pin->node;
+#warning "Oops this be broke come fix"
```
jeanlego, line 3249: this is failing Travis on compilation warnings, this can be solved with a for loop I believe

jeanlego:
:+1:  looking good so far. please ping me when this is ready for another review!

ODIN_II/SRC/simulate_blif.cpp
djns99: Sounds good. Will pick that up on Tuesday

ODIN_II/SRC/netlist_create_from_ast.cpp
djns99: I tried a few things to hit this, but I cant seem to find a case that has a non-null `assignment->pins[j]->node`. Any ideas?

```
module scratch(input clk, input in, output out);
reg a;
reg b;
always @(posedge clk)
begin
    a = in;
    b = a;
end
endmodule
```

Enters that loop. But I cant get it to actually do anything

ODIN_II/SRC/simulate_blif.cpp
djns99: Ok I have added a warning and stated that behaviour is undefined and left a TODO. Are you happy with that resolution for the time being? Or do you want me to explicitly update all values to unknown

jeanlego: I'd prefer update, its quick to do and it doesn't have a chance of being forgotten.

jeanlego: I have a PR up that should give you quad valued logic hopefully soon.

djns99: Ok hows that?

ODIN_II/SRC/netlist_create_from_ast.cpp
```diff
@@ -3130,15 +3132,18 @@ void terminate_registered_assignment(ast_node_t* always_node, signal_list_t* ass
...
 
-        if (pin->net->driver_pin) {
-            ref_string = pin->net->driver_pin->node->name;
+        if (pin->net->num_driver_pins) {
+            ref_string = pin->net->driver_pins[0]->node->name;
 
             for (j = i - 1; j >= 0; j--) {
-                if (list_dependence_pin[j] && list_dependence_pin[j]->net->driver_pin && list_dependence_type[j] && list_dependence_type[j] == BLOCKING_STATEMENT && strcmp(ref_string, assignment->pins[j]->node->name) == 0) {
+                if (list_dependence_pin[j] && list_dependence_pin[j]->net->num_driver_pins && list_dependence_type[j] == BLOCKING_STATEMENT && strcmp(ref_string, assignment->pins[j]->node->name) == 0) {
+                    if (list_dependence_pin[j]->net->num_driver_pins > 1 || pin->net->num_driver_pins > 1) {
```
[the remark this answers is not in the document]
jeanlego: I'm okay with that.

ODIN_II/SRC/partial_map.cpp
```diff
@@ -348,23 +347,31 @@ void instantiate_not_logic(nnode_t* node, short mark, netlist_t* /*netlist*/) {
...
- * (function: instantiate_buffer )
+ * (function: eliminate_buffer )
  * 	Buffers just pass through signals
+ * 	Returns true if the buffer could be eliminated
  *-------------------------------------------------------------------------------------------*/
-void instantiate_buffer(nnode_t* node, short /*mark*/, netlist_t* /*netlist*/) {
-    int width = node->num_input_pins;
-    int i;
-
+bool eliminate_buffer(nnode_t* node, short, netlist_t*) {
```
jeanlego, line 354: this should probably remove the buf node, since disconnecting it will make it unreachable from the net.

ODIN_II/SRC/partial_map.cpp
```diff
@@ -965,15 +972,15 @@ void instantiate_shift_left_or_right(nnode_t* node, operation_list type, short m
         remap_pin_to_new_node(node->output_pins[i], buf_node, i);
     }
     /* instantiate the buffer */
-    instantiate_buffer(buf_node, mark, netlist);
-
-    /* clean up */
-    for (i = 0; i < buf_node->num_input_pins; i++) {
-        buf_node->output_pins[i]->net = free_nnet(buf_node->output_pins[i]->net);
-        buf_node->input_pins[i] = free_npin(buf_node->input_pins[i]);
+    if( eliminate_buffer(buf_node, mark, netlist) ) {
```
jeanlego, line 975: this reaffirms that the eliminate buffer should clear the node.

jeanlego:
I think its close to ready. there are some failures, so if you can address that than I can merge it.

ODIN_II/SRC/partial_map.cpp
djns99: Im not sure how that would work? Line 136 doesn't clean it up so there must be some way of cleaning it up

djns99: Im pretty sure you are correct though, cause I am getting memory leaks by the looks

djns99:
Still some cases where this is broken which prevent InOut from being merged. Updated the test case to hit this error

ODIN_II/regression_test/benchmark/verilog/syntax/multi_assignment.v
```diff
@@ -0,0 +1,14 @@
+module select_bus(busout, bus0, bus1, bus2, bus3, enable, s);
```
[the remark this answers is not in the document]
djns99: When I tested with my inout cases I could reproduce with 1 bit wide. The problem appears to be the call to combine_nets() in the MUX setup. Im not entirely sure why it tries to merge two nets - Im still trying to understand that. The fix as far as I can see is to insert a buffer in combine_nets() if the number of driver pins is >1.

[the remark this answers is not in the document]
djns99: It fails with an error_message on the ast->netlist conversion so no BLIF

jeanlego:
I think the biggest issue actually appears to be the route we've taken. I wonder if letting inouts having separate nodes in both hash (input and output) would just solve this. Inputs don't get any buffer added so they are read straight from the the PI but if the output has multiple driver the buffer will coalesce them just fine. And having the buffer there causes no issue since it is in the right orientation to also drive the in of the inout

ODIN_II/SRC/partial_map.cpp
jeanlego: did you clear the memeory leak?

ODIN_II/SRC/netlist_create_from_ast.cpp
```diff
@@ -1925,12 +1926,15 @@ void connect_module_instantiation_and_alias(short PASS, ast_node_t* module_insta
                      * already been instantiated without any initial value. */
                     nnet_t* output_net = (nnet_t*)output_nets_sc->data[sc_spot_output];
                     nnet_t* input_new_net = (nnet_t*)output_nets_sc->data[sc_spot_input_new];
-                    if (output_net->driver_pin && output_net->driver_pin->node && input_new_net) {
-                        output_net->driver_pin->node->initial_value = input_new_net->initial_value;
+                    for (int k = 0; k < output_net->num_driver_pins; k++) {
+                        // TODO I am unsure if this is the correct behaviour
```
djns99, line 1930: @jeanlego I cant figure out what this code is trying to achieve so I cant say whether this is correct. Do you have any insight? The current behaviour is my best guess but I don't understand what scenario we would want to set all driver's initial values.

ODIN_II/regression_test/benchmark/task/preprocessor/simulation_result.json
```diff
@@ -201,6 +201,87 @@
         "Estimated LUTs": 1,
         "Total Node": 1
     },
+    "preprocessor/elsif_both_defined/no_arch": {
```
djns99, line 204: These test arent mine and they havent had results generated previously. Was there a reason for this or just someone forgot? Should I leave them out

ODIN_II/regression_test/benchmark/task/syntax/simulation_result.json
```diff
@@ -2312,6 +2300,33 @@
         "Estimated LUTs": 1020,
         "Total Node": 1315
     },
+    "syntax/multi_assignment/k6_frac_N10_frac_chain_mem32K_40nm": {
```
djns99, line 2303: These are mine, but they fail simulation at the moment (synthesis should be fine though). Should I leave the failing vectors, or should I remove them and come back in future once 4 valued simulation is implemented

ODIN_II/regression_test/benchmark/task/preprocessor/simulation_result.json
jeanlego: No, it seems there was some code changes that made the preprocessor work?? (Weird) and now since it synthesised it simulates, I'll double check

ODIN_II/regression_test/benchmark/task/syntax/simulation_result.json
jeanlego: No we keep failing tests in, (if you look around you'll see a few, and they pop up during verification (sig abort etc...) A lot of tests have been added even tho Odin doesn't support them, it's really just to speed up the development since the test are already in, easier to figure if you fixed an issue or not

ODIN_II/regression_test/benchmark/task/fpu/hard_logic/simulation_result.json
```diff
@@ -14,7 +14,7 @@
         "blif": "bgm.blif",
         "exit": 134,
         "errors": [
-            "[OUTPUT_BLIF] Could not resolve memory hard block fpu_mul~0 to a valid type."
+            "[OUTPUT_BLIF] Could not resolve memory hard block fpu_mul~6 to a valid type."
         ]
```
jeanlego, line 18: The pin changed here which worries me

ODIN_II/regression_test/benchmark/task/preprocessor/synthesis_result.json
```diff
@@ -157,57 +160,67 @@
...
+        "Longest Path": 3,
+        "Average Path": 3,
+        "Estimated LUTs": 12,
+        "Total Node": 12
     },
     "preprocessor/nested_ifdef_twodef/no_arch": {
         "test_name": "preprocessor/nested_ifdef_twodef/no_arch",
         "verilog": "nested_ifdef_twodef.v",
-        "exit": 1,
-        "errors": [
```
jeanlego, line 205: It seems it passes a bunch of test that didn't before now

ODIN_II/regression_test/benchmark/task/syntax/synthesis_result.json
```diff
@@ -2091,6 +2093,8 @@
         "verilog": "part_select.v",
         "warnings": [
             "part_select.v:15:1 [NETLIST] Net (null) driving node top.inst_3^bAND~0^lAND~7 is itself undriven.",
+            "part_select.v:15:1 [NETLIST] Net (null) driving node top.inst_3^bAND~0^lAND~7 is itself undriven.",
```
jeanlego, line 2096: How come this is duplicated?

djns99:
> The results look good because of the buffer, but I'm worried the changes might hide some of the impact of the multidriver, is there any way you can make a separate PR for the buffer stuff and then just use that and branch this PR off of that?
> It's hard to tell wether or not it's right otherwise...

https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1398#discussion_r453163306 I dont think we can remove it, its necessary for the case that there are two drivers on the output net, in which case I either have to insert some place holder buffers or do this. I can do the former if you would like to see the effect on the QoR, but this is strictly worse behaviour and will have to be changed back for In Outs anyway

ODIN_II/regression_test/benchmark/task/syntax/synthesis_result.json
djns99: print_dot_names_header() check this condition twice now, once in warn_undriven() and once in print_input_single_driver(). I can probably remove the first call without any issue though if it matters?

jeanlego:
> #1398 (comment) I dont think we can remove it, its necessary for the case that there are two drivers on the output net, in which case I either have to insert some place holder buffers (like I do for a MUX) or do this. I can do the former if you would like to see the effect on the QoR, but this is strictly worse behaviour and will have to be changed back for In Outs anyway

I don't want you to remove it at all, I'm just curious if you can do the buffer alone first, then add this PR other commits on top of it, if you can, then I could effectively merge the buffer stuff before the inout, and IF it affects QoR, it's easier to make a case for it. 
I know I've said this before, but I just wanna make sure you really can't. It'll make merging this a whole lot easier

ODIN_II/regression_test/benchmark/task/syntax/synthesis_result.json
jeanlego: Yeah, I don't think it's necessary to keep both. Given undriven will be either dropped or printed out, you should get the warning eventually

ODIN_II/regression_test/benchmark/task/fpu/hard_logic/simulation_result.json
djns99: I cant think of a good reason this would happen. I have a couple of theories, since I changed the read_blif slightly and also output slightly different which may cause the BLIF to encounter them in a different order

djns99: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1398/files#diff-ab25514ea7cdfbd91f356b27ad8c7ceeR2588 this and this https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1398/files#diff-f48accd8c411d18c3edaba9f7a21dc29R688 are my main thoughts

djns99:
> I don't want you to remove it at all, I'm just curious if you can do the buffer alone first, then add this PR other commits on top of it, if you can, then I could effectively merge the buffer stuff before the inout, and IF it affects QoR, it's easier to make a case for it.
> I know I've said this before, but I just wanna make sure you really can't. It'll make merging this a whole lot easier

Ok it looks like a relatively straightforward change anyway, just changing some if statements around. Will give it a shot

ODIN_II/SRC/netlist_create_from_ast.cpp
jeanlego: So nets technically don't have initial values, only latches can hold one. But when we build the circuit, we assign the init value to the net., With that in mind, we need to push that init value down to the node that the net is driving, of there is an init value on the net other then unknown, you should expect a latch bellow

djns99: So the driver should be a latch? In which case we would expect this to only apply when there one driver, else it would be an error since the net has multiple drivers one of which is a latch that is not guarded by a bufif/MUX

jeanlego: No, the driver you are looking at is driving a latch, the init value just hasn't been pushed down on the latch yet

djns99: From looking into it I can see a situation that we have multiple drivers one of which is a latch. There are cases were I can see it being valid, for example constant conditions or preprocessor behaviour that causes only one driver to resolve to connected. In these cases we need to propagate the initial value to whichever driver is the latch. I have added asserts elsewhere in the pipeline that should catch the case where two distinct initial values (not including Z) are associated with the net. This case should always be ill formed for our purposes to the best of my knowledge.

ODIN_II/regression_test/benchmark/task/fpu/hard_logic/simulation_result.json
djns99: @jeanlego Is there anything more you would like me to do on this?
I haven't been able to identify a specific reason why the count is different but I personally don't think this is an issue.

djns99:
@jeanlego If you are happy can we get this merged?

PR #1399, opened 2020-07-04
Improved the verify odin script

djns99:
+ Outputs MISMATCH instead of SIGABRT on simulation failure
+ Allows passing regex for tests to run

#### Description

Adds support for using regular expressions to specify a subset of tests to run.

e.g. `./verify_odin.sh -t regression_test/benchmark/task/syntax syntax/bm_*`

#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/verify_odin.sh
+++ b/ODIN_II/verify_odin.sh
@@ -47,14 +47,14 @@ global_failure="test_failures.log"
 ##############################################
 # Exit Functions
 function exit_program() {
-	
+
 	FAIL_COUNT="0"
 	if [ -f "${NEW_RUN_DIR}/${global_failure}" ]; then
 		FAIL_COUNT=$(wc -l "${NEW_RUN_DIR}/${global_failure}" | cut -d ' ' -f 1)
 	fi
 
 	FAILURE=$(( FAIL_COUNT ))
-	
+
 	if [ "_${FAILURE}" != "_0" ]
 	then
 		echo "Failed ${FAILURE} benchmarks"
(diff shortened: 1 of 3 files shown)


djns99:
@jeanlego Where should I be documenting this?

jeanlego:
> @jeanlego Where should I be documenting this?

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/doc/src/odin/dev_guide/verify_script.md

ODIN_II/verify_odin.sh
```diff
@@ -881,12 +881,20 @@ function sim() {
...
-				if [ "_${_SUBTEST_LIST[*]}" != "_" ] \
-				&& ! echo "${_SUBTEST_LIST[@]}" | grep "${TEST_FULL_REF}" &> /dev/null;
+				if [ "_${_SUBTEST_LIST[*]}" != "_" ]
 				then
-					# skip duplicate tests
 					run_this_test="off"
-				elif [ -d "${NEW_RUN_DIR}/${TEST_FULL_REF}" ]
+					for subtest in "${_SUBTEST_LIST[@]}";
+					do
+						if echo "_${TEST_FULL_REF}" | grep -E "_${subtest}" &> /dev/null
```
jeanlego, line 889: if you dropped the underscore, i think you could just grep the whole set, you wouldnt have to specify the suite_name either.

djns99: Im not sure what you mean? The `_SUBTEST_LIST[@]` is the list of regular expressions so I cant grep that set. The underscore seems unnecessary I agree, but I am not sure how that allows us to avoid needing the suite_name

jeanlego: I believe that if you also get rid of some of this, it will get rid of the requirement of the suite being passed in. 
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3e17e6dfa1cf79e9b925bf764b2880f51bbae2fb/ODIN_II/verify_odin.sh#L695-L715
you will need to check that at least one test was run for the task before continuing on with the parsing phase, but this would let you use just a regex without specifying the task name.

jeanlego:
```bash
 ########################################## 
 # check if we only run some subtask 
 run_benchmark="on" 
 run_subtest_only="" 
 if [ "_${_SUBTEST_LIST[*]}" != "_" ]; 
 then 
 	run_subtest_only="--subset" 
 fi 
```

djns99: Yeah already done

jeanlego: does it work? (you can pass --dry_run to not wait through all the shenanigans)

djns99: Looks like
`./verify_odin.sh -t regression_test/benchmark/task/syntax *bm_* *8_bit*`
Runs only the tests matching those regex

djns99: And
`./verify_odin.sh -t regression_test/benchmark/task/syntax *some_unmatchable_regex*`
Runs nothing and exits without error

jeanlego: > does it work? (you can pass --dry_run to not wait through all the shenanigans)

tested locally and it does :)

jeanlego: >Runs nothing and exits without error

looks perfect to me then! 
thanks for this! a much needed upgrade!

jeanlego:
changed the indentation?

djns99:
> changed the indentation?

Yes I removed the if statement that disabled the whole suite and just did the per test check

jeanlego:
> Yes I removed the if statement that disabled the whole suite and just did the per test check
check the diff, you use space or tabs? it seems to be now inconsistent

djns99:
> > Yes I removed the if statement that disabled the whole suite and just did the per test check
> 
> check the diff, you use space or tabs? it seems to be now inconsistent (from tab -> 2 space)

Oops yes. I have fixed that hopefully

jeanlego:
lgtm!

PR #1400, opened 2020-07-05
Stratix10 like archs

MohamedEldafrawy:
Added three new architectures that are created using COFFE_22nm delays:
-  Stratix-10-like architecture
-  Stratix-10-like architecture with 4-bit adders per ALM and two carry chains each spanning 5 ALMs
-  Stratix-10-like architecture with 5 ALMs having 4-bit adders each and 5 ALMs without arithmetic logic


vaughnbetz:
Thanks Mohamed.  Can you add some info to the comment at the top of the file about where the RAM block delays come from (Stratix IV I think).  Otherwise looks good to merge.

MohamedEldafrawy:
Yes that's true they come from Stratix IV. Will update the comment accordingly.

vaughnbetz:
Thanks.  At the same time it's also good to comment the area numbers.  I think it is COFFE for clb tile and routing, no data for DSP and RAM.  In any case, commenting what is and isn't there is a good idea.

vaughnbetz:
Hi Mohamed, sorry to nag, but if you could make those comment changes we can get this merged -- would be good to close it out.

MohamedEldafrawy:
Hi Vaughn, sorry for the delay. I'll get done by tomorrow at most :).

vaughnbetz:
Great, thanks Mohamed.

MohamedEldafrawy:
@vaughnbetz I modified the architecture file top comments to include the correct area and delay description. Please let me know if you want me to do any other modifications.

vaughnbetz:
Thanks Mohamed.  It looks good; I'll merge it once the tests finish.

PR #1429, opened 2020-07-15
Vtr flow script py rewrite

shadtorrie:
I converted run_vtr_flow.pl to run_vtr_flow.py

#### Description
Using the [old python rewrite branch](https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/vtr_script_py_rewrite_old) as a starting place, I converted the run_vtr_flow.pl to run_vtr_flow.py
Also from the old rewrite I imported and cleaned up the python library that is used by the run_vtr_flow.py script. 

#### Related Issue
[The issue this partially fixes](https://github.com/SymbiFlow/ideas/issues/16)

#### Motivation and Context
We want to be able to run the vtr test suite without requiring perl to be installed.

I enabled the regression tests to be run on the python script. I currently have two sets of the regression tests being run on travis, the perl script is run on one set of regression tests and the python the other.

#### Types of changes
- [x] New feature (change which adds functionality): run run_vtr_flow.py instead of run_vtr_flow.pl

#### Checklist:


vtr_flow/scripts/python_libs/doc/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
mithro, line 1: This should probably be part of the verilog-to-routing docs at https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/doc

vtr_flow/scripts/python_libs/vtr/abc/__init__.py
```diff
@@ -0,0 +1 @@
+from .abc import run, run_lec
```
mithro, line 1: Need to make sure you have newlines at the end of all your files

jgoeders:
> * Make sure we run a python linter on the code base (and it is checked by CI).
> * Start using a code auto-formatter for Python (something like https://github.com/google/yapf or https://github.com/psf/black)

Tim, I'll work on getting a python linter and auto-formatter set up.  Do you want to block this PR on having that set up?

mithro:
> > * Make sure we run a python linter on the code base (and it is checked by CI).
> > * Start using a code auto-formatter for Python (something like https://github.com/google/yapf or https://github.com/psf/black)
> 
> @mithro, I'll work on getting a python linter and auto-formatter set up. Do you want to block this PR on having that set up?

Nope, just want to track them as TODO items before we forget :-)

vaughnbetz:
Discussed today in the sync up meeting.  Next step is to write up a test plan and summarize the test results.  Here are some thoughts for that plan:
- CI passes when there are no qor or other failures in the vtr code.
- We get an execution error in the vtr flow (break any program you want; e.g. add a VPR_EXIT at the start of vpr).  Confirm regtests fail and are flagged in CI
- We get a qor failure in the vtr flow.  To do this, break one of the optimizers and then run and ensure we get QoR failures. For example, you could. change this line to set move_lim to 1: 
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/a67e72c9858f2c42cd4a2adb73467034cf2ad276/vpr/src/place/place.cpp#L674

vtr_flow/scripts/python_libs/vtr/abc/__init__.py
shadtorrie: This was fixed in the most recent commit

vtr_flow/scripts/vtr-flow.py
```diff
@@ -173,7 +173,30 @@ def vtr_command_argparser(prog=None):
...
+    #
+    # Iterative black-boxing flow arguments
+    #
+    iterative = parser.add_argument_group("Iterative", description="Iterative black-boxing flow for multi-clock circuits options")
+    iterative.add_argument("-iterative_bb",
+                            default=False,
+                            action="store_true",
+                            dest="iterative_bb",
+                            help="Use iterative black-boxing flow for multi-clock circuits")
+    iterative.add_argument("-once_bb",
```
jeanlego: If we are to port this over, I believe the three method were there just for experimenting, we settled on the itterative_bb method. 

side note: the blanket bb was black boxing all the latches (blanket black boxing).

vtr_flow/scripts/vtr-flow.py
```diff
@@ -203,6 +226,23 @@ def vtr_command_main(arg_list, prog=None):
                                    verbose=True if args.verbosity > 2 else False,
                                    echo_cmd=True if args.verbosity >= 4 else False)
     exit_status = 0
+    flow_type = 2 #Use iterative black-boxing flow for multi-clock circuits
+    use_new_latches_restoration_script = 1
```
jeanlego: It was agreed to enable it by default given not enabling gives the wrong circuit.
I don't think we should still provide a way to disable given that its "broken" without it, and doesn't affect the singly clocked circuit.

vaughnbetz: I think it's still good to be able to control this so we can experiment in future abc versions, which might change their clocking behaviour, or to experiment to see how much proper clock handling is costing in case we're comparing to naive tools that don't do it.  It should be on by default though.

vtr_flow/scripts/python_libs/verilogtorouting/flow.py
```diff
@@ -20,7 +21,9 @@ def run_vtr_flow(architecture_file, circuit_file,
                  parse_config_file=None,
                  temp_dir="./temp", 
                  verbosity=0,
-                 vpr_args=None):
+                 vpr_args=None,
+                 flow_type=2,
```
vaughnbetz: Can we give this a name instead of a number? I'm not sure what flow_type = 2 means, but a name could be much easier to understand.

vtr_flow/scripts/vtr-flow.py
```diff
@@ -203,6 +226,23 @@ def vtr_command_main(arg_list, prog=None):
                                    verbose=True if args.verbosity > 2 else False,
                                    echo_cmd=True if args.verbosity >= 4 else False)
```
vaughnbetz: Would be good to have a comment where verbosity is first defined that gives the legal range and meaning of various levels.

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -0,0 +1,510 @@
...
+    """
+    pass
+
+class CommandRunner(object):
+
+    def __init__(self, timeout_sec=None, max_memory_mb=None, track_memory=True, verbose_error=False, verbose=False, echo_cmd=False, indent="\t", show_failures=False,valgrind=False):
+        """
+        An object for running system commands with timeouts, memory limits and varying verbose-ness 
+
+        Arguments
```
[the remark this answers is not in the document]
shadtorrie: I have made all docstrings match the numpy format.

vtr_flow/scripts/python_libs/vtr/abc/abc.py
```diff
@@ -0,0 +1,302 @@
...
+
+def populate_clock_list(circuit_file,blackbox_latches_script,clk_list,command_runner,temp_dir,log_filename):
+    clk_list_path = Path(temp_dir) / "report_clk.out"
+    cmd = [blackbox_latches_script, "--input", circuit_file.name,"--output_list", clk_list_path.name]
+    command_runner.run_system_command(cmd, temp_dir=temp_dir, log_filename=log_filename, indent_depth=1)
+    with clk_list_path.open('r') as f:
+        for line in f.readlines():
+            clk_list.append(line.strip('\n'))
+
+def run_lec(reference_netlist, implementation_netlist, command_runner=CommandRunner(), temp_dir=".", log_filename="abc.dec.out", abc_exec=None):
```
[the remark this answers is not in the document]
shadtorrie: This has been fixed.

vtr_flow/scripts/vtr-flow.py
shadtorrie: If I understand correctly, I should remove all  the options and just use the itterative_bb setting? Iterative_bb is the default currently.

vaughnbetz: See my comment below.  I think we should still be able to turn this on and off for experiments (justification in comment below) so I'd leave the capability in, but it should be iterative_bb by default.

vtr_flow/scripts/vtr-flow.py
shadtorrie: Is this in regards to flow_type or use_new_latches_restoration_script?

vaughnbetz: flow_type.  bool variables that are 0 or 1 are fine, but flow_type = 2 is mysterious.  flow_type = iterative_bb is more self-explanatory, if that is what flow_type = 2 means.  Or if flow_type = 2 means implement twice in two different flows and compare, then it should be renamed flows_to_compare = 2 or flow_type = implement_twice (or something like that).

vtr_flow/scripts/python_libs/vtr/vpr/vpr.py
```diff
@@ -0,0 +1,305 @@
...
+    vpr_args['route'] = True #Re-route only
+    vpr_args['route_chan_width'] = relaxed_W #At a fixed channel width
+
+    #VPR does not support performing routing when fixed pins 
+    # are specified, and placement is not run; so remove the option
+
+    run(architecture, circuit_name, circuit, command_runner, temp_dir, log_filename=vpr_relaxed_W_log, vpr_exec=vpr_exec, vpr_args=vpr_args, check_for_second_run=False)
+    
+
+def run(architecture, circuit_name, circuit, command_runner=CommandRunner(), temp_dir=".", output_netlist=None,
```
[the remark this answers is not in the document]
shadtorrie: This has been fixed

vtr_flow/scripts/vtr-flow.py
shadtorrie: This has been fixed.

vtr_flow/scripts/vtr-flow.py
shadtorrie: This has been fixed

vtr_flow/scripts/python_libs/verilogtorouting/flow.py
shadtorrie: This has been fixed

vtr_flow/scripts/vtr-flow.py
shadtorrie: These levels were previous functionality that the perl script does not implement so I changed it to match the perl script, which uses a boolean to either output the error description or not.

vtr_flow/scripts/python_libs/vtr/flow.py
```diff
@@ -0,0 +1,288 @@
...
+            Stage of the flow to finish at
+        
+        temp_dir         : 
+            Directory to run in (created if non-existent)
+        
+        command_runner   : 
+            A CommandRunner object used to run system commands
+        
+        verbosity        : 
+            Weather to output error description or not
```
vaughnbetz, line 61: Nit: weather --> whether

vaughnbetz:
Thanks.  Changes look good to me; just found one typo (flagged above).

vtr_flow/scripts/python_libs/vtr/flow.py
```diff
@@ -0,0 +1,288 @@
...
+import shutil
+import subprocess
+import time
+import glob
+import vtr
+from pathlib import Path
+from collections import OrderedDict
+from vtr import CommandError
+
+VTR_STAGE = vtr.make_enum("odin", "abc", 'ace', "vpr")
```
jgoeders, line 11: Python has an enum class now.  Can we just use that instead?

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -0,0 +1,515 @@
...
+    Makes an enumeration
+
+    >> MY_ENUM = enum(ONE, TWO, THREE)
+    """
+    enums = dict(zip(sequential, range(len(sequential))), **named)
+    reverse = dict((value, key) for key, value in enums.items())
+    enums['reverse_mapping'] = reverse
+    return type('Enum', (), enums)
+
+def mkdir_p(path):
```
jgoeders, line 249: pathlib.mkdir(parents=True) should work here.  We don't need this function.

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -0,0 +1,515 @@
...
+                data_row = OrderedDict()
+
+                for i, value in enumerate(csv_row):
+                    data_row[header[i]] = value.strip()
+
+                data.append(data_row)
+
+    return data
+
+def make_enum(*sequential, **named):
```
jgoeders, line 238: I don't think we need this in Python3

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -0,0 +1,515 @@
...
+                        return str(file_path)
+    return None
+
+def find_vtr_root():
+    for env_var in ['VTR_ROOT']:
+        if env_var in os.environ:
+            return os.environ[env_var]
+
+    #We assume that this file is in <vtr_root>/vtr_flow/python_libs/verilogtorouting
+    inferred_vtr_root = Path(__file__).parent / '../../../..'
```
jgoeders, line 340: Use .parent.parent.parent.parent instead of '../../../..'

vtr_flow/scripts/python_libs/vtr/error.py
```diff
@@ -0,0 +1,32 @@
+class VtrError(Exception):
+    """
+    Base class for VTR related exceptions
+
+    Attributes:
+        msg -- An explanation of the error
+    """
+    def __init__(self, msg=""):
```
jgoeders, line 8: In these cases it's nice to do something like this, with variable arguments, so that you can build strings with multiple arguments like when you call print.  Make it nice so you don't have to convert everything to str, etc.

def error(*msg):
    self.msg =  " ".join(str(item) for item in msg)

vtr_flow/scripts/python_libs/vtr/error.py
```diff
@@ -0,0 +1,32 @@
...
+
+class CommandError(VtrError):
+    """
+    Raised when an external command failed.
+
+    Attributes:
+        returncode -- The return code from the command
+        cmd == The command run
+    """
+    def __init__(self, msg, cmd, returncode, log=None):
```
jgoeders, line 19: See previous

vtr_flow/scripts/python_libs/vtr/error.py
```diff
@@ -0,0 +1,32 @@
...
+        super(CommandError, self).__init__(msg=msg)
+        self.returncode = returncode
+        self.cmd = cmd
+        self.log = log
+
+class InspectError(VtrError):
+    """
+    Raised when some query (inspection) result is not found.
+    """
+    def __init__(self, msg, filename=None):
```
jgoeders, line 29: See previous.

vtr_flow/scripts/python_libs/vtr/inspect.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def type(self):
+        raise NotImplementedError()
+
+    def check_passed(golden_value, check_value):
+        raise NotImplementedError()
+
+class EqualPassRequirement(PassRequirement):
+    def __init__(self, metric):
+        super(EqualPassRequirement, self).__init__(metric)
```
jgoeders, line 49: Should use python3 super format, ie: super().__init__(args)

This occurs in many places.

vaughnbetz:
Not sure what the conflict is -- can you take a look Shad?  If this is ready to merge let me (or someone else who can merge it) know.

vtr_flow/scripts/python_libs/vtr/flow.py
shadtorrie: This has been fixed.

vtr_flow/scripts/python_libs/vtr/flow.py
shadtorrie: The python enum has now been implemented.

vtr_flow/scripts/python_libs/vtr/util.py
shadtorrie: mkdir_p has been removed and replaced with pathlib.mkdir

vtr_flow/scripts/python_libs/vtr/util.py
shadtorrie: This has been removed.

vtr_flow/scripts/python_libs/vtr/inspect.py
shadtorrie: This has been fixed

vtr_flow/scripts/python_libs/vtr/util.py
shadtorrie: This has been fied

vtr_flow/scripts/python_libs/vtr/error.py
shadtorrie: this has been fixed

vtr_flow/scripts/python_libs/vtr/error.py
shadtorrie: this has been fixed

vtr_flow/scripts/python_libs/vtr/error.py
shadtorrie: this has been fixed

jgoeders:
@vaughnbetz OK, this should be ready to go now.  Shad ran some extra checks as discussed, and it's ready to merge now.

PR #1432, opened 2020-07-16
Odin_II: Fix Issue1406, defparam's overriding localparams 

emacdo12:
#### Description

Made two changes. The first being odin now recognizes localparameters and doesn't error. The second fix is defparam cannot override localparams but instead throws a warning.

#### Related Issue

Issue 1406

#### Motivation and Context

Adhere to the verilog standard.

Added new benchmarks and passes old benchmarks that used to fail.

#### Types of changes

#### Checklist:


ODIN_II/SRC/ast_elaborate.cpp
```diff
@@ -2216,36 +2216,43 @@ void update_instance_parameter_table(ast_node_t* instance, STRING_CACHE* instanc
...
+                    /**
                  * TODO: is this needed?
                  * ast_node_t* temp = parameter_override_list->children[j];
                  * */
-                remove_child_from_node_at_index(parameter_override_list, j);
-                //temp = NULL;
+                    remove_child_from_node_at_index(parameter_override_list, j);
+                    //temp = NULL;
+                } else if (temp->types.variable.is_localparam) {
+                    warning_message(AST, parameter_override_list->loc,
```
jeanlego, line 2245: you should still free everything related, the only thing that you should guard  is swapping the node in the table

jeanlego: this would apply to the error thrown above also. but more importantly, can you answer the TODO?

ODIN_II/SRC/ast_elaborate.cpp
```diff
@@ -2216,36 +2216,43 @@ void update_instance_parameter_table(ast_node_t* instance, STRING_CACHE* instanc
...
+                                    parameter_override_list->children[j]->children[0]->types.identifier);
+                }
             }
         }
         // check if there are still overrides in the list that haven't been deleted
 
-        if (parameter_override_list->num_children > 0) {
-            oassert(false); // not reachable???
-        }
+        //if (parameter_override_list->num_children > 0) {
```
jeanlego, line 2253: make sure you keep this, guard since we should have covered everything otherwise there will be a memory leak

Issue #1435, opened 2020-07-16
Python auto-formatting


jgoeders:
We are currently auto-formatting C/C++ code using `clang-format`, but we aren't formatting Python code.  With the inclusion of more Python scripts (#1429), we should add Python formatting.

#### Proposed Behaviour
`make format` should also format Python code.

#### Current Behaviour
No automated Python formatting

#### Possible Solution
Try something like https://github.com/google/yapf or https://github.com/psf/black

PR #1437, opened 2020-07-17
Python auto-formatter

jgoeders:
Adding a Python auto-formatter

#### Description
Add new target `make format-py` which runs the `black` python auto-formatter on all .py files

#### Related Issue
#1435 

#### Motivation and Context
With the change of the scripts to python #1429, now is a good time to introduce a python automatter.

New CI test verifying python code is formatted

#### Types of changes

#### Checklist:


cmake/modules/AutoPyFormat.cmake
```diff
@@ -0,0 +1,7 @@
+# Use the black python formatter on the entire repo
+#
+# While black can perform it's own recursive search, it follows symlinks
+# which seems to be a problem with the Travis CI build
+add_custom_target(format-py 
+    COMMAND find ${PROJECT_SOURCE_DIR} -iname \"*.py\" -print0 | xargs -0 -P 0 -n 1 black
```
jeanlego, line 6: I think this shouldn't be applied to anything we import.

jgoeders: You mean excluding things like abc...?

jgoeders: Fixed in 4f6d2fd

jgoeders:
@jeanlego I see that you requested a review.  Have you had a chance to look at this further?  Thanks!

jeanlego:
> @jeanlego I see that you requested a review. Have you had a chance to look at this further? Thanks
Sorry fat thumb and small screen. Must've hit that by mistake

vaughnbetz:
I'm OK with the reformatting.  

One nit: Is it trying to fit everything into 88-character lines (it's default), and if so should we specify a somewhat longer line length?  Some not-too-long function calls are split into multiple lines now.  I don't have an extremely strong opinion on this, but if the C++ code formatter is allowing a longer than 88 character line then it would be good to line the python auto-formatter up with that line length if possible.  I am not sure what the C++ auto-formatter line length is set to, but I vaguely recall discussions around 100 characters so it might be that.

litghost:
> The auto-formatting looks good to me. @litghost: what line length are we using for C++ code? Looks like the python auto-formatter is defaulting to 88 characters. I vaguely recall using 100 for C++. If so, I think we should line up the line length at 100 (and the auto-formatter has an option for line length).
> Current format has a few too many line breaks to be ideal for my (highly subjective) taste :).

As discussed, the python formatter (`black`) has little to know configuration, and so we won't be able to change that.  I actually prefer shorter line length, and believe that code should probably be restructured if it spills over the line limit.

I do agree that the C++ formatter has a wider format limit, I believe it is somewhere between 100 and 120 characters.

jgoeders:
Good catch.  I didn't realize we were using non-standard options for clang-format for the C++ code.  Looks like we currently have no line limit for C++ code.   (ColumnLimit: 0 in https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.clang-format)

I don't see a documented option for unlimited in black (0 makes very short lines).  So we could use 100.  Or we could make it something like 500 to make it virtually unlimited.

I don't feel strongly, although I agree that 100 would be better than 88.

jgoeders:
Didn't see the most recent messages before posting mine.  What's the consensus? Leave at 88 or go larger?

vaughnbetz:
@litghost: it can be configured (https://black.readthedocs.io/en/stable/the_black_code_style.html#:~:text=You%20probably%20noticed%20the%20peculiar,used%20by%20the%20standard%20library).  --line_length

I prefer 100 to 88 (displays are wide these days), but don't have a super strong opinion if others disagree.
If we don't wrap any C++ lines I think we should change the format to 100 chars, but we can separate that.  Not sure there's a good reason for arbitrarily long lines in C++ -- if there is, let me know!

jeanlego:
I think the toml could be useful for this, you can pass in exclude there too.

https://black.readthedocs.io/en/stable/pyproject_toml.html

vaughnbetz:
This looks ready to go but has conflict; if you fix we can merge.

shadtorrie:
Black and pylint like two different indentations for arguments of a function. This is an issue if black is run after you have already cleaned up a python file. It causes a bad-continuation error with pylint after running black. 
This is what black likes:
image
and this is what pylint likes:
image

litghost:
Can you provide the output from pylint and the actual files for your example?  The second looks wrong to me, even by pylints standards.

vaughnbetz:
Is the difference that pylint indents two tab stops for the arguments of a function?  That doesn't seem like a good idea (wastes line width/inconsistent) so I'd go with the black version if that's the case.  Does pytlint have configuration settings to suppress this check/message?

mithro:
@shadtorrie -- FYI - You should be able to provide a configuration to pylint which makes it understand it.

jgoeders:
This black/pylint incompatibility for indenting of function arguments is actively being discussed:  https://github.com/psf/black/issues/48

There is no way to configure black to make pylint happy, so I think the answer for now is to disable this check in pylint...unless someone has a better solution.

jgoeders:
Is anyone able to provide any insight on this CI failure?  

The error in the log is:

```
graphics_commands:        k6_N10_mem32K_40nm.xml/stereovision3.v/common         		Error: Executable vpr failed
	full command:  /usr/bin/env time -v /tmpfs/src/github/vtr-verilog-to-routing/vpr/vpr k6_N10_mem32K_40nm.xml stereovision3 --circuit_file stereovision3.pre-vpr.blif --route_chan_width 100 --graphics_commands set_draw_block_outlines 0; set_draw_block_text 0; set_draw_block_internals 2; set_draw_net_max_fanout 128; save_graphics place.png; set_nets 1; save_graphics nets1.png; set_nets 2; save_graphics nets2.png; set_nets 0; set_cpd 1; save_graphics cpd1.png; --max_router_iterations 150
	returncode  :  23
	log file    :  vpr.out
failed: Executable vpr failed (took 8.31 seconds)
```

The Travis build is passing, and I've tried running `./run_reg_test.pl vtr_reg_strong -show_failures -skip_qor -j8` on two different machines, and neither have exhibited the same failure as the kokora run.

Any ideas/suggestions?

litghost:
That failure is running with the sanitizers on.  I assume you just compiled normally?  Anyways, that failure appears to be in master too, someone pushed or merged code without waiting for green CI.  FYI @vaughnbetz

jgoeders:
Thanks @litghost!  I wasn't familiar with the sanitizers, but I see you have to enable them in build.  Makes sense.

vaughnbetz:
@MahshadJ , can you take a look?  Since this is a graphics failure I am guessing it may be related to code you pushed.  Even if not, you are likely a good person to debug it :).
It looks like to reproduce it you may need to have the sanitizers enabled in the build.  I suggest running the regtest Jeff mentions above with and without the sanitizers to see what happens.  The developer guide on the VTR website explains how to turn on the sanitizers (extra memory checking etc.) when buiding.

vaughnbetz:
That failure is definitely unrelated -- memory leak (mostly deep in graphics libraries and harmless) with sanitizers and graphics on.  Doesn't gate this PR.

vaughnbetz:
@jgoeders @shadtorrie Is there anything left to do before merging this?  If not, you can click the ready to review button and I can merge it.  I think all reviews are done now, but if there's something else that needs doing you should list it here so we can move this PR along.

jgoeders:
Sorry, I've been away the last week.  There's a merge conflict I'll fix now and then merge it once things are passing again.

PR #1438, opened 2020-07-18
Remove pres cost

Bill-hbrhbr:
#### Related Issue
Final optimization decision on PR #1396.

#### Types of changes

- [x] Optimization (change which speeds up code)
#### Checklist:

### DIFF
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -8,8 +8,8 @@
  * As a note, looks into draw_global.c for understanding the data structures associated with drawing->
  *
  * 
- * Authors: Vaughn Betz, Long Yu (Mike) Wang, Dingyu (Tina) Yang
- * Last updated: June 2019
+ * Authors: Vaughn Betz, Long Yu (Mike) Wang, Dingyu (Tina) Yang, Bingran (Bill) Hu
+ * Last updated: July 2020
  */
 
 #include <cstdio>
@@ -45,6 +45,7 @@
 #include "save_graphics.h"
 #include "timing_info.h"
 #include "physical_types.h"
+#include "route_common.h"
 
 #ifdef WIN32 /* For runtime tracking in WIN32. The clock() function defined in time.h will *
               * track CPU runtime.														   */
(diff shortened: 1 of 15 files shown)


vpr/src/draw/draw.cpp
```diff
@@ -8,8 +8,8 @@
  * As a note, looks into draw_global.c for understanding the data structures associated with drawing->
  *
  * 
- * Authors: Vaughn Betz, Long Yu (Mike) Wang, Dingyu (Tina) Yang
- * Last updated: June 2019
+ * Authors: Vaughn Betz, Long Yu (Mike) Wang, Dingyu (Tina) Yang, Bingran (Bill) Hu
+ * Last updated: July 2020
```
kmurray, line 12: We might just want to drop these lines, since they aren't really up to date anyway.

vpr/src/draw/draw_types.h
```diff
@@ -193,6 +194,7 @@ struct t_draw_state {
     bool forced_pause = false;
     int sequence_number = 0;
     float net_alpha = 0.1;
+    float pres_fac = 1.;
```
kmurray, line 197: We'll have to be careful about how this is kept up to date. It would be really difficult to debug if this ever got out of sync with the router pres_fac as the graphics would be showing something different.

vpr/src/route/route_common.cpp
```diff
@@ -414,57 +412,60 @@ void pathfinder_update_path_cost(t_trace* route_segment_start,
...
-     * DATE.                                                                     */
+void pathfinder_update_acc_cost_and_overuse_info(float acc_fac, OveruseInfo& overuse_info) {
+    /* This routine recomputes the acc_cost (accumulated congestion cost) of each       *
+     * routing resource for the pathfinder algorithm after all nets have been routed.   *
+     * It updates the accumulated cost to by adding in the number of extra signals      *
+     * sharing a resource right now (i.e. after each complete iteration) times acc_fac. *
+     * THIS ROUTINE ASSUMES THE OCCUPANCY VALUES IN RR_NODE ARE UP TO DATE.             */
 
-    int occ, capacity;
+    int overuse;
```
kmurray, line 434: This seems to only be used in the loop body. Probably best to declare it in the loop scope (also avoids potential future bugs related to uninitialized use).

vpr/src/route/route_common.cpp
```diff
@@ -414,57 +412,60 @@ void pathfinder_update_path_cost(t_trace* route_segment_start,
...
+        overuse = route_ctx.rr_node_route_inf[inode].occ() - device_ctx.rr_nodes[inode].capacity();
 
-        if (occ > capacity) {
-            route_ctx.rr_node_route_inf[inode].acc_cost += (occ - capacity) * acc_fac;
-            route_ctx.rr_node_route_inf[inode].pres_cost = 1.0 + (occ + 1 - capacity) * pres_fac;
-        }
-
-        /* If occ == capacity, we don't need to increase acc_cost, but a change    *
-         * in pres_fac could have made it necessary to recompute the cost anyway.  */
+        if (overuse > 0) {
```
kmurray, line 442: Is there ever a scenario where overuse < 0 could occur (which wasn't a bug)? If not probably good to put a VTR_ASSERT_MSG() to that effect.

vpr/src/route/route_timing.cpp
```diff
@@ -1608,49 +1532,6 @@ static bool early_exit_heuristic(const t_router_opts& router_opts, const Wirelen
...
-
-                total_overuse += overuse;
-                worst_overuse = std::max(worst_overuse, size_t(overuse));
-            }
-        }
-    }
-
-    return OveruseInfo(device_ctx.rr_nodes.size(), overused_nodes, total_overuse, worst_overuse);
-}
-
```
kmurray, line 1653: It looks like the new version of this code has lost the incremental approach used here? Is that intended?

vpr/src/route/route_common.cpp
```diff
@@ -414,57 +412,60 @@ void pathfinder_update_path_cost(t_trace* route_segment_start,
...
+     * sharing a resource right now (i.e. after each complete iteration) times acc_fac. *
+     * THIS ROUTINE ASSUMES THE OCCUPANCY VALUES IN RR_NODE ARE UP TO DATE.             */
 
-    int occ, capacity;
+    int overuse;
     auto& device_ctx = g_vpr_ctx.device();
     auto& route_ctx = g_vpr_ctx.mutable_routing();
+    size_t overused_nodes = 0, total_overuse = 0, worst_overuse = 0;
 
     for (size_t inode = 0; inode < device_ctx.rr_nodes.size(); inode++) {
```
kmurray, line 439: It looks like this iterates through all nodes in the graph: O(n).

Can this be done incrementally (like the previously updated overuse code which only looked at the modified nodes?

vpr/src/route/route_timing.cpp
Bill-hbrhbr: While calculating the new acc_cost, we need to find all the overused nodes anyway. So what I decided to do is to merge these two functions together to avoid repeated traversals.

kmurray: That makes sense, and will probably help. But I think we can probably make the later code incremental too.

vpr/src/route/route_common.cpp
Bill-hbrhbr: same thing as above. I tried the incremental approach in another PR. Turns out that going through the traceback linked lists of rerouted nets are slower. I didn't really standardize my QoR testings (running them under the same load) so our previous results might be misleading.

Bill-hbrhbr: Oh now I recall, originally we went through the traceback linked lists of all the nets. The more efficient way is to simply go through all the RR nodes. It is O(n) but it's faster according to my new QoR testing results, which were produced by running compared versions simultaneously.

PR #1355, opened 2020-06-10
Post routing cluster block pin fix-up

tangxifan:
Add codes to fix up the net swapping due to routing optimization when logic equivalent pins are defined in `pb_type`.
The fix-up is called when routing is finished successfully.

#### Description

- Add post routing cluster block pin fix-up. The results are stored in the `ClusterContext` without modifying the original `clb_nlist`.
- Add post routing net annotation to `rr_nodes`. The results are stored in the `RoutingContext`. The modification is required as it is fast to create such annotation for all the nets and ``rr_nodes``. This database is frequently inquired by the post-routing fix-up function. This can avoid frequent searches on the `rr_graph` and ``ClusteringContext` data structures.
- Add util functions to find physical tile pins with given capacity offset to the `vpr_utils.cpp`. The modification is required because the old function `get_physical_tile_pin()` does not consider the `z_offset` in any sub tile's capacity when finding the pin index. This will cause a wrong pin index is founded when multiple sub tiles with `capacity > 0` are defined under a physical tile.

#### Related Issue

To address the [issue](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1347).

#### How Has This Been Tested?

- This feature has passed existing regression tests.
- New test cases should be added as planned in the [issue](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1347).

#### Types of changes

- [x] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [X] I have added tests to cover my changes
- [X] All new and existing tests passed

PR #1448, opened 2020-07-22
RR graph fix up

tangxifan:
Fix-up the bug in the Routing Resource Graph which occurs when an input/output pin locates on multiple sides.

#### Description

Modify the rr_graph builder to
- Allocate only a unique rr_node for those input/output pin locates on multiple sides.
- The unique rr_node can be indexed through the fast look-up on different sides.
- relax some constraints in check rr_graph to support the fix-up on rr_graph

#### Related Issue

This pull request is a splitted pull request from #1355 to address the bug found in rr_graph

#### Motivation and Context

Current rr_graph builder allocates individual rr_node for the input/output pins locating on multiple sides.
This causes that router can map different nets to these rr_nodes which corresponds to the same pin.
In such case, the routing results are illegal.
This bug is observed mainly in the Stratrix4 architecture but not limited to it.

Passed regression test with QoR failure. Expect to fix the architecture XML to meet QoR

#### Types of changes

#### Checklist:


tangxifan:
Hi @kmurray @vaughnbetz 
   I just created this pull request by splitting the commits on the rr_graph fix-up here.
   I believe this could provide a clean start for us to finish the rr_graph patching.

tangxifan:
Hi all,
   To help us understand which part of the architecture causes routing congestion, I have checked the log files of the four failed Titan benchmarks.
   The intermediate file sizes are typically very large, ranging from 0.9GB to 6.8GB. 
   After compression, it definitely exceeds the 10MB limit in file uploading.
   The .net file, which takes most of diskspace.
   Therefore, I attached here the log files, intermediated files of two benchmarks, which I think is small enough but very representative.
  
- neuron: the example benchmark in the Titan tutorial, which passed titan flow-run.
- openCV: the smallest benchmark that fails titan flow-run. I see in the log file (vpr_stdout.log) there is 1 node after 400 routing iterations. The 1 node remains in many iterations. Therefore I believe this is an example to look into, through which we can spot where the congestion coming from.

   
[neuron_stratixiv_arch_timing.blif.tar.gz](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4968287/neuron_stratixiv_arch_timing.blif.tar.gz)

[openCV_stratixiv_arch_timing.blif.tar.gz](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4968303/openCV_stratixiv_arch_timing.blif.tar.gz)

   I also attached the vpr log files of the other 3 failures benchmarks here.
   
[failed_titan_benchmarks_vpr_logs.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4968320/failed_titan_benchmarks_vpr_logs.zip)

vaughnbetz:
Thanks a lot Xifan.

I took a look through the logs, and it is interesting and encouraging.

circuit                     overused_rr                  re-routed nets           re-routed conn       routing iteration
opencv: 	              1                                   6                                14                            400
bitcoinminer:         10                                 27                              63                            400
directrf                   2                                   21                             125                          400
gaussianblur	       3394	                            8036                         39607	                     43 (qave up)

The 3 new failures are clearly highly localized congestion.  I imagine input or output pins (maybe OPINs, given the number of nets getting ripped up and rerouted).  

1. Xifan, if you can get data on what type (IPIN or OPIN, block type, pin #) of rr-nodes are overused at the end of those first three routes (opencv, bitcoinminer, directrf) that would be very useful in narrowing this further.  The good news is that it looks like some IPINs or OPINs are too hard to reach (or have too many signals they need to drive) so it's a localized problem that hopefully we can fix in the .arch file.

2. Xifan, if you can upload the vpr.out files for the 4 circuits above and neuron from the run on the master (no rr-graph change) that would also be very helpful, so we can directly compare router congestion resolution output.

vaughnbetz:
There is more discussion of the rr-graph changes and their impact in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1355

vaughnbetz:
Ideas:
#0: add code to print out exactly what nodes are overused at the end of routing if we have less than N (N defaults to 20, command line option) overused nodes. (Bill).
#1: limit the cluster output count below 100% (say to 80%).  Command line option, could change auto meaning if this helps. (Xifan?)
#2: change from instance equivalence on the logic block outputs to full equivalence (overly flexible, but easier to understand).  We do instance equivalence but without duplication right now, and we don't model the 2:1 muxes on the outputs. (Xifan?)
#3: Could change the Fc,out values (make 2x higher) to model output 2:1 muxes (Xifan?)
#4: Add input pin equivalence on DSP and RAM blocks in groups of 36 pins (Matches Stratix IV).  Kevin's recommendation is to do this with a wrapper / extra level of hierarchy where we have groups of equivalent pins that go into a complete (crossbar) and then drive the current pins.  (Mohamed)
#5: Successful changes from above should go into the Titan arch generation script which is currently in the Titan repo (check with Helen). (Mohamed)

The DSP and RAM block model is believed to be overly restrictive right now; just was easier and never caused problems before.

tangxifan:
Hi @vaughnbetz 
   - I attached the vpr.out log files of the 4 failed titan benchmarks using the current master.
   
[titan_master_vpr_out.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4973268/titan_master_vpr_out.zip)

   - Regarding to the overused node print-out, it seems that there is some overlap with Bill's work. Should I participate or I should merge Bill's branch?
   - Regarding to the cluster output utilization, do you expect to modify the flow-run parameters only for the titan regression tests but also for the strong regression tests which use the titan architecture?
   - I think using full equivalence for LAB outputs is reasonable as the <complete> crossbar is already defined in the architecture XML. In this case, the LAB outputs have literally full equivalence. But I am curious why it was defined as `instance` equivalence before. What is difference between the `instance` and `full` when building rr_graph source nodes? (Sorry, I rarely used `instance` but I am familiar with `full`).

Here is the XML codes of the titan arch. related to the LAB output connection in the current master. It is a complete crossbar.
```
         <!-- ALM Outputs: directly drive global routing -->
          <complete input="alm[9:0].data_out[1:2] alm[9:0].data_out[4:5]" name="LAB_dataout" output="LAB.data_out"/>
```
   - As for the 2:1 muxes, I did not their definition in the arch XML and I think Stratix4 does not have any. Do you expect me to add these muxes? If we increase the Fc_out values, it means that we move the local routing multiplexers to the switch blocks. It may increase the routing area. 

  -  I did not see any crossbars for the DSP inputs. Here is the XML codes of the titan arch. related to the DSP data input connection in the current master. We have only direct connections.
```
          <direct input="DSP.data_in[287:144]" name="data_in_bot" output="half_DSP[1].data_in">
            <delay_constant max="239e-12" min="132e-12" in_port="DSP.data_in[287:144]" out_port="half_DSP[1].data_in"/>
          </direct>
          <direct input="DSP.data_in[143:0]" name="data_in_top" output="half_DSP[0].data_in">
            <delay_constant max="239e-12" min="132e-12" in_port="DSP.data_in[143:0]" out_port="half_DSP[0].data_in"/>
          </direct>
```
      Adding the input crossbar will improve the routability but should we recalculate the DSP block area? These crossbars are very area consuming. Similar question for the RAM blocks.

    Regarding to the methodology, do you expect to apply all the modifications on my branch (in this pull request) or in separated branches and then merged together?

vaughnbetz:
@Bill-hbrhbr Adding Bill. Xifan, lets get Bill to make the overused rr-node print out feature.  I don't want to overload you too much!
@helen-23 Also adding Helen since she is likely to get involved with the Titan architecture changes.
1. Experiment 1: instance equivalence is less flexible than full equivalence, since you can't have signals go to multiple outputs (at least not without logic duplication, which we don't support).  Sounds like that's an easy one, so if you can try changing the output equivalence to "full" and run to get a comparison that would be great Xifan.
2. Experiment 2: Stratix IV actually has 36:1 crossbars on the inputs to DSP blocks and RAM blocks. I talked to Kevin today, and it wasn't intentional to leave them out -- we just never put them in through forgetting / running out of time.  It didn't seem to cause many problems, but now I think it is.  So the proper thing to do would be to group the inputs into 36:1 groups with input equivalence and crossbars, for both DSP and RAM blocks.  If that is a lot of work, we could do a quick-and-dirty experiment with all the inputs to a RAM and DSP made equivalent.
3: Experiment 3: You are right that increasing Fc,out will increase area, but Stratix IV does have 2:1 muxes between the LE outputs and the routing drivers.  Since we have a full crossbar already on the logic block outputs we shouldn't add anymore flexibility there.  But for the RAM and DSP blocks we should experiment with doubling Fc,out as it would better approximate having a 2:1 mux between the outputs and the routing drivers.  My recollection is that those 2:1 muxes are in those blocks as well, but I am not 100% sure.

Based on the results of those, we could decide which architecture changes to commit (or to combine some and do an experiment 4).  

Methodology:  Xifan, if you can make the architectures for exp 1 - 3 and run them on your branch, and also pass them to Helen to run on the master (or if it is easier you can also run them on the master) that would be great.  I'd like to get QoR data for all those items and then we can decide what to change in the architecture.  Helen has some new circuits that she's getting into the Titan flow, and at least one of those showed some strange routing to RAM blocks on the *master* so I think the lack of input pin equivalence on RAM blocks may already be an issue (maybe the current Titan designs don't have that many RAM blocks?).

vaughnbetz:
Also, experiment 0.5 (before 1 above) is to play with various options for the cluster input and output pin utilization limits.
--target_ext_pin_util has the option.  Right now we're using auto, which limits the logic block to 80% input usage when possible (with no other limits).  Other interesting cases:
 - logic block has input and output limits of 0.8
 - all blocks have input and output limits of 0.8

vaughnbetz:
Filed the reporting overused rr-node issue to Bill: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1453

tangxifan:
Hi @vaughnbetz 
   I just finished the experiment 0.5, where I set the LAB input and output utilization rate to 0.8.
   We have two benchmarks `directrf` and `opencv` passing the routing but a new benchmark `sparc_T2` failed in routing.
   I attached the zip file which include

- A spreadsheet that compares current QoR to the golden on master
- vpr.out of the new passed and failed benchmarks
- A summary of QoR check failure from `parse_vtr_trask.pl`

[titan_LAB_io_util0.8.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4989982/titan_LAB_io_util0.8.zip)

tangxifan:
Hi @vaughnbetz 
I havefinished another experiment 0.5, where I set all the logic block input and output utilization rate to 0.8 (--target_ext_pin_util 0.8,0.8).
This time, we have a benchmark opencv passing the routing but a new benchmark sparc_T2 failed in routing.
I attached the zip file which includes

A spreadsheet that compares current QoR to the golden on master
vpr.out of the new passed and failed benchmarks
A summary of QoR check failure from parse_vtr_trask.pl

[titan_util0.8.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4996291/titan_util0.8.zip)

tangxifan:
Hi @vaughnbetz
   The experiment 1 has been finished. After I used `full` equivalence to LAB outputs, all the titan benchmarks passed flow-run but have some QoR failures. A short summary on the QoR:

- The QoR failures are due to some improvements due to the use of `full` pin equivalence.

- Routing area is reduced by 5% on average

- Almost no impact on critical path delay

- Runtime overhead is still there. For example, the number of routing iterations is almost doubled.

    See detailed in the attached report.
[LAB_output_equivalent.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5002902/LAB_output_equivalent.zip)

vaughnbetz:
This looks great.  full instance equivalence definitely looks like a good thing to turn on.

vaughnbetz:
Just looked through the 0.8 input and 0.8 output pin utilization target experiment.  It didn't change much:  +2% cluster count, +2% placer estimated critical path, +5% routed wirelength, +9% routed critical path delay, +446% route time (probably really really +300% since machine load looks higher due to place time).  All comparisons to the master.

vaughnbetz:
I've created a spreadsheet that summarizes the 4 runs so far.  LAB output equivalence with the fixed rr-graph is very close to quality parity with the master (with no rr-graph fix).  Some metrics are slightly worse, some slightly better.  I'll keep this spreadsheet up to date with future results so we can track combinations might be promising, but output equivalence definitely looks good.
[titan_arch_xifan.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5004371/titan_arch_xifan.xlsx)

tangxifan:
Hi @vaughnbetz I am trying to re-run the titan tests using the latest master with the rr_graph fix-up (no arch. modification, only rr_graph modification), so that we can report overused nodes. But I detect a bug in the flow-run script, where the max_route_iterations is reduced from 400 to 150. This is critical to my tests because as you see in the spreadsheet, the required number of routing iteration is more than doubled. As a result, I see most of the benchmarks failed this time. I would like to learn your advice about how to proceed. 
- Should I wait for a feedback on the bug?
- Should I revert back to an old master and perform experiments 3 and 4? In this case, we may not have overused node reports.

tangxifan:
Hi @vaughnbetz 
    I have merged the latest master to my branch and run titan benchmarks with the `generate_overused_node_report` feature. Note that in this run, I am using the original titan architecture XML without any modification, and without any change on logic block pin utilization.
Still, there are four benchmarks failing, which are 
    
- gaussianblur: failed as usual
- sparcT1_core: was passing before
- sparcT2_core: was passing before
- bitcoin_miner: failed as usual

There are two benchmarks: `directrf` and `opencv` passed in this run.
So I am wondering if there is any router improvement from recently merged pull requests. 

A brief summary of the performance shift:

-  total wirelength is reduced by 4% on average (considering only passed benchmarks)
-  critical path delay is increased by 11% on average (considering only passed benchmarks)

Full details, including vpr.out, overused node reports and a spreadsheet, can be found in the attachment.
[titan_overused_nodes_report.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5023679/titan_overused_nodes_report.zip)

I will start trying experiment 3 and 4 this week.

vaughnbetz:
Thanks Xifan.  All the routing congestion is on OPINs and IPINs, with almost all the congested OPINs being driven by LABs.  So instance equivalence with the new (properly 2-sided) rr-graph is definitely a problem.  We didn't intentionally make any quality changes to the router, so I think it's random that 2 circuits passed that used to fail, while two others fail.  The failures of 3/4 designs have hundreds to 1800 of failures this time; only bitcoin_miner is very close to passing.
I've added this to the overall tracking spreadsheet.
[titan_arch_xifan.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5028921/titan_arch_xifan.xlsx)

Results are overall similar to what we had on / vs. the prior master. One strange thing: placement wire use went up 5% with the new rr-graph in this run (hasn't shown up before).  Hopefully that is random; I don't have any explanation for it.

tangxifan:
Hi @vaughnbetz 
  Thanks for the advice. I just rerun the experiment 1 after merging the master to my branch.
  You are right. There is randomness in the results. In this run, all the benchmarks passed and we have only two QoR failures:
```
stratixiv_arch_LAB_pin_eq.timing.xml/LU_Network_stratixiv_arch_timing.blif/common placed_CPD_est: golden = 6.64648 result = 9.70124
[Fail] 
 stratixiv_arch_LAB_pin_eq.timing.xml/LU_Network_stratixiv_arch_timing.blif/common placed_setup_WNS_est: golden = -6.64648 result = -9.70124 
```
Overall, the QoR is similar to our previous run, where 
- total routing wirelength is increased by 1% on average
- routing critical path delay is increased by 2% on average

I will wait for the architecture file from Helen and then continue my experiments 3 and 4.

Full results are uploaded here FYI:
[titan_LAB_output_eq_generate_overused_nodes.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5030042/titan_LAB_output_eq_generate_overused_nodes.zip)

vaughnbetz:
Added to summary spreadsheet: [titan_arch_xifan.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5030527/titan_arch_xifan.xlsx)

Results look good for this output equivalence change (again).  One strange thing (presumably unrelated): placement wire has gone up 8% vs. the master, but routed wire is only up 1%.  Gaussian blur shows a very strange placement estimated wire in the branch vs. the master:
Placement wire for Gaussian blur on rr-graph fixed case (4e7 placer wire) vs. 3.2e7 achieved by router in that code, and master estimate for placement wire is 2.3e7.  Xifan can you take a look at the placement wire estimates for gaussian_blur and make sure they're being parsed out correctly?  Any change in global signals that are ignored during placement or some such?

tangxifan:
Hi @vaughnbetz 
   I doubled checked the `vpr.out` for the benchmark `gaussianblur`. The placed wirelength is indeed 4e7, showing that the report parsing is correct.
`BB estimate of min-dist (placement) wire length: 39836016`

   Attached  the `vpr.out` FYI.
[gaussian_blur_vpr.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5031083/gaussian_blur_vpr.zip)

vaughnbetz:
Thanks Xifan.  I went through the logfile and you're right that it was parsed out correctly.  Nothing else jumps out as wrong, except the placement estimated wirelength doesn't line up with the routed wire or the usual value.  Probably unrelated to what you're doing, but strange.

tangxifan:
Hi @vaughnbetz 
  I have finished the experiment 3, where I modified the DSP inputs of the Titan architecture as follows:

- The `data_in[287:0]` port of DSP is splitted into two ports `data_in_top[143:0]` and `data_in_bot[143:0]`, where each ports is set to be logic equivalent, as show below:
```
      <input name="data_in_top" num_pins="144" equivalent="full"/>
      <input name="data_in_bot" num_pins="144" equivalent="full"/>
```

- The direct connection inside the DSP `pb_type` between the `data_in[287:0]` and the `data_in` of `half_DSP` is changed to complete connection, as shown below:

```
          <complete input="DSP.data_in_bot" name="data_in_bot" output="half_DSP[1].data_in">
            <delay_constant max="239e-12" min="132e-12" in_port="DSP.data_in_bot" out_port="half_DSP[1].data_in"/>
          </complete>
          <complete input="DSP.data_in_top" name="data_in_top" output="half_DSP[0].data_in">
            <delay_constant max="239e-12" min="132e-12" in_port="DSP.data_in_top" out_port="half_DSP[0].data_in"/>
          </complete>
```

- The pin location of tiles and direct connections between tiles have been updated for the splitted `data_in` ports.

After this change, the Titan regression test produces the following results:

- Benchmark `gaussianblur` and `bitcoin_miner` still failed.
- Benchmark `directrf`, `spartT2_core` and `openCV` now passed.
- Critical path delay has been increased by 39% on average.
- Routing wirelength is increased by 3% on average.

My understanding on the critical path delay degradation is that the routing congestion of LAB outputs has been alleviated by the equivalent DSP inputs. But the routing nets from LAB outputs squeeze the routing area of DSP inputs. Therefore, the routing congestion can be regarded as `shifted` to the DSP blocks. As a result, the routing is close to those achieved under minimum routable channel width.

I also notice that some passed benchmarks have critical path delay increased as well. My understanding is that there is limited multiple side pins for the DSP `data_in` port. When pin equivalence is set to true, the routing nets are uniquified. As a result, a net, which was mapped to multiple pins located on different sides, is now mapped to one pin located on one side. This imposed some limitation when routing these nets.

I attached my modified architecture file, spreadsheet of the results, and log files of failed benchmarks FYI.
[titan_dsp_input_eq.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5079267/titan_dsp_input_eq.zip)

tangxifan:
A minor problem detected when update the titan architecture file, w.r.t. to the `data_in` port location in DSP tiles.

LINE 4588 and LINE 4597 at the original Titan architecture file:
```
      <loc yoffset="2" side="right">DSP.data_in[144:108] DSP.control_in[8:6] DSP.scan_a_in[8:7]            DSP.data_out_top[71:54] DSP.scan_a_out[7:6] DSP.signal_out[2:2] </loc>
      <loc yoffset="0" side="right">DSP.data_in[287:252] DSP.control_in[20:16] DSP.scan_a_in[17:15] DSP.clk[3] DSP.data_out_bot[71:54] DSP.scan_a_out[15:14] DSP.signal_out[5:5]</loc>
```
You can see that the `data_in[144]` which belongs to the bottom half  `data_in` port is placed along with the top half `data_in[143:108]`. This is not consistent with the comments `the 'top' half o the DSP has pins connected at offset 2 and 3`.
I would like to learn your opinion. It may not create routing problems but it is good to clarify.

tangxifan:
Hi @vaughnbetz ,
   I have tried the experiment 4.1 where I set pin equivalence to the `data_out` port of M144K memory.
   This is a low hanging fruit for me because a complete crossbar for the `data_out` is already defined.
   The results are interesting, being similar to our experiment 3 results.

- Benchmark `gaussianblur`  and `bitcoin_miner` and `sparcT2_core` still failed
- A new benchmark `sparcT1_core`  failed.
- Benchmark `directrf` and `openCV` now passed.
- The critical path delay has been increased by 10x and routing wirelength is increased by 7% on average.

   I am studying the sparse crossbar for M9K and M144K memories but I see it is too sparse (almost every `data_in` pin has different connectivity). I may need more time to see how to group pins into equivalent ports.

   All the reports, spreadsheet and architecture XML is attached FYI.
   
[titan_M144K_output_eq.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5081429/titan_M144K_output_eq.zip)

vaughnbetz:
@tangxifan: thanks for the experimental results.  They are surprising though, as pin equivalence like this should make routing easier and should not hurt the critical path!
I am wondering if the router is confused by equivalent pins that are spread out on these large blocks, or if the pin locations were changed as a side effect of adding equivalence.  In (Altera's) Stratix IV, the equivalent pins are always in a single (x,y) location (they are equivalent in groups of about 36), as that lets them all feed the same crossbar or partial crossbar, and it also lets the router know where they are.
Perhaps the easiest way to check is to do a similar input pin equivalence test for the M9K as you did for the DSP blocks.  Since the M9K block is all at a single (x,y), if there was a problem with the pin locations in the DSP input equivalence test it shouldn't show up in the M9K test.

vaughnbetz:
Hi Xifan,

I looked at the code that creates SINKs & SOURCEs in the rr-graph:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/4a7c900d9b6e671203a7e98dddbedb7dc36790f3/vpr/src/route/rr_graph.cpp#L1409

and the map_lookahead code:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/4a7c900d9b6e671203a7e98dddbedb7dc36790f3/vpr/src/route/router_lookahead_map.cpp#L293

and there is a problem for large blocks.  All SINKs and SOURCEs of a large block (width and/or height > 1) are given the x,y coordinate of its lower-left corner.  The map lookahead also simply looks for the target location of a route as being the (x,y) of the SINK node it is heading toward.  For large blocks this means we're always heading toward the lower left corner, regardless of where the pins on the block are.

I think we need to change the rr-graph code to put each SINK or SOURCE at the location that is the average of the (x,y) locations of the pins it connects to.  That way, if we make proper SINKs / SOURCEs per pin we'll put the SINKs and SOURCEs at the right locations.  And if we have a group of logically equivalent pins, we'll again put the SINK or SOURCE at basically the correct position.

I think this is a strong candidate for the poor performance with equivalent DSP block input pins, as the router wouldn't have known where it was going.  I'm not sure if it also causes the poor OPIN equivalence behaviour of the M144K; it's not as obvious to me that having a bad SOURCE location would cause problems, but fixing this is a good idea for SOURCES too since it will be more logical in any case.

tangxifan:
@vaughnbetz This makes sense to me. The lookahead builder should have a more precise coordinate of the SOURCE and SINK nodes. On the other side, assigning a flexible coordinate to SOURCE and SINK may impact router or other clients, since it has been a convention to use the lowest corner (xlow, ylow). I will try to patch this and see if it cause any chain effects.

vaughnbetz:
Thanks Xifan.
In terms of landing this PR, I suggest we get your rr-graph fix and the fully equivalent OPINs for LABs in soon (but maybe not this week as there are lots of other PRs in flight).  Then the moving the SOURCE and SINK locations for large blocks to more appropriate locations could be a separate PR, and if we find good results with equivalent IPINs for some other blocks we can add them to the Titan architecture capture in a 3rd PR.

tangxifan:
@vaughnbetz Thanks for the suggestion. No problem from my side.

vpr/src/route/rr_graph2.cpp
```diff
@@ -1082,33 +1082,102 @@ static void load_block_rr_indices(const DeviceGrid& grid,
...
+                     * we just create a rr_node and make it indexable on these sides
+                     * As such, we can avoid redundant rr_node to be allocated
+                     * and multiple nets to be mapped to the pin
+                     *
+                     * Considering that some pin could be just dangling, we do not need
+                     * to create a void rr_node for it.
+                     * As such, we only allocate a rr node when the pin is indeed located
+                     * on at least one side
+                     */
+                    if (true == assigned_to_rr_node) {
```
HackerFoo, line 1178: The condition should just be `assigned_to_rr_node`.

tangxifan: Hello, I am sorry I do not get the exact issue. Would you mind providing more insights? I will make a change.

HackerFoo: `if (assigned_to_rr_node)` is more clear than `if (true == assigned_to_rr_node)`.

There isn't a good reason to compare with a boolean literal.

tangxifan:
@litghost I have fine-tuned the architecture `k6_N10_mem32K_40nm_i_and_o.xml` by 
- forcing 'EMPTY' tiles on the left and right sides 
- using 'clb' tiles to fill the empty locations in the columns of mem32K and DSP
This improves the routability of the architecture without conflicting what you want to test. 
Please let me know if this is fine for you.

FYI, I attached the grid layout (from VPR GUI) after architecture modification here.
- Global view of the architecture: 
image

- A close view on the architecture:
   - You can see that CLBs fill the empty locations in the columns of mem32K and DSP
   - CLB start from x=1, where x=0 is now filled by EMPTY tiles.
image

   - Routing architecture surrounds the CLBs as a regular FPGA fabric.
image

vaughnbetz:
Hi Xifan,

Mixing clbs and RAMs in a column like this is hard to lay out, so it's not normally done in commercial FPGAs.  I don't object to testing it in this architecture, but I think you should put a comment at the top of the file explaining what this architecture does that is different and what it is intended to test.  (Nobody did that for the dedicated input/output pads that it appears to have, so that's good to list in that comment too).

I don't think this architecture is used much in regtests and not in any automated QoR tests (which is good in my opinion, as with these changes it will be a bit more of an unusual architecture).  Is that right?

tangxifan:
@vaughnbetz I agree on this. It will challenge the physical design a lot. This is indeed rare in commercial FPGAs as well as any tape-out we have done. Usually, we properly size the grids so that there are no such EMPTY holes in the DSP/RAM columns. These holes are seen only during the architecture exploration. Therefore, I have added comments to the head of the architecture XML in order to clarify these points. If I miss anything, feel free to let me know.

vaughnbetz:
Thanks Xifan; the new comment is very clear.

tangxifan:
Hi all, I have address all the comments so far. The regression tests now passed. Please let me know if it is good to merge.

litghost:
Arch change looks fine to me

vaughnbetz:
Looks good.  

We can merge this now, but 
1. I have a minor comment to add to the arch file (can go in a different PR).
2. I think it would be good to run a final Titan QoR sweep of the current master vs. this new PR (is about to become the master) so we have it written down how QoR changed, and hence know how (roughly) to adjust expectations if someone has data compared to the old rr-graph and master.  I know you measured that a couple of months ago Xifan, but I think it's important enough to measure again s we commit this.

I suspect we can merge in advance of 1 & 2 though but should do them ASAP after the merge.  @tangxifan does that make sense?

tangxifan:
@vaughnbetz 
- No problem for the comments on the arch file. I can do it now. 
- I will rerun the titan regression test and collect the QoR. Do you want me to update the golden results for titan? Or just submit a spreadsheet comparing the QoR difference in this PR?

It will take ~1 day to finish and I think we can do it before merging.

vaughnbetz:
Thanks Xifan.  I think it would be good to upload a spreadsheet of Titan QoR first, and then if all looks good, update the golden and merge.

tangxifan:
Hi @vaughnbetz 
   I have finished the regression test for titan benchmark.
   In the attached spreadsheet, I compare the QoR results of the rr_graph fix-up branch to the golden results of current master.
   You can see that there is zero difference on the averaged critical path delay and an average of 5% improvement on the area.

[titan_parse_results_RRG_fixup_vs_master.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5234100/titan_parse_results_RRG_fixup_vs_master.xlsx)

vaughnbetz:
Thanks Xifan! This looks great, and I don't think we need to do an update golden as everything was close enough anyway.  I think this is ready to merge.  Let me know if you disagree, otherwise I'll merge it.

tangxifan:
@vaughnbetz Thanks for the advice. I am nothing against it.

vaughnbetz:
Excellent -- glad to land this one!

PR #1450, opened 2020-07-23
Add PlacerSetupSlacks interface

Bill-hbrhbr:
#### Description
Add an interface for a 1-to-1 mapping between CLB pins and setup slacks from the timing analyzer. Refactored place.cpp (replaced recompute_criticalities with three new routines) so that setup slacks and criticalities can be updated together or separately. 

#### Types of changes

#### Checklist:


Bill-hbrhbr:
The current question is that the three new routines in place.cpp:
1. update_setup_slacks
2. update_criticalities
3. update_setup_slacks_and_criticalities
can be merged into a single routine by utilizing some boolean flags.

The tradeoff is between code duplication and branch prediction. So far, only **update_criticalities** is called, and it is called once in a while (so it shouldn't be that hot of a function). 

I'm inclined to merge these three routines into a single one, but in the future, we might frequently call these update routines in the placement quench part of the code.

vpr/src/place/timing_place.h
```diff
@@ -81,6 +81,70 @@ class PlacerCriticalities {
...
+ * connections/AtomPinIds returned by SetupTimingInfo.
+ *
+ * The setup slacks of individual connections can then be queried by calling the 
+ * setup_slack() member function.
+ *
+ * It also supports iterating via pins_with_modified_setup_slack() through the 
+ * clustered netlist pins/connections which have had their setup slacks modified by 
+ * the last call to update_setup_slacks().
+ */
+class PlacerSetupSlacks {
```
vaughnbetz, line 108: Explain that these are raw slacks (can be negative, based on the timing constraint specified by the user or the very difficult automatically set timing constraints if no user-specified one exists).

vaughnbetz:
It is possible that this would be cleaner with one function that always updated criticalities and slacks.  Once you've played with the algorithm some, you could test if always updating both criticalities and slacks in one routine has a significant cost or not.

Bill-hbrhbr:
VTR benchmarks @ commit 112bde5: https://drive.google.com/file/d/1JRfVQWIme2yldLiOQBPWuD9R2-akIrl_/view?usp=sharing

vpr/src/place/place.cpp
```diff
@@ -119,6 +119,16 @@ struct t_annealing_state {
     int move_lim;             // Current move limit
 };
 
+struct t_placer_timing_update_mode {
```
vaughnbetz, line 122: You'll need a more detailed comment explaining each of these flags and how to set them.

vpr/src/place/place.cpp
```diff
@@ -321,15 +333,18 @@ static void update_move_nets(int num_nets_affected);
 static void reset_move_nets(int num_nets_affected);
 
 static e_move_result try_swap(float t,
+                              float crit_exponent,
```
vaughnbetz, line 336: The argument list is gettting very long. Could some of these variables be grouped in a struct and a pointer to that be passed around?

vpr/src/place/place.cpp
```diff
@@ -347,14 +362,17 @@ static int check_placement_consistency();
 static int check_block_placement_consistency();
 static int check_macro_placement_consistency();
 
-static float starting_t(t_placer_costs* costs,
+static float starting_t(float crit_exponent,
+                        t_placer_timing_update_mode* timing_update_mode,
```
vaughnbetz, line 366: Same comment on lots of variables.  Maybe time to group some into a struct?

vpr/src/place/place.cpp
```diff
@@ -821,31 +873,42 @@ void try_place(const t_placer_opts& placer_opts,
...
+                                      &outer_crit_iter_count,
+                                      place_delay_model.get(),
+                                      placer_criticalities.get(),
+                                      placer_setup_slacks.get(),
+                                      pin_timing_invalidator.get(),
+                                      timing_info.get());
 
         state.t = 0; /* freeze out */
 
+        //Use setup slack analysis if the placer is timing driven
```
vaughnbetz, line 889: yes, this will need to be a new command line option.

vpr/src/place/place.cpp
```diff
@@ -994,26 +1074,86 @@ static void outer_loop_recompute_criticalities(const t_placer_opts& placer_opts,
...
 
-//Update timing information based on current placement by running STA to get new slacks,
-//and calculate updated criticalities and timing costs
-static void recompute_criticalities(float crit_exponent,
-                                    const PlaceDelayModel* delay_model,
-                                    PlacerCriticalities* criticalities,
-                                    ClusteredPinTimingInvalidator* pin_timing_invalidator,
-                                    SetupTimingInfo* timing_info,
-                                    t_placer_costs* costs) {
+static void initialize_timing_info(float crit_exponent,
```
vaughnbetz, line 1077: Somewhere in this file (options listing?  type or variable definition of slacks and criticalities?) you'll need to detail exactly what a criticality and a slack are, and the difference (criticality is relative between 0 and 1, slack is absolute and can be +ve or negative (unrelaxed, raw slack based on timing constraint I think).

vpr/src/place/place.cpp
```diff
@@ -994,26 +1074,86 @@ static void outer_loop_recompute_criticalities(const t_placer_opts& placer_opts,
...
+                                          costs);
+
+    //Initialize the data structure that stores committed placer setup slacks
+    commit_setup_slacks(setup_slacks);
+
+    //Don't warn again about unconstrained nodes again during placement
+    timing_info->set_warn_unconstrained(false);
+}
+
+//Update timing information based on current placement by running STA.
```
vaughnbetz, line 1121: Better to use Doxygen compatible comments for function-level comments which start with /** for a block comment or are /// for a single line.
It's an ongoing project to build more Doxygen docs.

vpr/src/place/place.cpp
```diff
@@ -994,26 +1074,86 @@ static void outer_loop_recompute_criticalities(const t_placer_opts& placer_opts,
...
+//Record the new slack information as well as calculate the updated
```
vaughnbetz, line 1122: Should comment on any data that's needed by this routine.  Location of block according to block.x,.y etc. and the pin_timing_invalidator?  Explain what is necessary before you call this routine.

vpr/src/place/place.cpp
```diff
@@ -1445,21 +1615,67 @@ static e_move_result try_swap(float t,
         //Update the block positions
         apply_move_blocks(blocks_affected);
 
-        // Find all the nets affected by this swap and update their costs
+        //Find all the nets affected by this swap and update their costs
```
vaughnbetz, line 1618: Nice comment update -- thanks!

vpr/src/place/place.cpp
```diff
@@ -1491,14 +1717,43 @@ static e_move_result try_swap(float t,
...
+            /* Reset the net cost function flags first. */
             reset_move_nets(num_nets_affected);
 
             /* Restore the place_ctx.block_locs data structures to their state before the move. */
             revert_move_blocks(blocks_affected);
 
+            if (place_algorithm == SETUP_SLACK_ANALYSIS_PLACE) {
+                //Revert the timing delays and costs to pre-update values
+                //These routines must be called after reverting the block moves
+                //TODO: make this process incremental
```
vaughnbetz, line 1731: Is this TODO still relevant or are they incremental already?  They look incremental (with the invalidator etc.).

vpr/src/place/place.cpp
```diff
@@ -1674,6 +1929,64 @@ static void update_td_delta_costs(const PlaceDelayModel* delay_model,
     }
 }
```
vaughnbetz, line 1931: Explain the purpose of this routine and what it does.

vpr/src/place/place.cpp
```diff
@@ -1674,6 +1929,64 @@ static void update_td_delta_costs(const PlaceDelayModel* delay_model,
...
+        ClusterNetId net = clb_nlist.pin_net(clb_pin);
+        int ipin = clb_nlist.pin_net_index(clb_pin);
+
+        if (proposed_connection_delay[net][ipin] != connection_delay[net][ipin]) {
+            //Delay has changed. Must invalidate this sink pin.
+            sink_pins_affected.push_back(clb_pin);
+        }
+    }
+}
+
```
vaughnbetz, line 1954: Explain this routine and the cost function.

vpr/src/place/place.cpp
```diff
@@ -1775,45 +2088,55 @@ static void comp_td_connection_delays(const PlaceDelayModel* delay_model) {
...
-                    proposed_connection_delay[net_id][ipin] = INVALID_DELAY;
-                    connection_timing_cost[net_id][ipin] = proposed_connection_timing_cost[net_id][ipin];
-                    proposed_connection_timing_cost[net_id][ipin] = INVALID_DELAY;
-                }
-            } else {
-                //This pin is a net sink on a moved block
-                VTR_ASSERT_SAFE(cluster_ctx.clb_nlist.pin_type(pin_id) == PinType::SINK);
+        connection_setup_slack[net_id][pin_index_in_net] = setup_slacks->setup_slack(net_id, pin_index_in_net);
+    }
+}
```
vaughnbetz, line 2107: Explain what this routine is doing.  Confirming the cached setup slacks match what you get if you ask the timing analyzer?  Also explain when it is valid to expect these to match (timing_update with any new delays hasn't happened since the cached setup slacks were created?)

vpr/src/place/place.cpp
```diff
@@ -1775,45 +2088,55 @@ static void comp_td_connection_delays(const PlaceDelayModel* delay_model) {
...
-        } /* Finished going through all the pins in the moved block */
-    }     /* Finished going through all the blocks moved */
+        }
+    }
+
+    return true;
+}
+
+/* Update the connection_timing_cost values from the temporary *
+ * values for all connections that have changed.               */
```
vaughnbetz, line 2124: Explain this is part of the process of commiting a move.

vpr/src/place/place.cpp
```diff
@@ -1840,31 +2163,21 @@ static void revert_td_cost(const t_pl_blocks_to_be_moved& blocks_affected) {
 //
 //Relies on proposed_connection_delay and connection_delay to detect
 //which connections have actually had their delay changed.
```
vaughnbetz, line 2165: Explain purpose of the routine (to figure out which connections have had their delays changed so we can incrementally timing analyze).  Do we need to call this after every committed move or every proposed move?  Explain a bit more about its use here.

vpr/src/place/place.cpp
```diff
@@ -2096,6 +2409,8 @@ static void alloc_and_load_placement_structs(float place_cost_exp,
         connection_delay = make_net_pins_matrix<float>(cluster_ctx.clb_nlist, 0.f);
         proposed_connection_delay = make_net_pins_matrix<float>(cluster_ctx.clb_nlist, 0.f);
 
+        connection_setup_slack = make_net_pins_matrix<float>(cluster_ctx.clb_nlist, std::numeric_limits<float>::infinity());
```
vaughnbetz, line 2412: Make sure there is a comment somewhere on what this data structure is and give an example of how to access it e.g. matrix of setup slacks on all connections indexed from [0..num_nets-1][1..num_pins_on_net-1].  (check if I got that right).
Can comment where the variable is declared or here, depending on what seems like a better spot.

vpr/src/timing/timing_util.cpp
```diff
@@ -579,6 +579,23 @@ float calculate_clb_net_pin_criticality(const SetupTimingInfo& timing_info, cons
     return clb_pin_crit;
 }
 
+//Return the setup slack of a net's pin in the CLB netlist
```
vaughnbetz, line 582: Explain that it assumes the timing analysis is up to date (has already been performed with an update call with the right delays).

vpr/src/place/timing_place.cpp
```diff
@@ -100,6 +107,77 @@ PlacerCriticalities::pin_range PlacerCriticalities::pins_with_modified_criticali
     return vtr::make_range(cluster_pins_with_modified_criticality_);
 }
```
vaughnbetz, line 109: Are some of these routines rewrites / replacements for routines in place.cpp?  I haven't checked line by line but some seem similar.

vaughnbetz:
Hi Bill,

The code overall looks clean.  I've embedded a bunch of requests for comments and feedbacks though.
Themes:

- Comment each function and new data structure explaining what it does, and how it is to be used (e.g. do delays and/or the slacks on the timing graph have to be up to date before calling specific routines)?
- There are some long argument lists (which predate you, but are getting even longer).  Can you factor into a pointer to a higher-level structure to shorten, and would that be useful?  This is a bit of a judgement call, but please take a look.
- It looks like you've moved some of your code to timing_place.cpp.  If you can move as much as possible into timing_place.cpp (or a new file) it would be good, as place.cpp is already too big so avoiding adding to it would be good.

Bill-hbrhbr:
Hi @vaughnbetz, I managed to reduce down place.cpp from ~3300 lines to ~2800 lines by moving routines and structures to other files (mainly annealing state variables and timing update routines), and that's with new/enhanced documentation (also in Doxygen style) on a lot of these routines/data structures.
I also shortened the argument list of annealing routines by combining annealing state arguments and costs structures. The file is definitely cleaner than before.
I will continue to update documentation, notably the differences between slacks and criticalities and how to use them. But I will, for now, try not to mess around with ```try_swap()``` or ```placement_inner_loop()```, since they are other PRs going on that may affect them.

libs/libvtrutil/src/vtr_vec_id_set.h
```diff
@@ -2,6 +2,7 @@
 #define VTR_SET_H
 
 #include <vector>
+#include <algorithm>
```
HackerFoo, line 5: Why was this added?

vpr/src/place/place.cpp
```diff
@@ -91,94 +95,106 @@ struct t_placer_statistics {
...
-    double cost;
-    double bb_cost;
-    double timing_cost;
-};
+/*******************************************************************************
+ * Below is a list of definitions of data structures declared as `extern` in   *
+ * place_global.h. These variables were originally local to the current file.  *
+ * However, they were moved so as to facilitate moving some of the routines    *
+ * in the current file into other source files.                                *
+ *******************************************************************************/
```
HackerFoo, line 106: It seems like they were exported rather than moved. This seems to expose a lot of details; why not wrap these globals into a singleton (global) object? This will make it easier to duplicate these data structures e.g. to allow parallelism.

vpr/src/place/place_util.cpp
```diff
@@ -29,3 +30,174 @@ static vtr::Matrix<t_grid_blocks> init_grid_blocks() {
...
+ * @brief Update the range limited to keep acceptance prob. near 0.44.
+ *
+ * Use a floating point rlim to allow gradual transitions at low temps.
+ */
+static void update_rlim(float* rlim, float success_rat, const DeviceGrid& grid) {
+    float upper_lim = std::max(grid.width() - 1, grid.height() - 1);
+
+    *rlim *= (1. - 0.44 + success_rat);
+    *rlim = std::max(std::min(*rlim, upper_lim), 1.f);
+}
```
HackerFoo, line 203: Missing newline

vpr/src/place/place_global.h
```diff
@@ -0,0 +1,34 @@
...
+extern vtr::vector<ClusterNetId, char> bb_updated_before;
+extern ClbNetPinsMatrix<float> connection_delay;
+extern ClbNetPinsMatrix<float> proposed_connection_delay;
+extern ClbNetPinsMatrix<float> connection_setup_slack;
+extern PlacerTimingCosts connection_timing_cost;
+extern ClbNetPinsMatrix<double> proposed_connection_timing_cost;
+extern vtr::vector<ClusterNetId, double> net_timing_cost;
+extern vtr::vector<ClusterNetId, t_bb> bb_coords, bb_num_on_edges;
+extern vtr::vector<ClusterNetId, t_bb> ts_bb_coord_new, ts_bb_edge_new;
+extern std::vector<ClusterNetId> ts_nets_to_update;
```
HackerFoo, line 34: Missing newline

libs/libvtrutil/src/vtr_vec_id_set.h
Bill-hbrhbr: After moving around the code, the compiler gave me an error saying ```error: 'sort' is not a member of 'std'``` in L79 of this file. That's why I added it.

Bill-hbrhbr:
Hi @HackerFoo, this PR got quite big since I did a clean-up of ```place.cpp``` this past few days. All of the new features I've added are contained in the 
1. ```try_swap()``` routine.
2. The new ```PlacerSetupSlacks``` class in ```timing_place.h/cpp```
3. and the ```update_setup_slacks_and_criticalities()``` in ```place_timing_update.h/cpp```.

My steps before refactoring ```place.cpp``` (also writing this down for future reference):
1. ```PlacerSetupSlacks``` is an imitation of the class ```PlacerCriticalities```, and is utilized to get setup slacks from Kevin's Tatum timing analyzer.

2. I changed the original ```recompute_criticalities()``` to ```update_setup_slacks_and_criticalities()``` so that both categories of values can be updated.

3. One problem now occurs. Originally, each time the timing graph is updated by calling ```timing_info->update()```, the criticalities are updated as well. These updates use incremental techniques so that they only checkout modified atom pins returned by the timing analyzer. However, now the timing graph might be updated only for the setup slacks (without having criticalities updated). On the next iteration of criticalities update, it cannot use the incremental technique anymore, since more pins will have been modified than the ones the timing analyzer currently returns. In this case, the criticalities have to be recomputed from scratch (going through each sink pin).

4. To resolve this issue, I created the ```t_timing_update_mode``` class to keep track of the slack/criticality update status. This structure resides in ```place_timing_update.h```. It has a very long and detailed documentation already, so I guess I will not paste it here.

5. Now with the timing stuff down, I need to implement the setup slack analysis during the placement quench stage. So I made a vpr option ```--place_quench_metric``` to ask the user whether or not to turn on the new analysis technique in the quench.

6. I modified the ```try_swap()``` routine so that, if using the slack analysis, it will do a timing_update to get slacks, analyze, and then decide whether or not to accept the move based on the slack improvement. If accepted, the timing update will be kept; if rejected, I revert the timing update so that the slacks in the timing graph returns to their original values. 
Note that these timing updates will not involve changing criticalities (hence I had the consideration for point #3 and #4), and I've made a lot of testings to make sure that the timing update reversion works.

I would also like to refactor before adding changes, but right now a lot of my team members are trying to change ```place.cpp```, so getting that into the master might make the work more complicated for others. However, I believe that I am fairly familiar with ```place.cpp``` so that I will be able to resolve merge conflicts rather easily after they get their PRs in.

If it is still inconvenient for you to check and make comments, I can open a draft PR at one of my intermediate commits (before the refactorization) so that you can see only the new features being added. The downside to that is that I haven't written/enhanced a lot of my comments by then. Might be harder to interpret, but might not with what I've written down here.

vpr/src/place/place.cpp
Bill-hbrhbr: I can wrap them into a single global structure and provide accessor functions for them (so that I don't have to declare them as extern), but I don't think there's a need to duplicate them, as they all serve different purposes and are only used by ```place.cpp`` once per VPR flow.

Bill-hbrhbr: After giving it more thought, I think I'll first leave it as it is in the current PR, and then refactor it into something similar to vpr_context.h in a new PR (something like g_place_ctx or g_place_structs). That will reduce the amount of clutter in the code.

Bill-hbrhbr:
VTR benchmarks @74d279c: https://drive.google.com/file/d/1n2Txadj3pXm-CGq6kbEPbdT6hh7RzRb4/view?usp=sharing
Titan benchmarks @74d279c: https://drive.google.com/file/d/1igBYhXExdT2Uq4JyFeIbWXojWf5NZlS8/view?usp=sharing

The results have slightly degraded somehow. I will run the benchmarks again (with -j # parallel) to see if this is a recurring thing since I highly doubt that it should happen.

On another note, this PR is failing the golden results of ```benchmarks/blif/4/apex4.blif``` and ```benchmarks/blif/4/tseng.blif``` for ```arch/bidir/k4_n4_v7_bidir.xml```. However, it doesn't fail -check_golden when I run it on Wintermute. So it's kind of strange.

vaughnbetz:
The QoR results (vtr and titan benchmarks) you link above look like a tie to me (a few metrics marginally up a few marginally down) except perhaps Titan placement time which is up ~10% while pack time is up 4% (so machine load may be a factor, but placement slowed down by more than packing).  vtr benchmark placement time looks OK but they're much smaller.
Given the code cleanup and new feature I'd probably take the small placement slowdown for now and look for it later if necessary. If you're able to find something with profiling or other means and claw it back that would be great of course!

The QoR failure on apex4 and tseng is likely just CAD noise.  It can fail even if the results are better (but by what we consider an unusual amount).  I suggest taking a look at the qor compare and see if it looks out of whack, but on those small circuits if we got a change that's bigger than expected it is probably just noise and we can update the golden when you commit.  Not failing on wintermute likely means the qor failure comes and goes with slightly different floating point round-off (due to some library or compiler or OS difference) making a different decision somewhere in the flow.

Bill-hbrhbr:
VTR benchmarks @ commit 9f18666: https://drive.google.com/file/d/1GZHxZpBSBssbBNsPnb0aOdoBGRvmk3ia/view?usp=sharing
Titan benchmarks @ commit 9f18666: https://drive.google.com/file/d/1kmeJmhJ1BDdeA1GV8pDCk7j9FfTAYvHn/view?usp=sharing

I ran this set of benchmarks in a more controlled environment, and now the results make more sense: the pack time and the route time have no degradations. 

The place time is slower, as expected. Possibly since:
1. There are more branchings in the try_swap() routine.
2. The place.cpp crucial data structures are no longer file-scope.

Bill-hbrhbr:
Close this PR so that it becomes archived.

Issue #1032, opened 2019-11-04
Unbounded memory usage in case of pathelogical heap behavior


litghost:
#### Expected Behaviour

Router should have bounded memory usage in all circumstances

#### Current Behaviour

A pathological graph can trivially cause a memory explosion.

#### Possible Solution

I haven't thought about solutions, I wanted to document the issue first. 

#### Context

The core router expansion is roughly:
 1. Expansion lowest code node
 2. Add heap entries for all nodes that have cheaper cost than the previously seen entry (see https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/route_timing.cpp#L1931)
 3. Repeat until target is found

Step 2 is where the memory explosion can occur, because the same node can be added to the heap infinitely many times, as long as each time the expansion sees a lower cost, even if the cost is only lower by a small amount.  Even worse, the best cost is not updated when a node is placed on the heap, but when it is popped!

See check for lower cost is here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/route_timing.cpp#L1925-L1931

But costs are only updated once popped:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/route_timing.cpp#L1490-L1491

The heap will contain N-1 duplicate entries for a node, if the node is added N times prior to being popped.  This can add up pretty quickly, I've seen cases of memory balloning to ~100 GB.

The maximum number of heap entries should be bounded to the number of nodes, if only there was a way to prune duplicate nodes of higher cost.

litghost:
@vaughnbetz We are seeing failures of this class with semi-regular frequency.  In the case of non-RCV routers, we only need to keep the "best" copy of each rr node.  For RCV routers, what should we keep?  I could code a pruning routing to keep a copy of each rr node with a histogram of codes.

I'll work on a proposed fix for this pathological behavior.

vaughnbetz:
I think it's too soon to say what to do for RCV.  So long as we can turn the pruning on and off, we can experiment with it later when we have RCV, so your changes aren't gated by any future (impossible to predict) RCV requirements.

A quick thing to test: if you update the best_cost when you put something on the heap, and make sure the pruning test when you take things off the heap at
[...]

PR #1454, opened 2020-07-27
Prune the heap

litghost:
#### Description

VPR currently has unbounded memory usage as a result of #1032.  This PR caps the heap memory usage at 4x the number of nodes in the entire graph.  This is likely overkill for non-RCV routers.  For RCV based routers, this pruning will be problematic, but there likely needs to be a bounded memory limit for RCV based routers too.

#### Related Issue

#1032

#### Motivation and Context

The current heap implementations (both BinaryHeap and Bucket) can consume an unbounded amount of memory if the router is exploring bad paths that have incrementally better costs.  This has been observed to allow memory usage to grow to consume the entire RAM of ~128 GB machines in pathological cases.  Based on analysis, this memory usage around ~1000 heap entries for each node in the graph.

In #1032 other optimizations were identified, such as updating the best cost to node during insertion, rather than node pop, but even in that case the pathological behavior can result in unbounded memory usage.  This PR bounds the memory usage even in that pathological case.

- [ ] Exists tests pass
- [ ] Failing unbounded memory case no longer consumes unbounded memory

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/heap_type.h
+++ b/vpr/src/route/heap_type.h
@@ -191,6 +191,34 @@ class HeapInterface {
     // Note: Only invoke this method if all objects returned from this
     // HeapInterface instace have been free'd.
     virtual void free_all_memory() = 0;
+
+    // Set maximum number of elements that the heap should contain
+    // (the prune_limit).  If the prune limit is hit, then the heap should
+    // kick out duplicate index entries.
+    //
+    // The prune limit exists to provide a maximum bound on memory usage in
+    // the heap.  In some pathological cases, the router may explore
+    // incrementally better paths, resulting in many duplicate entries for
+    // RR nodes.  To handle this edge case, if the number of heap items
+    // exceeds the prune_limit, then the heap will compacts itself.
+    //
+    // The heap compaction process simply means taking the lowest cost entry
+    // for each index (e.g. RR node).  All nodes with higher costs can safely
+    // be dropped.
+    //
+    // The pruning process is intended to bound the memory usage the heap can
+    // consume based on the prune_limit, which is expected to be a function of
+    // the graph size.
+    //
+    // max_index should be the highest index possible in the heap.
+    // prune_limit is the maximuming number of heap entries before pruning
+    // should take place.
+    //
+    // The prune_limit should always be higher than max_index, likely by a
+    // significant amount.  The pruning process has some overhead, so
+    // prune_limit should be ~2-4x the max_index to prevent excess pruning
+    // when not required.
+    virtual void set_prune_limit(size_t max_index, size_t prune_limit) = 0;
 };
 
 enum class e_heap_type {
--- a/vpr/src/route/binary_heap.cpp
+++ b/vpr/src/route/binary_heap.cpp
@@ -9,7 +9,9 @@ static size_t right(size_t i) { return (i << 1) + 1; }
 BinaryHeap::BinaryHeap()
     : heap_(nullptr)
     , heap_size_(0)
-    , heap_tail_(0) {}
+    , heap_tail_(0)
+    , max_index_(std::numeric_limits<size_t>::max())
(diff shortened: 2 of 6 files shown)


vpr/src/route/heap_type.h
```diff
@@ -191,6 +191,27 @@ class HeapInterface {
     // Note: Only invoke this method if all objects returned from this
     // HeapInterface instace have been free'd.
     virtual void free_all_memory() = 0;
+
+    // Set maximum number of elements that the heap should contain
+    // (the prune_limit).  If the prune limit is hit, then the heap should
+    // kick out duplicate index entries.
+    //
+    // For non-RCV routers this simply means taking the lowest cost entry for
+    // each index.  For RCV routers this is ???.
```
vaughnbetz, line 200: For RCV routers we still want the lowest cost item in the heap (the router cost function is more complex, but we still want the lowest cost one).  So you can update or delete this comment (always want to keep the lowest cost items).

vaughnbetz: I think it would be good to give some context of why this is necessary: basically what your wrote in the issue (can get many slightly better paths in some architectures, leading to the same node being on the heap many times -- if this happens a lot, memory usage can get out of control so we prune and save only the lowest cost entry on the heap for each node).

vaughnbetz:
Code looks good; just requesting a couple more comments.  Should put in the qor comparisons on the vtr and titan benchmarks too (not just regtests passed) make sure we haven't degraded the achieved wiring or the router runtime.  I don't expect much change in either, but good to check now.

vpr/src/route/heap_type.h
litghost: Awesome, I've updated the comment.  Good to know that this will generalize to the RCV router too!

litghost:
I've pushed the latest version with some comment and code fixes.  I'll post the QoR results once the "Nightly Tests" completes.

litghost:
Found and fixed a bug in the `binary_heap` compaction logic.  I'll pull QoR tommorow and post a comparision.

litghost:
QoR data looks good: [comparison_output.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4995975/comparison_output.xlsx)

litghost:
Don't merge just yet, I think I found a missing edge case in an overnight test.  Looking now.

vaughnbetz:
Yup, QoR data looks good (probably pruning never kicked in, which is good).
Would also be good to do a stress test on at least one circuit where you set the prune size to be small (num_rr_nodes?  0.25 * num_rr_nodes?) and print a message every time the heap is pruned to ensure the pruning code has been exercised and doesn't crash or cause a route failure.

litghost:
> Yup, QoR data looks good (probably pruning never kicked in, which is good).
> Would also be good to do a stress test on at least one circuit where you set the prune size to be small (num_rr_nodes? 0.25 * num_rr_nodes?) and print a message every time the heap is pruned to ensure the pruning code has been exercised and doesn't crash or cause a route failure.

That's what I've been testing on.  The binary_heap pruning logic worked fine, but I got an assertion failure on the the bucket logic.  FYI, the binary_heap compacted a handful of times (5-6).

litghost:
> Yup, QoR data looks good (probably pruning never kicked in, which is good).
> Would also be good to do a stress test on at least one circuit where you set the prune size to be small (num_rr_nodes? 0.25 * num_rr_nodes?) and print a message every time the heap is pruned to ensure the pruning code has been exercised and doesn't crash or cause a route failure.

Just for reference, pruning 21 million binary heap elements takes ~1 second per prune.  So as long as the prunes are infrequenct, the overall impact is a small CPU hit, but bounded memory usage.

litghost:
@vaughnbetz Go ahead and merge, the assertion error I saw is unrelated to this PR.

Issue #1453, opened 2020-07-25
Print out information / report on which rr-nodes are overused at the end of a failed routing


vaughnbetz:
Sometimes routing fails with only a few overused routing resource nodes. This often indicates a problem with the architecture input and output pin connectivity or switch pattern, often on specific blocks only.  It would be useful to get more information about which routing resource nodes are overused in such cases.

#### Proposed Behaviour
We should have a new command line option --set_max_overused_rr_node_report (can't think of a shorter name!) that defaults to 20.  Up to this many overused_rr_nodes should have their details printed if routing fails.  

#### Current Behaviour
We say how many nodes are overused, and by clicking the congestion button you can see them in the GUI, but no details are reported.

#### Possible Solution
Make the command line option above.  It should print a nicely formatted part of the log file with a table like:
rr index    type (IPIN, OPIN, CHANX, ...)    (x,y) coordinates (or start and end (x,y) if not the same)    ptc (pin, or track number), ...  (other helpful info you can think of, such as the segment type name if it's a wire and the name of the block type (clb, M20k, ...) if it is a block.

#### Context
When a circuit fails with a few overused rr-nodes it is hard to figure out why, even though it's usually a localized architecture problem.  See https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1448 for an example.

Issue #83, opened 2015-06-26
.route file pin sequence


kmurray:
Originally reported on Google Code with ID 90
```
What steps will reproduce the problem?
1. Sinthesize any Verilog file to a target architecture.

What is the expected output? What do you see instead?
It's not quite an issue but more a understanding matter.
Whenever a .route file is generated, the node sequence within a net will eventually
have a series of IPIN and OPIN elements, each in turn having a Pin: # attribute that
reportedly informs which pin of a given CLB a net is tied to. My CLB block has an input
with four pins, an output and a clock pin, evenly spreaded through the sides (using
the pinlocation's pattern attribute). I'm trying to figure what do this numbers actually
means. I mean, do Pin #0 is the first pin of the input of my CLB? By analyzing the
files generated by VTR I can almost surely assert that the last pin (pin 4) is my solely
output. That also made me believe that pin 0 is the first pin of my input I[0:0], pin
1 is I[1:1] and so forth. And where do the clock pins come to play? Am I right? Thank
you very much!!

What version of the product are you using? On what operating system?
VTR 7.0


```

Reported by `ky.s.oliver` on 2014-07-05 19:03:06

PR #1455, opened 2020-07-28
Print out info/report on overused rr-nodes at the end of a failed routing iteration

Bill-hbrhbr:
#### Description
Add the option **--max_logged_overused_rr_nodes N**, which is set to N=20 by default.
When each routing attempt fails, if the # of overused nodes is less than or equal to N, then print out the diagnostics of the overused nodes.

Add the option **--generate_rr_node_overuse_report on/off**, which generates a detailed report on overused RR nodes if the final routing attempt fails (i.e. the routing process has failed).

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1453

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -1797,19 +1797,32 @@ static void print_single_overused_node_status(int overuse_index, int inode) {
 static void generate_overused_nodes_report(const t_router_opts& router_opts) {
     const auto& device_ctx = g_vpr_ctx.device();
     const auto& route_ctx = g_vpr_ctx.routing();
+    const auto& cluster_ctx = g_vpr_ctx.clustering();
+
+    //Create overused nodes to congested nets loop up
+    //by traversing through the net trackbacks
+    std::unordered_map<size_t, std::vector<ClusterNetId>> overused_nodes_to_net_lookup;
+    for (ClusterNetId net_id : cluster_ctx.clb_nlist.nets()) {
+        for (t_trace* tptr = route_ctx.trace[net_id].head; tptr != nullptr; tptr = tptr->next) {
+            size_t inode = size_t(tptr->index);
+
+            int overuse = route_ctx.rr_node_route_inf[inode].occ() - device_ctx.rr_nodes[inode].capacity();
+            if (overuse > 0) {
+                overused_nodes_to_net_lookup[inode].push_back(net_id);
+            }
+        }
+    }
 
     //Append to the current report file
     std::ofstream os("report_overused_nodes.rpt", std::ofstream::out | std::ofstream::app);
     os << "----------------------------------------------------------------" << '\n';
     os << "Routing failure index #" << num_routing_failed << '\n';
 
-    for (size_t inode = 0; inode < device_ctx.rr_nodes.size(); inode++) {
-        int overuse = route_ctx.rr_node_route_inf[inode].occ() - device_ctx.rr_nodes[inode].capacity();
+    for (auto lookup_pair : overused_nodes_to_net_lookup) {
+        size_t inode = lookup_pair.first;
+        const auto& congested_nets = lookup_pair.second;
 
-        // Skip nodes that are not overused
-        if (overuse <= 0) continue;
-
-        // Report Basic info
+        //Report Basic info
         os << '\n';
         os << "Overused RR node index #" << inode << '\n';
         os << "Occupancy = " << route_ctx.rr_node_route_inf[inode].occ() << '\n';
@@ -1819,7 +1832,7 @@ static void generate_overused_nodes_report(const t_router_opts& router_opts) {
         os << "Xlow = " << device_ctx.rr_nodes[inode].xlow() << ", Ylow = " << device_ctx.rr_nodes[inode].ylow() << '\n';
         os << "Xhigh = " << device_ctx.rr_nodes[inode].xhigh() << ", Yhigh = " << device_ctx.rr_nodes[inode].yhigh() << '\n';
 
-        // Report Selective info
+        //Report Selective info
         auto node_type = device_ctx.rr_nodes[inode].type();
         if (node_type == e_rr_type::CHANX || node_type == e_rr_type::CHANY) {
             os << "Direction = " << device_ctx.rr_nodes[inode].direction_string() << '\n';
@@ -1827,6 +1840,13 @@ static void generate_overused_nodes_report(const t_router_opts& router_opts) {
             os << "Capacitance = " << device_ctx.rr_nodes[inode].C() << '\n';
         } else if (node_type == e_rr_type::IPIN || node_type == e_rr_type::OPIN) {
             os << "Side = " << device_ctx.rr_nodes[inode].side_string() << '\n';
+        }
+
+        //Reported corresponding congested nets
+        os << "Number of nets passing through this RR node = " << congested_nets.size() << '\n';
+        os << "Net IDs:" << '\n';
+        for (auto net_id : congested_nets) {
+            os << size_t(net_id) << '\n';
         }
     }
 
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -131,6 +131,10 @@ static void print_route_status(int itry,
                                std::shared_ptr<const SetupHoldTimingInfo> timing_info,
                                float est_success_iteration);
 
+static void print_overused_nodes_status(OveruseInfo& overuse_info);
+static void print_overused_nodes_header();
+static void print_single_overused_node_status(int overuse_index, int inode);
+
 static void print_router_criticality_histogram(const SetupTimingInfo& timing_info,
                                                const ClusteredPinAtomPinsLookup& netlist_pin_lookup);
 
@@ -715,6 +719,8 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
         VTR_LOG("Successfully routed after %d routing iterations.\n", itry);
     } else {
         VTR_LOG("Routing failed.\n");
+        //If the routing fails, print the overused info
+        print_overused_nodes_status(overuse_info);
 #ifdef VTR_ENABLE_DEBUG_LOGGING
         if (f_router_debug) print_invalid_routing_info();
 #endif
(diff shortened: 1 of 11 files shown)


vpr/src/route/route_timing.cpp
```diff
@@ -1669,6 +1675,107 @@ static void print_route_status(int itry, double elapsed_sec, float pres_fac, int
...
+
+static void print_overused_nodes_header() {
+    VTR_LOG("\nRouting Failure Diagnostics: Printing Overused Nodes Information\n");
+    VTR_LOG("------ ------- ---------- --------- -------- ------------ ------- ------- ------- ------- ------- ------- ------------ -----------------\n");
+    VTR_LOG("   No.   Inode  Occupancy  Capacity  RR Node    Direction    Side     PTC    Xlow    Ylow   Xhigh   Yhigh   Resistance       Capacitance\n");
+    VTR_LOG("                                        type                          NUM                                                               \n");
+    VTR_LOG("------ ------- ---------- --------- -------- ------------ ------- ------- ------- ------- ------- ------- ------------ -----------------\n");
+}
+
+static void print_single_overused_node_status(int overuse_index, int inode) {
```
tangxifan, line 1713: I would suggest using the strong id of rr_node `RRNodeId` because integer id may be misused easily. People may give an invalid integer which crashes VPR. This will be compatible with the refactoring effort on rr_graph

vpr/src/route/route_timing.cpp
```diff
@@ -1669,6 +1675,107 @@ static void print_route_status(int itry, double elapsed_sec, float pres_fac, int
...
+    const auto& device_ctx = g_vpr_ctx.device();
+    const auto& route_ctx = g_vpr_ctx.routing();
+
+    //Determines if direction or side is available for printing
+    auto node_type = device_ctx.rr_nodes[inode].type();
```
tangxifan, line 1718: Using the strong id, type of `rr_nodes` can be inquired through `node_type(const RRNodeId&)` method. Same for the other node-level inquiries.
You can find more detailed here.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph_storage.h

vpr/src/route/route_timing.cpp
```diff
@@ -1669,6 +1675,107 @@ static void print_route_status(int itry, double elapsed_sec, float pres_fac, int
     fflush(stdout);
 }
 
+static void print_overused_nodes_status(const t_router_opts& router_opts, const OveruseInfo& overuse_info) {
+    //Display upper limit
+    if (int(overuse_info.overused_nodes) > router_opts.max_reported_overused_rr_nodes) {
```
tangxifan, line 1680: It seems that if the number of overused nodes is larger than the threshold, we will do nothing.
Should we still report the overuse status and leave a message that the listed nodes are limited by the thresholds set by users.

Bill-hbrhbr: I think we'll log the overuse info if the number is small as well as produce an overuse info report file.

tangxifan: Thanks! That would be perfect.

Bill-hbrhbr: I assume you want all the nets at the congested node as well? Do you want the report to be indexed by the congested nets or the overused RR nodes?

tangxifan: Yes. Showing all the congested nets will be helpful. I prefer to be indexed by the overused RR Nodes, where we can see what nets share the same the node.

Bill-hbrhbr: I've added the nets to the report. You can checkout the report by specifying **--generate_rr_node_overuse_report on**. I will deal with the strong Id issue soon.

vaughnbetz: Adding the congested nets is a good idea.  In addition to printing their ids, I think we should also print their names and the type of driving block. E.g.
net #83 (processor|alu|reset) driven by a block of type CLB
as that can help locate them in the netlist and see patterns faster.

vaughnbetz:
Good comments Xifan, thanks.
I think we should not print the congested report when there are too many overused nodes though, as it will lead to a lot of huge files in a binary search for minimum channel width route.

tangxifan:
Hi @vaughnbetz 
   Indeed, we should avoid printing reports after each try during the binary search phase. Should we just generate a report at the end of routing phase? For instance, when the final routing of binary search failed, we can have a report for the final try? In most cases, the binary search will end with success unless the rr_graph is not routable.

vaughnbetz:
If there are only a few overused rr-nodes in a binary search it can be useful to print them out (they can occur repeatedly, indicating a routing pattern problem).  But otherwise we should only print it on the final failed routing.

So we could do:
- always print to the log file if the overused rr-nodes are small enough
- print the separate report only on the final (binary search or regular) routing, if it fails.

Bill-hbrhbr:
@tangxifan How many overused nodes are you trying to analyze in your case? Also, if you only care about the final routing, I can generate the report only if the final routing fails, regardless of how many nodes/nets there are.

tangxifan:
@Bill-hbrhbr I would focus on the final routing and a full report of overused nodes only in that routing result. Normally, the titan benchmarks failed with less than 20 overused nodes. Therefore I think what you did is good enough.

Bill-hbrhbr:
@tangxifan If you are mainly working with titan benchmarks, then I don’t think we need to worry too much about the binary search on minimum width_fac. Trying to add detection for the last failed routing attempt would require more drastic code refactoring.

tangxifan:
@Bill-hbrhbr I would suggest thinking beyond titan benchmarks. In general, we do not need any node overuse reports during the binary search. In my opinion, it is also easy to do so. Your report generation is called after routing is done, e.g., somewhere in the function `vpr_route_flow()`, where you just do reporting when routing status flags a failure. By doing so, you can skip the binary search part.

tangxifan:
Another modification you may consider is to keep your functions in a separate source file, as your functions have little dependency on the functions in route_timing etc. As such, you can call your function freely in any source file with a header file included.

Bill-hbrhbr:
@tangxifan Ok thats indeed the easy way to do it. Thanks for the suggestion!

Bill-hbrhbr:
@tangxifan Hi Xifan, I think I've basically finished improving the PR according to your comments. If you need other changes, feel free to point them out.

tangxifan:
Hi @Bill-hbrhbr Thanks for the effort. It looks good for me.

Bill-hbrhbr:
@vaughnbetz I've added documentation for the new command-line options (one for logging and one for the separate report). If you think they are a little minor for the developers' page, then I'll revert the last commit. Otherwise, this PR is ready to merge.

vpr/src/route/route_timing.cpp
Bill-hbrhbr: Sorry, I forgot about printing more detailed info on the nets. Working on this right now

tangxifan:
Hi @Bill-hbrhbr Sorry. I forgot to check the nets. Would you mind uploading an example report as well as the print-out in vpr.out log after the modification? It will be much more straightforward to check. You can pick an architecture and a benchmark and specify a fixed route channel width, which is slightly smaller than the minimum routable channel width.

Bill-hbrhbr:
[report_overused_nodes.rpt.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4997666/report_overused_nodes.rpt.txt)

@tangxifan This is the current format of the report file.

tangxifan:
@Bill-hbrhbr This looks almost perfect on my side. The only comment I have is to add the overuse id for each node and a total number of overused nodes to the report.

Bill-hbrhbr:
[report_overused_nodes.rpt.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4997747/report_overused_nodes.rpt.txt)
[example_logging.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4997748/example_logging.txt)

@tangxifan The info in two files match each other. The former is the updated report, and the latter is an excerpt from the vpr.out. (I used Notepad++ to open the latter one and it would display the format nicely.)

tangxifan:
@Bill-hbrhbr I am satisfied with the log file and report file. Just a question, did you set a threshold of 230 or more when generating the log file? I was wondering by default, it should only output 20 overused nodes.

Bill-hbrhbr:
Yeah, I set it to 1000 for demonstration only. Default is 20.

tangxifan:
@Bill-hbrhbr I just double checked the codes. It seems that if we set the threshold to 20 which is smaller than the actual number of overused nodes, nothing will be reported than a warning. Should we still report the first 20 overused nodes? Otherwise, users have to re-run the flow and find the correct threshold?

Bill-hbrhbr:
@tangxifan You can see that there is a table of routing status info (printed for each routing iteration/trial with a single channel width) above the overuse info table. The user will be able to tell how many nodes are left overused from the last line entry in that table (in this case, it is 229).

Regarding printing overused nodes up to the threshold number, I can make that happen if you think it would be helpful.

tangxifan:
@Bill-hbrhbr I personally lean to be user-friendly. If a user set a threshold while the total is like 229, we just output the given number of overused nodes in the log and leave a message telling the user to check the report file for a complete information. Does that make sense to you?

Bill-hbrhbr:
Ok. But the report will only be generated if the last routing attempt fails. In this case, this routing iteration produced 229 overused nodes, but it is not the final attempt from the binary search strategy, so the report wouldn't be generated.

In this case, it would be misleading to leave a message and tell the users to use the report option to generate a report for the complete info.

Bill-hbrhbr:
I think I'll just make the change to print the # of nodes towards the threshold and leave a message if the whole routing process fails and the report option is off.

tangxifan:
@Bill-hbrhbr I understand what you mean now. I thought your log file was showing the final routing attempt. It is reasonable not to show anything overused details and reports for non-final routing attempts.
I also agree with you that if user does not require reporting, we just leave a short message.

Would you mind showing me complete log files? I would like to see the two test cases:
- VPR succeeds in final routing attempt. We see a short message as you show in the previous log file during the intermediate routing attempts and no report file will be generated.
- VPR fails in the final routing attempt. We will see the overused nodes printed out and there is a report file created as well.

If the log file matches our expectation, I think this pull request is in a good shape.

Bill-hbrhbr:
@tangxifan 

**1st case:** VPR failed, report generation is off.
**Command (from the VTR directory):** ./vtr_flow/scripts/run_vtr_flow.pl ./vtr_flow/benchmarks/verilog/diffeq2.v ./vtr_flow/arch/timing/k6_frac_N10_frac_chain_mem32K_40nm.xml -crit_path_router_iterations 100 --max_logged_overused_rr_nodes 50 --route_chan_width 32

**Logfile:** [vpr_fail_no_report.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4998015/vpr_fail_no_report.out.txt)
**Report:** None
Near the end of the logfile, there is a reminder for the report generation option. I fixed the channel width so that the routing will fail (no binary search).

**2nd case:** VPR failed, report generation is on.
**Command:** ./vtr_flow/scripts/run_vtr_flow.pl ./vtr_flow/benchmarks/verilog/diffeq2.v ./vtr_flow/arch/timing/k6_frac_N10_frac_chain_mem32K_40nm.xml -crit_path_router_iterations 100 --max_logged_overused_rr_nodes 50 --route_chan_width 32 --generate_rr_node_overuse_report on

**Logfile:**  [vpr_fail_yes_report.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4998018/vpr_fail_yes_report.out.txt)
**Report:** [report_overused_nodes.rpt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4998019/report_overused_nodes.rpt.txt)
Near the end of the logfile, there is a reminder for the user to check out the report.

**3rd case:** VPR succeeded. The report is not generated.
**Command:** ./vtr_flow/scripts/run_vtr_flow.pl ./vtr_flow/benchmarks/verilog/diffeq2.v ./vtr_flow/arch/timing/k6_frac_N10_frac_chain_mem32K_40nm.xml -crit_path_router_iterations 100 --max_logged_overused_rr_nodes 50 --generate_rr_node_overuse_report on

**Logfile:** [vpr_succeed.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/4998023/vpr_succeed.out.txt)
**Report:** None

Search **Failed routing attempt #" to quickly locate the overuse table in the logfile. The # of entries printed is the minimum between # of overused nodes and the threshold number, with special messages printed if the threshold is the smaller one.

tangxifan:
@Bill-hbrhbr I just checked the log files and reports. I am satisfied by your efforts.

Bill-hbrhbr:
@tangxifan Thanks! Do you want it to be merged ASAP? I can do that after the travis-CI has been passed.

tangxifan:
I am fine with the codes actually. It depends on you and @vaughnbetz when it is a good time to merge. It will surely help my pull request.

Bill-hbrhbr:
@tangxifan Ok. I think this PR can be brought up at the meeting tomorrow. All the checks should pass by then, and @vaughnbetz  will get a chance to take a look at it.

vaughnbetz:
Thanks Bill and thanks for the detailed review Xifan.

vaughnbetz:
The reports look great.  Xifan, if you can run with the detailed reporting on the rr-graph changes (with no arch file changes) on the 3 unroutable designs that nearly route and post the resulting reports that would be great.

PR #1456, opened 2020-07-29
Odin_II: resolve defparams before override

emacdo12:
#### Description

Added two new functions (one calling the other) that are called during create_netlist and only in the top module to solve parameters early.

#### Related Issue

issue1369

#### Motivation and Context

Odin was failing because it wasn't solving it's parameters early enough.

Passed make test.

#### Types of changes

#### Checklist:


jeanlego:
looks good :) as soon as travis green lights it i will merge. 

side note: try to keep your title for your commit within 80 characters, otherwise it truncates it.
title should be short, body can be more explicit of what the commit does

PR #1463, opened 2020-08-04
Iterate bucket heap

litghost:
#### Description

Make some improvements to the bucket heap.  This PR adds some dynamic rescaling to the bucket heap and adds fully randomized popping.

The fully randomized popping from the front bucket was always the intention, but it was initially believed that the cost of implementing that would be prohibitive.  It isn't that expensive to do, and improves the worst case handling a bit.

The other bit change is dynamic bucket rescaling, which was an issue identified originally, but didn't manifest in the initial test cases.  However, with some placement seeds, the non-rescaling version of the bucket heap can become cost insensitive, and as a result have poor convergence qualities.

#### Motivation and Context

In some cases of tougher congestion resolution, the semi-randomized pop and lack of bucket scaling resulted in the bucket heap degrading to a semi-random walk.  By adding some dynamic bucket scaling, this behavior is limited.   This new bucket heap balances the initial behavior early in the router flow where costs are well behaved (mostly cost == time), and then transitions well into the case where costs are poorly scaled (cost == congestion).

A failing 7-series circuit + bad seed was iterated on, and then tested with the board 7-series circuit list.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/bucket.cpp
+++ b/vpr/src/route/bucket.cpp
@@ -4,6 +4,9 @@
 #include "vtr_log.h"
 #include "vpr_error.h"
 
+static constexpr float kDivisionScaling = 100000.f;
+static constexpr ssize_t kMaxBuckets = 2000000;
+
 BucketItems::BucketItems() noexcept
     : alloced_items_(0)
     , num_heap_allocated_(0)
@@ -21,7 +24,8 @@ Bucket::Bucket() noexcept
     , max_cost_(0.f)
     , num_items_(0)
     , max_index_(std::numeric_limits<size_t>::max())
-    , prune_limit_(std::numeric_limits<size_t>::max()) {}
+    , prune_limit_(std::numeric_limits<size_t>::max())
+    , front_head_(std::numeric_limits<size_t>::max()) {}
 
 Bucket::~Bucket() {
     free_all_memory();
@@ -36,6 +40,7 @@ void Bucket::init_heap(const DeviceGrid& grid) {
     memset(heap_, 0, heap_size_ * sizeof(t_heap*));
 
     heap_head_ = std::numeric_limits<size_t>::max();
+    front_head_ = std::numeric_limits<size_t>::max();
     heap_tail_ = 0;
     num_items_ = 0;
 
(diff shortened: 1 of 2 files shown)


vaughnbetz:
Only error in vtr strong sanitized tests is a known (and unrelated) memory leak in graphics.  That's tracked in a separate issue and doesn't gate this.

vpr/src/route/bucket.h
```diff
@@ -257,18 +259,44 @@ class Bucket : public HeapInterface {
...
-    BucketItem** heap_; /* Buckets for linked lists*/
-    size_t heap_size_;  /* Number of buckets */
-    size_t heap_head_;  /* First non-empty bucket */
-    size_t heap_tail_;  /* Last non-empty bucket */
-    float conv_factor_; /* Cost bucket scaling factor */
+    BucketItem** heap_;      /* Buckets for linked lists*/
+    size_t heap_size_;       /* Number of buckets */
+    size_t heap_head_;       /* First non-empty bucket */
+    size_t heap_tail_;       /* Last non-empty bucket */
+    float conv_factor_;      /* Cost bucket scaling factor */
```
vaughnbetz, line 266: Can you explain this more?  Do you multiply cost by conv_factor to find the bucket?

vpr/src/route/bucket.h
```diff
@@ -257,18 +259,44 @@ class Bucket : public HeapInterface {
...
+    float division_scaling_; /* Scaling factor used during rescaling.
+                              * Larger division scaling results in larger
+                              * conversion factor.
+                              */
+    ssize_t max_buckets_;    /* Maximum number of buckets to control when to
```
vaughnbetz, line 271: Explaining how division_scaling and max_buckets interact would be useful.  I can imagine setting a max_buckets and a max_cost and computing conv_factor.  This is a bit more involved than that; can you give an example of how the parameters interact?

vpr/src/route/bucket.cpp
```diff
@@ -119,43 +158,48 @@ void Bucket::check_scaling() {
...
+    if (min_bucket < 0 || max_bucket < 0 || max_bucket > max_buckets_) {
+        rescale();
+    }
+}
 
-            std::fill(heap_ + heap_head_, heap_ + heap_tail_ + 1, nullptr);
-            heap_head_ = std::numeric_limits<size_t>::max();
-            heap_tail_ = 0;
+void Bucket::rescale() {
+    // Choose a scaling factor that accomidates 50k buckets between
```
vaughnbetz, line 167: This comment doesn't seem to reflect the new dynamic rescaling of the number of buckets (and the space between them).

vpr/src/route/bucket.cpp
```diff
@@ -4,6 +4,27 @@
 #include "vtr_log.h"
 #include "vpr_error.h"
 
+// Initial bucket scaling.  A larger division scaling results in smaller cost
```
vaughnbetz, line 7: Either here or later it seems like you need an overview comment on how the heap number of buckets and spacing between buckets is controlled. This describes the individual variables, but I still find division a bit mysterious.  I get the basic idea (I think) of you initially have 100k buckets (or you can grow to 100k buckets) but if the front bucket gets too full you'll increase the number of buckets by 2x and reduce their spacing, until you hit kMaxMaxBuckets.  But there is a bit more to what you're doing than that, and I think an overview comment would help.

litghost: Big comment written to address a couple of your comments.

litghost:
@vaughnbetz Take a look, comments have been updated

vpr/src/route/bucket.cpp
```diff
@@ -4,6 +4,63 @@
...
+ *    cost insenstive.
+ *
+ * The other consideration is to avoid rescaling the buckets too often, as
+ * that operation consumes time without delivering useful work.
+ *
+ * To prevent rescaling constantly, the bucket heap determines if a rescaling
+ * is needed based on two conditions:
+ *
+ *  - The maximum item cost (max_cost_) would require a bucket index that is
+ *    greater than max_buckets_.  When this occurs, a rescaling is done to
```
vaughnbetz, line 29: don't you rescale in this case to make the width of the buckets *larger* (not smaller)?

vaughnbetz:
Thanks; looks good.  I just have one question / comment (think a smaller should be a larger in one comment).

vpr/src/route/bucket.cpp
litghost: Yep, you are right.  Fixed.

PR #1466, opened 2020-08-05
Cleaning up code related to fix_pins and fix_clusters vpr options

sfkhalid:
Removed pad_loc_file string as it was no longer needed.

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -2253,11 +2253,6 @@ void set_conditional_defaults(t_options& args) {
         args.anneal_sched_type.set(AUTO_SCHED, Provenance::INFERRED); // Otherwise use the automatic schedule
     }
 
-    //Are the blocks locked to locations given by a constraints file?
-    if (std::string(args.constraints_file) != "") {
-        VTR_ASSERT(!args.constraints_file.value().empty());
-    }
-
     /*
      * Routing
      */
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -1608,7 +1608,7 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .help(
             "Fixes block locations during placement. Valid options:\n"
             " * path to a file specifying block locations (.place format with block locations specified).")
-        .default_value("not_locked")
+        .default_value("")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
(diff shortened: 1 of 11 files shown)


vpr/src/base/read_options.cpp
```diff
@@ -2253,24 +2253,8 @@ void set_conditional_defaults(t_options& args) {
...
-        VPR_FATAL_ERROR(VPR_ERROR_UNKNOWN, "Unknown I/O pad location type\n");
-    }
-
     //Are the blocks locked to locations given by a constraints file?
-    if (std::string(args.constraints_file) == "not_locked") {
-        args.block_loc_type.set(NOT_LOCKED, Provenance::INFERRED);
-    } else {
-        args.block_loc_type.set(LOCKED, Provenance::INFERRED);
+    if (std::string(args.constraints_file) != "") {
         VTR_ASSERT(!args.constraints_file.value().empty());
```
vaughnbetz, line 2258: This assert doesn't look useful.  You are checking if the string is not empty, and then assert that it is not empty (the .empty() call boils down to the same check as the != "" call.  Delete this code block (if and assert) as it no longer does anything.

vpr/src/base/vpr_types.h
```diff
@@ -837,6 +837,11 @@ enum e_place_algorithm {
     PATH_TIMING_DRIVEN_PLACE
 };
 
+enum e_pad_loc_type {
+	FREE,
```
vaughnbetz: Give a short comment of FREE and RANDOM (I/Os can move and optimized in placement, and I/Os are randomly locked down (to emulate board-level constraints).

vaughnbetz:
Sanitized test failure is the known graphics memory leak issue, so OK.
Travis CI: PR failure is the llvm fetch failure (only affects compilation warning check), so OK.
I'll merge this without the comment change as Sarah is going to work on some code refactoring here next, so I'd like to land this and defer the comment upgrade to that refactoring. Waiting for the nightly tests to pass.

sfkhalid:
> Sanitized test failure is the known graphics memory leak issue, so OK.
> Travis CI: PR failure is the llvm fetch failure (only affects compilation warning check), so OK.
> I'll merge this without the comment change as Sarah is going to work on some code refactoring here next, so I'd like to land this and defer the comment upgrade to that refactoring. Waiting for the nightly tests to pass.

@vaughnbetz Vaughn, the nightly tests have passed now. Would you be able to merge this code?

PR #1477, opened 2020-08-12
Avoid strcpy inside of atom netlist.

litghost:
#### Description

Saves around 20 out of 30 seconds in circuit cleaning on baselitex 7-series flow per VPR invocation.

#### Motivation and Context

We've noticed that the circuit cleaning time was large relative to the overall runtime (30 seconds into 120 second total run).  A profile run showed about 20 seconds being spent in a simple string copy to determine if a model is either inpad/outpad.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/atom_netlist.h
+++ b/vpr/src/base/atom_netlist.h
@@ -87,6 +87,9 @@ class AtomNetlist : public Netlist<AtomBlockId, AtomPortId, AtomPinId, AtomNetId
      */
     AtomNetlist(std::string name = "", std::string id = "");
 
+    AtomNetlist(const AtomNetlist& rhs) = default;
+    AtomNetlist& operator=(const AtomNetlist& rhs) = default;
+
   public: //Public types
     typedef std::vector<std::vector<vtr::LogicValue>> TruthTable;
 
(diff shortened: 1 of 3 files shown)


vpr/src/base/atom_netlist.h
```diff
@@ -257,6 +261,10 @@ class AtomNetlist : public Netlist<AtomBlockId, AtomPortId, AtomPinId, AtomNetId
     vtr::vector_map<AtomBlockId, const t_model*> block_models_;   //Architecture model of each block
     vtr::vector_map<AtomBlockId, TruthTable> block_truth_tables_; //Truth tables of each block
 
+    //INPAD and OUTPUT models
```
vaughnbetz, line 264: It would be good to comment these new member variables and why they exist:

Something like:
// Input IOs and output IOs always exist and have their own architecture models.
// While their models are already included in block_models_, we also store direct
// pointers to them to make checks of whether a block is an INPAD or OUTPAD fast,
// as such checks are common in some netlist operations (e.g. clean-up of an input netlist).

vaughnbetz:
Failure is spurious (just a known memory leak in the graphics test due to the lack of a sanitizer suppression file).

vpr/src/base/atom_netlist.h
litghost: Fixed

PR #1479, opened 2020-08-13
Vtr_reg_strong Some Regression tests results are not being checked

shadtorrie:
We are not running some qor tests that we should be run. This is on vtr_reg_strong that it can be seen. This has been fixed

#### Description
There are some tests that instead of comparing to the golden results, it compares to a second output from vtr. if one of these tests comes before a test that does need to compare to the golden results, the boolean value that tells the parser to compare golden results is still false and just doesn't do any qor tests on that test.
After fixing this bug, there are 36 failing tests on the vtr_reg_strong test. These same tests are also failing on my python version, the golden standards are completely different from what vtr is outputting.

#### Related Issue
No issue yet.

#### Motivation and Context
Vtr_reg_strong test passes when it actually is not testing the results of some tests and it in fact should actually be failing.

Run the vtr_reg_strong and make sure that all results are checked on the qor tests that should be

#### Types of changes

#### Checklist:
- [ ] All new and existing tests passed: fixing this bug brings up failures that we did not know about:

[This is the output](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071688/vtr_flow.txt) when vtr_reg_strong is run.

The following tests fail when vtr_reg_strong is run with their corisponding vpr.out file attatched:

- strong_eblif_vpr: the golden results may just need to be updated. 
[conn_order_vpr.out ](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071701/conn_order_vpr.txt) 
[test_eblif_vpr.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071704/test_eblif_vpr.txt)

- strong_unroute_analysis: qor tests might need to be turned off for this test, but be aware that we currently expect the flow to fail on one of these but it doesn't fail currently. that's something that needs to be looked into. 
[common_--route_chan_width_20_vpr.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071720/common_--route_chan_width_20.txt)
[common_--route_chan_width_8_vpr.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071722/common_--route_chan_width_8.txt)
[common_--route_chan_width_8_--analysis_vpr.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071723/common_--route_chan_width_8_--analysis.txt)
[common_--route_chan_width_20_--analysis_vpr.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071724/common_--route_chan_width_20_--analysis.txt)

- strong_clock_aliases_set_delay: this is failing with a file does not exist error.
[strong_clock_aliases_set_delay_vpr.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071731/strong_clock_aliases_set_delay_vpr.txt)

- strong_pack_disable
[strong_pack_disablek_6_frac_N10_40nml_mult_5x6_vpr.out](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5071735/strong_pack_disablek_6_frac_N10_40nml_mult_5x6_vpr.txt)


vtr_flow/scripts/parse_vtr_task.pl
```diff
@@ -150,6 +150,7 @@ sub parse_single_task {
 	my $qor_parse_file;
         my $second_parse_file;
         my $counter = 0;
+	my $single_check_golden = $check_golden;
```
vaughnbetz, line 153: It would be good to comment what this variable is (it's a complicated one).  

Should also have another comment should also probably explain the various ways this routine can operate 
 - (parse one result file for errors, parse a result and compare to a golden (prior) result, or parse two results files and make sure they are the same)?
 - I suspect the top of the file should also comment the various ways this parser can operate (may overlap with the function level comment -- use your judgement about where best to comment this, but we should comment that this parser can operate in different modes).

shadtorrie: @vaughnbetz  comments have been added and log files are now in the description above.

vaughnbetz: Thanks!

shadtorrie:
@vaughnbetz I attached log files above for the errors that are occurring,  is the next step for you to review them to decide what action needs to happen here? Let me know if there is anything else I can do or more log files to provide.

vaughnbetz:
Thanks for all the attached details and summary on this Shad.

The following tests fail when vtr_reg_strong is run with their corisponding vpr.out file attatched:

==================
strong_eblif_vpr: the golden results may just need to be updated.
conn_order_vpr.out
test_eblif_vpr.out

These are trivially small circuits; their QoR is worse due to what looks like hard-to-route rr-graphs (narrow channel widths, small device).  Updating their golden results is fine; if they fail after that (get too good a QoR results sometimes) we can turn off QoR checking for them.

Shad, one odd thing:  the critical path parsed out of conn_order is 1.4ns, but the regtest failure says it is 2 ns.
From vpr.out:  Final critical path delay (least slack): 1.40428 ns,
From vtr_reg_strong output:   k6_frac_N10_40nm.xml/conn_order.eblif/common critical_path_delay: golden = 1.17134 result = 2.01453

I think the difference is that we're probably running a binary search place and route and then running at a computed channel width to get the final critical path delay at an easier-to-route channel width.  Can you check if that's the case?  If so, it would be good to augment vtr_flow.py to bring back both vpr.out files when there are two in a multi-part flow (binary search routing, then routing at a fixed channel width).

====================
strong_unroute_analysis: 
common_--route_chan_width_20_vpr.out
common_--route_chan_width_8_vpr.out
common_--route_chan_width_8_--analysis_vpr.out
common_--route_chan_width_20_--analysis_vpr.out
VB: The --route option is not making it to vpr for these tests.  Looking at the config file:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_unroute_analysis/config/config.txt
I see 
# Script parameters
script_params_common = -starting_stage vpr --pack --place --route
script_params_list_add = --route_chan_width 20              #Routeable implicit analysis (analysis metrics should be produced)
script_params_list_add = --route_chan_width 20 --analysis   #Routeable explicit analysis (analysis metrics should be produced)
script_params_list_add = --route_chan_width 8               #Unrouteable implicit analysis (no analysis metrics should be produced)
script_params_list_add = --route_chan_width 8 --analysis    #Unrouteable explicit analysis (analysis result should be produced despite routing failure)

But the test vpr.out is not setting all those options to vpr.  Instead it is being told to pack and place only (not route and don't do analysis) which is not producing the expected output, and hence producing failures.  Here's the command line from the first test common_--route_chan_width_20_vpr.out
/home/shadtorrie/git/vtr-verilog-to-routing/vpr/vpr k6_N10_mem32K_40nm.xml traffic --circuit_file traffic.pre-vpr.blif --pack --place --route_chan_width 20 --max_router_iterations 150

As you can see, no --route option is set, but it should be.  The last two tests are missing both --route and --analysis options. So this looks like a vtr_flow.py issue.

=======

Still analyzing the others ....

vaughnbetz:
strong_clock_aliases_set_delay: this is failing with a file does not exist error.
strong_clock_aliases_set_delay_vpr.out
- Looks like this is missing the golden_result.txt file.  Can you generate it and check it in?

=========
strong_pack_disable
strong_pack_disablek_6_frac_N10_40nml_mult_5x6_vpr.out

- This design is getting the wrong architecture file.  vtr_flow.py seems like it has a bug for this one.
VPR's arguments are below:
/home/shadtorrie/git/vtr-verilog-to-routing/vpr/vpr k6_frac_N10_40nm.xml mult_5x6 --circuit_file mult_5x6.pre-vpr.blif --min_route_chan_width_hint 30

But the arch file specified by the config file to use is:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/timing/k6_frac_N10_40nm_disable_packing.xml

This regtest has two architecture files to sweep over, so maybe that's triggering the bug.

Add architectures to list to sweep
arch_list_add=k6_frac_N10_40nm.xml                 # A regular architecture that VPR should succeed
arch_list_add=k6_frac_N10_40nm_disable_packing.xml # Crucial LUT6 mode is disabled for packing, VPR should fail

shadtorrie:
@vaughnbetz thanks for getting me these results. Regarding the critical path delay that appears to be displaying the incorrect value, it is getting this value from the vpr.crit_path.out file instead of the vpr.out file. This file is the second easier to run vpr.out file that you spoke of. So both files are being saved and used. Is there any other action that I should take on this one?

vaughnbetz:
Thanks Shad.  When you run through CI, is this file brought back so you can click on it (as well as vpr.out) and see it?  That was my feature request, if it isn't there already.

vaughnbetz:
Hi Shad,

What do check_route and check_place mean in the recent commit?  If these mean "--route" and "--place" then I think check_route and check_place aren't great names for them (as we have routines to check the legality of routing and placement that have those names in the code, and some conservative checks can be disabled on the comman line.  If they have a complex meaning I think a comment would be helpful, and if they simply mean do_route or do_place then I think renaming them would be good.

shadtorrie:
@vaughnbetz  I don't quite understand what you mean by the file being "brought back so you can click on it". Would this be in travis? is this an existing feature? Do you mean just a path to the file that can be clicked on in an IDE like Visual Studio code? would this only be clickable if there was an error? 

As pertaining to check_place and check_route, these are the names of the arguments that go directly into run_vtr_flow.pl and .py, check_place enables routing and check_route enables analysis, not really sure why. there is a help  description in the Argument parser on run_vtr_flow.py. Would you like me to add more comments at other occurrences of these variables.

shadtorrie:
@vaughnbetz on strong_pack_disable, as you mentioned, the config file runs two architectures.  It seems like this is just running one of the two architectures it is supposed to.  Is there something in the logs you are seeing to suggest there is a bug?

vaughnbetz:
check_place and check_route: Yes, I'd at least comment them, and I'd prefer to rename them to something more intuitive.  Given their usage in the latest script, I think check_place means "read_first_placement" or "route_existing_placement" or some such.  I'm not sure what renaming them entails given that they go straight into the scripts; if it's going to break a whole bunch of tests I wouldn't both and would instead just comment them.  If it doesn't really matter what they're called (doesn't break dependencies) then I'd rename them.

vaughnbetz:
Bring back multiple .out files:  Not sure I know what I am talking about here -- I noticed you only included the binary search vpr.out files in your summary for strong_eblif_vpr.  If that is because it is easy to download the vpr.out after a failed CI test but not easy to download the vpr.crit_path_route.out (forget the exact name) then it is a request to make both easy to download from CI.  If both are equally easy to get from the CI system now then there's no action.

vaughnbetz:
strong_pack_disable:  you're right -- looking at the vtr_flow output it looks like both architectures were run (both are listed in the output).  The failure then is 

regression_tests/vtr_reg_strong/strong_pack_disable...[Fail] 
 k6_frac_N10_40nm.xml/mult_5x6.blif/common geomean_nonvirtual_intradomain_critical_path_delay: golden = -1 result = nan

So that seems like a change int he golden result used to be -1 (which I think means invalid) and is now nan.  Parsing error or return value error leading to the change?

I think strong_pack_disablek_6_frac_N10_40nml_mult_5x6_vpr.out is using the right architecture file (k6_frac_N10_40nm) -- I misread the .out filename earlier and thought it didn't match the arch file but you are right that it appears to match.

shadtorrie:
@vaughnbetz I don't quite understand what you mean when you say that golden result used to be -1 and is now nan, the golden result is currently -1 and the output from vpr is nan. Is this just a golden result issue? or is this some issue in the flow?

vaughnbetz:
I think it's fine to just update the golden result for that one.  I checked the vpr output, and as you said, it is nan for the geometric average non-virtual clock.  I checked the circuit and it does not have any clocks, so removing the virtual I/O clock results in nothing to average for speed (and hence the nan is correct).  The old vtr_flow.pl appears to have parsed out or set -1 in that case, and the new vtr_flow.py appears to parse out and set nan.  Nan is in the vpr.out, so what vtr_flow.py is doing is correct and I think you can just update the golden output.

You can take a look at vtr_flow.pl if you like to see if it converted all invalid values (missing, nan, etc.) to -1 when storing / passing to the qor comparison scripts.  If it did and there seems to be a valid reason why then you could duplicate that behaviour.  But if there doesn't seem to be any documented reason of why the old flow put a -1 in the golden results, then I'd just consider the new vtr_flow.py correct and update the golden results for this one.  Converting to -1 seems more mysterious than just putting the nan in the golden results, so if it's not clear why it was ever -1 I'd just keep the current behaviour and update golden.

shadtorrie:
@vaughnbetz  the comments have been added.

vtr_flow/scripts/python_libs/vtr/flow.py
```diff
@@ -119,7 +119,11 @@ def run(
 
         check_incremental_sta_consistency :
             Do a second-run of the incremental analysis to compare the result files
-
+        
+        check_route:
+            Check first placement by enabling VPR analysis.
```
vaughnbetz, line 124: Can you change this to "Check first placement and routing by enabling VPR analysis".

shadtorrie:
@vaughnbetz  this is ready to merge.

vaughnbetz:
Thanks Shad!

PR #1481, opened 2020-08-13
Fixed a typo

vaughnbetz:
documentation was incorrectly spelt.


vaughnbetz:
Errors are spurious (sanitizer graphics error that's now fixed in CI, and an llvm download failure for compiler warnings). Restarting the latter just to see if it will finish; can merge regardless.

PR #1482, opened 2020-08-15
[WIP] Odin: Add tristate gates and multiplexers in the backend

jeanlego:
#### Description
This adds buffif0, buffif1, notif0 and notif1 to the frontend
as well as the multiplexer ~onto which it maps to in the backend~.
The mapping ~to the multiplexer~ is done during ~partial mapping~*during output blif* as a soft implementation since the tristate buffers can have a hardware equivalent and it would be possible to map these in the future.
We map late so that instantiate a tristate should not require to manually add the unconn manually.

some liberty was taken to also move some decoded mux onto multiplexers to reduce the number of logic elements required.
The memory decoder was also simplified into a single logic element.

### TODO
- [ ] add test for bufifs

#### Motivation and Context
inouts require tri state logic

reg ligt and strong

#### Types of changes

#### Checklist:

- [ ] I have added tests to cover my changes **TODO**


jeanlego:
@emacdo12 this would need your eye once you have time,
I still need to create a test for the primitives that were added, but the PR should be near completion, see TODO's in the pr

@djns99 this should add what you need :)
also added a flag so you can emulate the tristate logic in the simulator

ODIN_II/SRC/ast_elaborate.cpp
```diff
@@ -976,6 +976,24 @@ ast_node_t* build_hierarchy(ast_node_t* node, ast_node_t* parent, int index, sc_
                     error_message(AST, node->loc,
                                   "%s", "Cannot perform operation with nonexistent value");
                 }
+                // TODO convert to bufifs
```
jeanlego, line 979: buffs are gate instantiation nodes and as such they expect their output to be a child, 
I don't think here is the right place to do this unless we want to do lookahead in the tree or unless this optimization is needed at the AST level for inouts.

@djns99 I don't recollect you needing that information this early on, if you feel it would help I'll work on this, otherwise, I might just remove it and do the recognition during the netlist creation phase.

ODIN_II/SRC/include/odin_types.h
```diff
@@ -141,6 +141,8 @@ struct global_args_t {
     argparse::ArgValue<int> sim_random_seed;
 
     argparse::ArgValue<bool> interactive_simulation;
+    // disable bufs in simulation so that unconn (z) are not filtered by buffers in multpilexers
+    argparse::ArgValue<bool> recover_tristate_logic;
```
jeanlego, line 145: @djns99 fyi

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -1375,9 +1372,43 @@ static void compute_flipflop_node(nnode_t* node, int cycle) {
...
+     * we buffer the output since the output should be a logical value,
+     * if we simply passed through the value, it would not be equivalent 
+     * to computing the PLA table
+     * NOTE: 
+     *   This breaks the "analogue-ish" behavior of tristate buffers.
+     *   since this is a digital logic simulator, it makes more sense to 
+     *   emulate digital behavior alone, but, if you need to have knowledge of the 'Z'
+     *   driver in some form, you may want to passthrough the value.
+     *   the buf will filter 'z' to 'x'
+     */
```
jeanlego, line 1398: @djns99

ODIN_II/SRC/netlist_create_from_ast.cpp
```diff
@@ -3225,106 +3225,127 @@ int alias_output_assign_pins_to_inputs(char_list_t* output_list, signal_list_t*
...
-            out_1 = create_output_pin(gate_instance->children[1], instance_name_prefix, local_ref);
+    output_list = (signal_list_t**)vtr::calloc(output_size, sizeof(signal_list_t*));
+    input_list = (signal_list_t**)vtr::calloc(input_size, sizeof(signal_list_t*));
 
-            oassert(in_1 != NULL);
-            oassert(out_1 != NULL);
+    // skip symbol
+    size_t child_id = 1;
+    for (size_t i = 0; i < output_size; i += 1) {
+        output_list[i] = create_output_pin(gate->children[child_id], instance_name_prefix, local_ref);
```
jeanlego, line 3295: this is the only reason the children cannot be done in the netlist_create_from_ast, It we can flag outputs from instances as such in the types, they could be recognized and handled there rather, reducing a lot of clutter.

ODIN_II/SRC/ast_elaborate.cpp
djns99: I don't do any checks that drivers are Z so I don't need to know about this anywhere I believe. As long as the simulator resolves the gate to a Z it should be fine.

jeanlego: just to clarify, I was wondering if you needed to know about the bufifs early on or not.

djns99: No inouts and multidriver do not care about any specific components. They only care about the simulator outputting the correct value

ODIN_II/SRC/netlist_create_from_ast.cpp
```diff
@@ -3361,10 +3382,26 @@ signal_list_t* create_operation_node(ast_node_t* op, signal_list_t** input_lists
...
             break;
+        case BUFIF0:
+            output_port_width = 1;
+            input_port_width = 1;
+            break;
+        case BUFIF1:
+            output_port_width = 1;
+            input_port_width = 1;
+            break;
+        case NOTIF0:
```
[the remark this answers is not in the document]
jeanlego: yes, and actually should be 2 bit wide.

ODIN_II/SRC/read_blif.cpp
```diff
@@ -579,11 +581,105 @@ void create_hard_block_nodes(hard_block_models* models, FILE* file, Hashtable* o
     vtr::free(names);
 }
 
+/*---------------------------------------------------------------------------------------------
```
jeanlego, line 584: @djns99 while this is rather ugly (don't know that I have a better way to write this tho)
this should do what you want now.
It still kind of breaks the spec, but this is the only way I can get the 'z' to go through during simulation.

PR #1484, opened 2020-08-18
Add model type

acomodi:
#### Description

VPR, by default, removes and prunes away all the blocks that have no fanout.
This PR makes it possible to annotate the models to have be marked as **special**, so that, during the sweeping process, they do not get removed.

By default, all the models are treated as **normal**, therefore no change in any architecture file regarding the regression tests is required.

#### Related Issue

Symbiflow issue related to VPR: https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1470

#### Motivation and Context

In Symbiflow, more specifically Series7 devices, there are special blocks that do not require output nets to function correctly. One example is the IDELAYCTRL, which takes as input a reference clock and a reset signal, without the need of outputting anything.

Given that the reference clock and the reset signal needs to be routed in order to have the IDELAYCTRL correctly functioning, this critical block cannot be swept away.

The usage of the annotation for the `model` is required as the netlist could potentially contain blocks that need to be swept, therefore a more generic approach wouldn't be possible.

#### Types of changes

basic and strong regression tests

#### Checklist:


doc/src/arch/reference.rst
```diff
@@ -28,10 +28,18 @@ The architecture tag contains the following tags:
...
 ---------------------------------
-The ``<models>`` tag contains ``<model name="string">`` tags.
+The ``<models>`` tag contains ``<model name="string" type="string">`` tags.
 Each ``<model>`` tag describes the BLIF ``.subckt`` model names that are accepted by the FPGA architecture.
 The name of the model must match the corresponding name of the BLIF model.
 
+The type of the model is not strictly required and it can assume the following values:
+
+* normal (default)
+* special
```
litghost, line 38: Rather than having a type, maybe an attribute like `never_prune="true"`?

acomodi: The main reason to have a type was that it might get expanded in the future to allow some other kind of operations on different model types, also to leave this as generic as possible.

Having a `never_prune` attribute might be too over-specific from an architecture description point of view.

litghost: But having it as a type is also never particularly generic.  A "type" works well if each addition type is very seperate.  But if each type enables some features but not others, attribute flags allow for better flexiblity.  So in this case, "special" is non-descriptive name, and the attribute name "type" is also not descriptive.

I'm open to other possibilities, but `type="special"` is not a good solution here.

acomodi: I see your point and I agree. I'll get the changes in and transition to use attribute flags instead of model types.

doc/src/arch/reference.rst
```diff
@@ -28,10 +28,17 @@ The architecture tag contains the following tags:
 
 Recognized BLIF Models (<models>)
 ---------------------------------
-The ``<models>`` tag contains ``<model name="string">`` tags.
+The ``<models>`` tag contains ``<model name="string" never_prune="string">`` tags.
 Each ``<model>`` tag describes the BLIF ``.subckt`` model names that are accepted by the FPGA architecture.
 The name of the model must match the corresponding name of the BLIF model.
 
+The never_prune flag of the model is not strictly required and it can assume the following values:
```
vaughnbetz, line 35: I think you could reword this comment to be a bit more clear.

"The never_prune flag is optional and can be either:"

... <text on options>

"Normally blocks with no output nets are pruned away by the netlist sweepers in vpr (removed from the netlist); this is the default behaviour.  If never_prune = true is set on a model, then blocks that are instances of that model will not be swept away during netlist cleanup. This can be helpful for some special blocks like I/Os <...add motivating case...> so space on the chip is still reserved for them, despite them not being connected to anything."

libs/libarchfpga/src/logic_types.h
```diff
@@ -49,6 +49,8 @@ struct t_model {
     vtr::t_linked_vptr* pb_types = nullptr; /* Physical block types that implement this model */
     t_model* next = nullptr;                /* next model (linked list) */
 
+    bool never_prune = false;
```
vaughnbetz, line 52: comment this again briefly:
/// don't remove from the netlist even if a block of this type is unconnected to the rest of the netlist

doc/src/arch/reference.rst
acomodi: Done, slightly changed also the suggested modification.

libs/libarchfpga/src/logic_types.h
acomodi: Done

vaughnbetz:
Thanks.  Looks good; will merge as soon as the CI finishes.

Issue #1467, opened 2020-08-05
Add block type for IPINs in overused rr-node report


vaughnbetz:
The new overused rr-node report file looks good.
One thing that would be good to add though: when an IPIN is overused, print the type of block it is part of.  This should be obtainable by checking the grid at the IPIN location (for large blocks, you'll have to be careful and may have to check the origin of the block, which the offset values in the grid will point you to). If more than one block is at that (x,y) location (multiple subtiles) you could print all the block types at that (x,y) location.  The common case will be one, but we shouldn't crash if there are more than one subtile and they have different types.

PR #1489, opened 2020-08-19
Added block information for the IPIN/OPIN grid location in the overuse report

Bill-hbrhbr:
#### Related Issue
#1467 

#### Example report
[ipin_report.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5099061/ipin_report.txt)

#### Reproduce
1. cd to the VTR root directory
2. ./vtr_flow/scripts/run_vtr_flow.pl ./vtr_flow/benchmarks/verilog/diffeq2.v ./vtr_flow/arch/timing/k6_frac_N10_frac_chain_mem32K_40nm.xml -crit_path_router_iterations 100 --max_logged_overused_rr_nodes 50 --route_chan_width 32 --generate_rr_node_overuse_report on -temp_dir report_test
3. ls report_test/report_overused_nodes.rpt

#### Note
I added block information for OPINs as well. If this is not desired, I'll discard this feature.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/overuse_report.cpp
+++ b/vpr/src/route/overuse_report.cpp
@@ -55,6 +55,8 @@ void log_overused_nodes_status(int max_logged_overused_rr_nodes) {
  *
  * Print all the overused RR nodes' info in the report file report_overused_nodes.rpt.
  * The report generation is turned on by the VPR option: --generate_rr_node_overuse_report on.
+ * This report will be generated only if the last routing attempt fails, which
+ * causes the whole VPR flow to fail.
  */
 void report_overused_nodes() {
     const auto& device_ctx = g_vpr_ctx.device();
@@ -166,9 +168,9 @@ static void report_overused_ipin_opin(std::ostream& os, RRNodeId node_id) {
     os << "Number of blocks currently occupying this grid location = " << grid_info.usage << '\n';
 
     size_t iblock = 0;
-    for (size_t isubtile = 0; isubtile < grid_blocks.size(); ++isubtile) {
-        //Check if the block is empty
-        if (grid_info.subtile_empty[isubtile]) {
+    for (size_t isubtile = 0; isubtile < grid_info.blocks.size(); ++isubtile) {
+        //Check if there is a valid block at this subtile location
+        if (grid_info.subtile_empty(isubtile)) {
             continue;
         }
 
--- a/vpr/src/route/overuse_report.cpp
+++ b/vpr/src/route/overuse_report.cpp
@@ -4,6 +4,11 @@
 #include "globals.h"
 #include "vtr_log.h"
 
+static void report_overused_ipin_opin(std::ostream& os, RRNodeId node_id);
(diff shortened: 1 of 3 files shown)


vpr/src/route/overuse_report.cpp
```diff
@@ -4,6 +4,11 @@
 #include "globals.h"
```
vaughnbetz, line 4: Add a file level comment about what this file is for, and what kinds of routines should go in it.

vpr/src/route/overuse_report.cpp
```diff
@@ -121,6 +112,80 @@ void generate_overused_nodes_to_congested_net_lookup(std::map<RRNodeId, std::set
...
+    //Add block type for IPINs/OPINs in overused rr-node report
+    const auto& clb_nlist = g_vpr_ctx.clustering().clb_nlist;
+    auto& grid_info = place_ctx.grid_blocks[grid_x][grid_y];
+    auto& grid_blocks = grid_info.blocks;
+
+    os << "Grid location: X = " << grid_x << ", Y = " << grid_y << '\n';
+    os << "Number of blocks currently at this grid location = " << grid_info.usage << '\n';
+    for (size_t iblock = 0; iblock < grid_blocks.size(); ++iblock) {
+        ClusterBlockId block_id = grid_blocks[iblock];
+        os << "Block #" << iblock << ": ";
```
vaughnbetz, line 137: I think you need to add a check for empty blocks here.  grid_blocks.size gives the capacity of that (x,y) location (how many things could be placed there).  But you could have locations where nothing was placed.  See 
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/ed1a373d9c3409eea098612b6530f734c125ca03/vpr/src/util/vpr_utils.cpp#L192
and the code just above:
            for (int z = 0; z < device_ctx.grid[x][y].type->capacity; ++z) {
                grid_block.blocks[z] = EMPTY_BLOCK_ID;
            }
        }

So you should check for EMPTY_BLOCK_ID.  I think adding a short function (grid_block.empty(i)) to t_grid_block would be a nice way to code this, rather than doing a check for EMPTY_BLOCK_ID (that small function would localize the EMPTY_BLOCK_ID check).

Bill-hbrhbr:
Hi @vaughnbetz. Thank you for taking the time to review these codes (as well as the larger PR, which I will get to in a sec). I fixed the issues in your comments. To summarize:

1. Improved the file-level and function-level comments in overuse_report.h and overuse_report.cpp.
2. Updated all the major comments (which include the ones above) to Doxygen style. 
3. Added the subtile occupancy checking method in the t_grid_blocks structure and added comments.
4. Fixed the bug regarding encountering EMPTY_BLOCK_ID while printing out IPIN/OPIN block info.

On the other hand, the report format/functionality hasn't changed. 

Please review it again if you have time. If you are satisfied with all these changes, this PR should be good to go.

vaughnbetz:
Looks very good -- thanks Bill!

PR #1490, opened 2020-08-19
Refactored read_place and read_user_block_loc functions into two new …

sfkhalid:
…functions

Two new functions - read_place_header and read_place_body - were created to replace read_place and read_user_block_loc. Read_place and read_user_block_loc had some similar code that could be condensed into the same function, which was done in read_place_body.

#### Description
Read_place_header now reads in the first two lines of a place file, this used to be done in read_place. Read_place_body reads in either a place file (minus the header) or a constraints file, which used to be done in two separate functions: read_place and read_user_block_loc. 

Tested with fix_clusters option (which calls read_place_body) on individual circuits. Also tested that functions work when loading a placement file. Also checked that it passes basic and strong reg tests.

#### Checklist:
- [x ] All new and existing tests passed


vaughnbetz:
Discussed and reviewed in our meeting. Please apply the changes discussed and update the commits.  Some of the changes:
- have read_place and read_constraints top level functions, to simplify callers use.
- Comment how to use them and any things they expect in the header file, using Doxygen-compatible /** comments
- Avoid use of .x == OPEN to count how many times a block appeared in a file. Use a vector<bool> instead (after checking that the block id space is dense, from 0 to num_blocks-1 or so.
- Should remove duplicate header processing code in read_place_body by passing a file pointer in (has already gone past the header)
- Should check subtile legality
- Check if constraints need to set the grid locations or just the block.x,.y etc.  Avoid setting grid for place and constraint uses if possible, to minimize differences.
- Should print out "Reading file ..." and "Finished reading file ..." for both placement and constraint files.
- Summarize your testing of error conditions.
  - Missing header, duplicate header, header in wrong spot, ...
  - Missing block, duplicate block, name not found for block, x,y, subtile out of range, two blocks placed at same grid location.  All the errors should be caught in the file reader (for both place and constraint files) except two blocks at same grid location, which hopefully is caught in check_place right after file reading (but check this).

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +97,129 @@ void read_place(const char* net_file,
...
 
-    fclose(fp);
-    free_hash_table(hash_table);
-    VTR_LOG("Successfully read %s.\n", constraints_file);
-    VTR_LOG("\n");
+    if (!is_place_file) {
+        VTR_LOG("Successfully read %s.\n", place_file);
+        VTR_LOG("\n");
+    }
+    place_ctx.placement_id = vtr::secure_digest_file(place_file);
```
vaughnbetz, line 222: This is a bug; shouldn't generate the signature (used to verify that a routing file uses a compatible placement if we later read in that routing file) for the placement if we're not reading a placement file.

sfkhalid:
1. Read_place tests done: duplicate blocks listed, block out of range of grid dimensions, block locations incompatible, subtile number out of range, incomplete/invalid lines, two blocks placed at same location (caught by check_place)

2. Invalid block name - not sure how to check for this. Previously a hash table was used to store block names and IDs and checked if names were valid by seeing if they were in the hash table.

3. Testing place file and constraints file at same time - constraints file is not read at all because read_place is only called in a situation where placer_opts.doPlacement == STAGE_LOAD

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +146,113 @@ void read_place(const char* net_file,
...
-            if (place_ctx.block_locs.size() != cluster_ctx.clb_nlist.blocks().size()) {
-                //Resize if needed
-                place_ctx.block_locs.resize(cluster_ctx.clb_nlist.blocks().size());
-            }
-
-            //Set the location
-            place_ctx.block_locs[blk_id].loc.x = block_x;
-            place_ctx.block_locs[blk_id].loc.y = block_y;
-            place_ctx.block_locs[blk_id].loc.sub_tile = sub_tile_index;
+            seen_grid_dimensions = true; //if you have read the grid dimensions you are done reading the place file header
```
sfkhalid, line 149: Fix comment at this line

vpr/src/base/read_place.cpp
```diff
@@ -15,25 +16,78 @@
...
+
+    read_place_body(fstream, constraints_file, is_place_file);
+
+    VTR_LOG("Successfully read %s.\n", constraints_file);
+    VTR_LOG("\n");
+}
+
+/**
+ * This function reads the header (first two lines) of a placement file.
+ * It checks whether the packed netlist file that generated the placement matches the current netlist file.
```
sfkhalid, line 74: Say what the header is - two lines in any order that give netlist and grid files used to generate placement. Also mention grid gives an error if it doesn't match

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +146,113 @@ void read_place(const char* net_file,
...
+            if (sub_tile_index >= physical_tile->capacity || sub_tile_index < 0) {
+                VPR_THROW(VPR_ERROR_PLACE, place_file, vtr::get_file_line_number_of_last_opened_file(), "Block %d subtile number (%d) is out of range. \n", blk_id, sub_tile_index);
+            }
 
-        ptr = vtr::fgets(buf, vtr::bufsize, fp);
-    }
+            //need to lock down blocks  and mark grid block usage if it is a constraints file
+            if (!is_place_file) {
+                place_ctx.block_locs[blk_id].is_fixed = true;
+                place_ctx.grid_blocks[block_x][block_y].blocks[sub_tile_index] = blk_id;
```
sfkhalid, line 238: Try setting grid all the time, check that it works

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +146,113 @@ void read_place(const char* net_file,
...
-                    }
-                }
-            }
-        }
-    }
+    std::string line;
+    int lineno = 0;
 
-    ptr = vtr::fgets(buf, vtr::bufsize, fp);
+    std::vector<ClusterBlockId> seen_blocks;
```
sfkhalid, line 177: Comment what this is

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +146,113 @@ void read_place(const char* net_file,
...
-        if (h_ptr == nullptr) {
-            VTR_LOG_WARN("[Line %d] Block %s invalid, no such block.\n",
-                         vtr::get_file_line_number_of_last_opened_file(), bname);
-            ptr = vtr::fgets(buf, vtr::bufsize, fp);
-            continue;
-        }
-        ClusterBlockId bnum(h_ptr->index);
-        int i = xtmp;
-        int j = ytmp;
+            ClusterBlockId blk_id = cluster_ctx.clb_nlist.find_block(block_name);
```
sfkhalid, line 201: find what this returns if it doesn't find the block name - use for invalid block name check

vpr/src/base/read_place.cpp
sfkhalid: should be a vtr::vector_map<ClusterBlockId, int>

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +146,113 @@ void read_place(const char* net_file,
...
+            if (block_x < 0 || block_x > int(device_ctx.grid.width() - 1)
+                || block_y < 0 || block_y > int(device_ctx.grid.height() - 1)) {
+                VPR_THROW(VPR_ERROR_PLACE, "The block with ID %d is out of range at location (%d, %d). \n", blk_id, block_x, block_y);
+            }
 
-        place_ctx.block_locs[bnum].loc.x = i; /* Will be reloaded by initial_placement anyway. */
-        place_ctx.block_locs[bnum].loc.y = j; /* We need to set .x only as a done flag.         */
-        place_ctx.block_locs[bnum].loc.sub_tile = k;
-        place_ctx.block_locs[bnum].is_fixed = true;
+            if (place_ctx.block_locs.size() != cluster_ctx.clb_nlist.blocks().size()) {
```
sfkhalid, line 214: check if we need this. if we do, should be out of the loop

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +146,113 @@ void read_place(const char* net_file,
...
+                place_ctx.grid_blocks[block_x][block_y].usage++;
+            }
 
-    for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
-        auto result = constrained_blocks.find(blk_id);
-        if (result == constrained_blocks.end()) {
-            continue;
-        }
+            //add to vector of blocks that have been seen
+            seen_blocks.push_back(blk_id);
```
sfkhalid, line 243: change once you refactor to vector map

vpr/src/base/read_place.cpp
```diff
@@ -97,194 +146,113 @@ void read_place(const char* net_file,
...
-        if (place_ctx.block_locs[blk_id].loc.x == OPEN) {
-            VPR_THROW(VPR_ERROR_PLACE_F, constraints_file, 0,
-                      "Block %s location was not specified in the constraints file.\n", cluster_ctx.clb_nlist.block_name(blk_id).c_str());
+        } else {
+            //Unrecognized
+            vpr_throw(VPR_ERROR_PLACE_F, place_file, lineno,
+                      "Invalid line '%s' in file",
+                      line.c_str());
         }
     }
```
sfkhalid, line 252: add in check for place file to see if it has checked all blocks

vpr/src/base/read_place.cpp
```diff
@@ -99,192 +150,145 @@ void read_place(const char* net_file,
...
+                AtomBlockId atom_blk_id = atom_ctx.nlist.find_block(block_name);
 
-        if (i < 0 || i > int(device_ctx.grid.width() - 1) || j < 0 || j > int(device_ctx.grid.height() - 1)) {
-            VPR_THROW(VPR_ERROR_PLACE_F, constraints_file, 0,
-                      "Block #%zu (%s) location, (%d,%d) is out of range.\n", size_t(bnum), bname, i, j);
-        }
+                if (atom_blk_id == AtomBlockId::INVALID()) {
+                    VPR_THROW(VPR_ERROR_PLACE, "Block %s has an invalid name.\n", c_block_name);
+                } else {
+                    blk_id = atom_ctx.lookup.atom_clb(atom_blk_id);
```
sfkhalid, line 229: add comment explaining this

vpr/src/base/read_place.cpp
```diff
@@ -99,192 +150,145 @@ void read_place(const char* net_file,
...
+
+            if (sub_tile_index >= physical_tile->capacity || sub_tile_index < 0) {
+                VPR_THROW(VPR_ERROR_PLACE, "Block %s subtile number (%d) is out of range. \n", c_block_name, sub_tile_index);
+            }
+
+            if (!is_sub_tile_compatible(physical_tile, logical_block, place_ctx.block_locs[blk_id].loc.sub_tile)) {
+                VPR_THROW(VPR_ERROR_PLACE, "Attempt to place block %s with ID %d at illegal location (%d, %d). \n", c_block_name, blk_id, block_x, block_y);
+            }
+
+            //need to lock down blocks  and mark grid block usage if it is a constraints file
```
sfkhalid, line 262: explain why you need to do this (if place file, initial placement does instead)

vpr/src/base/read_place.cpp
```diff
@@ -99,192 +150,145 @@ void read_place(const char* net_file,
...
+            if (!is_place_file) {
+                place_ctx.block_locs[blk_id].is_fixed = true;
+                place_ctx.grid_blocks[block_x][block_y].blocks[sub_tile_index] = blk_id;
+                place_ctx.grid_blocks[block_x][block_y].usage++;
+            }
+
+            //mark the block as seen
+            seen_blocks[blk_id] = 1;
```
sfkhalid, line 270: make ++ instead of 1

vpr/src/base/read_place.cpp
```diff
@@ -99,192 +150,145 @@ void read_place(const char* net_file,
...
+            vpr_throw(VPR_ERROR_PLACE_F, place_file, lineno,
+                      "Invalid line '%s' in file",
+                      line.c_str());
         }
+    }
 
-        if (place_ctx.block_locs[blk_id].loc.x == OPEN) {
-            VPR_THROW(VPR_ERROR_PLACE_F, constraints_file, 0,
-                      "Block %s location was not specified in the constraints file.\n", cluster_ctx.clb_nlist.block_name(blk_id).c_str());
+    //For place files, check that all blocks have been read
```
sfkhalid, line 280: For constraints, don't need to include all add to comment

vpr/src/base/read_place.cpp
```diff
@@ -99,192 +150,145 @@ void read_place(const char* net_file,
...
+                VPR_THROW(VPR_ERROR_PLACE, "Block %d has not been read from the place file. \n", block_id);
+            }
         }
     }
 
-    fclose(fp);
-    free_hash_table(hash_table);
-    VTR_LOG("Successfully read %s.\n", constraints_file);
-    VTR_LOG("\n");
+    if (is_place_file) {
```
sfkhalid, line 289: add comment to explain why

sfkhalid: for entire place file, want to make hash so we can write into routing file for later error checking, dont need for constraint

vpr/src/base/read_place.h
```diff
@@ -1,16 +1,25 @@
 #ifndef READ_PLACE_H
 #define READ_PLACE_H
 
+/**
+ * This function is used to read a placement file.
```
sfkhalid, line 5: so that we can skip placement, check netlist file, grid dimensions, if verify flag is set must give error

vpr/src/base/read_place.h
```diff
@@ -1,16 +1,25 @@
...
+ */
 void read_place(
```
sfkhalid, line 7: doesnt need to be passed is_place_file

vpr/src/base/read_place.h
```diff
@@ -1,16 +1,25 @@
...
     const char* place_file,
     bool verify_file_hashes,
-    const DeviceGrid& grid);
+    const DeviceGrid& grid,
+    bool is_place_file);
+
+/**
+ * This function is used to read a constraints file that specifies the desired locations of blocks.
+ */
+void read_constraints(
```
sfkhalid, line 17: doesnt need is_place_file to be passed

vaughnbetz:
One more update to make:  make --fix_pins give a good message pointing at --fix_clusters if the user tries to specify a file name.

sfkhalid:
Earlier we said that if a location is specified twice for the same block (ex. by two different primitives in the same cluster) we don't have to give an error as long as the locations don't conflict. But in this case (location of a block specified twice with no conflict) check_place gives an error later on during the placement consistency check. Given this, should I give an error in read_place_body if a location is specified twice for the same cluster?

vaughnbetz:
What is the error from check_place?
I am guessing this line is the culprit, if you didn't change it (guard it with an if so that usage is only increased the first time a block location is read):
	                place_ctx.grid_blocks[block_x][block_y].usage++;
If you increment usage (how many blocks are at that grid location) each time you read a primitive in the same cluster, you will corrupt the placement state and fail check_place.  But avoiding that is not hard, with small code changes in read_place.

sfkhalid:
> What is the error from check_place?
> I am guessing this line is the culprit, if you didn't change it (guard it with an if so that usage is only increased the first time a block location is read):
> place_ctx.grid_blocks[block_x][block_y].usage++;
> If you increment usage (how many blocks are at that grid location) each time you read a primitive in the same cluster, you will corrupt the placement state and fail check_place. But avoiding that is not hard, with small code changes in read_place.

Yes, that was the problem. Adding the if statement as you suggested fixed it

sfkhalid:
Merging with one failure in Travis CI - Branch. This failure is just the Python Lint failure that has been popping up.

PR #1491, opened 2020-08-20
Vpr allow post fit netlist

helen-23:
To allow VPR to run the large TItan designs with the post-fit netlist, some changes to VPR and the architecture had to be made.

#### Description

architecture changes:
1. added stratixiv_clkena clock control primitive
2. added four more PLLs, for a total of 16 to model the global clocks on the Stratix IV device

VPR changes
1. fixed forever loop issue in auto device resizing. It is possible for the loop to get stuck if the design used too many resources of a particular type (more than the device limit), so it will keep looking for larger devices without finding one that could fit the resource. Added a check in there to limit the loop

#### Motivation and Context

Allows for running larger circuits with the post-fit VQM. This gives results closer to that of Quartus II.

Tested on the FFT and DLA Titan benchmark variants
Passed Travis CI tests

#### Types of changes

- [x ] Bug fix (change which fixes an issue)

#### Checklist:

- [x ] All new and existing tests passed


helen-23:
Hi Vaughn,
I have been holding on to this branch for quite some time. Thought it should be a good time to merge it in now. I have a few other branches to merge in after this as well.
Thank you!

vpr/src/base/SetupGrid.cpp
```diff
@@ -150,6 +150,17 @@ static DeviceGrid auto_size_device_grid(const std::vector<t_grid_def>& grid_layo
 
         VTR_ASSERT_SAFE_MSG(std::find_if(auto_layout_itr + 1, grid_layouts.end(), is_auto_grid_def) == grid_layouts.end(), "Only one <auto_layout>");
 
+        //Determine maximum device size to try before concluding that the circuit cannot fit on any device
+        //Calculate total number of required instances
+        //Then multiply by a factor of 100 as overhead
```
vaughnbetz, line 155: I think you're multiplying by a factor of 1000 (should update comment).
Should say why you're doing this -- to avoid infinite loops if increasing the grid size never gets you more of the instance type you need and hence never lets you fit.

vpr/src/base/SetupGrid.cpp
```diff
@@ -183,10 +195,18 @@ static DeviceGrid auto_size_device_grid(const std::vector<t_grid_def>& grid_layo
...
             //Increase the grid size
             width++;
 
-        } while (true);
+        } while (grid_size < max_size);
+
+        //Maximum device size reached
+        VPR_FATAL_ERROR(VPR_ERROR_OTHER,
+                        "Device auto-fit aborted: device size already exceeds required resources count by 100 times yet still cannot fit the design. "
+                        "Might be using more instances of a particular type of resource than the StratixIV devices can support (e.g. PLLs)\n");
```
vaughnbetz, line 209: Should make this 100 (really 10,000 now) a defined constant so the message and code are guaranteed to agree.
Should make this message a bit more generic -- "This may be due to resources that do not grow as the grid size increases (e.g. PLLs in the Titan Stratix IV architecture capture."

vtr_flow/arch/titan/stratixiv_arch.timing.xml
```diff
@@ -4408,6 +4408,14 @@
         <port name="vcounderrange" clock="inclk"/>
       </output_ports>
     </model>
+    <model name="stratixiv_clkena">
```
vaughnbetz, line 4411: Useful to comment what this is: the clkena block at the root of block trees.

helen-23: Thank you, Vaughn, for catching these issues. I have made the changes as recommended.

vpr/src/base/SetupGrid.cpp
```diff
@@ -24,6 +24,8 @@
 #include "SetupGrid.h"
 #include "vtr_expr_eval.h"
 
+#define MAX_SIZE_FACTOR 10000
```
vaughnbetz, line 27: Look good, but also good to comment what this is here (grid size will be no more than this factor * the number of clustered block instances we expect to need).

vaughnbetz:
Changes look good; just suggesting one more comment.
kokoru had a QoR failure but it's just because this test used to have a grid limited by both io and labs and now has a grid limited only by io.  Probably due to some small change in the grid update logic; I suggest taking a quick look at it and then updating the golden results if all looks OK (grid size seems reasonable).
regression_tests/vtr_reg_nightly/titan_other...[Fail]
 stratixiv_arch.timing.xml/leon2_stratixiv_arch_timing.blif/common device_limiting_resources: golden = io LAB result = io

vaughnbetz:
Travis CI failure is spurious: util.py failed python lint (bizarre, but can ignore).

helen-23:
Hi Vaughn,
After rerunning the tests, the Travis CI failure was gone. Also safe to update golden result that was causing the qor failure in the nightly tests.
Now all tests have passed.

PR #1494, opened 2020-08-21
Manual move generator

MahshadJ:
The manual move generator allows the user to set moves during placement.

#### Description

The manual move generator allows the user to set moves during placement. After indicating the move, the user is presented with the costs of the move and can choose whether they want to accept or reject it.

#### Motivation and Context

Currently, the moves in placement are chosen quite randomly, and this feature allows the user to set specific moves.

#### Types of changes

#### Checklist:


vpr/src/place/place.cpp
```diff
@@ -1463,10 +1494,23 @@ static e_move_result try_swap(float t,
...
+            //update all the costs in the manual_move_info variable and open cost summary window
+            manual_move_info->delta_c = delta_c;
+            manual_move_info->bb_delta_c = bb_delta_c;
+            manual_move_info->timing_delta_c = timing_delta_c;
+            cost_summary_window();
+            update_screen(ScreenUpdatePriority::MAJOR, " ", PLACEMENT, nullptr);
+        }
+
         /* 1 -> move accepted, 0 -> rejected. */
         move_outcome = assess_swap(delta_c, t);
```
Bill-hbrhbr, line 1507: If you want the cost summary window to display the suggested move outcome, you should move the ```assess_swap``` routine before ```cost_sumary_window()```.

vpr/src/place/place.cpp
```diff
@@ -1463,10 +1494,23 @@ static e_move_result try_swap(float t,
...
 
-        if (move_outcome == ACCEPTED) {
+        if (manual_move) {
+            manual_move_outcome = (manual_move_info->move_outcome == 0 ? REJECTED : ACCEPTED);
+        }
```
Bill-hbrhbr, line 1511: It'd be nice to wrap these lines into a separate routine as well, and pass the ```move_outcome``` by reference so that you can modify it to user's chosen value.

vpr/src/place/manual_move_generator.h
```diff
@@ -0,0 +1,13 @@
+#ifndef VPR_MANUAL_MOVE_GEN_H
+#define VPR_MANUAL_MOVE_GEN_H
+#include "move_generator.h"
+#include "globals.h"
+
+void mmg_get_manual_move_info(ManualMoveInfo mmi);
+
+class ManualMoveGenerator : public MoveGenerator {
```
Bill-hbrhbr, line 8: If you can modify the original move generator to achieve your manual move purpose, then this derived class is not needed.

vpr/src/base/globals.h
```diff
@@ -9,4 +9,20 @@
...
+struct ManualMoveInfo {
+    int block_id;
+    int to_x;
+    int to_y;
+    int subtile = 0;
+    double delta_c = 0;
+    double bb_delta_c = 0;
+    double timing_delta_c = 0;
+    t_user_move_outcome move_outcome;
+};
```
Bill-hbrhbr, line 26: You should probably not put this in ```globals.h```. Maybe in another header file.

vpr/src/place/place.cpp
MahshadJ: good point. resolved

vpr/src/draw/draw.cpp
```diff
@@ -4026,6 +4040,251 @@ static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointe
...
+    GtkWidget* info = gtk_label_new(NULL);
+    gtk_label_set_markup((GtkLabel*)info, "<b>Move Costs and Outcomes</b>");
+    std::string dc_label = "Delta Cost: " + std::to_string(manual_move_globals.draw_manual_move_info.delta_c);
+    GtkWidget* dc = gtk_label_new(dc_label.c_str());
+    std::string dcn_label = "Delta Cost Norm: " + std::to_string(manual_move_globals.draw_manual_move_info.bb_delta_c);
+    GtkWidget* dcn = gtk_label_new(dcn_label.c_str());
+    std::string dtcn_label = "Delta Timing Cost Norm: " + std::to_string(manual_move_globals.draw_manual_move_info.timing_delta_c);
+    GtkWidget* dtcn = gtk_label_new(dtcn_label.c_str());
+    std::string outcome;
+    manual_move_globals.draw_manual_move_info.move_outcome == 0 ? outcome = "Rejected" : outcome = "Accepted";
```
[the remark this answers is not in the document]
MahshadJ: I actually have enum values for that, hadn't changed it there yet. Thanks for pointing that out. There are two move_outcomes, a user_move_outcome which is set after the cost summary window pops up and a placer_move_outcome which is just displayed in the cost summary window.

vpr/src/place/manual_move_generator.cpp
```diff
@@ -0,0 +1,44 @@
+#include "manual_move_generator.h"
+#include <iostream>
+
+t_pl_loc to;
+int block_id;
+
+void mmg_get_manual_move_info(ManualMoveInfo mmi) {
+    block_id = mmi.block_id;
+    to = t_pl_loc(mmi.to_x, mmi.to_y, mmi.subtile);
+}
```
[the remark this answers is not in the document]
MahshadJ: no, this function gets info from draw.cpp, mmg gets info from placer

vpr/src/draw/draw.cpp
```diff
@@ -4026,6 +4040,251 @@ static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointe
...
+        invalid_breakpoint_entry_window("y value out of bounds");
+        valid_input = false;
+    }
+    if (to_subtile < 0) {
+        invalid_breakpoint_entry_window("invalid subtile value");
+        valid_input = false;
+    }
+
+    //checks if all entries are full
+    if (std::string(gtk_entry_get_text((GtkEntry*)block_entry)).size() == 0 || std::string(gtk_entry_get_text((GtkEntry*)to_x_entry)).size() == 0 || std::string(gtk_entry_get_text((GtkEntry*)to_y_entry)).size() == 0 || std::string(gtk_entry_get_text((GtkEntry*)subtile_entry)).size() == 0) {
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw.cpp
```diff
@@ -4026,6 +4040,251 @@ static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointe
...
+//returns a pointer to the global variable draw_manual_move_info so that the placer can modify it
+ManualMoveInfo* get_manual_move_info() {
+    return &manual_move_globals.draw_manual_move_info;
+}
+
+//returns true when a string only contains numbers and no letters or symbols
+//this function is used to differentiate between block_id and block_name
+bool string_is_a_number(std::string block_id) {
+    for (int i = 0; i < block_id.size(); i++) {
+        if (isdigit(block_id[i]) == 0)
```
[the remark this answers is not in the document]
MahshadJ: I can't take out the ==0 cause the function returns an int and not a bool. and when I tried taking the ==0 out it wouldn't function properly

vpr/src/draw/draw.cpp
```diff
@@ -4026,6 +4040,251 @@ static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointe
...
+        valid_input = false;
+    }
+    GtkWidget* to_x_entry = gtk_grid_get_child_at((GtkGrid*)oldGrid, 2, 1);
+    GtkWidget* to_y_entry = gtk_grid_get_child_at((GtkGrid*)oldGrid, 2, 2);
+    GtkWidget* subtile_entry = gtk_grid_get_child_at((GtkGrid*)oldGrid, 2, 3);
+    int to_loc_x = std::atoi(gtk_entry_get_text((GtkEntry*)to_x_entry));
+    int to_loc_y = std::atoi(gtk_entry_get_text((GtkEntry*)to_y_entry));
+    int to_subtile = std::atoi(gtk_entry_get_text((GtkEntry*)subtile_entry));
+
+    //check for input validity
```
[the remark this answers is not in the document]
MahshadJ: i believe that is the range i'm considering

vpr/src/draw/draw.cpp
```diff
@@ -4026,6 +4040,251 @@ static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointe
...
+    auto& device_ctx = g_vpr_ctx.device();
+    if (to_loc_x < 0 || to_loc_x > int(device_ctx.grid.width())) {
+        invalid_breakpoint_entry_window("x value out of bounds");
+        valid_input = false;
+    }
+    if (to_loc_y < 0 || to_loc_y > int(device_ctx.grid.height())) {
+        invalid_breakpoint_entry_window("y value out of bounds");
+        valid_input = false;
+    }
+    if (to_subtile < 0) {
```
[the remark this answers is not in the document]
MahshadJ: i tried this but it would error on subtile == 0 as well, I believe the -1 shouldn't be there. Does that make sense?

vpr/src/draw/draw.cpp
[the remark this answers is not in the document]
Bill-hbrhbr: I think you had > here. I changed it to >=

vpr/src/draw/draw.cpp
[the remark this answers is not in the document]
Bill-hbrhbr: subtile index should range from [0...capacity), so I don't think I would error.
Either < capacity or <= capacity - 1

vpr/src/place/manual_move_generator.h
Bill-hbrhbr: Ok, in that case, I think the best solution is to fully utilize the inheritance, and pass in the manual move generator starting from the placement_inner_loop() function. In this way, you can move the branching to the try_place() level, instead of in the try_swap() level.

vpr/src/draw/draw.cpp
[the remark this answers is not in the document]
Bill-hbrhbr: Did you add the negation mark when you take out ==0? If it still doesn't work, then it's a little weird. But this is only a stylistic issue so no need to worry about this one.

vpr/src/place/manual_move_generator.h
Bill-hbrhbr: Also, if you are sticking with making a derived class, then I think making the code work is more important and should be done first. It's better to leave the duplicated code alone for now.

vpr/src/draw/draw.cpp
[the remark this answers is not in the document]
MahshadJ: i check for when it's invalid however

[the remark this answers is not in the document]
Bill-hbrhbr: yeah, the valid range is ```[0, device_ctx.grid.width() - 1]```
so the invalid range is ```(-inf, 0) U [device_ctx.grid.width(), inf)```, which requires ```>=```
Same for ```device_ctx.grid.height()```

vpr/src/place/manual_move_generator.h
Bill-hbrhbr: After giving it more thought, I think the branching in try_swap might be necessary, since you are displaying the graphics and also generating two move outcomes (one from VPR and one from user). But you can still choose to pass in the manual move generator using inheritance.

Basically somewhere in placement_inner_loop's argument list, you have a type ```MoveGenerator& move_generator```
Changing it to a pointer type ```MoveGenerator* move_generator``` will allow you to pass in manual move generator here, having ```MoveGenerator*``` function as a base pointer.

vpr/src/base/globals.h
```diff
@@ -9,4 +9,20 @@
 
 extern VprContext g_vpr_ctx;
 
+typedef enum UserMoveOutcome {
```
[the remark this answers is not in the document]
vaughnbetz: Agreed. Mahshad we usually name enumerated types with an e_ and recent C++ versions allow the enum class declaration which is a bit easier to read.

vpr/src/base/globals.h
vaughnbetz: Yup, globals.h is for really major data structures everybody sees and really just includes the contexts now.  This definitely should not go in globals.h.

vpr/src/draw/draw.cpp
```diff
@@ -4026,6 +4040,219 @@ static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointe
...
+    GtkWidget* info = gtk_label_new(NULL);
+    gtk_label_set_markup((GtkLabel*)info, "<b>Move Costs and Outcomes</b>");
+    std::string dc_label = "Delta Cost: " + std::to_string(manual_move_globals.draw_manual_move_info.delta_c);
+    GtkWidget* dc = gtk_label_new(dc_label.c_str());
+    std::string dcn_label = "Delta Cost Norm: " + std::to_string(manual_move_globals.draw_manual_move_info.bb_delta_c);
+    GtkWidget* dcn = gtk_label_new(dcn_label.c_str());
+    std::string dtcn_label = "Delta Timing Cost Norm: " + std::to_string(manual_move_globals.draw_manual_move_info.timing_delta_c);
+    GtkWidget* dtcn = gtk_label_new(dtcn_label.c_str());
+    std::string outcome;
+    manual_move_globals.draw_manual_move_info.placer_move_outcome == MOVE_REJECTED ? outcome = "Rejected" : outcome = "Accepted";
```
vaughnbetz, line 4187: Better to make a little utility routine in place/move_utils that can convert an e_move_result to a string.  A simple way to do that is to add another enum value NUM_MOVE_RESULTS and make an array of strings 
std::string move_result_strings[NUM_MOVE_RESULTS] = {"rejected", "accepted", "aborted"};
Then make a function that is passed in an enum value and returns a string (say pl_move_type_to_str( )).

vpr/src/place/manual_move_generator.cpp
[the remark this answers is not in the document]
vaughnbetz: Need to comment what this routine does.  Not quite sure why the placer needs to send move information on the block_id or the to_location back to draw.cpp so you should think about whether that is necessary.  If it is, comment what the information flow is.

vpr/src/place/manual_move_generator.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    //Each x/y location possibly contains multiple sub tiles, so we need to pick
+    //a z location within a compatible sub tile.
+    auto to_type = g_vpr_ctx.device().grid[to.x][to.y].type;
+    auto& compatible_sub_tiles = compressed_block_grid.compatible_sub_tiles_for_tile.at(to_type->index);
+    if (std::find(compatible_sub_tiles.begin(), compatible_sub_tiles.end(), to.sub_tile) == compatible_sub_tiles.end()) {
+        std::cout << "Move aborted due to uncompatible subtile\n";
+        return e_create_move::ABORT;
+    }
+    rlim += 0;
```
vaughnbetz, line 38: Why is there an rlim += 0 here?  There's no impact, and it seems out of place.

vpr/src/place/place.cpp
```diff
@@ -1409,14 +1427,26 @@ static e_move_result try_swap(float t,
         rlim = std::numeric_limits<float>::infinity();
     }
 
+    if (manual_move) {
+        //pops up the manual move window for the user to input set their move
```
vaughnbetz, line 1431: I'd move these four lines into a subroutine so this code is as easy to read as possible (readers who don't want manual moves can simply skip over the if and call).

vpr/src/place/place.cpp
```diff
@@ -1396,6 +1410,10 @@ static e_move_result try_swap(float t,
 
     MoveOutcomeStats move_outcome_stats;
 
+    bool manual_move = get_manual_move_flag(); //whether the manual move info has been enabled or not
+    e_move_result manual_move_outcome;         //move_outcome from the user
```
vaughnbetz, line 1414: Move these declarations into the function I'm proposing below (better to declare variables right where they are used instead of ahead of time, plus that moves these into the proposed subroutine for readability).

vpr/src/place/place.cpp
```diff
@@ -1409,14 +1427,26 @@ static e_move_result try_swap(float t,
...
+        create_move_outcome = move_generator.propose_move(blocks_affected, rlim);
+    else
+        create_move_outcome = manual_move_generator.propose_move(blocks_affected, rlim);
 
     LOG_MOVE_STATS_PROPOSED(t, blocks_affected);
-
     e_move_result move_outcome = ABORTED;
 
-    if (create_move_outcome == e_create_move::ABORT) {
+    if (create_move_outcome == e_create_move::ABORT || !manual_move_info->valid_input) {
```
vaughnbetz, line 1449: Could have manual_moves return the ABORT type if the input is illegal, to simplify this if statement.  I think that would be better.

vpr/src/place/place.cpp
```diff
@@ -1426,6 +1456,7 @@ static e_move_result try_swap(float t,
                                "ABORTED", "illegal move");
 
         move_outcome = ABORTED;
```
vaughnbetz, line 1458: Should be able to merge this code too:  just have one move_outcome that is used for both manual and regular moves.

vpr/src/place/place.cpp
```diff
@@ -1466,7 +1497,26 @@ static e_move_result try_swap(float t,
         /* 1 -> move accepted, 0 -> rejected. */
         move_outcome = assess_swap(delta_c, t);
 
-        if (move_outcome == ACCEPTED) {
+        if (manual_move && manual_move_info->valid_input) {
```
vaughnbetz, line 1500: Move the manual_move code into a subroutine so the control routine (here) is simpler.

vpr/src/place/place.cpp
```diff
@@ -1466,7 +1497,26 @@ static e_move_result try_swap(float t,
...
+                manual_move_info->placer_move_outcome = MOVE_ABORTED;
+            cost_summary_window();
+            update_screen(ScreenUpdatePriority::MAJOR, " ", PLACEMENT, nullptr);
+        }
+
+        if (manual_move && manual_move_info->valid_input) {
+            manual_move_outcome = (manual_move_info->user_move_outcome == MOVE_REJECTED ? REJECTED : ACCEPTED);
+        }
+
+        if ((!manual_move && move_outcome == ACCEPTED) || (manual_move && manual_move_outcome == ACCEPTED)) {
```
vaughnbetz, line 1519: Should be able to simplify this code by just having one move_outcome which is set by either manual moves or regular moves.

vpr/src/place/place.cpp
```diff
@@ -1491,8 +1541,8 @@ static e_move_result try_swap(float t,
             /* Update clb data structures since we kept the move. */
             commit_move_blocks(blocks_affected);
 
-        } else { /* Move was rejected.  */
-                 /* Reset the net cost function flags first. */
+        } else if ((!manual_move && move_outcome == REJECTED) || (manual_move && manual_move_outcome == REJECTED)) { /* Move was rejected.  */
```
vaughnbetz, line 1544: Simplify by having one move_outcome state for all moves.

vpr/src/place/place.cpp
```diff
@@ -1514,7 +1564,10 @@ static e_move_result try_swap(float t,
                                (move_outcome ? "ACCEPTED" : "REJECTED"), "");
     }
 
-    move_outcome_stats.outcome = move_outcome;
+    if (!manual_move)
```
vaughnbetz, line 1567: Simplify by having one move outcome for all moves.

vpr/src/place/place.cpp
```diff
@@ -1526,7 +1579,11 @@ static e_move_result try_swap(float t,
     check_place(*costs, delay_model, place_algorithm);
 #endif
 
-    return (move_outcome);
+    manual_move_info->valid_input = true;
+    if (manual_move)
```
vaughnbetz, line 1583: Simplify by having one move outcome for all moves.

vaughnbetz:
Overall comment: I think there are some duplicated types and ifs in the placement code, so you can simplify / make more common code between the regular and manual moves.  Also to keep the (already complex) flow through the placement perturbations as simple as possible, make overall control routines that group together places where you have multiple lines processing manual moves and just call the overall control routine so the reader of the placement code just sees one line/call to something related to manual moves.

vpr/src/draw/draw.cpp
```diff
@@ -4026,6 +4040,251 @@ static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointe
...
+    //create all widgets
+    GtkWidget* grid = gtk_grid_new();
+    GtkWidget* info = gtk_label_new(NULL);
+    gtk_label_set_markup((GtkLabel*)info, "<b>Move Costs and Outcomes</b>");
+    std::string dc_label = "Delta Cost: " + std::to_string(manual_move_globals.draw_manual_move_info.delta_c);
+    GtkWidget* dc = gtk_label_new(dc_label.c_str());
+    std::string dcn_label = "Delta Cost Norm: " + std::to_string(manual_move_globals.draw_manual_move_info.bb_delta_c);
+    GtkWidget* dcn = gtk_label_new(dcn_label.c_str());
+    std::string dtcn_label = "Delta Timing Cost Norm: " + std::to_string(manual_move_globals.draw_manual_move_info.timing_delta_c);
+    GtkWidget* dtcn = gtk_label_new(dtcn_label.c_str());
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/base/globals.h
```diff
@@ -9,4 +9,23 @@
...
+    MOVE_REJECTED,
+    MOVE_ACCEPTED,
+    MOVE_ABORTED
+} t_move_outcome;
+
+struct ManualMoveInfo {
+    int block_id;
+    int to_x;
+    int to_y;
+    int subtile = 0;
```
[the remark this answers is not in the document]
MahshadJ: resolved

vpr/src/draw/draw.cpp
MahshadJ: resolved

vpr/src/base/globals.h
```diff
@@ -9,4 +9,23 @@
 
 extern VprContext g_vpr_ctx;
 
+typedef enum MoveOutcome {
```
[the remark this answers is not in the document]
MahshadJ: You're right, I deleted the MoveOutcome struct completely and moved the info struct to move_utils

vpr/src/base/globals.h
MahshadJ: resolved

vpr/src/place/manual_move_generator.cpp
MahshadJ: resolved

vpr/src/place/place.cpp
MahshadJ: resolved

vpr/src/place/place.cpp
MahshadJ: resolved

vpr/src/place/place.cpp
MahshadJ: resolved

vpr/src/place/move_utils.cpp
```diff
@@ -651,3 +651,16 @@ bool find_to_loc_uniform(t_logical_block_type_ptr type,
 
     return true;
 }
+
+std::string e_move_result_to_string(e_move_result move_outcome) {
```
MohamedElgammal, line 655: You don't need such a function. Just define an array/vector of strings that holds the names of your enum.
char* move_result_to_string[] = {"ACCEPTED", "REJECTED", "ABORTED"}

Then, you can convert the move result into string by addressing this array e.g. move_result_to_string[move_result]

vpr/src/place/place.cpp
```diff
@@ -325,6 +326,7 @@ static e_move_result try_swap(float t,
                               t_placer_prev_inverse_costs* prev_inverse_costs,
                               float rlim,
                               MoveGenerator& move_generator,
+                              ManualMoveGenerator& manual_move_generator,
```
MohamedElgammal, line 329: Why do you need to have two different parameters: MoveGenerator and ManualMoveGenerator while MoveGenerator is the base class of ManualMoveGenerator?

vpr/src/place/move_utils.cpp
MohamedElgammal: Duplicated comment. Vaughn already highlighted that.

vaughnbetz: Agreed.  I still like a function wrapping the access to the array of strings though (so keep a utility function to turn move types into strings), but the indexing into an array is generally easier to read and probably a bit faster than a bunch of if statements.

MahshadJ: resolved

vaughnbetz:
Assigning to Mohamed to take over and complete this one.

vaughnbetz:
This was completed and merged by @PaulaPerdomo .  Closing.

PR #1501, opened 2020-08-24
Add raw setup slack analysis to placement quench

Bill-hbrhbr:
#### Related Issue
simplified version of #1450 for easier code review and merging.

#### Copy of the note in the original PR
Hi @HackerFoo. Please checkout this draft PR to see if it's easier for you to review. 
New features:
1. ```try_swap()``` routine along with ```analyze_setup_slack_cost()``` routine.
2. The new ```PlacerSetupSlacks``` class in ```timing_place.h/cpp```
3. Three new timing update related routines: ```update_timing_classes()```, ```update_timing_cost()```, ```perform_full_timing_update()``` in ```place.cpp```.
4. VPR option to turn on setup slack analysis during placement quench: ```--place_quench_metric {auto, timing_cost, setup_slack}```

My steps for this PR:
1. ```PlacerSetupSlacks``` is an imitation of the class ```PlacerCriticalities```, and is utilized to get setup slacks from Kevin's Tatum timing analyzer.

2. I changed the original ```recompute_criticalities()``` to ```update_setup_slacks_and_criticalities()``` so that both categories of values can be updated.

3. One problem now occurs. Originally, each time the timing graph is updated by calling ```timing_info->update()```, the criticalities are updated as well. These updates use incremental techniques so that they only checkout modified atom pins returned by the timing analyzer. However, now the timing graph might be updated only for the setup slacks (without having criticalities updated). On the next iteration of criticalities update, it cannot use the incremental technique anymore, since more pins will have been modified than the ones the timing analyzer currently returns. In this case, the criticalities have to be recomputed from scratch (going through each sink pin).

4. To resolve this issue, I created the ```t_timing_update_mode``` class to keep track of the slack/criticality update status. ***No documentation on this structure at this commit.***

5. Now with the timing stuff down, I need to implement the setup slack analysis during the placement quench stage. So I made a vpr option ```--place_quench_metric``` to ask the user whether or not to turn on the new analysis technique in the quench.

6. I modified the ```try_swap()``` routine so that, if using the slack analysis, it will do a timing_update to get slacks, analyze, and then decide whether or not to accept the move based on the slack improvement. If accepted, the timing update will be kept; if rejected, I revert the timing update so that the slacks in the timing graph returns to their original values. 
Note that these timing updates will not involve changing criticalities (hence I had the consideration for point #3 and #4), and I've made a lot of testings to make sure that the timing update reversion works.


vpr/src/place/place.cpp
```diff
@@ -321,15 +333,18 @@ static void update_move_nets(int num_nets_affected);
...
                               float rlim,
                               MoveGenerator& move_generator,
-                              TimingInfo* timing_info,
+                              SetupTimingInfo* timing_info,
                               ClusteredPinTimingInvalidator* pin_timing_invalidator,
                               t_pl_blocks_to_be_moved& blocks_affected,
                               const PlaceDelayModel* delay_model,
-                              const PlacerCriticalities* criticalities,
+                              PlacerCriticalities* criticalities,
+                              PlacerSetupSlacks* setup_slacks,
```
HackerFoo, line 347: `criticalities` and `setup_slacks` are always passed together, right? Maybe they should be packaged together in another class or struct.

Bill-hbrhbr: I will merge related arguments together in another PR as discussed in the meeting.

vpr/src/place/place.cpp
```diff
@@ -1674,6 +1927,64 @@ static void update_td_delta_costs(const PlaceDelayModel* delay_model,
...
+        original_setup_slacks.push_back(connection_setup_slack[net_id][ipin]);
+        proposed_setup_slacks.push_back(setup_slacks->setup_slack(net_id, ipin));
+    }
+
+    //Sort in ascending order, from worse slack value to best
+    std::sort(original_setup_slacks.begin(), original_setup_slacks.end());
+    std::sort(proposed_setup_slacks.begin(), proposed_setup_slacks.end());
+
+    //Check the first pair of slack values that are different
+    //If found, return their difference
```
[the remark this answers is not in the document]
Bill-hbrhbr: Added function level description.

vpr/src/place/place.cpp
```diff
@@ -994,29 +1050,135 @@ static void outer_loop_recompute_criticalities(const t_placer_opts& placer_opts,
...
+    /* Update the placer's criticalities (e.g. sharpen with crit_exponent). */
     criticalities->update_criticalities(timing_info, crit_exponent);
 
-    //Update connection, net and total timing costs based on new criticalities
+    /* Update the placer's raw setup slacks. */
+    setup_slacks->update_setup_slacks(timing_info);
+
+    /* Clear invalidation state. */
+    pin_timing_invalidator->reset();
+}
```
Bill-hbrhbr, line 1126: Important documentation for review

vpr/src/place/place.cpp
```diff
@@ -994,29 +1050,135 @@ static void outer_loop_recompute_criticalities(const t_placer_opts& placer_opts,
...
+                          pin_timing_invalidator);
+
+    /* Update the timing cost with new connection criticalities. */
+    update_timing_cost(delay_model,
+                       criticalities,
+                       &costs->timing_cost);
+
+    /* Commit the setup slacks since they are updated. */
+    commit_setup_slacks(setup_slacks);
 }
```
Bill-hbrhbr, line 1182: Important documentation for review

vpr/src/place/place.cpp
```diff
@@ -1674,6 +1932,82 @@ static void update_td_delta_costs(const PlaceDelayModel* delay_model,
...
+
+        if (slack_diff != 0) {
+            return slack_diff;
+        }
+    }
+
+    //If all slack values are identical (or no modified slack values),
+    //reject this move by returning an arbitrary positive number as cost.
+    return 1;
+}
```
Bill-hbrhbr, line 2009: Important documentation for review

vpr/src/place/place.cpp
```diff
@@ -1775,45 +2109,78 @@ static void comp_td_connection_delays(const PlaceDelayModel* delay_model) {
...
+
+    //Incremental: only go through sink pins with modified setup slack
+    auto clb_pins_modified = setup_slacks->pins_with_modified_setup_slack();
+    for (ClusterPinId pin_id : clb_pins_modified) {
+        ClusterNetId net_id = clb_nlist.pin_net(pin_id);
+        size_t pin_index_in_net = clb_nlist.pin_net_index(pin_id);
+
+        connection_setup_slack[net_id][pin_index_in_net] = setup_slacks->setup_slack(net_id, pin_index_in_net);
+    }
+}
```
Bill-hbrhbr, line 2140: Important documentation for review

vpr/src/place/place.cpp
```diff
@@ -1775,45 +2109,78 @@ static void comp_td_connection_delays(const PlaceDelayModel* delay_model) {
...
+    //Go through every single sink pin to check that the slack values are the same
+    for (ClusterNetId net_id : clb_nlist.nets()) {
+        for (size_t ipin = 1; ipin < clb_nlist.net_pins(net_id).size(); ++ipin) {
+            if (connection_setup_slack[net_id][ipin] != setup_slacks->setup_slack(net_id, ipin)) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
```
Bill-hbrhbr, line 2165: Important documentation for review

vpr/src/place/timing_place.h
```diff
@@ -55,40 +97,148 @@ class PlacerCriticalities {
...
+ *
+ * RAW setup slacks are unlike criticalities. Their values are not confined between
+ * 0 and 1. Their values can be either positive or negative.
+ *
+ * This class also provides iterating over the clustered netlist connections/pins that
+ * have modified setup slacks by the last call to update_setup_slacks(). However, this
+ * utility is mainly used for incrementally committing the setup slack values into the
+ * structure `connection_setup_slack` used by many placer routines.
+ */
+class PlacerSetupSlacks {
```
Bill-hbrhbr, line 176: Important documentation for review

vpr/src/place/timing_place.h
```diff
@@ -14,32 +43,45 @@ std::unique_ptr<PlaceDelayModel> alloc_lookups_and_criticalities(t_chan_width_di
...
- * To support incremental re-calculation the class saves the last criticality exponent
- * passed to update_criticalites(). If the next update uses the same exponent criticalities
- * can be incrementally updated. Otherwise they must be re-calculated from scratch, since
- * a change in exponent changes *all* criticalities.
+ * To support incremental re-calculation, the class saves the last criticality exponent
+ * passed to PlacerCriticalities::update_criticalites(). If the next update uses the same
+ * exponent, criticalities can be incrementally updated. Otherwise, they must be re-calculated
+ * from scratch, since a change in exponent changes *all* criticalities.
  */
 class PlacerCriticalities {
```
Bill-hbrhbr, line 86: Important documentation for review

vpr/src/timing/timing_util.cpp
```diff
@@ -579,6 +582,21 @@ float calculate_clb_net_pin_criticality(const SetupTimingInfo& timing_info, cons
...
+    //There may be multiple atom netlist pins connected to this CLB pin
+    float clb_pin_setup_slack = std::numeric_limits<float>::infinity();
+    for (const auto atom_pin : pin_lookup.connected_atom_pins(clb_pin)) {
+        //Take the worst/minimum of the atom pin slack as the CLB pin slack
+        clb_pin_setup_slack = std::min(clb_pin_setup_slack, timing_info.setup_pin_slack(atom_pin));
+    }
+
+    return clb_pin_setup_slack;
+}
+
```
Bill-hbrhbr, line 599: Important documentation for review

vpr/src/place/place.cpp
```diff
@@ -994,29 +1050,135 @@ static void outer_loop_recompute_criticalities(const t_placer_opts& placer_opts,
...
+                               const PlacerCriticalities* criticalities,
+                               double* timing_cost) {
 #ifdef INCR_COMP_TD_COSTS
-    update_td_costs(delay_model, *criticalities, &costs->timing_cost);
+    update_td_costs(delay_model, *criticalities, timing_cost);
 #else
-    comp_td_costs(delay_model, *criticalities, &costs->timing_cost);
+    comp_td_costs(delay_model, *criticalities, timing_cost);
 #endif
+}
```
Bill-hbrhbr, line 1148: Important doc for review

Bill-hbrhbr:
Titan benchmark @commit d88a38d: https://drive.google.com/file/d/1ZJzgyoVCQTIohAOHhZRGS8PmK26tyQ-x/view?usp=sharing

The place time still has degradation. I will comment out the commit_setup_slacks() routine as discussed and measure the new results.

vaughnbetz:
Thanks Bill.  In this run place time is 6.8% higher.  So ~7% place time increase seems consistent.  Other metrics look unaffected again.  The comment out setup slacks experiment will be useful to see.

Bill-hbrhbr:
Hi @vaughnbetz, I cannot see the partial review. Maybe the comments got lost? (I hope not...)

vaughnbetz:
Whoops you're right Bill; they are in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1450#pullrequestreview-477245679

Bill-hbrhbr:
Titan benchmark @commit d88a38d: https://drive.google.com/file/d/1oPvMIt3Eg5CVrY1veyFiotZxZZViv3DG/view?usp=sharing

I commented out the commit_setup_slacks() routine but forgot to call PlacerSetupSlacks->disable_update(). The place time still shows quite a bit of degradation. I will now completely turn off all slack related updates and run the benchmarks again.

vaughnbetz:
Yup, that still has the same (7.3%) placement time increase, so it eliminates that routine from the suspect list.

Bill-hbrhbr:
Hi @vaughnbetz, I ran another set of benchmarks with setup slacks update turned off for PlacerSetupSlacks, and I'm still getting the placer time degradation. It's weird. I'll go investigate some more.
On the other hand, I've reflected the comments you made in #1450 to this PR. Just fixing some small bugs and you should be able to see them soon.

vaughnbetz:
Thanks Bill.  If it turns out this is just a consequence of splitting up and cleaning up place.cpp I'd still merge it.  Obviously nice if we can find one or two changes that led to it and grab it back though!

Bill-hbrhbr:
Titan benchmark #run003 @commit d88a38d: https://drive.google.com/file/d/1LLtuIQ93gCg_aZw-idx8MIYcWpVt0K4o/view?usp=sharing

Difference:
1. Commented out ```commit_setup_slacks()```
2. Disabled **PlacerSetupSlacks** update by calling ```PlacerSetupSlacks::disable_update()```

I decided to share this result anyway. If you look for the time spent on timing analysis (the last 4 columns in this spreadsheet) you will find that two runs basically match each other's numbers. This is to be expected.

Yet the placer still gets runtime degradation by the roughly same amount as before. I've come to the conclusion that the added branching in the ```try_swap()``` is the main culprit. I'm going to get rid of the branching and do a final benchmarking run. If there is still degradation, then I probably need somebody else to look for where the runtime has increased.

On the other hand, if it is indeed the branching that is increasing the placer runtime, then I should say that I don't have adequate knowledge of the compiler/OS to do micro-optimizations. So somebody should probably come back to this PR in the future and optimize the new ```try_swap()``` routine.

Bill-hbrhbr:
I also ran profiling with two Titan benchmarks: **neuron** and **stereovision**. I ran them with ```--pack --place`` so as to get rid of the routing time in the final profile visualizations. I also didn't turn off the timing updates/commits to setup slacks. Here are the results:

**neuron**:
_Original_: https://drive.google.com/file/d/1ZfjKb1BmHoNo8GycULH8mjyPeOPypST7/view?usp=sharing
_New_: https://drive.google.com/file/d/1UVD1m5INESN5ZbIRvKAcPkOsdK_MnN1_/view?usp=sharing

**stereovision**:
_Original_: https://drive.google.com/file/d/1h__uurJmwEXEfzbfYobXMnCypZGNt6Ln/view?usp=sharing
_New_: https://drive.google.com/file/d/1HbmSyPUeRtnd1u4_KqCyw5CiBsBtT8iq/view?usp=sharing

I didn't draw meaningful conclusions from these results, as the ```try_swap()``` routine is kind of unstable in terms of taking up CPU runtime. However, it should become even more obvious that the timing updates are not degrading the placer runtime. If they take up a portion of the CPU time that is significant enough, related routines would show up on the profile (but they didn't).

Not being able to draw meaningful conclusions probably means that the benchmarks that I'm profiling are still way too small in size to create a noticeable difference between the ```try_swap()``` routine and other parts of VPR. I will simply do what I suggested in the previous comment, and make a concluding remark when that's done.

vpr/src/base/read_options.cpp
```diff
@@ -1680,8 +1684,14 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
...
     place_grp.add_argument<e_place_algorithm, ParsePlaceAlgorithm>(args.PlaceAlgorithm, "--place_algorithm")
         .help("Controls which placement algorithm is used")
-        .default_value("path_timing_driven")
-        .choices({"bounding_box", "path_timing_driven"})
+        .default_value("criticality_timing")
+        .choices({"bounding_box", "criticality_timing", "slack_timing"})
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
+    place_grp.add_argument<e_place_algorithm, ParsePlaceAlgorithm>(args.PlaceQuenchAlgorithm, "--place_quench_algorithm")
+        .help("Controls which placement algorithm is used during placement quench")
```
HackerFoo, line 1692: Please add a quick description of each option that can be used to decide which algorithm to use.

vpr/src/base/vpr_types.h
```diff
@@ -829,36 +829,84 @@ struct t_annealing_sched {
...
+ * to the placer algorithms if more timing driven placement strategies
+ * are added in tht future. This method is used across various placement
+ * setup files, and it can be useful for major placer routines as well.
+ *
+ * More methods can be added to this class if the placement strategies
+ * will be further divided into more categories the future.
+ *
+ * Also supports assignments and comparisons between t_place_algorithm
+ * and e_place_algorithm so as not to break down previous codes.
+ */
```
HackerFoo, line 872: What is the justification for this over the following?

```
bool is_timing_driven(e_place_algorithm algo) {
  return algo == CRITICALITY_TIMING_PLACE || algo == SLACK_TIMING_PLACE; 
}
```

What methods do you propose should be added to this class? Do they improve the current code? If so, please add them.

vpr/src/place/place.cpp
```diff
@@ -1089,7 +1251,7 @@ static void placement_inner_loop(float t,
             num_swap_rejected++;
         }
 
-        if (placer_opts.place_algorithm == PATH_TIMING_DRIVEN_PLACE) {
+        if (placer_opts.place_algorithm.is_timing_driven()) {
             /* Do we want to re-timing analyze the circuit to get updated slack and criticality values?
```
HackerFoo, line 1255: Both `place_algorithm` and `placer_opts.place_algorithm` are used?

vpr/src/place/place.cpp
```diff
@@ -1047,9 +1202,11 @@ static void placement_inner_loop(float t,
                                  ClusteredPinTimingInvalidator* pin_timing_invalidator,
                                  const PlaceDelayModel* delay_model,
                                  PlacerCriticalities* criticalities,
+                                 PlacerSetupSlacks* setup_slacks,
                                  MoveGenerator& move_generator,
                                  t_pl_blocks_to_be_moved& blocks_affected,
-                                 SetupTimingInfo* timing_info) {
+                                 SetupTimingInfo* timing_info,
+                                 const t_place_algorithm& place_algorithm) {
```
HackerFoo, line 1209: This overrides `placer_opts`? It's not clear how this works.

vpr/src/place/place.cpp
```diff
@@ -1290,14 +1451,16 @@ static bool update_annealing_state(t_annealing_state* state,
...
+                        t_placer_costs* costs,
                         t_placer_prev_inverse_costs* prev_inverse_costs,
                         t_annealing_sched annealing_sched,
                         int max_moves,
                         float rlim,
                         const PlaceDelayModel* delay_model,
-                        const PlacerCriticalities* criticalities,
-                        TimingInfo* timing_info,
+                        PlacerCriticalities* criticalities,
+                        PlacerSetupSlacks* setup_slacks,
```
HackerFoo, line 1462: `criticalities` and `setup_slacks` should be bundled together because they used together.

vpr/src/place/place.cpp
```diff
@@ -1463,21 +1635,64 @@ static e_move_result try_swap(float t,
...
+        } else if (place_algorithm == CRITICALITY_TIMING_PLACE) {
             /*in this case we redefine delta_c as a combination of timing and bb.  *
              *additionally, we normalize all values, therefore delta_c is in       *
              *relation to 1*/
 
             delta_c = (1 - timing_tradeoff) * bb_delta_c * prev_inverse_costs->bb_cost
                       + timing_tradeoff * timing_delta_c * prev_inverse_costs->timing_cost;
-        } else {
+
+        } else { //place_algorithm == BOUNDING_BOX_PLACE (wiring cost)
```
HackerFoo, line 1695:
```
VTR_ASSERT(place_algorithm == BOUNDING_BOX_PLACE);
```

vpr/src/place/place.cpp
```diff
@@ -1695,6 +1946,82 @@ static void update_td_delta_costs(const PlaceDelayModel* delay_model,
...
+    }
+
+    //Sort in ascending order, from the worse slack value to the best
+    std::sort(original_setup_slacks.begin(), original_setup_slacks.end());
+    std::sort(proposed_setup_slacks.begin(), proposed_setup_slacks.end());
+
+    //Check the first pair of slack values that are different
+    //If found, return their difference
+    for (size_t idiff = 0; idiff < original_setup_slacks.size(); ++idiff) {
+        float slack_diff = original_setup_slacks[idiff] != proposed_setup_slacks[idiff];
```
HackerFoo, line 2013: Should this be:

```
float slack_diff = original_setup_slacks[idiff] - proposed_setup_slacks[idiff];
```

(subtract instead of not equals)

HackerFoo (changes requested):

vpr/src/place/place.cpp
Bill-hbrhbr: Sorry, it should be ```place_algorithm``` here instead of ```placer_opts.place_algorithm```

vpr/src/place/place.cpp
Bill-hbrhbr: ```placement_inner_loop()``` is called for both normal annealing and the quench. The normal annealing now uses ```placer_opts.place_algorithm``` and the quench now uses ```placer_opts.place_quench_algorithm```. Instead of creating a new separate function, I have a new parameter ```place_algorithm``` passed in to override ```placer_opts.place_algorithm```

vpr/src/place/place.cpp
Bill-hbrhbr: Thanks for the catch!

vpr/src/place/place.cpp
Bill-hbrhbr: Added.

vpr/src/base/vpr_types.h
Bill-hbrhbr: 1. The timing driven check is performed in several places other than ```place.cpp```. So I incorporated this check into ```placer_opts```.
2. There might be future categorizations of placement algorithms/strategies. Currently, there are only two categories (timing and no timing). Having this class will make it easier to describe and implement new categorizations in the future. If you don't think this is necessary, I'll revert the change.

vpr/src/place/place.cpp
Bill-hbrhbr: I want to do this in another PR, where I will refactor place.cpp. But if you want it done here, I will implement it. On a side note, how would you prefer naming this new structure?

vpr/src/base/read_options.cpp
Bill-hbrhbr: Added.

vpr/src/place/timing_place.h
```diff
@@ -55,40 +97,148 @@ class PlacerCriticalities {
...
     vtr::vec_id_set<ClusterPinId> cluster_pins_with_modified_criticality_;
+
+    ///@brief Updates criticalities: incremental V.S. from scratch
+    void incr_update_criticalities(const SetupTimingInfo* timing_info);
+    void recompute_criticalities();
+
+    ///@brief Flag that turns on/off the update_criticalities() routine.
+    bool update_enabled = true;
+
+    ///@brief Flag that checks if criticalities needs to be recomputed for all connections.
```
vaughnbetz, line 155: Add "... i.e. they cannot be recomputed incrementally for only connections whose slacks have been updated in the last timing analysis"
(also, typo:  needs -> need)

vpr/src/place/timing_place.h
```diff
@@ -55,40 +97,148 @@ class PlacerCriticalities {
...
+     * @brief The matrix that stores raw setup slack values for each connection.
+     *
+     * Index range: [0..cluster_ctx.clb_nlist.nets().size()-1][1..num_pins-1]
+     */
+    ClbNetPinsMatrix<float> timing_place_setup_slacks_;
+
+    ///@brief Set of pins with raw setup slacks modified by last call to update_setup_slacks()
+    vtr::vec_id_set<ClusterPinId> cluster_pins_with_modified_setup_slack_;
+
+    ///@brief Updates setup slacks: incremental V.S. from scratch.
```
vaughnbetz, line 226: Please carry over any commenting enhancements for the placer_criticalities based on that discussion to here as well.

vpr/src/place/timing_place.h
```diff
@@ -99,53 +249,53 @@ class PlacerCriticalities {
...
  * calculated (and unchanged) value.
  *
- * For a circuit with 'K' connections, of which 'k' have changed (typically k << K), this can 
+ * For a circuit with 'K' connections, of which 'k' have changed (typically k << K), this can
  * be done in O(k log K) time.
  *
- * It is important to note that due to limited floating point precision, floating point 
+ * It is important to note that due to limited floating point precision, floating point
  * arithmetic has an order dependence (due to round-off). Using a binary tree to total
  * the timing connection costs allows us to incrementally update the total timign cost while
```
vaughnbetz, line 286: typo : timign cost

Bill-hbrhbr:
Hi @vaughnbetz, thank you for taking the time to look at the new changes. I will send you an e-mail soon regarding any leftover concerns for this PR.

vpr/src/place/timing_place.cpp
```diff
@@ -14,71 +19,49 @@
 
 #include "timing_info.h"
 
-//Use an incremental approach to updaing criticalities?
-constexpr bool INCR_UPDATE_CRITICALITIES = true;
-
-/**************************************/
+///@brief Use an incremental approach to updating criticalities and setup slacks?
```
vaughnbetz, line 22: Explain what this means.  Does it control incremental update_timing calls, or recomputing only the criticalities and setup slacks for connections whose slack changed in the last timing analysis, or both?

vpr/src/place/timing_place.cpp
```diff
@@ -90,16 +73,170 @@ void PlacerCriticalities::update_criticalities(const SetupTimingInfo* timing_inf
...
+ *
+ * Since timing_info uses shifted/relaxed criticality (which depends on max required
+ * time and worst case slacks), additional nodes may be modified when updating the
+ * atom pin criticalities.
+ */
+
+void PlacerCriticalities::incr_update_criticalities(const SetupTimingInfo* timing_info) {
+    cluster_pins_with_modified_criticality_.clear();
+
+    for (AtomPinId atom_pin : timing_info->pins_with_modified_setup_criticality()) {
```
vaughnbetz, line 97: This code and the code to find the pins_with_modified_setup_slack is almost the same, except one uses the timing_info->pins_with_modified_setup_criticality() and the other uses timing_info->pins_with_modified_setup_slack().
1. Do the timing_info->pins_with_modified_setup_criticality() and pins_with_modified_setup_slack() differ?  Why?  I expect they are close to the same; could we just use whichever one is the bigger set (maybe criticality, due to slack shifting) to track both slack and criticality pins that need updates?  Would simplify the code and likely make it faster if we are using both criticalities and slacks, if indeed the pin update lists are usually the same (less work and less memory footprint).

vaughnbetz: If we can just use the bigger of pins_with_modified_criticality_ or pins_with_modified_setup_slack, some data members and some duplicated code can be removed.

vpr/src/place/place.cpp
```diff
@@ -1509,14 +1734,40 @@ static e_move_result try_swap(float t,
...
 
-            if (place_algorithm == PATH_TIMING_DRIVEN_PLACE) {
+            if (place_algorithm == SLACK_TIMING_PLACE) {
+                //Revert the timing delays and costs to pre-update values
+                //These routines must be called after reverting the block moves
+                //TODO: make this process incremental
+                comp_td_connection_delays(delay_model);
+                comp_td_costs(delay_model, *criticalities, &costs->timing_cost);
+
+                //Re-invalidate the affected sink pins
```
vaughnbetz, line 1752: Say why (since we did a timing update with the proposed move locations, which was rejected)

vpr/src/place/place.cpp
```diff
@@ -1392,17 +1562,19 @@ static void reset_move_nets(int num_nets_affected) {
...
                               t_pl_blocks_to_be_moved& blocks_affected,
                               const PlaceDelayModel* delay_model,
-                              const PlacerCriticalities* criticalities,
+                              PlacerCriticalities* criticalities,
+                              PlacerSetupSlacks* setup_slacks,
                               float rlim_escape_fraction,
-                              enum e_place_algorithm place_algorithm,
+                              const t_place_algorithm& place_algorithm,
                               float timing_tradeoff) {
     /* Picks some block and moves it to another spot.  If this spot is   *
```
vaughnbetz, line 1579: This comment should go in front of the function header in the refactored code  (you probably did that already).

vpr/src/place/place.cpp
```diff
@@ -1463,21 +1635,64 @@ static e_move_result try_swap(float t,
...
                                                                     timing_delta_c);
-        if (place_algorithm == PATH_TIMING_DRIVEN_PLACE) {
+
+        //Find all the sink pins with changed connection delays from the affected blocks.
+        //These sink pins will be passed into the pin_timing_invalidator for timing update.
+        //They will also be added to the pin invalidator when we wish to revert a timing update.
+        std::vector<ClusterPinId> sink_pins_affected;
+        find_affected_sink_pins(blocks_affected, sink_pins_affected);
+
+        if (place_algorithm == SLACK_TIMING_PLACE) {
```
vaughnbetz, line 1656: I think an overview comment before this big if would be helpful.  The blocks have been temporarily moved, and now we're going to assess cost. We need to update all timing etc. information that would impact the cost, but remember what we did and undo if we reject the move.

vpr/src/place/place.cpp
```diff
@@ -1695,6 +1946,82 @@ static void update_td_delta_costs(const PlaceDelayModel* delay_model,
     }
 }
 
+static void find_affected_sink_pins(const t_pl_blocks_to_be_moved& blocks_affected,
```
vaughnbetz, line 1949: Needs a function level comment.

vpr/src/place/place.cpp
```diff
@@ -1796,45 +2123,78 @@ static void comp_td_connection_delays(const PlaceDelayModel* delay_model) {
...
-                //This pin is a net sink on a moved block
-                VTR_ASSERT_SAFE(cluster_ctx.clb_nlist.pin_type(pin_id) == PinType::SINK);
-
-                /* The following "if" prevents the value from being updated twice. */
-                if (!driven_by_moved_block(net_id, blocks_affected)) {
-                    int net_pin = cluster_ctx.clb_nlist.pin_net_index(pin_id);
+    //Go through all the sink pins affected
+    for (ClusterPinId pin_id : blocks_affected.affected_pins) {
+        ClusterNetId net_id = clb_nlist.pin_net(pin_id);
+        int ipin = clb_nlist.pin_net_index(pin_id);
```
vaughnbetz, line 2190: This is one of the few code differences in the base (CRITICALITY_PLACE) algorithm in the new code vs. the old code.  Code looks fine (cleaner), and affected_pins is pretty much the same pin list the other code would create dynamically.  One difference:  the old code did not update the timing cost if net_is_ignored(net_id) -- you could try putting that back in.  Doesn't seem like it would make much difference but may be worth a try.

vpr/src/place/place.cpp
```diff
@@ -1861,31 +2221,21 @@ static void revert_td_cost(const t_pl_blocks_to_be_moved& blocks_affected) {
...
 
-    auto& cluster_ctx = g_vpr_ctx.clustering();
-    auto& clb_nlist = cluster_ctx.clb_nlist;
-
-    //Inalidate timing graph edges affected by the move
-    for (ClusterPinId pin : blocks_affected.affected_pins) {
-        //It is possible that some connections may not have changed delay.(e.g.
+    //Invalidate timing graph edges affected by the move
+    for (ClusterPinId clb_pin : sink_pins_affected) {
+        //It is possible that some connections may not have changed delay.
```
vaughnbetz, line 2232: VB: changed from old code, but I went through this and am pretty sure the net algorithm is the same so I don't think this is the slowdown.

vaughnbetz:
The code looks good Bill.  I have no major concerns; some detailed comments (mostly about small bits of additional commenting).  I only identified one small difference from the original CRITCALITY_PLACE algorithm and the new code, in commit_td_costs (ignored nets aren't ignored).  You could try inserting the appropriate if to see if it speeds things up; seems unlikely, but easy to try.
If you can quickly go through my comments I think we can get this merged.  After that, looking for the 7% slowdown would probably be best done with a binary search of commits; I don't see any obvious cause of that slowdown either.

vpr/src/place/timing_place.h
```diff
@@ -1,3 +1,32 @@
...
+ *              Also maps atom pin crit. to clb pin crit.
+ *   @class PlacerTimingCosts
+ *              Hierarchical structure used by update_td_costs() to
+ *              maintain the order of addition operation of float values
+ *              (to avoid round-offs) while doing incremental updates.
+ *
+ * Calculating criticalities:
+ *      All the raw setup slack values across a single clock domain are gathered, shifted,
+ *      and rated from best to worst. The best shifted slack value (the most positive one)
+ *      will have a criticality of 0, while the worse shifted slack value (always 0)
```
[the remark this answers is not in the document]
Bill-hbrhbr: Sorry. I re-read ***calc_relaxed_criticality()***, and you are right. Only when the worst slack is negative does the shift happen. I'll update the documentation.

vpr/src/place/timing_place.h
```diff
@@ -55,40 +97,148 @@ class PlacerCriticalities {
...
-    //timing_info and the provided criticality_exponent.
+    /**
+     * @brief Updates criticalities based on the atom netlist criticalitites
+     *        provided by timing_info and the provided criticality_exponent.
+     */
     void update_criticalities(const SetupTimingInfo* timing_info, float criticality_exponent);
 
-    //Override the criticality of a particular connection
+    ///@brief Override the criticality of a particular connection.
     void set_criticality(ClusterNetId net, int ipin, float val);
```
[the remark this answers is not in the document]
Bill-hbrhbr: Added assertion level 3 check with error message for ipin.

vpr/src/place/timing_place.cpp
Bill-hbrhbr: I think it is safe to remove these flags. We can use incremental updates whenever it's possible now.

vpr/src/place/timing_place.h
```diff
@@ -55,40 +97,148 @@ class PlacerCriticalities {
...
+
+    ///@brief Set of pins with raw setup slacks modified by last call to update_setup_slacks()
+    vtr::vec_id_set<ClusterPinId> cluster_pins_with_modified_setup_slack_;
+
+    ///@brief Updates setup slacks: incremental V.S. from scratch.
+    void incr_update_setup_slacks(const SetupTimingInfo* timing_info);
+    void recompute_setup_slacks();
+
+    ///@brief Flag that turns on/off the update_setup_slacks() routine.
+    bool update_enabled = true;
```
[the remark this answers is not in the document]
Bill-hbrhbr: I turned it off during my runtime experiments and it isn't the cause of the runtime increase.

vpr/src/place/place.cpp
Bill-hbrhbr: I will do this in the subsequent PR

vpr/src/place/timing_place.cpp
Bill-hbrhbr: I don't recommend doing this since they are pre-computed (in ```slack_evaluation.cpp```) and are not necessarily the same. The efforts that go into merging these two sets are not worth it in my opinion, and it might make the code less readable. I agree with your feeling that a lot of codes are duplicated, but I don't have a good way of resolving the issue without sacrificing code readability.

vpr/src/place/place.cpp
```diff
@@ -332,40 +334,44 @@ static void update_move_nets(int num_nets_affected);
...
+                              SetupTimingInfo* timing_info,
                               ClusteredPinTimingInvalidator* pin_timing_invalidator,
                               t_pl_blocks_to_be_moved& blocks_affected,
                               const PlaceDelayModel* delay_model,
-                              const PlacerCriticalities* criticalities,
+                              PlacerCriticalities* criticalities,
+                              PlacerSetupSlacks* setup_slacks,
                               float rlim_escape_fraction,
-                              enum e_place_algorithm place_algorithm,
+                              const t_place_algorithm& place_algorithm,
```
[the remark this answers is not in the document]
Bill-hbrhbr: The slowdown already occurred in the commit before I implemented this algorithm wrapper, so this is likely not the cause of the slowdown.

vpr/src/place/place.cpp
Bill-hbrhbr: There's no difference due to the way ```affected_pins``` are gathered. 
This structure is populated by calling ```try_swap()```->```find_affected_nets_and_update_costs()```->```update_td_delta_costs()```.
The ignored nets are skipped in ```find_affected_nets_and_update_costs()```.

Bill-hbrhbr:
Hi @vaughnbetz, I fixed all the issues raised by your comments in the most recent commit. Some comments are unresolved due to more explanation on my side. Also, I added function level header for ```find_affected_sink_pins()```, ```commit_td_cost()```, and ```invalidate_affected_connection_delays()```. Please go through the comments/documentation and see if you have more concerns.

Note: the comment I deleted was not essential.

Bill-hbrhbr:
Hi @vaughnbetz , I finally found the culprit. The runtime increase is due to splitting out the routine ```find_affected_sink_pins()``` from the original ```invalidate_affected_connection_delays()```. However, the vector populated by ```find_affected_sink_pins()``` is useful if we wish to revert a timing analysis step. I'll think of a way to take care of this issue.

vaughnbetz:
Excellent, thanks Bill.  The changes look good.  There is one build failure in Travis CI: there are a few warnings in timing_place.cpp.  Can you get rid of those?
The nightly test failure is a QoR failure on one circuit (unrelated); Sarah already has the action to update the golden result for it, so it is not gating.
For the slowdown, do you prefer to merge this PR and then refactor to speed up, or wait for the speed up?

Bill-hbrhbr:
Hi @vaughnbetz, I can resolve the slow down in this PR. I think the runtime increase is due to the ```vector::push_back()``` call in **find_affected_sink_pins()** requiring a lot of dynamic reallocations. Currently trying out reserving/pre-allocating enough size for the vector and see if that alone makes the runtime increase go away. I don't really want to remove **find_affected_sink_pins()** yet, since it is useful for reverting timing analysis if we are using slack analysis in the placement inner loop.
If the ```vector::reserve()``` does not work, I'll find a way to skip **find_affected_sink_pins()** if slack timing analysis is not used. But either way, the code change is not going to be drastic, so I'll do it in this PR. I'm currently running Titan benchmarks on the new code, and this PR will get finished by tomorrow for sure.

vaughnbetz:
Great, thanks Bill!

vaughnbetz:
Thanks Bill.  Can you launch a QoR run to make sure all is well, and the placer sped back up?
Also, please check that blocks_affected.affected_pins does not affect the wirelength calculation (update_bb) as if it did we could get a bad wirelength cost estimate by skipping pins with unchanged delay.

vaughnbetz:
Just checked the update_bb routines and they don't use .affected_pins so that is fine.

Bill-hbrhbr:
Titan benchmark #run001 @commit 06c9cf6: https://drive.google.com/file/d/1kduvV2um0rjZRi7JVNZGHwCBIMwlATe5/view?usp=sharing

There is some fluctuation in place time so I'm running a second round of Titan. But it's pretty obvious that the source of the slowdown is gone.

Bill-hbrhbr:
Titan benchmark #run002 @commit 06c9cf6: https://drive.google.com/file/d/1nTu_tpMZdKigYwwJxhS5Go0eUXeudbnx/view?usp=sharing

There is still a tiny degradation. So I decided to look into the VPR output files for **bitcoin_miner** benchmark. Specifically, I'm looking for the inner placement loop log.

**Master:** [master.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5191577/master.txt)
**New:** [PR#1501.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5191579/PR.1501.txt)

I see that there are multiple random spikes in placement iteration runtime stats in both files. Also, I (greedily) ran with the option ```-j8``` (so there would be 16 concurrent threads). Hence, I think the runtime fluctuation is mainly due to the load on Wintermute, rather than the code.

To be more clear, before removing ```find_affected_sink_pins()```, I saw a consistent overhead in each placement iteration in the log file, but right now that is not the case anymore.

And even if the code is indeed a bit slower, here are my 4 arguments:
1. It's a _negligible and inconsistent_ overhead.
2. It might be due to the way the code is compiled with the added branching in the method ```update_td_delta_cost()```. The benefits brought by further micro-optimizations is not worth the time and efforts put into it.
3. ```blocks_affected.affected_pins``` should have fewer elements than before, which can potentially speed up other parts of the code, such as ```invalidate_affected_connections()```. This is what I hoped for, but it didn't happen.
4. **Most importantly,** the current code (with modified ```blocks_affected.affected_pins```) is the most elegant way I can think of to incorporate the slack timing analysis option and one-step timing analysis reversion.

vaughnbetz:
Thanks for tracking this down Bill. I agree this is close enough (1.5% and 2% placer runtime impact in the two runs, which is getting pretty close to noise).  The code is more capable and cleaner, so good to have.

vaughnbetz:
(FYI, the one regtest failure is a known qor compare update issue; Sarah has updated it on the master already).

Issue #1503, opened 2020-08-25
Analytic Placer in VPR


zhaiyif1:
#### Proposed Behaviour
Add analytic placer in VPR, enabled by option "--enable_analytic_placer true" on VPR command line
If enabled, AP should take random placement by initial_placement and place
Should return legal placement to annealer for further optimization
Can skip annealing and go straight to quench

#### Current Behaviour
Current placer (annealer) takes random placement by initial_placement and optimizes by simulated annealing followed by quench

#### Possible Solution
ANALYTICAL PLACEMENT FOR HETEROGENEOUS FPGAS by Marcel Gort and Jason H. Anderson
nextpnr project, placer_heap

#### Context
To research and experiment with analytic placer, for improvement of performance and qor of vpr.
AP can be an optimization on initial placement before annealer kicks in.
Or it could be a stand-alone algorithm if performance is satisfactory.
Or a combination of the two (somehow) by tuning parameters and adjusting the flow of placement algorithm.

PR #1504, opened 2020-08-25
Add Analytic Placer to VPR with new option, test, and dependency

zhaiyif1:
Analytic placer algorithm itself is in analytic_placer.cpp, and cut_spreader.cpp
See analytic_placer.h, cut_spreader.h for algorithm summary

An off-the-shelf linear algebra package, Eigen is added to solve matrix equations
A new CMake option, **VPR_ANALYTIC_PLACE** is added.
CMakeLists.txt are modified to check eigen dependecy, if VPR_ANALYTIC_PLACE is set to ON (current default is ON, can change in $VTR_ROOT/CMakeList.txt). If Eigen3 is not found, analytic placement will be disabled, until user manually install eigen3 and rebuild.

A new VPR option **--enable_analytic_placer** is added, false by default
When disabled, the original placement algorithm is run
When enabled, AP takes over after initial_placement, and returns a legal placement to annealer. Annealer is disabled by setting first_t = 0, and goes straight to quench for further optimization

VTR_LOG now captures stats at each iteration of AP.

A new strong test is added to test this option and flow. 

VTR benchmark results for various AP configurations are attached here:
[AP benchmarks (1).xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5185472/AP.benchmarks.1.xlsx)
Notable: for latest revision of AP, place_time for (AP stall=15, quench*25, quench timing recompute * 5) improved by **over 25%** compared to old result. Every AP configuration may need to be rerun to update all the runtime numbers.

Usage example:
(ensure CMake option VPR_ANALYTIC_PLACE is ON when building)
$VTR_ROOT/vpr/vpr  <architecture>.xml    <netlist>.blif   --enable_analytic_placer true

#### Related Issue
#1503#

#### Motivation and Context
To experiment and research Analytic Placer. See related issue for more details.

vtr_reg_basic, vtr_reg_strong, vtr_reg_titan, with --enable_analytic_placer true
runs are successful but fails golden test as it produces different results, which is expected
with --enable_analytic_placer false it passes all test.

#### Types of changes

#### Checklist:


vaughnbetz:
Dusty, it would be great if you can review this code.  We should review and improve the code no matter what.  After that we have a few options:
- Keep the code on a branch for now if the eigen dependency or any code coupling in it worries us
- Merge it in and compile it and include the eigen dependency
- Merge it but control it with an ifdef so the dependency doesn't impact anyone who doesn't need to work on this

vaughnbetz:
Daniel can you also add a summary of the results from the placer to date (circuits run through, QoR) with the settings you are proposing to put back?  Linking your result table here would also be good.

vpr/src/place/cut_spreader.cpp
```diff
@@ -0,0 +1,816 @@
+#include "cut_spreader.h"
```
vaughnbetz, line 1: Needs a top-level comment that explains what this file does, with an overview of the algorithm, references to HEAP / SimPL etc.
Subroutines also need comment as to what each does.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,73 @@
+#ifndef VPR_SRC_PLACE_LEGALIZER_H_
```
vaughnbetz, line 1: File needs more commenting.  Each data member should be commented to indicate what it is for.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,73 @@
...
+  private:
+    int32_t dont_solve = std::numeric_limits<int32_t>::max();
+
+    AnalyticPlacer* ap;
+    const DeviceContext& device_ctx;
+    const ClusteredNetlist& clb_nlist;
+    PlacementContext& place_ctx;
+    t_logical_block_type_ptr blk_type;
+
+    struct MacroExtent {
```
vaughnbetz, line 28: Commenting data members is particularly important.

vpr/src/place/cut_spreader.cpp
```diff
@@ -0,0 +1,816 @@
...
+            if (result != sub_tile.equivalent_sites.end()) {
+                for (auto loc : ap->legal_pos.at(tile->index).at(sub_tile.index)) {
+                    ft.at(loc.x).at(loc.y).emplace_back(loc);
+                }
+            }
+        }
+    }
+}
+
+void CutSpreader::run() {
```
vaughnbetz, line 33: This needs a big comment.  What is the expected input? (Placement state, but where is it stored and what state is it expected to be in? ).  Where is the output stored and is it fully legal or just more spread out?  What is the algorithm?  What is the work queue?

vaughnbetz: Also looks like regions had to previously be set up.  Who set them up and how big are they?
Some of these comments may be file scope (how does the whole cut spreader work)?

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,160 @@
...
+
+    struct BlockLocation {
+        t_pl_loc loc;
+        t_pl_loc legal_loc;
+        int rawz;
+        double rawx, rawy;
+        bool locked;
+    };
+
+    std::unordered_map<ClusterBlockId, BlockLocation> blk_locs;
```
vaughnbetz, line 70: Need to comments all these data members and their use.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,160 @@
...
+        int criticalityExponent;
+        int timingWeight;
+        float solverTolerance;
+        bool placeAllAtOnce;
+        int spread_scale_x, spread_scale_y;
+    };
+
+    PlacerHeapCfg tmpCfg;
+
+    int max_x = 0, max_y = 0; //size of the grid
```
vaughnbetz, line 35: Are max_x and max_y the same as the grid.width and height variables?  If so, why make a copy?  I think it would be more clear to just work with the originals.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,160 @@
...
+    VprContext& vpr_ctx;
+    const DeviceContext& device_ctx;
+    const ClusteredNetlist& clb_nlist;
+    PlacementContext& place_ctx;
```
vaughnbetz, line 39: These variables can always be accessed with function calls so not sure why you're caching them.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,160 @@
...
+        ClusterBlockId blkId;
+        int32_t udata = std::numeric_limits<int32_t>::max(); //dont_solve
+        t_logical_block_type_ptr type = nullptr;
+
+        // info for macros
+        MacroInfo* macroInfo = nullptr; // nullptr if not in macro
+    };
+    std::unordered_map<ClusterBlockId, BlockInfo> blk_info;
+
+    struct BlockLocation {
```
vaughnbetz, line 62: Need detailed comments for all these data members.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,160 @@
...
+    std::vector<std::vector<std::vector<std::vector<std::vector<t_pl_loc>>>>> fast_tiles;
+
+    // marker for blks not solved in current equation
+    int32_t dont_solve = std::numeric_limits<int32_t>::max();
+    // marker for blks not in any macros
+    size_t not_in_macros = std::numeric_limits<size_t>::max();
+
+    struct MacroInfo;
+
+    struct BlockInfo {
```
vaughnbetz, line 52: Need comments for all these data members.  Some seem to be copies of what is in the clb netlist; better not to copy those I think (unless you are getting ready to handle netlists other than the clb netlist -- i.e. unclustered / atom netlist).  If you're decoupling from a specific netlist for that reason it would make sense, but you would need to comment that so your reasoning is clear.

vpr/src/base/read_options.cpp
```diff
@@ -1677,6 +1677,12 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("0")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    place_grp.add_argument(args.enable_analytic_placer, "--enable_analytic_placer")
+        .help(
+            "Enables analytic placer, and skips simulated annealing. The result in improved by quenching")
```
vaughnbetz, line 1682: "result is improved by quenching"

vaughnbetz: Probably should rewrite this to say "Enables the analytic placer. Once analytic placement is done, the result passed through the quench phase of the annealing placer for local improvement".

vpr/src/place/place.cpp
```diff
@@ -732,7 +737,9 @@ void try_place(const t_placer_opts& placer_opts,
 
     first_rlim = (float)max(device_ctx.grid.width() - 1, device_ctx.grid.height() - 1);
 
-    float first_t = starting_t(&costs, &prev_inverse_costs,
+    float first_t = 0;
```
vaughnbetz, line 740: This needs a big comment on how the analytic placer and annealer interact.  Analytic placer: disables most of the anneal (leaves the quench) and passes its placement back to the quench via locations marked on both the clb_netlist and the grid.

vaughnbetz:
More commenting is needed (particularly on the big picture and data members). Eigen is very big; can we (and should we) bring it in as a package instead of a lot of source that we compile?  It's a lot of code to check into the external libraries.

HackerFoo (changes requested):
Implementation needs a lot more inline comments, so I can follow along and generally understand algorithms without reading the relevant papers. Good code is usually easier to read than papers describing the algorithms used.

There's a lot of bounding box math that would benefit from `vtr::Rect`. Please extend this if it doesn't do what you need.

Minor, but I prefer braces for `if` and avoiding `continue` and `break`, especially when they don't save any lines. Otherwise, unexpected things can happen when the code is later modified. Also, `goto` is great for handling cleanup, but expect it to get attention because of how complicated gotos can make control flow. So when using `goto`, it should be *more* clear than other forms of control flow.

vpr/src/place/cut_spreader.cpp
```diff
@@ -0,0 +1,857 @@
...
+                    std::queue<std::pair<ClusterBlockId, t_pl_loc>> visit;
+                    visit.emplace(blk, sub_t);
+                    while (!visit.empty()) {
+                        ClusterBlockId vb = visit.front().first;
+                        VTR_ASSERT(!isPlaced(vb));
+                        t_pl_loc target = visit.front().second;
+                        visit.pop();
+                        auto log_type = clb_nlist.block_type(blk);
+                        auto result = std::find(log_type->equivalent_tiles.begin(), log_type->equivalent_tiles.end(), g_vpr_ctx.device().grid[target.x][target.y].type);
+                        if (result == log_type->equivalent_tiles.end()) goto fail;
```
vaughnbetz, line 811: I don't like the goto here either.  Should make a cleanup routine and call it from two places and break if necessary.  Goto is almost never seen in C++ code nowadays.

I also prefer putting what an if controls in braces, or, if only one line you can do on the indented line under.  Putting on the same line is harder to read.  So:
if (result == log_type->equivalent_tiles.end()) {
    spreader_unlock ();   // Or some such
    continue;   // If you really need a continue
}



   
}

vpr/src/place/cut_spreader.cpp
```diff
@@ -0,0 +1,857 @@
...
+                        if (swap.second != EMPTY_BLOCK_ID)
+                            remaining.emplace(imacro(swap.second) != -1 ? place_ctx.pl_macros[imacro(swap.second)].members.size() : 1, swap.second);
+                    }
+
+                    placed = true;
+                    break;
+                }
+            }
+        }
+    }
```
vaughnbetz, line 854: This routine is about 240 lines long, without inline comments on what it's doing.
1. It should be factored into multiple smaller routines so the overall structure is more clear and you get out of these long loop nests, and avoid things like goto's.
2. It need inline comments (will be helped by calling sub-methods or sub-routines with meaningful names).

zhaiyif1:
Thanks @vaughnbetz and @HackerFoo, I'm on it right now.

Also, the vtr nightly tests failed at this test:
regression_tests/vtr_reg_nightly/vtr_reg_qor_chain_depop...[Fail] 
 k6_frac_N10_frac_chain_depop50_mem32K_40nm.xml/spree.v/common min_chan_width: golden = 76 result = 100

I ran the test again on master branch locally, and got the same results. Seems like golden-result needs to be updated on master.

vaughnbetz:
Thanks @zhaiyif1 .  Go ahead and update the golden qor (if you're familiar with the procedure).  Otherwise note that here and hopefully @MohamedElgammal can update it.

vpr/src/place/cut_spreader.cpp
```diff
@@ -0,0 +1,857 @@
...
+                    }
+
+                    for (auto& target : targets) {
+                        ClusterBlockId bound = place_ctx.grid_blocks[target.second.x][target.second.y].blocks[target.second.sub_tile];
+                        if (bound != EMPTY_BLOCK_ID) unbindTile(target.second);
+                        bindTile(target.second, target.first);
+                        swaps_made.emplace_back(target.second, bound);
+                    }
+
+                    if (false) {
```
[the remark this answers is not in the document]
HackerFoo: My main use of goto is as a local exception. I'm not sure if a throw would get optimized to a jump. Maybe.

The other use is where I have something like:
```
if(...) {
  if(...) {
    goto doA;
    // otherwise this would be a copy of A
  } else {
    B
  }
} else {
doA:
   A
}
```

because I hate repeating `A`. Obviously this can be rewritten, but the rewritten condition gets harder to understand as the nesting gets more complex. The other option is to put A into a function and call that, but that can be annoying if A modifies a bunch of variables.

I've never seen jumping into `if (false) { ... }`, though. I was kind of surprised that would work.

vaughnbetz:
Sarah also saw the qor failure; she'll update it.

zhaiyif1:
Awesome! Good to know.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,705 @@
...
+        ClusterBlockId head_id = macro.members[0].blk_index;
+        for (auto member = ++macro.members.begin(); member != macro.members.end(); ++member) {
+            blk_locs[member->blk_index].loc = blk_locs[head_id].loc + member->offset;
+        }
+    }
+}
+
+// Build and solve in one direction
+void AnalyticPlacer::build_solve_direction(bool yaxis, int iter, int build_solve_iter) {
+    for (int i = 0; i < build_solve_iter; i++) {
```
vaughnbetz, line 460: Why do you loop through multiple iterations of build_solve_iter?  How does setting up and solving the same matrix equation multiple times change things (or is build equations doing something different on each call)?
What does iter do?
Function level comment on all input variables: (yaxis true if y-directed, false for x-direction solve.  iter is ??, build_solve_iter controls how many times you build and solve this matrix equation; values > 1 are useful because ??

vaughnbetz (changes requested):

vaughnbetz:
Partway through my review, but posting now so you get some of the comments earlier.
Haven't gone through analytic_placer.h yet, where I see there are some big comments, so that may give more overview.  
In the .cpp file, more commenting on functions and inline on what is happening is needed (see detailed feedback).
There is also some code duplication that should be factored out (get_bb_from_scratch, part of initial placement that could be a helper function called from two places).

zhaiyif1:
Thank you. I'm also adding comments and refactoring the code based on
Dusty's comments.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,705 @@
...
+    PlacementContext& place_ctx = g_vpr_ctx.mutable_placement();
+
+    // Return the x or y position of a block
+    auto blk_p = [&](ClusterBlockId blk_id) { return yaxis ? blk_locs.at(blk_id).loc.y : blk_locs.at(blk_id).loc.x; };
+    // Return legal position from legalization, after first iteration
+    auto legal_p = [&](ClusterBlockId blk_id) { return yaxis ? blk_locs.at(blk_id).legal_loc.y : blk_locs.at(blk_id).legal_loc.x; };
+    es.reset();
+
+    for (auto net_id : clb_nlist.nets()) {
+        if (clb_nlist.net_is_ignored(net_id)) continue;
```
vaughnbetz, line 479: I don't like putting the logic on the same line as the if; please put on the line underneath, indented.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,705 @@
...
+        if (yaxis) {
+            blk_locs.at(solve_blks.at(i)).rawy = vals.at(i);
+            blk_locs.at(solve_blks.at(i)).loc.y = std::min(max_y - 1, std::max(0, int(vals.at(i) + 0.5)));
+        } else {
+            blk_locs.at(solve_blks.at(i)).rawx = vals.at(i);
+            blk_locs.at(solve_blks.at(i)).loc.x = std::min(max_x - 1, std::max(0, int(vals.at(i) + 0.5)));
+        }
+}
+
+// Debug use, finds # of blocks on 1 tile for all tiles
```
vaughnbetz, line 578: This routine seems to compute an unusual definition of overlap, as it counts all blocks (no matter whether they are overlapping or not) and then divides the number of blocks by that overlap (which I think is just the same as the number of blocks).  Check if this does something useful; if not I'd correct or delete it.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,705 @@
...
+    }
+
+    for (auto blk : clb_nlist.blocks()) {
+        overlap.at(blk_locs[blk].loc.y).at(blk_locs[blk].loc.x) += 1;
+    }
+
+    std::cout << "overlap:" << std::endl;
+    for (auto& row : overlap) {
+        for (int count : row) {
+            if (count > 0) OL++;
```
vaughnbetz, line 598: This seems like a bug; incrementing overlap if count > 0 means you count every block, no matter whether it is overlapping something else or not.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,705 @@
...
+}
+
+// Debug use, finds # of blocks on 1 tile for all tiles
+float AnalyticPlacer::find_overlap() {
+    const ClusteredNetlist& clb_nlist = g_vpr_ctx.clustering().clb_nlist;
+    int max_x = g_vpr_ctx.device().grid.width();
+    int max_y = g_vpr_ctx.device().grid.height();
+
+    int OL = 0;
+    overlap.resize(max_y);
```
vaughnbetz, line 585: overlap should be a variable allocated and returned by this routine, rather than a member variable of the class.  Having a big class like this and making more instance variables in it than necessary is akin to using lots of globals -- you have a bunch of variables accessible (read/write) throughout the class, which means it's hard to trace what variables are changed where or reason about the code.  Use local or passed in variables when you can, and promote to a member variable only for variables that really are needed througout the class. overlap is only computed by this routine and printed by another; that's a class case of a utility function that should just process more basic class state, but shouldn't create any member variables of its own.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,705 @@
...
+    for (auto& row : overlap) {
+        row.resize(max_x);
+        std::fill(row.begin(), row.end(), 0);
+    }
+
+    for (auto blk : clb_nlist.blocks()) {
```
vaughnbetz, line 591: the vtr::Matrix class would be a good fit to this variable (cleaner).  See https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/libvtrutil/src/vtr_ndmatrix.h if you want to see the code, but probably easiest to look for an example of its use (which is easy; behaves like:
my_matrix[i][j] = 20; // Access as if it's a 2D array.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,705 @@
...
+            if (count > 0) OL++;
+        }
+    }
+
+    return ((float)clb_nlist.blocks().size()) / (float)OL;
+}
+
+// prints a simple figure of FPGA fabric, with numbers on each tile showing usage
+// called in AnalyticPlacer::print_place()
+std::string AnalyticPlacer::print_overlap() {
```
vaughnbetz, line 607: Pass in overlap, and move it out of the class member variables.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+ * incompatible physical tiles. To legalize this solution, a geometric partitioning and spreading technique, introduced
+ * in SimPL, is used (@see cut_sreader.h). This completes the upper-bound placement.
+ *
+ * After the completion of 1 iteration of lower-bound & upper-bound placement, artificial pseudo connections are created
+ * between each block and its target location in the legalized overlap-free placement. When the mathematical system is
+ * again formulated and solved, the pseudo connections pull blocks towards their target locations, which tends to reduce
+ * overlaps in the placement.
+ *
+ * This process of formulating the system, solving, and legalizing is repeated until sufficiently good placement is
+ * acquired. Currently the stopping criterion is 15 iterations without improvement in total_hpwl.
```
vaughnbetz, line 49: Change 15 to your new defined constant name. (you can say 15 at time of this writing if you like too, but good to link to the constant name).

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+ * have been modified for the architecture and netlist specification of VTR, and better performance.
+ * nextpnr -- Next Generation Place and Route,  placer_heap, David Shah <david@symbioticeda.com>
+ * https://github.com/YosysHQ/nextpnr
+ */
+
+#    include "vpr_context.h"
+#    include "timing_place.h"
+#    include "PlacementDelayCalculator.h"
+
+typedef signed long int wirelen_t;
```
vaughnbetz, line 88: Comment what this is for (wirelength) and why you use it.  This is still only a 32 bit integer but that should be enough for integral wirelength of an FPGA.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+/*
+ * @brief Templated struct for constructing and solving matrix equations in analytic placer
+ * Eigen library is used in EquationSystem::solve()
+ */
+template<typename T>
+struct EquationSystem;
+
+/*
+ * @brief helper function to find the index of macro that contains blk
+ * returns index in placementCtx.pl_macros, -1 if blk not in any macros
```
vaughnbetz, line 99: Define NO_MACRO = -1 and use it instead of -1 as a sentinel.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+ */
+int imacro(ClusterBlockId blk);
+
+/*
+ * @brief helper fucntion to find the head of macro containing blk
```
vaughnbetz, line 104: Explain what the head block is, and why it is important (only one represented in the matrix equation). Typo: function.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+     * To tune these parameters, change directly in constructor
+     */
+    AnalyticPlacer();
+
+    /*
+     * @brief top function of analytic placement
+     * Takes the random initial placement from place.cpp through g_vpr_ctx
+     * Repeat the following until stopping criteria is met:
+     * 	* Formulate and solve equations in x, y directions for 1 type of logial block
+     * 	* Instantiate CutSpreader to spread and strict_legalize
```
vaughnbetz, line 122: ... and strict_legalize to fully legalize

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+        int criticalityExponent;            // not currently used, @see build_equations()
+        int timingWeight;                   // not currently used, @see build_equations()
+        float solverTolerance;              // parameter of the solver
+        int spread_scale_x, spread_scale_y; // see CutSpreader::expand_regions()
+    };
+
+    AnalyticPlacerCfg tmpCfg; // TODO: PlacerHeapCfg should be externally configured & supplied
+
+    // @see initial_placement.cpp load_legal_locations()
+    std::vector<std::vector<int>> num_legal_pos;
```
vaughnbetz, line 145: Should define what these are, not refer to another file where similar code exists.  This is a sign the code is duplicated and should be refactored.  Also, these shouldn't be member variables to a big class like this as it appears they are loaded in one place and used in another; they should be parameters / local variables that are passed around.
Finally, num_legal_pos seems unnecessary given that the last dimension of legal_pos is a vector anyway.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+
+    // marker for blks not solved in current iteration
+    int32_t dont_solve = std::numeric_limits<int32_t>::max();
+
+    // row number in the system of linear equations for each block
+    std::unordered_map<ClusterBlockId, int32_t> blk_row_num;
+
+    // Encapsulates 3 types of locations for each logic block
+    struct BlockLocation {
+        t_pl_loc loc;       // real, up-to-date location of the logic block
```
vaughnbetz, line 156: Explain (at a high level) how this is set.  Sometimes solver location, sometimes legalized?  Always last legalized?  Varies between initial placement, solved location, legalized?

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+        t_pl_loc legal_loc; // legalized location, used to create psudo connections
+        double rawx, rawy;  // raw location storing float result from matrix solver
+    };
+
+    // Lookup from blockID to block location
+    std::unordered_map<ClusterBlockId, BlockLocation> blk_locs;
```
vaughnbetz, line 162: vtr::vector_map would be faster.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+
+    // The set of blks actually placed.
+    // Excludes children of macros.
+    std::vector<ClusterBlockId> place_blks;
```
vaughnbetz, line 166: Explain what "children of macros" means (I think you mean non-head blocks of macros).  Does it also exclude blocks that are .fixed (locked down)?  I think you want to say "The ids of the blocks that are placed by the matrix solver, i.e. the free variable blocks. This excludes blocks that are part of placement macros but not the head of that macro, and fixed blocks."

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+
+    // blocks being solved in the current equation
+    // subset of all blocks being placed
+    std::vector<ClusterBlockId> solve_blks;
```
vaughnbetz, line 170: What's the difference between solve_blks and place_blks?  Explain what this subset of blocks is (give examples of how you're excluded).  If solve_blks and place_blks are the same, eliminate one variable.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+
+    /*
+     * number of blocks on each tile
+     * used to print a simple placement graph in print_place
+     */
+    std::vector<std::vector<int>> overlap;
```
vaughnbetz, line 176: Shouldn't be a class member variable; can reduce scope to a local variable that is returned by overlap() and printed by print_overlap(overlap&).

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+    void print_place(const char* place_file);
+
+    //build fast lookup of tiles/subtiles by tile, x, y, subtiles
+    void build_fast_tiles();
+
+    // build num_legal_pos, legal_pos like in initial_placement.cpp
+    void build_legal_locations();
+
+    // build blk_locs based on initial placement;
+    // build and place_blks for blocks that needs to be placed;
```
vaughnbetz, line 192: "build and place blks".  Not sure what that means.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+    void init();
+
+    wirelen_t get_net_hpwl(ClusterNetId net_id);
+
+    wirelen_t total_hpwl();
+
+    // Setup the blocks of type blkTypes to be solved
+    // Returns the number of rows (number of blks to solve)
```
vaughnbetz, line 200: Give an example; e.g. the type might be logic_blocks and this routine would ... and then return the number of rows (blocks to solve) which would be the number of logic blocks (or maybe it is the number of logic blocks that are free variables in the matrix?).

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+    int setup_solve_blks(t_logical_block_type_ptr blkTypes);
+
+    // Update the location of all members of all macros based on location of macro_head
+    // Since only macro_head is solved (connections to macro members are also taken into account
+    // when formulating the matrix equations)
+    void update_macros();
+
+    // Build and solve in one direction
+    // Solved solutions are written back to block_locs in rawx, rawy
+    void build_solve_direction(bool yaxis, int iter, int build_solve_iter);
```
vaughnbetz, line 210: Explain yaxis, ite,r build_solve_iter.

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+
+    // Build the system of equations for either X or Y
+    // When iter > 1, psudo-conenctions are formed, the strength is determined by alpha and iter
```
vaughnbetz, line 213: Explain pseudo connections (pseudo connections to the prior legal location for each movable block, with the  strength determined by alpha and iter).

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+    void build_equations(EquationSystem<double>& es, bool yaxis, int iter = -1);
+
+    // Solve the system of equations
+    void solve_equations(EquationSystem<double>& es, bool yaxis);
```
vaughnbetz, line 217: explain yaxis.  What does this update (must be a class member variable, as it appears to have no output from the prototype).

vaughnbetz (changes requested):
More comments on analytic_placer.h.
Main high-level one: be careful with class member variables in a large singleton class like analytic_placer.  They act as a form of global, and hence obscure who is reading and writing what in function calls.  I've given a couple of specific examples of where class members should be demoted to local variables and passed around; something to watch for in general.

vtr_flow/tasks/regression_tests/vtr_reg_strong/task_list.txt
```diff
@@ -70,3 +70,7 @@ regression_tests/vtr_reg_strong/strong_timing_update_type
 regression_tests/vtr_reg_strong/strong_timing_update_diff
 regression_tests/vtr_reg_strong/strong_blocks_with_no_inputs
 regression_tests/vtr_reg_strong/strong_fix_clusters
+regression_tests/vtr_reg_strong/strong_analytic_placer
+regression_tests/vtr_reg_strong/strong_timing_update_diff
```
vaughnbetz, line 74: It looks like some tests got duplicated (timing_update_diff, string_fix_clusters)?

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+ * are bordered on all sides by non-overutilized locations. Next, the area is expanded in both the x and y
+ * directions until it's large enough to accommodate all blocks contained. Overutilization is defined as follows:
+ * (Occupancy / Capacity) > beta, where beta is a constant <=1, currently defined in AnalyticPlacer::PlacerHeapCfg.
+ *
+ * cut_region
+ * ----------
+ * @see cut_region()
+ * @see run()
+ * In the second step, two cuts are generated: a source cut and a target cut. The source cut pertains to the blocks
+ * being places; the target cut pertains to the area into which the blocks are placed. The source cut splits the
```
vaughnbetz, line 41: typo: being placed

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+ * architecture. To achieve this, first assign all tiles to right region, then move tiles to the left one by one,
+ * calculating the difference in utilization for each move. The move with smallest delta_utilization is chosen as
+ * the target cut. To eliminate possible overutilization in either subarea, perturb the source cut by moving a
+ * single block from the over-utilized sub-area to the other, until neither is overutilized.
+ *
+ * Next, the blocks in sub-areas are spread to distribute them evenly. We split the sub-area into 10 equally-sized
+ * bins and logic blocks into 10 equal-capacity source bins. Then linearly interpolate to map blocks from their
+ * original_locations in their source bins to new spread location in target bins.
+ *
+ * This cut and spreading is repeated recursively. The cut spreading process returns the left and right subareas,
```
vaughnbetz, line 62: Do you alternate x and y dimensions, with one x-area (e.g. left) being cut into two y-areas, and then each of those two y-subareas being cut into two further x-subareas?

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * All blocks are sorted in descending macro lengths order and put in a priority queue (only macro heads are
+ * considered, while the rest of the macro members are ignored; single blocks have length 1). Each block goes through
+ * the following procedure:
+ *
+ * * Find all compatible sub_tile types, based on which all potential sub_tile locations are found (this process is
+ *   made computationally cheap by legal_pos data member in AnalyticPlacer)
+ * * Within a radius (starting from 0) of the block's currently location, randomly choose a location as candidate.
+ * * If the block is a single block (not in macro), multiple candidates are potentially chosen, and the one that
+ *   results in the smallest input wirelength for the block is chosen.
```
vaughnbetz, line 82: What is input wirelength?  wirelength to its input pins only?

vpr/src/place/analytic_placer.h
```diff
@@ -0,0 +1,249 @@
...
+
+    // @see initial_placement.cpp load_legal_locations()
+    std::vector<std::vector<int>> num_legal_pos;
+    std::vector<std::vector<std::vector<t_pl_loc>>> legal_pos;
+
+    // marker for blks not solved in current iteration
+    int32_t dont_solve = std::numeric_limits<int32_t>::max();
+
+    // row number in the system of linear equations for each block
+    std::unordered_map<ClusterBlockId, int32_t> blk_row_num;
```
vaughnbetz, line 152: vtr::vector_map would be faster.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+     * @return	result placement is passed to strict legalizer by modifying blk_locs in analytic_placer
+     */
+    void cutSpread();
+
+    /*
+     * @brief: 	Greedy strict legalize using algorithm described in algorithm overview above.
+     *
+     * 			Input illegal placement from data members (blk_locs) in analytic_placer
+     *
+     * @return	result placement is passed to annealer by modifying vpr_ctx.mutable_placement()
```
vaughnbetz, line 149: Don't you do a strict legalization in each solve/spread/legalize iteration?  (I thought that was the flow as analytic_placement.h comments).  If so, the result placement is presumably passed back through blk_locs, not vpr_ctx.mutable_placement().  Can you check this?
Explain which part of blk_locs is the input and which part is the output (there are multiple sub-members) for the cut spreader and the strict legalizer.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+     */
+    void strict_legalize();
+
+  private:
+    // marker for blocks that wasn't solved (blocks of different type, macro members)
+    int32_t dont_solve = std::numeric_limits<int32_t>::max();
```
vaughnbetz, line 155: I think it would be better to define this sentinel value once (say in analytic_placer.h) and then just use it here.

vpr/src/place/cut_spreader.h
vaughnbetz: Explain what x0,y0, x1,y1 are.  Is one the lower-left and the other the upper-right?  Or are they unordered where x0,y0 could be either corner?  Are they relative to the placement of the macro head or absolute values (i.e. if the macro head is at (1,6) and it's a two block macro with the second block offset by y_offset = -1, what are (x0,y0) and (x1,y1)?

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+
+    // struct that specifies the region occupied by a macro
+    struct MacroExtent {
+        int x0, y0, x1, y1;
+    };
+
+    // struct describing regions on FPGA to cut_spread
+    struct SpreaderRegion {
+        int id;              // index of regions in regions vector
+        int x0, y0, x1, y1;  // extent of the region
```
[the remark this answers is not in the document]
vaughnbetz: Good idea; vtr:rect would make this more clear, and has useful utilities.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+        int n_blks, n_tiles; // number of blocks and compatible tiles
```
vaughnbetz, line 172: I'd say "number of netlist blocks and compatible tiles (placement locations)" 
to differentiate blocks from tiles.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+        bool overused(float beta) const {
+            // determines whether region is overutilized:
+            // overused = (Occupancy / Capacity) > beta
+            if (n_blks > beta * n_tiles) return true;
```
vaughnbetz, line 176: Slightly cleaner code to write 
if ( ...) 
   return true
else 
   return false

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+            return false;
+        }
+    };
+
+    // Utilization of each tile, indexed by x, y
+    std::vector<std::vector<int>> occupancy;
+
+    // Region ID of each tile, indexed by x, y. -1 if not covered by any region
```
vaughnbetz, line 184: Best to define a AP_NO_REGION as -1 to avoid the magic number one in your code.

vpr/src/place/cut_spreader.h
vaughnbetz: could use vtr::rect for this.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+    std::vector<std::vector<int>> groups;
+
+    // Extent of macro at x, y location. If blk is not in any macros, MacroExtent only covers a single locations
+    std::vector<std::vector<MacroExtent>> macros;
```
vaughnbetz, line 188: macros isn't a great name for this, as this is in fact the region of a macro.  
`vtr::NdMatrix<vtr::Rect,2> `would handle this data structure (and be easier to read).  Not sure there's time, but it would be cleaner.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+
+    // List of logic blocks of blk_type at x, y location, indexed by x, y
```
vaughnbetz, line 190: Give an example of how to access (e.g. [x][y] -> gives vector of clustered block ids at that location.
Could use vtr::NdMatrix<ClusterBlockId,2>

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+    std::vector<std::vector<std::vector<ClusterBlockId>>> blks_at_location;
+
+    // List of sub_tiles compatible with blk_type at x, y location, indexed by x, y
```
vaughnbetz, line 193: ft is not a good name for a large scope variable like this.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+    std::vector<std::vector<std::vector<t_pl_loc>>> ft;
+
+    // List of all SpreaderRegion, index of vector members is the id of the region
+    std::vector<SpreaderRegion> regions;
+
+    // List of all merged_regions, these regions are merged in larger regions and should be skipped when
+    // recursively cur_spreading. Each entry is the region's ID, which is also the index into the regions vector
```
vaughnbetz, line 200: cur_spreading (typo).

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+    std::unordered_set<int> merged_regions;
+
+    // Lookup of macro's extent by block ID. If block is a single block, MacroExtent contains only 1 tile location
+    std::map<ClusterBlockId, MacroExtent> blk_extents;
```
vaughnbetz, line 204: vtr::vector_map<ClusterBlockId, MacroExtent> or 
vtr::vector_map<ClusterBlockID, vtr::Rect>  // even better
would be faster and smaller.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+    void init();
+
+    // Returns number of logical blocks at x, y location
+    int occ_at(int x, int y);
+
+    // Returns number of compatible sub_tiles at x, y location
+    int tiles_at(int x, int y);
+
+    /*
+     * Merge mergee into merged by:
```
vaughnbetz, line 217: Would be good to briefly explain or forward reference the expand_region member to explain why regions would merge.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+     * * adds all n_blks and n_tiles from mergee to merged region
+     * * grow merged to include all mergee grids
+     */
+    void merge_regions(SpreaderRegion& merged, SpreaderRegion& mergee);
+
+    /*
+     * Grow region r to include a rectangular region (x0, y0, x1, y1)
+     * Pass init = true if first time calling for a newly created region
+     */
+    void grow_region(SpreaderRegion& r, int x0, int y0, int x1, int y1, bool init = false);
```
vaughnbetz, line 228: x0,y0 etc. would be cleaner as a vtr::Rect

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+
+    /*
+     * Expand all utilized regions until they satisfy n_tiles * beta >= n_blocks
+     * If overutilized regions overlap in this process, they are merged
+     */
+    void expand_regions();
+
+    /*
+     * Find overutilized regions surrounded by non-overutilized regions
+     * Start off at an overutilized tile and expand in x, y directions until the region is surrounded by
```
vaughnbetz, line 238: Do you expand alternately in the x and y directions (or equally in both directions)?  Good to explain briefly how the x,y growth works.

vpr/src/place/cut_spreader.h
```diff
@@ -0,0 +1,259 @@
...
+
+    /*
+     * Recursive cut-based spreading in HeAP paper
+     * "left" denotes "-x, -y", "right" denotes "+x, +y" depending on dir
+     *
+     *  @param r	regrion to cut & spread
+     *  @param dir	direction, true for y, false for x
+     *
+     *  @return		a pair of sub-region IDs created from cutting region r.
+     *  			{-2, -2} if base case is reached
```
vaughnbetz, line 251: Make these values (-2,-2 and -1,-1) defined constants (pairs) so you can check for them with clean looking code. Or return a separate enumerated constant status code (e.g. with a by-reference outcome variable).

vaughnbetz (changes requested):

vaughnbetz:
I like the high level comments and data structure comments in cut_spreader.h, but have some requests for details in some places (see individual comments).  Time-permitting, various pieces of code could be made shorter and easier to read with use of vtr::NdMatrix and vtr::Rect.

vpr/src/place/cut_spreader.h
zhaiyif1: It's used in nextpnr, probably a heuristic. I did try adding up all wirelengths (input + output) and the result was slightly worse than input alone.

vaughnbetz: Thanks.  Worth adding that to the comment briefly here (what is input wirelength), and adding why either here or in the strict legalizer's comments.

zhaiyif1:
sorry Dusty and Vaughn, my last push had some changes from master and they are showing up in my changed files. I've "minimized" those files by selecting "reviewed" so they don't show up. In other files, I've added review comments for those changes that are not mine.

Other than the analytic_placer and cut_spreader files, there are 2 notable changes. One is I factored out the functions for building legal_pos from initial_placement to vpr_utils. The other is I added a contains(vtr::Rect<T> rect) function to vtr::Rect, which is handy for expanding regions in cut_spreading. 

I've basically addressed every comments you guys left, except the one regarding pin_offset for larger blocks by Vaughn. I've left comments where it needs to be worked on. The refactoring did improve the runtime a lot, mostly from an inefficient routine in strict_legalize(). On the circuit I ran locally it halfed AP runtime. I'm running VTR benchmark on wintermute to confirm this.

Otherwise everything should be done. Let me know if there are changes you still want me to make.

ODIN_II/SRC/netlist_create_from_ast.cpp
```diff
@@ -203,8 +203,9 @@ void look_for_clocks(netlist_t* netlist) {
     int i;
 
     for (i = 0; i < netlist->num_ff_nodes; i++) {
-        if (netlist->ff_nodes[i]->input_pins[1]->net->driver_pin->node->type != CLOCK_NODE) {
-            netlist->ff_nodes[i]->input_pins[1]->net->driver_pin->node->type = CLOCK_NODE;
+        oassert(netlist->ff_nodes[i]->input_pins[1]->net->num_driver_pins == 1);
```
zhaiyif1, line 206: Please ignore this file.

ODIN_II/SRC/netlist_statistic.cpp
```diff
@@ -232,10 +232,11 @@ static metric_t* get_upward_stat(nnet_t* net, netlist_t* netlist, uintptr_t trav
         if (traverse(net, traverse_mark_number)) {
             init(destination);
 
-            if (net->driver_pin) {
-                metric_t** parent_stat = (metric_t**)vtr::calloc(1, sizeof(metric_t*));
-                parent_stat[0] = get_upward_stat(net->driver_pin->node, netlist, traverse_mark_number);
-                aggregate(destination, parent_stat, 1);
+            if (net->num_driver_pins) {
```
zhaiyif1, line 235: Please ignore this file.

ODIN_II/SRC/netlist_utils.cpp
```diff
@@ -277,7 +277,8 @@ nnet_t* allocate_nnet() {
     nnet_t* new_net = (nnet_t*)my_malloc_struct(sizeof(nnet_t));
 
     new_net->name = NULL;
-    new_net->driver_pin = NULL;
+    new_net->driver_pins = NULL;
```
zhaiyif1, line 280: Please ignore this file.

ODIN_II/SRC/netlist_visualizer.cpp
```diff
@@ -314,28 +314,30 @@ void backward_traversal_net_graph_display(FILE* fp, uintptr_t marker_value, nnod
             if (current_node->input_pins[j] == NULL)
                 continue;
 
-            if ((current_node->input_pins[j] == NULL) || (current_node->input_pins[j]->net == NULL) || (current_node->input_pins[j]->net->driver_pin == NULL))
+            if ((current_node->input_pins[j] == NULL) || (current_node->input_pins[j]->net == NULL) || (current_node->input_pins[j]->net->num_driver_pins == 0))
```
zhaiyif1, line 317: Please ignore this file.

ODIN_II/SRC/node_creation_library.cpp
```diff
@@ -286,12 +286,18 @@ char* hard_node_name(nnode_t* /*node*/, char* instance_name_prefix, char* hb_nam
  * 	This creates the unique node name
  *-------------------------------------------------------------------------------------------*/
 char* node_name(nnode_t* node, char* instance_name_prefix) {
+    return op_node_name(node->type, instance_name_prefix);
```
zhaiyif1, line 289: Please ignore this file.

ODIN_II/SRC/odin_ii.cpp
```diff
@@ -113,7 +113,7 @@ static ODIN_ERROR_CODE synthesize_verilog() {
     parse_to_ast();
     /**
      *  Note that the entry point for ast optimzations is done per module with the
-     * function void next_parsed_verilog_file(ast_node_t *file_items_list) 
+     * function void next_parsed_verilog_file(ast_node_t *file_items_list)
```
zhaiyif1, line 116: Please ignore this file.

ODIN_II/SRC/output_blif.cpp
```diff
@@ -52,31 +52,75 @@ void define_ff(nnode_t* node, FILE* out);
 void define_decoded_mux(nnode_t* node, FILE* out);
 void output_blif_pin_connect(nnode_t* node, FILE* out);
 
-static void print_input_pin(FILE* out, nnode_t* node, long pin_idx) {
-    oassert(pin_idx < node->num_input_pins);
-    nnet_t* net = node->input_pins[pin_idx]->net;
-    if (!net->driver_pin || !net->driver_pin->node) {
+static bool warn_undriven(nnode_t* node, nnet_t* net) {
```
zhaiyif1, line 55: Please ignore this file.

ODIN_II/SRC/partial_map.cpp
```diff
@@ -48,7 +48,7 @@ void depth_first_traverse_partial_map(nnode_t* node, uintptr_t traverse_mark_num
 void partial_map_node(nnode_t* node, short traverse_number, netlist_t* netlist);
 
 void instantiate_not_logic(nnode_t* node, short mark, netlist_t* netlist);
-void instantiate_buffer(nnode_t* node, short mark, netlist_t* netlist);
+bool eliminate_buffer(nnode_t* node, short, netlist_t*);
```
zhaiyif1, line 51: Please ignore this file.

ODIN_II/SRC/read_blif.cpp
```diff
@@ -685,12 +685,13 @@ void create_internal_node_and_driver(FILE* file, Hashtable* output_nets_hash) {
 
         add_output_pin_to_node(new_node, new_pin, 0);
 
-        nnet_t* new_net = allocate_nnet();
-        new_net->name = new_node->name;
-
-        add_driver_pin_to_net(new_net, new_pin);
-
-        output_nets_hash->add(new_node->name, new_net);
+        nnet_t* out_net = (nnet_t*)output_nets_hash->get(names[input_count - 1]);
```
zhaiyif1, line 688: Please ignore this file.

ODIN_II/SRC/simulate_blif.cpp
```diff
@@ -131,6 +131,8 @@ static int get_num_covered_nodes(stages_t* s);
 static int is_node_ready(nnode_t* node, int cycle);
 static int is_node_complete(nnode_t* node, int cycle);
 
+static bool pin_is_driver(npin_t* pin, nnet_t* net);
```
zhaiyif1, line 134: Please ignore this file.

vpr/test/test_vpr.cpp
```diff
@@ -125,8 +125,7 @@ TEST_CASE("read_rr_graph_metadata", "[vpr]") {
             kArchFile,
             "wire.eblif",
             "--route_chan_width",
-            "100",
-        };
+            "100"};
```
zhaiyif1, line 128: Please ignore this file.

vpr/src/route/rr_graph_util.cpp
```diff
@@ -169,4 +169,11 @@ void reorder_rr_graph_nodes(const t_router_opts& router_opts) {
             }
         }
     }
+
```
zhaiyif1, line 172: Please ignore this file.

vpr/src/route/routing_predictor.h
```diff
@@ -25,14 +25,15 @@ class RoutingPredictor {
 
     void add_iteration_overuse(size_t iteration, size_t overused_rr_node_count);
 
-    float get_slope();
+    float get_slope() const;
```
zhaiyif1, line 28: Please ignore this file.

vpr/src/route/routing_predictor.cpp
```diff
@@ -63,12 +63,9 @@ float covariance(std::vector<size_t> x_values, std::vector<float> y_values, floa
     return cov;
 }
 
-float RoutingPredictor::get_slope() {
-    if (iterations_.size() > min_history_) {
-        auto model = fit_model(iterations_, iteration_overused_rr_node_counts_, history_factor_);
-        return model.get_slope();
-    }
-    return -1;
+float RoutingPredictor::get_slope() const {
```
zhaiyif1, line 66: Please ignore this file.

vpr/src/route/route_timing.h
```diff
@@ -12,6 +12,7 @@
 #include "spatial_route_tree_lookup.h"
 #include "connection_router_interface.h"
 #include "heap_type.h"
+#include "routing_predictor.h"
```
zhaiyif1, line 15: Please ignore this file.

vpr/src/route/route_timing.cpp
```diff
@@ -88,11 +88,12 @@ static bool timing_driven_route_sink(
     unsigned itarget,
     int target_pin,
     const t_conn_cost_params cost_params,
-    int high_fanout_threshold,
+    const t_router_opts& router_opts,
```
zhaiyif1, line 91: Please ignore this file.

vpr/src/route/check_route.cpp
```diff
@@ -20,7 +20,7 @@
 /******************** Subroutines local to this module **********************/
 static void check_node_and_range(int inode, enum e_route_type route_type);
 static void check_source(int inode, ClusterNetId net_id);
-static void check_sink(int inode, ClusterNetId net_id, bool* pin_done);
+static void check_sink(int inode, int net_pin_index, ClusterNetId net_id, bool* pin_done);
```
zhaiyif1, line 23: Please ignore this file.

vpr/src/base/metadata_storage.h
```diff
@@ -28,6 +28,26 @@ class MetadataStorage {
         data_.push_back(std::make_tuple(lookup_key, meta_key, meta_value));
     }
 
+    // Use the given mapping function to change the keys
```
zhaiyif1, line 31: Please ignore this file.

ODIN_II/regression_test/benchmark/verilog/syntax/multi_assignment_output
```diff
@@ -0,0 +1,6 @@
+busout
```
zhaiyif1, line 1: Please ignore this file.

ODIN_II/regression_test/benchmark/verilog/syntax/multi_assignment_input
```diff
@@ -0,0 +1,6 @@
+GLOBAL_SIM_BASE_CLK bus0 bus1 bus2 bus3 enable s
```
zhaiyif1, line 1: Please ignore this file.

ODIN_II/regression_test/benchmark/verilog/syntax/multi_assignment.v
```diff
@@ -0,0 +1,14 @@
+module select_bus(busout, bus0, bus1, bus2, bus3, enable, s);
```
zhaiyif1, line 1: Please ignore this file.

ODIN_II/regression_test/benchmark/task/syntax/synthesis_result.json
```diff
@@ -1702,7 +1702,7 @@
             "matrix_multiplication.v:132:1 [NETLIST] Assignment (matrix_multiplication^tmp7~0) is missing driver",
             "matrix_multiplication.v:69:8 [NETLIST] This output is undriven (matrix_multiplication^data_out~16) and will be removed",
             "matrix_multiplication.v:69:8 [NETLIST] This output is undriven (matrix_multiplication^data_out~17) and will be removed",
-            "matrix_multiplication.v:166:16 [NETLIST] Net (null) driving node matrix_multiplication^nMUX~49^MUX_2~318 is itself undriven.",
+            "matrix_multiplication.v:166:16 [NETLIST] Net (null) driving node matrix_multiplication^nMUX~49^MUX_2~380 is itself undriven.",
```
zhaiyif1, line 1705: Please ignore this file.

ODIN_II/regression_test/benchmark/task/syntax/simulation_result.json
```diff
@@ -2312,6 +2312,33 @@
         "Estimated LUTs": 1020,
         "Total Node": 1315
     },
+    "syntax/multi_assignment/k6_frac_N10_frac_chain_mem32K_40nm": {
```
zhaiyif1, line 2315: Please ignore this file.

ODIN_II/regression_test/benchmark/task/preprocessor/synthesis_result.json
```diff
@@ -13,13 +13,16 @@
...
         "test_name": "preprocessor/elsif_both_defined/no_arch",
         "verilog": "elsif_both_defined.v",
-        "exit": 1,
-        "errors": [
-            "elsif_both_defined.v:18:3 [NETLIST] You've defined this driver simple_op^out twice."
-        ],
         "expectation": [
             "ELSIF this test is expected to pass, if it fails the `elsif are not functioning properly."
-        ]
+        ],
```
zhaiyif1, line 18: Please ignore this file.

ODIN_II/regression_test/benchmark/task/preprocessor/simulation_result.json
```diff
@@ -201,6 +201,87 @@
         "Estimated LUTs": 1,
         "Total Node": 1
     },
+    "preprocessor/elsif_both_defined/no_arch": {
```
zhaiyif1, line 204: Please ignore this file.

ODIN_II/regression_test/benchmark/task/large/synthesis_result.json
```diff
@@ -25,7 +25,6 @@
         "architecture": "k6_frac_N10_frac_chain_mem32K_40nm.xml",
         "verilog": "arm_core.v",
         "warnings": [
```
zhaiyif1, line 27: Please ignore this file.

ODIN_II/regression_test/benchmark/task/full/synthesis_result.json
```diff
@@ -437,7 +437,6 @@
         "architecture": "k6_frac_N10_frac_chain_mem32K_40nm.xml",
         "verilog": "matmul.v",
         "warnings": [
```
zhaiyif1, line 439: Please ignore this file.

ODIN_II/regression_test/benchmark/task/fpu/hard_logic/simulation_result.json
```diff
@@ -14,7 +14,7 @@
         "blif": "bgm.blif",
         "exit": 134,
         "errors": [
-            "[OUTPUT_BLIF] Could not resolve memory hard block fpu_mul~0 to a valid type."
+            "[OUTPUT_BLIF] Could not resolve memory hard block fpu_mul~6 to a valid type."
```
zhaiyif1, line 17: Please ignore this file.

ODIN_II/SRC/subtractions.cpp
```diff
@@ -405,11 +405,12 @@ void split_adder_for_sub(nnode_t* nodeo, int a, int b, int sizea, int sizeb, int
         for (i = 0; i < b; i++) {
             /* If the input pin of not gate connects to gnd, replacing the input pin and the not gate with vcc;
              * if the input pin of not gate connects to vcc, replacing the input pin and the not gate with gnd.*/
-            if (not_node[i]->input_pins[0]->net->driver_pin->node->type == GND_NODE) {
+            oassert(not_node[i]->input_pins[0]->net->num_driver_pins == 1);
```
zhaiyif1, line 408: Please ignore this file.

vpr/src/util/vpr_utils.cpp
```diff
@@ -2356,3 +2356,80 @@ void print_timing_stats(std::string name,
             current.num_full_hold_updates - past.num_full_hold_updates,
             current.num_full_setup_hold_updates - past.num_full_setup_hold_updates);
 }
+
+void alloc_legal_placement_locations(std::vector<std::vector<std::vector<t_pl_loc>>>& legal_pos,
```
vaughnbetz, line 2360: Please add a TODO comment: replace num_legal_pos by using the .size() of legal_pos instead. We'll leave this for the future but good to mark it down.

vaughnbetz (approved):

vaughnbetz:
Thanks Daniel!  The code readability is greatly improved; nice work.  Thanks for fitting in the refactorings to vtr::Rect, matrix and pulling out common initial_placement code.
This all looks good.  I made a few suggestions for minor additional commenting.  Please make those changes and then this is good to be merged.

vpr/src/place/analytic_placer.cpp
```diff
@@ -0,0 +1,852 @@
...
+            colnnz.push_back(int(A_col.size()));
+        mat.reserve(colnnz); // reserve memory for mat depending on number of entries in each row
+        for (int col = 0; col < int(A.size()); col++) {
+            auto& A_col = A.at(col);
+            for (auto& row_entry : A_col)
+                mat.insert(row_entry.first, col) = row_entry.second;
+        }
+
+        // use current value of x as guess for iterative solver
+        // feeding next solution back to solver and repeat can achieve better results
```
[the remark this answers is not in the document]
zhaiyif1: sorry, feeding solution back to solver happends outside EquationSystem::solve in AnalyticPlacer::build_solve_direction. The exact number of build_solve_iter can be tuned in AP cfg. I will remove line 93 here to avoid confusion.

vaughnbetz:
Thanks Daniel.  Looks good; just waiting for CI to finish and will merge.

Bill-hbrhbr:
Hey Daniel @zhaiyif1 , I just merged my PR, so now you might have a couple conflicting files waiting to be merged. If you have any questions resolving these merge conflicts, feel free to @ me in this thread.

vpr/src/place/place.cpp
```diff
@@ -745,15 +758,20 @@ void try_place(const t_placer_opts& placer_opts,
...
+#endif /* ENABLE_ANALYTIC_PLACE */
+        first_t = starting_t(&costs, &prev_inverse_costs,
+                             annealing_sched, move_lim, first_rlim,
+                             place_delay_model.get(),
+                             placer_criticalities.get(),
+                             timing_info.get(),
+                             *move_generator,
+                             pin_timing_invalidator.get(),
+                             blocks_affected,
+                             placer_opts);
```
Bill-hbrhbr, line 774: It seems like if you are using the analytic placer, then the starting temperature for the annealing is 0. Is your intention jumping straight to the placement quench part (1X placement inner loop), or perform the usual annealing (**huge number**X placement inner loop), albeit with a temperature of 0?

zhaiyif1:
Hi Bill,

I'll checkout the conflicts soon. My reason for setting start_t to 0 is to
go straight to quench.

vaughnbetz:
Will merge once the conflicts are resolved and CI finishes.
Daniel, it's not gating, but it would also be good if you ran a benchmark sweep with your favourite options for the analytic placer + quench so we have final data attached to this PR. Also further tests the final code.

vpr/src/place/place.cpp
zhaiyif1: I'll checkout the conflicts soon. My reason for setting start_t to 0 is to
go straight to quench.

vpr/src/place/place.cpp
```diff
@@ -891,7 +891,11 @@ void try_place(const t_placer_opts& placer_opts,
     } while (update_annealing_state(&state, success_rat, costs, placer_opts, annealing_sched));
     /* Outer loop of the simmulated annealing ends */
 
+#ifdef ENABLE_ANALYTIC_PLACE
+// guard quench label, otherwise compiler complains about unused label
 quench:
```
vaughnbetz, line 896: Didn't realize there was now a goto in place.cpp.  This would be good to restructure. It could be restructured by:

#ENABLE_ANALYTIC_PLACE wouldn't guard the command line option, but would print a message saying you have to compile with that #define to use it.  Now you can test on placer_opts.enable_analytic_placer anywhere you want.

Then restructure the do_while for the annealer into a while loop controlled like 
exit_anneal = !placer_opts.enable_analytic_placer  // Skip the anneal with analytic placer, otherwise start the anneal
while (!exit_anneal) {

 .... lots of stuff
  exit+anneal = update_annealing_state (...)

Not sure if it is best to land the PR first or restructure first -- the restructuring is pretty local.

zhaiyif1: I had a conversation with Bill and he told me to use goto for now and he will take care of it when he refactors place.cpp later. I think he changed the original while loop to do_while so perhaps he has some preferences. @Bill-hbrhbr.

vaughnbetz: OK, thanks.  Will merge when kokoro finishes then.  Thinking about this more, we don't need to change the command line option guards to get rid of the goto.  Just need:
```
exit_anneal = false;
#ifdef ENABLE_ANALYTIC_PLACE;
  if (placer_opts.enable_analytic_placer)
     exit_anneal = true;
# endif

// change to while loop as above.
```

zhaiyif1: Hi Vaughn, the CI tests have passed. Bill confirmed with me that he will remove the goto when he refactors place.cpp. I'm currently running more tests and will update the result in PR description probably by tomorrow. It there's nothing else to change, perhaps you could merge the branch.

vaughnbetz:
Thanks Daniel -- Merged!

Issue #1475, opened 2020-08-07
VPR Memory Error when input pin equivalence applied to DSPs and RAMs


helen-23:
When running the titan_new DLA benchmark variants with the stratixiv_arch.timing.xml architecture under Titan flow, VPR succeeded. However, after input pin equivalence is applied to the M9K block in stratixiv_arch.timing.xml, VPR reported memory access failures during routing.

#### Expected Behaviour
VPR should pass when input pin equivalence is applied to the DSP and RAM blocks in stratixiv_arch.timing.xml,

#### Current Behaviour
when input pin equivalence is applied to the M9K block in stratixiv_arch.timing.xml,

1. For the DLA_BSC and DLA_ELT benchmarks, VPR reported a segmentation fault during prune_route_tree_recurr() in route_tree_timing.cpp. This occurred at the beginning of routing. Please see attached VPR log files and screenshot of command line error messages for details.

2. For the DLA_LRN benchmark, VPR aborted at assert(node) due to node being NULL in the same function mentioned above. Please see attached VPR log file and screenshot of command line error messages for details.

#### Possible Solution
Maybe there are char-type variables in the router that are used to specify size, instead of uint16, so some values ran over the top.

#### Steps to Reproduce
1. checkout my branch, **"vqm2bliff_one_lut_removal"**, which contains all changes required to run the DLA variants
2. unzip attached DLA circuits (DLA_BSC, DLA_ELT, and DLA_LRN)
3. unzip the modified architecture file (stratixiv_arch.timing.experiment2.xml)
4. Run titan_flow script with the DLA circuits and the modified architecture file. DO NOT run titan_flow.py with sanitizer build turned on, because there is currently integer overflow in the hash function due to multiply-add. Please do note to turn on options **--fit** and **--gen_post_fit_netlist**, because the DLA circuits need post-fit netlist for VPR. An example of the command looks like the following:

<titan_dir>/scripts/titan_flow.py \\
	-q DLA_BSC/quartus2_proj/DLA.qpf \\
	-a stratixiv_arch.timing.experiment2.xml \\
	--fit \\
	--gen_post_fit_netlist \\
	--titan_dir <titan_dir> \\
	--vqm2blif_dir <vtr_root_dir>/build/utils/vqm2blif \\
	--quartus_dir /tools/intel/install/fpga/18.1/standard/quartus/bin \\
	--vpr_dir <vtr_root_dir>/vpr

6. unzip vpr.sdc
[...]

PR #1508, opened 2020-08-27
Refactor rt node

helen-23:
**[Edited]**

In this pull request, I have refactored the rt_node and traceback structures, adding a new member in the structure to keep track of the net pin index  (ranging from 1 to fanout) of sinks. I've also updated all parts of the code to properly store and access this new member.

#### Description

- Fixed **traceback_to_route_tree()** function to create and connect separate rt_nodes if the same rr node index is hit more than once walking the traceback.
- Added a new member in the rt_node and traceback structures to keep track of the net pin index of sinks (ranging from 1 to fanout).
- When a new rt_node is created, made sure to add the node's associated net pin index to the new member if it is a sink node. If it's not a sink, -1 is stored. Same applies for traceback instances.
- Updated various other functions to use this new member instead of using the **rr_sink_node_to_pin** mapping,
- Modified **print_route** routine to print out sink net pin index to the router file (.route).
- Modified **load_routing** routine to read in sink net pin index from the router file (.route) and store it in the traceback.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1475

#### Motivation and Context

As part of a Stratix IV architecture experiment, we tried forcing a greedy and false pin-equivalence to the M9K, M144K, and DSP inputs (meaning that although equivalence is set to "full" on the top level input pins, the local routing of these blocks is still partially fully connected).

When running benchmarks with this experiment, VPR failed in routing because it was attempting to delete the same node twice from the route tree. This happened because the the clustered netlist builder for logic blocks mapped the same net to multiple pins in a port, resulting in multiple rt nodes being created with the same rr node index. However, the code is inconsistent: when building up the route tree from traceback, only one rt_node is created for each rr node index found in the traceback, so edges are created between the same two nodes multiple times; this is problematic during deletion.

This never happened in the past because pin equivalence had always been applied to the LABs. The LABs have proper pin equivalence set up in the architecture capture. In a fully equivalent port, each pin should be mapped to a unique port, so the clustered netlist builder doesn't connect a net more than once to a port if it goes to multiple equivalent inputs in this case.

While investigating the case, I found multiple other issues that would arise if multiple sink-type rt nodes share the same rr node index. Several parts of the routing code need to use the net pin index of a node, but the rt_node structure only retains the node's rr node index value, so the rr node index is translated into its corresponding net pin index through  the **rr_sink_node_to_pin** mapping, which could only map a rr-node index to ONE of its corresponding net pin indices.

If the architecture capture is modified properly with true equivalent ports to the DSP and ram blocks, this problem can be avoided. However, this is still a bug that we should address. If we one day take different speed paths to the same sink for different target_pins, we want to associate the route tree branch with the target_pin we were trying to route to (and hence make sure we get the branch with the timing we expect).

One way to fix this problem is to make **rr_sink_node_to_pin** a multimap, but that still makes it difficult to determine which pin a particular rt_node corresponds to if its node index maps to several net pin indices.

The best way to fix this once and for all is to have net pin index stored in the rt_node structure as well, so no more mapping is required.

- Tried running large ML designs (e.g. DLA variants) through VPR with greedy and false pin-equivalence to the M9K, M144K, and DSP inputs, which succeeded with results that make sense.
- Tried running large ML designs (e.g. DLA variants) through VPR with the existing SIV architecture capture, which succeeded with the same resource usage results as before. Fix did not increase VPR runtime.
- Passed regression test and QoR checks

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/read_route.cpp
+++ b/vpr/src/base/read_route.cpp
@@ -216,7 +216,7 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
 
     /*remember the position of the last line in order to go back*/
     std::streampos oldpos = fp.tellg();
-    int inode, x, y, x2, y2, ptc, switch_id, net_pin_id, offset;
+    int inode, x, y, x2, y2, ptc, switch_id, net_pin_index, offset;
     std::string prev_type;
     int node_count = 0;
     std::string input;
@@ -348,23 +348,26 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
                 switch_id = atoi(tokens[7 + offset].c_str());
             }
 
-            /*Process net pin index for sinks*/
+            /* Process net pin index for sinks                                   *
+             * If you have an old .route file, it may not have this information  *
+             * Please check your .route file to see if it contains Net_pin_index *
+             * information for sinks. If not, plrase re-generate the routing.    */
             if (tokens[2] == "SINK") {
                 if (tokens[8 + offset] == "Net_pin_index:") {
-                    net_pin_id = atoi(tokens[9 + offset].c_str());
+                    net_pin_index = atoi(tokens[9 + offset].c_str());
                 } else {
                     vpr_throw(VPR_ERROR_ROUTE, filename, lineno,
-                              "%d (sink) node does not have net pin index", inode);
+                              "%d (sink) node does not have net pin index. If you are using an old .route file without this information, please re-generate the routing.", inode);
                 }
             } else {
-                net_pin_id = OPEN;
+                net_pin_index = OPEN; //net pin index is invalid for non-SINKs
             }
 
             /* Allocate and load correct values to trace.head*/
             if (node_count == 0) {
                 route_ctx.trace[inet].head = alloc_trace_data();
                 route_ctx.trace[inet].head->index = inode;
-                route_ctx.trace[inet].head->ipin = net_pin_id;
+                route_ctx.trace[inet].head->net_pin_index = net_pin_index;
                 route_ctx.trace[inet].head->iswitch = switch_id;
                 route_ctx.trace[inet].head->next = nullptr;
                 tptr = route_ctx.trace[inet].head;
@@ -373,7 +376,7 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
                 tptr->next = alloc_trace_data();
                 tptr = tptr->next;
                 tptr->index = inode;
-                tptr->ipin = net_pin_id;
+                tptr->net_pin_index = net_pin_index;
                 tptr->iswitch = switch_id;
                 tptr->next = nullptr;
                 node_count++;
--- a/vpr/src/base/read_route.cpp
+++ b/vpr/src/base/read_route.cpp
@@ -356,6 +356,8 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
                     vpr_throw(VPR_ERROR_ROUTE, filename, lineno,
                               "%d (sink) node does not have net pin index", inode);
                 }
+            } else {
+                net_pin_id = OPEN;
             }
 
             /* Allocate and load correct values to trace.head*/
--- a/vpr/src/base/read_route.cpp
+++ b/vpr/src/base/read_route.cpp
@@ -216,7 +216,7 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
 
     /*remember the position of the last line in order to go back*/
     std::streampos oldpos = fp.tellg();
-    int inode, x, y, x2, y2, ptc, switch_id, offset;
+    int inode, x, y, x2, y2, ptc, switch_id, net_pin_id, offset;
     std::string prev_type;
     int node_count = 0;
     std::string input;
@@ -348,10 +348,21 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
                 switch_id = atoi(tokens[7 + offset].c_str());
             }
 
+            /*Process net pin index for sinks*/
+            if (tokens[2] == "SINK") {
+                if (tokens[8 + offset] == "Net_pin_index:") {
+                    net_pin_id = atoi(tokens[9 + offset].c_str());
+                } else {
+                    vpr_throw(VPR_ERROR_ROUTE, filename, lineno,
+                              "%d (sink) node does not have net pin index", inode);
+                }
+            }
+
             /* Allocate and load correct values to trace.head*/
             if (node_count == 0) {
                 route_ctx.trace[inet].head = alloc_trace_data();
                 route_ctx.trace[inet].head->index = inode;
+                route_ctx.trace[inet].head->ipin = net_pin_id;
                 route_ctx.trace[inet].head->iswitch = switch_id;
                 route_ctx.trace[inet].head->next = nullptr;
                 tptr = route_ctx.trace[inet].head;
@@ -360,6 +371,7 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
                 tptr->next = alloc_trace_data();
                 tptr = tptr->next;
                 tptr->index = inode;
+                tptr->ipin = net_pin_id;
                 tptr->iswitch = switch_id;
                 tptr->next = nullptr;
                 node_count++;
--- a/vpr/src/base/vpr_types.h
+++ b/vpr/src/base/vpr_types.h
@@ -1315,15 +1315,16 @@ typedef std::array<vtr::NdMatrix<std::vector<int>, 3>, NUM_RR_TYPES> t_rr_node_i
  * @brief Basic element used to store the traceback (routing) of each net.
  *
  *   @param index    Array index (ID) of this routing resource node.
- *   @param ipin:    Net pin index associated with the node. This value      *
- *                   ranges from 1 to fanout [1..num_pins-1]. For cases when *
- *                   different speed paths are taken to the same sink for    *
- *                   different pins, node index cannot uniquely identify     *
- *                   each sink, so the net pin index guarentees an unique    *
- *                   identification for each sink-type node. For non-sink-   *
- *                   type nodes and for sink-type nodes with no associated   *
- *                   net pin index, the value for this member should be set  *
- *                   to OPEN (-1).                                           *
+ *   @param net_pin_index:    Net pin index associated with the node. This value       
+ *                            ranges from 1 to fanout [1..num_pins-1]. For cases when  
+ *                            different speed paths are taken to the same SINK for     
+ *                            different pins, node index cannot uniquely identify      
+ *                            each SINK, so the net pin index guarantees an unique     
+ *                            identification for each SINK node. For non-SINK nodes    
+ *                            and for SINK nodes with no associated net pin index      
+ *                            (i.e. special SINKs like the source of a clock tree      
+ *                            which do not correspond to an actual netlist connection),
+ *                            the value for this member should be set to OPEN (-1).
  *   @param iswitch  Index of the switch type used to go from this rr_node to
  *                   the next one in the routing.  OPEN if there is no next node
  *                   (i.e. this node is the last one (a SINK) in a branch of the
@@ -1333,7 +1334,7 @@ typedef std::array<vtr::NdMatrix<std::vector<int>, 3>, NUM_RR_TYPES> t_rr_node_i
 struct t_trace {
     t_trace* next;
     int index;
-    int ipin = OPEN;
+    int net_pin_index = OPEN;
     short iswitch;
 };
 
--- a/vpr/src/base/vpr_types.h
+++ b/vpr/src/base/vpr_types.h
@@ -1315,8 +1315,15 @@ typedef std::array<vtr::NdMatrix<std::vector<int>, 3>, NUM_RR_TYPES> t_rr_node_i
  * @brief Basic element used to store the traceback (routing) of each net.
  *
  *   @param index    Array index (ID) of this routing resource node.
- *   @param index    Index of the pin for a sink node. Only used for rt_node
- *                   tree traceback. Otherwise value is OPEN.
+ *   @param ipin:    Net pin index associated with the node. This value      *
+ *                   ranges from 1 to fanout [1..num_pins-1]. For cases when *
+ *                   different speed paths are taken to the same sink for    *
+ *                   different pins, node index cannot uniquely identify     *
+ *                   each sink, so the net pin index guarentees an unique    *
+ *                   identification for each sink-type node. For non-sink-   *
+ *                   type nodes and for sink-type nodes with no associated   *
+ *                   net pin index, the value for this member should be set  *
+ *                   to OPEN (-1).                                           *
  *   @param iswitch  Index of the switch type used to go from this rr_node to
  *                   the next one in the routing.  OPEN if there is no next node
  *                   (i.e. this node is the last one (a SINK) in a branch of the
--- a/vpr/src/base/vpr_types.h
+++ b/vpr/src/base/vpr_types.h
@@ -1315,6 +1315,8 @@ typedef std::array<vtr::NdMatrix<std::vector<int>, 3>, NUM_RR_TYPES> t_rr_node_i
  * @brief Basic element used to store the traceback (routing) of each net.
  *
  *   @param index    Array index (ID) of this routing resource node.
+ *   @param index    Index of the pin for a sink node. Only used for rt_node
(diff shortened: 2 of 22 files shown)


helen-23:
Hi Vaughn,
I have this branch ready for review. The only Travis CI failure is the sporadic Python Lint issue. Did not change any python scripts in this pull request. Thank you.

vpr/src/route/route_common.cpp
```diff
@@ -86,7 +86,7 @@ static int num_linked_f_pointer_allocated = 0;
  *                                                                          */
 
 /******************** Subroutines local to route_common.c *******************/
-static t_trace_branch traceback_branch(int node, std::unordered_set<int>& main_branch_visited);
+static t_trace_branch traceback_branch(int node, int target_pin, std::unordered_set<int>& main_branch_visited);
```
vaughnbetz, line 89: I'd use a more descriptive variable name:  target_net_pin_index would be better I think

vpr/src/route/route_common.cpp
```diff
@@ -494,7 +494,7 @@ void init_route_structs(int bb_factor) {
     route_ctx.net_status.resize(cluster_ctx.clb_nlist.nets().size());
 }
 
-t_trace* update_traceback(t_heap* hptr, ClusterNetId net_id) {
+t_trace* update_traceback(t_heap* hptr, int target_pin, ClusterNetId net_id) {
```
vaughnbetz, line 497: Rename and explain target_pin in the comment.  Also, we've changed style to comments before function so better to move the comment above the function.

vpr/src/route/route_common.cpp
```diff
@@ -532,7 +532,7 @@ t_trace* update_traceback(t_heap* hptr, ClusterNetId net_id) {
 
 //Traces back a new routing branch starting from the specified 'node' and working backwards to any existing routing.
 //Returns the new branch, and also updates trace_nodes for any new nodes which are included in the branches traceback.
-static t_trace_branch traceback_branch(int node, std::unordered_set<int>& trace_nodes) {
+static t_trace_branch traceback_branch(int node, int target_pin, std::unordered_set<int>& trace_nodes) {
```
vaughnbetz, line 535: Rename and explain target_pin.

vpr/src/route/route_common.h
```diff
@@ -22,7 +22,7 @@ void pathfinder_update_acc_cost_and_overuse_info(float acc_fac, OveruseInfo& ove
 
 float update_pres_fac(float new_pres_fac);
 
-t_trace* update_traceback(t_heap* hptr, ClusterNetId net_id);
+t_trace* update_traceback(t_heap* hptr, int target_pin, ClusterNetId net_id);
```
vaughnbetz, line 25: target_pin -> target_net_pin_index 
and add a comment on usage.

vpr/src/route/route_tree_timing.cpp
```diff
@@ -45,6 +52,7 @@ static t_linked_rt_edge* alloc_linked_rt_edge();
 static void free_linked_rt_edge(t_linked_rt_edge* rt_edge);
 
 static t_rt_node* add_subtree_to_route_tree(t_heap* hptr,
+                                            int target_pin,
```
vaughnbetz, line 55: Rename target_pin to target_net_pin_index.  If you understand enough about what the routine does, write a comment here on what it does and what the input parameters are (if you don't feel confident in writing that, skip it).

vpr/src/route/route_tree_timing.cpp
```diff
@@ -199,7 +208,7 @@ t_rt_node* init_route_tree_to_source(ClusterNetId inet) {
  * updates the Tdel, etc. numbers for the rest of the routing tree.  hptr
  * is the heap pointer of the SINK that was reached.  This routine returns
  * a pointer to the rt_node of the SINK that it adds to the routing.        */
-t_rt_node* update_route_tree(t_heap* hptr, SpatialRouteTreeLookup* spatial_rt_lookup) {
+t_rt_node* update_route_tree(t_heap* hptr, int target_pin, SpatialRouteTreeLookup* spatial_rt_lookup) {
```
vaughnbetz, line 211: rename target_pin and add a description of it to the comment.

vpr/src/route/route_tree_timing.cpp
```diff
@@ -241,7 +250,12 @@ t_rt_node* update_route_tree(t_heap* hptr, SpatialRouteTreeLookup* spatial_rt_lo
 
 void add_route_tree_to_rr_node_lookup(t_rt_node* node) {
     if (node) {
-        VTR_ASSERT(rr_node_to_rt_node[node->inode] == nullptr || rr_node_to_rt_node[node->inode] == node);
+        auto& device_ctx = g_vpr_ctx.device();
+        if (device_ctx.rr_nodes[node->inode].type() == SINK) {
```
vaughnbetz, line 254: Need to explain the logic of this ASSERT.  The rr_node that is used in this rt_node has either (1) not been added into the routing (rt_tree) for this net before, or if it was added the rr_node_to_rt_node mapping structure will point back at the rt_node we're processing so we're just branching off that point.  Exception: for SINK nodes, some netlists and input pin equivalences can lead to us routing to the same SINK more than once on a net.  Hence for SINKs we assert on a weaker condition that if this SINK is already in the rt_tree, the rr_node_to_rt_node mapping structure points to a legal rt_node (but not necessarily the only one) containing the SINK.

vpr/src/route/route_tree_timing.cpp
```diff
@@ -252,7 +266,7 @@ void add_route_tree_to_rr_node_lookup(t_rt_node* node) {
 }
 
 static t_rt_node*
-add_subtree_to_route_tree(t_heap* hptr, t_rt_node** sink_rt_node_ptr) {
+add_subtree_to_route_tree(t_heap* hptr, int target_pin, t_rt_node** sink_rt_node_ptr) {
```
vaughnbetz, line 269: Rename target_pin to target_net_pin_index, update comment to mention it, and move comment above function.

vpr/src/route/route_tree_timing.cpp
```diff
@@ -316,6 +331,7 @@ add_subtree_to_route_tree(t_heap* hptr, t_rt_node** sink_rt_node_ptr) {
 
         rt_node->u.child_list = linked_rt_edge;
         rt_node->inode = inode;
+        rt_node->ipin = OPEN;
```
vaughnbetz, line 334: Explain why:  not valid for non-SINK nodes.

vpr/src/route/route_tree_timing.cpp
```diff
@@ -347,6 +363,7 @@ add_subtree_to_route_tree(t_heap* hptr, t_rt_node** sink_rt_node_ptr) {
 
     //Expand (recursively) each of the main-branch nodes adding any
     //non-configurably connected nodes
+    //Sink is not included, so no need to pass in the node's ipin value.
```
vaughnbetz, line 366: I don't see any pin index passed in, so I don't understand the comment.

vpr/src/route/route_tree_timing.h
```diff
@@ -19,7 +19,7 @@ void free_route_tree(t_rt_node* rt_node);
 void print_route_tree(const t_rt_node* rt_node);
 void print_route_tree(const t_rt_node* rt_node, int depth);
 
-t_rt_node* update_route_tree(t_heap* hptr, SpatialRouteTreeLookup* spatial_rt_lookup);
+t_rt_node* update_route_tree(t_heap* hptr, int target_pin, SpatialRouteTreeLookup* spatial_rt_lookup);
```
vaughnbetz, line 22: Rename target_pin to target_net_pin_index, and if you know what the routine does, write a comment above the function name.

vpr/src/route/route_tree_type.h
```diff
@@ -27,6 +27,13 @@ struct t_linked_rt_edge {
  * parent_switch:  Index of the switch type driving this node (by its        *
  *                 parent).                                                  *
  * inode:  index (ID) of the rr_node that corresponds to this rt_node.       *
+ * ipin:  Net pin index associated with the rt_node. This value ranges from  *
```
vaughnbetz, line 30: net_pin_index would be a better name than ipin, both here and in t_trace, in retrospect.  The comment is good; you could change the member name too (here and in t_trace) if you have the energy for it :).

vpr/src/timing/net_delay.cpp
```diff
@@ -77,29 +77,29 @@ static void load_one_net_delay(ClbNetPinsMatrix<float>& net_delay, ClusterNetId
...
+    ipin_to_Tdel_map.clear(); // clear the map
 }
 
 static void load_one_net_delay_recurr(t_rt_node* node, ClusterNetId net_id) {
-    /* This routine recursively traverses the route tree, and copies the Tdel of the node into the map.  */
-
-    inode_to_Tdel_map[node->inode] = node->Tdel; // add to the map, process current node
+    /* This routine recursively traverses the route tree, and copies the Tdel of the sink_type nodes *
+     * into the map.                                                                                 */
+    if (node->ipin != OPEN) {
```
vaughnbetz, line 100: Comment that this means non-SINK.

vpr/src/route/connection_based_routing.h
```diff
@@ -19,6 +19,10 @@ class Connection_based_routing_resources {
     // each net maps SINK node index -> PIN index for net
     // only need to be built once at the start since the SINK nodes never change
     // the reverse lookup of route_ctx.net_rr_terminals
+    // be careful: it is possible for multiple sinks to share the same node index in some cases.
```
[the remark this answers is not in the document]
helen-23: Actually, nothing uses this anymore, not even the BF router. There is just one function that does a sanity check of the map, called by an assertion in route_timing.cpp, which no longer uses this anyways.

I can go ahead and remove this altogether.

vpr/src/route/route_tree_timing.cpp
helen-23: My apologies, I sort of put this in as a reminder for myself that there is no need to pass in any net pin info to this function. But I agree it's confusing. I am taking it out.

vpr/src/route/route_tree_timing.h
helen-23: I never added a comment because the description for these functions are in the .cpp. But I've just added to the comment to describe what the target_net_pin_index parameter means.

vpr/src/base/read_route.cpp
```diff
@@ -348,10 +348,23 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
                 switch_id = atoi(tokens[7 + offset].c_str());
             }
 
+            /*Process net pin index for sinks*/
+            if (tokens[2] == "SINK") {
+                if (tokens[8 + offset] == "Net_pin_index:") {
+                    net_pin_id = atoi(tokens[9 + offset].c_str());
+                } else {
+                    vpr_throw(VPR_ERROR_ROUTE, filename, lineno,
```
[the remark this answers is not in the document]
helen-23: Added message in the vpr_throw.

vpr/src/base/vpr_types.h
```diff
@@ -1315,6 +1315,15 @@ typedef std::array<vtr::NdMatrix<std::vector<int>, 3>, NUM_RR_TYPES> t_rr_node_i
  * @brief Basic element used to store the traceback (routing) of each net.
  *
  *   @param index    Array index (ID) of this routing resource node.
+ *   @param ipin:    Net pin index associated with the node. This value      *
+ *                   ranges from 1 to fanout [1..num_pins-1]. For cases when *
+ *                   different speed paths are taken to the same sink for    *
+ *                   different pins, node index cannot uniquely identify     *
+ *                   each sink, so the net pin index guarentees an unique    *
+ *                   identification for each sink-type node. For non-sink-   *
+ *                   type nodes and for sink-type nodes with no associated   *
```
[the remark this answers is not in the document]
helen-23: Added comment. Also added the same comments for rt_node.

vpr/src/route/route_common.cpp
helen-23: Good idea, while I'm at it, I also changed the parameter name in update_traceback, which is the routine a level above this.

vpr/src/route/route_common.cpp
helen-23: Ah, so this is an odd one out with comments inside the function brackets. I was wondering why it looked a bit different :)
Done changing parameter name and adding/fixing comment.

vpr/src/route/route_common.cpp
helen-23: Yup, also done.

vpr/src/route/route_common.h
helen-23: Done.

vpr/src/route/route_tree_timing.cpp
helen-23: There is also already a comment at where the function is defined, which is in this file as well. I guess this is why none of the function declarations at the top of the file had comments to them. But if you think it's a good idea and a good convention to do so, I can add one in.

At where the function is defined, I moved the comment on top of the function and updated comments with the new parameter meaning.

vpr/src/route/route_tree_timing.cpp
helen-23: Done.

vpr/src/route/route_tree_timing.cpp
helen-23: Now that I think about it, this function serves to record ALL nodes from the rt_tree (built from traceback) in the rr_node_to_rt_node lookup, this is to get the mapping ready so that new branches can be added knowing where the branching points are in the existing tree. This is done recursively in a depth-first traversal, so technically, wouldn't it be illegal to reach the exact same rt_node twice no matter what? I think that the assertion should actually be changed to just

```
VTR_ASSERT( rr_node_to_rt_node[node->inode] == nullptr ||
        (rr_node_to_rt_node[node->inode] != node && rr_node_to_rt_node[node->inode]->inode == node->inode) );

```

helen-23: Or something like
```
if (device_ctx.rr_nodes[node->inode].type() == SINK) {
   VTR_ASSERT( rr_node_to_rt_node[node->inode] == nullptr ||
        (rr_node_to_rt_node[node->inode] != node && rr_node_to_rt_node[node->inode]->inode == node->inode) );
} else {
    VTR_ASSERT(rr_node_to_rt_node[node->inode] == nullptr);
}
```
so that we make sure to check it's a SINK node to use the weak assert condition.

vpr/src/route/route_tree_timing.cpp
helen-23: Done.

vpr/src/route/route_tree_timing.cpp
helen-23: Done.

vpr/src/timing/net_delay.cpp
helen-23: Agreed, done.

helen-23:
Hi Vaughn,
I marked most of the above as done just so I can keep track of which ones I've changed locally, though I do not currently have anything new committed yet. I also replied to some o your comments with questions as well. Thank you.

vpr/src/route/route_tree_type.h
helen-23: Agreed and done.

helen-23:
The refactored code increased memory usage less than 1% for all tests conducted. VPR runtime was not effected (placement time, pack time, and route time fluctuates in every run depending on the machine's load)

image

vpr/src/route/route_tree_timing.h
vaughnbetz: Your call; you can keep following the .cpp convention for these files if you like.

vaughnbetz:
Code format failed Helen; can you run make format?
pylint failed too (unrelated, Shad is fixing that so hopefully works after rebase but not your issue anyway).

vaughnbetz:
Thanks for the QoR checks; those look good.

helen-23:
Experiment: re-run routing (just routing and analysis) with astar_fac=0.5
This did improve Fmax for the DLA variants as it does not reach for the bottom-left corner as rigorously, especially for DLA_BSC, which had the lowest Fmax. I think it is reasonable to hypothesize that once the sink coordinates are refactored, it would improve routing decisions. Here is a comparison between the Fmax results:
image

vaughnbetz:
Great news Helen -- thanks.
The Travis CI failure is an unrelated python lint failure (Shad already fixed that on the master).  
Once kokoro finishes (assuming all goes well) I'll merge this.

vaughnbetz:
Everything looks good (except the spurious pylint failure.  Merging.

PR #1511, opened 2020-08-27
travis: Upgrade OS to Bionic

### LINKED ISSUE #1505: Move travis build to xenial or newer as trusty has left supported status
The travis build is currently using `dist: trusty`; see 
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/c5c1c0b59460e31f89e64d6da54b43c7b8b3cb41/.travis.yml#L4

The "End of Standard Support" for Ubuntu 14.04.6 LTS was April 2019 which has meant things like the llvm upstream packages have started to remove support. For example from https://apt.llvm.org/;
> Aug 20th 2019 - Ubuntu Trusty removed (EOL)

We should move to a new LTS version. The options are;
 - xenial - Ubuntu 16.04.7 LTS -- ends April 2021
 - bionic - Ubuntu 18.04.5 LTS -- ends April 2023
 - focal - Ubuntu 20.04.1 LTS -- ends April 2025

I recommend we move to focal as we won't be forced to update for 5 years :-)

mithro:
This moves Travis to use `bionic` which is Ubuntu 18.04.5 LTS and has support until April 2023.

The "End of Standard Support" for Ubuntu 14.04.6 LTS was April 2019 which has meant things like the llvm upstream packages have started to remove support. For example from https://apt.llvm.org/;
> Aug 20th 2019 - Ubuntu Trusty removed (EOL)

Fixes #1505

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ jobs:
         - ./dev/check-format-py.sh
     - stage: Test
       name: "Python Lint"
-      script:     
+      script:
         - python3 ./dev/pylint_check.py
     - stage: Test
       name: "C++ Unit Tests"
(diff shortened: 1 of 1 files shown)


vaughnbetz:
@mithro This seems like a good idea, but CI failed (looks like infrastructure issues).  Can you take a look?

mithro:
@vaughnbetz - The failing Travis check is confused because I accidentally pushed to vtr-verilog-to-routing repo, it is passing now otherwise.

PR #1516, opened 2020-08-31
Apply reordering to node and edge metadata

HackerFoo:
Update RR graph node IDs in node and edge metadata when using RR graph node reordering.

#### Description
Bitstreams produced with node reordering (a `--reorder_rr_graph_nodes_algorithm` other than `none`) were not building with `fasm2bels` due to scrambled metadata.

#### Related Issue
#1270 

#### Motivation and Context
This fixes metadata when using RR node reordering.

`fasm2bels` tests now pass on [`fpga-tool-perf`](https://hydra.vtr.tools/jobset/dusty/fpga-tool-perf#tabs-jobs).

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/metadata_storage.h
+++ b/vpr/src/base/metadata_storage.h
@@ -28,6 +28,26 @@ class MetadataStorage {
         data_.push_back(std::make_tuple(lookup_key, meta_key, meta_value));
     }
 
+    // Use the given mapping function to change the keys
+    void remap_keys(std::function<LookupKey(LookupKey)> key_map) {
+        if (map_.empty()) {
+            for (auto& entry : data_) {
+                std::get<0>(entry) = key_map(std::get<0>(entry));
+            }
+        } else {
+            VTR_ASSERT(data_.empty());
+            for (auto& dict : map_) {
+                for (auto& entry : dict.second) {
+                    for (auto& value : entry.second) {
+                        data_.push_back(std::make_tuple(key_map(dict.first), entry.first, value.as_string()));
+                    }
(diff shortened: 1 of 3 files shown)


vpr/src/base/metadata_storage.h
```diff
@@ -28,6 +28,25 @@ class MetadataStorage {
...
+            }
+        } else {
+            VTR_ASSERT(data_.empty());
+            for (auto& dict : map_) {
+                for (auto& entry : dict.second) {
+                    for (auto& value : entry.second) {
+                        data_.push_back(std::make_tuple(key_map(dict.first), entry.first, value.as_string()));
+                    }
+                }
+            }
```
litghost, line 45: Don't you need to clear the map just before `build_map`?

HackerFoo: Yes, I've added it. Thanks.

vpr/test/test_vpr.cpp
```diff
@@ -126,6 +126,8 @@ TEST_CASE("read_rr_graph_metadata", "[vpr]") {
             "wire.eblif",
             "--route_chan_width",
             "100",
+            "--reorder_rr_graph_nodes_algorithm",
+            "random_shuffle"
```
litghost, line 130: This is too early in the flow, as the metadata isn't added until line 147.

HackerFoo: Fixed. I also tested that commenting out the `std::shuffle` on line 186 causes the test to fail.

PR #1518, opened 2020-09-01
Simplify and strengthen check_sink

helen-23:
Refactored check_sink to use the traceback's net_pin_index so that code can be simplified. Also fixed the weak sink check.

#### Description

- Allow net_pin_index of the traceback to be passed into the check_sink function.
- use the net_pin_index so we can get rid of the loop through all sinks for a match.
- implemented strong check by making sure that we hit a netlist pin exactly once.

#### Motivation and Context

The old check was a weaker check as it wasn't actually checking that we find a sink matching each pin in the net; it just checks that we find N sinks for a fanout N net, and each sink appears to connect to a block that is reasonable (matches some pin on this net).  it was really just counting them and making sure we hit N sinks that were reasonable, but not that each was different. So a strong check replacement is needed.

- Passed Travis CI tests
- VPR succeeded for the larger new Titan DLA_BSC benchmark using post-fit netlist and input pin equivalence to RAMs and DSPs

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/check_route.cpp
+++ b/vpr/src/route/check_route.cpp
@@ -20,7 +20,7 @@
 /******************** Subroutines local to this module **********************/
 static void check_node_and_range(int inode, enum e_route_type route_type);
 static void check_source(int inode, ClusterNetId net_id);
-static void check_sink(int inode, ClusterNetId net_id, std::set<vtr::StrongId<cluster_pin_id_tag>>& pin_done);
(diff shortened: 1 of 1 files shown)


helen-23:
Hi Vaughn,
Good news, the strong check worked with net_pin_index is passed into and used in check_sink. This branch is ready for review. Thank you!

vaughnbetz:
Great, thanks Helen.

PR #1520, opened 2020-09-01
Enabled BLIF buffer elimination in preparation for InOuts

djns99:
#### Description

Enables the logic for eliminating the redundant final buffer in the BLIF output. This is a necessary precursor to InOuts as the extra buffer is not driven by the input correctly

#### Types of changes

#### Checklist:


djns99:
@kmurray This changes Odin II's BLIF output slightly which seems to have had some QoR changes for VPR. The issues on Travis appear to be improvements. What do I need to do to regenerate the golden results to make this pass?

djns99:
Running the regression tests (with -create_golden to regenerate golden) for basic, strong and nightly I get this error:

```vtr_func_formal:          k6_frac_N10_40nm.xml/stereovision3.v/common           			file        :  abc.lec.out
failed: Couldn't determine Logical Equivalence status between stereovision3.odin.blif <-> sv_chip3_hierarchy_no_mem_post_synthesis.blif (took 3.21 seconds)```

I assume this means ABC somehow broke with these changes?

My concern here is that I believe this modification must definitionally be identical behaviour, since the driving net can only have a fanout of one (otherwise I do not do this optimisations) so the drivers of this net must necessarily assign directly to the output wire.

The only thing I can think is that somehow there is a combinational loop formed somewhere, but I dont see how this could be possible since fanout = 1 and we dont have InOuts yet (and if we did it would be a bug in stereovision not Odin II)

@jeanlego Thoughts?

File: ODIN_II/SRC/output_blif.cpp [hunk 1]
```diff
@@ -65,25 +65,25 @@ static bool warn_undriven(nnode_t* node, nnet_t* net) {
...
-//}
+static void merge_with_inputs(nnode_t* node, long pin_idx) {
+    oassert(pin_idx < node->num_input_pins);
+    nnet_t* net = node->input_pins[pin_idx]->net;
+    warn_undriven(node, net);
+    // Merge node with all inputs with fanout of 1
+    if (net->num_fanout_pins <= 1) {
+        for (int i = 0; i < net->num_driver_pins; i++) {
+            npin_t* driver = net->driver_pins[i];
+            if (driver->name != NULL && ((driver->node->type == MULTIPLY) || (driver->node->type == HARD_IP) || (driver->node->type == MEMORY) || (driver->node->type == ADD) || (driver->node->type == MINUS))) {
```
jeanlego, line 76: given you are calling it on the output node only, this will never be reached.
I also don't forsee this being used for anything else but output. can you assert that node->type is OUTPUT_NODE
I would rather not see this used anywhere else in the future unless theres a good reason for it, so guarding it should make that clear

vaughnbetz:
Does @jeanlego's suggested code change above fix the logical equivalence failure or is there still an issue with that?

jeanlego:
> Does @jeanlego's suggested code change above fix the logical equivalence failure or is there still an issue with that?

Sadly no but I'm out so I can't pull the code to diff the resulting blif

ODIN_II/SRC/output_blif.cpp
djns99: I assume you mean the input node? The driver can be anything that drives an output yeah?

vaughnbetz:
Assigning this to @sdamghan to see if he wants to finish this PR, or abandon it.

sdamghan:
@vaughnbetz not sure why this is left, I was not aware of the progress of this PR. Will investigate more in the future

PR #1522, opened 2020-09-02
Refactor the placer debug code

MohamedElgammal:
#### Description

This PR follows #1287 . In this PR, some code refactoring was done while preserving the same functionality.
#### Motivation and Context

Taking into consideration some code review comments on #1287 

test that the UI is working for multiple VPR benchmarks
test that the feature is disabled if the compilation or the code flag is disabled

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -998,16 +998,16 @@ static void drawplace(ezgl::renderer* g) {
                 ezgl::color block_color;
                 t_logical_block_type_ptr logical_block_type = nullptr;
 #    ifdef VTR_ENABLE_DEBUG_LOGGING
-                if (f_placer_debug) {
-                    t_pl_loc curr_loc;
-                    curr_loc.x = i;
-                    curr_loc.y = j;
-                    auto it = std::find_if(draw_state->colored_blocks.begin(), draw_state->colored_blocks.end(), [&curr_loc](const std::pair<t_pl_loc, ezgl::color>& a) { return (a.first.x == curr_loc.x && a.first.y == curr_loc.y); });
-                    if (it != draw_state->colored_blocks.end()) {
+                t_pl_loc curr_loc;
+                curr_loc.x = i;
+                curr_loc.y = j;
+                auto it = std::find_if(draw_state->colored_blocks.begin(), draw_state->colored_blocks.end(), [&curr_loc](const std::pair<t_pl_loc, ezgl::color>& a) { return (a.first.x == curr_loc.x && a.first.y == curr_loc.y); });
+                if (placer_debug_enabled() && it != draw_state->colored_blocks.end()) {
                         block_color = it->second;
                         auto tile_type = device_ctx.grid[i][j].type;
                         logical_block_type = pick_best_logical_type(tile_type);
                     } else {
+#   endif
                         if (bnum != EMPTY_BLOCK_ID) {
                             block_color = draw_state->block_color(bnum);
                             logical_block_type = cluster_ctx.clb_nlist.block_type(bnum);
@@ -1018,31 +1018,9 @@ static void drawplace(ezgl::renderer* g) {
                             auto tile_type = device_ctx.grid[i][j].type;
                             logical_block_type = pick_best_logical_type(tile_type);
                         }
-                    }
-                } else {
-                    if (bnum != EMPTY_BLOCK_ID) {
-                        block_color = draw_state->block_color(bnum);
-                        logical_block_type = cluster_ctx.clb_nlist.block_type(bnum);
-                    } else {
-                        block_color = get_block_type_color(device_ctx.grid[i][j].type);
-                        block_color = lighten_color(block_color, EMPTY_BLOCK_LIGHTEN_FACTOR);
-
-                        auto tile_type = device_ctx.grid[i][j].type;
-                        logical_block_type = pick_best_logical_type(tile_type);
-                    }
+#   ifdef VTR_ENABLE_DEBUG_LOGGING
                 }
-#    else
-                if (bnum != EMPTY_BLOCK_ID) {
-                    block_color = draw_state->block_color(bnum);
-                    logical_block_type = cluster_ctx.clb_nlist.block_type(bnum);
-                } else {
-                    block_color = get_block_type_color(device_ctx.grid[i][j].type);
-                    block_color = lighten_color(block_color, EMPTY_BLOCK_LIGHTEN_FACTOR);
-
-                    auto tile_type = device_ctx.grid[i][j].type;
-                    logical_block_type = pick_best_logical_type(tile_type);
-                }
-#    endif
+#   endif
                 g->set_color(block_color);
                 /* Get coords of current sub_tile */
                 ezgl::rectangle abs_clb_bbox = draw_coords->get_absolute_clb_bbox(i, j, k, logical_block_type);
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -48,9 +48,7 @@
 #include "route_common.h"
 #include "breakpoint.h"
 
-#ifdef VTR_ENABLE_DEBUG_LOGGING
-#    include "move_utils.h"
-#endif
+#include "move_utils.h"
 
 #ifdef WIN32 /* For runtime tracking in WIN32. The clock() function defined in time.h will *
               * track CPU runtime.														   */
@@ -78,6 +76,8 @@
 /****************************** Define Macros *******************************/
 
 #    define DEFAULT_RR_NODE_COLOR ezgl::BLACK
+#   define OLD_BLK_LOC_COLOR blk_GOLD
+#   define NEW_BLK_LOC_COLOR blk_GREEN
 //#define TIME_DRAWSCREEN /* Enable if want to track runtime for drawscreen() */
 
 /********************** Subroutines local to this module ********************/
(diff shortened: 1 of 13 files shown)


vpr/src/draw/draw.cpp
```diff
@@ -998,16 +998,16 @@ static void drawplace(ezgl::renderer* g) {
                 ezgl::color block_color;
                 t_logical_block_type_ptr logical_block_type = nullptr;
 #    ifdef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 1000: Do we need this guard?  Does it only control the breakpoint functionality or something else?  Does compiling in breakpoints slow the tool down?
If not, we can just eliminate this #ifdef throughout the code.

vaughnbetz: I'd factor this into a subroutine; shorter routines with meaningful names are easier to understand and change.

vpr/src/draw/draw.cpp
```diff
@@ -998,16 +998,16 @@ static void drawplace(ezgl::renderer* g) {
                 ezgl::color block_color;
```
vaughnbetz, line 998: Put the continue for if (bnum == ...) on a separate, indented line.

vpr/src/draw/draw.cpp
```diff
@@ -998,16 +998,16 @@ static void drawplace(ezgl::renderer* g) {
...
-                if (f_placer_debug) {
-                    t_pl_loc curr_loc;
-                    curr_loc.x = i;
-                    curr_loc.y = j;
-                    auto it = std::find_if(draw_state->colored_blocks.begin(), draw_state->colored_blocks.end(), [&curr_loc](const std::pair<t_pl_loc, ezgl::color>& a) { return (a.first.x == curr_loc.x && a.first.y == curr_loc.y); });
-                    if (it != draw_state->colored_blocks.end()) {
+                t_pl_loc curr_loc;
+                curr_loc.x = i;
+                curr_loc.y = j;
+                auto it = std::find_if(draw_state->colored_blocks.begin(), draw_state->colored_blocks.end(), [&curr_loc](const std::pair<t_pl_loc, ezgl::color>& a) { return (a.first.x == curr_loc.x && a.first.y == curr_loc.y); });
```
vaughnbetz, line 1004: Use a better name than a.
colored_blocks should really be called colored_locations (I think).

vpr/src/draw/draw.cpp
```diff
@@ -998,16 +998,16 @@ static void drawplace(ezgl::renderer* g) {
...
+                if (placer_debug_enabled() && it != draw_state->colored_blocks.end()) {
```
vaughnbetz, line 1005: Is placer_debug_enabled() the right name?  Does this really mean we are in a breakpoint?  If so worth renaming.

vaughnbetz, line 1005: Comment: found a colored location at the spot I am drawing (currently used for drawing the current move).  This overrides any block color.

vpr/src/draw/draw.cpp
```diff
@@ -998,16 +998,16 @@ static void drawplace(ezgl::renderer* g) {
...
                         block_color = it->second;
                         auto tile_type = device_ctx.grid[i][j].type;
                         logical_block_type = pick_best_logical_type(tile_type);
                     } else {
```
vaughnbetz, line 1009: No color specified at this location; use the block color.

vpr/src/draw/draw.cpp
```diff
@@ -4157,4 +4135,16 @@ static void run_graphics_commands(std::string commands) {
     ++draw_state->sequence_number;
 }
 
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+// location highlighting functions 
+void set_draw_loc_color(t_pl_loc loc, ezgl::color clr){
+    t_draw_state* draw_state = get_draw_state_vars();
+    draw_state->colored_blocks.push_back(std::make_pair(loc,clr));
```
vaughnbetz, line 4142: change variable name to colored_locs

vpr/src/draw/draw.cpp
```diff
@@ -4157,4 +4135,16 @@ static void run_graphics_commands(std::string commands) {
     ++draw_state->sequence_number;
 }
 
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+// location highlighting functions
```
vaughnbetz, line 4139: Expand this:  pass in an (x,y,subtile) location and the color in which it should be drawn.  This overrides the color of any block placed in that location, and also applies if the location is empty.

vpr/src/draw/draw.cpp
```diff
@@ -4157,4 +4135,16 @@ static void run_graphics_commands(std::string commands) {
     ++draw_state->sequence_number;
 }
 
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+// location highlighting functions 
+void set_draw_loc_color(t_pl_loc loc, ezgl::color clr){
+    t_draw_state* draw_state = get_draw_state_vars();
+    draw_state->colored_blocks.push_back(std::make_pair(loc,clr));    
+}
+void clear_colored_blocks(){
```
vaughnbetz, line 4144: Add a blank line (vertical spacing).
Add a function level comment.

vpr/src/draw/draw.h
```diff
@@ -84,6 +84,12 @@ float get_net_alpha();
 
 ezgl::color get_block_type_color(t_physical_tile_type_ptr type);
 
+#ifdef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 87: Remove the #ifdef (unless you show a slowdown).

vpr/src/draw/draw.h
```diff
@@ -84,6 +84,12 @@ float get_net_alpha();
...
+// location highlighting functions 
+void set_draw_loc_color(t_pl_loc , ezgl::color );
```
vaughnbetz, line 89: Move the function level comments (or duplicate them) here.

vpr/src/place/move_utils.cpp
```diff
@@ -8,9 +8,8 @@
 //Run-time flag to control when placer debug information is printed
```
vaughnbetz, line 8: First two lines of this comment can probably be deleted. (get rid of #ifdef and second line seems less detailed / accurate than the rest of the comment).

vpr/src/place/move_utils.h
```diff
@@ -57,4 +51,8 @@ bool find_to_loc_uniform(t_logical_block_type_ptr type,
                          float rlim,
                          const t_pl_loc from,
                          t_pl_loc& to);
+
+//Accessor and setter for f_placer_debug
```
vaughnbetz, line 55: Explain what setting this to true or false means.  (idea: person who wants to call these should only have to read the .h file.

vaughnbetz: Is placer_debug_enabled the right name?  Or should it be placer_breakpoint_reached? (or something like that).

vpr/src/place/place.cpp
```diff
@@ -45,13 +45,17 @@
 #include "tatum/TimingReporter.hpp"
 
 #ifdef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 47: Get rid of ifdef if possible.

vpr/src/place/place.cpp
```diff
@@ -45,13 +45,17 @@
...
+#    include "draw.h"
 #    include "draw_types.h"
 #    include "draw_global.h"
 #    include "draw_color.h"
 #    include "breakpoint.h"
 //map of the available move types and their corresponding type number
 std::map<int, std::string> available_move_types = {
     {0, "Uniform"}};
+
+#define OLD_BLK_LOC_COLOR blk_GOLD
```
vaughnbetz, line 57: Shouldn't these colors be defined in a draw routine?

vpr/src/place/place.cpp
```diff
@@ -3046,13 +3050,13 @@ void stop_placement_and_check_breakopints(t_pl_blocks_to_be_moved& blocks_affect
         get_bp_state_globals()->get_glob_breakpoint_state()->from_block = size_t(blocks_affected.moved_blocks[0].block_num);
 
         //check for breakpoints
-        f_place_debug = check_for_breakpoints(true);
-        if (f_place_debug)
+        set_placer_debug(check_for_breakpoints(true));
```
vaughnbetz, line 3053: Seems like you should do something like:
if (check_for_breakpoints (true)) {  // What does the true flag passed in mean?  Comment this.
   set_placer_debug (true);
   ....
else {

vpr/src/place/place.cpp
```diff
@@ -3037,7 +3041,7 @@ void transform_blocks_affected(t_pl_blocks_to_be_moved blocksAffected) {
 }
 
 #ifdef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 3043: Minimize added code in try_swap (already a critical routine and overly long).  Create high-level functions, if possible in a new place_debug.cpp (or some such) file and call them from here.  Should only see a few calls / condition checks in try_swap for breakpoints and they should be well commented.

vaughnbetz: Looking closer, I think Mahshad already did this.

vpr/src/place/place.cpp
```diff
@@ -3067,6 +3071,10 @@ void stop_placement_and_check_breakopints(t_pl_blocks_to_be_moved& blocks_affect
```
vaughnbetz, line 3071: Consider if this routine should be moved to another file, in /draw or in a placer breakpoints .cpp file (can't recall the exact names).

vaughnbetz:
Various inline comments above ... please take a crack at them and let me know when it's ready for merging.

vpr/src/place/place.cpp
MohamedElgammal: After reviewing the code, I found that the true flag means that the function is called from placer. I added a comment by this meaning.

MohamedElgammal:
@vaughnbetz , I have pushed a new commit that addresses all your comments. please let me know if there are any further comments.

vaughnbetz:
Looks good.  The QoR failure was just a golden result that needed updating.  Merging.

PR #1529, opened 2020-09-09
Refactor place.cpp Part I

Bill-hbrhbr:
#### Description
PR in progress. More to come...
- [X] Created placer context for accessing various data structures.
- [X] Moved timing update related routines to place_timing_update.h/cpp.
- [X] Moved connection delay related routines to place_delay_model.h/cpp.
- [ ] Merge t_placer_costs and t_placer_inverse_costs.
- [ ] Use t_annealing_state to reduce function argument list's length.

#### Motivation and Context
Refactor place.cpp to modularize it and reduce its length and complexity.

#### Types of changes


Bill-hbrhbr:
Hi @vaughnbetz @HackerFoo , this is a PR in progress. Please drop by and leave some comments when you are available so that I can fix potential issues as early as possible!

vpr/src/place/place.cpp
```diff
@@ -83,10 +85,6 @@ using std::min;
 #define UPDATED_ONCE 'U'
 #define GOT_FROM_SCRATCH 'S'
```
vaughnbetz, line 87: Good to add a comment about what should go in this file.  (This is an annealing based, cluster-level placer.  This file ...)

vpr/src/place/place.cpp
```diff
@@ -2479,25 +2003,26 @@ static void alloc_and_load_placement_structs(float place_cost_exp,
     if (placer_opts.place_algorithm.is_timing_driven()) {
         /* Allocate structures associated with timing driven placement */
         /* [0..cluster_ctx.clb_nlist.nets().size()-1][1..num_pins-1]  */
-        connection_delay = make_net_pins_matrix<float>(cluster_ctx.clb_nlist, 0.f);
-        proposed_connection_delay = make_net_pins_matrix<float>(cluster_ctx.clb_nlist, 0.f);
 
-        connection_setup_slack = make_net_pins_matrix<float>(cluster_ctx.clb_nlist, std::numeric_limits<float>::infinity());
+        p_timing_ctx.connection_delay = make_net_pins_matrix<float>(cluster_ctx.clb_nlist, 0.f);
```
vaughnbetz, line 2007: I'd measure speed regularly on this refactoring to watch for slowdowns.  p_timing_ctx is a global, so if the compiler can't figure out if someone else might be changing the pointer value in connection_delay (called pointer aliasing --maybe some routine has made a copy of p_timing_ctx.connection_delay and is updating the values or the connection_delay pointer itself when you call a subroutine).  If the compiler thinks there is some chance that some routine being called could change some pointer you're using, it will reload the pointer a lot, slowing things down some.

This means sometimes making a local copy of a pointer and using it to fill in or access a data structure can be faster.  It's very hard to reason about because you're guessing what the compiler can and can't do, but it's why it's good to measure speed a lot.  Innocuous seeming changes from a local variable to a larger scope variable can slow things down due to conservative reloading of values by the compiler.

vpr/src/place/place_delay_model.cpp
```diff
@@ -18,10 +20,7 @@
 #    include "serdes_utils.h"
 #endif /* VTR_ENABLE_CAPNPROTO */
```
vaughnbetz, line 22: Please add a comment at the top of the file saying what kind of routines should go in here and what the purpose of the file is.

vpr/src/place/place_delay_model.cpp
```diff
@@ -300,3 +297,83 @@ void OverrideDelayModel::write(const std::string& file) const {
...
+        int sink_block_ipin = cluster_ctx.clb_nlist.pin_logical_index(sink_pin);
+
+        int source_x = place_ctx.block_locs[source_block].loc.x;
+        int source_y = place_ctx.block_locs[source_block].loc.y;
+        int sink_x = place_ctx.block_locs[sink_block].loc.x;
+        int sink_y = place_ctx.block_locs[sink_block].loc.y;
+
+        /**
+         * This heuristic only considers delta_x and delta_y, a much better
+         * heuristic would be to to create a more comprehensive lookup table.
```
vaughnbetz, line 342: Can you take a look in delay_model and see if this is still true? (only uses dx,dy no matter what command line options are used)?  Since we pass in the block pins we could in theory take carry chains into account.

vpr/src/place/place_delay_model.cpp
vaughnbetz: Update: Now I see you have one in the header file.  Might be enough, although I tend to think it's good to put it in the .cpp too.

vpr/src/place/place_delay_model.h
```diff
@@ -123,8 +155,11 @@ class OverrideDelayModel : public PlaceDelayModel {
 
     vtr::flat_map2<t_override, float> delay_overrides_;
```
vaughnbetz, line 156: Good to comment what these are if you can figure this delay_overrides_ data structure out.

vpr/src/place/place_delay_model.h
```diff
@@ -123,8 +155,11 @@ class OverrideDelayModel : public PlaceDelayModel {
...
 
-    //operator< treats memory layout of t_override as an array of short
-    //this requires all members of t_override are shorts and there is no padding between members of t_override
+    /**
```
vaughnbetz, line 158: should mention this was done for speed (array of shorts / tricky code here)

vpr/src/place/place_delay_model.cpp
vaughnbetz: Looked in the delay model file.  True for the delta delay model, but perhaps not for derived classes written in Symbiflow etc.

vaughnbetz (commented):
Looks good Bill, thanks!
The style and commenting looks good.  My two main comments (details embedded) are:
- Please run cpu tests regularly and attach them as you do this refactoring.  I gave one example above of a particularly tricky run time impact (pointer alias analysis) which is impacted by the scope of the variables used in key loops and if their address has ever been taken.  Hard to reason about, so good to check regularly.
- Splitting into multiple files is very good.  I think it's good to think about what code should go in each.  For example, is place.cpp only for annealing? (I think it should be).  place_utils is for utilities that are reasonably general and may be useful for more than one placement algorithm (e.g. anneal and analytic).  Putting those guidelines in the comments will help future developers stick with your organization.

Bill-hbrhbr:
Titan benchmark run001@commit bdaf63c: https://drive.google.com/file/d/1F1U-P6b6LLMa3rCzoPEeeJZjWBgqaiGE/view?usp=sharing

Comment: No runtime or performance degradation.

vaughnbetz:
Great, thanks Bill.  Should this be merged now?

vaughnbetz:
I suggest addressing the comments I have in the review (which aren't too big, pretty much all commenting I think) and then we merge.  After that you can do another PR for more refactoring, but probably best to do this in stages.  What do you think?

Bill-hbrhbr:
Doing this in multiple stages sounds good. I'll just reference this PR in the new PR.

vaughnbetz:
Sounds good.  Let me know when the comments above are addressed and I'll merge this.

vpr/src/place/place.cpp
Bill-hbrhbr: This might become an issue when Dusty designs the parallel execution placer. According to the benchmark results so far everything is good.

vaughnbetz: That's good news, thanks.

Bill-hbrhbr:
Hi @vaughnbetz, I added more comments for the override delay model class. Regarding comments on what should be included in place.cpp, I think maybe it's more suitable for me to do that once the refactorization is mostly done.

QoR test can be skipped since I only changed comments. **This PR can be merged now.** The next step for me is to shorten the argument lists for various core placer routiners.

vaughnbetz:
Great, thanks Bill.

Issue #1527, opened 2020-09-07
Need better commenting of t_physical_tile_type and t_sub_tile


vaughnbetz:
Exactly how physical tiles are described, and what a sub_tile within each is, are core concepts.  Physical tiles were recently generalized to make it possible to place different types of logic blocks at one x,y location, and to allow multiple physical implementations of a logical block (e.g. CLB and MCLB).  However, the commenting of these types and how to use them isn't sufficient (or I'm not finding the big comment block). 

#### Proposed Behaviour
We should improve the commenting of these concepts. The commenting of sub_tile in https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/bb51a8b5b3ed65a8cb2ffebc6aa4b96c30f40630/libs/libarchfpga/src/physical_types.h#L668 should be improved.  The commenting of t_physical_type should similarly be improved around this concept.

t_logical_block should also have a clear distinction between logical and physical blocks.

#### Current Behaviour
I either can't find the data model comments on this, or they don't exist.

#### Possible Solution
We need both high-level data model comments around this, and clear commenting of all the types and data members involved.

acomodi:
@vaughnbetz FYI, opened this PR: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1531.

Please let me know if more details need to be added.

On a side note, there is more explanation on the sub tile and equivalent tiles in the tutorial section of the documentation that might provide more information:
- [Heterogeneous tiles](https://docs.verilogtorouting.org/en/latest/tutorials/arch/heterogeneous_tiles/)
- [Equivalent sites](https://docs.verilogtorouting.org/en/latest/tutorials/arch/equivalent_sites/)

vaughnbetz:
Closed as completed.

PR #1531, opened 2020-09-10
arch: additional comments on physical tiles, logical blocks, sub tiles

acomodi:

#### Description

Additional in-code comments around physical tiles, logical blocks and sub tiles.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1527

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -576,6 +576,23 @@ constexpr int DEFAULT_SWITCH = -2;
  * num_receivers: Total number of input receivers supplied
  * index: Keep track of type in array for easy access
  * logical_tile_index: index of the corresponding logical block type
+ *
+ * In general, the physical tile is a placeable physical resource on the FPGA device,
+ * and it is allowed to contain an heterogeneous set of logical blocks (pb_types).
+ *
+ * Each physical tile must specify at least one sub tile, that is a physical location
+ * on the z axis. This means that a physical tile occupies an (x, y) location on the grid,
+ * and it has at least one sub tile slot that allows for a placement within the (x, y) location.
+ *
+ * Therefore, to identify the location of a logical block within the device grid, we need to
+ * specify three different coordinates:
+ *      - x: horizontal coordinate
+ *      - y: vertical coordinate
+ *      - z: depth coordinate, which refers to the sub tile instance that is used within the (x, y) location
+ *
+ * A physical tile is heterogeneous as it allows the placement of different kinds of logical blocks within,
+ * that can share the same (x, y) placement location.
+ *
  */
 struct t_physical_tile_type {
     char* name = nullptr;
@@ -615,6 +632,7 @@ struct t_physical_tile_type {
 
     int index = -1; /* index of type descriptor in array (allows for index referencing) */
 
+    // vector of the different types of sub tiles allowed for the physical tile.
     std::vector<t_sub_tile> sub_tiles;
 
     /* Unordered map indexed by the logical block index.
@@ -624,6 +642,7 @@ struct t_physical_tile_type {
     /* Returns the indices of pins that contain a clock for this physical logic block */
     std::vector<int> get_clock_pins_indices() const;
 
+    // Returns the sub tile location (z coordinate) given an input pin of the physical tile
     int get_sub_tile_loc_from_pin(int pin_num) const;
 
     // TODO: Remove is_input_type / is_output_type as part of
@@ -662,8 +681,25 @@ struct t_capacity_range {
     }
 };
 
-/** Describes the possible placeable blocks within a physical tile type.
- *  A sub tile adds flexibility in the tile composition description.
+/**
+ * @brief Describes the possible placeable blocks within a physical tile type.
+ *
+ * Heterogeneous blocks:
+ *
+ * The sub tile allows to have heterogeneous blocks placed at the same grid location.
+ * Heterogeneous blocks are blocks which do not share neither the same functionality nor the
+ * IO interface, but do share the same (x, y) grid location.
+ * For each heterogeneous block than, there should be a corresponding sub tile to enable
+ * its placement within the physical tile.
+ *
+ *
+ * Equivalent sites:
+ *
+ * Moreover, the same sub tile enables to allow the placement of different implementations
+ * of a logical block.
+ * This means that two blocks that have different internal functionalities, but the IO interface of one block
+ * is a subset of the other, they can be placed at the same sub tile location within the physical tile.
+ * These two blocks can be identified as equivalent, hence they can belong to the same sub tile.
  */
 struct t_sub_tile {
     char* name = nullptr;
@@ -763,6 +799,16 @@ struct t_physical_tile_port {
  *
  * index: Keep track of type in array for easy access
  * physical_tile_index: index of the corresponding physical tile type
+ *
+ * A logical block is the implementation of a component'ss functionality of the FPGA device
+ * and it identifies its logical behaviour and internal connections.
+ *
+ * The logical block type is mainly used during the packing stage of VPR and is used to generate
+ * the packed netlist and all the corresponding blocks and their internal structure.
+ *
+ * The logical blocks than get assigned to a possible physical tile for the placement step.
+ *
+ * A logical block must correspond to at least one physical tile.
  */
 struct t_logical_block_type {
     char* name = nullptr;
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -581,14 +581,14 @@ constexpr int DEFAULT_SWITCH = -2;
  * and it is allowed to contain an heterogeneous set of logical blocks (pb_types).
  *
  * Each physical tile must specify at least one sub tile, that is a physical location
- * on the z axis. This means that a physical tile occupies an (x, y) location on the grid,
+ * on the sub tiles stacks. This means that a physical tile occupies an (x, y) location on the grid,
  * and it has at least one sub tile slot that allows for a placement within the (x, y) location.
  *
  * Therefore, to identify the location of a logical block within the device grid, we need to
  * specify three different coordinates:
- *      - x: horizontal coordinate
- *      - y: vertical coordinate
- *      - z: depth coordinate, which refers to the sub tile instance that is used within the (x, y) location
+ *      - x         : horizontal coordinate
+ *      - y         : vertical coordinate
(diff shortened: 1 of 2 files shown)


libs/libarchfpga/src/physical_types.h
```diff
@@ -576,6 +576,23 @@ constexpr int DEFAULT_SWITCH = -2;
  * num_receivers: Total number of input receivers supplied
  * index: Keep track of type in array for easy access
  * logical_tile_index: index of the corresponding logical block type
+ *
+ * In general, the physical tile is a placeable physical resource on the FPGA device,
+ * and it is allowed to contain an heterogeneous set of logical blocks (pb_types).
+ *
+ * Each physical tile must specify at least one sub tile, that is a physical location
+ * on the z axis. This means that a physical tile occupies an (x, y) location on the grid,
```
vaughnbetz, line 584: Lets not use z for the third axis (we may have 3D FPGAs in the future) and instead use terminology the tutorial uses (index in the sub_block stack).

libs/libarchfpga/src/physical_types.h
```diff
@@ -763,6 +799,16 @@ struct t_physical_tile_port {
  *
  * index: Keep track of type in array for easy access
  * physical_tile_index: index of the corresponding physical tile type
+ *
+ * A logical block is the implementation of a component'ss functionality of the FPGA device
```
vaughnbetz, line 803: typo --> double 'ss

libs/libarchfpga/src/physical_types.h
```diff
@@ -662,8 +681,25 @@ struct t_capacity_range {
...
 
-/** Describes the possible placeable blocks within a physical tile type.
- *  A sub tile adds flexibility in the tile composition description.
+/**
+ * @brief Describes the possible placeable blocks within a physical tile type.
+ *
+ * Heterogeneous blocks:
+ *
+ * The sub tile allows to have heterogeneous blocks placed at the same grid location.
+ * Heterogeneous blocks are blocks which do not share neither the same functionality nor the
```
vaughnbetz, line 690: typo: neither -> either
nor -> or
(otherwise double negatives)

libs/libarchfpga/src/physical_types.h
```diff
@@ -576,6 +576,23 @@ constexpr int DEFAULT_SWITCH = -2;
...
+ *
+ * Each physical tile must specify at least one sub tile, that is a physical location
+ * on the z axis. This means that a physical tile occupies an (x, y) location on the grid,
+ * and it has at least one sub tile slot that allows for a placement within the (x, y) location.
+ *
+ * Therefore, to identify the location of a logical block within the device grid, we need to
+ * specify three different coordinates:
+ *      - x: horizontal coordinate
+ *      - y: vertical coordinate
+ *      - z: depth coordinate, which refers to the sub tile instance that is used within the (x, y) location
```
vaughnbetz, line 591: rename to sub_tile index, within the sub_tile stack.

libs/libarchfpga/src/physical_types.h
```diff
@@ -763,6 +799,16 @@ struct t_physical_tile_port {
...
+ *
+ * The logical block type is mainly used during the packing stage of VPR and is used to generate
+ * the packed netlist and all the corresponding blocks and their internal structure.
+ *
+ * The logical blocks than get assigned to a possible physical tile for the placement step.
+ *
+ * A logical block must correspond to at least one physical tile.
  */
 struct t_logical_block_type {
     char* name = nullptr;
```
vaughnbetz, line 814: explain equivalent_tiles (below):  comment that it is a lit of the physical tiles at which one could place this type of netlist block?

libs/libarchfpga/src/physical_types.h
```diff
@@ -662,8 +681,25 @@ struct t_capacity_range {
...
+ * Equivalent sites:
+ *
+ * Moreover, the same sub tile enables to allow the placement of different implementations
+ * of a logical block.
+ * This means that two blocks that have different internal functionalities, but the IO interface of one block
+ * is a subset of the other, they can be placed at the same sub tile location within the physical tile.
+ * These two blocks can be identified as equivalent, hence they can belong to the same sub tile.
  */
 struct t_sub_tile {
     char* name = nullptr;
```
vaughnbetz, line 705: explain equivalent_sites (below).  These are really the types of logical blocks (i.e. blocks types in the netlist) which could be placed in this type of subtile (I think).

vaughnbetz: Explain capacity below.  Give how many of these sub_tiles there are?  (i.e. capacity_range of 4 .. 7 means we could place compatible (logical) netlist blocks at sub_tile indices 4 to 7 in the grid?

libs/libarchfpga/src/physical_types.h
```diff
@@ -662,8 +681,25 @@ struct t_capacity_range {
     }
 };
 
-/** Describes the possible placeable blocks within a physical tile type.
- *  A sub tile adds flexibility in the tile composition description.
+/**
+ * @brief Describes the possible placeable blocks within a physical tile type.
+ *
+ * Heterogeneous blocks:
```
vaughnbetz, line 687: Add a pointer to the heterogeneous tiles tutorial on the web (can just say look for it rather than giving a url), as that tutorial does help a lot in figuring this out.

vpr/src/util/vpr_utils.h
```diff
@@ -155,7 +155,23 @@ void print_usage_by_wire_length();
...
+ * to syncronize the pins related to the logical block with the actual connection interface of
+ * the belonging physical tile with the RR graph.
+ *
+ * This step is required as the logical block can be placed at any compatible sub tile locations
+ * within a physical tile.
+ * Given that it is possible to have equivalent logical blocks within a specific sub tile, with
+ * a different subset of IO pins, the various pins offsets must be correctly computed and assigned
+ * to the physical pins vector, so that, when the net RR terminals are computed, the correct physical
+ * tile IO pins are selected.
+ */
```
vaughnbetz, line 172: Does it get the location of block iblk from the netlist (I assume it does)?  I assume it needs the block to be placed at a legal (x,y,isub) but it is OK if there is some overuse of the grid (i.e. it does not need the grid_blocks data to be up to date) -- is that right?
This is relevant for things like analytic placers which produce illegal placements in multiple ways (block.x and .y at a location of the wrong type, overused grid_blocks).  I am guessing the block.x, block.y at the wrong type of tile/subtile would be fatal, but it is OK for grid_blocks to not be up to date.

vaughnbetz:
Thanks for the fast turnaround and enhanced comments @acomodi .
I have some additional questions and commenting suggestions for you to consider.

libs/libarchfpga/src/physical_types.h
```diff
@@ -624,6 +642,7 @@ struct t_physical_tile_type {
     /* Returns the indices of pins that contain a clock for this physical logic block */
     std::vector<int> get_clock_pins_indices() const;
 
+    // Returns the sub tile location (z coordinate) given an input pin of the physical tile
```
MohamedElgammal, line 645: We no longer have the z coordinate notation. We can change the comment to be:
// Returns the sub tile location of the physical tile given an input pin

libs/libarchfpga/src/physical_types.h
```diff
@@ -662,8 +681,25 @@ struct t_capacity_range {
...
- *  A sub tile adds flexibility in the tile composition description.
+/**
+ * @brief Describes the possible placeable blocks within a physical tile type.
+ *
+ * Heterogeneous blocks:
+ *
+ * The sub tile allows to have heterogeneous blocks placed at the same grid location.
+ * Heterogeneous blocks are blocks which do not share neither the same functionality nor the
+ * IO interface, but do share the same (x, y) grid location.
+ * For each heterogeneous block than, there should be a corresponding sub tile to enable
```
MohamedElgammal, line 692: typo: heterogeneous block type

vpr/src/util/vpr_utils.h
acomodi: So, this routine does not check for illegal placement, in the sense that it just looks for the block and, given it's placement location derives the value of the physical pins to make it match with the pins of the logical block.

It does not strictly check whether the block is compatible with the sub tile and it does not require that a physical location gets overused.

If I remember correctly, all the checks are done prior to this step, so this routine assumes that placement is legal.

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

libs/libarchfpga/src/physical_types.h
acomodi: Done

vpr/src/util/vpr_utils.h
vaughnbetz: It's worth writing this in; the routine uses the .x,.y and .sub_tile of the clb netlist, and expects those to put each netlist block at a legal location that can accomodate it. It does not check for overuse of locations, so it can be used with placements that have resource overuse (e.g. global placement algorithms).

vaughnbetz:
Thanks; one more suggested comment to add above. After that this is good to go.

vpr/src/util/vpr_utils.h
acomodi: Sure, comment added.

vaughnbetz:
Thanks.  Will merge as soon as CI goes green (or you can).

acomodi:
CI went green on all targets, merging.

PR #1535, opened 2020-09-11
Created draft of new constraints class header file

sfkhalid:
A draft header file containing an overview of the new constraints class was created. The constraints class will read in constraints information from an XML file. This information will be used during the packing and placement stages of VPR

#### Motivation and Context
This class is being created to address issue #932 Add Placement Constraints to VPR

### DIFF
--- a/libs/libvtrutil/src/vtr_geometry.h
+++ b/libs/libvtrutil/src/vtr_geometry.h
@@ -141,7 +141,7 @@ class Rect {
 template<class T>
 Rect<T> bounding_box(const Rect<T>& lhs, const Rect<T>& rhs);
 
-//Return the intersection (union) of two given rectangles
+//Return the intersection of two given rectangles
 template<class T>
 Rect<T> intersection(const Rect<T>& lhs, const Rect<T>& rhs);
 
--- a/libs/libvtrutil/src/vtr_geometry.h
+++ b/libs/libvtrutil/src/vtr_geometry.h
@@ -141,6 +141,10 @@ class Rect {
 template<class T>
 Rect<T> bounding_box(const Rect<T>& lhs, const Rect<T>& rhs);
 
+//Return the intersection (union) of two given rectangles
+template<class T>
+Rect<T> intersection(const Rect<T>& lhs, const Rect<T>& rhs);
+
 //Sample on a uniformly spaced grid within a rectangle
 //  sample(vtr::Rect(l, h), 0, 0, M) == l
 //  sample(vtr::Rect(l, h), M, M, M) == h
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -34,7 +34,7 @@ class Partition {
     bool contains_atom(AtomBlockId atom_id);
 
     //get the atom_blocks vector
-    std::vector<AtomBlockId> get_atoms();
+    const std::vector<AtomBlockId> get_atoms();
 
     //set the union of regions for this partition;
     void set_part_regions(PartitionRegions pr);
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -7,32 +7,41 @@
 #include "partition_region.h"
 /**
  * @file
- * @brief This file defines the data for a partition: a grouping of atoms that are constrained to a portion of an FPGA. A partition defines
- * the atoms that are assigned to it, and the locations in which they can be placed. The locations are a union of rectangles in x, y,
- * sub_tile space. Common cases include a single rectangle or a single x, y, sub_tile (specific location). More complex partitions
+ * @brief This file defines the data for a partition: a grouping of atoms that are constrained to a portion of an FPGA.
+ *
+ * A partition defines the atoms that are assigned to it, and the locations in which they can be placed.
+ * The locations are a union of rectangles in x, y, sub_tile space.
+ * Common cases include a single rectangle or a single x, y, sub_tile (specific location). More complex partitions
  * with L, T or other shapes can be created with a union of multiple rectangles.
  */
 
-//Type tag for PartitionId
+/// @brief Type tag for PartitionId
 struct partition_id_tag;
 
-//A unique identifier for a partition
+/// @brief A unique identifier for a partition
 typedef vtr::StrongId<partition_id_tag> PartitionId;
 
 class Partition {
   public:
     const std::string get_name();
+
     void set_name(std::string _part_name);
 
-    //set the union of regions for this partition;
-    void set_part_regions(PartitionRegion pr);
+    /**
+     * @brief Set the PartitionRegion (union of rectangular regions) for this partition
+     *
+     *   @param pr     The PartitionRegion belonging to the partition
+     */
+    void set_part_region(PartitionRegion pr);
 
-    //get the union of regions of this partition
-    const PartitionRegion get_part_regions();
+    /**
+     * @brief Get the PartitionRegion (union of rectangular regions) for this partition
+     */
+    const PartitionRegion get_part_region();
 
   private:
-    std::string name;             //name of the partition, name will be unique across partitions
-    PartitionRegion part_regions; //the union of regions that the partition can be placed in
+    std::string name;            ///< name of the partition, name will be unique across partitions
+    PartitionRegion part_region; ///< the union of regions that the partition can be placed in
 };
 
 #endif /* PARTITION_H */
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -21,10 +21,12 @@ typedef vtr::StrongId<partition_id_tag> PartitionId;
 
 class Partition {
   public:
-    PartitionId get_partition_id();
+    Partition();
+
+    const PartitionId get_partition_id();
     void set_partition_id(PartitionId _part_id);
 
-    std::string get_name();
+    const std::string get_name();
     void set_name(std::string _part_name);
 
     //add an id to the atom_blocks vector
@@ -40,7 +42,7 @@ class Partition {
     void set_part_regions(PartitionRegions pr);
 
     //get the union of regions of this partition
-    PartitionRegions get_part_regions();
+    const PartitionRegions get_part_regions();
 
   private:
     PartitionId id;                       //unique id for this partition
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -3,8 +3,8 @@
 
 #include "vtr_strong_id.h"
 #include "region.h"
-#include "partition_regions.h"
 #include "atom_netlist_fwd.h"
+#include "partition_region.h"
 /**
  * @file
  * @brief This file defines the data for a partition: a grouping of atoms that are constrained to a portion of an FPGA. A partition defines
@@ -21,34 +21,18 @@ typedef vtr::StrongId<partition_id_tag> PartitionId;
 
 class Partition {
   public:
-    Partition();
-
-    const PartitionId get_partition_id();
-    void set_partition_id(PartitionId _part_id);
-
     const std::string get_name();
     void set_name(std::string _part_name);
 
-    //add an id to the atom_blocks vector
-    void add_to_atoms(AtomBlockId atom_id);
-
-    //check if a given atom is in the partition
-    bool contains_atom(AtomBlockId atom_id);
-
-    //get the atom_blocks vector
-    const std::vector<AtomBlockId> get_atoms();
-
     //set the union of regions for this partition;
-    void set_part_regions(PartitionRegions pr);
+    void set_part_regions(PartitionRegion pr);
 
     //get the union of regions of this partition
-    const PartitionRegions get_part_regions();
+    const PartitionRegion get_part_regions();
 
   private:
-    PartitionId id;                       //unique id for this partition
-    std::string name;                     //name of the partition, name will be unique across partitions
-    std::vector<AtomBlockId> atom_blocks; //atoms that belong to this partition, each atom should only belong to one partition
-    PartitionRegions part_regions;        //the union of regions that the partition can be placed in
+    std::string name;             //name of the partition, name will be unique across partitions
+    PartitionRegion part_regions; //the union of regions that the partition can be placed in
 };
 
 #endif /* PARTITION_H */
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -1,0 +1,27 @@
+#ifndef PARTITION_H
+#define PARTITION_H
+
+#include "vtr_strong_id.h"
+#include "region.h"
+
+/**
+ * This class stores the data for each constraint partition -
+ * partition name, the atoms that are in the partition,
+ * and the regions that the partition can be placed in.
+ */
+
+//Type tag for PartitionId
+struct partition_id_tag;
+
+//A unique identifier for a partition
+typedef vtr::StrongId<partition_id_tag> PartitionId;
+
+class Partition {
+  public: //accessors
+  private:
+    PartitionId id;
+    std::string name;
+    std::vector<AtomBlockId> atomBlocks; //atoms that belong to this partition
+    std::vector<Region> regions;         //regions that this partition can be placed in
+};
+#endif /* PARTITION_H */
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -3,11 +3,14 @@
 
 #include "vtr_strong_id.h"
 #include "region.h"
+#include "partition_regions.h"
 
 /**
- * This class stores the data for each constraint partition -
- * partition name, the atoms that are in the partition,
- * and the regions that the partition can be placed in.
+ * @file
+ * @brief This file defines the data for a partition: a grouping of atoms that are constrained to a portion of an FPGA. A partition defines
+ * the atoms that are assigned to it, and the locations in which they can be placed. The locations are a union of rectangles in x, y,
+ * sub_tile space. Common cases include a single rectangle or a single x, y, sub_tile (specific location). More complex partitions
+ * with L, T or other shapes can be created with a union of multiple rectangles.
  */
 
 //Type tag for PartitionId
@@ -17,11 +20,28 @@ struct partition_id_tag;
 typedef vtr::StrongId<partition_id_tag> PartitionId;
 
 class Partition {
-  public: //accessors
+  public:
+    PartitionId get_partition_id();
+    void set_partiton_id(PartitionId _part_id);
+
+    std::string get_name();
+    void set_name(std::string _part_name);
+
+    //append to the atom_blocks vector
+    void add_to_atom_blocks(AtomBlockId id);
+
+    //check if a given atom is in the partition
+    bool is_atom_in_part(AtomBlockId id);
+
+    std::vector<AtomBlockId> get_partition_atoms();
+
+    PartitionRegions get_partition_regions();
+
   private:
     PartitionId id;
-    std::string name;
-    std::vector<AtomBlockId> atomBlocks; //atoms that belong to this partition
-    std::vector<Region> regions;         //regions that this partition can be placed in
+    std::string name;                     //name of the partition
+    std::vector<AtomBlockId> atom_blocks; //atoms that belong to this partition
+    PartitionRegions part_regions;        //the union of regions that the partition can be placed in
 };
+
 #endif /* PARTITION_H */
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -4,7 +4,7 @@
 #include "vtr_strong_id.h"
 #include "region.h"
 #include "partition_regions.h"
-
+#include "atom_netlist_fwd.h"
 /**
  * @file
  * @brief This file defines the data for a partition: a grouping of atoms that are constrained to a portion of an FPGA. A partition defines
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -21,26 +21,33 @@ typedef vtr::StrongId<partition_id_tag> PartitionId;
 
 class Partition {
   public:
+
     PartitionId get_partition_id();
-    void set_partiton_id(PartitionId _part_id);
+    void set_partition_id(PartitionId _part_id);
 
     std::string get_name();
     void set_name(std::string _part_name);
 
-    //append to the atom_blocks vector
-    void add_to_atom_blocks(AtomBlockId id);
+    //add an id to the atom_blocks vector
+    void add_to_atoms(AtomBlockId atom_id);
 
     //check if a given atom is in the partition
-    bool is_atom_in_part(AtomBlockId id);
+    bool contains_atom(AtomBlockId atom_id);
 
-    std::vector<AtomBlockId> get_partition_atoms();
+    //get the atom_blocks vector
+    std::vector<AtomBlockId> get_atoms();
 
-    PartitionRegions get_partition_regions();
+    //set the union of regions for this partition;
+    void set_part_regions(PartitionRegions pr);
+
+    //get the union of regions of this partition
+    PartitionRegions get_part_regions();
+
 
   private:
-    PartitionId id;
-    std::string name;                     //name of the partition
-    std::vector<AtomBlockId> atom_blocks; //atoms that belong to this partition
+    PartitionId id;                       //unique id for this partition
+    std::string name;                     //name of the partition, name will be unique across partitions
+    std::vector<AtomBlockId> atom_blocks; //atoms that belong to this partition, each atom should only belong to one partition
     PartitionRegions part_regions;        //the union of regions that the partition can be placed in
 };
 
--- a/vpr/src/base/partition.h
+++ b/vpr/src/base/partition.h
@@ -21,7 +21,6 @@ typedef vtr::StrongId<partition_id_tag> PartitionId;
 
 class Partition {
   public:
-
     PartitionId get_partition_id();
     void set_partition_id(PartitionId _part_id);
 
@@ -42,7 +41,6 @@ class Partition {
 
     //get the union of regions of this partition
     PartitionRegions get_part_regions();
-
 
   private:
     PartitionId id;                       //unique id for this partition
--- a/vpr/src/base/partition_region.h
+++ b/vpr/src/base/partition_region.h
@@ -10,10 +10,10 @@
  * of regions that a partition can be placed in.
  */
 
-class PartitionRegions {
+class PartitionRegion {
   public:
     //Returns the intersection of two PartitionRegions vectors that are passed to it.
-    PartitionRegions get_intersection(PartitionRegions part_region);
+    PartitionRegion get_intersection(PartitionRegion part_region);
 
     //function to add to the partition_regions vector
     void add_to_part_regions(Region region);
--- a/vpr/src/base/partition_region.h
+++ b/vpr/src/base/partition_region.h
@@ -12,17 +12,27 @@
 
 class PartitionRegion {
   public:
-    //Returns the intersection of two PartitionRegions vectors that are passed to it.
+    /**
+     * @brief Return the intersection of two PartitionRegions
+     *
+     *   @param part_region     The PartitionRegion that the calling object will be intersected with
+     */
     PartitionRegion get_intersection(PartitionRegion part_region);
 
-    //function to add to the partition_regions vector
-    void add_to_part_regions(Region region);
+    /**
+     * @brief Add a region to the union of regions belonging to the partition
+     *
+     *   @param region     The region to be added to the calling object
+     */
+    void add_to_part_region(Region region);
 
-    //function to get the partition_regions vector
-    std::vector<Region> get_partition_regions();
+    /**
+     * @brief Return the union of regions
+     */
+    std::vector<Region> get_partition_region();
 
   private:
-    std::vector<Region> partition_regions; //vector of regions that a partition can be placed in
+    std::vector<Region> partition_region; ///< union of rectangular regions that a partition can be placed in
 };
 
 #endif /* PARTITION_REGIONS_H */
--- a/vpr/src/base/partition_regions.h
+++ b/vpr/src/base/partition_regions.h
@@ -1,33 +1,36 @@
 #ifndef PARTITION_REGIONS_H
 #define PARTITION_REGIONS_H
 
+#include "region.h"
+
 /**
  * @file
  * @brief This file defines the PartitionRegions class. The PartitionRegions class is used to store the union
- * of regions that a particular partition can be placed in.
+ * of regions that a partition can be placed in.
  */
 
 class PartitionRegions {
   public:
+
     /**
      * Returns the intersection of two PartitionRegions vectors that are passed to it.
      */
     PartitionRegions get_intersection(PartitionRegions region);
 
     /**
-     * A test for intersection
-     * For example, test if the PartitionRegions of an atom intersect with PartitionRegions of a cluster.
+     * A test for intersection of the PartitionRegions of two blocks
      */
     bool intersects(PartitionRegions region);
 
     /**
-     * Takes Atom Id and PartitonRegions vector and returns the intersection of the atom's PartitionRegions
-     * and the net PartitionRegions of the cluster (if any exists)
+     * Takes Atom Id returns intersection of atom's PartitionRegions with the PartitionRegions of the object
      */
-    PartitionRegions get_intersection(const AtomBlockId& id, PartitionRegions part_regions);
+    PartitionRegions get_intersect(const AtomBlockId atom_id);
 
+    //method to add to the partition_regions vector
     void add_to_part_regions(Region region);
 
+    //method to get the partition_regions vector
     std::vector<Region> get_partition_regions();
 
   private:
--- a/vpr/src/base/partition_regions.h
+++ b/vpr/src/base/partition_regions.h
@@ -11,7 +11,6 @@
 
 class PartitionRegions {
   public:
-
     /**
      * Returns the intersection of two PartitionRegions vectors that are passed to it.
      */
--- a/vpr/src/base/partition_regions.h
+++ b/vpr/src/base/partition_regions.h
@@ -2,6 +2,7 @@
 #define PARTITION_REGIONS_H
 
 #include "region.h"
+#include "atom_netlist_fwd.h"
 
 /**
  * @file
@@ -15,16 +16,6 @@ class PartitionRegions {
      * Returns the intersection of two PartitionRegions vectors that are passed to it.
      */
     PartitionRegions get_intersection(PartitionRegions region);
-
-    /**
-     * A test for intersection of the PartitionRegions of two blocks
-     */
-    bool intersects(PartitionRegions region);
-
-    /**
-     * Takes Atom Id returns intersection of atom's PartitionRegions with the PartitionRegions of the object
-     */
-    PartitionRegions get_intersect(const AtomBlockId atom_id);
 
     //method to add to the partition_regions vector
     void add_to_part_regions(Region region);
--- a/vpr/src/base/partition_regions.h
+++ b/vpr/src/base/partition_regions.h
@@ -15,7 +15,7 @@ class PartitionRegions {
     /**
      * Returns the intersection of two PartitionRegions vectors that are passed to it.
      */
-    PartitionRegions get_intersection(PartitionRegions region);
+    PartitionRegions get_intersection(PartitionRegions part_region);
 
     //method to add to the partition_regions vector
     void add_to_part_regions(Region region);
--- a/vpr/src/base/partition_regions.h
+++ b/vpr/src/base/partition_regions.h
@@ -12,15 +12,13 @@
 
 class PartitionRegions {
   public:
-    /**
-     * Returns the intersection of two PartitionRegions vectors that are passed to it.
-     */
+    //Returns the intersection of two PartitionRegions vectors that are passed to it.
     PartitionRegions get_intersection(PartitionRegions part_region);
 
-    //method to add to the partition_regions vector
+    //function to add to the partition_regions vector
     void add_to_part_regions(Region region);
 
-    //method to get the partition_regions vector
+    //function to get the partition_regions vector
     std::vector<Region> get_partition_regions();
 
   private:
--- a/vpr/src/base/partition_regions.h
+++ b/vpr/src/base/partition_regions.h
@@ -1,0 +1,15 @@
+#ifndef PARTITION_REGIONS_H
+#define PARTITION_REGIONS_H
+
+/**
+ * Vectors of type PartitionRegions are used to return information on the intersection of regions
+ * (ex. between a cluster and an atom).
+ */
+
+class PartitionRegions {
+  public: //accessors
+  private:
+    std::vector<Region> partition_regions;
+};
+
+#endif /* PARTITION_REGIONS_H */
--- a/vpr/src/base/partition_regions.h
+++ b/vpr/src/base/partition_regions.h
@@ -2,14 +2,36 @@
 #define PARTITION_REGIONS_H
 
 /**
- * Vectors of type PartitionRegions are used to return information on the intersection of regions
- * (ex. between a cluster and an atom).
+ * @file
+ * @brief This file defines the PartitionRegions class. The PartitionRegions class is used to store the union
+ * of regions that a particular partition can be placed in.
  */
 
 class PartitionRegions {
-  public: //accessors
+  public:
+    /**
+     * Returns the intersection of two PartitionRegions vectors that are passed to it.
+     */
+    PartitionRegions get_intersection(PartitionRegions region);
+
+    /**
+     * A test for intersection
+     * For example, test if the PartitionRegions of an atom intersect with PartitionRegions of a cluster.
+     */
+    bool intersects(PartitionRegions region);
+
+    /**
+     * Takes Atom Id and PartitonRegions vector and returns the intersection of the atom's PartitionRegions
+     * and the net PartitionRegions of the cluster (if any exists)
+     */
+    PartitionRegions get_intersection(const AtomBlockId& id, PartitionRegions part_regions);
+
+    void add_to_part_regions(Region region);
+
+    std::vector<Region> get_partition_regions();
+
   private:
-    std::vector<Region> partition_regions;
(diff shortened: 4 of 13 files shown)


vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,93 @@
...
+
+		/**
+		 * Method to create a new placement region (assign a new id) when a new placement region is declared in XML file
+		 */
+		void create_placement_region();
+
+		/**
+		 * Next two methods are for loading the above data structures
+		 */
+		void load_constrained_primitives(const char* constraints_file);
```
vaughnbetz, line 62: Lower level helper functions for reading the constraint file probably can't be defined until you define the file.  They aren't as important and don't need to be in this design review (or only belong there if you have an outline of how to parse).

vaughnbetz: Also helper functions should be private.

vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,93 @@
...
+		void load_partitions(const char* constraints_file);
+
+		/**
+		 * Higher level function to call above two functions if constraints file is being read
+		 */
+		void read_constraints_file(const char* constraints_file);
+
+		/**
+		 * Method to see if regions of primitives intersect for packing
+		 * Should take in any number of primitives and see if any have intersecting regions
```
vaughnbetz, line 73: Think about how this will be used.  
1. Gradually insert atoms in a cluster.  Keep track of the net PartitionRegions of the cluster (intersection of all primitives' PartitionRegions so far), and intersect with the PartitionRegions or the new atom.  So you need a routine to intersection PartitionRegions and return the new PartitionRegions (can be empty, and you need to have a nice routine to check for that such as .is_empty() . 

2. Maybe want a utility routine that builds on that and takes an AtomBlockId and a PartitionRegions and returns their intersection.

3. Start with data structures private, add the accessors you need.

vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,93 @@
...
+		 * Can check constrained_primitives vector map for this - if vector length is 0 for a primitive, it is not constrained.
+		 */
+		bool is_atom_constrained(const AtomBlockId);
+
+		/**
+		 * Method to check if an atom is found in a given partition
+		 */
+		bool is_atom_in_partition(const AtomBlockId, int);
+
+	private: //Private Members
```
vaughnbetz, line 88: Data should all be private.
Then define the accessors you need (setters/getters).

vaughnbetz: Add minimal public functions; can always add more later but hard to take away.  More public functions is a wider interface which makes updating the class harder.

vaughnbetz (changes requested):

sfkhalid:
Sample XML 1:

<vpr_constraints>
 
    <partition_list>
        
        <partition name="partition_name_or_id">
            
            <add_atoms name_pattern="my_primitive1"/> 
            <add_atoms name_pattern="my_processor/alu32/*/"> 
        </partition>
    </partition_list>

    <placement_constraints>
        
        <placement_region partition_pattern="regex" || atom_pattern="regex"> 
            <rect xlow="int" ylow="int" subtilelow="int" xhigh="int" yhigh="int" subtilehigh="int">
        </placement_region>
    </placement_constraints>

</vpr_constraints>

sfkhalid:
> Sample XML 1:
> 
> <vpr_constraints>
> 
> ```
> <partition_list>
>     
>     <partition name="partition_name_or_id">
>         
>         <add_atoms name_pattern="my_primitive1"/> 
>         <add_atoms name_pattern="my_processor/alu32/*/"> 
>     </partition>
> </partition_list>
> 
> <placement_constraints>
>     
>     <placement_region partition_pattern="regex" || atom_pattern="regex"> 
>         <rect xlow="int" ylow="int" subtilelow="int" xhigh="int" yhigh="int" subtilehigh="int">
>     </placement_region>
> </placement_constraints>
> ```
> 
> </vpr_constraints>

Based on Kevin's draft XML in issue 932 with minor changes

sfkhalid:
Sample XML 2:

<vpr_constraints>
 
    <partition_list>
        
        <partition name="partition_name_or_id">
            
            <add_atoms name_pattern="my_primitive1"/> 
            <add_atoms name_pattern="my_processor/alu32/*/"> 
        </partition>
    </partition_list>
	
	<region_list>
		
		<region name="region_name_or_id">
            <rect xlow="int" ylow="int" subtilelow="int" xhigh="int" yhigh="int" subtilehigh="int">
        </region>
	<region_list>
	
	<placement_constraints>
        
        <constraint> 
            <add_partition partition_pattern="partition_1">
			<add_region region_pattern="region_1">
        </constraint>
    </placement_constraints>
	

</vpr_constraints>

vpr/src/base/partition.h
```diff
@@ -0,0 +1,27 @@
+#ifndef PARTITION_H
+#define PARTITION_H
+
+#include "vtr_strong_id.h"
+#include "region.h"
+
+/**
+ * This class stores the data for each constraint partition -
```
vaughnbetz, line 8: Give more context.  
This file defines the data for a partition: a grouping of atoms that are constrained to some portion of an FPGA. A partition defines the atoms that are assigned to it, and the locations in which they can be placed.  The locations are be a union of rectangles in x,y,sub_tile space. Common cases include a single rectangle or a single x,y,sub_tile (specific location); more complex partitions with L, T or other shapes can be created with a union of multiple rectangles.

vpr/src/base/partition.h
```diff
@@ -0,0 +1,27 @@
...
+//Type tag for PartitionId
+struct partition_id_tag;
+
+//A unique identifier for a partition
+typedef vtr::StrongId<partition_id_tag> PartitionId;
+
+class Partition {
+  public: //accessors
+  private:
+    PartitionId id;
```
vaughnbetz, line 22: Comment every data member.

vpr/src/base/partition.h
```diff
@@ -0,0 +1,27 @@
...
+    std::string name;
+    std::vector<AtomBlockId> atomBlocks; //atoms that belong to this partition
+    std::vector<Region> regions;         //regions that this partition can be placed in
```
vaughnbetz, line 25: if multiple regions, legal to place in any of them (i.e. union).

vpr/src/base/partition_regions.h
```diff
@@ -0,0 +1,15 @@
+#ifndef PARTITION_REGIONS_H
+#define PARTITION_REGIONS_H
+
+/**
+ * Vectors of type PartitionRegions are used to return information on the intersection of regions
```
vaughnbetz, line 5: This comment seems odd; need to describe what this stores (a region, which is a union of rectangles).

vpr/src/base/partition.h
vaughnbetz: Should use PartitionRegions type here -- that's why you made a class.  Decide on best place to comment fact that a PartitionRegions is the union of multiple regions (should be in the PartitionRegions class mostly, can have brief comments outside).

vpr/src/base/region.h
```diff
@@ -0,0 +1,25 @@
+#ifndef REGION_H
+#define REGION_H
+
+#include "vtr_strong_id.h"
+#include "globals.h"
+
+/**
+ * This class stores the data for each constraint region - the region bounds and the subtile bounds.
```
vaughnbetz, line 8: Explain a single rectangle.

vpr/src/base/region.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+//Type tag for RegionId
+struct region_id_tag;
```
vaughnbetz, line 12: Why do I need ids for Regions?

vaughnbetz: Delete for id for Regions --> does not seem useful.

vpr/src/base/region.h
```diff
@@ -0,0 +1,25 @@
...
+
+//A unique identifier for a region
+typedef vtr::StrongId<region_id_tag> RegionId;
```
vaughnbetz, line 15: Delete the RegionID --> not useful

vpr/src/base/region.h
```diff
@@ -0,0 +1,25 @@
...
+
+class Region {
+  public: //accessors
+  private:
+    //may need to include zmin, zmax for future use in 3D FPGA designs
+    vtr::Rect<int> region_bounds; //xmin, xmax, ymin, ymax
```
vaughnbetz, line 21: Are these inclusive (I think they should be).

vpr/src/base/region.h
```diff
@@ -0,0 +1,25 @@
...
+    int subTilemin, subTilemax;
```
vaughnbetz, line 22: Inclusive?  If you don't care about which sub_tile, you can use 0 and Region::HUGE_SUBTILE and all subtiles will be legal.

If you want a specific location, you can specific xmin = xmax = x, ymin=ymax=y, subTilemin = subTilemax = sub_tile.

vaughnbetz: I'd rename subTilemin to sub_tile_min since all other names in this class use underscores.

vaughnbetz: subTileMin is more camelcase.  But other variables use underscores, so I'd just make them all underscore separated for consistency.

vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+
+class VprConstraints {
+  public:
+    /**
+     * Returns the intersection of two PartitionRegions vectors that are passed to it.
+     * Can be used by atom_cluster_intersection.
+     */
+    PartitionRegions do_regions_intersect(PartitionRegions part_regions_old, PartitionRegions part_regions_new);
```
vaughnbetz, line 30: Rename to get_intersection.
Make a member function of PartitionRegions.
part_regions_old and new don't seem like good names.  region1 and region2 would be better.  Once a member function, you just need one (the other region is the member variables of the class instance).

You probably also want a test for intersection.  Could be a member function:
bool intersects (PartitionRegion region)

vaughnbetz: Objects you are not going to change that are passed in are best passed in as const &  (constant references).

vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Takes Atom Id and PartitonRegions vector and returns the intersection of the atom's partition regions
+     * and the net PartitionRegions of the cluster (if any exists)
+     * Used by clusterer during packing.
+     */
+    PartitionRegions atom_cluster_intersection(const AtomBlockId id, PartitionRegions part_regions);
```
vaughnbetz, line 37: Change to member function.  Name should describe what it does, not who you think might call it. Better prototype would be:
PartitionRegions .get_intersection (const AtomBlockID id);

vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,52 @@
...
+
+  private:
+    /**
+     * Store constraints information of each  constrained atom.
+     * Store ID of the partition the atom belongs to/invalid entry if it doesn't belong to any partition.
```
vaughnbetz, line 42: invalid entry --> doesn't exist in unordered_map

vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,52 @@
...
+     */
+    std::unordered_map<AtomBlockId, PartitionId> constrained_atoms;
+
+    /**
+     * Store partitions in a vector
```
vaughnbetz, line 47: Index in the vector is the partition ID.

vaughnbetz: Use vtr::vector <PartitionID, Partition> instead for type checking.

vtr::vector <size_t, Partition> --> exactly same as vector<Partition>

vaughnbetz:
XML: I think XML 2 has too many distinct concepts that are given Ids (don't think individual regions need names/ids).
Think there should be an XML 0 option too where the PartitionRegion is inside the Partition definition.  That would match your current class description.  

These are reasonable options, so showing both in a design review is reasonable.

sfkhalid:
<vpr_constraints>

    <partition_list>
        
        <partition name="partition_name_or_id">
            
            <add_atoms name_pattern="my_primitive1"/> 
            <add_atoms name_pattern="my_processor/alu32/*/"> 
			
			<add_region xlow="int" ylow="int" subtilelow="int" xhigh="int" yhigh="int" subtilehigh="int"/>
        </partition>
    </partition_list>

</vpr_constraints>

XML_0 option where the PartitionRegion is inside the Partition definition.

vaughnbetz:
I would remove the indenting of the regions; they are directly underneath the partition hierarchy, just like add_atoms.

sfkhalid:
XML_0 with indenting fixed
`<vpr_constraints>
 
    <partition_list>
        
        <partition name="partition_name_or_id">
            
            <add_atoms name_pattern="my_primitive1"/> 
            <add_atoms name_pattern="my_processor/alu32/*/"/> 
	        
	        <add_region xlow="int" ylow="int" subtilelow="int" xhigh="int" yhigh="int" subtilehigh="int"/>
        </partition>
    </partition_list>

</vpr_constraints>`

vpr/src/base/partition.h
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * @file
+ * @brief This file defines the data for a partition: a grouping of atoms that are constrained to a portion of an FPGA. A partition defines
+ * the atoms that are assigned to it, and the locations in which they can be placed. The locations are a union of rectangles in x, y,
+ * sub_tile space. Common cases include a single rectangle or a single x, y, sub_tile (specific location). More complex partitions
+ * with L, T or other shapes can be created with a union of multiple rectangles.
+ */
+
+//Type tag for PartitionId
```
sfkhalid, line 16: mention that this is to create a strongid

vpr/src/base/partition.h
```diff
@@ -0,0 +1,47 @@
...
+    void set_name(std::string _part_name);
+
+    //append to the atom_blocks vector
+    void add_to_atom_blocks(AtomBlockId id);
+
+    //check if a given atom is in the partition
+    bool is_atom_in_part(AtomBlockId id);
+
+    std::vector<AtomBlockId> get_partition_atoms();
+
```
sfkhalid, line 37: Add method to add regions, or have constructor do it (probably easier to do first). Since I have set functions, can just have a simple constructor that just sets everything to invalid.

vpr/src/base/partition.h
```diff
@@ -0,0 +1,47 @@
...
+    PartitionId get_partition_id();
+    void set_partiton_id(PartitionId _part_id);
+
+    std::string get_name();
+    void set_name(std::string _part_name);
+
+    //append to the atom_blocks vector
+    void add_to_atom_blocks(AtomBlockId id);
+
+    //check if a given atom is in the partition
```
sfkhalid, line 33: can shorten some names since in partition class. ex. .contains_atom

vpr/src/base/partition.h
```diff
@@ -0,0 +1,47 @@
...
+    bool is_atom_in_part(AtomBlockId id);
+
+    std::vector<AtomBlockId> get_partition_atoms();
+
+    PartitionRegions get_partition_regions();
+
+  private:
+    PartitionId id;
```
sfkhalid, line 41: comment this too - unique id for this partition

vpr/src/base/partition.h
```diff
@@ -0,0 +1,47 @@
...
+    std::string name;                     //name of the partition
```
sfkhalid, line 42: name will be unique across partitions

vpr/src/base/partition.h
```diff
@@ -0,0 +1,47 @@
...
+    std::vector<AtomBlockId> atom_blocks; //atoms that belong to this partition
```
sfkhalid, line 43: atom should only belong to one partition - mention this

vpr/src/base/partition_regions.h
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * A test for intersection
+     * For example, test if the PartitionRegions of an atom intersect with PartitionRegions of a cluster.
+     */
+    bool intersects(PartitionRegions region);
+
+    /**
+     * Takes Atom Id and PartitonRegions vector and returns the intersection of the atom's PartitionRegions
+     * and the net PartitionRegions of the cluster (if any exists)
```
sfkhalid, line 25: Update the comment (don't need to mention cluster)

vpr/src/base/partition_regions.h
```diff
@@ -0,0 +1,37 @@
...
+     */
+    PartitionRegions get_intersection(const AtomBlockId& id, PartitionRegions part_regions);
```
sfkhalid, line 27: don't need the second argument - will only need atomblockid so you can get the part _regions of that, the other will come from the object you're calling for

vpr/src/base/partition_regions.h
```diff
@@ -0,0 +1,37 @@
...
+#define PARTITION_REGIONS_H
+
+/**
+ * @file
+ * @brief This file defines the PartitionRegions class. The PartitionRegions class is used to store the union
+ * of regions that a particular partition can be placed in.
+ */
+
+class PartitionRegions {
+  public:
```
sfkhalid, line 11: add a simple constructor - takes no arguments and sets everything to empty.
comment every member function, or group them together and comment accordingly

vpr/src/base/region.h
```diff
@@ -0,0 +1,31 @@
+#ifndef REGION_H
+#define REGION_H
+
+#include "globals.h"
```
sfkhalid, line 4: get rid of this

vpr/src/base/region.h
```diff
@@ -0,0 +1,31 @@
...
+ * @file
+ * @brief This file defines the Region class. The Region class stores the data for each constraint region.
+ *        This includes the bounds of the region rectangle and the sub_tile bounds. To lock a block down to a specific location,
+ *        when defining the region specify xmin = xmax, ymin = ymax, sub_tile_min = sub_tile_max
+ *
+ */
+
+class Region {
+  public:
+    vtr::Rect get_region_rect();
```
sfkhalid, line 16: comment the member functions, have a simple constructor and comment that too

vpr/src/base/region.h
```diff
@@ -0,0 +1,31 @@
...
+    void set_region_rect(int _xmin, int _xmax, int _ymin, int _ymax);
```
sfkhalid, line 17: Want to have intersect helper functions here that PartitionRegions can call

vpr/src/base/partition.h
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+//Type tag for PartitionId
+struct partition_id_tag;
+
+//A unique identifier for a partition
+typedef vtr::StrongId<partition_id_tag> PartitionId;
+
+class Partition {
+  public:
```
sfkhalid, line 23: have a member function to see whether a block is locked to a specific x, y, subtile location - comment can be called by placer to see if it can be immediately locked down

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,185 @@
...
+    REQUIRE(regions[1].get_xmin() == 1);
+    REQUIRE(regions[1].get_ymin() == 1);
+    REQUIRE(regions[1].get_xmax() == 2);
+    REQUIRE(regions[1].get_ymax() == 2);
+}
+
+TEST_CASE("RegionLocked", "[vpr]") {
+    Region r1;
+
+    r1.set_region_rect(0, 1, 0, 1); //xmin = xmax = 0, ymin = ymax =1
```
vaughnbetz, line 149: Avoid symmetric inputs.

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,185 @@
...
+    r1.set_sub_tile(3);             //set the region to specific x, y, subtile to lock down
+
+    bool is_r1_locked = false;
+
+    is_r1_locked = r1.locked();
+
+    REQUIRE(is_r1_locked == true);
```
vaughnbetz, line 156: Should also test that a region with no subtiles gives .locked == false.

vaughnbetz (changes requested):
Added some requested test cases.
Should also comment what the test cases do (comment above each function; can be brief for obvious ones).  For things like complex regions (L-shaped, etc.) some more commenting helpful.

vpr/src/base/partition.h
vaughnbetz: Done.

vpr/src/base/partition.h
vaughnbetz: Done.

HackerFoo:
Here's an example of how to get a `std::vector` index from a pointer to one of the elements: https://gist.github.com/HackerFoo/3e1ccd5872a1149c54330942a88a595d

sfkhalid:
> @sfkhalid Here's an example of how to get a `std::vector` index from a pointer to one of the elements: https://gist.github.com/HackerFoo/3e1ccd5872a1149c54330942a88a595d

Thanks, I'll take a look at this.

vaughnbetz:
Thanks Dusty.  I guess there's no alternative to pointer arithmetic if we need this.  I was hoping the standard library wrapped a nice function around things like this (like std::distance, but it only works for iterators which won't quite do what we need -- http://www.cplusplus.com/reference/iterator/distance/). 
Sarah, you should also consider if you really need to return a partition id from a partition.  If callers ask for the partition id for an atom, they will get an id that way and then pass that into the a get_partition function by id or some such function.  But they won't need to ask for an id given a partition object; they'll be doing the opposite pattern of access.  So I suggest not coding the get_partition_id function until you see you have callers that need it.

mithro:
@litghost - Could you also take a look at this?

mithro:
@andrewb1999 -- Could you take a look at this and see if it is interesting / helps at all with the work you are doing at UPenn?

vpr/src/base/partition.h
```diff
@@ -0,0 +1,52 @@
...
+    PartitionId get_partition_id();
+    void set_partition_id(PartitionId _part_id);
+
+    std::string get_name();
+    void set_name(std::string _part_name);
+
+    //add an id to the atom_blocks vector
+    void add_to_atoms(AtomBlockId atom_id);
```
litghost, line 31: This feels like it should be `add_atom` or `add_atom_to_partition`.

The description of the method describes the contents of the method, which is not helpful.  If someone wanted to understand the implementation, they can read the code.  The method comment should describe the why more than the how.

Couple example questions that are relevant to this method:
 - What happens if the same atom_id is added twice?  What happens then?
 - What is the implication of many partitions with few atoms versus fewer partitions with many atoms?
 - What is the implication of an atom being added to two Partition?

litghost: Question: Is there a use case for providing a range of atoms, versus one at a time?

vpr/src/base/partition.h
```diff
@@ -0,0 +1,52 @@
...
+
+    //check if a given atom is in the partition
+    bool contains_atom(AtomBlockId atom_id);
+
+    //get the atom_blocks vector
+    std::vector<AtomBlockId> get_atoms();
```
litghost, line 37: This returns a new copy of the `std::vector`?  I assume this not the intention?  Better to have this return either `const std::vector<AtomBlockId> &` or `vtr:array_view<const AtomBlockId>`?

Method should be const?

vpr/src/base/partition.h
```diff
@@ -0,0 +1,52 @@
...
+ * with L, T or other shapes can be created with a union of multiple rectangles.
+ */
+
+//Type tag for PartitionId
+struct partition_id_tag;
+
+//A unique identifier for a partition
+typedef vtr::StrongId<partition_id_tag> PartitionId;
+
+class Partition {
```
litghost, line 22: Many getters on this class are missing their `const` annotations?

vpr/src/base/partition.h
```diff
@@ -0,0 +1,52 @@
...
+    void add_to_atoms(AtomBlockId atom_id);
+
+    //check if a given atom is in the partition
+    bool contains_atom(AtomBlockId atom_id);
+
+    //get the atom_blocks vector
+    std::vector<AtomBlockId> get_atoms();
+
+    //set the union of regions for this partition;
+    void set_part_regions(PartitionRegions pr);
```
litghost, line 40: This is a nit, but I think it is worth consider.  In general, plural methods like `add_atoms` or `set_part_regions` imply an array context.  However in this case, the array context is a scalar object "PartionRegions".  Does it make more sense to rename `PartitionRegions` to `PartitionCollection`, which is a non-plural name, and makes clear it is a scalar that contains a collection?

litghost: To illustrate what I'm talking about, consider `atom_blocks`, which is plural and is an array-like object.  However part_regions is also plural, but is a scalar object.

vpr/src/base/partition_regions.h
```diff
@@ -0,0 +1,39 @@
...
+ * @brief This file defines the PartitionRegions class. The PartitionRegions class is used to store the union
+ * of regions that a partition can be placed in.
+ */
+
+class PartitionRegions {
+  public:
+    /**
+     * Returns the intersection of two PartitionRegions vectors that are passed to it.
+     */
+    PartitionRegions get_intersection(PartitionRegions region);
```
litghost, line 17: Consider making this a static method that consumes two arguments?

vpr/src/base/partition_regions.h
```diff
@@ -0,0 +1,39 @@
...
+    PartitionRegions get_intersect(const AtomBlockId atom_id);
+
+    //method to add to the partition_regions vector
+    void add_to_part_regions(Region region);
+
+    //method to get the partition_regions vector
+    std::vector<Region> get_partition_regions();
+
+  private:
+    std::vector<Region> partition_regions; //vector of regions that a partition can be placed in
```
litghost, line 36: I'm not sure a vector of Region is the right data structure here.  I feel like you want a spatial tree, e.g. :
 - https://en.wikipedia.org/wiki/Binary_space_partitioning
 - https://en.wikipedia.org/wiki/K-d_tree

You probably want to prototype a couple different data structures balancing simplicity of implementation, data size overhead, etc.

At a minimum, I believe that PartitionRegion needs a `t_bb` that encompasses the vector of regions to provide a quick no if the query it outside any of the supplied regions.

vpr/src/base/region.h
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * @file
+ * @brief This file defines the Region class. The Region class stores the data for each constraint region.
+ *        This includes the bounds of the region rectangle and the sub_tile bounds. To lock a block down to a specific location,
+ *        when defining the region specify xmin = xmax, ymin = ymax, sub_tile_min = sub_tile_max
+ *
+ */
+
+//sentinel value for indicating that a subtile had not been specified
+extern int NO_SUBTILE;
```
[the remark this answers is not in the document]
litghost: Also rather than `extern int` you probably want:
```
static constexpr int NO_SUBTILE = ...;
```

litghost (commented):
Comments in

HackerFoo:
@vaughnbetz Yeah, I saw `std::distance` but it seemed more complicated for `O(1)` access. I've used the method I gave in C and it is defined behavior as long as the pointer is into the vector. One drawback is that the elements should be aligned to a power of 2 for the fastest computation of the index, but this is probably not an issue here, and there is a possibility that storing indices could be slower than division anyway.

If you don't want to promote pointer arithmetic scattered across the codebase, this could be wrapped in a `vtr::vector` method e.g. `ssize_t vector<T>::getIndex(const& T)` and return `-1` on a failed bounds check.

vpr/src/base/partition.h
vaughnbetz: Good point. Yes, should be const, and should just return a reference.  Haven't used array_view; may be more appropriate still.

vpr/src/base/partition.h
vaughnbetz: Good points.  Adding an atom a second time doesn't really make sense; we could check and not add it, but that may get slow with a vector.  Could have a validate() method in vpr_constraints.h that checks for conditions like this efficiently (same atom in a partition twice, same atom in two different partitions) and prints a useful error message.

I don't think having an atom in two partitions makes sense.  It will complicate the data structures and the use case isn't very clear to me.  It makes it possible to have wildcards that overlap, e.g.:
*alu* --> partition 1
*alu|foo --> partition 2
That would only make sense if partition 2 is constrained to a region that is a subset of partition 1.  So you could make cases where this kind-of makes sense, but documenting it is going to be pain and it makes more complex code for not much gain.

Instead we could say:
A:  assigning the same atom to two partitions is an error (and then you have to check during parsing)
B:  The last assignment to an atom wins (removes any prior assignment).  That means if you code some relatively loose and some relatively specific constraints, you would order them from loose to specific to take advantage of this behaviour.

Either A or B seems like a reasonable choice.  Sarah, I suggest trying such a constraint in Quartus to see what it does; we don't have to do the same thing but it's useful to know what Quartus chose in such a case.

vpr/src/base/partition_regions.h
vaughnbetz: Common case is that the regions are simple (0 or 1 rectangle will be the dominant cases).  T-shaped and L-shaped will be possible but relatively rare.  So I think a k-d tree etc. is overkill and will be slower and more complex for the common case.

vpr/src/base/partition.h
litghost: I want to be clear, my examples where not about the implementation, but about the documentation and intention behind how to use the method.  If I find the "add_to_atoms" method, and I want to call it, what is valid, what isn't, etc.  Reading the implementation won't tell me answers to these questions, especially if the validation is done on the complete set of data.  It is totally fine in the comment to say, "see this other method that maintains the following invariant, etc".

sfkhalid: Sure, I'll check how Quartus handles this. In the case of choosing option B, would we then specify to users in documentation that their constraints should go from more general to more specific?

vpr/src/base/partition.h
vaughnbetz: You're right; all should have const.

vpr/src/base/partition.h
vaughnbetz: Could rename PartitionRegions as PartitionRegion, and PartitionRegion as RectPlusSubtile (or something like that).  Now PartitionRegion is singular, indicating it stores a single (possibly complex) region, and RectPlusSubtile indicates the basic elements used to describe a (possibly composite) region.

vpr/src/base/partition.h
vaughnbetz: Sarah: yes, we'd document the behavior and why (no matter what is chosen). 
Keith: agreed, this should be documented for the function in the .h file. But your comment also made me realize that the implementation also hasn't really made a choice on this; it's just letting the callers do whatever they want in terms of multiple additions of an atom. So we should both comment what invariants this function expects, and then enforce them either in the underlying add_atom function or in a validate function.

vpr/src/base/partition_regions.h
litghost: Given your statement, should we consider a dynamic dispatch approach here?  Where the simple 0/1 case is implemented without the vector at all, and the more complicated case gets a more complicated object?

vpr/src/base/partition.h
sfkhalid: Right, Region i.e. RectPlusSubtile does not store a composite region - it stores a single rectangular area and optionally a subtile. So changing the name would specify this.

I do think it would be better to rename PartitionRegions to something non-plural since it is a collection of regions belonging to one particular partition. PartitionCollection could work, I could comment that the class stores the collection of regions (RectPlusSubtile's) belonging to the partition.

vpr/src/base/partition_regions.h
vaughnbetz: We could.  I'd leave it simple for now though, and based on future need we can update if we hit cpu bottlenecks, memory footprint etc.  The class interface should hide all that from the callers, so re-coding would be limited.

vpr/src/base/partition.h
vaughnbetz: I think PartitionCollection is also the wrong name.  This is really storing the (possibly complex) region that goes with one partition.  So it really should be called PartitionRegion.  PartitionRegions and PartitionCollection both let part of the implementation (currently a union of simple rectangles) leak through the name, which isn't great.  If we someday change to more complex region representations like those Keith suggests, PartitionRegion still works as a name, but PartitionCollection doesn't work as well.

libs/libvtrutil/src/vtr_geometry.h
```diff
@@ -141,6 +141,10 @@ class Rect {
 template<class T>
 Rect<T> bounding_box(const Rect<T>& lhs, const Rect<T>& rhs);
 
+//Return the intersection (union) of two given rectangles
```
HackerFoo, line 144: This is not a union operation.

vpr/src/base/region.cpp
```diff
@@ -0,0 +1,95 @@
...
+    if (region_bounds.ymin() != region_bounds.ymax()) {
+        return locked;
+    }
+
+    if (sub_tile == NO_SUBTILE) {
+        return locked;
+    }
+
+    return locked = true;
+}
```
HackerFoo, line 91: This is a lot easier to understand:

```c++
return
  region_bounds.xmin() == region_bounds.xmax() &&
  region_bounds.ymin() == region_bounds.ymax() &&
  sub_tile != NO_SUBTILE;
```

The `locked` variable is unnecessary and it is strange to set it on return.

sfkhalid:
@vaughnbetz This PR should be ready to merge now - made changes to the Partition and VprConstraints classes as we discussed, changed the name of PartitionRegions -> PartitionRegion, and made the comments for the classes doxygen compatible.

vpr/src/base/partition_region.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+class PartitionRegion {
+  public:
+    /**
+     * @brief Return the intersection of two PartitionRegions
+     *
+     *   @param part_region     The PartitionRegion that the calling object will be intersected with
+     */
+    PartitionRegion get_intersection(PartitionRegion part_region);
```
vaughnbetz, line 20: Lets follow Keith's suggestion and be more like Rect.  
Move this to the bottom of the header file (outside the class), and make it 
PartitionRegion insersection (PartitionRegion a, PartitionRegion b);

vpr/src/base/region.cpp
```diff
@@ -0,0 +1,81 @@
+#include "region.h"
+
+/// @brief sentinel value for indicating that a subtile has not been specified
+constexpr int NO_SUBTILE = -1;
+
+Region::Region() {
+    sub_tile = NO_SUBTILE;
+
+    //default rect for a region is (-1, -1, -1, -1)
```
vaughnbetz, line 9: Explain why (invalid, to help catch uninitialized use).

vpr/src/base/region.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+int Region::get_sub_tile() {
+    return sub_tile;
+}
+
+void Region::set_sub_tile(int _sub_tile) {
+    sub_tile = _sub_tile;
+}
+
+bool Region::do_regions_intersect(Region region) {
```
vaughnbetz, line 35: Similar to intersect, should probably make this 
bool do_regions_intersection (Region a, Region b);
and it would go outside the class, at the bottom of this header.
If it's accessing private variables, then you'll need to make it a friend function.

vpr/src/base/region.cpp
```diff
@@ -0,0 +1,81 @@
...
+     * the regions do not intersect
+     */
+    if (intersect_rect.empty() || sub_tile != region.get_sub_tile()) {
+        return intersect = false;
+    }
+
+    return intersect;
+}
+
+Region Region::regions_intersection(Region region) {
```
vaughnbetz, line 54: Rename to intersection, and make it a global function at the bottom of the header file (outside class):
Region intersection (Region a, Region b);
// May have to be a friend function of class Region.

vpr/src/base/region.h
```diff
@@ -0,0 +1,61 @@
...
+ * @brief This file defines the Region class. The Region class stores the data for each constraint region.
+ *
+ * This includes the x and y bounds of the region rectangle and its sub_tile. To lock a block down to a specific location,
+ * when defining the region specify xmin = xmax, ymin = ymax, and specify a subtile value.
+ *
+ */
+
+class Region {
+  public:
+    Region();
```
vaughnbetz, line 17: Add @brief on these methods too so they show up in Doxygen (then build Doxygen to the vtr docs and make sure it shows up).

vpr/src/base/vpr_constraints.cpp
```diff
@@ -0,0 +1,45 @@
+#include "vpr_constraints.h"
+
+void VprConstraints::add_constrained_atom(const AtomBlockId blk_id, const PartitionId part_id) {
+    constrained_atoms.insert({blk_id, part_id});
+
+    auto got = constrained_atoms.find(blk_id);
+
+    if (got == constrained_atoms.end()) {
```
vaughnbetz, line 8: Each atom can be in only one partition; if it's already in a partition it will be switched to the new partition instead.

vpr/src/base/vpr_constraints.cpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+PartitionId VprConstraints::get_atom_partition(AtomBlockId blk_id) {
+    PartitionId part_id;
+
+    auto got = constrained_atoms.find(blk_id);
+
+    if (got == constrained_atoms.end()) {
+        return part_id = PartitionId::INVALID();
```
vaughnbetz, line 21: // Not in a partition, i.e. unconstrained.

vpr/src/base/vpr_constraints.h
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * @brief Store the id of a constrained atom with the id of the partition it belongs to
+     *
+     *   @param blk_id      The atom being stored
+     *   @param part_id     The partition the atom is being constrained to
+     */
+    void add_constrained_atom(const AtomBlockId blk_id, const PartitionId part_id);
+
+    /**
+     * @brief Return id of the partition the atom belongs to
```
vaughnbetz, line 34: // If an atom is not in a partition (unconstrained), PartitionId::Invalid() is returned.

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,414 @@
...
+#include "partition.h"
+#include "region.h"
+
+/**
+ * This file contains unit tests that check the functionality of all classes related to vpr constraints. These classes include
+ * VprConstraints, Region, PartitionRegions, and Partition.
+ */
+
+TEST_CASE("Region", "[vpr]") {
+    Region r1;
```
vaughnbetz, line 14: Comment what each test is testing (briefly).

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,414 @@
...
+    bool is_def_empty = false;
+
+    vtr::Rect<int> def_rect = def_region.get_region_rect();
+    is_def_empty = def_rect.empty();
+    REQUIRE(is_def_empty == true);
+    REQUIRE(def_rect.xmin() == -1);
+}
+
+TEST_CASE("PartitionRegion", "[vpr]") {
+    Region r1;
```
vaughnbetz, line 39: Add comment on what this does.

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,414 @@
...
+    vtr::Rect<int> rect;
+    rect = pr_regions[0].get_region_rect();
+    REQUIRE(rect.xmin() == 2);
+    REQUIRE(rect.ymin() == 3);
+    REQUIRE(rect.xmax() == 6);
+    REQUIRE(rect.ymax() == 7);
+}
+
+TEST_CASE("Partition", "[vpr]") {
+    Partition part;
```
vaughnbetz, line 59: Comment what it tests.

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,414 @@
...
+    vtr::Rect<int> rect;
+    rect = regions[0].get_region_rect();
+    REQUIRE(rect.xmin() == 2);
+    REQUIRE(rect.ymin() == 3);
+    REQUIRE(rect.xmax() == 7);
+    REQUIRE(rect.ymax() == 8);
+}
+
+TEST_CASE("VprConstraints", "[vpr]") {
+    PartitionId part_id(0);
```
vaughnbetz, line 86: Comment what it tests.

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,414 @@
...
+    int_pr = pr1.get_intersection(pr2);
+    std::vector<Region> regions = int_pr.get_partition_region();
+
+    vtr::Rect<int> int_rect(0, 0, 1, 1);
+    vtr::Rect<int> int_rect_2(1, 1, 2, 2);
+    REQUIRE(regions[0].get_region_rect() == int_rect);
+    REQUIRE(regions[1].get_region_rect() == int_rect_2);
+}
+
+TEST_CASE("PartRegionIntersect2", "[vpr]") {
```
vaughnbetz, line 214: Comment what it tests.

vpr/test/test_vpr_constraints.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+    REQUIRE(regions.size() == 4);
+    REQUIRE(regions[0].get_region_rect() == int_r1r3);
+    REQUIRE(regions[1].get_region_rect() == int_r1r4);
+    REQUIRE(regions[2].get_region_rect() == int_r2r3);
+    REQUIRE(regions[3].get_region_rect() == int_r2r4);
+}
+
+TEST_CASE("RegionLocked", "[vpr]") {
+    Region r1;
```
vaughnbetz, line 386: Comment what it tests.

vaughnbetz (commented):
Some final clean up / consistency suggestions. You can do those in a new PR.

PR #1536, opened 2020-09-11
Add extended lookahead map followup

acomodi:
#### Description

Followup of https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1449.

This PR is a followup to solve some outstanding comments on the lookahead extended map that require some more work.
To not have to wait a full CI build once again, these changes are applied to this followup PR

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/connection_router.cpp
+++ b/vpr/src/route/connection_router.cpp
@@ -621,7 +621,7 @@ float ConnectionRouter<Heap>::compute_node_cost_using_rcv(const t_conn_cost_para
 
     const t_conn_delay_budget* delay_budget = cost_params.delay_budget;
 
-    std::tie(expected_delay, expected_cong) = router_lookahead_.get_expected_delay_and_cong(to_node, target_node, cost_params, R_upstream);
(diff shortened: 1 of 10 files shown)


vaughnbetz:
@acomidi: let me know when this is ready for review.  Also, it appears to include all commits for the extended lookahead map; can you rebase so only the new commits for this change are visible (or let me know what subset of commits to view)?  Finally, if you can note which remaining issues / comments are addressed by this PR that would be good.  I can see the clamp function has been moved and it looks like more RRNodeIDs are being used, but a list would be helpful.

acomodi:
@vaughnbetz I have applied the required changes to have lookahead methods get as inputs RRNodeIDs.

There is also the assertion addition to the get_median_entry and a missing comment requested in the previous extended lookahead PR.

vaughnbetz:
Thanks for the changes, and they look good.  But there are some regtest failures (a bunch with reorder_rr_nodes, but they're not hte only ones) and they look like crashes, not QoR changes.

acomodi:
@vaughnbetz CI went green on all tests now. I have also added a fix that should solve the coverty scan issue with the seg_count_ member of the cost map.

vaughnbetz:
Yes, the second travis CI seems hung, but looking at the individual tests in it they look like they passed anyway.  Strange.

Issue #1533, opened 2020-09-10
RCV should report a successful routing even if hold isn't completely resolved


dpbaines:
#### Expected Behaviour

If a successful routing is achieved, i.e. there are no overused nodes and all nets have been routed the router should report a successful routing.

#### Current Behaviour

Right now when the RCV router is enabled, there is an exit condition that checks to see if hTNS is resolved before allowing the router to complete. There is a function to tell RCV to stop early and report a successful routing in some circumstances, however if this isn't tripped before the max iteration count is reached the router will report a failed routing. 

This has caused fails on LU8PEEng and LU32PEEng which reach the maximum iteration count and report fails. These report feasible routing, but don't completely resolve hold.

#### Possible Solution

The easiest patch right now would be to have the router report successful routing if the routing is feasible and the maximum iteration count is reached, regardless of hTNS. This would ensure routing doesn't finish before RCV has resolved hold, while also serving as a fallback if RCV can't resolve hold but the circuit is otherwise fine.

A more robust fix would be making the RCV early exit function mentioned above smarter.

#### Steps to Reproduce

1. Run vtr_qor_reg_chain test
2. Observe LU8PEEng and LU32PEEng circuits log files

#### Context

I've been developing RCV this summer, and didn't get around to patching this. The point of this is to make the RCV router less failure prone.

#### Your Environment

* VTR revision used: VTR 8.0
* Operating System and version: Linux-5.8.5-arch1-1 x86_64
* Compiler version: GNU 10.2.0

PR #1540, opened 2020-09-13
Allow router to report successful routing even if RCV fails to resolve hold

dpbaines:
#### Description

Makes it so that RCV will report a successful routing regardless of whether hold has been resolved. This is just a simple check that allows it to report a successful routing if routing is feasible, and the maximum iteration count is reached. This doesn't fix the underlying problem of **why** RCV fails to resolve hold for some circuits (LU8PEEng and LU32PEEng) but it's a quick patch.

I want to figure out the underlying issue and also address that in this PR so I don't want to merge quite yet. 

#### Related Issue

Closes https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1533

#### Motivation and Context

Make RCV more robust and versatile on multiple circuits. 

I tested this on LU32PEEng and LU8PEEng circuits with harsh hold constraints and RCV enabled, and verified that it reports a successful routing even though hold isn't resolved. I don't see a point in running the full qor tests as this isn't a possibly breaking change yet.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -533,7 +533,10 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
         /*
          * Are we finished?
          */
-        if (is_iteration_complete(routing_is_feasible, router_opts, itry, timing_info, rcv_finished_count == 0)) {
+
+        bool can_router_finish = rcv_finished_count == 0 || itry >= router_opts.max_router_iterations;
+
+        if (is_iteration_complete(routing_is_feasible, router_opts, itry, timing_info, can_router_finish)) {
             auto& router_ctx = g_vpr_ctx.routing();
 
             if (is_better_quality_routing(best_routing, best_routing_metrics, wirelength_info, timing_info)) {
--- a/vpr/src/route/route_timing.cpp
+++ b/vpr/src/route/route_timing.cpp
@@ -533,10 +533,7 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
         /*
          * Are we finished?
          */
-
-        bool can_router_finish = rcv_finished_count == 0 || itry >= router_opts.max_router_iterations;
-
-        if (is_iteration_complete(routing_is_feasible, router_opts, itry, timing_info, can_router_finish)) {
+        if (is_iteration_complete(routing_is_feasible, router_opts, itry, timing_info, rcv_finished_count == 0)) {
             auto& router_ctx = g_vpr_ctx.routing();
 
             if (is_better_quality_routing(best_routing, best_routing_metrics, wirelength_info, timing_info)) {
@@ -1990,9 +1987,11 @@ bool is_iteration_complete(bool routing_is_feasible, const t_router_opts& router
     //in addition to routing being legal and the correct budgeting algorithm being set.
 
     if (routing_is_feasible) {
+        bool can_router_finish = rcv_finished || itry >= router_opts.max_router_iterations;
+
         if (router_opts.routing_budgets_algorithm != YOYO) {
             return true;
-        } else if (router_opts.routing_budgets_algorithm == YOYO && (timing_info->hold_worst_negative_slack() == 0 || rcv_finished) && itry != 1) {
+        } else if (router_opts.routing_budgets_algorithm == YOYO && (timing_info->hold_worst_negative_slack() == 0 || can_router_finish) && itry != 1) {
             return true;
         }
     }


dpbaines:
I don't want to merge this yet, as I'm hoping to get around to figuring out why the LU circuits are failing, and fix those.

mithro:
@dpbaines -- Any chance we could fail by default (and instead add a flag which allows things to report success even if there are hold violations)? Designs with hold violations are **broken**. Reporting that PnR is successful when the design is broken violates the contract that a successful exit code means a successfully place and routed design which will **work**.

vaughnbetz:
@mithro I think we should report routing success in this case; the hold violation will still be reported, so reporting routing success does not imply timing success. 

This matches what commercial tools do -- Quartus will report a successful routing but failed hold (or setup) timing constraints if any failed constraints are present. You can still analyze routability, look at the routing, get a timing report etc.  Simply stopping makes it harder to look at the output (don't get a full analysis), and precludes programming a board (useful as hold margins usually have guardband, so you could still test to see if it works on a board. If the failure is small, it usually will work).

vpr/src/route/route_timing.cpp
```diff
@@ -1987,9 +1987,11 @@ bool is_iteration_complete(bool routing_is_feasible, const t_router_opts& router
     //in addition to routing being legal and the correct budgeting algorithm being set.
 
     if (routing_is_feasible) {
```
vaughnbetz, line 1989: I would re-order this if
Test the non-RCV (routing_budgets_algorithm != YOYO) first, so someone who is not thinking of RCV has less to read.  Then the rest of the code.

vpr/src/route/route_timing.cpp
```diff
@@ -1987,9 +1987,11 @@ bool is_iteration_complete(bool routing_is_feasible, const t_router_opts& router
...
+        bool can_router_finish = rcv_finished || itry >= router_opts.max_router_iterations;
+
         if (router_opts.routing_budgets_algorithm != YOYO) {
             return true;
-        } else if (router_opts.routing_budgets_algorithm == YOYO && (timing_info->hold_worst_negative_slack() == 0 || rcv_finished) && itry != 1) {
+        } else if (router_opts.routing_budgets_algorithm == YOYO && (timing_info->hold_worst_negative_slack() == 0 || can_router_finish) && itry != 1) {
```
vaughnbetz, line 1994: Shouldn't the test for hold_worst_negative_slack be >= 0?  Seems a lot safer than a test for precisely 0.

vaughnbetz (changes requested):

vaughnbetz:
I think a cleaner fix is possible.
1. is_iteration_complete is not a very clear name for this function.  The header file also doesn't have a function description.  A better name would be consider_routing_legal ()

2. A larger restructuring (maybe more than you want to do right now) would be:

- replace the call to is_iteration_complete with a call to is_feasible_routing
- make sure the is_better_quality_routing routine checks for hold and store the routing that has the best timing (first priority = hold if hold slack < 0, second priority = setup slack / current code).
- if you can resolve more hold violations with more iteration, then you'd restructure some later code to also check a new hold_not_legal variable (not just how many convergences there are).

I think this would be cleaner and more powerful, but it would take more testing.

3. I think code that says the routing is legal, even if Th is not resolved in the end, should go in no matter what.  No algorithm is perfect and sometimes hold violations are not resolvable without running out of wiring.  So even if you can resolve LU8 and LU32's hold problems, code to say the routing is legal should be added.

vpr/src/route/route_timing.cpp
dpbaines: I would agree, but worst_negative_slack is equal to zero if worst hold slack is positive. Therefore they would equate to the same thing, but I could change it in case that changes in the future.

dpbaines:
@vaughnbetz Your suggestions make sense, I'll try to take a better look at them this week at some point.

vpr/src/route/route_timing.cpp
vaughnbetz: OK, lets do >= for safety and readability though.

dpbaines:
@vaughnbetz I've implemented your suggestions so far. I've eliminated the is_iteration_complete function, and instead modified the is_better_quality_routing function. If RCV is enabled the is_better_quality_routing function will prioritize routing with improved hold slack, and then setup slack. Otherwise it uses the old code path.

It will also indicate that routing was successful, however it won't allow the router to break early unless either hold slack >= 0, or RCV detects it can finish early, as exists in the current code. I haven't had a lot of time to figure out why LU8 and LU32 struggle to resolve, so I wanna keep picking away at it. 

Let me know what you think of the current code. Or if you think is_better_quality should always prioritize hold even in the regular router (which would clean up code). 

I have tested this on a few key circuits and it seems to be working well. For LU8 and LU32 it returns the routing that had the lowest negative hold slack.

vpr/src/route/route_timing.cpp
```diff
@@ -2007,35 +2006,66 @@ bool should_setup_lower_bound_connection_delays(int itry, const t_router_opts& /
...
                                       const RoutingMetrics& best_routing_metrics,
                                       const WirelengthInfo& wirelength_info,
-                                      std::shared_ptr<const SetupHoldTimingInfo> timing_info) {
+                                      std::shared_ptr<const SetupHoldTimingInfo> timing_info,
+                                      const t_router_opts& router_opts) {
     if (best_routing.empty()) {
         return true; //First legal routing
     }
 
     //Rank first based on sWNS, followed by other timing metrics
```
vaughnbetz, line 2015: This comment doesn't seem to match the code below.

vpr/src/route/route_timing.cpp
```diff
@@ -2007,35 +2006,66 @@ bool should_setup_lower_bound_connection_delays(int itry, const t_router_opts& /
...
-        }
+        if (router_opts.routing_budgets_algorithm != YOYO) {
+            // If RCV is disabled prioritize setup slack and don't worry about hold
 
-        if (timing_info->setup_total_negative_slack() > best_routing_metrics.sTNS) {
-            return true;
-        } else if (timing_info->setup_total_negative_slack() < best_routing_metrics.sTNS) {
-            return false;
-        }
+            if (timing_info->hold_worst_negative_slack() > best_routing_metrics.hWNS) {
```
vaughnbetz, line 2020: This is the RCV off case?  If so, why is it making decisions based on hold?

vpr/src/route/route_timing.cpp
```diff
@@ -2007,35 +2006,66 @@ bool should_setup_lower_bound_connection_delays(int itry, const t_router_opts& /
...
+                return true;
+            } else if (timing_info->hold_worst_negative_slack() > best_routing_metrics.hWNS) {
```
vaughnbetz, line 2022: Both the if and else if condition seem to be the same?  Copy and paste error?

vpr/src/route/route_timing.cpp
```diff
@@ -2007,35 +2006,66 @@ bool should_setup_lower_bound_connection_delays(int itry, const t_router_opts& /
...
+                return false;
+            }
 
-        if (timing_info->hold_worst_negative_slack() > best_routing_metrics.hWNS) {
-            return true;
-        } else if (timing_info->hold_worst_negative_slack() > best_routing_metrics.hWNS) {
-            return false;
-        }
+            if (timing_info->hold_total_negative_slack() > best_routing_metrics.hTNS) {
```
vaughnbetz, line 2026: Needs a comment as this looks like a repeat of the prior 5 lines of code until you look very closely (total negative slack instead of worst negative slack).

vpr/src/route/route_timing.cpp
```diff
@@ -2007,35 +2006,66 @@ bool should_setup_lower_bound_connection_delays(int itry, const t_router_opts& /
...
+                return true;
+            } else if (timing_info->hold_total_negative_slack() > best_routing_metrics.hTNS) {
+                return false;
+            }
 
-        if (timing_info->hold_total_negative_slack() > best_routing_metrics.hTNS) {
-            return true;
-        } else if (timing_info->hold_total_negative_slack() > best_routing_metrics.hTNS) {
-            return false;
+            if (timing_info->setup_worst_negative_slack() > best_routing_metrics.sWNS) {
```
vaughnbetz, line 2032: You could reduce code duplication and make the code more readable by making helper functions: setup_comparison and hold_comparison that encapsulate the (1) test worst case slack first and then check total negative slack, and return better, worse or tied.

vaughnbetz: Those routines would also give you a place to comment that we don't care about positive slack (consider it a tie) and only check to see if negative slacks have improved.

vpr/src/route/route_timing.cpp
```diff
@@ -579,11 +588,17 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
...
+        }
+
         //Have we converged the maximum number of times, did not make any changes, or does it seem
         //unlikely additional convergences will improve QoR?
-        if (legal_convergence_count >= router_opts.max_convergence_count
-            || router_iteration_stats.connections_routed == 0
-            || early_reconvergence_exit_heuristic(router_opts, itry_since_last_convergence, timing_info, best_routing_metrics)) {
+        if ((legal_convergence_count >= router_opts.max_convergence_count
+             || router_iteration_stats.connections_routed == 0
+             || early_reconvergence_exit_heuristic(router_opts, itry_since_last_convergence, timing_info, best_routing_metrics))
```
vaughnbetz, line 600: Could move hold_legal into the early_reconvergence_exit_heuristic routine, which I think would be easier to read by simplifying this condition and giving you a place to comment why you're not saying you're exiting early.

vaughnbetz (changes requested):

vaughnbetz:
Thanks David.  I have some questions and suggestions above.  One is major: I think the code is backwards, and others are useful for readability.

vpr/src/route/route_timing.cpp
dpbaines: I guess that check was there as a secondary quality check if the setup slacks were tied, to allow the router to pick routing with improved hold. I get that this would be unlikely though, so I guess it makes sense to remove this check for non-RCV routing.

vpr/src/route/route_timing.cpp
dpbaines: I agree that would be easier to read, but because the default max_convergence_count is 1, the router would stop early regardless. What I could do is to move all of these conditions into a seperate function?

dpbaines: Still outstanding

vaughnbetz: When RCV is selected you could change the legal_convergence_count default to some larger number (we can have dependent option defaults; take a look at the option parsing code and you should find an example).  Then if the early_reconvergence_exit_heuristic understands hold and rcv it can do the right thing.  The if statement won't look as long.

You could also move the whole if into a control routine; that would also work.  But I think changing legal_convergence_count makes sense no matter what, as I believe you've found RCV wants a higher value for this setting.  Better to do that by actually changing the value than writing if statements to work around it.

vpr/src/route/route_timing.cpp
```diff
@@ -579,11 +588,17 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
             pres_fac = update_pres_fac(router_opts.initial_pres_fac);
         }
 
+        if (budgeting_inf.if_set()) {
+            // If total negative slack is greater than or equal to 0, or RCV is finished, allow router to complete
+            hold_legal = timing_info->hold_total_negative_slack() >= 0 || rcv_finished_count == 0;
```
vaughnbetz, line 593: If this variable doesn't really mean hold is legal (it seems to mean RCV should stop trying) then you should give it a different name.

vpr/src/route/route_timing.cpp
```diff
@@ -2004,38 +2003,100 @@ bool should_setup_lower_bound_connection_delays(int itry, const t_router_opts& /
...
-        }
+            // Compare hold of current routing to best routing
+            int hold_comparison = compare_hold(best_routing_metrics, timing_info);
 
-        if (timing_info->hold_worst_negative_slack() > best_routing_metrics.hWNS) {
-            return true;
-        } else if (timing_info->hold_worst_negative_slack() > best_routing_metrics.hWNS) {
-            return false;
-        }
+            if (hold_comparison > 0) {
```
vaughnbetz, line 2081: This code could be shortened by checking if YOYO is enabled, and if so, checking hold and returning whatever is better, or doing nothing in a tie.  Then check setup (no matter what), then wirelength.

vaughnbetz:
The compare_setup and compare_hold routines have cleaned up some code.  I left some more comments on ways to shorten / clarify further.

vpr/src/route/route_timing.cpp
dpbaines: You're right, I should have seen this! Will fix thanks.

PR #1541, opened 2020-09-14
read_place: do not throw error if block name is invalid

acomodi:

#### Description

This fixes an issue when reading placement files.
Placement files can be read for two different reasons:

1. Prior to the routing stage if the flow is divided in three diffrent calls to VPR P&R stages
2. To read placement constraints files. 

In SymbiFlow, there are a set of tests which are automated and the verilog design alongside with the PCF constraints is automatically generated. The PCF constraints will translate into a placement constraints file.

The issue is that, during the various test design generation, some parameters may allow for some IO signals to be unconnected after synthesis, conflicting with the fact that VPR throws an error if a block specified in the placement constraints file is not present in the packed netlist.

Instead of throwing an error, this PR downgrades the error to a warning.

Another possible solution would be to encapsulate the check in a sub-routine that can be disabled at run-time with a flag (https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/25efd6670f14c1f23d976ef820ead8ef9b7c5e37/vpr/src/base/read_options.cpp#L1309-L1314).
This would require a sub-routine as the errors would otherwise be disabled in all the original routine, and those are errors which should not be disabled.

SymbiFlow tests

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/read_place.cpp
+++ b/vpr/src/base/read_place.cpp
@@ -226,7 +226,8 @@ void read_place_body(std::ifstream& placement_file,
                 AtomBlockId atom_blk_id = atom_ctx.nlist.find_block(block_name);
 
                 if (atom_blk_id == AtomBlockId::INVALID()) {
-                    VPR_THROW(VPR_ERROR_PLACE, "Block %s has an invalid name.\n", c_block_name);
+                    VTR_LOG_WARN("Block %s has an invalid name and it is going to be skipped.\n", c_block_name);
+                    continue;
                 } else {
                     blk_id = atom_ctx.lookup.atom_clb(atom_blk_id); //getting the ClusterBlockId of the cluster that the atom is in
                 }


vaughnbetz:
Downgrading to a warning is fine.  Code looks fine.  Can be merged as soon as CI goes green.

mithro:
@acomodi -- I'm not sure this is the correct solution to the problem here? Why are we getting invalid block names?

acomodi:
@mithro Specifically, the error I get is when running the [ice40 carry-stress tests](https://github.com/SymbiFlow/symbiflow-arch-defs/tree/master/tests/7-carry_stress).

For what I understand those tests are generated automatically for each board, as well as the PCF files. For some of those tests it happens that, after synthesis, some of the inputs remain unconnected, therefore not added to the packed netlist. 

The constraints file though contain those IO top level pins, generating the error described above.

Before recent changes applied to the ways placement files are being read and processed, if a block specified in the placement file was not found in the netlist, it resulted in a warning as well.

PR #1545, opened 2020-09-15
Remove temporary files added.

litghost:


vaughnbetz:
Thanks.  I just noticed those too -- not sure when they snuck in.

PR #1547, opened 2020-09-16
Fix initial placement for heterogeneous tiles

andrewb1999:
Signed-off-by: Andrew Butt <butta@seas.upenn.edu>

#### Description

In the case where a design contains more SLICEMs than there are CLBLM_Ls in the architecture, the initial placement would still accept the CLBLM_L as a valid placement tile, resulting in a lack of sub_tiles to make the placement. This happens because there was no distinction of valid sub_tiles for a SLICEM, so the CLBLM_L still has free SLICELs allowing CLBLM_L to be chosen as a placement tile for a SLICEM while not having any free SLICEM sub_tiles. This fix only checks the free locations of valid sub_tiles, causing CLBLM_L to be rejected for placement of a SLICEM and allowing the CLBLM_R to be used instead.

I made this fix when using the symbiflow-arch-defs defined architectures, but the issue is relevant for all architectures with heterogeneous tiles.

#### Related Issue

#1298 

#### Motivation and Context

This fix allows designs that use a large percentage of available heterogeneous sub tiles to be placed without error.

The fix has been tested locally, but it might be useful to add a test with some small test architecture.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -330,8 +330,9 @@ static t_physical_tile_type_ptr pick_placement_type(t_logical_block_type_ptr log
                                                     std::vector<std::vector<int>>& free_locations) {
     // Loop through the ordered map to get tiles in a decreasing priority order
     for (auto& tile : logical_block->equivalent_tiles) {
-        for (auto sub_tile : tile->sub_tiles) {
-            if (free_locations[tile->index][sub_tile.index] >= num_needed_types) {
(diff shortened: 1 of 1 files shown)


vaughnbetz:
The code looks good to me, although I think @acomodi wrote the underlying functions and hence is in the best position to comment.  kokoro seems hung; forcing it to restart.

acomodi (approved):
LGTM, let's wait for the nightly tests to complete

PR #1553, opened 2020-09-18
Replace lingering references to pl scripts in travis

jgoeders:
Although we moved to the Python flow scripts, there are a couple references to the Perl scripts in `.travis.yml`.  This fixes those.

#### Types of changes

#### Checklist:


vaughnbetz:
Looks like the disk on kokoro filled up.  I'll try rerunning.  Other kokoro tests have been passing, so if this fails again it may mean there is some problem with the change.

Issue #1546, opened 2020-09-16
Packer assertion failure on clock buffer primitive


mkurc-ant:
I want to make a net of a top-level input port become a clock source so that I can constrain it using SDC directive(s).

The target FPGA architecture requires that there is a `CLOCK_CELL` IO buffer primitive at a clock input that drives a global clock distribution buffer `GMUX_IP`. From there the clock can be routed to flip-flops. I want the clock to be propagated through both of these primitive cells.

I modeled both `CLOCK_CELL` and `GMUX_IP` primitives as clock buffers according to: https://docs.verilogtorouting.org/en/latest/tutorials/arch/timing_modeling/#clock-buffers-gates

When VPR is run it correctly identifies top-level inputs as clock sources but fails on an assertion during pre-packing.

#### Expected Behaviour
Packer succeeds, input ports `clk(0)` and `clk(1)` are considered as netlist clock sources.

#### Current Behaviour
Netlist clock sources are identified fine but packing fails with `vpr/src/pack/prepack.cpp:1127 get_driving_block: Assertion 'pin_number == 1' failed.` error.

#### Possible Solution
Unclear. Simply commenting out the assertion makes the packer pass (at least for the attached design).

The problem seems not to be related to "multi-clock primitives" as the TODO comment in the code suggests: `//TODO: support multi-clock primitives`. I'd like to know why the assertion is there and why `pin_number` has to be 1.

#### Steps to Reproduce
1. Use the architecture file and EBLIF netlist from the attached archive:  [pack.tar.gz](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5231842/pack.tar.gz)
2. Point the env. variable `VPR` to the VPR binary and run `runme.sh`

#### Context
I want to make a net of a top-level input port become a clock source so that I can constrain it using SDC directive(s).

#### Your Environment
* VTR revision used: 8.1.0-dev+7c167f8c8
* Operating System and version: Linux Ubuntu 18.04 LTS
* Compiler version: GNU 7.4.0 on Linux-4.15.0-45-generic x86_64

mkurc-ant:
@kmurray Could you take a look at it?

mkurc-ant:
@vaughnbetz Could you take a look at it?

vaughnbetz:
The proposed fix looks OK to me; see PR for comments.

PR #1556, opened 2020-09-21
prepack: relax assertion on model port clock input pin_number

acomodi:

#### Description

This PR relaxes the assertion about the pin number when the model port is a clock input during prepacking.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1546

#### Motivation and Context

If an input PAD needs to be treated as a clock generator to add the possibility to have SDC constraints, VPR would throw an error on the PAD primitive as the clock pin would not be valid (according to the assertion), given that the input clock pin would be at position 0 instead of 1.

With this PR, the assertion is relaxed, in a way that now, if the driver block is of an `.input` type, the assertion does not get triggered.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/pack/prepack.cpp
+++ b/vpr/src/pack/prepack.cpp
@@ -1123,10 +1123,6 @@ static AtomBlockId get_sink_block(const AtomBlockId block_id, const t_model_port
  *      pin_number : the pin_number of the pin driven by the net (pin index within the port)
  */
 static AtomBlockId get_driving_block(const AtomBlockId block_id, const t_model_ports* model_port, const BitIndex pin_number) {
-    if (model_port->is_clock) {
-        VTR_ASSERT(pin_number == 1); //TODO: support multi-clock primitives
-    }
-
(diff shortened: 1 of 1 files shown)


vaughnbetz:
The change is fine with me.  To be honest I don't understand the original assert anyway, as the concept of pin_number is not well commented (what are the rules for pin_number?  What does it represent?).  That's prior code outside this PR, but if you know the answer to those questions updating the comment on t_physical_type would be helpful.

acomodi:
@vaughnbetz I tried to track the pin_number down as well and try and get an understanding of its purposes and where it is used, but couldn't find documentation as well.

One thing for sure is that this is not related at all to physical types, as physical types are used during the placement step, and this piece of code concerns prepacking, therefore atom netlist and models/pb_types.

vaughnbetz:
OK, releaxing the assert is certainly fine as none of us know exactly what it is for!  Adding @kmurray and @luujason in case either can give insight on what this pin_number is so we can comment it.

Issue #1539, opened 2020-09-13
Warning from g++ in vtr_vector_map.h due to unsigned integer sign check 


dpbaines:
#### Expected Behaviour

There should be no or very minimal warnings during the build process. This means that when there are warnings or errors we can see them clearly.

#### Current Behaviour

Updating g++ has created the warning output I posted below when building VPR. This warning is generated everytime the vector_map class is referenced, which is a lot. This means the output is completely spammed with warnings.

```
/home/dbaines/Documents/research/vtr-verilog-to-routing/libs/libvtrutil/src/vtr_vector_map.h: In instantiation of ‘vtr::vector_map<K, V, Sentinel>::const_reference vtr::vector_map<K, V, Sentinel>::operator[](K) const [with K = vtr::StrongId<cluster_port_id_tag>; V = vtr::StrongId<cluster_block_id_tag>; Sentinel = vtr::DefaultSentinel<vtr::StrongId<cluster_block_id_tag> >; vtr::vector_map<K, V, Sentinel>::const_reference = const vtr::StrongId<cluster_block_id_tag>&]’:
/home/dbaines/Documents/research/vtr-verilog-to-routing/vpr/src/base/netlist.tpp:217:24:   required from ‘BlockId Netlist<BlockId, PortId, PinId, NetId>::port_block(PortId) const [with BlockId = vtr::StrongId<cluster_block_id_tag>; PortId = vtr::StrongId<cluster_port_id_tag>; PinId = vtr::StrongId<cluster_pin_id_tag>; NetId = vtr::StrongId<cluster_net_id_tag>]’
/home/dbaines/Documents/research/vtr-verilog-to-routing/vpr/src/base/netlist.tpp:327:12:   required from ‘BlockId Netlist<BlockId, PortId, PinId, NetId>::pin_block(PinId) const [with BlockId = vtr::StrongId<cluster_block_id_tag>; PortId = vtr::StrongId<cluster_port_id_tag>; PinId = vtr::StrongId<cluster_pin_id_tag>; NetId = vtr::StrongId<cluster_net_id_tag>]’
[...]

PR #1559, opened 2020-09-22
Remove index >= 0 check from vtr_vector_map.h

dpbaines:
#### Description

This check seems to be causing an issue on the latest version of g++, where it's complaining about checking the sign on an unsigned type. See the issue linked for more information.

#### Related Issue

Closes https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1539

#### Motivation and Context

Cleans up the build output, otherwise you get a warning everytime vector_map is referenced.

I built it successfully, and tested it on a super small circuit, everything seems in order.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libvtrutil/src/vtr_vector_map.h
+++ b/libs/libvtrutil/src/vtr_vector_map.h
@@ -62,6 +62,11 @@ class vector_map {
     //Indexing
     const_reference operator[](const K n) const {
         size_t index = size_t(n);
+
+        // Shouldn't check for index >= 0, since size_t is unsigned thus won't be negative
+        // A negative input to n would result in an absurdly large number close the maximum size of size_t, and be caught by index < vec_.size()
+        // http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf chapter 4.7 para 2
+
         VTR_ASSERT_SAFE_MSG(index < vec_.size(), "Out-of-range index");
         return vec_[index];
     }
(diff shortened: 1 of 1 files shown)


libs/libvtrutil/src/vtr_vector_map.h
```diff
@@ -62,7 +62,7 @@ class vector_map {
     //Indexing
     const_reference operator[](const K n) const {
         size_t index = size_t(n);
```
vaughnbetz, line 64: I'm nervous about losing the check.  I'd move the VTR_ASSERT_SAFE check for index>=0 to before the size_t conversion (i.e.
VTR_ASSERT_SAFE (n>=0)
would come right after the function header.

dpbaines: I would agree, but vector_map is only used by StrongId types, which only allow conversion to size_t, thus will never be negative. If a negative number got involved, it would wrap around and be caught by the `index < vec_.size()` check. This could result in undefined behaviour, but this would have to be handled by the StrongId type, not vector_map.

dpbaines: I tried adding the ASSERT_SAFE (n >= 0) to the top, but it won't compile since StrongId doesn't have a < operator.

vaughnbetz: OK, sounds good.  Adding @kmurray in case he has any objections, but that sounds reasonable.

vaughnbetz: David, can you add a comment explaining this (we're converting to size_t, which can't be negative, so just check if it is too big).

dpbaines: Yeah sure that's not a bad idea. Will do!

vaughnbetz:
Thanks David.

PR #1561, opened 2020-09-22
Add a unit test for the ConnectionRouter

HackerFoo:
Adds a unit test for the ConnectionRouter.

#### Description
This unit test is both to verify functionality, and as a starting point to experiment with the router.

#### Motivation and Context
Unit tests allow quicker iteration because they run quickly and make it easier to locate bugs. Furthermore, they serve as documentation on to do something, and a starting point for changes.

Specifically, I'm working on parallel routing algorithms, so this is a starting point for that.

The unit test passes in `make test`.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/test/test_connection_router.cpp
+++ b/vpr/test/test_connection_router.cpp
@@ -1,0 +1,163 @@
+#include "catch.hpp"
+
+#include "vpr_api.h"
+#include "vpr_signal_handler.h"
+#include "globals.h"
+#include "net_delay.h"
+#include "place_and_route.h"
+#include "route_tree_timing.h"
+#include "timing_place_lookup.h"
+
+static constexpr const char kArchFile[] = "../../vtr_flow/arch/timing/k6_frac_N10_mem32K_40nm.xml";
+static constexpr int kMaxHops = 10;
+
+namespace {
+
+// Route from source_node to sink_node, returning either the delay, or infinity if unroutable.
+static float do_one_route(int source_node, int sink_node, const t_router_opts& router_opts, const std::vector<t_segment_inf>& segment_inf) {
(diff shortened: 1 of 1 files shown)


vpr/test/test_connection_router.cpp
```diff
@@ -0,0 +1,140 @@
+#include "catch.hpp"
```
mithro, line 1: This needs comments.

Should explain what the test is testing and why it exists.

HackerFoo: Done.

vpr/test/test_connection_router.cpp
```diff
@@ -0,0 +1,131 @@
...
+    return delay;
+}
+
+// Test that the router can route nets individually, not considering congestion.
+// This is a minimal timing driven routing test that can be used as documentation,
+// and as a starting point for experimentation.
+TEST_CASE("connection_router", "[vpr]") {
+    // Source and sink VPR would have chosen for wire.eblif
+    // Determined by running VPR with the same parameters.
+    int source_rr_node = 198;
```
litghost, line 84: This is dangerous because a change in the rr graph generator would change these numbers.  Consider your own randomize the rr graph.

A couple examples you could do:
 - Run the pack/place for `wire.eblif` to populate the net terminals
 - Pick random SOURCE / SINK from the graph to route
 - Search the cluster list for IO tiles (e.g. model == ".input" / ".output", etc) and use SOURCE and SINK rr nodes from those tiles

HackerFoo: I've added a more robust method for picking the source and sink by walking the RR graph.

PR #1564, opened 2020-09-28
Add symbiflow tests

acomodi:
#### Description

This is an initial PR to have the SymbiFlow XC7 architecture integrated within the `vtr_flow`.

There is an initial download step of the latest SymbiFlow package, with the extraction of the Artix 50T data to have it working within VPR:
- Arch XML
- RR graph
- Router lookahead
- Place delay matrix lookup

A very simple initial circuit file has been added, which represent a synthesized counter.

This PR is still marked as WIP for two main reasons:
- There is an issue with the architecture or too strict VPR checks around buffered switches. In fact, to let the test pass, I had to temporarily demote a VPR_ERROR to a warning. ([Issue](https://github.com/SymbiFlow/vtr-verilog-to-routing/issues/548))
- Need to either allow to correctly read symlinks within VPR to get the RR graph and other data files, or have the vtr_flow accept the addition of additional <flags, file> pairs and add them to the VPR command. The data files (RR graph, etc.) should be located in the `arch` directory.

#### Motivation and Context

With the Symbiflow/VTR fork being almost equal to the upstream version, it is possible now to insert SymbiFlow tests within the vtr_flow. This helps to track possible regressions against SymbiFlow architectures, as well as increasing the robustness of the test suite. 

New symbiflow counter test added.

#### Types of changes

#### Checklist:


acomodi:
@mithro @litghost @vaughnbetz FYI. This is still a WIP and requires some more adjustments, but it is a starting point to begin the introduction of SymbiFlow architectures and benchmarks.

In this PR, the nightly tests have been reduced to only run the symbiflow counter example, to check whether everything is working fine.

vtr_flow/tasks/regression_tests/vtr_reg_nightly/symbiflow/config/config.txt
```diff
@@ -0,0 +1,30 @@
...
+# Parse info and how to parse
+parse_file=vpr_standard.txt
+
+# How to parse QoR info
+qor_parse_file=qor_standard.txt
+
+# Pass requirements
+pass_requirements_file=pass_requirements.txt
+
+#The Titan benchmarks are run at a fixed channel width of 300 to simulate a Stratix IV-like routing architecture
```
litghost, line 26: This comment is out of date

acomodi:
@litghost @vaughnbetz According to what discussed, I am trying to get the RR indexed data factored out to be used only for the classic router lookahead.
The problem I am facing is that this RR indexed data is used not only in the classic lookahead, but in other locations:
1. https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/469ab561714b954fdb9cc59cd1c33b437bcbd11d/vpr/src/route/router_lookahead_map_utils.cpp#L93

In this case, the base cost is taken through the `get_single_rr_cong_base_cost` function which in turn uses the RR indexed data for the base cost.
2. https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/469ab561714b954fdb9cc59cd1c33b437bcbd11d/vpr/src/route/route_timing.cpp#L1550-L1556
3. and other locations

I am unsure how to proceed, as, for instance, the base cost calculation depends on the T_linear and T_quadratic values as well when building the RR indexed data structure.

litghost:
> @litghost @vaughnbetz According to what discussed, I am trying to get the RR indexed data factored out to be used only for the classic router lookahead.

As stated before, the base cost should be per cost index, and the base cost compuation doesn't depend on any of the other variables:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/469ab561714b954fdb9cc59cd1c33b437bcbd11d/vpr/src/route/rr_graph_indexed_data.cpp#L192-L232

The idea is everything except for the base cost can likely be split.

This case is a little hairy:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/469ab561714b954fdb9cc59cd1c33b437bcbd11d/vpr/src/route/route_timing.cpp#L1550-L1556

@vaughnbetz Can you explain why `factor` is only multiplied in the T_quadratic calculation?

acomodi:
> As stated before, the base cost should be per cost index, and the base cost compuation doesn't depend on any of the other variables

Actually, as the base cost computation depends on the normalization which, in turn, depends on T_quadratic, T_del and so on (this only if `base_cost_type != DEMAND_ONLY` and `!= DEMAND_ONLY_NORMALIZED_LENGTH`).

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/469ab561714b954fdb9cc59cd1c33b437bcbd11d/vpr/src/route/rr_graph_indexed_data.cpp#L267-L307

litghost:
> > As stated before, the base cost should be per cost index, and the base cost compuation doesn't depend on any of the other variables
> 
> Actually, as the base cost computation depends on the normalization which, in turn, depends on T_quadratic, T_del and so on (this only if `base_cost_type != DEMAND_ONLY` and `!= DEMAND_ONLY_NORMALIZED_LENGTH`).
> 
> https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/469ab561714b954fdb9cc59cd1c33b437bcbd11d/vpr/src/route/rr_graph_indexed_data.cpp#L267-L307

Ugh

acomodi:
Given that the error returned should be affecting only specific run cases (classic lookahead, or any lookahead with base_cost_types that make use of the T_quadratic/linear values, etc.), does it make sense to have the VTR_ERROR changed to VPR_THROW and disable that with the `--disable_errors` flag?

litghost:
> Given that the error returned should be affecting only specific run cases (classic lookahead, or any lookahead with base_cost_types that make use of the T_quadratic/linear values, etc.), does it make sense to have the VTR_ERROR changed to VPR_THROW and disable that with the `--disable_errors` flag?

So based on the examples you showed, the inconsistent buffering will affect the router in the following was:
 - It corrupts the delay normalization factor a small amount.  This is likely fine, as the affect is likely small
 - It will cause T_quadratic to be incorrect for `update_rr_base_costs` case

It is unclear how important the code in `update_rr_base_costs` is, and whether having this wrong on the 7-series graph will have dramatic effects.

@vaughnbetz , thoughts?

acomodi:
@litghost @vaughnbetz I have done some digging and found out that the problem resides in the fact that in SymbiFlow we add `short` switches to connect wires. These short switches are "fake" in the sense that their cost values are all set to zero.

The problem might be that, when calculating the average T_del, T_quadratic and C for a node, we end up adding up in the sum of the number of switches also the "fake" switches:

```
NODE: RR node: 223968 type: CHANX location: (17,40) <-> (46,40) track: 23 len: 29 longline: 0 seg_type: LH dir: BI_DIR capacity: 1 fan-in: 26 fan-out: 40, COST_INDEX: 49
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0012375_C9.404e-09_Tdel1.6800000000000001e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7699999999999998e-10
        SWITCH NAME: routing_R0.0012375_C9.404e-09_Tdel1.65e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7699999999999998e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.89e-10
        SWITCH NAME: routing_R0.0012375_C9.404e-09_Tdel1.6800000000000001e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7599999999999999e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.75e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.8e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.8e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.75e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7599999999999999e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.85e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7799999999999998e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7699999999999998e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.8e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0012375_C9.404e-09_Tdel1.6700000000000002e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.79e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.89e-10
        SWITCH NAME: routing_R0.0012375_C9.404e-09_Tdel1.6700000000000002e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0012375_C9.404e-09_Tdel1.65e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7699999999999998e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.89e-10
        SWITCH NAME: routing_R0.0015845231875_C9.979e-09_Tdel1.7699999999999998e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: routing_R0.0007569375_C1.3478e-08_Tdel3.01e-10
        SWITCH NAME: short
Warning 21: Inconsitent buffering of children of rr node LH (RR node: 223968 type: CHANX location: (17,40) <-> (46,40) track: 23 len: 29 longline: 0 seg_type: LH dir: BI_DIR capacity: 1 fan-in: 26 fan-out: 40)
        SWITCH NAME: short
Warning 22: Inconsitent buffering of children of rr node LH (RR node: 223968 type: CHANX location: (17,40) <-> (46,40) track: 23 len: 29 longline: 0 seg_type: LH dir: BI_DIR capacity: 1 fan-in: 26 fan-out: 40)

```

In the example above, for the LH node, we have the last two switches which are short, and the warning is displayed, but, even though those have zero delay, I am unsure whether the number of switches to perform the avarage calculation should be taken into account.

Moreover, the error we are getting about the buffering inconsistencies is due to the fact that, for the same cost index, we have only short swithces:

```
NODE: RR node: 610507 type: CHANX location: (16,81) <-> (58,81) track: 10 len: 42 longline: 0 seg_type: LH dir: BI_DIR capacity: 1 fan-in: 2 fan-out: 2, COST_INDEX: 49
        SWITCH NAME: short
        SWITCH NAME: short

```

acomodi:
@vaughnbetz @litghost @mithro I have rebased this PR and modified it to get the latest symbiflow packages.

I have also added the possibility to add additional data files to the VPR commands with their corresponding flag.

E.g, to have a custom rr_graph file, the config file should have the following:
```
additional_files_list_add=--read_rr_graph,<rr_graph_file>
```

The nightly tests passed with only the symbiflow test enabled ([link to the build instance](https://source.cloud.google.com/results/invocations/9948500a-dbbc-4a51-bedf-d70a20790eef/targets/foss-fpga-tools%2Fverilog-to-routing%2Fupstream%2Fpresubmit%2Fnightly/log)). I will re-enable all the nightly tests as well and remove the WIP label

litghost:
CI is red?

vtr_flow/benchmarks/symbiflow/symbiflow_counter.eblif
```diff
@@ -0,0 +1,1132 @@
+# Generated by Yosys 0.9+2406 (git sha1 d8b2d1a2, x86_64-conda_cos6-linux-gnu-gcc 1.23.0.449-a04d0 -fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -fdebug-prefix-map=/tmp/really-really-really-really-really-really-really-really-really-really-really-really-really-long-path/conda/conda-bld/symbiflow-yosys_1594197534228/work=/usr/local/src/conda/symbiflow-yosys-0.8_6021_gd8b2d1a2 -fdebug-prefix-map=/data/tmp/archdefs/env/conda/envs/symbiflow_arch_def_base=/usr/local/src/conda-prefix -fPIC -Os -fno-merge-constants)
```
litghost, line 1: What's the story for how to regenerate this file?

vtr_flow/scripts/upgrade_arch.py
```diff
@@ -47,6 +47,7 @@ def __init__(self):
     "add_tile_tags",
     "add_site_directs",
     "add_sub_tiles",
+    "remove_penalty_cost",
```
litghost, line 50: I believe we can remove this now?

vtr_flow/benchmarks/symbiflow/symbiflow_counter.eblif
litghost: Because the download script pulls the latest, it is possible (and likely) that this eblif will become out of date automatically.  How will that be resolved?  @HackerFoo has a similiar issue for the PGO VTR issue, consider consulting with each other?

HackerFoo: We could generate a tarball with eblifs along with each -install tarball, as part of the same build.

HackerFoo (approved):

vtr_flow/benchmarks/symbiflow/symbiflow_counter.eblif
litghost: > We could generate a tarball with eblifs along with each -install tarball, as part of the same build.

I think that makes sense to me.

acomodi: Yeah, I think this is the best way to keep things synchronized.
I'd say that, given we only enable small size counter test, for the time being, we could push this as-is, while we wait for a valid package containing the circuits is generated by arch-defs master. As soon as more circuits are available, we can update the `get_symbiflow_benchmarks` target to get also the circuits automatically and remove the counter test.

litghost: > I'd say that, given we only enable small size counter test, for the time being, we could push this as-is,

No, I do not believe we should merge this as is.  This will break the VTR CI as soon as the blif and tarball get out of date.

Adding example blifs for testing shouldn't be a heavy lift, and I think doing this correctly such that it won't break the VTR CI is valuable.

acomodi:
@litghost @HackerFoo @vaughnbetz I have added the benchmarks artifact from GCS as well and enabled several tests.

There are some of them which require the placement constraints (e.g. minilitex_ddr) and for the time being they will not get executed, until the benchmarks package contains the constraints.

[This](https://source.cloud.google.com/results/invocations/bd2af504-4fe1-4c7a-ae77-8aadb146fd61/targets/foss-fpga-tools%2Fverilog-to-routing%2Fupstream%2Fpresubmit%2Fnightly/log) is a nightly build with only symbiflow enabled.

There seem to be some issues with the latest link generation as the CI on symbiflow-arch-defs master doesn't seem to have produced them, therefore, for the time being I have added static links to gather the symbiflow packages.

vtr_flow/scripts/python_libs/vtr/task.py
```diff
@@ -331,10 +347,29 @@ def create_jobs(args, configs, longest_name=0, longest_arch_circuit=0, after_run
...
-                ]
+                sdc_name = "{}.sdc".format(Path(circuit).stem)
+                try:
+                    sdc_file = resolve_vtr_source_file(config, sdc_name, config.sdc_dir)
+
+                    cmd += [
+                        "-sdc_file",
+                        "{}".format(sdc_file)
+                    ]
+                except InspectError:
```
litghost, line 358: I don't understand why you are allowing the `except`?

acomodi: In general, It Is possible that some of the tests do not have an sdc file. In such a case, the resolve path function would fail.

litghost: The original script has `sdc_dir` and `place_constr_dir` to control if the SDC and/or place is expected?  So I don't think this change is right.

acomodi: Ok, I know what is happening. There are two issues here:

1. From symbiflow-arch-defs there are two ways of using the SDC, one is with an explicit constraint file, and the other is through the SDC yosys plugin. In the latter case, the file that gets installed in the benchmarks directory is a symlink, which generates an error when trying to resolve the path.
2. Some tests do not make use of the SDC file (namely counter, murax_full_50, murax_full_100).

Both of these issues should be resolved in arch-defs. We should start cleaning up everything and transition to use the SDC plugin instead of relying on explicit manually written SDCs, as well as adding constraints where they are missing.

For the time being I will temporarily disable tests that do have SDC issues and

litghost:
Latest failure looks like the workspace disk issue:

```
Test 'vtr_reg_nightly' had 0 qor test failures

Test 'vtr_reg_nightly' had 0 run failures

All tests passed
++ kill 11388
github/vtr-verilog-to-routing/.github/kokoro/run-vtr.sh: line 29: 11388 Terminated              ( while :; do
    date; uptime; free -h; sleep 300;
done )


[ID: 6118664] Build finished after 66386 secs, exit value: 0


Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
[22:57:45] Executor 1152195486 is unresponsive. Build terminated.
```

See https://github.com/SymbiFlow/symbiflow-arch-defs/blob/27c47842f479fc01f5765ea8b9039cba337e7532/.github/kokoro/package_results.sh#L42-L47

acomodi:
@litghost @vaughnbetz @mithro Nightly finally went green. I have added a remove step that deletes all data necessary to run ISPD, Titan and SymbiFlow benchmarks (circuits and arch data).

Nightly CI took 21 hours to complete with 3 parallel cores.

mithro:
@acomodi - In the future, rather than having a "nightly" test suite, it might make sense to split the nightly into `VTR ISPD`, `VTR Titan`, `VTR SymbiFlow` (which can run in parallel)? Can you log a github issue to look into that?

mithro:
@acomodi - Another non-blocking issue -- we have kokoro "big" machines which have 96 cores. It would be worth understanding if we can use them to make these runs much faster without affecting the quality of the result? (Also log a github issue for that)

acomodi:
@mithro Sure, I'll open the issues.

>  we have kokoro "big" machines which have 96 cores.

If we are going towards splitting nightly into ISPD, Titan and Symbiflow we could potentially use multiple cores without getting into out-of-memory issues. At the moment, there are 3 parallel cores due to excesses in RAM utilization. If the "big" machines also have bigger RAM, I'd say we should move to those.

vaughnbetz:
Makes sense.  When you say VTR ISPD, do you mean the VTR benchmarks, or do you mean a new test?  I don't know of a test on ISPD benchmarks right now.

mithro:
> @mithro Sure, I'll open the issues.
> 
> > we have kokoro "big" machines which have 96 cores.
> 
> If we are going towards splitting nightly into ISPD, Titan and Symbiflow we could potentially use multiple cores without getting into out-of-memory issues. At the moment, there are 3 parallel cores due to excesses in RAM utilization. If the "big" machines also have bigger RAM, I'd say we should move to those.

@acomodi - Yes the big machines have a lot more RAM. 
 * 16 vCPUs, 104 GB - `n1-highmem-16`
 * 96 vCPUs, 360 GB - `n1-standard-96`

mithro:
> Makes sense. When you say VTR ISPD, do you mean the VTR benchmarks, or do you mean a new test? I don't know of a test on ISPD benchmarks right now.

The nightly benchmarks include the ISPD benchmarks if I understand correctly.

acomodi:
> When you say VTR ISPD, do you mean the VTR benchmarks, or do you mean a new test?

@vaughnbetz Currently, in the nightly run, there are these ISPD tests enabled: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/0b7e606904cdf0f0896759d97a566f6a276ae48c/.github/kokoro/steps/vtr-full-setup.sh#L4

vaughnbetz:
Ah, I see.  I found the test in vtr_reg_weekly:
regression_tests/vtr_reg_weekly/vpr_ispd

litghost:
> @litghost @vaughnbetz @mithro Nightly finally went green. I have added a remove step that deletes all data necessary to run ISPD, Titan and SymbiFlow benchmarks (circuits and arch data).
> 
> Nightly CI took ~21 hours to complete with 3 parallel cores.

Even after the cleanup, the output directory is quiet big: 
```
+++ du -sh
++ echo 'Working directory size: 89G	.'
Working directory size: 89G	.
```

We should probably try to figure out why this is the case, and shrinking the overhead.  For reference, it took ~45 minutes to move the 90G directory:
```
[05:30:45] Collecting build artifacts from build VM
[06:17:25] Kokoro builder finished
```

vaughnbetz:
That is really big.  Yes, I think it's important to shrink that; I can't think of a good reason for the output to be that large, and it will cause problems (e.g. if someone runs the test locally to debug something) to have such large amount sizes.

acomodi:
I believe this might be due to the fact that the VTR flow copies the circuit under test in the build directory, instead of generating a symlink?

This could result in the various `runN` directories to be prettly loaded with lots of data.

For reference, the whole `symbiflow` build directory weights ~500 MB.

I don't have data on the Titan benchmarks, but, if they are in total sth like ~10GB and if they get replicated in the build directory, the sum of all the circuits size and VPR outputs might generate that amount of data.

litghost:
@acomodi To separate whether the large output directory is new or not, can you open a new PR with just the changes to the working directory?  Assuming that the output directory was large before, I don't believe it blocks this PR, just something to fix in the future.

vaughnbetz:
Agreed, if this was an existing issue then it's not blocking.

acomodi:
PR opened here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1610.

I have also rebased this on top of master

acomodi:
@litghost @vaughnbetz CI went green and the "big" data usage happens also in master: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1610#issuecomment-745158418

PR #1576, opened 2020-10-23
rr_graph: implement more robust delay normalization calculation

acomodi:

#### Description

This PR aims at making the delay normalization factor used in the based cost calculation to be more representative of the real underneath device fabric.

Previous to this PR, the calculation of the const_indices' delays/resistance/capacitance was based on heuristics, which proved to be insufficient for complex RR Graphs such as the Series7 one.

All the nodes in the RR graph are now checked and their switches delays averaged, therefore we check now for every node instead of picking them as samples.

The main differences between the previous indexed data calculation are:
- `short` switches are no longer increasing the total delay values.
- If a node does not have any switch (due to the previous point), it's delay contribution won't be added to the whole sum.
- The delay normalization calculation is solely based on the previous rr_graph_indexed data instead of being based to some fixed values (such as `clb_dist`).
  - If nodes were not added and a cost index has null values, it won't contribute to the final delay normalization value.

#### Motivation and Context

Base costs for Series7 device had a 15x reduction making the delay and base costs not comparable anymore.
The 15x reduction was due to an uncaught bug in the VtR/SymbiFlow fork, which get caught as soon as I had tested the upstream master against SymbiFlow tests.

Currently, all strong and basic test run through completion, but some of the strong tests do report a change in the observed metrics, which in all cases but one seem to be a positive change (less CPD, wire length, etc).

#### Types of changes

#### Checklist:


acomodi:
@litghost @vaughnbetz @HackerFoo FYI, I have tested this against SymbiFlow tests and it seems to be solving the base cost value reduction and the consequent runtime degradation.

With this PR, the base cost reported for the A35T device is now only ~1.06x lower than the previous values.
Another thing I have observed is that the base cost seems to be consistent across different RR graphs (for instance the ROI-based one and the 50T). 
I have observed that previously, the various RR graphs reported a 5x difference in the base costs between these two RR graphs.

acomodi:
Failing tests include:

```
regression_tests/vtr_reg_strong/strong_custom_grid...[Fail]
[Fail]
non_column.xml/raygentop.v/common min_chan_width_routing_area_total relative value 0.7979175341181404 outside of range [0.8,1.3] and not equal to golden value: 3434390.0
[Fail]
non_column.xml/raygentop.v/common min_chan_width_routing_area_per_tile relative value 0.7979173735061246 outside of range [0.8,1.3] and not equal to golden value: 3153.71

regression_tests/vtr_reg_strong/strong_power...[Fail]
[Fail]
k6_frac_N10_mem32K_40nm.xml/ch_intrinsics.v/common total_power relative value 0.8824649298597196 outside of range [0.9,1.1] and not equal to golden value: 0.00998

regression_tests/vtr_reg_strong/strong_bidir...[Fail]
[Fail]
k4_n4_v7_longline_bidir.xml/styr.blif/common critical_path_delay relative value 1.4492353123212178 outside of range [0.5,1.4] and not equal to golden value: 8.75717
[Fail]
k4_n4_v7_longline_bidir.xml/styr.blif/common geomean_nonvirtual_intradomain_critical_path_delay relative value 1.4492353123212178 outside of range [0.5,1.4] and not equal to golden value: 8.75717
[Fail]
k4_n4_v7_longline_bidir.xml/styr.blif/common setup_WNS relative value 1.4492353123212178 outside of range [0.5,1.4] and not equal to golden value: -8.75717

regression_tests/vtr_reg_strong/strong_clock_modeling...[Fail]
[Fail]
timing/k6_N10_40nm.xml/verilog/mkPktMerge.v/common_--clock_modeling_ideal_--route_chan_width_60 routed_wirelength relative value 0.47368421052631576 outside of range [0.6,1.5], above absolute threshold 5.0 and not equal to golden value: 38.0
```

I am also running the `quick_titan` QoR benchmarks to have more data.

vpr/src/route/rr_graph_indexed_data.cpp
```diff
@@ -264,101 +250,67 @@ static std::vector<size_t> count_rr_segment_types() {
...
-            continue;
-        cost_index = device_ctx.rr_nodes[inode].cost_index();
-        frac_num_seg = clb_dist * device_ctx.rr_indexed_data[cost_index].inv_length;
-        Tdel = frac_num_seg * device_ctx.rr_indexed_data[cost_index].T_linear
-               + frac_num_seg * frac_num_seg
-                     * device_ctx.rr_indexed_data[cost_index].T_quadratic;
-        Tdel_sum += Tdel / (float)clb_dist;
+        if (rr_indexed_data[cost_index].number_of_nodes == 0 || T_value == 0.0) continue;
+
+        Tdel_vector.push_back(T_value * 1e10);
```
litghost, line 265: Where did this magic number of come from?  It should be `constexpr double kNameOfConstant = 1e10`, along with a comment about the value should be.

vpr/src/route/rr_graph_indexed_data.cpp
```diff
@@ -264,101 +250,67 @@ static std::vector<size_t> count_rr_segment_types() {
...
+    for (size_t cost_index = CHANX_COST_INDEX_START; cost_index < rr_indexed_data.size(); cost_index++) {
+        float T_value = rr_indexed_data[cost_index].T_linear + rr_indexed_data[cost_index].T_quadratic;
 
-    for (itrack = 0; itrack < nodes_per_chan; itrack++) {
-        inode = find_average_rr_node_index(device_ctx.grid.width(), device_ctx.grid.height(), CHANX, itrack,
-                                           L_rr_node_indices);
-        if (inode == -1)
-            continue;
-        cost_index = device_ctx.rr_nodes[inode].cost_index();
-        frac_num_seg = clb_dist * device_ctx.rr_indexed_data[cost_index].inv_length;
```
litghost, line 286: The `inv_length` doesn't seem to be accounted for anymore?

vpr/src/route/rr_graph_indexed_data.cpp
```diff
@@ -264,101 +250,67 @@ static std::vector<size_t> count_rr_segment_types() {
...
-               + frac_num_seg * frac_num_seg
-                     * device_ctx.rr_indexed_data[cost_index].T_quadratic;
-        Tdel_sum += Tdel / (float)clb_dist;
+    double mult = 1.0f;
+    for (auto Tdel : Tdel_vector) {
+        mult *= Tdel;
     }
 
-    return (Tdel_sum / (2. * nodes_per_chan));
+    return std::pow(mult, 1.0 / Tdel_vector.size()) / 1e10;
```
litghost, line 273: Can you explain where this equation came from and why it is right?

vpr/src/route/rr_graph_indexed_data.cpp
acomodi: Right, I forgot to fix that. Anyway, this was needed to temporary shift delay values, as during the geo mean computation, the subsequent multiplication of the delays would eventually bring the `mult` value to zero. 
Unsure whether this is the best way to achieve this though.

vpr/src/route/rr_node.h
```diff
@@ -185,6 +185,7 @@ struct t_rr_indexed_data {
     float T_linear = std::numeric_limits<float>::quiet_NaN();
     float T_quadratic = std::numeric_limits<float>::quiet_NaN();
     float C_load = std::numeric_limits<float>::quiet_NaN();
+    int number_of_nodes = 0;
```
litghost, line 188: Does this value need to be kept around?  If not, remove it from this structure and keep it temporarily.

litghost (commented):
Comments in.  @vaughnbetz needs to do a detailed review as well.

vpr/src/route/rr_graph_indexed_data.cpp
litghost: Floating point doesn't need shifts like this.  Rather than using 1e10, just a `double` during the intermediate calcuation.

acomodi: Actually I tried with a double for the intermediate calculation, but it went actually beyond the limit which is around `1e-300`

litghost: So the issue is that maybe `product(Tdel_vector)` is the wrong math to use here?  I think you may have a numerical convergence error here.

acomodi: The fact is that I wanted to use the geometric mean to take into account possible delay values outliers of some segment types which, I have seen, may differ by two orders of magnitude.

Given that delay values are usually in the order of 1e-10, I applied the 1e10 correction for the intermediate step and than restored it at the end of the geo mean calculation.

litghost: The solution here is to change from `(product(T_del))^(1/len(T_del))` to https://en.wikipedia.org/wiki/Geometric_mean#Relationship_with_logarithms `e^(1/len(T_del)*sum(map(logn(T_del))))`.   `logn(0)` is a violation, but those need to be skipped in either case.  Again, the problem here is the naive geomean algorithm converges always to zero as N increases.  This is numerically denegenerate, and you can always increase N to drive the product to zero.

vaughnbetz: I don't think geometric mean is the right approach here.  It does fairly average numers with widely varying magnitude ranges (treats them all as equally significant) but it is very dangerous if numbers can go to or approach 0.  We could get some small/weird wires or fragments with delay near 0, and we really do not want those dropping the average way down.  I think that's why you're special casing 0 values -- otherwise the geomean blows up.  But a value of epsilon is almost as destructive.  I'd change to the arithmetic mean and things should be a lot more stable.  It is much less prone to going to tiny averages due to some tiny values, and as overly small base costs are the problem I think we want to stop using an average that weights small values very heavily (infinitely for 0).

vaughnbetz:
QoR regtest failures above look OK. 3 of 4 are improvements, and the last one is a 44% critical path delay degradation on a small, bidirectional architecture circuit, and that's known to be a noisy test, and on a less important architecture.

vpr/src/route/rr_graph_indexed_data.cpp
acomodi: Ok, I have changed the code to use the arithmetic mean instead

vpr/src/route/rr_graph_indexed_data.cpp
acomodi: I was unsure about this. Basically, the inv_length is taken into account in a later step of the base_cost calculation and IMO it should be taken into account based on the base_cost_type parameter value.

vpr/src/route/rr_graph_indexed_data.cpp
acomodi: This was the geo mean, but I have changed it to use the arithmetic mean instead.

vpr/src/route/rr_node.h
acomodi: No, it was a leftover. Removed

acomodi:
@vaughnbetz @litghost Titan quick QoR have finished and has been run with the current PR's status against master:
- Changes (https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1576/commits/4c55d01eb694758c0def39a3daec57f2329fe0c3)
- Baseline (https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/ab5f508db7e405925a45d1da918e6dca78730b44).

Results summary are presented in [this colab page](https://colab.research.google.com/drive/1_Vn0k-VHT-KdCPP5wYNromouPHDd3xFV?usp=sharing).

For what I see there seems to be a general degradation of CPD and Placement runtime, in favor of a reduced routing run-time. I'll investigate why this is so.

In general, would it be acceptable to add yet another parameter to switch between different rr_indexed_data (and consequently also the delay normalization) computation methods?

litghost:
I don't see geomean comparisions for the relevant QoR values?

litghost:
The geomean table should probably be transposed, as it current hard to read in it's current form.

litghost:
While the routing time is down ~20%, the overall flow time is up 6%.  Where did the time go?

litghost:
> While the routing time is down ~20%, the overall flow time is up ~6%. Where did the time go?

Place time is up ~16% :(

litghost:
So overall:
 - CPD is up 8% (bad)
 - Route wire length about the same
 - Flow time is up 6% (bad)
 - Place time is up 16% (bad)
 - Route time is down 20% (good)

I'd say this change seems pretty bad on the Titan benchmarks.  What is the latest symbiflow results?

acomodi:
@litghost SymbiFlow tests results seemed to be promising, achieving better run-time as well as CPD w.r.t. baseline. I am currently gathering more consistent data and add that to the colab page, so to display also the SymbiFlow tests results.

I think that the issue might be related to the missing inv_length during the delay_normalization, as you have noticed in one of the review comments. Investigating whether this is the case.

vaughnbetz:
Yes, this does seem good to me too.  I'm a bit baffled as to why placement time would go up though -- did the placement delay matrix computation get slower?  Otherwise I wouldn't expect placement to be affected.  If the placement delay matrix computation slowed down we may be able to claw it back as it seems like it wouldn't be fundamental.  In any case placement delay matrix can be cached, so it could be saved for some flows.  Can you parse out the place time components to see where the slowdown comes from.

acomodi:
@vaughnbetz I think I cannot access the log files from the hydra runs, but I have run locally some of the smaller designs and indeed, the place_delay matrix computation has increased:

**cholesky_mc benchmark**:

Baseline:
- Lookahead time: 78.59
- Place delay matrix: 42.18
- Place time: 169.10

Changes:
- Lookahead time: 78.04
- Place delay matrix: 117.00
- Place time: 242.40

vaughnbetz:
Thanks.  That's really strange.  I can't figure out why a different base cost normalization would slow down the place delay matrix calculation.  Presumably the routing paths being explored are different, but still not sure how that leads to this behaviour. Probably worth double-checking that any data (rr_indexed_data?) needed by the place_delay_matrix is properly loaded, and printing out the base costs of all resoures to make sure none have funny data (should just all be scaled up or down some, uniformly) at the point where the place delay matrix is calculated.

acomodi:
@vaughnbetz The default for the `place_delay_matrix` calculation is set to `astar` by default. This means that the delay calculation method is based on the router, which uses, by default, the lookahead computed at the previous step.
I think the delay matrix runtime is affected as the lookahead itself got computed with the new base_cost values.

I have double checked that the base costs in the indexed data do have no funny values prior to calculating the delay matrix

vaughnbetz:
Thanks.  The astar search using a high criticality so it is looking for low delay paths.  So the base cost shouldn't really matter a lot to it.  So something odd is happening to slow it down so much.

acomodi:
I think I understand where the problem is. https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/37095c742471af2e2c8d90faf10ab13156a47942/vpr/src/route/router_lookahead_map.cpp#L167

The code snippet above is in the lookahead_map, and the `delay` cost of the Cost_Entry is taken from the rr_indexed_data. Basically, with the changes in this PR, this delay cost gets changed as well in the lookahead.

I think we should switch to use the delay computation used for instance in the extended_lookahead_map: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/37095c742471af2e2c8d90faf10ab13156a47942/vpr/src/route/router_lookahead_map_utils.cpp#L107-L132

litghost:
> I think we should switch to use the delay computation used for instance in the extended_lookahead_map:

I agree this should help.

acomodi:
@litghost, @vaughnbetz I got some nice table printed out in the echo files regarding RR indexed data:

**stereo_vision** titan design:
Baseline:
```
Delay normalization: 3.28462e-11
Cost Index                    Base Cost           Ortho Cost Index    Seg Index           Inv. Length         T. Linear           T. Quadratic        C. Load
0 SOURCE                      3.28462e-11         -1                  -1                  nan                 0                   0                   0
1 SINK                        0                   -1                  -1                  nan                 0                   0                   0
2 OPIN                        3.28462e-11         -1                  -1                  nan                 0                   0                   0
3 IPIN                        3.12039e-11         -1                  -1                  nan                 0                   0                   0
4 CHANX L4                    1.31385e-10         6                   0                   0.25                1.3917e-10          0                   0
5 CHANX L16                   5.25539e-10         7                   1                   0.0625              3.25101e-10         0                   0
6 CHANY L4                    1.31385e-10         4                   0                   0.25                1.39018e-10         0                   0
7 CHANY L16                   5.25539e-10         5                   1                   0.0625              3.25098e-10         0                   0
```

Changes:
```
Delay normalization: 7.7409e-11
Cost Index                    Base Cost           Ortho Cost Index    Seg Index           Inv. Length         T. Linear           T. Quadratic        C. Load
0 SOURCE                      7.7409e-11          -1                  -1                  nan                 0                   0                   0
1 SINK                        0                   -1                  -1                  nan                 0                   0                   0
2 OPIN                        7.7409e-11          -1                  -1                  nan                 0                   0                   0
3 IPIN                        7.35386e-11         -1                  -1                  nan                 0                   0                   0
4 CHANX L4                    3.09636e-10         6                   0                   0.25                1.36732e-10         0                   0                   
5 CHANX L16                   1.23854e-09         7                   1                   0.0625              2.86368e-10         0                   0                   
6 CHANY L4                    3.09636e-10         4                   0                   0.25                1.35758e-10         0                   0                   
7 CHANY L16                   1.23854e-09         5                   1                   0.0625              2.75063e-10         0                   0    
```

Delay normalization seems to be almost double w.r.t baseline, indicating that there might be something wrong with the delay_normalization_factor calculation, as the T_Linear data doesn't seem to vary by a huge factor.

vaughnbetz:
Thanks. 
T_linear also drops a little for the L4s (~2%) and quite a bit (~15%) for the L16s. That lines up with Keith's theory that we're getting a lower T_linear and that's effectively reducing Astar_fac / causing back-tracking while creating the placer delay matrix.  Not necessarily wrong if it leads to higher quality placements; we could run the new code with old placements to see if the cpd gain is coming from the placer or the router.

acomodi:
> we could run the new code with old placements to see if the cpd gain is coming from the placer or the router.

Sure, makes sense. I can test this.

Moreover, I fixed a potential bug in my code that resulted higher values of the delay normalization factor then they should be. I have removed the clb_dist constant. 
For what I understand, it should only affect the T_quadratic value of the delay, but I do not get why `3` would be the correct value for this constant.
Also, for what I have seen, there is a negligible amount of cost indices with T_quadratic, making this constant useless?

By removing the constant in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1576/commits/3c9164376bca7144cdbad95a43f3ecc52447fa08 the delay value obviously get three times smaller, and this, in consequence, is making also SymbiFlow tests run slower.

I am running once again the Titan benchmarks to get more data.

On a side note, using the switch delay information during the lookahead map generation resulted in a [test failure](https://hydra.vtr.tools/build/41039)

vaughnbetz:
Tquadratic is nonzero only for unbuffered switches that have a finite resistance, which is generally pass transistors.  These are less commonly used these days, so you are right that Tquadratic is 0 for all wire segments for most architectures these days.

acomodi:
Updated results came back from the Hydra builds and I have summed them up in the same [colab page](https://colab.research.google.com/drive/1_Vn0k-VHT-KdCPP5wYNromouPHDd3xFV?usp=sharing).

With https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1576/commits/e747b4a3e25bebc409d6e3d9d7d6666957e9f82a I have removed the clb_dist factor when calculating the delay_normalization factor and so fixed a bug in the previous code that resulted in the delay normalization being 3x the value it should have had. This resulted in an increase of the vtr_flow run-time by 11% w.r.t. baseline and a ~3.8% improvement in CPD.

Given that the only change was related to the delay_normalization_factor, and the fact that with higher base costs value resulted in better run-time as well as CPD for both symbiflow and titan tests, I have run also a test for which I have doubled the delay_cost_normalization. This indeed resulted in a run-time improvement ending up in a 4.5% increase in the vtr_flow run-time w.r.t baseline, with a CPD improvement of ~2.2%.

One more note to add is that the Kokoro Nightly tests showed a deterioration of CPD for some of the tests that instead showed an improvement in the Hydra runs. The only difference between the two builds, as far as I know, is the compiler version:
- Hydra builds: GNU 9.2.0
- Kokoro Nightly: GNU 5.4.0

I am thinking whether it would be worth adding a flag to control the indexed_data calculation method (and consequently also the delay_normalization_factor), even though this is less than ideal.

vaughnbetz:
These do look like good results.  I think the doubled delay_cost_normalization will be very close (within numerical noise) to just using a doubled initial_pres_cost so I don't think it would be worth bringing out this option to the command line. You can run the latest code with a doubled initial_pres_cost to see if it's roughly equivalent to the doubled delay_cost_normalization.

vaughnbetz:
Is the placement time increase still due to the placement_delay_matrix calculation time?

litghost:
@acomodi Can I make a couple requests:

 - The colab presentation of the data from the Titan benchmarks is hard to read.  Can you please reformat the table to be transposed like the symbiflow results are
 - What is the current state of experiments?  The colab page shows 3 rows, but it is unclear what exactly those 3 rows represent.
 - What are your next steps to closing this?

acomodi:
@vaughnbetz Ok, I am getting all the data from the Hydra runs, and double check where the placement time is gone. 
On the re-running the tests, I believe it should be the other way around: the initial pres factor should be halved to get roughly the same results as for the doubled delay normalization, correct? (the latest code in this PR includes the doubled delay normalization factor)

@litghost 

> The colab presentation of the data from the Titan benchmarks is hard to read. Can you please reformat the table to be transposed like the symbiflow results are

This actually made results quite more unreadable as well (as there are many metrics in there). I have instead reduced the tests in a dropdown widget, so that you can select one circuit at the time, making it easier to read.

> What is the current state of experiments? The colab page shows 3 rows, but it is unclear what exactly those 3 rows represent.

I have added a more detailed description on top of the code
Basically, we have two experiments:
1. robust RR indexed data calculation with all bug fixes I removed
2. same as 1 with a doubled delay_normalization_factor

> What are your next steps to closing this?

First I think I should verify where the actual extra time is spent. In case the whole time increase is in the place_delay_matrix and lookahead computation, I might try to add multi-thread capabilities to those steps as it happens for the extended lookahead.

vaughnbetz:
Not sure we're using the same numerator and denominator :).
Since cong_cost is proportional to base_cost * pres_fac_mult, doubling the delay_normalizer (which increases all base_costs) is roughly the same as doubling initial_pres_fac.  So while the delay_normalizer is showing a useful CPU/quality trade-off, it is one that we should already be able to explore from the command line with initial_pres_fac, so I don't think it will be necessary to bring this new knob out to the command line.  Hopefully we can pick a reasonable default, and then those who want a different trade-off can use initial_pres_fac.

acomodi:
@vaughnbetz All right, all clear.
The command line flag I was referring to, was actually intended to switch between this PR's indexed data calculation method and the baseline's one. Mainly because, despite the delay_normalization (which, I agree, shouldn't be adjustable with a flag), the `T_linear` values have changed w.r.t. baseline, causing the increase in `place_delay_matrix` run-time.

vaughnbetz:
Got it.  That's the one I really hope to run to ground -- hopefully we can figure out what's going on there, and if we have a quality/CPU trade-off there or are just estimating a little too low and wasting CPU.

Another thing we could do to speed up the place_delay matrix is to switch for repeated astar invocations to a Dijkstra flood more like the lookahead calculation.  Kevin tried that a while ago and saw a moderate QoR loss so he didn't turn it on.  I don't thinkt he QoR loss should be fundamental though -- we should be able to explore the routing graph more efficiently and/or leverage the lookahead we've already built so we don't pay this CPU time.  I'd explore that before multi-threading the current algorithm as I think we're being fairly inefficient in creating the place delay matrix.

acomodi:
As discussed, I have computed some histograms relative to the various properties of nodes/switches during the indexed data calculation. Here is an example of L16 wire values:

```
COST index: 5   SEGMENT: L16
R_TOTAL HISTOGRAM
[       50:  1.3e+02)   952 (  2.8%) |**
[  1.3e+02:    2e+02)   948 (  2.8%) |**
[    2e+02:  2.8e+02)   474 (  1.4%) |*
[  2.8e+02:  3.5e+02)   950 (  2.8%) |**
[  3.5e+02:  4.3e+02)   476 (  1.4%) |*
[  4.3e+02:    5e+02)   948 (  2.8%) |**
[    5e+02:  5.8e+02)   474 (  1.4%) |*
[  5.8e+02:  6.6e+02)   950 (  2.8%) |**
[  6.6e+02:  7.3e+02)   476 (  1.4%) |*
[  7.3e+02:  8.1e+02) 27076 ( 80.3%) |*********************************************


C_TOTAL HISTOGRAM
[  3.6e-14:  6.8e-14)   952 (  2.8%) |**
[  6.8e-14:  9.9e-14)   948 (  2.8%) |**
[  9.9e-14:  1.3e-13)   474 (  1.4%) |*
[  1.3e-13:  1.6e-13)   526 (  1.6%) |*
[  1.6e-13:  1.9e-13)   900 (  2.7%) |*
[  1.9e-13:  2.3e-13)   543 (  1.6%) |*
[  2.3e-13:  2.6e-13)   879 (  2.6%) |*
[  2.6e-13:  2.9e-13)   560 (  1.7%) |*
[  2.9e-13:  3.2e-13)   866 (  2.6%) |*
[  3.2e-13:  3.5e-13) 27076 ( 80.3%) |*********************************************


siwtch_R_TOTAL HISTOGRAM
[    2e+02:  2.2e+02)    36 (  0.1%) |
[  2.2e+02:  2.5e+02)   111 (  0.3%) |
[  2.5e+02:  2.8e+02)   726 (  2.2%) |**
[  2.8e+02:    3e+02)  8921 ( 26.5%) |*****************************
[    3e+02:  3.2e+02) 13760 ( 40.8%) |*********************************************
[  3.2e+02:  3.5e+02)  8554 ( 25.4%) |****************************
[  3.5e+02:  3.8e+02)   786 (  2.3%) |***
[  3.8e+02:    4e+02)   108 (  0.3%) |
[    4e+02:  4.2e+02)   677 (  2.0%) |**
[  4.2e+02:  4.5e+02)    45 (  0.1%) |


switch_T_TOTAL HISTOGRAM
[  5.9e-11:  6.1e-11)    45 (  0.1%) |
[  6.1e-11:  6.4e-11)   693 (  2.1%) |**
[  6.4e-11:  6.6e-11)   207 (  0.6%) |*
[  6.6e-11:  6.8e-11)  2066 (  6.1%) |*****
[  6.8e-11:  7.1e-11)  7526 ( 22.3%) |*******************
[  7.1e-11:  7.3e-11) 17914 ( 53.1%) |*********************************************
[  7.3e-11:  7.5e-11)  4457 ( 13.2%) |***********
[  7.5e-11:  7.8e-11)   765 (  2.3%) |**
[  7.8e-11:    8e-11)    38 (  0.1%) |
[    8e-11:  8.2e-11)    13 (  0.0%) |


switch_Cinternal_TOTAL HISTOGRAM
[        0:        0) 33724 (100.0%) |*********************************************
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |

```

litghost:
> @litghost @vaughnbetz @kmurray As discussed, I have computed some histograms relative to the various properties of nodes/switches during the indexed data calculation. Here is an example of L16 wire values:

Did I miss something, or should there also be a T_linear histogram?

acomodi:
@litghost So, T linear is built with the following function (for non-buffered nodes):

```
rr_indexed_data[cost_index].T_linear = Tsw + Rsw * (Cinternalsw + Cnode) + 0.5 * Rnode * (Cnode + Cinternalsw);                                                                                                                                                                                         
```

All the components of this equation come from averaging all the values found for each node of the same cost index.

acomodi:
By taking the most common bucket of the histogram and averaging between its min/max values I get for the stereo_vision titan design the following indexed data:

```
Delay normalization factor: 2.67516e-11
Cost Index                    Base Cost           Ortho Cost Index    Seg Index           Inv. Length         T. Linear           T. Quadratic        C. Load
0 SOURCE                      2.67516e-11         -1                  -1                  nan                 0                   0                   0
1 SINK                        0                   -1                  -1                  nan                 0                   0                   0
2 OPIN                        2.67516e-11         -1                  -1                  nan                 0                   0                   0
3 IPIN                        2.54141e-11         -1                  -1                  nan                 0                   0                   0
4 CHANX L4                    1.07007e-10         6                   0                   0.25                1.36194e-10         0                   0
5 CHANX L16                   4.28026e-10         7                   1                   0.0625              3.07135e-10         0                   0
6 CHANY L4                    1.07007e-10         4                   0                   0.25                1.38505e-10         0                   0
7 CHANY L16                   4.28026e-10         5                   1                   0.0625              3.06175e-10         0                   0

```

For reference, these were previous values:

Baseline:
```
Delay normalization: 3.28462e-11
Cost Index                    Base Cost           Ortho Cost Index    Seg Index           Inv. Length         T. Linear           T. Quadratic        C. Load
0 SOURCE                      3.28462e-11         -1                  -1                  nan                 0                   0                   0
1 SINK                        0                   -1                  -1                  nan                 0                   0                   0
2 OPIN                        3.28462e-11         -1                  -1                  nan                 0                   0                   0
3 IPIN                        3.12039e-11         -1                  -1                  nan                 0                   0                   0
4 CHANX L4                    1.31385e-10         6                   0                   0.25                1.3917e-10          0                   0
5 CHANX L16                   5.25539e-10         7                   1                   0.0625              3.25101e-10         0                   0
6 CHANY L4                    1.31385e-10         4                   0                   0.25                1.39018e-10         0                   0
7 CHANY L16                   5.25539e-10         5                   1                   0.0625              3.25098e-10         0                   0
```

New indexed data calculation with averaging:
```
Delay normalization: 7.7409e-11
Cost Index                    Base Cost           Ortho Cost Index    Seg Index           Inv. Length         T. Linear           T. Quadratic        C. Load
0 SOURCE                      7.7409e-11          -1                  -1                  nan                 0                   0                   0
1 SINK                        0                   -1                  -1                  nan                 0                   0                   0
2 OPIN                        7.7409e-11          -1                  -1                  nan                 0                   0                   0
3 IPIN                        7.35386e-11         -1                  -1                  nan                 0                   0                   0
4 CHANX L4                    3.09636e-10         6                   0                   0.25                1.36732e-10         0                   0                   
5 CHANX L16                   1.23854e-09         7                   1                   0.0625              2.86368e-10         0                   0                   
6 CHANY L4                    3.09636e-10         4                   0                   0.25                1.35758e-10         0                   0                   
7 CHANY L16                   1.23854e-09         5                   1                   0.0625              2.75063e-10         0                   0    
```

vaughnbetz:
Thanks.  Most common / median values look more reasonable for Tlinear etc.
The variation in Rtotal and Ctotal make sense -- cut off wires.

Why does Rswitch vary, and why does Tswitch vary?  I would have thought we'd be driving the L16's with the same switch all the time.

Here are the values from the arch file.  All the L16 drivers should have R = 150 and Tsw = 87 ps.  (second type of switch).  The L4 drivers are also listed below.

    <switch type="mux" name="seg4_driver" R="450" Cin="0.60e-15" Cout="4.82e-15" Tdel="59e-12" mux_trans_size="2.630740" buf_size="27.645901"/>
    
    <switch type="mux" name="seg16_driver" R="150" Cin="1.80e-15" Cout="14.5e-15" Tdel="87e-12" mux_trans_size="2.630740" buf_size="27.645901"/>

Instead, Tsw and Rsw show variations, and their means are halfway between the L4 and L16 values.  Are you averaging over all the fanout switches from a node?  All its fan-in switches?  Averaging over its fanin switches is probably better, as they really represent the delay of *this* node.  If you average over fan-in switches only, I think you'll get much tigher Tsw and Rsw distributions and they'll converge to the seg16_driver values for the L16 wires.

acomodi:
@vaughnbetz the relevant code is here to calculate the average switches of a node: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/5e2b7eab3809f946ed57b648a4187986096cfdf9/vpr/src/route/rr_graph_indexed_data.cpp#L415-L451

I am unsure though whether the considered edges correspond to the node's `fanout` or `fanin`.

vaughnbetz:
Thanks. That is averaging over the fan out of inode. 

There is an rr graph function to generate the fanin lists; we don’t need them as much so they are generated on demand for all nodes in an O(n) operation and should be destroyed after you’re done with them.

acomodi:
@vaughnbetz All right. I have looked for this function, but I am unable to find it, unfortunately. If you know where this function is, can you please point me to that?

vaughnbetz:
You're right; it looks like this utility function doesn't exist (or is very well hidden).  I think Xifan wrote a version of it in an in-progress rr-graph refactoring: adding him and Keith.  @tangxifan @litghost 
The code to create this would be similar to 
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/5bded3ed00b3c010770a8528dd84ccf991fd8d1b/vpr/src/route/rr_graph_storage.cpp#L384
which loads a fanin count for each switch.  Instead of storing a fan_in count for each rr_node, it would store a vector of input switch ids for each rr_node (so push the edge id index into a vector, where there is a different vector for every dest_node id).

This would be a useful utility function to have.  It should have a destroy_ function though so we can free the memory separately from the rest of the rr_graph as it isn't needed most of the time.

acomodi:
@vaughnbetz I have added the fan-in generation to this PR as well, and the histogram generation.

By using the fan-in instead of fan-out, the following histograms are generated:
```
Cost Index: 7   Segment info: L16
        Node R Total
[       50:  1.3e+02)  1280 (  3.7%) |**
[  1.3e+02:    2e+02)  1280 (  3.7%) |**
[    2e+02:  2.8e+02)   640 (  1.8%) |*
[  2.8e+02:  3.5e+02)  1280 (  3.7%) |**
[  3.5e+02:  4.3e+02)   640 (  1.8%) |*
[  4.3e+02:    5e+02)  1280 (  3.7%) |**
[    5e+02:  5.8e+02)   640 (  1.8%) |*
[  5.8e+02:  6.6e+02)  1280 (  3.7%) |**
[  6.6e+02:  7.3e+02)   640 (  1.8%) |*
[  7.3e+02:  8.1e+02) 25920 ( 74.3%) |*********************************************

        Node C Total
[  3.6e-14:  6.8e-14)  1280 (  3.7%) |**
[  6.8e-14:  9.9e-14)  1280 (  3.7%) |**
[  9.9e-14:  1.3e-13)   640 (  1.8%) |*
[  1.3e-13:  1.6e-13)   707 (  2.0%) |*
[  1.6e-13:  1.9e-13)  1213 (  3.5%) |**
[  1.9e-13:  2.3e-13)   748 (  2.1%) |*
[  2.3e-13:  2.6e-13)  1172 (  3.4%) |**
[  2.6e-13:  2.9e-13)   765 (  2.2%) |*
[  2.9e-13:  3.2e-13)  1155 (  3.3%) |**
[  3.2e-13:  3.5e-13) 25920 ( 74.3%) |*********************************************

        Switch R Total
[  1.5e+02:  1.5e+02) 34880 (100.0%) |*********************************************
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |
[  1.5e+02:  1.5e+02)     0 (  0.0%) |

        Switch T Total
[  8.7e-11:  8.7e-11) 34880 (100.0%) |*********************************************
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |
[  8.7e-11:  8.7e-11)     0 (  0.0%) |

        Switch Cinternal Total
[        0:        0) 34880 (100.0%) |*********************************************
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |
[        0:        0)     0 (  0.0%) |

```

The switches seem to have correct single values, corresponding to those defined in the arch definition.

The problem is that these values now produce even lower T_linear data:
```
Delay normalization factor: 2.66461e-11
Cost Index                    Base Cost           Ortho Cost Index    Seg Index           Inv. Length         T. Linear           T. Quadratic        C. Load
0 SOURCE                      2.66461e-11         -1                  -1                  nan                 0                   0                   0
1 SINK                        0                   -1                  -1                  nan                 0                   0                   0
2 OPIN                        2.66461e-11         -1                  -1                  nan                 0                   0                   0
3 IPIN                        2.53138e-11         -1                  -1                  nan                 0                   0                   0
4 CHANX L4                    1.06584e-10         6                   0                   0.25                1.41435e-10         0                   0
5 CHANX L16                   4.26337e-10         7                   1                   0.0625              2.82613e-10         0                   0
6 CHANY L4                    1.06584e-10         4                   0                   0.25                1.43596e-10         0                   0
7 CHANY L16                   4.26337e-10         5                   1                   0.0625              2.82613e-10         0                   0
```

vaughnbetz:
That makes sense for L16's, as the more powerful switches drive them.  We should still take median R & C for each wire type as that will get the wires that aren't cut off.

I think then it becomes a matter of tuning:
1.  astar_fac during place_delay_matrix calculation.  If the Tlinear values are correct (and they look correct to me now) then we just want to tune how much time to spend exploring looking for good L16 paths vs. accepting a slower path.
2. an extra guardband on the estimated delay in the place_delay_matrix to move us from best delay to "average" or "almost guaranteed" delay.  That can be done more understandably with a modification like:  
  place_delay_est(dx,dy) = place_delay_est_found_now * mult_guardband + constant_guardband
  (e.g. constant_guardband = 0, mult_guardband = 1.2)

Longer term, I think Djikstra + optimism reduction (part 2 above)  is likely to provide a superior combination of cpu time to calculate place delay matrix values and placement optimization.

vpr/src/base/read_options.cpp
```diff
@@ -2000,11 +2000,19 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
...
-            "Controls the directedness of the the timing-driven router's exploration."
+            "Controls the directedness of the timing-driven router's exploration."
             " Values between 1 and 2 are resonable; higher values trade some quality for reduced run-time")
         .default_value("1.2")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument(args.router_profiler_astar_fac, "--router_profiler_astar_fac")
+        .help(
+            "Controls the directedness of the timing-driven router's exploration"
+            " when doing router profiling."
```
vaughnbetz, line 2011: I think you should explain what "router profiling" means.  Does this mean during calculation of point-to-point delays for use during placement delay estimation & optimization?  Does it mean calculation of terms used by the router lookahead?  Does it mean both?

vaughnbetz (changes requested):

vpr/src/base/read_options.cpp
acomodi: Sure, makes sense.
One thing I see is that the RouterDelayProfiler is used only in the timing_place_lookup to generate the place delay matrix actually. I am unsure if its original purpose was to serve as a general interface to run the router, or if it was intended only for the delay matrix.

vaughnbetz: I think we only ever used it for the place delay matrix.  One day there is probably some code that can be unified between the lookahead builder and the place delay matrix (use a single Djikstra expansion and get all the info perhaps ...).  But for now just saying what we use this parameter for would be good.

acomodi:
@litghost @vaughnbetz The latest runs completed. I have separated them in the following colab pages:
- [VTR reg QoR chain tests](https://colab.research.google.com/drive/1BicA1YxnnPNSNct-pddYe34eacmXCo8D?usp=sharing)
- [Titan quick QoR](https://colab.research.google.com/drive/1_s55wt59d60Smj18Z4pFBpf6gmpQh72w?usp=sharing)

The best results are obtained with the following settings:
- Higher a_star during the place_delay_matrix calculation: this parameter is adjustable, and I think that a lower value can yield good enough results as well, probably achieving better CPD estimates and final CPDs.
- Doubling of `initial_pres_fac`.
- New indexed data calculation method.

vaughnbetz:
Thanks @acomodi .
Just to be check: the circuit name in the tables is not meaningful (all values are averages over multiple circuits), correct?

Titan results show all 3 changes are beneficial.
VTR results show the new normalization is beneficial (slightly lower Wmin, faster runtime).  Slightly worse critical path delay, but that could be due to the smaller Wmin (126 vs. 130).  Flow runtime is also down.
VTR results show the higher astar_fac and double pres_fac are not beneficial; higher Wmin, no improvement in runtime or cpd vs. the new normalization.  They aren't too far off though (Wmin = 136 instead of the default 130 with all 3 changes, and cpd is up ~2%).

So the new normalization should go in.  Higher astar_fac and doubled pres_fac is beneficial for Titan, may be mildly negative for VTR.  Could set different defaults for those?

acomodi:
> So the new normalization should go in. Higher astar_fac and doubled pres_fac is beneficial for Titan, may be mildly negative for VTR. Could set different defaults for those?

@vaughnbetz Ok. Just to be sure, you mean changing the default router_delay_profiler astar_fac and the pres_fac to get VTR tests in line?

> Just to be check: the circuit name in the tables is not meaningful (all values are averages over multiple circuits), correct?

So, there are two different tables:
- the first one lets you select the various circuits' results, to have some more detailed information.
- the last one is a collection of normalized data of all the tests to see the overall trends.

I have also double-checked that, with the current implementation, some of strong tests do not meet QoR, but, by lowering the initial_pres_fac (where the initial_pres_fac default is 1.0). Would it be better to update the golden results, or change the initial_pres_fac for the failing strong tests?

vaughnbetz:
Ah, got it, thanks.

Results are good for all 3 changes for Titan (every change improves things).
New normalization scheme is a slight improvement for VTR; higher astar_fac during place_delay_matrix and higher pres_fac are slight degradations.  So I'd recommend having Titan run with the higher astar_fac and pres_fac (set at command line), but leaving the defaults for both alone so VTR uses the old values.  Hopefully that resolves the QoR failures too (although there might still be some random ones).

acomodi:
Ok, I will run another set of tests on Hydra, to get the final results.

In the meantime I collected some SymbiFlow tests with and without the changes in this PR. The results are collected [here](https://docs.google.com/spreadsheets/d/1Hb0aAtqDwnnNlkozWQ_TGE3EEw0eCABwITZUKYHwRQ4/edit?usp=sharing)

NOTE:
- the changes tests seem to have way too high run-times, probably due to the fact that there might have been some additional loads on my local machine which slowed everything down. I will re-run everything tomorrow and update the results table.
- tests have been run on the 50T and 100T devices

@litghost FYI

acomodi:
@vaughnbetz Final runs on Hydra have completed, more details in the colab pages linked below:

- [VTR reg QoR chain](https://colab.research.google.com/drive/1BicA1YxnnPNSNct-pddYe34eacmXCo8D?usp=sharing)
- [Titan quick QoR](https://colab.research.google.com/drive/1_Vn0k-VHT-KdCPP5wYNromouPHDd3xFV?usp=sharing)

vaughnbetz:
VTR and Titan results look good.  Symbiflow results look OK to me too, but I haven't analyzed them as much / am not as familar with them.

vpr/src/route/rr_graph_indexed_data.cpp
```diff
@@ -264,119 +260,104 @@ static std::vector<size_t> count_rr_segment_types() {
...
+    return delay_norm_fac;
 }
 
-static void load_rr_indexed_data_T_values(int index_start,
-                                          int num_indices_to_load,
-                                          t_rr_type rr_type,
-                                          int nodes_per_chan,
-                                          const t_rr_node_indices& L_rr_node_indices) {
+static void load_rr_indexed_data_T_values() {
     /* Loads the average propagation times through segments of each index type   *
```
vaughnbetz, line 296: This comment looks out of date.  I think now you're going through all rr_nodes and getting medians for their delay. Good to update this comment and check for other out-of-date comments. Should also describe what we're trying to do here (get typical values of T_linear etc. for different wire types, which are used for things like constructing lookaheads and place delay matrices).

vpr/src/route/rr_graph_indexed_data.cpp
```diff
@@ -264,119 +260,104 @@ static std::vector<size_t> count_rr_segment_types() {
...
+    auto fan_in_list = get_fan_in_list();
 
-    num_nodes_of_index = (int*)vtr::calloc(device_ctx.rr_indexed_data.size(), sizeof(int));
-    C_total = (float*)vtr::calloc(device_ctx.rr_indexed_data.size(), sizeof(float));
-    R_total = (float*)vtr::calloc(device_ctx.rr_indexed_data.size(), sizeof(float));
+    std::vector<int> num_nodes_of_index(rr_indexed_data.size(), 0);
+    std::vector<std::vector<float>> C_total(rr_indexed_data.size());
+    std::vector<std::vector<float>> R_total(rr_indexed_data.size());
 
     /* August 2014: Not all wire-to-wire switches connecting from some wire segment will
```
vaughnbetz, line 313: Again, I think we're averaging / doing a median over all rr_nodes, not just some in one chosen channel. Should update the comment.

vpr/src/route/rr_graph_indexed_data.cpp
```diff
@@ -453,44 +409,41 @@ static void load_rr_indexed_data_T_values(int index_start,
                 // the combined capacitance of the node and internal capacitance of the switch. The
                 // second transient response is the result of the Rnode being distributed halfway along a
```
vaughnbetz, line 410: Change "second transient response" to "multiplication by the second term by 0.5"

vpr/src/route/rr_graph_indexed_data.cpp
```diff
@@ -453,44 +409,41 @@ static void load_rr_indexed_data_T_values(int index_start,
...
-    free(C_total);
-    free(R_total);
-    free(switch_R_total);
-    free(switch_T_total);
-    free(switch_Cinternal_total);
-    free(switches_buffered);
 }
 
-static void calculate_average_switch(int inode, double& avg_switch_R, double& avg_switch_T, double& avg_switch_Cinternal, int& num_switches, short& buffered) {
+static void calculate_average_switch(int inode, double& avg_switch_R, double& avg_switch_T, double& avg_switch_Cinternal, int& num_switches, short& buffered, vtr::vector<RRNodeId, std::vector<RREdgeId>>& fan_in_list) {
```
vaughnbetz, line 429: Add comment saying what this routine does, and why it's useful.

vaughnbetz (commented):
Suggest some additional comments; otherwise looks good.

litghost (approved):
LGTM.  @vaughnbetz, we good to merge?

acomodi:
We can potentially merge this and then I will open a follow-up PR with the additional detailed comments.

vaughnbetz:
Sounds good, merging.

PR #1579, opened 2020-10-29
check_rr_graph: fix bug when checking multi-connections between nodes

acomodi:

#### Description

There seems to be a bug in check_rr_graph related to the check on the multiple connections between RR nodes.

While dealing with https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1711 to fix the RR graph due to the presence of only one node for all the possible sides of a pin, I encountered the VPR error below:

```
Message: in check_rr_graph: node 279 (CHANX) connects to node 149 (IPIN) 2 times - multi-connections only expected for CHAN->CHAN.
```

Given that the comment prior to the check say that this pair of nodes can have multiple connections, I spotted a possible bug in the if statement.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/check_rr_graph.cpp
+++ b/vpr/src/route/check_rr_graph.cpp
@@ -123,17 +123,27 @@ void check_rr_graph(const t_graph_type graph_type,
 
             t_rr_type to_rr_type = device_ctx.rr_nodes[to_node].type();
 
-            /* Only expect the following cases to have multiple edges
-             * - chan <-> chan connections 
-             * - IPIN <-> chan connections (unique rr_node for IPIN nodes on multiple sides)
-             * - OPIN <-> chan connections (unique rr_node for OPIN nodes on multiple sides)
+            /* It is unusual to have more than one programmable switch (in the same direction) between a from_node and a to_node,
+             * as the duplicate switch doesn't add more routing flexibility.
+             *
+             * However, such duplicate switches can occur for some types of nodes, which we allow below.
+             * Reasons one could have duplicate switches between two nodes include:
+             *      - The two switches have different electrical characteristics.
+             *      - Wires near the edges of an FPGA are often cut off, and the stubs connected together.
+             *        A regular switch pattern could then result in one physical wire connecting multiple
+             *        times to other wires, IPINs or OPINs.
+             *
+             * Only expect the following cases to have multiple edges
+             * - CHAN <-> CHAN connections
+             * - CHAN  -> IPIN connections (unique rr_node for IPIN nodes on multiple sides)
+             * - OPIN  -> CHAN connections (unique rr_node for OPIN nodes on multiple sides)
              */
-            if (((to_rr_type != CHANX && to_rr_type != CHANY && rr_type != IPIN)
-                 || (rr_type != CHANX && rr_type != CHANY))
-                && ((to_rr_type != CHANX && to_rr_type != CHANY)
-                    || (rr_type != CHANX && rr_type != CHANY && rr_type != OPIN))) {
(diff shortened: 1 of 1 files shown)


vpr/src/route/check_rr_graph.cpp
```diff
@@ -124,16 +124,16 @@ void check_rr_graph(const t_graph_type graph_type,
             t_rr_type to_rr_type = device_ctx.rr_nodes[to_node].type();
 
             /* Only expect the following cases to have multiple edges
-             * - chan <-> chan connections 
-             * - IPIN <-> chan connections (unique rr_node for IPIN nodes on multiple sides)
-             * - OPIN <-> chan connections (unique rr_node for OPIN nodes on multiple sides)
+             * - CHAN <-> CHAN connections
+             * - CHAN  -> IPIN connections (unique rr_node for IPIN nodes on multiple sides)
+             * - OPIN  -> CHAN connections (unique rr_node for OPIN nodes on multiple sides)
```
litghost, line 129: My suggestion for clarity is to break this into many lines, e.g.:

```
bool is_chan_to_chan = ...;
bool is_chan_to_ipin = ...;
bool is_opin_to_chan = ...;
```

That will make the intention of each boolean statement clear, and easier to follow.  Then the final if statement will be very simple too!

acomodi: Sure, makes total sense. Done

vpr/src/route/check_rr_graph.cpp
```diff
@@ -124,16 +124,16 @@ void check_rr_graph(const t_graph_type graph_type,
...
              */
-            if (((to_rr_type != CHANX && to_rr_type != CHANY && rr_type != IPIN)
-                 || (rr_type != CHANX && rr_type != CHANY))
-                && ((to_rr_type != CHANX && to_rr_type != CHANY)
-                    || (rr_type != CHANX && rr_type != CHANY && rr_type != OPIN))) {
+            bool is_chan_to_chan = (rr_type == CHANX && to_rr_type == CHANY) || (rr_type == CHANY && to_rr_type == CHANX);
```
litghost, line 131: `is_chan_to_chan` should be `(rr_type == CHANX || rr_type == CHANY) && (to_rr_type == CHANX || to_rr_type == CHANY)`.  It is legal to have CHANX <-> CHANX for adjacent tracks.

acomodi: Oh right, thanks. Fixed

vpr/src/route/check_rr_graph.cpp
```diff
@@ -124,16 +124,16 @@ void check_rr_graph(const t_graph_type graph_type,
             t_rr_type to_rr_type = device_ctx.rr_nodes[to_node].type();
 
             /* Only expect the following cases to have multiple edges
```
vaughnbetz, line 126: Suggest a longer comment here.  Here's a proposal.
/* It is unusual to have more than one programmable switch (in the same direction) between a from_node and a to_node, as the duplicate switch doesn't add more routing flexibility. However, such duplicate switches can occur for some types of nodes, which we allow below. Reasons one could have duplicate switches between two nodes include:
 - The two switches have different electrical characteristics.
 - Wires near the edges of an FPGA are often cut off, and the stubs connected together. A regular switch pattern could then result in one physical wire connecting multiple times to other wires, IPINs or OPINs.
*/

tangxifan (commented):
Look clean to me.

acomodi:
All CIs went green and I have updated the comment. Merging.

vpr/src/route/check_rr_graph.cpp
acomodi: Done

PR #1585, opened 2020-11-03
Added comments and changed intersection member functions to global fu…

sfkhalid:
…nctions

This PR has some changes that were requested on the new_constraints_class PR that has been merged. The changes include adding comments and turning some class member functions into global functions.

### DIFF
--- a/vpr/src/base/vpr_constraints.cpp
+++ b/vpr/src/base/vpr_constraints.cpp
@@ -5,6 +5,10 @@ void VprConstraints::add_constrained_atom(const AtomBlockId blk_id, const Partit
 
     auto got = constrained_atoms.find(blk_id);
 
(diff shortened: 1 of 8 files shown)


vpr/src/base/vpr_constraints.cpp
```diff
@@ -5,6 +5,10 @@ void VprConstraints::add_constrained_atom(const AtomBlockId blk_id, const Partit
 
     auto got = constrained_atoms.find(blk_id);
 
+    /**
+     * Each atom can only be in one partition. If the atoms already has a partition id assigned to it,
```
vaughnbetz, line 9: If the *atom* already has a partition id assigned to it the *id* will be swithched ...

PR #1586, opened 2020-11-03
Add missing gcov library for PGO and configurable data directory

HackerFoo:
#### Description

This adds a missing library (gcov) used by the profile guided optimization configuration, and adds a new variable to specify where to store the profile data (VPR_PGO_DATA_DIR.)

#### Motivation and Context

This is to enable PGO and make it more convenient to collect profile data.

I have tested `VPR_PGO_CONFIG=prof_gen` and `VPR_PGO_CONFIG=prof_use` on my machine.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/CMakeLists.txt
+++ b/vpr/CMakeLists.txt
@@ -10,6 +10,8 @@ option(VPR_USE_SIGNAL_HANDLER "Should VPR use a signal handler to intercept sign
 set(VPR_PGO_CONFIG "none" CACHE STRING "Configure VPR Profile-Guided Optimization (PGO). prof_gen: built executable will produce profiling info, prof_use: built executable will be optimized based on generated profiling info, none: disable pgo")
 set_property(CACHE VPR_PGO_CONFIG PROPERTY STRINGS prof_gen prof_use none)
 
+set(VPR_PGO_DATA_DIR "." CACHE PATH "Where to store and retrieve PGO data")
+
 #Handle graphics setup
 set(GRAPHICS_DEFINES "")
 
@@ -143,16 +145,18 @@ endif()
 #
 set(PROF_GEN_FLAGS_TO_CHECK
     #GCC-like
-    "-fprofile-generate" #Build will generate profiling information
+    "-fprofile-generate=${VPR_PGO_DATA_DIR}" #Build will generate profiling information
     )
 set(PROF_USE_FLAGS_TO_CHECK
     #GCC-like
-    "-fprofile-use"     #Build will use previously generated profiling information to guide code optimization
+    "-fprofile-use=${VPR_PGO_DATA_DIR}"     #Build will use previously generated profiling information to guide code optimization
     "-Wmissing-profile" #Warn if the used profiling information doesn't match the source code or is missing
     )
 
 if (VPR_PGO_CONFIG STREQUAL "prof_gen")
     message(STATUS "VPR: building to generate profiling data for PGO")
+    target_link_libraries(libvpr gcov)
+    set(CMAKE_REQUIRED_LIBRARIES gcov)
     foreach(flag ${PROF_GEN_FLAGS_TO_CHECK})
         CHECK_CXX_COMPILER_FLAG(${flag} CXX_COMPILER_SUPPORTS_${flag})
         if(CXX_COMPILER_SUPPORTS_${flag})
--- a/vpr/CMakeLists.txt
+++ b/vpr/CMakeLists.txt
@@ -156,7 +156,10 @@ set(PROF_USE_FLAGS_TO_CHECK
 if (VPR_PGO_CONFIG STREQUAL "prof_gen")
     message(STATUS "VPR: building to generate profiling data for PGO")
     target_link_libraries(libvpr gcov)
+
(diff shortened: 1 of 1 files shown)


vpr/CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ option(VPR_USE_SIGNAL_HANDLER "Should VPR use a signal handler to intercept sign
 set(VPR_PGO_CONFIG "none" CACHE STRING "Configure VPR Profile-Guided Optimization (PGO). prof_gen: built executable will produce profiling info, prof_use: built executable will be optimized based on generated profiling info, none: disable pgo")
 set_property(CACHE VPR_PGO_CONFIG PROPERTY STRINGS prof_gen prof_use none)
 
+set(VPR_PGO_DATA_DIR "." CACHE PATH "Where to store and retrieve PGO data")
```
litghost, line 13: I believe this should be "${CMAKE_CURRENT_SOURCE_DIR}"?

vpr/CMakeLists.txt
```diff
@@ -143,16 +145,19 @@ endif()
...
     #GCC-like
-    "-fprofile-use"     #Build will use previously generated profiling information to guide code optimization
+    "-fprofile-use=${VPR_PGO_DATA_DIR}"     #Build will use previously generated profiling information to guide code optimization
     "-Wmissing-profile" #Warn if the used profiling information doesn't match the source code or is missing
     )
 
 if (VPR_PGO_CONFIG STREQUAL "prof_gen")
     message(STATUS "VPR: building to generate profiling data for PGO")
+    target_link_libraries(vpr gcov)
+    target_link_libraries(libvpr gcov)
```
litghost, line 159: Shouldn't need to add gcov to both `libvpr` and `vpr`.  Just adding it to `libvpr` should be enough.

What is `CMAKE_REQUIRED_LIBRARIES` also required, versus just the `target_link_libraries` call?

vpr/CMakeLists.txt
HackerFoo: The default is to put the profile data in the same place as the object files; one for each .o file, which requires a relative path. "." is the default if unspecified.

vpr/CMakeLists.txt
HackerFoo: Without `CMAKE_REQUIRED_LIBRARIES`, the flag test fails and PGO is disabled. This is the only way I could find to add `-lgcov` to the test, which should be something like `$(CC) test.c -fprofile-generate -lgcov`.

HackerFoo: I've removed `target_link_libraries(vpr gcov)`.

litghost: Is this a case where you want to use `find_library(GCOV_LIB gcov REQUIRED)`?  Is that what you are trying to accomplish with `CMAKE_REQUIRED_LIBRARIES`?

HackerFoo: No, it still fails with `find_library(GCOV_LIB gcov REQUIRED)`

```
-- VPR: building to generate profiling data for PGO
-- Performing Test CXX_COMPILER_SUPPORTS_-fprofile-generate=/tmp/pgo
-- Performing Test CXX_COMPILER_SUPPORTS_-fprofile-generate=/tmp/pgo - Failed
```
and in `CMakeError.log`:
```
g++   -DCXX_COMPILER_SUPPORTS_-fprofile-generate=/tmp/pgo    CMakeFiles/cmTC_996f9.dir/src.cxx.o  -o cmTC_996f9
ld: CMakeFiles/cmTC_996f9.dir/src.cxx.o: in function `main':
rect_call'
ld: src.cxx:(.text.startup+0x27): undefined reference to `__gcov_indirect_call_profiler_v3'
ld: src.cxx:(.text.startup+0x40): undefined reference to `__gcov_time_profiler_counter'
ld: CMakeFiles/cmTC_996f9.dir/src.cxx.o: in function `_GLOBAL__sub_I_00100_0_src.cxx':
t'
ld: CMakeFiles/cmTC_996f9.dir/src.cxx.o: in function `_GLOBAL__sub_D_00100_1_src.cxx':

ld: CMakeFiles/cmTC_996f9.dir/src.cxx.o:(.data.rel+0x20): undefined reference to `__gcov_merge_add'
ld: CMakeFiles/cmTC_996f9.dir/src.cxx.o:(.data.rel+0x58): undefined reference to `__gcov_merge_time_profile'
collect2: error: ld returned 1 exit status
make[1]: *** [CMakeFiles/cmTC_996f9.dir/build.make:87: cmTC_996f9] Error 1
```

litghost: I think I understand what is happening here that I was missing before. Can you add a comment above this `set` that this is only affecting the `CHECK_CXX_COMPILER_FLAG` below.  I think with that comment future readers will get what is going on here.

litghost (commented):
Awaiting clarification comment, otherwise good.

litghost (approved):
LGTM, merge if green

vpr/CMakeLists.txt
HackerFoo: I've added a comment to clarify.

PR #1592, opened 2020-11-12
rr_nodes: add fan-in list generator

acomodi:

#### Description

It might be useful to get only the fan-in edges of a certain destination node, and currently the RR graph APIs do not provide such a possibility.

This PR adds a fan-in list constructor and its destroyer so that the extra memory is freed once the purpose of the fan-in lists has been fulfilled.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/rr_graph_util.cpp
+++ b/vpr/src/route/rr_graph_util.cpp
@@ -177,3 +177,20 @@ void reorder_rr_graph_nodes(const t_router_opts& router_opts) {
                                std::get<2>(edge));
     });
 }
+
+vtr::vector<RRNodeId, std::vector<RREdgeId>> get_fan_in_list() {
+    auto& rr_nodes = g_vpr_ctx.device().rr_nodes;
+
+    vtr::vector<RRNodeId, std::vector<RREdgeId>> node_fan_in_list;
+
+    node_fan_in_list.resize(rr_nodes.size(), std::vector<RREdgeId>(0));
+    node_fan_in_list.shrink_to_fit();
+
+    //Walk the graph and increment fanin on all dwnstream nodes
+    rr_nodes.for_each_edge(
+        [&](RREdgeId edge, __attribute__((unused)) RRNodeId src, RRNodeId sink) {
+            node_fan_in_list[sink].push_back(edge);
+        });
+
+    return node_fan_in_list;
+}
--- a/vpr/src/route/rr_graph_util.h
+++ b/vpr/src/route/rr_graph_util.h
@@ -1,10 +1,16 @@
 #ifndef RR_GRAPH_UTIL_H
 #define RR_GRAPH_UTIL_H
 
+#include "vpr_types.h"
+
 int seg_index_of_cblock(t_rr_type from_rr_type, int to_node);
 
 int seg_index_of_sblock(int from_node, int to_node);
 
 void reorder_rr_graph_nodes(const t_router_opts& router_opts);
+
+// This function generates and returns a vector indexed by RRNodeId
+// containing a list of fan-in edges for each node.
+vtr::vector<RRNodeId, std::vector<RREdgeId>> get_fan_in_list();
 
 #endif


vpr/src/route/rr_graph_storage.cpp
```diff
@@ -393,6 +393,24 @@ void t_rr_graph_storage::init_fan_in() {
...
 }
 
+void t_rr_graph_storage::init_fan_in_list() {
+    //Reset all fan-ins to zero
+    edges_read_ = true;
+    node_fan_in_list_.resize(node_storage_.size(), std::vector<RREdgeId>(0));
+    node_fan_in_list_.shrink_to_fit();
+
+    //Walk the graph and increment fanin on all dwnstream nodes
+    for (const auto& edge_id : edge_dest_node_.keys()) {
```
vaughnbetz, line 403: If node_fan_in (count of incoming edges) has already been loaded (should have been since we don't appear to have a separate free_ call for it) then you can make this a bit more memory-efficient by getting the right size of vector storage by doing something like the below before the main edge loop that loads the vectors.
```
 for (const auto& edge_id : edge_dest_node_.keys()) {
       node_fan_in_list_[edge_id].reserve(node_fan_in[edge_id];
   }
```

vaughnbetz (changes requested):
Looks good.  Just have a few suggestions (most important of which is a not-loaded check).
I think we should also add a test of this to a unit test if possible (I think we have some rr-graph unit tests already, which would make this pretty easy to add).

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -393,6 +393,24 @@ void t_rr_graph_storage::init_fan_in() {
     }
 }
 
+void t_rr_graph_storage::init_fan_in_list() {
```
litghost, line 396: Because this can be computed using the public interface, just make this a utility class that lives outside of `t_rr_graph_storage` completely.

vaughnbetz: Good idea.

acomodi: I do not think this can be computed using the public interface (unless I am missing something). This method requires the `edge_dest_node_` vector, which is private to the graph storage

litghost: `for_each_edge` should work here?

vpr/src/route/rr_graph_util.h
```diff
@@ -1,10 +1,31 @@
...
 int seg_index_of_cblock(t_rr_type from_rr_type, int to_node);
 
 int seg_index_of_sblock(int from_node, int to_node);
 
 void reorder_rr_graph_nodes(const t_router_opts& router_opts);
 
+// t_rr_graph_util is a utility class that collects utility
+// methods that do not strictly belong to the RR graph storage,
+// as they can use its public interface.
+class t_rr_graph_util {
```
litghost, line 15: You don't need a class or destructor.  Just make a function that returns the vector?

acomodi: Sure, I just though that it might be possible to populate this class with additional utilities that might be necessary in the future, so to keep everything well encapsulated.

Maybe it is an overkill for now and the function is good enough.

acomodi: Done

vpr/src/route/rr_graph_util.cpp
```diff
@@ -2,14 +2,15 @@
...
+#include "rr_graph_util.h"
+
 #include "vtr_memory.h"
 #include "vtr_time.h"
 
 #include "vpr_types.h"
 #include "vpr_error.h"
 
 #include "globals.h"
-#include "rr_graph_util.h"
```
litghost, line 12: Why did you move this?  If there is an include order problem that this solved, change the header order does not solving the underlying problem.

acomodi: Ah, leftover, while implementing this I bumped in some external library linking, but this was not the problem, I'll revert this change

litghost: I'm aluding to the fact that you may have discovered a problem in the header files.  What exactly did you run into?

acomodi: The issue was actually unrelated to the header files. It was a missing semicolon (typical :) )while implementing the rr_graph_util class, part of the previous code. This missing semicolon was producing this error:

```
In file included from /data/tmp/vtr-verilog-to-routing/vpr/src/route/rr_graph2.h:8:0,
                 from /data/tmp/vtr-verilog-to-routing/vpr/src/route/rr_graph_storage.h:8,
                 from /data/tmp/vtr-verilog-to-routing/vpr/src/base/vpr_context.h:12,
                 from /data/tmp/vtr-verilog-to-routing/vpr/src/base/globals.h:8,
                 from /data/tmp/vtr-verilog-to-routing/vpr/src/base/CheckSetup.cpp:5:
/data/tmp/vtr-verilog-to-routing/vpr/src/route/rr_types.h:6:57: error: invalid declarator before ‘t_pin_to_track_lookup’
 typedef std::vector<vtr::NdMatrix<std::vector<int>, 4>> t_pin_to_track_lookup;
```

litghost: Very good.

litghost (approved):
LGTM

vaughnbetz:
Looks good to me. Merging.

Issue #1587, opened 2020-11-05
The need on a Re-arrangement for the APIs in vpr_utils.h/cpp


tangxifan:
#### Current Behaviour

Currently, there are many APIs in [vpr_util.cpp](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/util/vpr_utils.cpp) but are not categorized well.
Based on the input argument and return values, the APIs can be classified into two categories:
- Pure device-level inquiry: where input arguments and return values are all device-level parameters (from the architecture definition). These functions could be just APIs in the [libarchfpga ](https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/libs/libarchfpga) library rather than in VPR, because the APIs may be called outside VPR.
- Device-level inquiry based on mapping results: where input arguments include both device-level parameters and VPR's mapping results. These functions should be part of the APIs that VPR is offering.

#### Expected Behaviour

-  APIs in libarchfpga: these functions should be made for device-level inquiry **without** VPR's mapping results.
-  APIs in VPR: these functions should be made for device-level inquiry **with** VPR's mapping results.

#### Possible Solution

Move the pure device-level inquiry APIs to libarchfpga as a separated file. 
Then unit test can be added to libarchfpga for these APIs.

#### Context

The clean definition on APIs will ease the development of tools that use VTR as a library, e.g., [OpenFPGA](https://github.com/LNIS-Projects/OpenFPGA).
These APIs offer a standard way to access VTR's data structure.

#### Your Environment

* VTR revision used: Latest master
* Operating System and version: CentOS 7
* Compiler version: gcc-8.4

PR #1593, opened 2020-11-17
utils: move device specific utilities to arch_util

acomodi:

#### Description

There are several utility functions in `vpr_utils` which belong to the `arch_util` file instead.

This because these utility functions should work only at the device level, and should be callable without the need of invoking VPR.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1587

#### Types of changes

#### Checklist:


vaughnbetz:
Adding Mohamed as another possible reviewer.

libs/libarchfpga/src/arch_util.h
```diff
@@ -89,4 +89,107 @@ bool block_type_contains_blif_model(t_logical_block_type_ptr type, const std::st
...
+int get_max_num_pins(t_logical_block_type_ptr logical_block);
+
+//Verifies whether a given logical block is compatible with a given physical tile
+bool is_tile_compatible(t_physical_tile_type_ptr physical_tile, t_logical_block_type_ptr logical_block);
+
+//Verifies whether a logical block and a relative placement location is compatible with a given physical tile
+bool is_sub_tile_compatible(t_physical_tile_type_ptr physical_tile, t_logical_block_type_ptr logical_block, int sub_tile_loc);
+
+//Returns the physical tile type which 'best' matches logical_block
+t_physical_tile_type_ptr pick_best_physical_type(t_logical_block_type_ptr logical_block);
```
tangxifan, line 132: Hi, I think the notion of 'best' is vague here and additional comments are needed to clarify what 'best' means in the context. I see that the function will return the first tile in the `equivalent_tile` list. Does the `equivalent_tile` list follow the same sequence as users define the equivalent sites in architecture XML file?  Or does the parser does any re-ordering? These explanation can help developers to know when to use the util functions.

libs/libarchfpga/src/arch_util.h
```diff
@@ -89,4 +89,107 @@ bool block_type_contains_blif_model(t_logical_block_type_ptr type, const std::st
...
+
+//Returns the logical block type which 'best' matches the physical tile
+t_logical_block_type_ptr pick_best_logical_type(t_physical_tile_type_ptr physical_tile);
```
tangxifan, line 135: Same comment as the function `pick_best_physical_type()`

tangxifan (changes requested):

tangxifan:
Hi @acomodi ,
   In general, the pull request is clean.
   Since we start rebuilding the utils functions for `libarchfpga`, I would suggest reorganizing the files as well. 
   As a developer, I am scared by long `.cpp` files where functions in different usages are placed together.
   There have already been many data structures involved in the `arch_utils.h/cpp` files. 
   This causes some issues like:
   - If development goes on, we may see these files keep growing. For example, `vpr_utils.h/cpp` already contains 4.5k lines. If I just some APIs for a data structure, I have to include many unrelated APIs.
   - Developer will just add more APIs to these files. And these files become somehow dirty corners of the code base. No one can and want to maintain a clean organization there since there are 4k+ lines codes.
    
   Therefore, I would suggest to create dedicated `utils.h/cpp` for each data structure. For example, for `physical_tile` data structure, there should be a `physical_tile_utils.h/cpp` where APIs directly related to `physical_tile_type` should be placed. For example,  function `get_best_physical_tile()`  is considered directly related to `physical_tile_type`  while function `get_best_logic_block(t_logical_block_type, t_physical_tile_type)` may not be considered if it is data query for the `t_logical_block_type`. 

    Also, we should add some comments in the source files to developers what functions should be placed in these `utils.h` files. For example: 
    `/************************************************************************
      * Function to perform fundamental operation for the physical_tile 
      * These functions are not universal methods for the physical_tile class
      * They are made to ease the development in some specific purposes
      * Please classify such functions in this file
      ***********************************************************************/`

    These are my opinions on organizing these most utilized functions.
    Feel free to comment and let me know what you think.

libs/libarchfpga/src/arch_util.h
acomodi: Done, added a comment and also renamed the function to `pick_physical_tile` instead.

libs/libarchfpga/src/arch_util.h
acomodi: Done (see comment above)

tangxifan (commented):
@acomodi Thanks for the effort! Looks clean to me.

tangxifan (approved):

acomodi:
@tangxifan Hi, thanks for the review.

I think you are totally right, I have moved all the `physical tile`/`logical block` functions to their own file. I kept them together it might be confusing probably to separate them in different files, as there are some utilities that do access both of the object types.

I am also thinking if it is actually better to incorporate these routines within the `logical block` / `physical tile` objects, instead of keeping them in a utility file, but this could be an even further improvement for the future.

tangxifan:
> @tangxifan Hi, thanks for the review.
> 
> I think you are totally right, I have moved all the `physical tile`/`logical block` functions to their own file. I kept them together it might be confusing probably to separate them in different files, as there are some utilities that do access both of the object types.
> 
> I am also thinking if it is actually better to incorporate these routines within the `logical block` / `physical tile` objects, instead of keeping them in a utility file, but this could be an even further improvement for the future.

I agree as well. This PR should focus on what the title defines. Moving some utility functions to a method of `logical block` / `physical tile` objects should be a future work and driven by developers' needs as well.

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -0,0 +1,426 @@
+#include "vtr_assert.h"
+#include "vtr_memory.h"
+#include "vtr_util.h"
+
+#include "arch_types.h"
+#include "arch_util.h"
+#include "arch_error.h"
+
```
vaughnbetz, line 8: Good to comment what the purpose of this file is / what routines should go in it.

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -0,0 +1,426 @@
...
+#include "physical_types_util.h"
+
+struct t_pin_inst_port {
```
vaughnbetz, line 11: Should comment what this data structure is for.  Gives data for a single physical pin I think -- what port it came from, what index within that port, what sub_tile and what capacity_instance?  Not quite sure what capacity_instance is -- which sub_tile within a certain type of sub_tile?

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -0,0 +1,132 @@
...
+
+/********************************************************************
+ *                                                                  * 
+ *  Physical types utility functions                                *
+ *                                                                  *
+ *  This source file contains several utilities that enable the     *
+ *  interaction with the architecture's physical types.             *
+ *  Mainly, the two classes of objects accessed by the utility      *
+ *  functions in this file are the following:                       *
+ *    - physical_tile_type                                          *
```
vaughnbetz, line 14: Probably good to explain what a physical_tile_type is (place on the grid I think) vs. a logical_block_type (netlist item I think).  Are logical_block_types only in the clb_netlist or can they be in either the atom_netlist or clb_netlist?

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -0,0 +1,132 @@
...
+ *  All the following utilities are intended to ease the            *
+ *  developement to access the above mentioned classes and perform  *
+ *  some required operations with their data.                       *
+ *                                                                  *
+ *  Please classify such functions in this file                     *
+ *                                                                  *
+ * ******************************************************************/
+
+bool is_opin(int ipin, t_physical_tile_type_ptr type);
+bool is_input_type(t_physical_tile_type_ptr type);
```
vaughnbetz, line 26: Should comment what these functions do.  is_input_type is an I/O pad test I think, while is_opin is a test for a single pin being an output pin of a block I think.  What index is ipin on -- a pin index on a block, or some other kind of index?)

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -0,0 +1,132 @@
...
+bool is_output_type(t_physical_tile_type_ptr type);
+bool is_io_type(t_physical_tile_type_ptr type);
+
+//Returns the corresponding physical pin based on the input parameters:
+// - physical_tile
+// - relative_pin: this is the pin relative to a specific sub tile
+// - capacity location: absolute sub tile location
+int get_physical_pin_from_capacity_location(t_physical_tile_type_ptr physical_tile, int relative_pin, int capacity_location);
```
vaughnbetz, line 34: I think it would be helpful to give an example here where you showed how to pass in relative pin and capacity_location for a specific case.  Also define what the physical_pin index returned can be used for (it's an index into what data structure?).

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -0,0 +1,132 @@
...
+
+//Returns a pair consisting of the absolute capacity location relative to the pin parameter
+//and the relative pin within the sub tile
+std::pair<int, int> get_capacity_location_from_physical_pin(t_physical_tile_type_ptr physical_tile, int pin);
```
vaughnbetz, line 38: Again an example here would help I think (can define an example and use it for a few of these routines.  I'm not quite sure what "capacity location" means so precisely defining that (or giving an example) would help.

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -0,0 +1,132 @@
...
+
+//Returns the name of the class_index'th pin class on the specified block type
+std::vector<std::string> block_type_class_index_to_pin_names(t_physical_tile_type_ptr type, int class_index);
+
+//Returns the physical tile type matching a given physical tile type name, or nullptr (if not found)
+t_physical_tile_type_ptr find_tile_type_by_name(std::string name, const std::vector<t_physical_tile_type>& types);
+
+int find_pin_class(t_physical_tile_type_ptr type, std::string port_name, int pin_index_in_port, e_pin_type pin_type);
+
+int find_pin(t_physical_tile_type_ptr type, std::string port_name, int pin_index_in_port);
```
vaughnbetz, line 51: Define what the int returned is (id / index of pin, but into what data structure or what kind of id?)

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -0,0 +1,132 @@
...
+//Returns the physical pin index (within 'physical_tile') corresponding to the
+//logical index ('pin' of the first instance of 'logical_block' within the physcial tile.
+//This function considers if a given offset is in the range of sub tile capacity
+//
+//  (First pin index at current sub-tile)                                     (The wanted pin index)
+//
+//  |                                                               |<----- pin ------->|
+//  v                                                                                   v
+//
+//  |<----- capacity.low ----->|<----- capacity.low + 1 ----->| ... |<----- sub_tile_capacity ---->|
```
vaughnbetz, line 104: I like the example here but still don't fully understand it.  Can you make it more concrete or add a concrete example?

libs/libarchfpga/src/physical_types_util.h
vaughnbetz: I think we need a big comment that defines logical_pin_index, physical_in_index, sub_tile, capacity etc. Can refer to a block comment in another file if we already have all that or otherwise define those terms.  Those terms are used in comments below, but without clear definitions and an example of how they're used its hard to understand the comments and notation in the routines.

vaughnbetz:
I have some suggestions for additional comments, example, and definitions.  I think it's pretty important to add that to make the usage of these functions more clear.

vaughnbetz:
Hi @acomodi,
Any thoughts on the feedback above (extra commenting)?  Would be good to land this ...

acomodi:
@vaughnbetz Right, thanks for the ping, this was left behind in the todo list. I'll resume it and resolve all the comments.

libs/libarchfpga/src/physical_types_util.cpp
acomodi: This description was added in the header file, I think that is a better place for this kind of comment as it has all the utils prototype.

libs/libarchfpga/src/physical_types_util.cpp
acomodi: Done

libs/libarchfpga/src/physical_types_util.h
acomodi: Done

libs/libarchfpga/src/physical_types_util.h
acomodi: Done

libs/libarchfpga/src/physical_types_util.h
acomodi: Done

libs/libarchfpga/src/physical_types_util.h
acomodi: Done

libs/libarchfpga/src/physical_types_util.h
acomodi: Done

libs/libarchfpga/src/physical_types_util.h
acomodi: Done

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -0,0 +1,444 @@
...
+#include "vtr_util.h"
+
+#include "arch_types.h"
+#include "arch_util.h"
+#include "arch_error.h"
+
+#include "physical_types_util.h"
+
+/**
+ * @brief Data structure that holds information about a phyisical pin
```
vaughnbetz, line 12: minor nit: typo in physical (can fix after merge).

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * With the assumption that there is no 1:1 mapping between logical block and sub tile pins
+ * (e.g. EN pin in the BUFGCTRL sub tile), there is some extra computation and data structures
+ * needed to correctly identify the relation between the pins.
+ *
+ * For instance, the following information are required:
+ *   - mapping between logical and sub tile pins.
+ *   - mapping between sub tile pins and absoulte physical pin
+ *   - capacity instance of the sub tile
+ *
```
vaughnbetz, line 112: Great, comment. Thanks!

vaughnbetz (approved):
Thanks, the new comments help a lot!

PR #1597, opened 2020-11-19
Adding doxygen documentation for VTR utilities

MohamedElgammal:
#### Description

In this PR, I am adding a documentation for the supported VTR utilities in libs/libvtrutil/ path. The documentation was added by adding doxygen-compatible comments to the source. 

The main target of the docs are:
- Give a high-level comment about each utility
- Give a detailed description of the public method of each utility

#### Motivation and Context

This PR should help new developers to easily understand the utilities supported by vtr and use them more efficiently.

To generate the docs:
- clone the branch
- `cd  docs/ `directory
- `make html`
- The updated html is in doc/_build/html/index.html

The newly added documentations are in VPR API section, under Utilities title

#### Types of changes

#### Checklist:


libs/libvtrutil/src/generate.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+for filename in *.h
```
kmurray, line 3: Seems like this is a one off use script? If so it probably shouldn't be checked in.

Otherwise, we should give more documentation to clarify what this is for.

libs/libvtrutil/src/generate.sh
```diff
@@ -0,0 +1,10 @@
...
+do
+    echo "$filename" >> t
+    echo "" >> t
+    echo ".. doxygenfile:: $filename" >> t
+    echo "   :project: vpr" >> t
```
kmurray, line 8: `:project: vtr`?

kmurray (commented):
Overall looks good Mohamed.

It seems like all the .rst files are under api/vpr, since these are all part of libvtrutil (which is separate from VPR), it seems like they should be under `doc/src/api/vtrutil` or  `doc/src/api/libvtrutil`.

Similarly in RTD, this documentation shows up under the `VPR API` header, It should probably be under a separate heading (e.g. vtrutil API)

libs/libvtrutil/src/generate.sh
MohamedElgammal: Thanks for catching that. This file shouldn't be upstreamed as it was only used to generate an initial version for the .rst file.

MohamedElgammal:
@kmurray I have moved the docs to a new header that is called VTRUTIL API instead of VPR API and changed the .rst files paths accordingly.

libs/libvtrutil/src/vtr_assert.h
```diff
@@ -20,15 +23,16 @@
  *      VTR_ASSERT_LEVEL == 2: VTR_ASSERT_OPT, VTR_ASSERT enabled
  *      VTR_ASSERT_LEVEL == 1: VTR_ASSERT_OPT enabled
  *      VTR_ASSERT_LEVEL == 0: No assertion checking enabled
- * Note that an assertion levels beyond 4 are currently treated the same as level 4
+ *
+ * @Note that an assertion levels beyond 4 are currently treated the same as level 4
  */
 
-//Set a default assertion level if none is specified
+// brief Set a default assertion level if none is specified
```
acomodi, line 30: Missing` /@` before `brief`, if this was intended as a doxygen comment

libs/libvtrutil/src/vtr_cache.h
```diff
@@ -5,19 +5,21 @@
```
acomodi, line 5: Missing a file description? (Not sure if it is needed in this case, but it is worth keeping consistency with other headers)

libs/libvtrutil/src/vtr_expr_eval.h
```diff
@@ -13,29 +13,50 @@
...
 //function declarations
-//returns the global variable that holds all values that can trigger a breakpoint and are updated by the router and placer
+///@brief returns the global variable that holds all values that can trigger a breakpoint and are updated by the router and placer
 BreakpointStateGlobals* get_bp_state_globals();
 
 namespace vtr {
 
 /**** Structs ****/
 
+////@brief a class to hold the formula data
```
acomodi, line 40: One too many `/`

acomodi (approved):
LGTM. There are some typos, but the doxygen setup looks good. I can see that the docs generated do contain the full VTR APIs, and that looks good as well.

There are several API header files that do lack a file-scope doxygen comment. Is this intended to be left out? Not a big issue, but there is a little inconsistency with other files which do have a file-scope doxygen comment.

PR #1601, opened 2020-11-24
rr_graph: avoide div-by-zero issues while getting the delay norm fac

acomodi:

#### Description

This PR solves a coverty issue on the delay normalization factor calculation function for which, in corner cases, there might be a division by zero.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/rr_graph_indexed_data.cpp
+++ b/vpr/src/route/rr_graph_indexed_data.cpp
@@ -278,6 +278,11 @@ static float get_delay_normalization_fac() {
         Tdel_num += 1;
     }
 
+    if (Tdel_num == 0) {
+        VTR_LOG_WARN("No valid cost index was found to get the delay normalization factor. Setting delay normalization factor to 1e-9 (1 ns)\n");
+        return 1e-9;
+    }
+
     float delay_norm_fac = Tdel_sum / Tdel_num;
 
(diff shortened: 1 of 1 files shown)


vaughnbetz:
Would returning 0 result in 0 base costs?  If so, I think it's better to return 1 or 1e-9 (1 ns) if this happens.  Since we're issuing a warning and continuing we should make sure the returned value will let routing complete.

acomodi:
> I think it's better to return 1 or 1e-9 (1 ns) if this happens

@vaughnbetz Sure, makes sense. Fixed

vaughnbetz:
LGTM.  Whenever CI goes green this can be merged.

vaughnbetz:
Not sure what is going on in Travis (looks like one test hung) but this is clearly safe and passed the more extensive tests.  Merging.

acomodi:
@vaughnbetz Travis is currently having some troubles for what I understood, and this is happening pretty much everywhere. I think we might need to consider to move towards using GH actions instead.

vaughnbetz:
Hmmm ... won't let me merge unless I change permissions.  @acomodi if you can't merge either can you relaunch the hung travis CI build?

vaughnbetz:
That would be fine with me.  Good topic for Thursday.

vaughnbetz:
OK, changed permissions temporarily to let me merge this.

PR #1602, opened 2020-12-03
Vpr constraints fixes

sfkhalid:
Added a change to the XML Schema so that atoms and regions can be added in any order. Also added ability to provide regular expressions in place of atoms.

### DIFF
--- a/vpr/src/base/vpr_constraints_serializer.h
+++ b/vpr/src/base/vpr_constraints_serializer.h
@@ -211,6 +211,11 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
 
     virtual inline void set_vpr_constraints_tool_name(const char* /*tool_name*/, void*& /*ctx*/) final {}
 
+    virtual inline void set_vpr_constraints_constraints_comment(const char* /*constraints_comment*/, void*& /*ctx*/) final {}
+
+    virtual inline const char* get_vpr_constraints_constraints_comment(void*& /*ctx*/) final {
+        return temp_.c_str();
+    }
     virtual inline void* init_vpr_constraints_partition_list(void*& /*ctx*/) final {
         return nullptr;
     }
--- a/vpr/src/base/vpr_constraints_serializer.h
+++ b/vpr/src/base/vpr_constraints_serializer.h
@@ -58,9 +58,34 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
     virtual inline void set_add_atom_name_pattern(const char* name_pattern, void*& /*ctx*/) final {
         auto& atom_ctx = g_vpr_ctx.atom();
         std::string atom_name = name_pattern;
+
+        auto atom_name_regex = std::regex(atom_name);
+
+        std::vector<AtomBlockId> clear_atoms;
+        atoms_ = clear_atoms;
+
         atom_id_ = atom_ctx.nlist.find_block(name_pattern);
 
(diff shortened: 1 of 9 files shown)


vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -58,9 +58,34 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
     virtual inline void set_add_atom_name_pattern(const char* name_pattern, void*& /*ctx*/) final {
         auto& atom_ctx = g_vpr_ctx.atom();
         std::string atom_name = name_pattern;
+
+        auto atom_name_regex = std::regex(atom_name);
+
+        std::vector<AtomBlockId> clear_atoms;
```
sfkhalid, line 64: use atoms_.clear() instead

sfkhalid: in other places in this file, use this same method

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -58,9 +58,34 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
...
+        atoms_ = clear_atoms;
+
         atom_id_ = atom_ctx.nlist.find_block(name_pattern);
 
-        if (atom_id_ == AtomBlockId::INVALID()) {
+        if (atom_id_ != AtomBlockId::INVALID()) {
```
sfkhalid, line 69: comment here

sfkhalid: might have a specific name, might have a regex, this code could get slow if you have a lot of wildcards

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -58,9 +58,34 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
...
+            atoms_.push_back(atom_id_);
+        } else {
+            /*If the atom name returns an invalid ID, it might be a regular expression, so loop through the atoms blocks
+             * and see if any block names match atom_name_regex.
+             */
+            for (auto block_id : atom_ctx.nlist.blocks()) {
+                auto block_name = atom_ctx.nlist.block_name(block_id);
+
+                if (std::regex_match(block_name, atom_name_regex)) {
+                    VTR_LOG("Matched block %s with ID %d to regex\n", block_name.c_str(), block_id);
```
sfkhalid, line 79: take this out, debug statement

sfkhalid: or make it vtr verbose level, can check comments for vtr utils

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -128,8 +153,10 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
     virtual inline void finish_partition_add_atom(void*& /*ctx*/) final {
         PartitionId part_id(num_partitions_);
 
-        if (atom_id_ != AtomBlockId::INVALID()) {
-            constraints_.add_constrained_atom(atom_id_, part_id);
+        for (unsigned int i = 0; i < atoms_.size(); i++) {
+            if (atoms_[i] != AtomBlockId::INVALID()) {
```
sfkhalid, line 157: shouldn't need if statement because this should never happen, get rid of the if statement. can use vtr assert -> blockid is not invalid. if atom netlist has an obvious check that block id is valid, can use that here. if there is no easy check, don't need to do that

PR #1604, opened 2020-12-04
Add missing includes.

hzeller:


mithro:
@hzeller - Any idea if this could be automatically checked by icwyu or similar tool?

hzeller:
Yes, [iwyu](https://include-what-you-use.org/) should be able to catch these things early. I suspect that it shouldn't be too hard to have it part of the CI-systems.

vaughnbetz:
Looks like everything passed except travis (which is almost certainly a travis issue).

Issue #1493, opened 2020-08-20
Explore converting the Travis CI / Kokoro CI to instead use GitHub actions


mithro:
@FFY00 has set up GitHub actions to generate Arch Linux packages for SymbiFlow at https://github.com/FFY00/symbiflow-arch-pkgs

This has a couple of really nice aspects;
 - It generates a list of tasks to run based on which packages have been modified.
 - It can potentially have a much clearer dependency tree described.

It seems that GitHub actions are the future here. @FFY00 has said he would be interesting to explore using it instead.

mithro:
@acomodi - I believe this is now done? We have converted to using GitHub Actions (with self hosted runners on Google Cloud Platform replacing Kokoro).

PR #1608, opened 2020-12-11
ci: add GitHub Actions workflow 'Test', deprecate Travis CI

### LINKED ISSUE #847: Travis build cache issue causing sporadic CI failures
#### Expected Behaviour

Travis should use a single build step to produce the VTR executables (VPR, ODIN etc.), which are then re-used during the following testing steps (see #619).

#### Current Behaviour

The above works most of the time, but sporadically fails (at least two instances since #619 was merged in January), with the steps subsequent to the build step failing to find any of the generated executables. For example: https://travis-ci.org/verilog-to-routing/vtr-verilog-to-routing/builds/562741339

image

where the build succeeded, but the test failed (inspection shows that the required executables were missing).

Since this has sporadically appeared and then gone, its possible this is an issue within the TravisCI infrastructure. But it could potentially be an issue with our configuration (not sure).

#### Possible Solution

One option would be to disable build caching, this would slow down the CI times but side step the issue if its a TravisCI problem.

#### Context

Having inconsistent CI makes it very difficult to fairly evaluate new PRs and detect regressions.

### LINKED ISSUE #1473: Travis CI intermittently failing to fetch llvm packages, causing fails
#### Expected Behaviour

Travis CI normally tests pull requests to ensure they build correctly, and on many different compilers like GNU GCC and LLVM clang. 

#### Current Behaviour

On a number of PR's Travis is reporting failed jobs due to an inability to fetch LLVM libraries from apt-get. This doesn't happen always, and sometimes restarting the job fixes it, sometimes it causes another job to fail. It's not always the same files that it fails to fetch.

For example: https://travis-ci.com/github/verilog-to-routing/vtr-verilog-to-routing/jobs/369808911#L771 which is a part of my RCV branch merge PR. 

Also this https://travis-ci.com/github/verilog-to-routing/vtr-verilog-to-routing/jobs/369149961#L761

Both above jobs fail in a similar way, for example:
`E: Failed to fetch http://apt.llvm.org/trusty/pool/main/l/llvm-toolchain-8/clang-8_8.0.1~svn360950-1~exp1~20190517011409.67_amd64.deb  Could not connect to apt.llvm.org:80 (199.232.66.49), connection timed out`

It is usually clang and clang format that fail. I've also encountered this failure: https://travis-ci.com/github/verilog-to-routing/vtr-verilog-to-routing/jobs/369808913 where apt-get is also unable to locate the clang package. 

This can be fixed by brute force by force restarting jobs.

#### Possible Solution

It could be an issue with the ubuntu repository mirrorlist, changing this to use a different mirror could present results. One temporary option is to remove the clang jobs for now. 

#### Steps to Reproduce

1. Create a pull request 
2. Observe CI test results

#### Context

It's effected two of my PR's so far, as well as another pending PR: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1466

#### Your Environment

* VTR revision used:
* Operating System and version:
* Compiler version:

umarcor:
#### Description

This early work in progress for migrating from Travis CI to GitHub Actions. Not perfect yet, but it seems to work so far. I'm an absolute ignorant on the VTR codebase and on previous CI usage. I just dropped some hours here. Hence, any experienced user please review!

#### Related Issue

- Ref #1493
- Close #847
- Close #1473

#### Motivation and Context

/cc @mithro 

Discussion in verilator: https://github.com/verilator/verilator/issues/2658

It's CI plumbing, so self-tested. See below 👇🏼 

#### Types of changes

This is a breaking change regarding CI, but should not affect the codebase at all. Not sure what to check below.

#### Checklist:

Too early to check any of the items below.


vaughnbetz:
Thanks a lot -- this looks great and will be super useful!
Getting off travis and on github actions will be great, and the more fine-grained actions will allow faster iteration on CI failures (better initial triage). 
@lighost has some ideas on why some CI actions (sanitized builds of the strong regtests) will probably time out on github actions (may need to be run on kokoro only for those ones); mentioning him here so he can take a look and fill in the details.

.github/workflows/test.yml
```diff
@@ -0,0 +1,292 @@
...
+        CMAKE_PARAMS: '-DVTR_ASSERT_LEVEL=3 -DVTR_ENABLE_SANITIZE=on -DVTR_IPO_BUILD=off -DWITH_BLIFEXPLORER=on'
+        MATRIX_EVAL: "CC=gcc-5 && CXX=g++-5"
+        BUILD_TYPE: debug
+        LSAN_OPTIONS: 'exitcode=42' #Use a non-standard exit code to ensure LSAN errors are detected
+      run: |
+        source .github/travis/common.sh
+        ./.github/travis/setup.sh
+        ./.github/travis/build.sh
+        # We skip QoR since we are only checking for errors in sanitizer runs
+        ./run_reg_test.py ${{ matrix.suite }} -show_failures -j2
```
litghost, line 198: This needs `-skip_qor` because the sanitized version doesn't have good runtime, which is why the basic sanitized version failed.  The vtr_reg_strong takes too long on GH Actions, and so should be dropped.  The Kokoro CI "Strong Sanitized" has ultimited runtime, so I don't think we need the strong sanitized here.

mithro: FYI - GitHub actions has a significantly longer maximum timeout than Travis.
The default is 360 minutes, but it can be extended to be much longer by using https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

 * Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.
 * Workflow run time - Each workflow run is limited to 72 hours. If a workflow run reaches this limit, the workflow run is cancelled.

umarcor: In fact, `Regression · Strong · vtr_reg_strong` run rather fast: ~30min https://github.com/umarcor/vtr-verilog-to-routing/runs/1534764902?check_suite_focus=true. Timeout happened with `Sanitized · Strong · vtr_reg_strong`. Let's see with `-skip_qor` now.

umarcor:
@litghost would it be possible to split `Sanitized · Strong · vtr_reg_strong` into several groups so we can execute them in multiple jobs of up to 6h each?

litghost:
> @litghost would it be possible to split `Sanitized · Strong · vtr_reg_strong` into several groups so we can execute them in multiple jobs of up to 6h each?

I think such a thing is possible, but for now, let's just leave those tests as disabled.

umarcor:
> I think such a thing is possible, but for now, let's just leave those tests as disabled.

Then, I think this PR is ready. I believe the failing `F: Python` is unrelated to the PR, and I disabled `S: Basic` already.

litghost (approved):
@vaughnbetz I think this PR is ready for review.  Python formatting failure appears to be unrelated.

PR #1609, opened 2020-12-11
Don't compute congestion for non-configurable node sets up front.

HackerFoo:
#### Description

This PR improves performance by not summing congestion on entry of a nonconfigurable node set.

#### Motivation and Context

Symbiflow architectures use nonconfigurable node sets to model complex channel geometry, such as L-shaped wires. When the router enters a nonconfigurable node set, it sums the congestion cost for all nodes in the set.

This simple operation is performed many billions of times, and is one of the top time consumers in the CPU profile.

#### How Has This Been Tested?
This has been tested with [fpga-tool-perf](https://github.com/SymbiFlow/fpga-tool-perf) on the Hydra cluster with SymbiFlow's modifications, and compared to the baseline, an addition to tests run on CI.

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1613, opened 2020-12-15
Don't sum deterministic non-cofigurable node congestion.

HackerFoo:
#### Description
Alternative approach to https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1609

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/connection_router.cpp
+++ b/vpr/src/route/connection_router.cpp
@@ -677,10 +677,13 @@ void ConnectionRouter<Heap>::evaluate_timing_driven_node_costs(t_heap* to,
      * new_costs.R_upstream: is the upstream resistance at the end of this node
      */
 
+    auto& route_ctx = g_vpr_ctx.routing();
+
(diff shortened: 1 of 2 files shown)


HackerFoo:
@litghost What do you think of this approach? The idea is to calculate congestion normally when nonconfigurable nodes are used to represent a wire. I'll clean this PR up when I see the results, but I wanted your opinion on the approach.

litghost:
> @litghost What do you think of this approach? The idea is to calculate congestion normally when nonconfigurable nodes are used to represent a wire. I'll clean this PR up when I see the results, but I wanted your opinion on the approach.

I think approach will work on a subset of the relevant non-configurable node sets, so I expect to see some improvement.  However, I believe we discussed approaches in the previous meeting that would be fully generalized to all non-configurable node sets, and would enable the full performance gain in all cases.

This is a small incremental change, but isn't robust to graph layout.  Let's see the results.

PR #1614, opened 2020-12-16
Use std::string::find(char c) when searching for a single character.

hzeller:
Faster, and also clearer states intention.

#### Description
Very small cleanup.


vaughnbetz:
Thanks.  Travis CI is a bit messed up right now; doing a kokoro force to run the remaining regtests on it. This looks very safe in any case.

hzeller:
Uh, the CI is crunching on this three line change for over a month now ...

vaughnbetz:
Sorry, the move away from travis caused some slowdown.  Updated to the latest, and the tests have restarted.  These should pass (and if irrelevant things fail I'll override to merge).

hzeller:
Wonderful, looks like the test restart did work!

vaughnbetz:
:).

PR #1615, opened 2020-12-17
[WIP] Precalculate non-configurable node congestion base costs

HackerFoo:
#### Description
VPR spends a lot of time adding costs in `get_rr_cong_cost()` when used with SymbiFlow architectures, which is called many billions of times.

This PR precomputes the sum of base costs in each node set, and uses this to compute the congestion cost more quickly.

In addition, this fixes a bug where occupation costs are not updated for all nodes in a non-configurable set.

#### Motivation and Context
This is to improve VPR's performance.

Performance measurements are still in progress.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -344,6 +344,9 @@ struct RoutingContext : public Context {
      */
     vtr::dynamic_bitset<> non_configurable_bitset; /*[0...device_ctx.num_rr_nodes] */
 
+    ///@brief Congestion base cost of each non-configurably connected node
+    std::vector<float> rr_non_config_node_cong_base_costs;
+
     ///@brief Information about current routing status of each net
     t_net_routing_status net_status;
 
--- a/vpr/src/route/route_common.cpp
+++ b/vpr/src/route/route_common.cpp
@@ -99,7 +99,6 @@ static void adjust_one_rr_occ_and_acc_cost(int inode, int add_or_sub, float acc_
 bool validate_traceback_recurr(t_trace* trace, std::set<int>& seen_rr_nodes);
 static bool validate_trace_nodes(t_trace* head, const std::unordered_set<int>& trace_nodes);
 static vtr::vector<ClusterNetId, uint8_t> load_is_clock_net();
-static void invalidate_non_configurable_cong_costs(size_t inode);
 
 /************************** Subroutine definitions ***************************/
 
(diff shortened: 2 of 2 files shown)


HackerFoo:
@litghost This approach should preserve the same VPR's previous behavior. I'm collecting data to see if it is faster.

HackerFoo:
I've updated the algorithm based on discussion today.

litghost:
> I've updated the algorithm based on discussion today.

How does the performance look?  Hopefully the various tests go green.

litghost:
Looks like the strong tests are failing?  

Example:
```
Error: Executable vpr failed
	full command:  /usr/bin/env time -v /tmpfs/src/github/vtr-verilog-to-routing/vpr/vpr k4_N8_topology-0.85sL2-0.15gL4-on-cb-off-sb_22nm_22nm_nonLR.xml const_false --circuit_file const_false.pre-vpr.blif --gen_post_synthesis_netlist on --sweep_dangling_primary_ios off --sweep_constant_primary_outputs off --min_route_chan_width_hint 2
	returncode  :  139
	log file    :  /tmpfs/src/github/vtr-verilog-to-routing/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_post_routing_sync/run001/k4_N8_topology-0.85sL2-0.15gL4-on-cb-off-sb_22nm_22nm_nonLR.xml/const_false.blif/common/vpr.out
failed: Executable vpr failed (took 0.15 seconds)
```

HackerFoo:
Segfault in get_rr_cong_cost(). I forgot to change the size of `rr_non_config_node_cong_base_costs`. One per node, not one per set.

HackerFoo:
It doesn't seem to converge. I have more work to do.

HackerFoo:
I suspect that non-configurable nodes are now being over-adjusted, because multiple paths in the same routing tree could pass through the same set, but I'm not sure if this is possible.

litghost (changes requested):
CI failure?
```
[Fail]
buffered_flyover_wires.xml/raygentop.v/common vpr_status Task value 'exited with return code 134' does not match golden 'success'
```

HackerFoo:
I have marked this WIP because I don't think the simple fix here worked for https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1619, so I'll have to try something more sophisticated.

Issue #1577, opened 2020-10-27
RRNode side method no longer correct


litghost:
#### Context

#1448 changed the rr graph build to only generate one IPIN/OPIN RR node even if the relevant cluster had pins on multiple sides of the tile.  However when this change was made, the underlying RRNode method for getting the sides of a node was not changed, see:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/37095c742471af2e2c8d90faf10ab13156a47942/vpr/src/route/rr_graph_storage.h#L189-L195

`node_side` still returns just one value, even if the underlying RR node is actually representing multiple sides.

Code like: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/37095c742471af2e2c8d90faf10ab13156a47942/vpr/src/route/check_rr_graph.cpp#L565-L576 no longer work when a node represents multiple sides.

FYI @tangxifan / @vaughnbetz 

#### Possible Solution

Now that IPIN/OPIN nodes can represent pins on multiple sides of the tile, the relevant node methods need to be updated to account for this.  Possible solutions:
 - Change `side()` and `node_side()` to `sides()` and `node_sides()`
 - Remove `side()` and `node_side()` and add `is_on_side()`
 
Etc

tangxifan:
@litghost I noticed this issue when fixing the bug in routing resource graph. I have the same idea on changing the `side()` and `node_side()` method to `sides()` and `node_sides()`. But I am not sure what the `is_on_side()` will do? Does it record which side is used by the router?

litghost:
The signature of `is_on_side()` would be `bool RRNode::is_on_side(RRNode node, Side side)`, and returns true if the rr node connects to that side.  This would be a scalar version of `sides` or `node_sides`.

tangxifan:
@litghost Got it. Do you expect me to fix this on another pull request? Or you have a solution in the Symbiflow version?

litghost:
> @litghost Got it. Do you expect me to fix this on another pull request? Or you have a solution in the Symbiflow version?

We are have a workaround, but the current VPR data model in upstream is broken until this is fixed.  I'm surprised `check_rr_route` didn't fail somewhere, because we both agree that `node_side` no longer works correctly.  I suspect we've just gotten lucky so far.

vaughnbetz:
[...]

PR #1621, opened 2020-12-30
Node Sides Patch to Show All the Sides that a RR Node may Appear

tangxifan:
Improve the routing resource graph data structure to store all the node sides on which a ``rr_node`` may appear.

#### Description

Add a truth table to the ``rr_node_storage`` data structure which indicates on which sides a ``rr_node`` may appear.
The node sides are stored in an encoded format, where a `char` is used to represent if a `rr_node` appears on {TOP, RIGHT, BOTTOM, LEFT} sides. This keeps the rr_node storage size no changed.

This pull request introduces the following changes in the ``rr_node`` methods:
- **The current method ``node_side()`` is changed to access the first valid side for a given ``rr_node``.** The priority list is {TOP, RIGHT, BOTTOM, LEFT}. For example, if a node is on both TOP and BOTTOM sides. The method `node_side()` will return TOP. A fatal error will be thrown if the given ``rr_node`` has multiple sides.
- **The current method ``set_node_side()`` will set only 1 side to a given ``rr_node``, which overwrite all the previous side attributes.** For example, if a node was set to TOP side, using ``set_node_side(BOTTOM)`` will change the side attribute to BOTTOM side only.
-  Add a new method ``node_sides()`` through which developers can get a list of sides that a given ``rr_node`` may appear.
-  Add a new method ``node_on_specific_side()`` which is a shortcut for developers to find if a given ``rr_node`` appears on a specific side.
- Add a new method ``set_node_sides()`` which allows developers to set multiple sets to a given ``rr_nodes``.
- Add a new method ``add_node_side()`` through which developers can update side attributes incrementally.

More methods may be added upon the needs of routers as well as other client modules.

RR graph reader and writer have been updated to support the multiple sides
- XML file now accepts strings for multiple sides, e.g., ``TOP_RIGHT``
- Add a tiny RRGraph modeling a 11x8 array of Titan architecture in testing purpose

#### Related Issue

To address issue #1577 

#### Motivation and Context

Previously, the routing resource graph assumes that a ``rr_node`` appears ONLY on a side when it is an output/input pin.
However, as the advancing of routing architecture modeling in #1448 , a ``rr_node`` may appears on several sides in Titan architectures.
This hits the limitation of current data structure. As a result, only 1 side is stored, leading to a chain of problems, such as 
- Displaying these multi-sided nodes in GUI
- Router needs all the sides for a node to create accurate lookahead mapper
- etc.

Two new tests are added in this pull request, which focuses on testing the reader and writer of RRGraph XML
Instead, sanity checks are added to RR graph build-up functions, which ensure that node sides to set are consistent with the storage in RR graph.

The pull request will change the way that client functions in placer, router and other VPR modules to access the sides for a given node in routing resource graphs.

#### Types of changes

#### Checklist:


tangxifan:
@litghost I patched the node_side() method in rr-graph to show all the sides for a given rr-node.
The PR is ready for review.
See if the current status is good to merge or not. We can update client functions, such as GUI and lookahead mapper, in follow-up PRs.

litghost (changes requested):
You haven't updated https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.xsd , and so the `--read_rr_graph` and `--write_rr_graph` logic hasn't been updated.

vpr/src/route/rr_graph_storage.h
```diff
@@ -585,7 +616,50 @@ class t_rr_graph_storage {
...
         }
-        return node_storage[id].dir_side_.side;
+        std::bitset<NUM_SIDES> side_tt = node_storage[id].dir_side_.sides;
+        for (const e_side& side : SIDES) {
+            if (side_tt[size_t(side)]) {
+                return side;
+            }
+        }
+        /* No valid sides, return an invalid value */
+        return NUM_SIDES;
```
litghost, line 626: This should throw an exception or call `VTR_FATAL_ERROR`?

tangxifan: No problem. Should throw a fatal error as other methods

tangxifan:
> You haven't updated https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.xsd , and so the `--read_rr_graph` and `--write_rr_graph` logic hasn't been updated.

Thanks for reminding. I will patch the  `--read_rr_graph` and `--write_rr_graph` functions. But before doing that, we should agree on the file format changes. Since the node side is now a list of strings, current format `xs:string` may not work well for nodes with several sides. I am thinking about alternatives:
- Use a list of strings, such as `<attribute name="side" type="TOP,LEFT"/>`. But in this case, we have to enumerate all the  16 cases w.r.t. sides in the `.xsd` file.
```
 <xs:simpleType name="loc_side">
    <xs:restriction base="xs:string">
      <xs:enumeration value="TOP"/>
      <xs:enumeration value="RIGHT,TOP"/>
      <xs:enumeration value="RIGHT"/>
      ...
    </xs:restriction>
  </xs:simpleType>
```
-  Use a bitmask, such as `<attribute name="side" type="1001"/>`. It will change the data type to `xs:int`. We have to return a bitmask in the `node_sides()` method, which I am trying to avoid because it is ambiguous for developers.
```
    <xs:attribute name="side" type="xs:int"/>
```

I am open to discussion here.

litghost:
> > You haven't updated https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.xsd , and so the `--read_rr_graph` and `--write_rr_graph` logic hasn't been updated.
> 
> Thanks for reminding. I will patch the `--read_rr_graph` and `--write_rr_graph` functions. But before doing that, we should agree on the file format changes. Since the node side is now a list of strings, current format `xs:string` may not work well for nodes with several sides. I am thinking about alternatives:
> 
> * Use a list of strings, such as `<attribute name="side" type="TOP,LEFT"/>`. But in this case, we have to enumerate all the  16 cases w.r.t. sides in the `.xsd` file.
> 
> ```
>  <xs:simpleType name="loc_side">
>     <xs:restriction base="xs:string">
>       <xs:enumeration value="TOP"/>
>       <xs:enumeration value="RIGHT,TOP"/>
>       <xs:enumeration value="RIGHT"/>
>       ...
>     </xs:restriction>
>   </xs:simpleType>
> ```
> 
> * Use a bitmask, such as `<attribute name="side" type="1001"/>`. It will change the data type to `xs:int`. We have to return a bitmask in the `node_sides()` method, which I am trying to avoid because it is ambiguous for developers.
> 
> ```
>     <xs:attribute name="side" type="xs:int"/>
> ```
> 
> I am open to discussion here.

There is an advantage to leaving this as an enumeration, and leaving the old order in place.  It means that old rr graph files will work with the new schema.

tangxifan:
> > > You haven't updated https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.xsd , and so the `--read_rr_graph` and `--write_rr_graph` logic hasn't been updated.
> > 
> > 
> > Thanks for reminding. I will patch the `--read_rr_graph` and `--write_rr_graph` functions. But before doing that, we should agree on the file format changes. Since the node side is now a list of strings, current format `xs:string` may not work well for nodes with several sides. I am thinking about alternatives:
> > 
> > * Use a list of strings, such as `<attribute name="side" type="TOP,LEFT"/>`. But in this case, we have to enumerate all the  16 cases w.r.t. sides in the `.xsd` file.
> > 
> > ```
> >  <xs:simpleType name="loc_side">
> >     <xs:restriction base="xs:string">
> >       <xs:enumeration value="TOP"/>
> >       <xs:enumeration value="RIGHT,TOP"/>
> >       <xs:enumeration value="RIGHT"/>
> >       ...
> >     </xs:restriction>
> >   </xs:simpleType>
> > ```
> > 
> > 
> > 
> > * Use a bitmask, such as `<attribute name="side" type="1001"/>`. It will change the data type to `xs:int`. We have to return a bitmask in the `node_sides()` method, which I am trying to avoid because it is ambiguous for developers.
> > 
> > ```
> >     <xs:attribute name="side" type="xs:int"/>
> > ```
> > 
> > 
> > I am open to discussion here.
> 
> There is an advantage to leaving this as an enumeration, and leaving the old order in place. It means that old rr graph files will work with the new schema.

No problem then. I will patch it.

vpr/src/route/rr_graph.xsd
```diff
@@ -234,12 +234,59 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <!-- The enumeration MUST follow the order of side maskcode
```
litghost, line 237: Do not rely on the enumeration order matching the mask.  Preserve the previous order, and use a `std::array<>` array map to translate from the enum to the bitmask.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1626,39 +1631,84 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
-                return uxsd::enum_loc_side::BOTTOM;
-            default:
-                report_error(
-                    "Invalid side %d", side);
+    uxsd::enum_loc_side to_uxsd_loc_side(std::bitset<NUM_SIDES> sides) {
+        if (0 == sides.to_ulong()) {
+            report_error(
+                "Invalid side %ld", sides.to_ulong());
         }
+        return static_cast<uxsd::enum_loc_side>(sides.to_ulong());
```
litghost, line 1711: The fact that you are writing out a `static_cast` here should tell you this is not a good strategy.

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -711,7 +711,13 @@ void t_rr_graph_storage::set_node_side(RRNodeId id, e_side new_side) {
     if (node_type(id) != IPIN && node_type(id) != OPIN) {
         VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Attempted to set RR node 'side' for non-channel type '%s'", node_type_string(id));
     }
-    node_storage_[id].dir_side_.side = new_side;
```
litghost, line 714: I think this is not the right strategy.  I believe set_node_side should replace all existing sides (to match the description and previous behavior).  A new method, e.g. `set_node_sides` or a method like `add_node_side` should be added to handle the "more than one side case".

vpr/src/route/rr_graph.xsd
tangxifan: Not a problem for me. I was trying to compress the codes. I use `if-else` tree to translate.

vpr/src/route/rr_graph_storage.cpp
tangxifan: Make senses to me. Will add the two new methods and rework the `set_node_side()`

vpr/src/route/rr_graph_uxsdcxx_serializer.h
tangxifan: I was trying to save some codes. An explicit translation may work better.

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -711,7 +711,42 @@ void t_rr_graph_storage::set_node_side(RRNodeId id, e_side new_side) {
...
+    std::bitset<NUM_SIDES> side_bits;
+    /* The new side will overwrite existing side storage */
+    side_bits[size_t(new_side)] = true;
+    if (side_bits.to_ulong() > CHAR_MAX) {
+        VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Invalid side '%s' to be added to rr node %u", SIDE_STRING[new_side], size_t(id));
+    }
+    node_storage_[id].dir_side_.sides = static_cast<unsigned char>(side_bits.to_ulong());
+}
+
+void t_rr_graph_storage::set_node_sides(RRNodeId id, std::vector<e_side> new_sides) {
```
litghost, line 723: Why not take `std::bitset<NUM_SIDES>`?

vpr/src/route/rr_graph_storage.h
```diff
@@ -585,7 +623,53 @@ class t_rr_graph_storage {
...
-        return node_storage[id].dir_side_.side;
+        std::bitset<NUM_SIDES> side_tt = node_storage[id].dir_side_.sides;
+        for (const e_side& side : SIDES) {
+            if (side_tt[size_t(side)]) {
+                return side;
+            }
+        }
+        /* No valid sides, return an invalid value */
+        VPR_FATAL_ERROR(VPR_ERROR_ROUTE,
+                        "Invalid side for RR node '%d':\n\ttype='%s'\txlow,ylow=(%d,%d)\n\txhigh,yhigh=(%d,%d)",
```
litghost, line 634: The error should report that the user request "node_side" on a node that has more than one side.

vpr/src/route/rr_graph_storage.h
```diff
@@ -585,7 +623,53 @@ class t_rr_graph_storage {
                             "Attempted to access RR node 'side' for non-IPIN/OPIN type '%s'",
                             rr_node_typename[node_data.type_]);
         }
-        return node_storage[id].dir_side_.side;
+        std::bitset<NUM_SIDES> side_tt = node_storage[id].dir_side_.sides;
```
litghost, line 626: I think you need to detect if more than one side is present, and generate an error if this method is used in that case.

vpr/src/route/rr_graph_storage.h
```diff
@@ -585,7 +623,53 @@ class t_rr_graph_storage {
...
+        vtr::array_view_id<RRNodeId, const t_rr_node_data> node_storage,
+        const RRNodeId& id) {
+        auto& node_data = node_storage[id];
+        if (node_data.type_ != IPIN && node_data.type_ != OPIN) {
+            VPR_FATAL_ERROR(VPR_ERROR_ROUTE,
+                            "Attempted to access RR node 'side' for non-IPIN/OPIN type '%s'",
+                            rr_node_typename[node_data.type_]);
+        }
+        // Return a vector showing only the sides that the node appears
+        std::vector<e_side> exist_sides;
```
litghost, line 654: `exist_sides` is unused?

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -578,12 +578,17 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
                     inode, node.type());
             }
         } else {
-            node.set_side(from_uxsd_loc_side(side));
+            std::bitset<NUM_SIDES> side_mask = from_uxsd_loc_side(side);
+            for (const e_side& candidate_side : SIDES) {
+                if (side_mask[size_t(candidate_side)]) {
+                    node.set_side(candidate_side);
```
litghost, line 584: This should be "node.add_side".  How did a unit test not catch this failure?  Add another test case?

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1626,39 +1631,139 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
-                return uxsd::enum_loc_side::TOP;
-            case BOTTOM:
-                return uxsd::enum_loc_side::BOTTOM;
-            default:
-                report_error(
-                    "Invalid side %d", side);
+        return side_mask;
+    }
+
+    /* The enumeration MUST follow the order of side maskcode
```
litghost, line 1706: This comment is not right anymore?

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1626,39 +1631,139 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
+     *  12   |  X      X             | 1100
+     * ======+======================+======
+     *  13   |  X      X          X  | 1101
+     * ======+======================+======
+     *  14   |  X      X     X       | 1110
+     * ======+======================+======
+     *  15   |  X      X     X    X  | 1111
+     */
+    uxsd::enum_loc_side to_uxsd_loc_side(std::bitset<NUM_SIDES> sides) {
+        std::array<uxsd::enum_loc_side, 16> side_map = {uxsd::enum_loc_side::UXSD_INVALID,
```
litghost, line 1742: Can I make a suggestion on how to construct this?

Maybe in the constructor have something like:
```
std::array<uxsd::enum_loc_side, 16> side_map;
side_map[0] = uxsd::enum_loc_side::UXSD_INVALID;
side_map[(1 << TOP)] = uxsd::enum_loc_side::TOP;
side_map[(1 << RIGHT)] = uxsd::enum_loc_side::RIGHT;
side_map[(1 << TOP) | (1 << RIGHT)] = uxsd::enum_loc_side::TOP_RIGHT;
...
```

vpr/src/route/rr_graph_storage.cpp
tangxifan: Agree.

vpr/src/route/rr_graph_storage.h
tangxifan: No problem. I will throw a fatal error in this case

vpr/src/route/rr_graph_storage.h
tangxifan: Indeed. Will remove.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
tangxifan: Good catch. I forgot to update this line. It is not detected because most rr_graph XML has only 1 side for rr_nodes.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
tangxifan: Indeed. Will update

vpr/src/route/rr_graph_uxsdcxx_serializer.h
litghost: The second part of my comment was to suggest we specifically need a basic / strong test to check round tripping a design with more than 1 side on the rr_nodes, otherwise this code could easily go stale.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
tangxifan: Make senses. It is more straightforward to read and maintain.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
tangxifan: I agree. As far as I know, the Titan architecture has multiple sides for many ``rr_nodes``. But its ``rr_graph.xml`` could be very large which may be an overkill in terms of complexity in basic/strong test. Do you have any small examples? I can add a new test.

tangxifan:
@vaughnbetz @litghost I have added test cases and convert fatal errors to warnings. It is ready for review.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_read_rr_graph/config/config.txt
```diff
@@ -0,0 +1,27 @@
+#
```
litghost, line 1: The "verify_rr_graph" tests do what you want: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/aec608ff60a0c01b3c93407b722684d810f04b23/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_verify_rr_graph_bin/config/config.txt#L29

The `-verify_rr_graph` flag writes out a rr graph and reads it back and and verifies its the same rr graph.

litghost: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/aec608ff60a0c01b3c93407b722684d810f04b23/vtr_flow/scripts/run_vtr_flow.py#L576-L578
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/08f054c85e22ddf33811d91b2dd45daf5ee2341e/vtr_flow/scripts/python_libs/vtr/vpr/vpr.py#L271-L299

Use this structure also means you don't need to add the rr graph to the repo.

tangxifan: Great. I have changed the test to use `-verify_rr_graph` feature. Now we just need a test `verify_rr_graph_titan` to do the job.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1626,39 +1626,139 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
+     *  12   |  X      X             | 1100
+     * ======+======================+======
+     *  13   |  X      X          X  | 1101
+     * ======+======================+======
+     *  14   |  X      X     X       | 1110
+     * ======+======================+======
+     *  15   |  X      X     X    X  | 1111
+     */
+    uxsd::enum_loc_side to_uxsd_loc_side(std::bitset<NUM_SIDES> sides) {
+        std::array<uxsd::enum_loc_side, 16> side_map;
```
litghost, line 1736: This should be initialized in the constructor for this class and re-used, rather than reconstructed everytime.  Other option is to make this a return from a constexpr function, but I don't know where the limit of constexpr functions and std::array.

tangxifan: It makes sense. I know how to create a `constexpr` function like the `SIDES` variables. But I do not know if a constant side_map can be built using the same way. If you want, we can add the side_map as an internal data of the data structure and create it when initialize the object.

litghost: > If you want, we can add the side_map as an internal data of the data structure and create it when initialize the object.

That was what I meant when I said "This should be initialized in the constructor for this class and re-used"

vpr/src/route/rr_graph_storage.h
```diff
@@ -192,6 +203,19 @@ class t_rr_graph_storage {
                 node_storage_.data(), node_storage_.size()),
             id);
     }
+    std::bitset<NUM_SIDES> node_sides(RRNodeId id) const {
```
litghost, line 206: Need method descriptions.

litghost (changes requested):
Looks really good!  Please add method descriptions to the new methods on rr_graph_storage.  Otherwise, I think this is ready for final review and merge if green.

tangxifan:
@litghost I have added the comments about the new APIs about how the return values should be used. It is ready for you and @vaughnbetz to review.

litghost (approved):
LGTM

tangxifan:
It seems that all the pull requests are staled because the updates on CI, I merged latest master and it contains errors on the Python test. If this has been resolved on the master branch, I can update my local one.

litghost:
> It seems that all the pull requests are staled because the updates on CI, I merged latest master and it contains errors on the Python test. If this has been resolved on the master branch, I can update my local one.

The Python CI is not marked as required, so it shouldn't block anything.  The other CIs are still running.

tangxifan:
> > It seems that all the pull requests are staled because the updates on CI, I merged latest master and it contains errors on the Python test. If this has been resolved on the master branch, I can update my local one.
> 
> The Python CI is not marked as required, so it shouldn't block anything. The other CIs are still running.

Got it. No worries then. Seeing some red cross is so friendly to OCD patients :D

litghost:
> > > It seems that all the pull requests are staled because the updates on CI, I merged latest master and it contains errors on the Python test. If this has been resolved on the master branch, I can update my local one.
> > 
> > 
> > The Python CI is not marked as required, so it shouldn't block anything. The other CIs are still running.
> 
> Got it. No worries then. Seeing some red cross is not so friendly to OCD patients :D

Someone does need to do a reformat to get it green.

vaughnbetz:
Looks good. As discussed in the meeting, I think in the follow up commit it is good to add a comment on why you went with the 16 enum solution for the sides of a block that a node appears in (compatibility with old rr_graph parsing, by keeping an enum with the first four values I think).

Also, when coding the multi-side support in a follow-on PR, please check if the node_sides (returns a bitset for iteration) is indeed used and cleaner than just looping over the sides and calling node_on_specific_side().  I don't have a big objection to the bitset, but a narrower interface is a bit simpler to maintain if it doesn't cost us anything. But if the bitset is used and works better in some callers, then keeping it makes sense.

I'm not sure why the nightly tests haven't run yet; trying a kokoro restart.

tangxifan:
@vaughnbetz I have added the comments to clarify 
- the use of 16 enumeration for node sides.
- the design purpose of the node_sides() API. And comment that it may be deprecated later depending on the client function.

Still, waiting for the nightly test to finish.

vaughnbetz:
Are the nightly tests taking unusually long?  Lots of changes in CI and in placement settings (new RL stuff) so worth flagging it if things have slowed down.

vaughnbetz:
Changes look good, thanks.  Just waiting for CI to finish ... perhaps pedantic since the last updates were only comments and I believe CI went fully green before that?

tangxifan:
> Are the nightly tests taking unusually long? Lots of changes in CI and in placement settings (new RL stuff) so worth flagging it if things have slowed down.

According to my experience, it was almost 1 day run. I checked the log. It started 3PM yesterday. So I expect to finish this afternoon. Otherwise, something may go wrong.

tangxifan:
@vaughnbetz I have removed the warning message from log files, as part of the issue #1647

PR #1629, opened 2021-01-12
Move initialization before block that can be skipped by 'goto'.

hzeller:
'goto' over variable initializations leave these in an undefined
state. This fixes the build in case ENABLE_ANALYTIC_PLACE is defined.

#### Description
This just fixes the build in case `ENABLE_ANALYTIC_PLACE` is enabled. No functional changes.


hzeller:
Are these tests also stuck in the Travis blackhole ?

MohamedElgammal:
@hzeller this issue is now fixed in master.

PR #1645, opened 2021-01-28
CMake: do not install test binaries

Xiretza:
#### Description

The test binaries used for `make test`/ctest are also installed to the destination system using `install()` directives. This makes little sense, as the tests do not need to be run outside a development environment.

#### Related Issue

N/A

#### Motivation and Context

I am packaging vtr in the [vtr-git AUR package](https://aur.archlinux.org/packages/vtr-git/). The installed package should not contain unit test binaries.

`make install` no longer installs test binaries, `make test` still works as expected.

#### Types of changes

#### Checklist:


vaughnbetz:
Thanks!

PR #1653, opened 2021-02-02
pb cleaning during clustering

mkurc-ant:
Recently I've encountered the assertion failure during clustering: `vpr/src/pack/cluster.cpp:1437 try_place_atom_block_rec: Assertion 'parent_pb->mode == pb_graph_node->pb_type->parent_mode->index' failed.` This PR is an attempt to fix it.

#### Description
While debugging I discovered that the issue seems to be caused by the packer changing mode of a parent pb without re-allocating appropriate children for that mode which results in a conflict. In other words: when trying mode B after mode A there are already children for mode A allocated while they shouldn't be.

With this PR I added cleaning of the pb being used to form a cluster using the `free_pb` function. The cleanup is done each time when `try_pack_molecule` fails to pack for a given mode instead of after all available modes are exhausted.

The proposed fix seems to be working. However, I'm not that familiar with the packer code to tell whether it is the appropriate way of fixing the issue. I'd welcome any suggestions if a better solution is more feasible.

#### Related Issue

None

#### Motivation and Context

I've been working for supporting a VPR architecture that is used in the OpenFPGA project in SymbiFlow (https://github.com/lnis-uofu/SOFA/blob/master/ARCH/vpr_arch/k4_N8_tileable_reset_softadder_register_scan_chain_nonLR_caravel_io_skywater130nm.xml)

For now only locally using a few designs that triggered the assertion.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/pack/cluster.cpp
+++ b/vpr/src/pack/cluster.cpp
@@ -1992,15 +1992,15 @@ static void start_new_cluster(t_cluster_placement_stats* cluster_placement_stats
         auto type = candidate_types[i];
 
         t_pb* pb = new t_pb;
-        pb->pb_graph_node = type->pb_graph_head;
-        alloc_and_load_pb_stats(pb, feasible_block_array_size);
-        pb->parent_pb = nullptr;
 
         *router_data = alloc_and_load_router_data(&lb_type_rr_graphs[type->index], type);
 
         //Try packing into each mode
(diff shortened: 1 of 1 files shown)


mkurc-ant:
This PR however still does not fix the issue when there is a mode with `disable_packing="true"`. Investigating the issue further...

mkurc-ant:
The fix I proposed won't work in general. Im closing the PR.

Issue #1655, opened 2021-02-03
Assertion 'parent_pb->mode == pb_graph_node->pb_type->parent_mode->index' failure


mkurc-ant:
Recently I've encountered the assertion failure during clustering: `vpr/src/pack/cluster.cpp:1437 try_place_atom_block_rec: Assertion 'parent_pb->mode == pb_graph_node->pb_type->parent_mode->index' failed.`

I've spent some time debugging the code and I've roughly narrowed the problem down. For some reason `child_pbs` together with a selected mode of a `pb`  are not cleared correctly. Those are allocated and set as packer tries to place an atom. Every time a placement fails those should be cleared but in rare cases they are not.

The same assertion gets triggered in some cases for architectures with disabled modes (with `disable_packing = "true"`). The packer allocates `pb` and its `child_pbs` for such modes and bails out later.

#### Expected Behaviour
Packer should either succeed or fail but without triggering assertions.

#### Current Behaviour
The assertion 'parent_pb->mode == pb_graph_node->pb_type->parent_mode->index'  at ``vpr/src/pack/cluster.cpp:1437 try_place_atom_block_rec` gets triggered.

#### Possible Solution
Add (or fix?) cleanup code that removes `child_pbs` created during atom placement if it fails for any reason.

#### Steps to Reproduce

Use `arch.xml` and `design.blif` from the [design.tar.gz](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/5926031/design.tar.gz) archive. Run VPR with the following command:
```
vpr arch.xml design.blif --device ql-openfpga-k4_N8 --constant_net_method route --route_chan_width 60 --clock_modeling ideal --pack
```

It will trigger the assertion: `vpr/src/pack/cluster.cpp:1449 try_place_atom_block_rec: Assertion 'pb->pb_graph_node == pb_graph_node' failed.` which is not the same as the one reported in the issue but the underlying problem is likely the same.

#### Context

I've been working for supporting a VPR architecture that is used in the OpenFPGA project in SymbiFlow (https://github.com/lnis-uofu/SOFA/blob/master/ARCH/vpr_arch/k4_N8_tileable_reset_softadder_register_scan_chain_nonLR_caravel_io_skywater130nm.xml)

#### Your Environment

* VTR revision used: GIT `fa683da07`
* Operating System and version: Linux Ubuntu 18.04
[...]

PR #1656, opened 2021-02-03
Added cleaning of a pb after unsuccessful molecule packing

mkurc-ant:
Attempt to fix https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1655


litghost (changes requested):
This needs a test.

mkurc-ant:
> This needs a test.

@litghost Test added.

mkurc-ant:
I've run `vtr_reg_nightly` on the same machine. Both for master (@53780c0b9) and this PR. There was no difference in any of the QoR metrics except runtime (this was expected).

| metric | baseline | modified (this PR) |
| --- | --- | --- |
| total_runtime | 1.000 | 1.010 |
| pack_time | 1.000 | 1.013 |

litghost (approved):
LGTM, waiting for other reviews.

acomodi (approved):
LGTM. I think we need to re-run CI as the Nightly one errored out for some reason ([link](https://source.cloud.google.com/results/invocations/51c65e56-c8af-4437-a806-dc05d45ae26c/targets)).

vpr/src/pack/cluster.cpp
```diff
@@ -1179,6 +1179,64 @@ static void alloc_and_load_pb_stats(t_pb* pb, const int feasible_block_array_siz
 }
 /*****************************************/
 
+/**
+ * Cleans up a pb after unsuccessful molecule packing
+ */
```
vaughnbetz, line 1184: Good to comment this routine and how it works as much as you can.

vaughnbetz:
LGTM too.  My one request is to comment it as much as you can.  I realize that may be limited if you don't have full understanding of exactly how / why these structures are set up, but if you give as much insight as you can into why this routine is being called, what it does and when it is safe to call it that would be good.

vpr/src/pack/cluster.cpp
```diff
@@ -1358,6 +1416,10 @@ static enum e_block_pack_status try_pack_molecule(t_cluster_placement_stats* clu
                         revert_place_atom_block(molecule->atom_block_ids[i], router_data, atom_molecules);
                     }
                 }
+
+                /* Placement failed, clean the pb */
```
sfkhalid, line 1420: To clarify, is the cleanup_pb function supposed to be called when a molecule fails try_pack_molecule for any reason? Or only when it fails during atom placement (try_place_atom_block_rec)? I’m asking because in the issue it says “Add (or fix?) cleanup code that removes child_pbs created during atom placement if it fails for any reason.”

If it is meant to be called when a molecule fails for any reason, it should also be called if the atom fails the detailed routing legality check in try_pack_molecule. (line 1305/1363)

If the cleanup function is only meant to be called if the atom fails during try_place_atom_block_rec , then maybe the call for cleanup_pb should be at the point right after try_place_atom_block_rec is called. So right after the atom placement attempt, you could have a check of the block_pack_status and call the cleanup function if it needed. Right now, where the cleanup_pb function is called, it could have failed the placement check or the pin feasibility filter check.

mkurc-ant: You're right. It should be called whenever `try_place_atom_block_rec` fails. But if you look at the loop that invokes it for each atom you'll see that it terminates immediately on failure: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/8e00e82a03ce12b7987a244620a0e3624083ae79/vpr/src/pack/cluster.cpp#L1241

Later in the flow there are calls to some atom placement reversal routines (https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/8e00e82a03ce12b7987a244620a0e3624083ae79/vpr/src/pack/cluster.cpp#L1350) executed on failure so I figured out that the pb cleanup should be invoked from there. It doesn't matter what is the reason of the atom placement failure. If no pbs were allocated the function simply won't do anything. BTW: it does not touch pbs with already placed atoms.

Diving into the low-level details: The problem I observed with the assertion failure in `try_place_atom_block_rec` was that for some reason the function was hitting already allocated pbs with their modes set causing a conflict with the mode required by the currently packed atom. Digging deeper I've discovered that the function is called for a pb_graph_node corresponding to that atom. Then it recurses up the hierarchy allocating pbs and setting their modes until the root CLB is reached. But these pbs are never freed between subsequent molecule packing. So if packing succeeds everything is fine but when it fails there are leftover pbs set to specific modes. This may cause mode mismatch if the mode required for the next molecule is different than the last one checked.

I typed "Add (or fix?)..." in the issue as I was surprised that the problem haven't manifested itself until now. I'm still not sure whether I'm adding the actually missing data structure cleanup or if the cleanup is already there and requires fixing and I've missed it.

mkurc-ant:
I added more comments to the cleanup function that I've added.

vpr/src/pack/cluster.cpp
sfkhalid: Ah I see, that makes sense. In that case LGTM.

vaughnbetz:
Looks good to me.  Looks like it needs to be updated to match the base branch and then can be merged.
Also looks like vtr_nightly didn't run again.  flagging that to @litghost

mkurc-ant:
@vaughnbetz Updated.

vaughnbetz:
vtr_reg_nightly got stuck again, but based on the testing and QoR data I'm confident this is OK and we should merge it.  @litghost: seems like this is another case of a stuck nightly reg.

mkurc-ant:
@vaughnbetz thanks.

Issue #1580, opened 2020-10-29
Router lookahead inefficiency for large blocks


vaughnbetz:
Currently the router lookahead takes two things:
- rr_node you are currently on
- the rr_node you are trying to get to, which is a SINK.

For large blocks, VPR's rr-graph builder sets the (x,y) of all SINK nodes to the lower-left corner of the block. This may be far from the actual pins we're trying to get to, making the router slower and less effective.

#### Expected Behaviour
Router should route these connections efficiently.

#### Current Behaviour
Titan results suggest this is hurting us on large blocks; logically equivalent pins on those blocks don't help us as much as expected and we get some circuitous routes to large blocks both with and without logical equivalence.

#### Possible Solution
I think the cleanest solution would be for the rr-graph builder to assign a smarter (x,y) for SINKs in large blocks. It could be the average (x,y) of the IPINs that can reach that SINK, or the (x,y) of a randomly chosen IPIN that reaches the SINK.

vaughnbetz:
@litghost @acomodi @tangxifan FYI.
Looking for an owner ...

tangxifan:
Hi @vaughnbetz , 
   I have reading the router lookahead map codes.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/574525c9f14e4c2c8856ee992d37888eaa0b3526/vpr/src/route/router_lookahead_map.cpp#L1055

It seems that @acomodi has patched the issue as you said. Currently, the sink coordinate to be used in calculating the ``delta_xy()`` is the center coordinate of a multi-height and multi-width block.
It may resolve the QoR problem. I am running the titan regression tests now.

Something we can try is to elaborate the ``adjust_rr_src_sink_posistion()`` function
- We can visit all the fan-out nodes of the SINK node, which are OPINs. The we average the ``x_low()``, ``ylow()`` of these OPINs as the final coordinate to return.
- We can visit all the fan-in nodes of the SOURCE node, which are IPINs. The we average the ``x_low()``, ``ylow()`` of these IPINs as the final coordinate to return.

Looking forward to your reply.

vaughnbetz:
Thanks Xifan.  
[...]

PR #1661, opened 2021-02-11
Deploy new API rr_node.node_sides() to GUI

tangxifan:
This is a follow up pull request for the PR #1621 
The purpose of this PR is to deploy the new APIs about the node sides of RRGraph to the *Graph User Interface* (GUI)

**Note**
This PR will not deprecate the old APIs and the unnecessary node_sides() APIs.
The deprecation will be done after lookahead map has been fixed, as the final PR in this series of patches

#### Description

Major changes are landed in the GUI part 
In the GUI part, the major changes to enable the drawing for lines between the IPIN/OPIN and CHANX/CHANY routing resource nodes.
As an IPIN/OPIN may have multiple sides, the drawer should identify the correct line (from the multiple possible connections) to display.

** A Side Note**
A separated PR will be created to solve the issue on the router lookahead map (as stated in issue #1580)
In the router lookahead map, a proper side should be picked from multiple sides of a rr_node so that the lookahead map can have an accurate estimation on the routing congestion.
As multiple sides are used in Titan architectures, a proof of the QoR changes is required!

#### Related Issue

To address issue #1577 

#### Motivation and Context

To show correct RRGraph display for nodes/edges with multiple sides in GUI.

**As multiple sides are used in Titan architectures, a proof of the GUI display is required!**

My major focus is on the Titan architecture where multiple sides are heavily used for rr_nodes.
- Before GUI patching (LAB in Titan Architecture): 
image

- After GUI patching (LAB in Titan Architecture): You can see the connections on the top side has been correctly shown, which are driven by LAB.data_in (locate on TOP, LEFT, RIGHT sides)
image

- Before GUI patching (PLL and I/Os in Titan Architecture): 
image

- After GUI patching (PLL and I/Os in Titan Architecture): 
image

- Before GUI patching (M9K in Titan Architecture): 
image

- After GUI patching (M9K in Titan Architecture):
image

- Before GUI patching (DSP36 in Titan Architecture): 
image

- After GUI patching (DSP36 in Titan Architecture): 
image

**Note that there is no difference on normal architectures.
- Before GUI patching (k6_N10_mem32K_40nm.xml): 
image

- After GUI patching (k6_N10_mem32K_40nm.xml): 
image

The changes will only affect GUI codes under ``vpr/src/draw``.

#### Types of changes

#### Checklist:


tangxifan:
@vaughnbetz @litghost I have finished my patch on the GUI using the new node_side() APIs.
All the graphic proofs have been uploaded. This PR is ready for your review.

vpr/src/draw/draw.cpp
```diff
@@ -2955,19 +2963,83 @@ static void draw_pin_to_chan_edge(int pin_node, int chan_node, ezgl::renderer* g
 
     const t_grid_tile& grid_tile = device_ctx.grid[pin_rr.xlow()][pin_rr.ylow()];
     t_physical_tile_type_ptr grid_type = grid_tile.type;
-    VTR_ASSERT_MSG(grid_type->pinloc[grid_tile.width_offset][grid_tile.height_offset][pin_rr.side()][pin_rr.pin_num()],
-                   "Pin coordinates should match block type pin locations");
 
     float draw_pin_offset;
-    if (pin_rr.side() == TOP || pin_rr.side() == RIGHT) {
```
litghost, line 2962: I think this logic needs a comment.  Now that pin rr nodes have multiple edges, I don't think this logic is right.  I recommend you write a short comment explaining why you believe this is correct.

tangxifan: Indeed. In the later part of this code block, I have logics to identify which side to be used for GUI. Therefore, I move the offset calculation there.

litghost (approved):
LGTM

vaughnbetz (commented):
LGTM.  Only had one suggestion for a comment (same place Keith flagged).

vpr/src/draw/draw.cpp
tangxifan: @vaughnbetz No problem. I have added comments about why the positive/negative offset is needed

vaughnbetz:
This is all graphics, so it won't affect the nightly tests.  Will merge as soon as sanitized complete, although I think this code is not active in them either.

PR #1672, opened 2021-02-24
vpr: fix direct connections at the same tile location

acomodi:
#### Description

This PR solves an issue for which VPR cannot have directs within the same X, Y tile locations, specifically in heterogeneous tiles, where two different sub tiles have a direct connection (varying only in the sub_tile_offset).

#### Motivation and Context

With the addition in SymbiFlow of new kinds of tiles and primitives, which are mostly heterogeneous (at the same X, Y location live different kinds of sub tiles), there exists connectivity between these sub tiles which was not present and verified before.

For instance, there is the GTP tile which have the following sub tiles:
- GTP_CHANNEL (primitive)
- IPAD
- OPAD

The IPADs and OPADs are connected to the GTP_CHANNEL primitive through a direct connection, hence no routing CHANX/CHANY is present.

The changes in this PR address and fix some bugs corresponding to this situation.

Basic, strong tests.

#### Types of changes

#### Checklist:


litghost (changes requested):
Change itself looks fine, needs a new test to check both the old behavior and the fix

acomodi:
> needs a new test to check both the old behavior and the fix

@litghost Done

litghost (approved):
LGTM

acomodi:
@vaughnbetz Apart from some QoR failures that I see also on master, all the other Nightly tests passed. Is this good to merge?

vpr/src/route/rr_graph.cpp
```diff
@@ -2765,11 +2766,12 @@ static int get_opin_direct_connecions(int x,
                             }
                         }
 
+                        int target_sub_tile = z + directs[i].sub_tile_offset;
+                        if (relative_ipin >= target_type->sub_tiles[target_sub_tile].num_phy_pins) continue;
```
vaughnbetz, line 2770: It would be good to comment why this can occur (what leads to this continue firing)?

vaughnbetz (commented):
LGTM. I think there's one spot where a comment would be helpful (not obvious what the continue statement is handling -- maybe no tile of the right type there in the grid?).  This can be merged and then the comment added.

Issue #1666, opened 2021-02-17
vtr nightly runs not completing


sfkhalid:
The vtr nightly runs are not completing. They may not be running at all in some cases. For example, the nightly run in this pull request #1641 says that the tool failed and duration is 00:00. The same thing is happening after trying to run nightly a few times on this pull request. The same issue also seems to be occurring on other pull requests - #1663 and #1658 are some examples of this. 
Also, it is not possible to look at the invocation log to see what is happening because it says "Unable to retrieve invocation log"

#### Expected Behaviour
The nightly run should have an elapsed runtime and either pass or fail. If the nightly test fails, some reasons for the failure should be able to be seen in the invocation log (ex. to show if some QoR tests failed).

#### Current Behaviour
The nightly test gets stuck on the yellow status and after some time just shows failed and duration 00:00. All other checks seem to be working fine.

#### Possible Solution
Not sure of what might resolve this. First step is probably to run nightly test on a pull request and see what is happening on kokoro when it is running.

#### Steps to Reproduce
Run nightly test on a pull request.

#### Context
Haven't been able to merge code on a pull request because nightly is not passing.

#### Your Environment
Kokoro CI

#### Additional Information
When I run the nightly regression test on my local machine, I get the following error:
![image](https://user-images.githubusercontent.com/61606358/108148203-5483ce80-709e-11eb-81e0-169f3a60bdbc.png)

I ran make get_titan_benchmarks and make get_ispd_benchmarks to make sure I had the proper benchmarks to run nightly so the issue is not from that. Also, other reg tests such as basic and strong run fine.

mithro:
FYI - @litghost

sfkhalid:
Solution: clean up log files on CI and run nightly again
Link with instructions: 
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/45c83a2592923b9a785aa05c9464b5f453ebd151/.github/kokoro/package_results.sh#L43-L48

sfkhalid:
@litghost There seems to be an issue with vtr_reg_coffe having QoR failures that is contributing to the nightly test failures.

I first saw these failures in this pull request #1637. I thought they were odd since in this pull request, I only changed the order of how things are printed in the vpr log file.

[...]

PR #1674, opened 2021-02-25
Fix cleanup if `run_vtr_test.py` generates an error.

litghost:
#### Description

There is logic to cleanup the VTR working space after `run_reg_test.py`, but if the `e` flag (terminate on error) is set, then the cleanup logic never runs.  This change un-sets the `e` flag around `run_reg_test.py` and forwards the result.

#### Related Issue

Might help with #1666


mithro (approved):
LGTM.

mithro:
Merge if tests go green?

PR #1677, opened 2021-03-01
Fix missing include to unbreak build.

hzeller:
Fix broken build not finding fprintf() due to missing stdio.h include.


libs/libvtrutil/src/vtr_geometry.h
```diff
@@ -3,6 +3,8 @@
 #include "vtr_range.h"
 #include "vtr_assert.h"
 
+#include <stdio.h> // vtr_geometry.tpp uses printf()
```
litghost, line 6: Should probably be `cstdio`?

litghost (approved):
LGTM

libs/libvtrutil/src/vtr_geometry.h
hzeller: Looks like it is common in this directory, so let's use `<cstdio>`. Updated.
(usually, I recommend `stdio.h` for cleanly stating the origins, but seem that this part of the project is not doing that)

litghost: In C++, `cstdio` should always be preferred, at least by the standard: https://stackoverflow.com/questions/7596406/stdio-h-not-standard-in-c

hzeller: Yes, but it is important to know the rules to know when it is ok and better to break them :D

Issue #1681, opened 2021-03-05
VPR ignores some IO placement constraints


Xiretza:
A regression seems to cause VPR to ignore some IO placement constraints, producing a `.place` file with wrong positions.

#### Expected Behaviour
The constraints file is applied correctly.

#### Current Behaviour

With the following constraints file:

```
#Block Name  x   y   z    pcf_line
#----------- --  --  -    ----
$iopadmap$test.leds.t_inv  114  104  0  # set_property LOC H5 [get_ports {leds[0]}]
$iopadmap$test.leds_2.t_inv   2  156  0  # set_property LOC T9 [get_ports {leds[2]}]
$iopadmap$test.leds_3.t_inv   2  157  0  # set_property LOC T10 [get_ports {leds[3]}]
switches[2]   2  29  0  # set_property LOC C10 [get_ports {switches[2]}]
switches[1]   2  28  0  # set_property LOC C11 [get_ports {switches[1]}]
$iopadmap$test.leds_1.t_inv  114  106  0  # set_property LOC J5 [get_ports {leds[1]}]
switches[0]   2  26  0  # set_property LOC A8 [get_ports {switches[0]}]
switches[3]   2  30  0  # set_property LOC A10 [get_ports {switches[3]}]
```

executing the test script below on adbaf771e gives a wrong placement result:

```
Netlist_File: test.net Netlist_ID: SHA256:9344ad5632787c112f2ab44183554efb8375317a67ca1823fad565d9a4953214
Array size: 117 x 162 logic blocks

#block name	x	y	subblk	block number
#----------	--	--	------	------------
$iopadmap$test.leds.t_inv	114	104	0	#0
$iopadmap$test.leds_1.t_inv	114	106	0	#1
$iopadmap$test.leds_2.t_inv	2	156	0	#2
$iopadmap$test.leds_3.t_inv	2	157	0	#3
$iopadmap$test.other.t_inv	114	83	0	#4
switches[0]	2	26	0	#5
switches[1]	2	79	0	#6
switches[2]	2	132	0	#7
switches[3]	2	131	0	#8
$true		1	1	0	#9
```

As you can see, the placement of `switches[1..3]` is wrong.

Executing the same test on de62b152a gives the correct result:

```
Netlist_File: test.net Netlist_ID: SHA256:9344ad5632787c112f2ab44183554efb8375317a67ca1823fad565d9a4953214
Array size: 117 x 162 logic blocks

[...]

PR #1682, opened 2021-03-05
place: fix illegal swap of fixed blocks

acomodi:

#### Description

This PR fixes an issue for which a fixed block could get swapped during placement

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1681

#### Motivation and Context

Critical blocks and IOs cannot get swapped and must preserve the initial constraints.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/test/test_vpr_constraints.cpp
+++ b/vpr/test/test_vpr_constraints.cpp
@@ -1,6 +1,8 @@
 #include "../src/base/partition_region.h"
 #include "catch.hpp"
 
+#include "vpr_api.h"
+#include "globals.h"
 #include "vpr_constraints.h"
 #include "partition.h"
 #include "region.h"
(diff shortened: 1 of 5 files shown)


litghost (changes requested):
Change looks good, this needs a test case?

litghost:
> > this needs a test case?
> 
> I am trying to get this done, but as far as I can see there is no infrastructure stood up to check whether place constraints have been preserved in the vtr_flow, making it quite complex to add such a test.

Is there a way to compare place files against a golden?  That should be easy to add if its missing?

litghost:
@vaughnbetz / @jgoeders Any ideas to make @acomodi's test idea easier?  Or will @acomodi need to extend the test framework?

vaughnbetz:
The change looks good. @sfkhalid is adding checks for floorplanning constraints to check_placement().  I think right now you're testing that the in-memory constraints have been respected.  Another level of rigour in that test would be to reload the constraints from file to ensure they are respected; not sure if we need that or not.  
Your constraint test should flag this; a good test would be to insert a bug in placement where you move a locked block (or just force its location to change in the debugger) and make sure it fails your test.  A locked block has a specified x,y, and sub_tile.

jgoeders:
OK, keep me posted on how this goes.  If the overall flow needs some extra hooks in order to allow these types of comparisons I could try and find a student to do that.

acomodi:
I opted to get a C++ regression test using the vpr APIs and calling the placer with the input place constraints file. The test fails if the final placement does not respect the input constraints.

I have verified the test failing without the fix in this PR

vpr/test/test_vpr_constraints.cpp
```diff
@@ -421,3 +423,60 @@ TEST_CASE("RegionLocked", "[vpr]") {
...
+
+    vpr_initialize_logging();
+
+    // Command line arguments
+    const char* argv[] = {
+        "test_vpr",
+        kArchFile,
+        "wire.eblif",
+        "--fix_clusters", "wire.constraints",
+        "--place_static_move_prob", "0", "0", "0", "0", "0", "100", "0",
```
litghost, line 443: Can you explain why these arguments are here?

acomodi: As described [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/fe2bcfd16975cc1a93e057c8c9194f5d13ba4679/vpr/src/base/read_options.cpp#L1792-L1800), this option defines the probabilities to select a move type. The 6th field corresponds to the [feasible region constraints move type](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/place/feasible_region_move_generator.cpp), where there was no check on the `is_fixed` attribute of the selected block.

Also, the RL_agent seems to be enabled by default, and it does not implement the "feasible region constraints move type" hence I disabled it.

I'll update the PR with a comment on that.

acomodi:
The feasible region move has a fix already with this PR, and I also added an additional check when determining whether a move should be committed or not (and this is common for all types of moves), so the illegal swap behaviour should not happen anymore with this PR. If it is ok I'd proceed with merging this and opening a followup with maybe more C++ unit tests to cover more move types.

litghost:
@vaughnbetz / @MohamedElgammal Is this ready to be merged?  If not, what is left to be done?

MohamedElgammal:
@litghost I have added a fix to the critical uniform move as well. I think it is now good to go.

PR #1693, opened 2021-03-24
RR Graph Overlay Creation and Deploy new node_type() API of Overlay

tangxifan:
This PR kicks off the refactoring effort of the routing resource graph -related data structures.
A detailed technical plan can be found at [link](https://docs.google.com/document/d/15m7IbVRbQYLxFQjIVNIZT3VjYDhAqK6cjqCxVuSVDGU/edit?usp=sharing)

The overall refactoring effort aims to 
- create a unified data structure ``RRGraphView`` as a centralized storage for all the routing resource graph -related information. See Fig. 1
- create a set of frame view object from the centralized storage ``RRGraph`` for client functions, which are suitable for ``rr_graph builder``, ``placer``, ``router``, ``GUI``, ``timing analyzer`` etc. See Fig. 2.
- This is to avoid massive updates to the codes of client functions when there is a change on the unified data structure. 
- Also it avoid large memory footprint for client functions, since each client function may only use a small portion (typically <50%) APIs of the unified object.

image
Fig. 1 Illustration of the relationship between data structures

image
Fig. 2 Different levels of frame views of routing resource graphs to satisfy various needs from client functions.

The result/benefits of the refactoring efforts is
- The routing resource graph can be decoupled from VPR's core engine as a library. Unit tests can be enabled
- It is much easier for developers to develop custom routing resource graph builders thanks to the APIs of the unified data structure ``RRGraph``. A routing resource graph builder can be a library, decoupled from VPR's core engine. Many checking codes can be efficiently merged into the data structure and developers can save a lot of efforts in writing the atom-level sanity checks.
- Ensure that each client functions have a clean view of the routing resource graph, i.e. ``RRGraphView``. In other words, routing resource graph will be read-only and only accessors are exposed to client functions. Developers have no worries on developing their own placer/router etc.

#### Description

This PR adds:
- [x] A new object ``RRGraphBuilder`` as the writable data structure modeling a routing resource graph. It will encapsulate ``rr_nodes`` and ``rr_node_indices``. (To keep this PR small, currently just bring ``rr_node_indices`` in. This ``RRGraphBuilder`` will be **the one and only one interface** for routing resource graph builder functions.
- [x] A new object ``RRGraphView`` as the read-only data structure modeling a routing resource graph. It shares the data storage as the ``RRGraphBuilder``. This ``RRGraphView`` is designed to be a full featured frame-view for client functions. A few variants may be introduced in later pull requests, such as reduced frame-views for client functions.
- [x] The new object ``RRGraphBuilder`` and ``RRGraphView`` are added to ``DeviceContext`` and updated each time a routing resource graph is built
- [x] Develop a basic mutator in the ``RRGraphBuilder`` for updating the ``rr_node_indices``.
- [x] Deploy the mutator in routing resource graph building function. (May reduce to applying an example code change to validate the new object if massive code changes are required)
- [x] Develop a basic accessor in the ``RRGraphBuilder`` and ``RRGraphView`` for data queries on ``rr_node_indices``.
- [x] Deploy the accessor in placer
- [ ] Deploy the accessor in router (Deferred if massive code changes are required)

#### Types of changes

#### Checklist:


tangxifan:
@vaughnbetz @litghost 
I have updated the description of this PR and applied code changes, according to our discussion.
Please note that:
- I have not added the ``RRGraphBuilderView`` to ``DeviceContext`` because currently rr_node_indices builder creates a local copy of indices. Therefore, the  ``RRGraphBuilderView`` is created locally. 
- I have not added fancy methods for node fast look-up builder, such as deposit an array of default node ids, which can simplify the current rr_node_indices builders. This is to avoid massive code changes. Once we agree on the data structures, it is easy to do so.

vpr/src/device/rr_graph_builder_view.h
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Note that the builder does not own the storage
+ * It serves a virtual protocol for
+ * - rr_graph builder, which requires both mutators and accessors
+ *
+ * Note:
+ * - This is the only data structre allowed to modify a routing resource graph
+ *
+ */
+class RRGraphBuilderView : public RRGraphView {
```
litghost, line 28: No inheritance!

vaughnbetz:
April 1 discussion:
- Make a class that encapsulates RRSpatialLookup
- RRGraphView: has reference to this class, can return a const&.  Callers who want fast lookups can call the class using that const rrSpatialLookup&
- RRGraphBuilder: will also have a reference to the RRSpatialLookup data structure. It will mediate changes to the RRSpatialLookup; they'll be made as nodes are added or deleted. Could use dirty bits and bulk updates, or immediate updates.  Your call, as implementation needs dictate.

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,99 @@
...
+    /* The ONLY mutators allowed */
+    void set_internal_data(t_rr_graph_storage* node_storage,
+                           t_rr_node_indices* rr_node_indices);
+
+    /****************
+     * internal data storage, can be accessed by parent class RRGraphBuilderView
+     ****************/
+  protected:
+    /* node-level storage including edge storages */
+    t_rr_graph_storage* node_storage_;
```
hzeller, line 94: Strive for making members that are only set _once_ in the constructor, `const`.

```
t_rr_graph_storage* const node_storage_;
t_rr_node_indices* const rr_node_indices_;
```

That way, it is immediately clear for someone reading the code that these values will never be assigned to again, and it helps make it easier to reason about the code.

this also means, that you will have to have a
`RRGraphView(t_rr_graph_storage* , t_rr_node_indices*)` constructor that initializes via the initializer list, because otherwise assignment will not be possible anymore aftewards: So e.g. this works

```
RRGraphView(t_rr_graph_storage* storage, t_rr_node_indices* indices) 
    : node_storage_(storage), rr_node_indices_(indices) {}
```

While the following won't, as it attempts to assign things _after_ the initializer list:
```
RRGraphView(t_rr_graph_storage* storage, t_rr_node_indices* indices) {
  node_storage_ = storage;
  rr_node_indices_ = indices;
}
```
So using this technique of making as much `const` as can be (the stuff that will only be set once in the constructor) also lets us adopt the correct habit of initializing as much as possible in the constructor initializer list. And it is easy to find the one place where it happens -> overall improved code quality.

vpr/src/device/rr_graph_view.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /* Reaching here, it means that node exists in the look-up, return the id */
+    return RRNodeId((*rr_node_indices_)[type][node_x][node_y][node_side][ptc]);
+}
+
+/****************************
+ * Mutators
+ ****************************/
+void RRGraphView::set_internal_data(t_rr_graph_storage* node_storage,
```
hzeller, line 83: Is this something we want to allow ? It sounds like that is something that we want to have set once. After all, the `RRGraphView` is (at this point) merely a tuple of these pointers. And to make things easy to reason, we want to have data containers mostly immutable (common best practice).

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,99 @@
+#ifndef _RR_GRAPH_OVERLAY_
+#define _RR_GRAPH_OVERLAY_
+
+#include <exception>
+
+#include "rr_graph_fwd.h"
```
hzeller, line 6: (like in the other place, keep the headers sorted. Since this is a new file, we can start right away using best practices here)

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,99 @@
+#ifndef _RR_GRAPH_OVERLAY_
```
hzeller, line 1: The class you define here is called `RRGraphView`. The header is calls `rr_graph_view.h`. so you want the include guard also be called something like `RR_GRAPH_VIEW_H_`

Also, avoid _leading_ underscores, as any identifiers and in particular macros with leading underscores are reserved for internal use in C/C++.

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,99 @@
...
+     ****************/
+  public:
+    /* The ONLY mutators allowed */
+    void set_internal_data(t_rr_graph_storage* node_storage,
+                           t_rr_node_indices* rr_node_indices);
+
+    /****************
+     * internal data storage, can be accessed by parent class RRGraphBuilderView
+     ****************/
+  protected:
```
hzeller, line 92: `protected` on data members is what we rarely should have to do. If the derived classes need to have access to the raw members, make protected getters, but keep the members private.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1534,6 +1537,10 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         VTR_ASSERT(read_rr_graph_name_ != nullptr);
         read_rr_graph_filename_->assign(read_rr_graph_name_);
 
+        /* Update RRGraph viewer with new data pointers */
+        rr_graph_->set_internal_data(rr_nodes_,
```
hzeller, line 1541: See comment above, but this sounds like we want to avoid that and consider making a fresh `rr_graph_` with pointers that are only assigned once.

vpr/src/device/rr_graph_builder_view.h
```diff
@@ -0,0 +1,53 @@
+#ifndef _RR_GRAPH_BUILDER_VIEW_
```
hzeller, line 1: Avoid leading undrescores in the include-guards as they are reserved for c/c++ compiler internals.

vpr/src/device/rr_graph_builder_view.h
```diff
@@ -0,0 +1,53 @@
...
+#define _RR_GRAPH_BUILDER_VIEW_
+
+#include <exception>
```
hzeller, line 4: why do you include `<exception>` ? I don't see that needed anywhere below.

vpr/src/device/rr_graph_builder_view.h
```diff
@@ -0,0 +1,53 @@
...
+#include "rr_graph_fwd.h"
+#include "rr_node_fwd.h"
+#include "rr_graph2.h"
+#include "vtr_log.h"
+#include "vtr_memory.h"
+#include "vpr_utils.h"
+#include "vtr_strong_id_range.h"
+#include "vtr_array_view.h"
+#include "rr_graph_storage.h"
+#include "rr_graph_view.h"
```
hzeller, line 15: Below, the types `rr_graph_storage`, `rr_node_indices`, `RRNodeId`, `t_rr_type` and `e_side` are used. Do we have to include all the above to get access to these ?

IWYU, but not more.

hzeller (commented):
Hi, let me introduce myself: I am Henner, working with @litghost on Symbiflow FPGA stuff. Since Keith is going to work on other interesting things and can't be as much involved, I'll be helping with guidance and reviews on this project. 
I'll be in the next meetings, so then I'll say hello in person :)

WRT to this project, I still have to go through all the docs; in general we probably want to go in small steps, abstracting one aspect at a time, providing accessors instead of direct access to members etc. Small pull requests (100-250 lines max) are probably good stepping stones to transform the code-base into the right direction. Starting from transforming the 'consumer' part of the RR-graph to have more abstract ways to access it, then allows in further steps to address the builder to provide things we need.
Once I get into the project I can help guide towards a robust and future-compatible API.

So below, the comments are mostly directly the C++ specific comments that come to mind immediately and that are somewhat independent of wherever the project goes. I hope they can be valuable feedback even if the scope will change. Not using inheritance unless for interface implementations, initializing things as much as possible in the constructor initializer list, sorting headers, IWUY etc. are universal good practices that can be applied right away.

vpr/src/device/rr_graph_builder_view.cpp
```diff
@@ -0,0 +1,47 @@
...
+                                       t_rr_node_indices* rr_node_indices) {
+    node_storage_ = node_storage;
+    rr_node_indices_ = rr_node_indices;
+}
+
+/****************************
+ * Mutators
+ ****************************/
+void RRGraphBuilderView::add_node_to_fast_lookup(const RRNodeId& node,
+                                                 const int& x,
```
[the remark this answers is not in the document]
tangxifan: Agree. Updated the codes

vpr/src/base/vpr_context.h
```diff
@@ -9,6 +9,7 @@
 #include "vtr_vector.h"
 #include "atom_netlist.h"
 #include "clustered_netlist.h"
+#include "rr_graph_view.h"
```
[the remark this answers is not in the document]
tangxifan: It makes sense to me. I will avoid modifying the ``vpr_context.h`` because it is not the focus of this PR. But I will force the alphabetical order of header files in newly created files in this PR.

vpr/src/device/rr_graph_builder_view.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * Methods for Object RRGraphBuilderView
+ ***************************************/
+#include "vtr_assert.h"
+#include "rr_graph_builder_view.h"
+
+/****************************
+ * Constructors
+ ****************************/
+RRGraphBuilderView::RRGraphBuilderView(t_rr_graph_storage* node_storage,
+                                       t_rr_node_indices* rr_node_indices) {
```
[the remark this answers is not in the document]
tangxifan: Thanks for the comment. I was looking for something like this and failed to find related tutorials. I have updated the codes.

vpr/src/device/rr_graph_builder_view.h
```diff
@@ -0,0 +1,53 @@
...
+ */
+class RRGraphBuilderView : public RRGraphView {
+    /****************
+     * Constructors
+     ****************/
+  public:
+    RRGraphBuilderView(t_rr_graph_storage* node_storage,
+                       t_rr_node_indices* rr_node_indices);
+
+    /****************
```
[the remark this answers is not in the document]
tangxifan: The organization was actually suggested when prototyping the data structure RRGraphObject. We want to guide developers quickly to the constructors and accessors, which the developers care most. I am open to discussion still. This is mainly due to the suggested organization is not clear to me yet. We can have more discussion.

vpr/src/device/rr_graph_builder_view.h
```diff
@@ -0,0 +1,53 @@
...
+     * Accessors all come from RRGraph View
+     ****************/
+
+    /****************
+     * Mutators
+     ****************/
+  public:
+    /* Register a node in the fast look-up */
```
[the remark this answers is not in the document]
tangxifan: Indeed. I will see what comments to added.

vpr/src/device/rr_graph_view.h
tangxifan: Agree. Updated.

vpr/src/device/rr_graph_view.cpp
tangxifan: Agree. Removed

vpr/src/device/rr_graph_view.h
tangxifan: Agree

vpr/src/device/rr_graph_view.h
tangxifan: Agree. Updated.

vpr/src/device/rr_graph_view.h
tangxifan: Agree. Changed to private now.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
tangxifan: Agree. Removed.

vpr/src/device/rr_graph_builder_view.h
tangxifan: Agree. Updated.

vpr/src/device/rr_graph_builder_view.h
tangxifan: Agree. Removed all the unused headers.

vpr/src/device/rr_graph_builder_view.h
tangxifan: Agree. Removed all the unused headers.

vpr/src/device/rr_graph_builder.h
```diff
@@ -0,0 +1,45 @@
...
+    t_rr_graph_storage& node_storage();
+    /* Return a writable object for update the fast look-up of rr_node */
+    RRSpatialLookup& node_lookup();
+
+    /****************
+     * internal data storage
+     ****************/
+  private:
+    /* node-level storage including edge storages */
+    t_rr_graph_storage& node_storage_;
```
hzeller, line 40: Only use references if they are const. If you have a modifiable object, use pointers instead (essentially like you did before).

Using non-const references creates a lot of confusion, in particular if used as instance variables.

tangxifan: I fully agree this convention. This rule can improve code readability and help developers to know what they are doing easily. However, as you see in the figure I showed (copied to this PR as well), the builder function will be the owner of the data storages. That is why I temporarily use the reference here. It can avoid a lot of code changes once the refactoring is finished (there is no function get data directly through the ``node_storage`` in ``DeviceContext``). My concern is that if I use pointers, it may cause many codes in client functions or inside the data structures to be changed later. 

Here are my concerns and I am open to discussion.

tangxifan:
Hi @hzeller @litghost @vaughnbetz 
   
Just a summary what we talked and what we agreed today:

- Work should be done on RRGraphView to provide fast access to internal data. But it will be schedule after the current refactoring on builder functions is finished. 
- Current refactoring effort will not modify the data structure organizations. In the end, the internal data storage should be organized efficiently for client functions. But this will be scheduled after the current refactoring on builder functions is finished. 
- Keep only one issue to solve per PR
- Avoid large code changes per PR

From my point of view, we can plan the refactoring in this way:
- I recommend to let this PR focused on a simple ``RRGraphBuild`` and ``RRGraphView`` as well as ``RRSpatialLookup``, as a proof of the concept about the data structures that we all agree on.
- If we all agree that the ``RRSpatialLookup`` is a right way to implement the node-level fast look-up, I can deploy the ``RRSpatialLookup`` in builder functions and client functions, as follow-up PRs. 
 
After that, we have achieved the milestone where the builder has been improved to avoid direct write/read on the lookup data structure. 
And we can discuss if we should prioritize the API addition to ``RRGraphView`` or ``RRGraphBuild``.

Looking forward to your feedbacks.

vaughnbetz:
That makes sense to me.  
I think a logical order is:
1. Create RRSpatialLookup, as a wrapper/re-factoring around the current (data-intrusive, no API) spatial lookup code.
2. Next I think we want RRBuilder to have an API that allows us to do what the builder code already does: add nodes, edges, do bulk updates.  I'd focus on giving the functionality we already use, but with an API around it instead of direct data structure manipulation. 
3. Start improving APIs / removing direct data structure manipulation.  Pick one at a time.  Might be the geometry we talked about today, might be a cleaner RRBuilder api for node and edge addition (or maybe we'll be happy with stage 2 already).

tangxifan:
Hi @vaughnbetz @hzeller 
  This PR is ready for your review now.   
   - [X] Added the ``RRSpatialLookup`` with an example API ``add_node`` to register a node to lookup
   - [X] Added the ``RRGraphBuilder`` which owns an ``RRSpatialLookup`` object
   - [X] Deployed ``RRGraphBuilder``  to a routing resource graph building function, to showcase the use and functional correctness
   - [X] Added the ``RRGraphView`` which shares the storage with ``RRGraphBuilder`` object
   - [X] Deployed ``RRGraphView``  to a placement function, to showcase the use and functional correctness
 
   This PR showcased the use of the new data structures in both builder and client functions

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -0,0 +1,106 @@
...
+                               e_side side) {
+    VTR_ASSERT_SAFE(3 == rr_node_indices_[type].ndims());
+
+    /* Expand the fast look-up if the new node is out-of-range
+     * This may seldom happen because the rr_graph building function
+     * should ensure the fast look-up well organized  
+     */
+    VTR_ASSERT(type < rr_node_indices_.size());
+
+    if ((size_t(x) >= rr_node_indices_[type].dim_size(0))
```
hzeller, line 92: We should probably `VTR_ASSERT` above that `x` and `y` are both >= 0. Otherwise casting it to an unsigned type might mean that `size_t(x)` is always essentially >= whatever we're comparing it to.

Or we do the other way around, casting `rr_node_indices_[type].dim_size(1)` to `int` and compare with the signed `x`; then it is more likely doing what we mean.

hzeller (commented):
Looks pretty good so far.

vpr/src/base/vpr_context.h
```diff
@@ -145,6 +146,12 @@ struct DeviceContext : public Context {
     /*
      * Structures to define the routing architecture of the FPGA.
      */
+    RRGraphView rr_graph = RRGraphView(rr_nodes, rr_spatial_lookup); // A read-only view of routing resource graph to be the ONLY database for client functions: GUI, placer, router, timing analyzer etc.
```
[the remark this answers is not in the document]
tangxifan: Make senses to me. Changed.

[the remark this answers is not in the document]
tangxifan: Make sense to me. I have moved the ``RRGraphView`` and ``RRSpatialLookup`` after the ``rr_node_indices``, so that they create references only after the dependent datastuctures have been initalized.

vpr/src/device/rr_graph_builder.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ */
+class RRGraphBuilder {
+    /****************
+     * Constructors
+     ****************/
+  public:
+    RRGraphBuilder(t_rr_graph_storage& node_storage,
+                   RRSpatialLookup& node_lookup);
+
```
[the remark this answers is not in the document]
tangxifan: You are right. The protection rules should be made! Changed both RRGraphBuilder and RRGraphView

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,62 @@
...
+
+    /* Return a read-only object for performing fast look-up in rr_node */
+    const RRSpatialLookup& node_lookup() const;
+
+    /****************
+     * internal data storage
+     ****************/
+  private:
+    /* node-level storage including edge storages */
+    t_rr_graph_storage& node_storage_;
```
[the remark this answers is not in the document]
tangxifan: Make sense. It will be clear and avoid any recurring engineer in later stages. Changed

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,62 @@
...
+  public:
+    RRGraphView(t_rr_graph_storage& node_storage,
+                RRSpatialLookup& node_lookup);
+
+    /****************
+     * Accessors
+     ****************/
+  public:
+    /* Get the type of a routing resource node */
+    t_rr_type node_type(const RRNodeId& node) const;
```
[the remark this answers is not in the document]
tangxifan: Agree. Changed

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,91 @@
...
+ * The data structure allows users to 
+ * - Update the look-up with new nodes
+ * - Find the id of a node with given information, e.g., x, y, type etc.
+ ********************************************************************/
+class RRSpatialLookup {
+    /****************
+     * Constructors
+     ****************/
+  public:
+    RRSpatialLookup(t_rr_node_indices& rr_node_indices);
```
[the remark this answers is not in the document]
tangxifan: Yes. Made changes.

vpr/src/route/rr_graph2.cpp
```diff
@@ -982,15 +990,15 @@ static void load_block_rr_indices(const DeviceGrid& grid,
...
                 //Assign indices for SINKs and SOURCEs
                 // Note that SINKS/SOURCES have no side, so we always use side 0
-                for (const auto& class_inf : type->class_inf) {
-                    auto class_type = class_inf.type;
+                for (size_t iclass = 0; iclass < type->class_inf.size(); ++iclass) {
+                    auto class_type = type->class_inf[iclass].type;
                     if (class_type == DRIVER) {
-                        indices[SOURCE][x][y][0].push_back(*index);
-                        indices[SINK][x][y][0].push_back(OPEN);
+                        rr_graph_builder.node_lookup().add_node(RRNodeId(*index), x, y, SOURCE, iclass, SIDES[0]);
```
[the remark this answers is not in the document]
tangxifan: Actually, I prefer to create the ``RRGraphBuilder::add_node()`` API in later pull request.
The detailed reasons are:
- The ``RRGraphBuilder::add_node()`` API is not a simple chained call using ``RRSpatialLookup``. It will have two functionality: (a) add a node to the ``node_storage``; and (b) register the node in the ``RRSpatialLookup``.
- Current builder functions are written in the way that the addition to ``node_storage`` is separated from the addition to  ``RRSpatialLookup``. If we created the ``RRGraphBuilder::add_node()`` API, it will cause massive changes on the  builder functions. So, I created the new API, it will not be useful in builder functions without changing.

That is why I prefer to do it in later pull request.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,91 @@
...
+     * - type is the type of a node
+     * - ptc is a feature number of a node, which can be
+     *   1. pin index in a tile when type is OPIN/IPIN
+     *   2. track index in a routing channel when type is CHANX/CHANY
+     * - side is the side of node on the tile, applicable to OPIN/IPIN 
+     *
+     * Note that the add node here will not create a node in the node list
+     * You MUST add the node in the t_rr_node_storage so that the node is valid  
+     */
+    void add_node(RRNodeId node,
```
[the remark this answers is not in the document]
tangxifan: I think it is a good approach. We can surely try it in later pull request. I will add comments to the header file as a reminder.

vpr/src/device/rr_spatial_lookup.h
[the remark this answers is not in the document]
tangxifan: Agree. I also learned how to use the new keyword ``explicit``

vpr/src/device/rr_spatial_lookup.cpp
tangxifan: Make senses to me. Changed.

vpr/src/device/rr_graph_view.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+/****************************
+ * Accessors
+ ****************************/
+t_rr_type RRGraphView::node_type(RRNodeId node) const {
+    return node_storage_.node_type(node);
+}
+
+const RRSpatialLookup& RRGraphView::node_lookup() const {
+    /* Return a constant object rather than a writable one */
```
hzeller, line 23: mmh, I am not a fan of comments stating the obvious. Problem with that is of course, that comments become noise instead of actual useful information, so readers of the code might learn to ignore comments even though it might contain something useful.

Also, the vertical space used is a lot, so that makes it harder to navigate the code (similar with the massive three-line comments about constructors and accessors which should't really be needed as anyone sees that anyway - but it sounds like this was decided for some other reason ? Still not a fan :) ).

vpr/src/device/rr_graph_builder.h
```diff
@@ -0,0 +1,63 @@
...
+ */
+class RRGraphBuilder {
+    /****************
+     * Constructors
+     ****************/
+  public:
+    RRGraphBuilder(t_rr_graph_storage* node_storage,
+                   RRSpatialLookup* node_lookup);
+
+    /* Disable copy constructors
```
hzeller, line 26: This looks like a lot of comment for something fairly common.
I think the disabled copy constructor and assignment operator can be there without comment as they speak for themselve.

vpr/src/device/rr_graph_builder.h
```diff
@@ -0,0 +1,63 @@
...
+ * Note:
+ * - This is the only data structre allowed to modify a routing resource graph
+ *
+ */
+class RRGraphBuilder {
+    /****************
+     * Constructors
+     ****************/
+  public:
+    RRGraphBuilder(t_rr_graph_storage* node_storage,
```
hzeller, line 23: Might be useful to add a comment about what the parameters are (even if it changes in the future, but as reference for the reader).

vpr/src/device/rr_graph_builder.h
```diff
@@ -0,0 +1,63 @@
...
+ * - node_lookup: store a fast look-up for the nodes
+ *
+ * Note:
+ * - This is the only data structre allowed to modify a routing resource graph
+ *
+ */
+class RRGraphBuilder {
+    /****************
+     * Constructors
+     ****************/
```
hzeller, line 21: ... Still not a fan of these massive comments. They don't really serve a purpose (as for a well-structured class it is immediately clear where is what), so they are mostly just [counter the goal to reduce vertical whitespace](https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace). We have only so much vertical space on our monitors, so we should try to maximize information.

If it helps readability I  sometimes do that with a single line comment that looks a bit emphasized, maybe with dashes:
```
  // --- Constructors
```
Sounds like these large comments were decided before me, but I'd strongly encourage to re-evaluate :)

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,71 @@
...
+ */
+class RRGraphView {
+    /****************
+     * Constructors
+     ****************/
+  public:
+    RRGraphView(const t_rr_graph_storage& node_storage,
+                const RRSpatialLookup& node_lookup);
+
+    /* Disable copy constructors
```
hzeller, line 42: Similar here: these constructors can just be two lines without comment.

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -0,0 +1,108 @@
...
+    VTR_ASSERT_SAFE(3 == rr_node_indices_[type].ndims());
+
+    /* Sanity check to ensure the x, y, side and ptc are in range 
+     * Data type of rr_node_indice:
+     *   typedef std::array<vtr::NdMatrix<std::vector<int>, 3>, NUM_RR_TYPES> t_rr_node_indices;
+     * Note that x, y and side are the 1st, 2nd and 3rd dimensions of the vtr::NdMatrix
+     * ptc is in the std::vector<int>
+     */
+    if (size_t(type) >= rr_node_indices_.size()) {
+        /* Node type is out of range, return an invalid index */
```
hzeller, line 47: I'd avoid comments that just restate exactly what the code does. They will train readers to ignore comments, which is dangerous. And they use up vertical space, at least 6 lines alone here when counting also the `/* Reaching here, must...`

To further reduce vertical space usage, you can consider the this if permitted by this project's readability rules: simple return based on on a condition is more compact (and as readable) with the following:
```
   if (size_t(type) >= rr_node_indices_.size()) return RRNodeId::INVALID();
```

So, where interesting stuff is happening, we should emphasize it with code and comments, but boring things are nice when compact.

vpr/src/route/rr_graph2.cpp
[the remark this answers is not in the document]
hzeller: SGTM.

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -0,0 +1,108 @@
...
+    /* Find actual side to be used */
+    e_side node_side = side;
+    if (type == IPIN || type == OPIN) {
+        VTR_ASSERT_MSG(side != NUM_SIDES, "IPIN/OPIN must specify desired side (can not be default NUM_SIDES)");
+    } else {
+        VTR_ASSERT(type != IPIN && type != OPIN);
+        node_side = SIDES[0];
+    }
+
+    /* Currently need to swap x and y for CHANX because of chan, seg convention */
```
hzeller, line 31: _this_ is a place where I'd expect a more detailed comment, as this is clearly something that can surprise.

hzeller (commented):
This looks pretty good to me.
Mostly my comments are now down to comments that are too many and other comments that are too few :)

vpr/src/device/rr_graph_view.h
```diff
@@ -0,0 +1,71 @@
...
+    RRGraphView(const RRGraphView&) = delete;
+
+    /* Disable copy assignment operator */
+    void operator=(const RRGraphView&) = delete;
+
+    /****************
+     * Accessors
+     ****************/
+  public:
+    /* Get the type of a routing resource node */
```
hzeller, line 55: Probably good to add a TODO comment saying that the goal is to have these accessors private at some point totally replaced with the 'questioin' kind of methods.

vpr/src/device/rr_graph_view.cpp
tangxifan: I think you are right. I have not written the comments carefully so that there are some redundant comments. Should remove this line. Also, I would like to share my opinions about writing comments (feel free to bug and share your experience)
- The comments about API's usage/functionality should be left in the header file, being close to the API declaration.
- The comments in source file should **NOT**  repeat what header files say. The comments in source file should focus on details of an API implementation, which is not easy to follow.

As such, developers can focus on reading header files for the key data structure and only look into specific functions in source files when necessary.

hzeller: Fully agree.

vpr/src/device/rr_graph_builder.h
tangxifan: I left the comment about the copy constructor is because it is the first time I used these C++ syntax. 
I did so because I assume that there could be many guys do not know the syntax as well. 
On the other sides, we cannot expect every developer is an expert of a C++ when reading the source codes (considering VPR is a research project, it is perhaps the first C++ project that students touched).
Some comments may be appreciated for these non-popular syntax.
Would you mind share your experience? 
My idea is to find a good trade-off in the comments between entry-level engineers and experts.

vpr/src/device/rr_graph_builder.h
tangxifan: Yes. Since we will comment each private data structure, I suggest just simply redirect developers to the comments in the internal data storage section of the header file.

vpr/src/device/rr_graph_builder.h
tangxifan: Agree. These dummy lines should be removed.

vpr/src/device/rr_spatial_lookup.cpp
tangxifan: Yes. I removed all the redundant comments. 
Regarding the 1-line if-clause, I personally do not have a strong opinion on it. I am not used to read 1-line if-clauses. Unless it is necessary, I prefer if-clauses with brackets.

vpr/src/device/rr_spatial_lookup.cpp
tangxifan: Indeed. More explanation has been added.

vpr/src/device/rr_graph_view.h
tangxifan: Yes. The TODO comment is useful. But not every API will be converted to the ``questionin`` methods. But the ``questionin`` style is preferred.

tangxifan:
@hzeller I have applied your comments on the code comments.

@kmurray If you have time to review this PR, feel free to comment. 
Here is a brief of the PR's contribution. Hope it can make your reviewing process easier.
- [X] Added the ``RRSpatialLookup`` with an example API ``add_node`` to register a node to lookup
- [X] Added the ``RRGraphBuilder `` to ``DeviceContext`` which owns an ``RRSpatialLookup`` object, but only provide mutators.
- [X] Deployed ``RRGraphBuilder`` to a routing resource graph building function, to showcase the use and functional correctness
- [X] Added the ``RRGraphView``  to ``DeviceContext`` which shares the storage with ``RRGraphBuilder`` object, but only provide accessors.
- [X] Deployed ``RRGraphView`` to a placement function, to showcase the use and functional correctness

vpr/src/device/rr_graph_builder.h
hzeller: No worries, leave it as you feel comfortable.

As for the additional explanation
```
 This is to avoid any duplication of the object
 as it is only interface allowed to modify routing resource graph
```

I'd probably emphasize that we want to avoid an accidental copy not because of the interface situation, but because (once the builder actually owns the data), it will be a very expensive operation, copying gigabytes of data. If we prevent copying, then accidental 'pass-by-value' are immediately caught by the compiler for instance.

vpr/src/device/rr_spatial_lookup.cpp
hzeller: SGTM as well.

vpr/src/device/rr_graph_builder.h
tangxifan: No problem. I think it is more accurate. Changed.

vpr/src/route/rr_graph2.cpp
```diff
@@ -970,8 +971,12 @@ static void load_chan_rr_indices(const int max_chan_width,
     }
 }
 
-static void load_block_rr_indices(const DeviceGrid& grid,
+/* As the rr_indices builders modify a local copy of indices, use the local copy in the builder
```
vaughnbetz, line 974: I don't quite understand this.  There is both a local couple of the rr_indices data structure and another copy in the RRGraphBuilder.  It looks like some data is put in the version in the builder (with add_node) and some is in the local copy (with direct data structure access). I'm not sure where the local copy is copied back to the RRGraphBuilder copy.

It would be good to clean this up sooner rather than later.  The circular dependency in the header files (rr_graph2.h etc.) should be resolvable by refactoring the header files / making other headers.  Ideally we'd just use the builder functions when adding nodes to the rr_indices data structure, and would have only one copy.

vaughnbetz (changes requested):

vaughnbetz:
@tangxifan : just finished reviewing this. Most of my feedback is minor (requests for comments or clarification), but I think there's one substantive thing: I'm not sure how the two copies of rr_indices (one in builder, one local) are synced up, and I think it would be good to get down to just one copy (in the RRGraphBuilder) sooner rather than later.

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -0,0 +1,93 @@
...
+                                    int y,
+                                    t_rr_type type,
+                                    int ptc,
+                                    e_side side) const {
+    /* Find actual side to be used */
+    e_side node_side = side;
+    if (type == IPIN || type == OPIN) {
+        VTR_ASSERT_MSG(side != NUM_SIDES, "IPIN/OPIN must specify desired side (can not be default NUM_SIDES)");
+    } else {
+        VTR_ASSERT(type != IPIN && type != OPIN);
```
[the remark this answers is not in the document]
tangxifan: Right. The assertation is just a safety check. Currently, I prefer to changing it to VTR_ASSERT_SAFE, which will not be considered unless in debug mode (no runtime cost in release mode). I am also o.k. to remove it. Please let me know what you think.

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -0,0 +1,93 @@
...
+        node_side = SIDES[0];
```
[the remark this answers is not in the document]
tangxifan: Agree. Added comments to clarify this.

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -0,0 +1,93 @@
...
+        return RRNodeId::INVALID();
+    }
+
+    if (node_y >= rr_node_indices_[type].dim_size(1)) {
+        return RRNodeId::INVALID();
+    }
+
+    if (node_side >= rr_node_indices_[type].dim_size(2)) {
+        return RRNodeId::INVALID();
+    }
```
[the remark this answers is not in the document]
tangxifan: Agree. However, the ``node_x``, ``node_y`` are already in the ``size_t`` data type, they cannot be less than 0 (converted from ``x ``, ``y`` which are ``int``). If ``x`` is negative, when converted to ``size_t``, it will be a huge number and out-of-range in the fast look-up. But it is really tricky. So, I added a pre-check on the ``x``,  ``y``, ``node_side`` and ``ptc`` to return invalid ids when one of them is negative.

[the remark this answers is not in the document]
vaughnbetz: Ah, I didn't think of that.  In that case I'm OK with just a comment noting that will happen.  If you've added the pre-check for < 0 already that's fine too.

vpr/src/device/rr_spatial_lookup.cpp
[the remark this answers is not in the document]
vaughnbetz: OK, your call.  I'm OK with VTR_ASSERT_SAFE if you prefer, although I personally don't put asserts right in an if-chain where it is obvious the assert can't fire based on the if/else condition.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,97 @@
+#ifndef RR_SPATIAL_LOOKUP_H
+#define RR_SPATIAL_LOOKUP_H
+
+#include "vpr_types.h"
+
+/******************************************************************** 
+ * A data structure storing the fast look-up for the nodes
```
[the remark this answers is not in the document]
tangxifan: Indeed. Reworked the comments.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,97 @@
...
+     * This is to avoid any duplication of the object
+     * as it is only interface allowed to access node look-up of a routing resource graph
+     */
+    RRSpatialLookup(const RRSpatialLookup&) = delete;
+    void operator=(const RRSpatialLookup&) = delete;
+
+    /* -- Accessors -- */
+  public:
+    /* Returns the index of the specified routing resource node.  (x,y) are
+     * the location within the FPGA, rr_type specifies the type of resource,
```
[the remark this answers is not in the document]
tangxifan: Comment applied.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,97 @@
...
+     * and ptc gives the number of this resource.  ptc is the class number,
+     * pin number or track number, depending on what type of resource this
+     * is.  All ptcs start at 0 and go up to pins_per_clb-1 or the equivalent.
+     * There are class_inf size SOURCEs + SINKs, type->num_pins IPINs + OPINs,
+     * and max_chan_width CHANX and CHANY (each).
+     *
+     * Note that for segments (CHANX and CHANY) of length > 1, the segment is
+     * given an rr_index based on the (x,y) location at which it starts (i.e.
+     * lowest (x,y) location at which this segment exists).
+     * This routine also performs error checking to make sure the node in
```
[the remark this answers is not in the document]
tangxifan: Comment applied.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,97 @@
...
+    RRSpatialLookup(const RRSpatialLookup&) = delete;
+    void operator=(const RRSpatialLookup&) = delete;
+
+    /* -- Accessors -- */
+  public:
+    /* Returns the index of the specified routing resource node.  (x,y) are
+     * the location within the FPGA, rr_type specifies the type of resource,
+     * and ptc gives the number of this resource.  ptc is the class number,
+     * pin number or track number, depending on what type of resource this
+     * is.  All ptcs start at 0 and go up to pins_per_clb-1 or the equivalent.
```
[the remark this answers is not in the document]
tangxifan: Agree. I have reworked the comments to provide more insights on the ptc number depending on the ``node_type``.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,97 @@
...
+     *   1. a valid one is to register a node in the lookup
+     *   2. an invalid id means the removal of a node from the lookup
+     * - (x, y) are the coordinate of the node to be indexable in the fast look-up
+     * - type is the type of a node
+     * - ptc is a feature number of a node, which can be
+     *   1. pin index in a tile when type is OPIN/IPIN
+     *   2. track index in a routing channel when type is CHANX/CHANY
+     * - side is the side of node on the tile, applicable to OPIN/IPIN 
+     *
+     * Note that the add node here will not create a node in the node list
```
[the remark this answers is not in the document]
tangxifan: Comment applied.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,97 @@
...
+                       t_rr_type type,
+                       int ptc,
+                       e_side side = NUM_SIDES) const;
+
+    /* -- Mutators -- */
+  public:
+    /* Register a node in the fast look-up 
+     * - You must have a node id
+     *   1. a valid one is to register a node in the lookup
+     *   2. an invalid id means the removal of a node from the lookup
```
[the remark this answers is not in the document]
tangxifan: You are right. The use of invalid id is very tricky. It just unregisters a node from the lookup.
Actually, I should not state it because we should not promote such use of the API.
Therefore, I removed this line. We should add an API ``remove_node()``. I put this TODO item in the comment.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,97 @@
...
+     * Note that for segments (CHANX and CHANY) of length > 1, the segment is
+     * given an rr_index based on the (x,y) location at which it starts (i.e.
+     * lowest (x,y) location at which this segment exists).
+     * This routine also performs error checking to make sure the node in
+     * question exists.
+     *
+     * The 'side' argument only applies to IPIN/OPIN types, and specifies which
+     * side of the grid tile the node should be located on. The value is ignored
+     * for non-IPIN/OPIN types
+     */
```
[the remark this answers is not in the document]
tangxifan: Agree. Comment applied.

vpr/src/route/rr_graph2.cpp
tangxifan: This is indeed confusing. We actually share the data pointers between the ``rr_indices`` and ``RRGraphBuilder`` when initializing the ``DeviceContext``.  So the two data structures are always synchronized.

https://github.com/tangxifan/vtr-verilog-to-routing/blob/4a8443b135bdbd97d8e97da73f33ce2cedd0d7e6/vpr/src/base/vpr_context.h#L166-L180

Currently, the data ownership remains in the ``rr_indices`` because I avoid to modify many codes in rr_graph2.cpp which will make this PR tedious. I just modify a very small part to showcase that the new data structure is working fine. 

And you are right. This code should be cleaned up ASAP. I also stated in the TODO of the comment in LINE 975. I will start massively deployment of the ``RRGraphBuilder`` in the function ``alloc_and_load_rr_indices()`` in next PR.

tangxifan:
@vaughnbetz Thanks a lot for the constructive comments. I have updated the PR and clarify as much as I can. Feel free to bug if you see anything confusing.

vpr/src/route/rr_graph2.cpp
vaughnbetz: OK.  I think you should comment the pointer aliasing right now though, so it's understandable how it works at this point. I didn't figure out that you'd aliased the pointers (both point at the same storage) and that's why the code worked.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -0,0 +1,109 @@
...
+     * - You must have a valid node id to register the node in the lookup
+     * - (x, y) are the coordinate of the node to be indexable in the fast look-up
+     * - type is the type of a node
+     * - ptc is a feature number of a node, which can be
+     *   - the class number of a common SINK/SOURCE node of grid, 
+     *   - pin index in a tile when type is OPIN/IPIN
+     *   - track index in a routing channel when type is CHANX/CHANY
+     * - side is the side of node on the tile, applicable to OPIN/IPIN 
+     *
+     * An invalid id will be returned if the node does not exist
```
vaughnbetz, line 74: I think this comment (an invalid id will be returned if the node does not exist) is in the wrong spot.  It should be with find_node, not add_node.

vpr/src/device/rr_spatial_lookup.h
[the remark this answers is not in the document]
vaughnbetz: I think the comment is in the wrong spot (add_node, but should be in find_node).

vaughnbetz (commented):
Looks good Xifan, thanks.  I added a few follow-up comments (one comment is in the wrong spot, think pointer aliasing of rr_node_indices should be commented even though it's a temporary fix).  Other than that looks good to go.

vaughnbetz:
Spoke too soon: looks like CI also caught a bug in an assert:  
	</home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/vpr/src/device/rr_spatial_lookup.cpp:83 add_node: Assertion 'node' failed.

vpr/src/device/rr_spatial_lookup.h
tangxifan: You are right. I moved to the find_node() API.

tangxifan:
@vaughnbetz I have patched the comment and clarify the dirty sync between ``rr_node_indices`` and ``rr_spatial_lookup``: 

https://github.com/tangxifan/vtr-verilog-to-routing/blob/49e3339488e2f02a3b07df3ec53a6fa0b7f53a43/vpr/src/base/vpr_context.h#L166-L172

I will keep watching the CI and fix anything red.

tangxifan:
@vaughnbetz CI is green now. The pull request is ready for your review again.

vaughnbetz:
Updating branch (pretty sure no related changes went in, so this should be fine).  I'll merge as soon as the fast running tests pass and show no compile or basic errors; no need to wait for vtr_reg_nightly I think.

vaughnbetz:
Everything passed except vtr nightly was in progress; updated again due to other changes.  *really* will merge once the fast running tests are done.

vaughnbetz:
The Odin-II failure is unrelated; due to moving some runs onto kokoro (in progress).  This has passed vtr_nightly before so merging it now (it's been rebased twice already).

PR #1667, opened 2021-02-20
Deploy new node.side() API to rr_graph builder and router

tangxifan:
This PR aims to deploy the new node_side() API to rr_graph builder and router.

#### Description

This PR:
- [X] Replace all the node.side() API with node.is_node_on_specific_side() API in rr_graph builder and router.
- [ ] Patch routers' lookahead map building for multi-width and multi-height blocks. This will be patched in later pull requests, because key roadblocks have been identified: require prev_node list to be cached for each rr_node so that lookahead builder can know accurate positions for SINK nodes.

#### Related Issue

To address issue #1580

#### Motivation and Context

In Titan architecture, a ``rr_node`` may appear in multiple sides of a logic block. However, VPR's rr_graph builder and router only assumes that a ``rr_node`` locates on only 1 side. There has been patches (#1621 and #1661) to extend the side support for ``rr_node`` and deploy to GUIs. This PR is a follow-up PR as planned to deploy the new API node.is_node_on_specific_side() in place of the to-be-deprecated API node_side() in the rr_graph builder and router.
After this PR, the old API can be safely removed.

#### Types of changes

- [X] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1694, opened 2021-03-24
Deprecate node_side() API

tangxifan:
This is the final PR in the PR series #1667 and #1661  to replace node_side() API with is_node_on_specific_side() APIs

#### Description

- [X] Deprecate node_side() API
- [x] Deprecate node_sides() API

#### Related Issue

To address issue #1577 

#### Motivation and Context

To address issue #1577 

Should have no impact to existing codes as the node_side() API is almost unused.

#### Types of changes

#### Checklist:


vpr/src/route/rr_node.cpp
```diff
@@ -25,7 +25,13 @@ const char* t_rr_node::direction_string() const {
 }
 
 const char* t_rr_node::side_string() const {
```
litghost, line 27: Probably need to deprecate this one too?

vaughnbetz: Seems reasonable to deprecate.

tangxifan: @litghost @vaughnbetz The method has been deprecated

vaughnbetz:
Thanks Xifan; looks good.  Will merge as soon as CI goes green.

PR #1697, opened 2021-03-26
update golden results for vtr_req_qor_chain(_depop) and titan_quick_qor

sdamghan:
#### Description

Regenerating new golden results for vtr_req_qor_chain_depop which is the cause of vtr_reg_nightly failure in CI

#### Related Issue

Vtr_reg_nightly failure

#### Types of changes

#### Checklist:


sdamghan:
Hi @vaughnbetz ,
I have updated the vtr_reg_qor_chain and vtr_reg_qor_chain_depop which were caused of vtr_nightly failure. The arm_core failure appeared in my second PR related to the expression support (b0debed516a38e4241979985a89a930ddcad07c5). This is actually why we did not see the arm_core failure in the implicit_memory PR (f38aa8f3eb023008542a02b5fd492eaf5e4b67fd).
Moreover, I re-updated titan_quick_qor. There was only a mismatch for minres in **setup_WNS** parameter. TBH, I have no idea why only this parameter did not update in the previous PR while all others got updated. Anyway, everything is now good with a green checkmark :)
Please let me know your thoughts.

sdamghan:
@vaughnbetz 
Actually, I have checked the arm_core testbench through Odin. Although, in the implicit memory PR, I added the behaviour to Odin to not infer a SPRAM as a DPRAM when it has both read and write access in the Verilog code, the results of the arm_core benchmark did not change. The arm_core includes only SPRAMs. These SPRAMs appeared in Odin stat after the expression support PR, while they should have appeared after the implicit memory PR. This is actually because I used the memory-connected signals' net name (unique names) to check whether a signal is connected to a memory port or not. In the second PR, I needed to use net names again, but this time was more general since I needed access not only for memory but for all variables. I noticed that several nets did not have names (they connect to a memory after multiple stages, like being buffered), so I added names for them. As a result, the second PR detects more similar memory connectivities (net names are not NULL anymore) and did the same scenario that we expected in the first PR. Anyway, my second PR fixed a bug from the previous one automatically.

sdamghan:
The added net names resulted in more memory optimization:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/bdf51aa9b1338de5b630414fe80bff5018a4680b/ODIN_II/SRC/netlist_create_from_ast.cpp#L2709

vaughnbetz:
Great, thanks for the detailed investigation Seyed!  Merging.

sdamghan:
Anytime! BTW, I should have investigated more when I created the second PR. However, because we already had issues with VTR nightly test detecting such behaviour was a bit confusing and tricky. Anyway, sorry about the long term failure :)

vaughnbetz:
No worries, glad it is sorted now and we should be back to a clean CI.  Having CI fail is risky!

PR #1701, opened 2021-03-31
Updated files for artifact collection to include zipped files in the …

sfkhalid:
…nightly config file

The configuration files for the regression tests needed to be updated to include files ending with ".gz", since the output files from nightly files are zipped. This change is needed because no data is being collected because the filters for artifact collection are missing ".gz"


vaughnbetz:
Updating to latest branch. Prior CI had (known I think) failures in the symbiflow tests due to a timing problem (missing) in the arch file.  Should be gone now.

PR #1704, opened 2021-04-08
Floorplanning during place moves

sfkhalid:
#### Description
Adding floorplanning constraints checks to the placement move generators. When a move is being proposed, the blocks affected by the move are checked to see if they are moved out of their respective floorplan regions. If a move would cause a block to be moved out of its legal floorplan region, the move is aborted.

#### Related Issue
Changes related to issue #932


vpr/src/place/centroid_move_generator.cpp
```diff
@@ -35,5 +36,19 @@ e_create_move CentroidMoveGenerator::propose_move(t_pl_blocks_to_be_moved& block
...
+    e_create_move create_move;
+    create_move = ::create_move(blocks_affected, b_from, to);
+
+    //Check if the move is legal from a floorplan perspective
+    //Check that all of the blocks affected by the move would still be in a legal floorplan region after the swap
+    bool floorplan_legal = true;
+    for (int i = 0; i < blocks_affected.num_moved_blocks; i++) {
+        floorplan_legal = cluster_floorplanning_legal(blocks_affected.moved_blocks[i].block_num, blocks_affected.moved_blocks[i].new_loc);
+        if (!floorplan_legal) {
+            VTR_LOG("Move aborted for block %zu, location tried was x: %d, y: %d, subtile: %d \n", size_t(blocks_affected.moved_blocks[i].block_num), blocks_affected.moved_blocks[i].new_loc.x, blocks_affected.moved_blocks[i].new_loc.y, blocks_affected.moved_blocks[i].new_loc.sub_tile);
```
sfkhalid, line 48: make this log message only for verbose - it will affect cpu time

vpr/src/place/centroid_move_generator.cpp
```diff
@@ -35,5 +36,19 @@ e_create_move CentroidMoveGenerator::propose_move(t_pl_blocks_to_be_moved& block
         return e_create_move::ABORT;
     }
 
-    return ::create_move(blocks_affected, b_from, to);
+    e_create_move create_move;
+    create_move = ::create_move(blocks_affected, b_from, to);
+
+    //Check if the move is legal from a floorplan perspective
+    //Check that all of the blocks affected by the move would still be in a legal floorplan region after the swap
+    bool floorplan_legal = true;
```
sfkhalid, line 44: make this chunk of lines a helper function in place_constraints.cpp so that you can just get this with one line

sfkhalid: and pass blocks_affected by const reference

sfkhalid: can make it an inline function in the header file - say it is performance critical

vpr/src/base/read_options.cpp
```diff
@@ -1413,6 +1413,10 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .show_in(argparse::ShowIn::HELP_ONLY);
 
     file_grp.add_argument(args.read_vpr_constraints_file, "--read_vpr_constraints")
+        .help("Writes the floorplanning constraints from placement to the specified XML file.")
```
sfkhalid, line 1416: Switch the comments for read and write

vpr/src/base/vpr_constraints_writer.h
```diff
@@ -0,0 +1,12 @@
+/*
+ * vpr_constraints_writer.h
+ *
+ *      Author: khalid88
+ */
+
+#ifndef VPR_SRC_BASE_VPR_CONSTRAINTS_WRITER_H_
+#define VPR_SRC_BASE_VPR_CONSTRAINTS_WRITER_H_
+
```
sfkhalid, line 9: Add overview comment explaining what file is for

vpr/src/place/feasible_region_move_generator.cpp
```diff
@@ -122,5 +123,28 @@ e_create_move FeasibleRegionMoveGenerator::propose_move(t_pl_blocks_to_be_moved&
             return e_create_move::ABORT;
     }
 
-    return ::create_move(blocks_affected, b_from, to);
+    /*e_create_move create_move;
```
sfkhalid, line 126: Remove commented out code

vpr/src/place/weighted_median_move_generator.cpp
```diff
@@ -101,7 +102,16 @@ e_create_move WeightedMedianMoveGenerator::propose_move(t_pl_blocks_to_be_moved&
     if (!find_to_loc_centroid(cluster_from_type, from, w_median_point, range_limiters, to)) {
         return e_create_move::ABORT;
     }
-    return ::create_move(blocks_affected, b_from, to);
+    //return ::create_move(blocks_affected, b_from, to);
```
sfkhalid, line 105: remove commented line

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+void write_vpr_floorplan_constraints(const char* file_name, int expand, int subtile) {
+    auto& place_ctx = g_vpr_ctx.placement();
+    auto& atom_ctx = g_vpr_ctx.atom();
+    auto& cluster_ctx = g_vpr_ctx.clustering();
+
+    //Fill in all the info needed for locking atoms to their locations
+    VprConstraints constraints;
+
+    std::map<ClusterBlockId, std::vector<AtomBlockId>> cluster_atoms;
```
sfkhalid, line 28: have another helper routine for filling in the VprConstraints object

vpr/src/base/read_options.cpp
sfkhalid: make a little more detailed, what floorplanning constraints (read floorplanning constraints that packing and placement must respect from the specified XML file)

sfkhalid: writes out new floorplanning constraints based on current placement to the specified xml

vpr/src/base/read_options.cpp
```diff
@@ -1873,6 +1877,21 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
             "Its range equals to [0., 1.].")
         .default_value("0.7")
         .show_in(argparse::ShowIn::HELP_ONLY);
+
+    place_grp.add_argument(args.place_constraint_expand, "--place_constraint_expand")
+        .help(
+            "The value used to decide how much to expand the floorplan constraint region when writing"
```
sfkhalid, line 1883: add in that the range is 0 to inf, and that it is an int. ( an int greater than or equal to zero that controls how specific the floorplan constraint region written out is, if a floorplan constraint file is generated)

sfkhalid: say what zero means (same x, y location) and give an example of what a number higher than zero would do (like if you had 1)

vpr/src/base/read_options.cpp
```diff
@@ -1873,6 +1877,21 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
...
+            "a floorplan constraint XML file. If it is zero, the block stays locked in its place. If it is"
+            "greater than zero the constraint region expands by the specified value in each direction.")
+        .default_value("0")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
+    place_grp.add_argument(args.place_constraint_subtile, "--place_constraint_subtile")
+        .help(
+            "The value used to specify a subtile constraints when writing a floorplan constraints XML file.")
```
sfkhalid, line 1891: should be a boolean option (on or off) otherwise you will have illegal placements (because some blocks can be at the same x, y but diff subtile)

sfkhalid: if it is off, don't print subtile, if it is on, you print what is in current placement

vpr/src/base/vpr_api.cpp
```diff
@@ -660,11 +660,17 @@ void vpr_place(t_vpr_setup& vpr_setup, const t_arch& arch) {
...
 
     auto& filename_opts = vpr_setup.FileNameOpts;
+    auto& placer_opts = vpr_setup.PlacerOpts;
     auto& cluster_ctx = g_vpr_ctx.clustering();
 
     print_place(filename_opts.NetFile.c_str(),
                 cluster_ctx.clb_nlist.netlist_id().c_str(),
                 filename_opts.PlaceFile.c_str());
+
+    //Write out a vpr floorplanning constraints file if the option is specified
```
sfkhalid, line 670: move to vpr_place_flow

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -65,6 +70,11 @@ struct VprConstraintsContextTypes : public uxsd::DefaultVprConstraintsContextTyp
 
 class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstraintsContextTypes> {
   public:
+    //VprConstraintsSerializer() : report_error_(nullptr) {}
+    VprConstraintsSerializer(VprConstraints constraints)
+        : constraints_(constraints)
```
sfkhalid, line 75: pass in by reference instead of copying in (the constraints object)

sfkhalid: when you call the writer, call with a reference to the constraints object so that it doesn't copy the whole thing. Can also make it a pointer.

sfkhalid: Don't put it into the constructor if you don't have to (because during loading you don't need to pass in a constraints object). Instead, pass it by reference to the top level writing routine.

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -49,11 +49,16 @@
...
+    Partition part;
+    std::vector<AtomBlockId> atoms;
+    PartitionId part_id;
+};
 
 struct VprConstraintsContextTypes : public uxsd::DefaultVprConstraintsContextTypes {
-    using AddAtomReadContext = void*;
-    using AddRegionReadContext = void*;
-    using PartitionReadContext = void*;
+    using AddAtomReadContext = AtomBlockId;
```
sfkhalid, line 59: comment that readcontext is for writing file and vice versa, can email Fahri and verify that it is from point of view of memory

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -256,12 +269,20 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
...
     }
     virtual inline size_t num_partition_list_partition(void*& /*ctx*/) final {
-        int i = 0;
-        return i;
+        return constraints_.get_num_partitions();
     }
 
-    virtual inline void* get_partition_list_partition(int /*n*/, void*& /*ctx*/) final {
-        return nullptr;
+    virtual inline partition_info get_partition_list_partition(int n, void*& /*ctx*/) final {
```
sfkhalid, line 275: maybe comment this one, n is the partition id, get all the data for partition n

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -298,16 +319,16 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
...
 
     //temp data structures to be loaded during file reading
     Region loaded_region;
     Partition loaded_partition;
     PartitionRegion loaded_part_region;
-    VprConstraints constraints_;
 
     //temp string used when a method must return a const char*
-    std::string temp_;
+    std::string temp_ = "test";
```
sfkhalid, line 331: can take this out

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -298,16 +319,16 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
     }
 
     //temp data for loads
```
sfkhalid, line 321: comment is out of date

sfkhalid: comment what it is, what it does for load and write cases.
Just have two constructors so you don't have to pass in an empty object for the load case.

vpr/src/base/vpr_constraints_serializer.h
sfkhalid: disregard above, discussed that constraints object is need for both load and write. But have two constructors so you don't have to pass in an empty object when doing load_vpr_constraints

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
+/*
```
sfkhalid, line 1: have file level comment, doxygen compatible

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "globals.h"
+#include "pugixml.hpp"
+#include "pugixml_util.hpp"
+#include "echo_files.h"
+
+#include <fstream>
+#include "vpr_constraints_writer.h"
+
+void write_vpr_floorplan_constraints(const char* file_name, int expand, int subtile) {
```
sfkhalid, line 20: subtile will change to bool

vpr/src/base/vpr_constraints_writer.cpp
sfkhalid, line 28: could also just do this with a vtr::vector, would be faster and smaller

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+    VprConstraints constraints;
+
+    std::map<ClusterBlockId, std::vector<AtomBlockId>> cluster_atoms;
+
+    for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
+        cluster_atoms.insert({blk_id, std::vector<AtomBlockId>()});
+    }
+
+    for (auto atom_blk_id : atom_ctx.nlist.blocks()) {
+        ClusterBlockId clb_index = atom_ctx.lookup.atom_clb(atom_blk_id);
```
sfkhalid, line 35: this lookup may be looking into a vector that is just like the map I'm using, take a look

sfkhalid: make sure there is no data structure that is already doing this, if there is, make a utility function that does this on call. the utility function can go in with some kind of clb netlist utils file,

vpr/src/base/vpr_constraints_writer.h
sfkhalid: change subtile to bool, explain what all the variables are for

vpr/src/base/vpr_constraints_writer.cpp
sfkhalid: also look at clb netlist to see if it has this

sfkhalid: can add to clustered_netlist_utils if you add the utility function

sfkhalid: can also make a small class whose constructor creates the vector as a private member, and some function to say give me the atoms in this cluster, can add this class to clustered_netlist_utils.cpp

sfkhalid: for clusters_echo, will need to use this vector as well, so go change that code in cluster.cpp

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        cluster_atoms[clb_index].push_back(atom_blk_id);
+    }
+
+    int part_id = 0;
+    for (auto i = cluster_atoms.begin(); i != cluster_atoms.end(); i++) {
+        std::string part_name;
+        part_name = cluster_ctx.clb_nlist.block_name(i->first);
```
sfkhalid, line 43: this will become simpler once you make it a vector, can just use the indices

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+        Partition part;
+        part.set_name(part_name);
+
+        PartitionRegion pr;
+        Region reg;
+
+        auto loc = place_ctx.block_locs[i->first].loc;
+
+        reg.set_region_rect(loc.x - expand, loc.y - expand, loc.x + expand, loc.y + expand);
+        reg.set_sub_tile(subtile);
```
sfkhalid, line 56: add some logic around here for when subtile is bool

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    for (auto atom_blk_id : atom_ctx.nlist.blocks()) {
+        ClusterBlockId clb_index = atom_ctx.lookup.atom_clb(atom_blk_id);
+
+        cluster_atoms[clb_index].push_back(atom_blk_id);
+    }
+
+    int part_id = 0;
+    for (auto i = cluster_atoms.begin(); i != cluster_atoms.end(); i++) {
```
sfkhalid, line 41: add some comments on what this loop is doing

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+        int num_atoms = i->second.size();
+
+        for (auto j = 0; j < num_atoms; j++) {
+            AtomBlockId atom_id = i->second[j];
+            constraints.add_constrained_atom(atom_id, partid);
+        }
+        part_id++;
+    }
+
+    VprConstraintsSerializer reader(constraints);
```
sfkhalid, line 71: change name to writer

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    if (vtr::check_file_name_extension(file_name, ".xml")) {
+        std::fstream fp;
+        fp.open(file_name, std::fstream::out | std::fstream::trunc);
+        fp.precision(std::numeric_limits<float>::max_digits10);
+        void* context;
```
sfkhalid, line 77: comment why you are passing a context in, also make it null to make it more clear

vpr/src/base/vpr_types.h
```diff
@@ -1064,6 +1065,8 @@ struct t_placer_opts {
     //int place_timing_cost_func;
     std::string place_reward_fun;
     float place_crit_limit;
+    int place_constraint_expand;
```
sfkhalid, line 1068: add short comments if the other options have comments

vpr/src/place/place_constraints.cpp
```diff
@@ -36,7 +36,7 @@ bool is_cluster_constrained(ClusterBlockId blk_id) {
     return (!pr.empty());
 }
 
-bool is_macro_constrained(t_pl_macro pl_macro) {
+bool is_macro_constrained(t_pl_macro& pl_macro) {
```
sfkhalid, line 39: for all the things being passed by reference, make it a const reference to indicate that you are doing this for efficiency

sfkhalid: also better for the compiler to have it const if you are not changing it

vpr/src/place/place_constraints.h
```diff
@@ -23,16 +24,31 @@ bool is_cluster_constrained(ClusterBlockId blk_id);
 /*
  * Check if the placement location would respect floorplan constraints of the block, if it has any
  */
-bool cluster_floorplanning_check(ClusterBlockId blk_id, t_pl_loc loc);
+bool cluster_floorplanning_legal(ClusterBlockId blk_id, t_pl_loc& loc);
```
sfkhalid, line 27: make them all const ref

vpr/src/place/place_constraints.h
```diff
@@ -23,16 +24,31 @@ bool is_cluster_constrained(ClusterBlockId blk_id);
...
  * Check whether any member of the macro has floorplan constraints
  */
-bool is_macro_constrained(t_pl_macro pl_macro);
+bool is_macro_constrained(t_pl_macro& pl_macro);
 
 /*
  * Returns region of valid locations for the head of the macro based on floorplan constraints
  */
-PartitionRegion constrained_macro_locs(t_pl_macro pl_macro);
+PartitionRegion constrained_macro_locs(t_pl_macro& pl_macro);
```
sfkhalid, line 37: make it a const ref

vpr/src/place/place_constraints.h
```diff
@@ -23,16 +24,31 @@ bool is_cluster_constrained(ClusterBlockId blk_id);
...
+
+inline bool floorplan_legal(t_pl_blocks_to_be_moved& blocks_affected) {
+    bool floorplan_legal = true;
```
sfkhalid, line 40: don't need to set this here, delete

sfkhalid: just keep the declaration

vpr/src/base/clustered_netlist_utils.h
```diff
@@ -27,4 +27,15 @@ class ClusteredPinAtomPinsLookup {
     vtr::vector<AtomPinId, ClusterPinId> atom_pin_connected_cluster_pin_;
 };
 
+class ClusterAtomsLookup {
```
sfkhalid, line 30: comment what this is for

sfkhalid: what the private data is, automatically intialized, order 1 lookups, comment data

sfkhalid: state that it should only be called after clustered netlist is created

vaughnbetz:
1. Comment the new cluster_atoms_lookup class and its data/functions.
2. Need to add a testing summary. Minimum testing to put this back: ensure nothing that used to work is broken.
  - QoR before and after this change (attach a spreadsheet and summary on the VTR benchmarks), with no floorplan constraints. Everything should be identical, except CPU time, which should be extremely close.
  - Run the sanitizers on the new code (1) with no floorplan constraints, (2) when generating a floorplan file (3) when running with an (easy/big) floorplan file to confirm memory is clean.
3. Additional testing, that can be done after this is merged if it holds things up, but can be done now if it doesn't hold things up:
  - Run VTR benchmarks with a big floorplan region on each atom (shouldn't change anything significantly)
------------------------
_4. After that, QoR testing can proceed **post-merge**
  - Run VTR benchmarks with each atom constrained to a specific x,y,sub_tile.  Should work, but more chance something goes wrong and needs algorithm updates.
  - Moderate size floorplan regions_

vaughnbetz:
@sfkhalid : can you add the testing summary above so we can get this merged?  Also, close any comments that you've addressed in the code review.

vaughnbetz:
machine load can impact time. Marius killed a bunch of stale processes, so the load may be lower.  It's best to run the baseline and the new run under as similar conditions as you can (either run them at the same time, or run one after the other in rapid sucession, and watch the machine load some).

sfkhalid:
Testing with sanitizers:

I ran vpr on some test constraint cases, and some cases with no constraints on this branch and the master branch.
There were some memory leaks that occurred, however, the same memory leaks occurred on both the master and new branch, and so I don't think they are cause by the code in this pull request. 

mem_leaks_1
mem_leaks_2
mem_leaks_summary

vaughnbetz:
Looks good.  I agree the sanitizer errors aren't related to your code.  I believe we suppress those ones in CI (we can't make them go away as they're in libraries).  QoR looks OK.
I'll merge once CI goes fully green.
@sfkhalid: it's not in this PR, but did you also make the change to only check the the constraint on one block of a macro (could be head or just the block passed in) when you try to move a place macro?  Probably cleanest to just have the floorplan_legal check only check the one block passed in, and comment why that works for place_macros (and make sure you have constriant propagation fully done!).

sfkhalid:
@vaughnbetz Can we re-start the tests on this PR? The CI had gone completely green before but I've had to restart it a couple of times because by the time it finishes it needs to be updated with the new master branch. Some of the clang tests are failing now, I think they will pass if we run it again though.

sfkhalid:
Actually, looking at the log files on the clang tests, it seems there is an error coming up. I am not sure where it is coming from, this PR passed those tests twice in the past few days, so it may be from something that was updated recently in the master branch.

vaughnbetz:
Strange errors: looks like clang can't find basic things like size_t?  I set kokoro:force_rerun to try restarting CI, but looks like that only re-runs the big cpu tests, not clang tests. I'm confident this is unrelated so I'm merging anyway; hopefully this was a transient clang install issue.

PR #1706, opened 2021-04-11
Removing agilex_arch.xml

aman26kbm:
Removing the agilex_arch.xml file that was added a few months back.

#### Description
1. The naming of this file isn't correct. This is not really an agilex architecture. The main thing that's agilex like is the complex DSP slice. Based on recent discussions with Andrew Boutros, we prefer to rename this something like k6_N10*complex_dsp*.xml.

2. During some recent work, we found some issues with this architecture and fixed them. And we also created different variations of this architecture. These updated arch files will be checked in into the repo soon.

3. This arch is only used in the vtr_reg_coffe regression test, which is currently disabled. I'll enable vtr_reg_coffe when we checkin the new architectures. So, it's safe to delete this file.

#### Types of changes

#### Checklist:


aman26kbm:
@vaughnbetz , please review and approve.

Issue #1699, opened 2021-03-29
[ODIN] Remove the requirement that module instantiation port assignment order must match port order in module definition.


dlrauch15:
Right now, ODIN requires that port assignments in module instantiations must match the order in which the ports are defined in their module definitions. While this is a simple requirement to satisfy for hand-written RTL designs, it is a challenge when dealing with large HLS-generated designs that can be 10s of thousands of lines long with 10s to 100s of module instantiations that do not obey this rule.

@sdamghan @aman26kbm

sdamghan:
@dlrauch15 , would you please attach a small Verilog sample code plus relevant input/output files? If memory helps me this wasn't a problem. Odin can only do either port connection by name OR by ordered list. However, it cannot perform the combination of them.
```
module module_x (input a, output q);
---------------------------------------------------------
module_x uut (.in(a), .out(q) ) -> working
module_x uut (.out(q), .in(a) ) -> working
module_x uut ( a, q ) -> working
module_x uut (.in(a), q ) -> does not work
```

aman26kbm:
Hi @sdamghan , just wanted to clarify. The issue is not with port connection by name or ordered list. The issue is with the requirement of matching the order in which the ports are defined in the module definition and the order in which they are provided on the module instantiation.

Example:
Let's say this is the module definiton:
module mymodule(
input x,
input y,
output z);
...
endmodule

Or it is this:
module mymodule(x,y,z);
input x;
input y;
output z;
...
endmodule

While instantiating:

If we use:
mymodule inst (.x(), .y(), .z())
then it works.

But if we use:
mymodule inst(.z(), .y(), .x())  //Note that the order is not x,y,z
then it fails.

@dlrauch15 please correct me if I am not explaining it correctly.

dlrauch15:
That is correct but it sounds like Seyed is saying he believes that works (see the second line in his provided code). I will investigate this more next week.

dlrauch15:
Hi @sdamghan,

I confirmed that mixed port instantiation order works for simple cases, but not more complex ones. Here is a minimal test case that reproduces the issue:

```
module m (
   output wire o1,
   input wire i1,
   input wire i2,
   output wire o2
);
   assign o1 = ~i1;
   assign o2 = ~i2;
endmodule

module top (
   input wire i1,
   input wire i2,
   output wire o1,
   output wire o2
);

   m m_inst (
     .i1(i1),
     .i2(i2),
[...]

PR #1710, opened 2021-04-17
[Odin]: Sorting the instance ports based on the related module, function or task

sdamghan:
#### Description

Previously, the simple swap of two items at a time has been used as the sort approach for instance port connection, to reorder them according to the main module (function or task) that they are instantiated from. That would cause an error since we alter the array of ports each time, which results in an unexpected change in the array for future swapping. The new approach uses a helper array to order the ports and then reorder the instantiated instance ports.

#### Related Issue

 #1699

#### Motivation and Context

bug fix

1. make test (ODIN-II)
2. Basic Regression Tests
3. Strong Regression Test
4. Basic Valgrind Memory Tests (also, Valgrind memory tests for all other regression suits)
5. Sanitized Basic Regression Tests
6. Odin-II Micro Tests

#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/SRC/netlist_create_from_ast.cpp
+++ b/ODIN_II/SRC/netlist_create_from_ast.cpp
@@ -5418,12 +5418,25 @@ signal_list_t* create_hard_block(ast_node_t* block, char* instance_name_prefix, 
     return return_list;
 }
 
+/**
+ * --------------------------------------------------------------------------
+ * (function: reorder_connections_from_name)
+ * 
+ * @brief reorder the instantiated instance ports based on 
+ * the order of the main item(module, function or task)
+ * 
+ * @param instance_node_list including the list of the main item ports
+ * @param instantiated_instance_list including the list of the instantiated instance ports
+ * @param ids the type of the main item or instance (module, function or task)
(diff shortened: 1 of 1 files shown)


sdamghan:
@aman26kbm and @dlrauch15
The bug was related to using a simple swap at a time for re-ordering instance ports. So, by swapping two ports each time, the main array was altered, and afterwards, the array was not valid for the previously achieved indexes for swapping. I have provided the solution in this PR.  
The provided test case in the issue is passed. Would you please have it confirmed based on the other failure cases? 

Please let me know if you have any thoughts about the code as well.

sdamghan:
@vaughnbetz this is ready for merge, please let me know your thoughts :)

vaughnbetz:
@sdamghan: Looks good. Thanks for the new commenting.
If no automatic test caught this, could one be created and added to the CI regtests?

sdamghan:
> @sdamghan: Looks good. Thanks for the new commenting.
> If no automatic test caught this, could one be created and added to the CI regtests?
@vaughnbetz 
Yes, for sure. I'll add an appropriate test to regtests.

PR #1712, opened 2021-04-20
There is a mistake on the true dual port mode of BRAM.

ZhangHao-Oreo:
#### Description
--------------------- Describe your changes in detail ----------------

<mode name="mem_2048x8_dp">
   I think it may be a mistake. The memory mode should be 4096x8. In this mode, I replace 2048x8 with 4096x8.
Moreover, I find the same mistake in other architecture files. For example k6_frac_N10_frac_chain_mem32K_40nm.xml.


vaughnbetz:
That does look wrong -- thanks!  Will merge once CI goes green again (had to update to the latest branch so CI kicked off again).

Issue #1711, opened 2021-04-19
vtr_reg_nightly run failures


sfkhalid:
vtr_reg_nightly is currently having six run failures. 
#### Expected Behaviour
All of the runs should pass.

#### Current Behaviour

Six runs are failing. 

The runs that are failing are all run with similar command line arguments. Many command line arguments are specified, including --fix_clusters and --sdc_file.

The failures can be seen in the following log file:
[log_with_nightly_failures.log](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6337279/log_with_nightly_failures.log)

#### Possible Solution

1. The failures could possibly be occurring because the files that are specified with the vpr command line options are not in the specified directories. For example, the command specifies a file /vtr_flow/benchmarks/symbiflow/sdc/picosoc_basys3_full_100.sdc, but I do not see this file in the directory when I check the repo on github.
 
2. The failure could also be coming from the move generators moving blocks that are marked as fixed. Only one of the move generators (feasible region move generator) has a check to see whether the block about to be moved is fixed. The check can be seen here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/aeddc0229b47b22de3b80ae93ec51f8399060b87/vpr/src/place/feasible_region_move_generator.cpp#L23-L25

Even if the failures are not coming from the second issue, this check should be added to all the move generators.
#### Steps to Reproduce

Run vtr_reg_nightly CI

sfkhalid:
@acomodi Are the sdc files used in the commands mentioned in point 1 coming in from the symbiflow repo? I don't see the sdc files in the directory /vtr_flow/benchmarks/symbiflow/sdc/ on github and I was thinking that may be the reason for the vtr_reg_nightly failures

sfkhalid:
@MohamedElgammal Could you add in the necessary checks in the move generators as mentioned in point 2? This may be the source of the vtr_reg_nightly failures, but also the .is_fixed check should be added to each move generator anyway.

vaughnbetz:
@kgugala can you take a look?  As you noted in the meeting, there is an error embedded in a warning (which is itself a bug I think), and it flags some issue with the arch.xml.

MohamedElgammal:
@sfkhalid The checks are done in the pick_from_block() function that is used to select the moving block for all moves except the feasible_region and the critical_unifrom moves. That's why we added the check to both of them explicitly.
[...]

PR #1713, opened 2021-04-23
place: fix case in which PR region does not have constraints

acomodi:

#### Description

The new partition region code might not take into account that partition regions are not being set and constraints are not initialized. This PR is an attempt at fixing this.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1711

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/place_constraints.cpp
+++ b/vpr/src/place/place_constraints.cpp
@@ -31,8 +31,13 @@ int check_placement_floorplanning() {
 /*returns true if cluster has floorplanning constraints, false if it doesn't*/
 bool is_cluster_constrained(ClusterBlockId blk_id) {
     auto& floorplanning_ctx = g_vpr_ctx.floorplanning();
-    PartitionRegion pr;
-    pr = floorplanning_ctx.cluster_constraints[blk_id];
+
+    auto keys = floorplanning_ctx.cluster_constraints.keys();
+    auto res = std::find(keys.begin(), keys.end(), blk_id);
+
+    if (res == keys.end()) return false;
+
+    auto pr = floorplanning_ctx.cluster_constraints[blk_id];
     return (!pr.empty());
 }
 


acomodi:
cc @sfkhalid @vaughnbetz

sfkhalid:
@acomodi During clustering each new cluster block is assigned an empty PartitionRegion to start with, which can then be either populated with constraint regions or not populated with constraint regions. I think this check is a good one to have, but I am not sure if it is the cause of the issue

acomodi:
@sfkhalid Ok, I'd need to investigate a bit more and find the source of the issue. The question I am trying to answer is also why this behaviour was not caught previously by CI, so either some change not related to the Partition Regions might have indirectly caused the root issue that leads to this failure.

One thing I wanted to test was whether this was a fix in the right direction, but kokoro did not start for some reason

vaughnbetz (changes requested):
I don't think we want the new code. cluster_constraints is stored as a vector indexed by cluster block number.  So this change makes the code longer and I think takes us from an O(1) lookup to an O(n) search.

vaughnbetz:
I looked into the .keys() implementation. The code change would definitely take us from O(1) to O(N) in code that can be performance critical.  We should not make this change.

acomodi:
@vaughnbetz Sure, I will change the fix. As it turns out, there is some other kind of problem preventing the Nightly build to complete successfully.

vaughnbetz:
@acomodi, @sfkhalid  : I think this fix is no longer needed, as Sarah now has put back code to ensure the original function (which checked if you had a partition region on a block) works in all cases / flows. @sfkhalid : can you confirm that the code is now safe (will return false) for blocks with no partition region?  If so, we can close this PR and not put back the code (which looks slow).

sfkhalid:
> @acomodi, @sfkhalid : I think this fix is no longer needed, as Sarah now has put back code to ensure the original function (which checked if you had a partition region on a block) works in all cases / flows. @sfkhalid : can you confirm that the code is now safe (will return false) for blocks with no partition region? If so, we can close this PR and not put back the code (which looks slow).

Yes, this fix is no longer needed. The function returns false for blocks with an empty partition region. This will work in any flow because there is a routine now that assigns each cluster a partition region (empty or otherwise if it has constraints) in the case where the pack stage is loaded rather than performed.

vaughnbetz:
Closing this PR as unneeded.

PR #1714, opened 2021-04-26
VPR test when starting from placement stage

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

#### Description

This is to test an initial placement failure case described in https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1711

PR #1715, opened 2021-04-27
Add routine to initialize cluster_constraints data structure in the c…

sfkhalid:
…ase where packing stage of VPR is skipped

In the vpr api, if the packing stage of vpr is skipped, the cluster_constraints data structure, which stores the floorplanning constraints of each block is initialized with an empty PartitionRegion for each cluster block.

#### Related Issue
This is an attempt to solve the issues discussed in #1711. Currently, there are seg faults during initial placement when individual stages of VPR are run. This is causing the nightly tests to fail.

#### Motivation and Context
Normally, the cluster_constraints data structure is loaded during packing. If packing is not done, the structure will remain empty, leading to seg faults when it is used in initial placement. This code should solve that since it loads the structure with empty PartitionRegions if the packing stage of vpr is skipped.


acomodi:
This LGTM, I believe that the corresponding regression test from https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1714 might be added to this PR, so that it can catch possible regressions when packer and placer are called separately (probably it would be good to have something similar for the router).

> This is causing the nightly tests to fail

Actually, we'd need to dig a bit more about the nightly failures, as the issue fixed by this PR is not actually causing the failure itself. Not sure why the timing model of the architecture causes issues in CI. For instance, the current architecture model that is responsible for the failure is the IPAD_GTP_VPR:

```
    <model name="IPAD_GTP_VPR">
      <input_ports>
        <port name="I"/>
      </input_ports>
      <output_ports>
        <port name="O"/>
      </output_ports>
    </model>
```

It does not have any timing specification and the model is used in this pb type:

```
      <pb_type name="IPAD" num_pb="1">
        <output name="O" num_pins="1"/>
        <pb_type blif_model=".subckt IPAD_GTP_VPR" name="IPAD_GTP" num_pb="1">
          <input name="I" num_pins="1"/>
          <output name="O" num_pins="1"/>
          <metadata>
            <meta name="type">bel</meta>
          </metadata>
        </pb_type>
        <pb_type blif_model=".input" name="inpad" num_pb="1">
          <output name="inpad" num_pins="1"/>
          <metadata>
            <meta name="type">bel</meta>
            <meta name="subtype">input</meta>
          </metadata>
        </pb_type>
        <interconnect>
          <direct input="IPAD_GTP.O" name="IPAD_GTP_to_IPAD" output="IPAD.O"/>
          <direct input="inpad.inpad" name="inpad.inpad_to_IPAD_GTP.I" output="IPAD_GTP.I">
            <pack_pattern in_port="inpad.inpad" input="inpad.inpad" name="inpad_to_IPAD_GTP" out_port="IPAD_GTP.I" output="IPAD_GTP.I"/>
          </direct>
        </interconnect>
        <metadata>
          <meta name="type">bel</meta>
          <meta name="subtype">input</meta>
        </metadata>
      </pb_type>  
```


The failre in CI that I can see is the following ([from here](https://source.cloud.google.com/results/invocations/03953482-0c3e-489f-ac12-87c1d6ba4875/log)):

```
Error 1: arch.timing.xml:-1 <pb_type> 'IPAD_GTP' timing-annotation/<model> mismatch on port 'I' of model 'IPAD_GTP_VPR', input port 'I' has combinational connections to port 'O'; specified in model, but no combinational delays found on pb_type
```

Not sure why the error says there is a combinational connection between I and O, as the model does not specify that. Moreover, I still wasn't able to reproduce the issue locally.

vaughnbetz:
I think this might lead to memory corruption so good to fix ASAP so we can rule it out as a cause of the symbiflow CI failure.

sfkhalid:
This code fixes the segmentation fault issues that occurred when the pack stage was skipped.
I ran Alessandro's test case from #1714 on this branch and no segmentation faults occurred.
I also ran vpr with a few other blif circuits while skipping the pack stage, and no issues occurred during those runs either.

@vaughnbetz Does this code look good to merge?

vpr/src/place/place_constraints.cpp
```diff
@@ -141,3 +141,18 @@ bool cluster_floorplanning_check(ClusterBlockId blk_id, t_pl_loc loc) {
 
     return floorplanning_good;
 }
+
+void check_cluster_constraints_init() {
```
vaughnbetz, line 145: Why are we assigning an empty partition (no constraint) to all clustered blocks?  Shouldn't we instead be reading the vpr_constraints file, and applying those constraints (or if there is no constraint, setting the empty constraint)?  We'll want the flow where you have a constraint file and do:

pack in one run
place in another run

to work, but that is only going to work if the second run (which skips packing) reads the constraint file.

vpr/src/base/vpr_api.cpp
```diff
@@ -502,6 +503,8 @@ bool vpr_pack_flow(t_vpr_setup& vpr_setup, const t_arch& arch) {
 
     if (packer_opts.doPacking == STAGE_SKIP) {
```
vaughnbetz, line 504: what does STAGE_SKIP mean?  If it means we don't do a packing or load a packing (just stop the flow) then we probably don't need to run check_cluster_constraints_init() as we didn't actually load or create a packing, so we can't proceed anyway.

vaughnbetz (commented):
This doesn't look right to me; seems like it is not really loading in the constraints and applying them to clusters when you load in a packing.

vpr/src/place/place_constraints.cpp
sfkhalid: Ah ok, so load_vpr_constraints should also be called here, and then we call a routine that fills in cluster_constraints accordingly.

sfkhalid: But what if some atoms are packed together in cluster that have incompatible region constraints according to the constraints XML file?

sfkhalid:
@vaughnbetz I changed the flow of the code based on the comments above. Now, the check_cluster_constraints_init() routine checks whether a constraints file has been specified. If no file has been specified, each cluster is just assigned an empty PartitionRegion. If a file is specified, each cluster is assigned a PartitionRegion based on the constrained atoms it contains (similar to what takes place during clustering).

vpr/src/place/place_constraints.cpp
```diff
@@ -141,3 +141,68 @@ bool cluster_floorplanning_check(ClusterBlockId blk_id, t_pl_loc loc) {
 
     return floorplanning_good;
 }
+
+void check_cluster_constraints_init(std::string constraints_file) {
```
sfkhalid, line 145: rename to cluster_constraints_init

sfkhalid: just delete this and call load_cluster_constraints in the vpr api

vpr/src/place/place_constraints.cpp
```diff
@@ -141,3 +141,68 @@ bool cluster_floorplanning_check(ClusterBlockId blk_id, t_pl_loc loc) {
...
+        //If an atom returns a valid PartitionId it is part of a partition (i.e. has floorplan constraints)
+        if (partid != PartitionId::INVALID()) {
+            ClusterBlockId clb_index = atom_ctx.lookup.atom_clb(atom_blk_id);
+            cluster_constrained_atoms[clb_index].push_back(atom_blk_id);
+        }
+    }
+
+    floorplanning_ctx.cluster_constraints.resize(cluster_ctx.clb_nlist.blocks().size());
+
+    for (auto cluster_id : cluster_ctx.clb_nlist.blocks()) {
```
sfkhalid, line 185: refactor: for every cluster, you have an empty PR, go through the atoms of that cluster in a loop and intersect with the empty PR and update.

mkurc-ant:
Hi, Is there a chance that this PR lands anytime soon? It is important from the perspective of SymbiFlow.

sfkhalid:
> @sfkhalid @vaughnbetz Hi, Is there a chance that this PR lands anytime soon? It is important from the perspective of SymbiFlow.

Yes, I have to refactor some things related to this code. I'll do that asap today.

mkurc-ant:
Allright, thanks.

vaughnbetz:
Code looks OK.  I'll merge it once CI goes fully green.

Issue #1716, opened 2021-04-30
Python Lint CI Failures


sfkhalid:
The python lint tests are currently failing on the CI because of two scripts that have python lint errors.

#### Expected Behaviour
The python lint tests should pass.

#### Current Behaviour
The python lint tests fail because two scripts have lint errors - run_vtr_task.py and util.py

![image](https://user-images.githubusercontent.com/61606358/116712391-ced7a580-a9a1-11eb-8402-166937892ff9.png)

![image](https://user-images.githubusercontent.com/61606358/116712446-dac36780-a9a1-11eb-9b9a-0a73d7f41f0a.png)

PR #1719, opened 2021-05-06
requirements: pin pylint

acomodi:
This PR fixes https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1716


vaughnbetz:
Thanks Alessandro!

PR #1731, opened 2021-05-18
Add missing post routing cluster pin fixup to genfasm

### LINKED ISSUE #1730: genfasm does not do post routing pb pin fixup
For blocks with equivalent pins cluster routing gets updated after device routing to accommodate for changes in which pins are actually used. This is done when the analysis flow is run via the `sync_netlists_to_routing` function. However this is not done in genfasm which leads to incorrect FASM faeture emission.

#### Expected Behaviour
genfasm should do post routing pb pin fixup so that FASM features emitted reflect the actual updated cluster routing.

#### Current Behaviour
genfasm does not do post routing pb pin fixup and uses cluster routing taken directly from the packer which yields in incorrect FASM features being emitted

#### Possible Solution
Call `sync_netlists_to_routing` in genfasm

#### Steps to Reproduce
Use data from the included archive: [case.tar.gz](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6500145/case.tar.gz)

The simple architecture included there defines a `clb` block with equivalent input pins followed by a set of N-to-1 muxes that effectively implement a fully connected crossbar. Each mux contains very informative and fake FASM annotation so that its configuration can be easily inferred from a generated FASM file.

1. Run the `runme.sh` (change paths to `vpr` and `genfasm`). This runs the full VPR flow and genfasm
2. Inspect the `design.route` file and see which pins are used for signals that come into the `clb` block used
3. Inspect the `top.fasm` file and see how input muxes of the `clb` are configured
4. Pins reported in the route file should not match mux configuration in the FASM file (It may match if VPR chooses not to take advantage of the pin equivalence but this is very unlikely).

#### Context
Trying to have an architecture with blocks that have fully connected crossbars at their inputs and be able to generate correct FASM files.

#### Your Environment

* VTR revision used: `v8.0.0-3437-ge7d45e013`
* Operating System and version: Linux Ubuntu 18.04
* Compiler version: `GNU 7.4.0 on Linux-4.15.0-45-generic x86_64`

mkurc-ant:
#### Description
Added execution of post routing cluster pin fixup to genfasm via the `sync_netlists_to_routing()` function.

#### Related Issue

#1730

#### Motivation and Context

Correct FASM generation for annotated blocks that have ports with equivalent input pins.

Using the testcase attached to the issue #1730. Also this PR contains update to the `genfasm` test which verifies that FASM features from clb muxes correspond to those from rr graph edges.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/EXTERNAL/libargparse/src/argparse.cpp
+++ b/libs/EXTERNAL/libargparse/src/argparse.cpp
@@ -4,6 +4,7 @@
 #include <cassert>
 #include <string>
 #include <set>
+#include <limits>
 
 #include "argparse.hpp"
 #include "argparse_util.hpp"
--- a/utils/fasm/test/test_fasm.cpp
+++ b/utils/fasm/test/test_fasm.cpp
@@ -7,6 +7,7 @@
 #include "fasm_utils.h"
 #include "arch_util.h"
 #include "rr_graph_writer.h"
+#include "post_routing_pb_pin_fixup.h"
 #include <sstream>
 #include <fstream>
 #include <regex>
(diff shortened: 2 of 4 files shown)


libs/EXTERNAL/libargparse/src/argparse.cpp
```diff
@@ -4,6 +4,7 @@
 #include <cassert>
 #include <string>
 #include <set>
+#include <limits>
```
acomodi, line 7: This looks like a leftover

acomodi (approved):
Apart from a small comment, LGTM

libs/EXTERNAL/libargparse/src/argparse.cpp
mkurc-ant: But it's not. It's actually required as without it compilation using Clang on the recent GH ubuntu-latest VM fails.

It's the same issue that was hitting `symbiflow-yosys-plugins` but here we can solve it properly by adding the missing include to the source code (for plugins we couldn't as we just clone and build Yosys there).

acomodi: Aha, right, I didn't notice the commit's comment. (good that it was only a one line change to fix the Clang compilation issue)

vaughnbetz (commented):
Thanks, I noticed that was breaking clang -- just bugged Kevin about it but I see you've already fixed it.

libs/EXTERNAL/libargparse/src/argparse.cpp
vaughnbetz, line 7: Good fix.

utils/fasm/test/test_fasm.cpp
```diff
@@ -168,6 +169,54 @@ TEST_CASE("match_lut_init", "[fasm]") {
     CHECK(match_lut_init("16'b0000000011111111", "16'b0000111100001111"));
 }
 
+std::string get_pin_feature (size_t inode) {
```
vaughnbetz, line 172: Should have a comment saying what this function does.
Also, I prefer declaring even local routines at the top of the file (then we can turn on missing prototype warnings, if they aren't on already).  If it's a local helper routine, best to declare it static. So I suggest:
- Declare this static if it's only intended for this file.
- Declare it at the top of the file, and comment what it does there.
- Comment should say what inode is (an rr_node index, or you could rename it to irr_node) and what the string format is and why you want it.

vaughnbetz:
Requesting some commenting updates but merging now as those can be handled in a follow up.

vaughnbetz:
@mkurc-ant : I merged this but please see the request for follow up comments / function pre-declaration.

mkurc-ant:
@vaughnbetz Thanks. I'll create a second PR with the comments you requested.

PR #1741, opened 2021-05-27
Floorplan constraints tuning

sfkhalid:
#### Description
This branch will have changes to improve the usability and performance of user-defined floorplan constraints in vpr. 

Changes were made to floorplan-related routines in clustering like atom_cluster_floorplan_check to reduce the pack time, which was increasing when floorplan constraints were specified. Now the pack time is similar when running vpr with no constraints or large constraints.

Changes will also be made to the clustering process to improve performance with tight floorplan constraints. Attraction groups will be used to ensure that molecules with similar floorplan regions are packed together.

#### Related Issue
This is part of a series of PRs to add floorplan placement constraints to VPR. It is related to issue #932


vpr/src/place/place_constraints.cpp
```diff
@@ -114,6 +117,10 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
 }
 
 /*returns true if location is compatible with floorplanning constraints, false if not*/
+/*
```
sfkhalid, line 120: Improve comment

vpr/src/base/partition_region.cpp
```diff
@@ -26,20 +26,18 @@ bool PartitionRegion::is_loc_in_part_reg(t_pl_loc loc) {
...
 }
 
-PartitionRegion intersection(PartitionRegion& pr1, PartitionRegion& pr2) {
+PartitionRegion intersection(const PartitionRegion& cluster_pr, const PartitionRegion& new_pr) {
     /**for N regions in part_region and M in the calling object you can get anywhere from
      * 0 to M*N regions in the resulting vector. Only intersection regions with non-zero area rectangles and
      * equivalent subtiles are put in the resulting vector
      * Rectangles are not merged even if it would be possible
      */
     PartitionRegion pr;
```
sfkhalid, line 35: make another version of this function that changes cluster_pr instead of creating new pr and returning (that will be more efficient because it avoids extra copying)

vpr/src/base/partition_region.h
```diff
@@ -46,7 +46,7 @@ class PartitionRegion {
      *   @param pr1     One of the PartitionRegions to be intersected
      *   @param pr2     One of the PartitionRegions to be intersected
      */
-    friend PartitionRegion intersection(PartitionRegion& pr1, PartitionRegion& pr2);
+    friend PartitionRegion intersection(const PartitionRegion& cluster_pr, const PartitionRegion& new_pr);
```
sfkhalid, line 50: add declaration of void version of this function

sfkhalid: also change the name and the first parameter would not be const (could also have it be a bool that returns whether or not there is an intersect)

vpr/src/place/place_constraints.cpp
```diff
@@ -11,6 +11,9 @@
 #include "globals.h"
 #include "place_constraints.h"
 
+/// @brief sentinel value for indicating that a subtile has not been specified
+constexpr int NO_SUBTILE = -1;
```
sfkhalid, line 15: put this in a header file that both region.cpp and place_constraints.cpp both include. Logically in region.h

sfkhalid:
QoR data for this pull request: 
[large_constraints_updated_QoR.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6572144/large_constraints_updated_QoR.xlsx)

To summarize the above - it was found that using the vpr constraints option to specify large floorplan constraints for blocks was increasing pack time. The changes made on this pull request got rid of the pack time increase that was occurring when specifying large constraints. The spreadsheet above shows the results of running three circuits with no constraints specified vs. large constraints specified, and it is shown that there is no significant increase in pack time anymore.

sfkhalid:
@vaughnbetz Can this PR be merged now? I added the changes we discussed last week, as well as the new QoR results.

vpr/src/base/partition_region.cpp
```diff
@@ -48,6 +50,20 @@ PartitionRegion intersection(PartitionRegion& pr1, PartitionRegion& pr2) {
     return pr;
 }
```
vaughnbetz, line 52: Let's talk about this in our meeting this week.  I can merge this to get it going, but this update_cluster_part_reg has a few problems:
1. name is overly specific; this is an "in-place" intersection, rather than having anything to do with clustering specifically. We should at least change the name to reflect what the routine does, rather than who calls it right now.
2. As written it is probably not faster than the other intersection routine, since it still creates a new partition_region and assigns it to the cluster_pr.  So if the other routine is leading to a negligible cluster time increase we can delete this one.  I know I requested an in-place version for speed, but looking at this I don't think it is "in place" enough to be faster.

vaughnbetz (approved):
Will merge, but see comment for us to discuss on Thursday.

vpr/src/base/partition_region.cpp
```diff
@@ -48,6 +50,20 @@ PartitionRegion intersection(PartitionRegion& pr1, PartitionRegion& pr2) {
...
 
+void update_cluster_part_reg(PartitionRegion& cluster_pr, const PartitionRegion& new_pr) {
```
sfkhalid, line 53: just delete this one and keep the original

PR #1742, opened 2021-05-27
RRGraphView::node_type() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::node_type()` throughout VTR. Every time `device_ctx.rr_nodes[index].type()` was called has been replaced with `rr_graph.node_type(RRNodeId(index))`. In order to do this, I followed a pattern like that shown below in every case where `device_ctx.rr_nodes[index].type()` needed to be replaced.

### Example Replacement
```diff
static void check_source(int inode, ClusterNetId net_id) {
    auto& device_ctx = g_vpr_ctx.device();
+   const auto& rr_graph = device_ctx.rr_graph;
    auto& cluster_ctx = g_vpr_ctx.clustering();
    auto& place_ctx = g_vpr_ctx.placement();

-   t_rr_type rr_type = device_ctx.rr_graph.node_type(RRNodeId(inode));
+   t_rr_type rr_type = rr_graph.node_type(RRNodeId(inode));
    if (rr_type != SOURCE) {
        VPR_FATAL_ERROR(VPR_ERROR_ROUTE,
                        "in check_source: net %d begins with a node of type %d.\n", size_t(net_id), rr_type);
    }
```

### Motivation and Context

These changes are a continuation of the work began by @tangxifan in PR #1693. Although this PR only implements one function through VTR, future pull requests of this nature that implement the RRGraphView API will include multiple functions. The changes made in this PR are fairly repetitive, so once they are looked over and merged into VTR, the rest of the API functions should be quite similar to implement.

### How Has This Been Tested?

I have run the regression tests for `vtr_reg_strong` along with the QoR testing found at `    $ ../scripts/run_vtr_task.py regression_tests/vtr_reg_nightly/vtr_reg_qor_chain`. I don't know the range of uncertainty that is to be expected in these tests, but the results from these QoR tests can be found below. The file containing all results can be found [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6555913/rr_graph_view_node_type.xlsx). And the file containing results with the inlined function is [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6562514/node_type_inline.xlsx).

|  |VTR Before These Changes| WIth Changes in this PR| After inlining `node_type()`|
|--|--|--|--|
|vtr_flow_elapsed_time|	1|	1.01204436| 0.985755221|
|odin_synth_time|	1|	1.010760259| 1.001577787|
|abc_depth|	1|	1| 1 |
|abc_synth_time|	1|	0.965919937| 0.970080695 |
|num_clb|	1|	1| 1 |
|num_memories|	1|	1| 1 |
|num_mult|	1|	1| 1 |
|max_vpr_mem|	1|	0.998708931| 0.998329365 |
|num_pre_packed_blocks|	1|	1| 1 |
|num_post_packed_blocks|	1|	1| 1 |
|device_grid_tiles|	1|	1| 1 |
|pack_time|	1|	0.998771255| 0.987595496 |
|placed_wirelength_est|	1|	1| 1 |
|place_time|	1|	1.006280619| 0.98858469 |
|placed_CPD_est|	1|	1| 1 |
|min_chan_width|	1|	1| 1 |
|routed_wirelength|	1|	1| 1 |
|min_chan_width_route_time|	1|	1.01422609| 0.983911635 |
|crit_path_routed_wirelength|	1|	1| 1 |
|critical_path_delay|	1|	1| 1 |
|geomean_nonvirtual_intradomain_critical_path_delay|	1|	1| 1 |
|crit_path_route_time|	1|	1.013738605|  1.002929197  |

#### Types of changes

#### Checklist:


vpr/src/device/rr_graph_view.h
```diff
@@ -53,10 +53,14 @@ class RRGraphView {
      */
   public:
     /* Get the type of a routing resource node */
-    t_rr_type node_type(RRNodeId node) const;
+    inline t_rr_type node_type(RRNodeId node) const {
```
tangxifan, line 56: Should we comment why inline is used here? e.g. for runtime optimization etc. This can help other developers to understand and avoid them to remove inline property randomly.

vpr/src/route/check_rr_graph.cpp
```diff
@@ -583,7 +586,9 @@ static bool has_adjacent_channel(const t_rr_node& node, const DeviceGrid& grid)
     /* TODO: this function should be reworked later to adapt RRGraphView interface 
      *       once xlow(), ylow(), side() APIs are implemented
      */
-    VTR_ASSERT(node.type() == IPIN || node.type() == OPIN);
+    auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
```
tangxifan, line 590: We may combine the two lines to one line:
```
const auto& rr_graph = g_vpr_ctx.device().rr_graph;
```

The ``device()`` method will give you a read-only ``DeviceContext`` data structure. So we can use ``const auto&`` rather than a reference like ``auto&``.

vpr/src/route/route_timing.cpp
```diff
@@ -546,6 +546,24 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
...
+                    auto current_node = best_routing[net_id].head;
+                    std::cout << "-------\n";
+                    while (current_node != nullptr) {
+                        auto current_type = device_ctx.rr_nodes[current_node->index].type_string();
+                        auto xlow = device_ctx.rr_nodes[current_node->index].xlow();
+                        auto xhigh = device_ctx.rr_nodes[current_node->index].xhigh();
+                        auto ylow = device_ctx.rr_nodes[current_node->index].ylow();
+                        auto yhigh = device_ctx.rr_nodes[current_node->index].yhigh();
+                        if ((xlow == xhigh) && (ylow == yhigh)) {
+                            printf("Node: %d %s (%d,%d) \n", current_node->index, current_type, xlow, ylow);
```
tangxifan, line 559: Not sure why we need the echo here. A few comments:
- VTR use the ``VTR_LOG()`` for messages rather than ``printf``. Please check the [vtr_log.h](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/libvtrutil/src/vtr_log.h) for detailed usages.
- These codes will print a message for each net that is routed. This may cause our log file very long. As a result, the regression test may take longer to run (it will read the log file line by line and extract information)
- If these codes are needed, I suggest to leave sufficient comments about why these codes are needed and how we should use it. We may output these messages to a separated log file.

vpr/src/route/route_timing.cpp
```diff
@@ -786,7 +804,7 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
             VTR_LOG("Critical path: %g ns\n", 1e9 * best_routing_metrics.critical_path.delay());
         }
 
-        VTR_LOG("Successfully routed after %d routing iterations.\n", itry);
+        VTR_LOG("timing Successfully routed after %d routing iterations.\n", itry);
```
tangxifan, line 807: Not sure why the message is changed here.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -627,13 +628,15 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
 
     inline void set_node_loc_side(uxsd::enum_loc_side side, int& inode) final {
         auto node = (*rr_nodes_)[inode];
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 631: I suggest not to use any global variable in this file.
This data structures in this head file caches the routing resource graph. As you see, the ``rr_nodes_`` in the line above is the cached version.
You can add the RRGraphView to the data cache:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/c18b2f24cecd2bdfc57c1b0ad20828932cb07433/vpr/src/route/rr_graph_uxsdcxx_serializer.h#L250-L297
And then use it in the same as the line above.

When refactoring is finished, we remove the data cache for ``rr_nodes``, ``rr_node_indices``, etc.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -645,7 +648,9 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         }
     }
     inline uxsd::enum_loc_side get_node_loc_side(const t_rr_node& node) final {
-        if (node.type() == IPIN || node.type() == OPIN) {
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 651: Same as my previous comment.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -690,6 +695,8 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
      * </xs:complexType>
      */
     inline int init_node_segment(int& inode, int segment_id) final {
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 698: Same as my previous comment.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -801,23 +810,29 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
-                    inode, node.type());
+                    inode, rr_graph.node_type(node.id()));
             }
         } else {
             node.set_direction(from_uxsd_node_direction(direction));
         }
     }
     inline uxsd::enum_node_direction get_node_direction(const t_rr_node& node) final {
-        if (node.type() == CHANX || node.type() == CHANY) {
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 833: Same as my previous comment.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -801,23 +810,29 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
     }
     inline uxsd::enum_node_type get_node_type(const t_rr_node& node) final {
-        return to_uxsd_node_type(node.type());
+        auto& device_ctx = g_vpr_ctx.device();
+        const auto& rr_graph = device_ctx.rr_graph;
+        return to_uxsd_node_type(rr_graph.node_type(node.id()));
     }
 
     inline void set_node_direction(uxsd::enum_node_direction direction, int& inode) final {
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 819: Same as my previous comment.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -801,23 +810,29 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         return size_t(node.id());
     }
     inline uxsd::enum_node_type get_node_type(const t_rr_node& node) final {
-        return to_uxsd_node_type(node.type());
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 813: Same as my previous comment.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -752,13 +759,15 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         // was invoked, but on formats that lack size on read,
         // make_room_in_vector will use an allocation pattern that is
         // amoritized O(1).
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 762: Same as my previous comment.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1554,6 +1571,8 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
     /*Allocates and load the rr_node look up table. SINK and SOURCE, IPIN and OPIN
      *share the same look up table. CHANX and CHANY have individual look ups */
     void process_rr_node_indices() {
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 1574: Same as my previous comment.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -930,6 +945,8 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         /*initialize a vector that keeps track of the number of wire to ipin switches
          * There should be only one wire to ipin switch. In case there are more, make sure to
          * store the most frequent switch */
+        auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 948: Same as my previous comment.

tangxifan (changes requested):
@ethanroj23 In general, most modifications are clear. My comments are most on adding comments or avoid the use of global variables in some specific file.

vpr/src/device/rr_graph_view.h
ethanroj23: Yes, that makes sense. I'll add that.

vpr/src/route/route_timing.cpp
ethanroj23: Thanks for pointing this out. I added this snippet of code for my own understanding of VTR, but it is not needed. I will remove it.

vpr/src/route/route_timing.cpp
ethanroj23: This also was left over from when I began trying to understand VTR. It will be removed.

tangxifan:
@ethanroj23 Thanks for editing. Look good to me now. @hzeller Would you mind reviewing this PR? It is a follow-up PR on the RRGraph refactoring effort. Thanks!

ethanroj23:
@hzeller @vaughnbetz

Would you be able to take a look through this PR? Xifan has already reviewed it, and some changes have been made in response.

vaughnbetz:
QoR data looks good.  Another good regtest to check QoR on is the Titan set; see https://docs.verilogtorouting.org/en/latest/dev/developing/#evaluating-quality-of-result-qor-changes

tangxifan:
@vaughnbetz Do we have a way to download artifacts from the github actions? This can save us a lot of time when checking the QoR.

vaughnbetz (approved):
Looks good to me.  Thanks Ethan!

vaughnbetz:
@tangxifan : I haven't pulled down artifacts from github actions, so not sure.  @mithro might now or be able to point you to a github action expert; also adding @MohamedElgammal in case he knows.

mithro:
You need to use the [`actions/upload-artifact`](https://github.com/actions/upload-artifact) in the action somewhere.

utils/fasm/test/test_fasm.cpp
```diff
@@ -260,7 +261,7 @@ TEST_CASE("fasm_integration_test", "[fasm]") {
 
                 // Add additional features to edges that go to CLB.I[11:0] pins
                 // to correlate them with features of CLB input mux later.
-                auto sink_type = device_ctx.rr_nodes[sink_inode].type();
+                auto sink_type = rr_graph.node_type(RRNodeId(sink_inode));
```
hzeller, line 264: (In all these places where we now manually convert an integer of sorts into an RRNodeId, we should see that the place we get it from already returns an RRNodeId in the future; but I suppose that is somewhere on the TODO list)

vpr/src/power/power.cpp
```diff
@@ -825,7 +826,7 @@ static void power_usage_routing(t_power_usage* power_usage,
                     auto next_node = device_ctx.rr_nodes[node.edge_sink_node(edge_idx)];
                     t_rr_node_power* next_node_power = &rr_node_power[node.edge_sink_node(edge_idx)];
 
-                    switch (next_node.type()) {
+                    switch (rr_graph.node_type(RRNodeId(node.edge_sink_node(edge_idx)))) {
```
hzeller, line 829: the duplication of `node.edge_sink_node(edge_idx)` and above when we get the next_node is a bit concerning as it might get out of sync and it is harder to see that we're talking about the same node.
We should probably have a `const next_node_id = node.edge_sink_node(edge_idx);` and use that when getting next_node and in this switch.

(there might be other places in this change where this might be a good idea)

hzeller (commented):
question:  now we don't call `type()` anymore on nodes (e.g. next_node.type() is replaced by asking the rr_graph).
Does this mean we can make `type()` now private or remove entirely ?

utils/fasm/test/test_fasm.cpp
ethanroj23: Yes, I believe the plan is to do that in the future. With the potential overhead of changing many functions to accept RRNodeIds instead of ints, we are just casting the node_id for now.

vpr/src/power/power.cpp
ethanroj23: I have updated this instance. Let me know if you had something else in mind than what I have changed it to.

ethanroj23:
> question: now we don't call `type()` anymore on nodes (e.g. next_node.type() is replaced by asking the rr_graph).
> Does this mean we can make `type()` now private or remove entirely ?

@hzeller I have now commented out `type()` from the `rr_node.h` file, and the tests still run successfully (since that function is no longer used). Would you be able to review these changes again?

vpr/src/draw/draw.cpp
```diff
@@ -3069,7 +3078,7 @@ static void draw_pin_to_chan_edge(int pin_node, int chan_node, ezgl::renderer* g
     ezgl::rectangle chan_bbox = draw_get_rr_chan_bbox(chan_node);
 
     float x2 = 0, y2 = 0;
-    switch (chan_rr.type()) {
+    switch (rr_graph.node_type(RRNodeId(chan_node))) {
```
hzeller, line 3081: could we re-use `channel_type` from above ?

In general it is harder to follow the code if this is a return value of a call to an object method that we pass a type-cast input value vs. a simple variable with a name. And if it is the same name we use everywhere and that value is made `const` it is much easier for someone reading the code (as it has human-understandable names and it is clear that the constant  values don't change in-between).

vpr/src/route/route_timing.cpp
```diff
@@ -1661,10 +1662,12 @@ static bool check_hold(const t_router_opts& router_opts, float worst_neg_slack)
 
 static size_t calculate_wirelength_available() {
     auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
 
     size_t available_wirelength = 0;
+    // But really what's happening is that this for loop iterates over every node and determines the available wirelength
     for (size_t i = 0; i < device_ctx.rr_nodes.size(); ++i) {
-        if (device_ctx.rr_nodes[i].type() == CHANX || device_ctx.rr_nodes[i].type() == CHANY) {
+        if (rr_graph.node_type(RRNodeId(i)) == CHANX || rr_graph.node_type(RRNodeId(i)) == CHANY) {
```
hzeller, line 1670: Here is another example where putting things in a const variable and then compare `(channel_type == CHANX || channel_type == CHAN_Y)` would really improve readability.

Also the multiple array lookups below `device_ctx.rr_nodes[i]` will be much more readable (and possibly faster) if put in a const reference (though outside the scope of this change, but something to look out for in the future)

hzeller (commented):
LGTM, but now things that were relatively short before (foo.type()) became a lot longer and less human readable (I've picked on one of the instances). In particular where the same long-ish call is repeated more than once only 0-5 lines apart).

It is worthwhile to consider giving things some meaningful names by assigning them to const variables and then use these in the code-snippets. Maybe write a script that picks out the same kind node_type() calls, and counts if the exact same call happens multiple times in any given 10 LOC group, so these might be good candidates for cleaning that.

Having said that, since this is mostly a mechanical translation from old to new, this is fine as is. But definitely something to put on the TODO list for later as we risk that things reduce in readability.

vpr/src/draw/draw.cpp
```diff
@@ -3016,13 +3025,13 @@ static void draw_pin_to_chan_edge(int pin_node, int chan_node, ezgl::renderer* g
         pin_side = pin_candidate_sides[0];
     } else {
         VTR_ASSERT(1 < pin_candidate_sides.size());
-        if (CHANX == chan_rr.type() && pin_rr.ylow() <= chan_rr.ylow()) {
+        if (CHANX == rr_graph.node_type(RRNodeId(chan_node)) && pin_rr.ylow() <= chan_rr.ylow()) {
```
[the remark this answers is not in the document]
ethanroj23: I have added a `const t_rr_type channel_type` constant here.

vpr/src/draw/draw.cpp
ethanroj23: This has also been resolved.

vpr/src/route/route_timing.cpp
ethanroj23: Additionally, I have fixed this.

vpr/src/draw/draw.cpp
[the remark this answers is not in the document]
hzeller: ah, much easier to read.

hzeller (approved):
LGTM

tangxifan (approved):

ethanroj23:
@vaughnbetz This  PR has now been approved by all the reviewers. Let me know if anything else needs to be done before merging it in to master. Thanks!

vaughnbetz:
Sounds good.  I assume QoR is unchanged with the recent changes -- is that right?
If it is, I'll merge as soon as CI goes green.

ethanroj23:
> Sounds good. I assume QoR is unchanged with the recent changes -- is that right?
> If it is, I'll merge as soon as CI goes green.

Yes, QoR remains the same as before with very slight fluctuations. Also, by CI do you mean the ODIN-II Strong Tests, VtR Strong Tests, and VtR Strong Sanitized Tests? These tests appear to have stayed yellow for a while now. Perhaps someone needs to manually start them?

vaughnbetz:
OK if they're stuck I'll merge, since this has been well tested before already.

PR #1728, opened 2021-05-16
Reorganizing vtr_flow/tasks 

ArashAhmadian:
Reorganized vtr-tasks by removing/moving the outdated tasks not run in CI. 
#### Description
Deleted tasks: timing_small,timing_chain,checkin_reg,basic_flow,regression_bidir, regression_fpu_hard_block_arch,regression_fpu_soft_block_arch

Moved tasks:arithmetic_tasks,func_multiclock,power,multiclock_mcnc to vtr_reg_weekly 
timing to vtr_reg_nightly,regression_mcnc to vtr_reg_basic 

#### Motivation and Context
To make sure that all the tasks are updated by being included in the CI and deleting outdated ones.

PR #1747, opened 2021-05-28
Deploy RRGraphBuilder in place of rr_node_indices in routing resource graph builder functions

tangxifan:
This pull request is a follow-up PR on the #1693 

#### Description

- [X] Fully use ``RRGraphBuilder`` when building the lookup in routing resource graph. Eliminate the use of ``rr_node_indice``
- [X] Cut off the dead loop between two header files: ``rr_graph2.h`` and ``rr_graph_storage.h``. Previously, ``rr_graph2.h`` is included in the ``rr_graph_storage.h`` due to a shared data structure ``rr_edge_info_set``. However,  ``rr_graph_storage.h`` should be the primitive header files. Therefore, a new header file ``rr_edge.h`` is created to host the shared data structure ``rr_edge_info_set``.
- [X] Added an API ``resize_nodes()`` to ``RRSpatialLookup``  to pre-allocate memory efficiently
- [X] Added an API ``mirror_nodes()`` to `RRSpatialLookup`` to enable batch copy of indices.

#### Related Issue

This is part of the PR request plan on routing resource graph, being a follow-up PR on the #1693 

#### Motivation and Context

Motivated by @vaughnbetz 's suggestion, the top priority is to deploy the new data structure ``RRGraphBuilder`` in place of the legacy data structure ``rr_node_indices`` in routing resource graph builder functions.
This PR addresses the comment, where we fully remove the use of ``rr_node_indices`` in builder functions.
As a result, ``RRGraphBuilder`` data structure is the only one that builder functions are talking to.

Should pass all the existing tests, since this PR is about refactoring.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/device/rr_spatial_lookup.h
+++ b/vpr/src/device/rr_spatial_lookup.h
@@ -127,9 +127,8 @@ class RRSpatialLookup {
                       e_side side);
 
     /**
-     * Resize three dimensions of the lookup under a given type of node to be memory efficient
-     * This function is called to expand the matrix when x, y or side 
-     * when one or more of them beyond current capacity 
+     * Resize the given 3 dimensions (x, y, side) of the RRSpatialLookup data structure for the given type
+     * This function will keep any existing data
      *
      * Strongly recommend to use when the sizes of dimensions are deterministic
      *
--- a/vpr/src/device/rr_spatial_lookup.h
+++ b/vpr/src/device/rr_spatial_lookup.h
@@ -30,7 +30,8 @@ class RRSpatialLookup {
 
     /* -- Accessors -- */
   public:
-    /* Returns the index of the specified routing resource node.  
+    /**
+     * Returns the index of the specified routing resource node.  
      * - (x, y) are the grid location within the FPGA
      * - rr_type specifies the type of resource,
      * - ptc gives a unique number of resources of that type (e.g. CHANX) at that (x,y).
@@ -64,7 +65,8 @@ class RRSpatialLookup {
 
     /* -- Mutators -- */
   public:
-    /* Register a node in the fast look-up 
+    /**
+     * Register a node in the fast look-up 
      * - You must have a valid node id to register the node in the lookup
      * - (x, y) are the coordinate of the node to be indexable in the fast look-up
      * - type is the type of a node
@@ -90,20 +92,23 @@ class RRSpatialLookup {
                   int ptc,
                   e_side side);
 
-    /* Mirror the last dimension of a look-up, i.e., a list of nodes, from a source coordinate to 
+    /**
+     * Mirror the last dimension of a look-up, i.e., a list of nodes, from a source coordinate to 
      * a destination coordinate.
      * This function is mostly need by SOURCE and SINK nodes which are indexable in multiple locations.
      * Considering a bounding box (x, y)->(x + width, y + height) of a multi-height and multi-width grid, 
      * SOURCE and SINK nodes are indexable in any location inside the boundry.
      *
-     * Note: currently this function only accept SOURCE/SINK nodes. May unlock for other depending on needs
+     * Note: currently this function only accepts SOURCE/SINK nodes. May unlock for the other types 
+     * depending on needs
      */
     void mirror_nodes(const vtr::Point<int>& src_coord,
                       const vtr::Point<int>& des_coord,
                       t_rr_type type,
                       e_side side);
 
-    /* Resize three dimensions of the lookup under a given type of node to be memory efficient
+    /**
+     * Resize three dimensions of the lookup under a given type of node to be memory efficient
      * This function is called to expand the matrix when x, y or side 
      * when one or more of them beyond current capacity 
      *
--- a/vpr/src/device/rr_spatial_lookup.h
+++ b/vpr/src/device/rr_spatial_lookup.h
@@ -1,6 +1,7 @@
 #ifndef RR_SPATIAL_LOOKUP_H
 #define RR_SPATIAL_LOOKUP_H
 
+#include "vtr_geometry.h"
 #include "vpr_types.h"
 
 /******************************************************************** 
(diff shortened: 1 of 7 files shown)


vpr/src/device/rr_spatial_lookup.h
```diff
@@ -89,7 +90,32 @@ class RRSpatialLookup {
...
+     * Note: currently this function only accept SOURCE/SINK nodes. May unlock for other depending on needs
+     */
+    void mirror_nodes(const vtr::Point<int>& src_coord,
+                      const vtr::Point<int>& des_coord,
+                      t_rr_type type,
+                      e_side side);
+
+    /* Resize three dimensions of the lookup under a given type of node to be memory efficient
+     * This function is called to expand the matrix when x, y or side 
+     * when one or more of them beyond current capacity
```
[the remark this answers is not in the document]
vaughnbetz: Should also say exactly what this does.  resizes the given 3 dimensions of the RRNodeLookup data structure for the given type.  Any existing data in the lookup for this type is copied over (? I think?)

vpr/src/route/rr_graph2.cpp
```diff
@@ -928,44 +927,40 @@ static void load_chan_rr_indices(const int max_chan_width,
...
     for (int chan = 0; chan < num_chans - 1; ++chan) {
         for (int seg = 1; seg < chan_len - 1; ++seg) {
             /* Assign an inode to the starts of tracks */
             int x = (type == CHANX ? seg : chan);
             int y = (type == CHANX ? chan : seg);
             const t_chan_seg_details* seg_details = chan_details[x][y].data();
 
-            for (unsigned track = 0; track < indices[type][chan][seg][0].size(); ++track) {
+            for (int track = 0; track < max_chan_width; ++track) {
                 if (seg_details[track].length() <= 0)
```
vaughnbetz, line 940: We likely should let the .length = 0 case go through eventually, to model muxes.  Could comment that now.

vpr/src/route/rr_graph2.cpp
```diff
@@ -1132,8 +1098,14 @@ static void load_block_rr_indices(RRGraphBuilder& rr_graph_builder,
                 int root_x = x - width_offset;
```
vaughnbetz, line 1098: Could the whole double-for loop comtaining these calls be moved into RRGraphBuilder as a high-level API?  Call it something like dupLargeBlkLookups().  Then mirror_nodes could be removed or made a (private if possible) helper function.  

Right now the comment for mirror_node describes a good chunk of the body of this for loop, despite the for loop being in the caller.  I think that's a sign we really want a higher-level API.

vaughnbetz (changes requested):

vaughnbetz:
Thanks Xifan.  Looks mostly good, but I have some comments for you to consider.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -89,7 +90,32 @@ class RRSpatialLookup {
...
                   e_side side);
 
-    /* TODO: Add an API remove_node() to unregister a node from the look-up */
+    /* Mirror the last dimension of a look-up, i.e., a list of nodes, from a source coordinate to 
+     * a destination coordinate.
+     * This function is mostly need by SOURCE and SINK nodes which are indexable in multiple locations.
+     * Considering a bounding box (x, y)->(x + width, y + height) of a multi-height and multi-width grid, 
+     * SOURCE and SINK nodes are indexable in any location inside the boundry.
+     *
+     * Note: currently this function only accept SOURCE/SINK nodes. May unlock for other depending on needs
```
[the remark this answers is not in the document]
tangxifan: Thanks for correction. I have adapted the comments about the APIs by following the style in ``vpr_context.h``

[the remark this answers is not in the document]
tangxifan: Yes. This function will copy a vector of node indices from a source ``[src_x][src_y][type][side]`` to a destination ``[des_x][des_y][type][side]``. It is required when building the lookup for SOURCE and SINK nodes, which are supposed to be indexable at any location inside a large block ``(x,y)->(x+width, y+height)``.

I agree that we should have a high-level API to do the job, while this low-level API can be a private one. Here are my thoughts:
- We can create a high-level API like ``expand_nodes()`` which will copy the vector from a source coordinate to all the coordinates in a bounding box.
- I also think that it is better to rework the ``find_node()`` API so that we always search the lowest (x,y) corner when dealing with large blocks. But this may require the data structure to be dependent on ``DeviceGrid`` information (it needs to identify if  a grid has height > 1 as well as width >1)

We can keep brainstorming on this front. I have left comments in the header file as a TODO item.

I have also added comments (a quick example) about how to use the API.

vpr/src/device/rr_spatial_lookup.h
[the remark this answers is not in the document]
tangxifan: Agree. I have reworked the comment to be precise and also comment that any existing data will be maintained without modification.

vpr/src/route/rr_graph2.cpp
tangxifan: No problem. Added this as a TODO item. I will check details about the ``t_chan_details`` data structure and see what we can do.

vpr/src/route/rr_graph2.cpp
tangxifan: Yes. This is what I thought:

- We can create a high-level API like expand_nodes() which will copy the vector from a source coordinate to all the coordinates in a bounding box.

I will keep thinking about what high-level API we need.  I am positive for moving the for-loops into API.

tangxifan:
@vaughnbetz Thanks for the comments. I have addressed all of them and this PR is ready for your review.

Also, we can discuss if more should be added to this PR or leave them in next PRs:
- Use ``find_node()`` API in place of the legacy function ``get_rr_node_index()``. Then we can fully shadow ``rr_node_indices`` and remove it from the ``DeviceContext``
- Get rid of the legacy convention on swapping (x, y) for CHANX nodes when searching its index.
- Develop high-level APIs in place of the ``mirror_nodes()``.

vpr/src/route/rr_graph2.cpp
vaughnbetz: Sounds good.  Let's land this and keep iterating on the higher-level api.

vaughnbetz (approved):
Looks good, thanks! Let's leave further improvements for the follow-on PRs you mention.

vaughnbetz:
Will merge when CI goes green.

tangxifan:
@vaughnbetz  Just the nightly test failed with an error:

```
13:10:43 up  4:06,  0 users,  load average: 3.04, 3.02, 3.00
++ free -h
              total        used        free      shared  buff/cache   available
Mem:           102G        5.5G         60G        8.6M         36G         96G
Swap:          1.0G          0B        1.0G
++ sleep 300
bash: : command not found


[ID: 7125009] Build finished after 14654 secs, exit value: 127


Warning: Permanently added 'localhost' (ED25519) to the list of known hosts.
[13:11:09] Collecting build artifacts from build VM
Build script failed with exit code: 127
```


But this happened after merging the latest master to my branch. I see similar failures in other PRs:  https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1728
Not sure why the error happens.

vaughnbetz:
This looks like another spurious kokoro failure; exit code 127 is apparently a command not found one.  I'll merge this and hopefully kokoro recovers soon.

tangxifan:
Thanks @vaughnbetz. I will keep an eye on the regression test and also start working on the next PR.

PR #1749, opened 2021-05-29
aligned_allocator: add operator==

Xiretza:
#### Description
Fixes #1745, see the commit message or the issue for details.

#### How Has This Been Tested?
The test suite still passes, and compilation is now successful with GCC 11.

#### Types of changes
- [x] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1750, opened 2021-05-29
CI: Test compatibility with recent compilers

Xiretza:
#### Description

The CI compatibility tests currently only go up to GCC 9 and clang 8, which are both quite old. While clang 11 and 12 aren't in Ubuntu 18.04, clang 10 and GCC 11 are possible.

#### Related Issue
This will only work once #1749 is merged.

#### Motivation and Context
To test vtr with modern compilers.

Since this only adds new tests, there's not much to test.

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -236,8 +236,9 @@ jobs:
         - { name: 'GCC 9 (Ubuntu Focal - 20.04)',     eval: 'CC=gcc-9 && CXX=g++-9',           build: ''               }
         - { name: 'GCC 10 (Ubuntu Hirsute - 21.04)',  eval: 'CC=gcc-10 && CXX=g++-10',         build: ''               }
         - { name: 'GCC 11 (Latest)',                  eval: 'CC=gcc-11 && CXX=g++-11',         build: ''               }
-        - { name: 'Clang 6 (Debian + Ubuntu common)', eval: 'CC=clang-6.0 && CXX=clang++-6.0', build: ''               }
-        - { name: 'Clang 8 (Latest Release)',         eval: 'CC=clang-8 && CXX=clang++-8',     build: ''               }
+        - { name: 'Clang 6 (Ubuntu Bionic - 18.04)',  eval: 'CC=clang-6.0 && CXX=clang++-6.0', build: ''               }
+        - { name: 'Clang 7 (Debian Buster)',          eval: 'CC=clang-7 && CXX=clang++-7',     build: ''               }
+        - { name: 'Clang 10 (Ubuntu Focal - 20.04)',  eval: 'CC=clang-10 && CXX=clang++-10',   build: ''               }
     name: 'B: ${{ matrix.name }}'
     steps:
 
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -232,10 +232,14 @@ jobs:
         - { name: 'GCC 5 (Ubuntu Xenial - 16.04)',    eval: 'CC=gcc-5 && CXX=g++-5',           build: 'release_strict' }
         - { name: 'GCC 6 (Debian Stretch)',           eval: 'CC=gcc-6 && CXX=g++-6',           build: 'release_strict' }
         - { name: 'GCC 7 (Ubuntu Bionic - 18.04)',    eval: 'CC=gcc-7 && CXX=g++-7',           build: ''               }
-        - { name: 'GCC 9 (Latest Release)',           eval: 'CC=gcc-9 && CXX=g++-9',           build: ''               }
         - { name: 'GCC 8 (Ubuntu Latest)',            eval: 'CC=gcc-8 && CXX=g++-8',           build: ''               }
+        - { name: 'GCC 9',                            eval: 'CC=gcc-9 && CXX=g++-9',           build: ''               }
```
mithro, line 236: Could you add comments about which distro have these compilers as default (see other examples)?

mithro (approved):
Generally looks good to me (LGTM). One minor comment.

mithro:
Also looks like some cleanup is needed to make the latest compilers work?

Xiretza:
> Also looks like some cleanup is needed to make the latest compilers work?

As mentioned in the pull request, #1749 should take care of that.

.github/workflows/test.yml
Xiretza: I looked up the defaults for all supported ubuntu and debian releases, turns out GCC 5 and clang 8/9 aren't actually used anywhere.

Xiretza:
Removing gcc-5 breaks a bunch of workflows that use it - does it even make sense to use an ancient compiler to check e.g. compilation warnings? A newer compiler would definitely have better diagnostics.

mithro:
According to https://wiki.ubuntu.com/Releases -- Ubuntu - 16.04.7 LTS - Xenial Xerus - End of Standard Support: April 2021. Thus I think we can drop 16.04 support.

IIRC our policy has been to support any Ubuntu version that is currently supported. That includes;
 - Ubuntu 18.04.5 LTS, Bionic Beaver
 - Ubuntu 20.04.2 LTS, Focal Fossa
 - Ubuntu 20.10, Groovy Gorilla
 - Ubuntu 21.04, Hirsute Hippo

mithro:
Looking at https://wiki.debian.org/DebianReleases, we could probably drop support for Debian Stretch tool
 - Debian 9 “Stretch”: July 6, 2020 to June 30, 2022
 - Debian 10 “Buster”: July, 2022 to June, 2024

mithro:
Looks like Centos-8 is the earliest version of CentOS / RHEL that makes sense too.

mithro:
https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux#Product_life_cycle

PR #1751, opened 2021-06-01
Add constraints propagation to update the PartitionRegions of blocks …

sfkhalid:
…in macros at the start of initial placement

This code does constraint propagation at the start of initial placement. It goes through all the macros, and checks for floorplanning constraints on each one. For the macros that have floorplan constraints, it calculates the smallest possible floorplan region for the entire macro, based on the floorplan constraints of each block within the macro. This is all done before initial placement starts so that during initial placement, when macros are being placed, the floorplan legality can easily be checked by seeing whether the head member of the macro is in the floorplan region that was calculated earlier.

#### Related Issue

#### Motivation and Context
This change makes it easier to check floorplan legality for macros during initial placement.

### DIFF
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -82,20 +82,17 @@ static int check_macro_can_be_placed(t_pl_macro pl_macro, int itype, t_pl_loc he
     // Every macro can be placed until proven otherwise
     int macro_can_be_placed = true;
 
+    //Check whether head macro is in appropriate floorplan region
+    //return true if macro_can_be_placed = yes, false if no
     bool macro_constrained = is_macro_constrained(pl_macro);
-    PartitionRegion macro_pr;
-
-    if (macro_constrained) {
-        macro_pr = constrained_macro_locs(pl_macro);
-    }
 
     // Check whether all the members can be placed
     for (size_t imember = 0; imember < pl_macro.members.size(); imember++) {
         t_pl_loc member_pos = head_pos + pl_macro.members[imember].offset;
 
-        //if the macro is constrained, check if the member position is within the PartitionRegion for the macro
-        if (macro_constrained) {
-            bool member_loc_good = macro_pr.is_loc_in_part_reg(member_pos);
+        //if the macro is constrained, check if the head position is within the PartitionRegion for the macro
+        if (macro_constrained && imember == 0) {
+            bool member_loc_good = cluster_floorplanning_legal(pl_macro.members[imember].blk_index, member_pos);
             if (!member_loc_good) {
                 macro_can_be_placed = false;
                 break;
@@ -416,6 +413,9 @@ void initial_placement(enum e_pad_loc_type pad_loc_type, const char* constraints
     //Sort blocks
     vtr::vector<ClusterBlockId, t_block_score> block_scores = assign_block_scores();
     std::vector<ClusterBlockId> sorted_blocks = sort_blocks(block_scores);
+
+    //Perform constraints_propagation
+    constraints_propagation();
 
     // Loading legal placement locations
     zero_initialize_grid_blocks();
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -419,8 +419,8 @@ void initial_placement(enum e_pad_loc_type pad_loc_type, const char* constraints
     vtr::vector<ClusterBlockId, t_block_score> block_scores = assign_block_scores();
     std::vector<ClusterBlockId> sorted_blocks = sort_blocks(block_scores);
 
-    //Perform constraints_propagation
-    constraints_propagation();
+    //Go through blocks with floorplan constraints to ensure we have the tightest constraint on each one.
+    propagate_place_constraints();
 
     // Loading legal placement locations
     zero_initialize_grid_blocks();
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -419,7 +419,9 @@ void initial_placement(enum e_pad_loc_type pad_loc_type, const char* constraints
     vtr::vector<ClusterBlockId, t_block_score> block_scores = assign_block_scores();
     std::vector<ClusterBlockId> sorted_blocks = sort_blocks(block_scores);
 
-    //Go through blocks with floorplan constraints to ensure we have the tightest constraint on each one.
+    /* Go through cluster blocks to calculate the tightest placement
+     * floorplan constraint for each constrained block
+     */
     propagate_place_constraints();
 
     // Loading legal placement locations
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -82,15 +82,20 @@ static int check_macro_can_be_placed(t_pl_macro pl_macro, int itype, t_pl_loc he
     // Every macro can be placed until proven otherwise
     int macro_can_be_placed = true;
 
-    //Check whether head macro is in appropriate floorplan region
-    //return true if macro_can_be_placed = yes, false if no
+    //Check whether macro contains blocks with floorplan constraints
     bool macro_constrained = is_macro_constrained(pl_macro);
 
     // Check whether all the members can be placed
     for (size_t imember = 0; imember < pl_macro.members.size(); imember++) {
         t_pl_loc member_pos = head_pos + pl_macro.members[imember].offset;
 
-        //if the macro is constrained, check if the head position is within the PartitionRegion for the macro
+        /*
+         * If the macro is constrained, check that the head member is in a legal position from
+         * a floorplanning perspective. It is enough to do this check for the head member alone,
+         * because constraints propagation was performed to calculate smallest floorplan region for the head
+         * macro, based on the constraints on all of the blocks in the macro. So, if the head macro is in a
+         * legal floorplan location, all other blocks in the macro will be as well.
+         */
         if (macro_constrained && imember == 0) {
             bool member_loc_good = cluster_floorplanning_legal(pl_macro.members[imember].blk_index, member_pos);
             if (!member_loc_good) {
--- a/vpr/src/place/place_constraints.cpp
+++ b/vpr/src/place/place_constraints.cpp
@@ -140,11 +140,13 @@ PartitionRegion update_macro_member_pr(PartitionRegion& head_pr, const t_pl_offs
 }
 
 void print_macro_constraint_error(const t_pl_macro& pl_macro) {
+    auto& cluster_ctx = g_vpr_ctx.clustering();
     VTR_LOG(
         "Feasible floorplanning constraints could not be calculated for the placement macro. \n"
         "The placement macro contains the following blocks: \n");
     for (unsigned int i = 0; i < pl_macro.members.size(); i++) {
-        VTR_LOG("Block #%zu ", size_t(pl_macro.members[i].blk_index));
+        std::string blk_name = cluster_ctx.clb_nlist.block_name((pl_macro.members[i].blk_index));
+        VTR_LOG("Block %s (#%zu) ", blk_name, size_t(pl_macro.members[i].blk_index));
     }
     VTR_LOG("\n");
     VPR_ERROR(VPR_ERROR_PLACE, " \n Check that the above-mentioned placement macro blocks have compatible floorplan constraints.\n");
--- a/vpr/src/place/place_constraints.cpp
+++ b/vpr/src/place/place_constraints.cpp
@@ -146,7 +146,7 @@ void print_macro_constraint_error(const t_pl_macro& pl_macro) {
         "The placement macro contains the following blocks: \n");
     for (unsigned int i = 0; i < pl_macro.members.size(); i++) {
         std::string blk_name = cluster_ctx.clb_nlist.block_name((pl_macro.members[i].blk_index));
-        VTR_LOG("Block %s (#%zu) ", blk_name, size_t(pl_macro.members[i].blk_index));
+        VTR_LOG("Block %s (#%zu) ", blk_name.c_str(), size_t(pl_macro.members[i].blk_index));
     }
     VTR_LOG("\n");
     VPR_ERROR(VPR_ERROR_PLACE, " \n Check that the above-mentioned placement macro blocks have compatible floorplan constraints.\n");
--- a/vpr/src/place/place_constraints.cpp
+++ b/vpr/src/place/place_constraints.cpp
@@ -80,18 +80,11 @@ PartitionRegion update_macro_head_pr(const t_pl_macro& pl_macro, const Partition
 
                 vtr::Rect<int> reg_rect = block_regions[i].get_region_rect();
 
-                t_pl_loc min_pl_loc(reg_rect.xmin(), reg_rect.ymin(), block_regions[i].get_sub_tile());
+                modified_reg.set_region_rect(reg_rect.xmin() - offset.x, reg_rect.ymin() - offset.y, reg_rect.xmax() - offset.x, reg_rect.ymax() - offset.y);
 
-                t_pl_loc modified_min_pl_loc = min_pl_loc - offset;
-
-                t_pl_loc max_pl_loc(reg_rect.xmax(), reg_rect.ymax(), block_regions[i].get_sub_tile());
-
-                t_pl_loc modified_max_pl_loc = max_pl_loc - offset;
-
-                modified_reg.set_region_rect(modified_min_pl_loc.x, modified_min_pl_loc.y, modified_max_pl_loc.x, modified_max_pl_loc.y);
                 //check that subtile is not an invalid value before changing, otherwise it just stays -1
                 if (block_regions[i].get_sub_tile() != NO_SUBTILE) {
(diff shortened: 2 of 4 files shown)


vpr/src/place/initial_placement.cpp
```diff
@@ -417,6 +419,9 @@ void initial_placement(enum e_pad_loc_type pad_loc_type, const char* constraints
     vtr::vector<ClusterBlockId, t_block_score> block_scores = assign_block_scores();
     std::vector<ClusterBlockId> sorted_blocks = sort_blocks(block_scores);
 
+    //Perform constraints_propagation
```
vaughnbetz, line 422: Not a great comment as it is pretty close to just the function name.  Explain what constraints_propagation is briefly (go through floorplan and .locked constraints to ensure we have the tightest constraint on each block when there are interacting constraints).

vpr/src/place/place_constraints.h
```diff
@@ -34,9 +34,19 @@ bool cluster_floorplanning_legal(ClusterBlockId blk_id, const t_pl_loc& loc);
 bool is_macro_constrained(const t_pl_macro& pl_macro);
 
 /*
- * Returns region of valid locations for the head of the macro based on floorplan constraints
+ * Returns PartitionRegion for the head of the macro based on the floorplan constraints
+ * of all blocks in the macro.
  */
```
vaughnbetz, line 39: Shouldn't just do the head -- should figure out the tightest constraint on each member of the macro to give the most info possible to later code.

vpr/src/place/initial_placement.cpp
vaughnbetz: propagate_placement_cons or propagate_place_constraints() are better names I think; they read as an action.

vpr/src/place/place_constraints.cpp
```diff
@@ -113,6 +113,25 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
+        if (is_macro_constrained(pl_macro)) {
+            /*
+             * Update the PartitionRegion for the head of the macro
+             * based on the constraints of all blocks contained in the macro
+             */
+            PartitionRegion macro_head_pr = update_macro_pr(pl_macro);
+
+            //Get block id of head macro member and update its PartitionRegion
+            ClusterBlockId blk_id = pl_macro.members[0].blk_index;
+            floorplanning_ctx.cluster_constraints[blk_id] = macro_head_pr;
```
vaughnbetz, line 130: Should set the tightest constraint possible on all members of the macro, not just the head.  It gives a stronger invariant to later code (checking any block in a macro is good enough).

vaughnbetz: Should also 
1. set the .locked flag for any blocks that have a floorplan constraint that implies a single location
2. set a floorplan_ctx.cluster_constraints for any block that has its .locked flag set (read in from a cluster-level constraints file / placement file).  That again helps later code -- the cluster_constraints data will encapsulate everything we know about cluster location constraints.
That should probably be done in a separate helper function, which can be called by the top-level propagate_place_constraints() (proposed name) function.

vaughnbetz (changes requested):

vpr/src/place/place_constraints.cpp
sfkhalid: That makes sense, I will implement the constraints propagation so that we have the tightest constraints for all members of the macro.

I put the code for setting locked flags on a different pull request - #1759 - because there are other changes I am implementing in that one and it felt like it was getting too big for one pull request.

sfkhalid:
To check that the code in this pull request did not increase the place time, I ran vtr_reg_qor_chain on this branch and master and compared the results. The comparison can be seen here:
[constraints_prop_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6604666/constraints_prop_qor_comparison.xlsx)
The results show that the overall run time and place time did not increase. 
Furthermore, the place time decreased a bit for bigger circuits such as mcml, LU32, and stereovision2.

vpr/src/place/place_constraints.cpp
```diff
@@ -54,8 +54,8 @@ bool is_macro_constrained(const t_pl_macro& pl_macro) {
 }
 
 /*Returns PartitionRegion of where the head of the macro could go*/
-PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
-    PartitionRegion macro_pr;
+PartitionRegion update_macro_head_pr(const t_pl_macro& pl_macro) {
```
vaughnbetz, line 57: Should update the partition regions on all macro members, not just the head.  Easy to do: after you've found the implied (tightest) constraint on the macro head, just shift it the appropriate amount for each other member of the macro (i.e. if the next member is at delta_y = +1, then shift the y-values by 1),. Also should intersect the resulting constraint per member with the grid dimensions, unless you can prove that the code that figures out the constraint on the macro_head guarantees that later shifting it can never go outside the grid.

vpr/src/place/place_constraints.cpp
```diff
@@ -82,37 +82,95 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
 
                 modified_pr.add_to_part_region(modified_reg);
             }
 
             if (num_constrained_members == 1) {
-                macro_pr = modified_pr;
+                macro_head_pr = modified_pr;
             } else {
-                macro_pr = intersection(macro_pr, modified_pr);
+                macro_head_pr = intersection(macro_head_pr, modified_pr);
```
vaughnbetz, line 103: Should also intersect the modified_pr with the grid dimensions to make sure it never goes outside the grid, and you get the tightest possible constraints.

vaughnbetz: And even if some member of a macro has no constraint, you should consider it to have a constraint of the size of the grid, as it still implies a constraint for other members (if I'm the end of a vertical carry chain of length 2, it means the head of the chain can never be placed in the top row, even if I have no floorplan constraint on the second block in the chain).
If no member of a macro at all has a constraint, then you probably still want to avoid setting any region constraint as it might slow down the placer in that case.

vpr/src/place/place_constraints.cpp
```diff
@@ -82,37 +82,95 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
     //if the intersection is empty, no way to place macro members together, give an error
-    if (macro_pr.empty()) {
+    if (macro_head_pr.empty()) {
         VPR_ERROR(VPR_ERROR_PLACE, " \n Feasible floorplanning constraints could not be calculated for the placement macro.\n");
     }
 
+    return macro_head_pr;
+}
+
+PartitionRegion update_macro_member_pr(PartitionRegion& head_pr, const t_pl_offset& offset) {
```
vaughnbetz, line 116: Good for a function to comment to note that the tightest possible constraint (implied by the whole chain) must already have been set on the head of the macro that you pass in.

vpr/src/place/place_constraints.h
```diff
@@ -34,9 +34,29 @@ bool cluster_floorplanning_legal(ClusterBlockId blk_id, const t_pl_loc& loc);
...
+ * based on the constraint that was calculated for the head region.
+ * The constraint that was calculated for the head region considered all
+ * constrained blocks in the macro, so to calculate the constraint for the rest
+ * of the blocks, it is just the head constraint with the offset of the member
+ * applied.
+ */
+PartitionRegion update_macro_member_pr(PartitionRegion& head_pr, const t_pl_offset& offset);
+
+/*
+ * Updates the floorplan constraints information for all constrained macros.
```
vaughnbetz, line 53: Giving an example would help explain what this does (e.g. macro of length 2, both with floorplan constraints, implies a tighter floorplan constraint on the head block of the macro that we can compute by shifting and intersecting the per-member constraints).

vpr/test/test_vpr_constraints.cpp
```diff
@@ -425,6 +426,28 @@ TEST_CASE("RegionLocked", "[vpr]") {
     REQUIRE(is_r2_locked == false);
 }
```
vaughnbetz, line 428: Nice ... glad to see a unit test :).

vpr/src/place/place_constraints.cpp
sfkhalid: Updating the PartitionRegions of the other members is done with the helper function update macro_member_pr. Yes, I can update that function to intersect with the grid dimensions. I'll also add a check to see whether the updated PartRegion is empty, just like I have in the head macro function, because that would indicate that the floorplan constraints are incompatible as well.

vpr/src/place/place_constraints.cpp
sfkhalid: Yes, I'll update the function to have an intersection with the grid dimensions

sfkhalid: If a member of a macro has no constraint (when other members do), then it is updated to have a PartitionRegion that is the head macro, shifted by its offset. Basically, it is treated the same as the other macro members. This is because wherever the other members are constrained to, the non-constrained member will essentially have to be placed there too. 

If the macro has no members with floorplan constraints, then none of the updating PartitionRegion functions are called (update_macro_head_pr, update_macro_member_pr), so the PartitionRegions of all the members just remain empty.

vaughnbetz: OK, that sounds good.

vpr/src/place/place_constraints.cpp
```diff
@@ -82,29 +82,69 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
+PartitionRegion update_macro_member_pr(PartitionRegion& head_pr, const t_pl_offset& offset, const PartitionRegion& grid_pr) {
+    std::vector<Region> block_regions = head_pr.get_partition_region();
+    PartitionRegion macro_pr;
+
+    for (unsigned int i = 0; i < block_regions.size(); i++) {
+        Region modified_reg;
+
+        vtr::Rect<int> reg_rect = block_regions[i].get_region_rect();
+
+        t_pl_loc min_pl_loc(reg_rect.xmin(), reg_rect.ymin(), block_regions[i].get_sub_tile());
```
sfkhalid, line 128: See if you can refactor this by creating an offset helper function for vtr::Rect

vpr/src/place/place_constraints.cpp
```diff
@@ -82,29 +82,69 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
 
                 t_pl_loc min_pl_loc(reg_rect.xmin(), reg_rect.ymin(), block_regions[i].get_sub_tile());
 
-                t_pl_loc modified_min_pl_loc = min_pl_loc + offset;
```
sfkhalid, line 85: See if you can refactor this by creating an offset helper function for vtr::Rect

vpr/src/place/place_constraints.cpp
```diff
@@ -82,29 +82,69 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
+                macro_head_pr = intersection(macro_head_pr, modified_pr);
             }
         }
     }
 
+    //intersect to ensure the head pr does not go outside of grid dimensions
+    macro_head_pr = intersection(macro_head_pr, grid_pr);
+
+    //if the intersection is empty, no way to place macro members together, give an error
+    if (macro_head_pr.empty()) {
```
sfkhalid, line 112: Add detail to error message - say what blocks are in the macro that failed. Macro with x blocks can't be placed, print names of the blocks, and say that floorplan region is empty for the macro. Say to check the floorplan constraints

vpr/src/place/place_constraints.cpp
```diff
@@ -82,29 +82,69 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
+        }
+
+        macro_pr.add_to_part_region(modified_reg);
+    }
+
+    //intersect to ensure the head pr does not go outside of grid dimensions
+    macro_pr = intersection(macro_pr, grid_pr);
+
     //if the intersection is empty, no way to place macro members together, give an error
     if (macro_pr.empty()) {
```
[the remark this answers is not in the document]
sfkhalid: can make a helper function to print this error message (or the macro) and call it in the two places you need it

vpr/src/place/place_constraints.cpp
```diff
@@ -82,29 +82,69 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
+        //check that subtile is not an invalid value before changing, otherwise it just stays -1
+        if (block_regions[i].get_sub_tile() != NO_SUBTILE) {
+            modified_reg.set_sub_tile(macro_min_pl_loc.sub_tile);
+        }
+
+        macro_pr.add_to_part_region(modified_reg);
+    }
+
+    //intersect to ensure the head pr does not go outside of grid dimensions
+    macro_pr = intersection(macro_pr, grid_pr);
```
sfkhalid, line 146: change comment - this is for member pr not head

vpr/src/place/place_constraints.cpp
```diff
@@ -113,6 +153,45 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
+    grid_reg.set_region_rect(0, 0, width, height);
+    PartitionRegion grid_pr;
+    grid_pr.add_to_part_region(grid_reg);
+
+    for (auto pl_macro : place_ctx.pl_macros) {
+        if (is_macro_constrained(pl_macro)) {
+            /*
+             * Update the PartitionRegion for the head of the macro
+             * based on the constraints of all blocks contained in the macro
+             */
```
sfkhalid, line 175: don't update PR here, just get the head macro

vpr/src/place/place_constraints.cpp
```diff
@@ -113,6 +153,45 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
+            PartitionRegion macro_head_pr = update_macro_head_pr(pl_macro, grid_pr);
+
+            //Update PartitionRegions of all members of the macro
+            for (size_t imember = 0; imember < pl_macro.members.size(); imember++) {
+                ClusterBlockId iblk = pl_macro.members[imember].blk_index;
+                auto offset = pl_macro.members[imember].offset;
+
+                //Update head PR
```
sfkhalid, line 183: put one line below (beside the if)

vpr/src/place/place_constraints.cpp
sfkhalid: Instead of adding a offset helper function for vtr::Rect, I directly added or subtract the offset to the rectangle values as needed. By doing this, the creation of the different locs was avoided, and applying the offset to the rectangle was reduced to one line.
I chose to do this instead because otherwise I would have needed to add two offset helper functions to vtr::Rect (add and subtract) and also added a mutator to the Region class to set the rectangle by passing in a vtr::Rect instead of four individual values.

sfkhalid:
@vaughnbetz Do the changes on this PR look good?

vpr/src/place/place_constraints.cpp
vaughnbetz: OK, if you prefer this style you can leave it.  It is explicit.

vpr/src/place/place_constraints.cpp
```diff
@@ -80,39 +80,114 @@ PartitionRegion constrained_macro_locs(const t_pl_macro& pl_macro) {
...
 
     return macro_pr;
 }
 
+void print_macro_constraint_error(const t_pl_macro& pl_macro) {
+    VTR_LOG(
+        "Feasible floorplanning constraints could not be calculated for the placement macro. \n"
+        "The placement macro contains the following blocks: \n");
+    for (unsigned int i = 0; i < pl_macro.members.size(); i++) {
+        VTR_LOG("Block #%zu ", size_t(pl_macro.members[i].blk_index));
```
vaughnbetz, line 147: Should also print the block name. Ids are OK for developers but hostile for users.
Usually I do that as "Block %s (#%d)" where the %s is the name and the %d is the id.

vpr/test/test_vpr_constraints.cpp
```diff
@@ -425,6 +426,34 @@ TEST_CASE("RegionLocked", "[vpr]") {
...
 
+//Test calculation of macro constraints
+TEST_CASE("MacroConstraints", "[vpr]") {
+    t_pl_macro pl_macro;
```
vaughnbetz, line 431: What is pl_macro initialized to?  Maybe you're relying on the default constructor?  Better to initialize it I think, and describe what the test does (macro of size 1, macro of size 2, ??).

vaughnbetz (commented):
Two minor comments on the unit test and the error message.  I can merge when CI goes green, but I'd like you to incorporate those two comments as well.

vpr/test/test_vpr_constraints.cpp
sfkhalid: The pl_macro wasn't initialized to anything. It is only passed in to update_macro_member_pr so that if there is an error the error message can be printed with the macro. This won't happen in the case of this unit test, since the test is just checking that the PartRegion is updated properly for a particular case. I can initialize it to something though, and yes, I'll add some comments about what the test is doing.

sfkhalid:
@vaughnbetz Can this be merged now?

PR #1759, opened 2021-06-02
Mark fixed blocks

sfkhalid:
#### Description
Before starting initial placement, the added code loops through the cluster blocks netlist to see whether any of the blocks have floorplan constraints that lock them to one grid location. For the blocks that are thus locked, the .is_fixed flag is set, and their locations are marked, meaning that the blocks will be skipped during initial placement, and they will not be moved during simulated annealing.

#### Related Issue


vpr/src/place/initial_placement.cpp
```diff
@@ -466,6 +466,9 @@ void initial_placement(enum e_pad_loc_type pad_loc_type, const char* constraints
         read_constraints(constraints_file);
     }
 
+    /*Mark fixed blocks*/
```
sfkhalid, line 469: Improve comment

vpr/src/base/region.cpp
```diff
@@ -35,7 +35,7 @@ bool Region::locked() {
 }
 
 bool Region::empty() {
-    return region_bounds.empty();
+    return region_bounds.xmax() < region_bounds.xmin() || region_bounds.ymax() < region_bounds.ymin() || (region_bounds.xmin() == -1 && region_bounds.ymin() == -1 && region_bounds.xmax() == -1 && region_bounds.ymax() == -1);
```
sfkhalid, line 38: instead of having the default check, change the default values of the region_bounds so that they fail this test -> xmax and ymax can be -1, and xmin ymin can be a big number

sfkhalid: another option - could also have this empty as calculating area

vpr/src/base/region.h
```diff
@@ -44,7 +44,8 @@ class Region {
     void set_sub_tile(int _sub_tile);
 
     /**
-     * @brief Return whether the region is empty, based on whether the region rectangle is empty
+     * @brief Return whether the region is empty, meaning the rectangle has impossible dimensions
```
sfkhalid, line 47: change this comment once you make the changes above

vpr/src/place/place_constraints.cpp
```diff
@@ -180,3 +180,36 @@ void load_cluster_constraints() {
...
+    for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
+        PartitionRegion pr = floorplanning_ctx.cluster_constraints[blk_id];
+        std::vector<Region> part_region = pr.get_partition_region();
+        if (part_region.size() == 1) {
+            int subtile = part_region[0].get_sub_tile();
+            vtr::Rect<int> rect = part_region[0].get_region_rect();
+
+            if (part_region[0].locked()) {
+                //Set the location of the block
+                place_ctx.block_locs[blk_id].loc.x = rect.xmin();
```
sfkhalid, line 198: can put the helper function in place_utils

vpr/src/place/place_constraints.cpp
```diff
@@ -180,3 +180,36 @@ void load_cluster_constraints() {
...
+
+void mark_fixed_blocks() {
+    auto& cluster_ctx = g_vpr_ctx.clustering();
+    auto& place_ctx = g_vpr_ctx.mutable_placement();
+    auto& floorplanning_ctx = g_vpr_ctx.floorplanning();
+
+    for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
+        PartitionRegion pr = floorplanning_ctx.cluster_constraints[blk_id];
+        std::vector<Region> part_region = pr.get_partition_region();
+        if (part_region.size() == 1) {
```
sfkhalid, line 192: possible that you could have two of the same rectangles, could have an area function (or two) for PartitionRegion - one that just checks area, one that checks how many tiles it covers because if it covers more than one tile, you would just break out.

vpr/src/place/place_constraints.cpp
```diff
@@ -180,3 +180,36 @@ void load_cluster_constraints() {
...
+                //Mark the grid location of the block
+                place_ctx.grid_blocks[rect.xmin()][rect.ymin()].blocks[subtile] = blk_id;
+
+                //Set as fixed
+                place_ctx.block_locs[blk_id].is_fixed = true;
+
+                //Mark grid location usage
+                place_ctx.grid_blocks[rect.xmin()][rect.ymin()].usage++;
+
+                std::string block_name = cluster_ctx.clb_nlist.block_name(blk_id);
```
sfkhalid, line 211: don't need block name

vpr/src/place/place.cpp
```diff
@@ -2644,8 +2644,8 @@ static int check_block_placement_consistency() {
     for (size_t i = 0; i < device_ctx.grid.width(); i++)
         for (size_t j = 0; j < device_ctx.grid.height(); j++) {
             if (place_ctx.grid_blocks[i][j].usage > device_ctx.grid[i][j].type->capacity) {
-                VTR_LOG_ERROR("Block at grid location (%zu,%zu) overused. Usage is %d.\n",
-                              i, j, place_ctx.grid_blocks[i][j].usage);
+                VTR_LOG_ERROR("%d blocks were placed at grid location (%zu,%zu), but location capacity is %d.\n",
```
sfkhalid, line 2647: Check exactly what capacity means here - is it the number of the distinct type of subtiles, or the total number of things that can go there

vpr/src/place/place_constraints.cpp
```diff
@@ -180,3 +181,28 @@ void load_cluster_constraints() {
...
+    auto& floorplanning_ctx = g_vpr_ctx.floorplanning();
+
+    for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
+        PartitionRegion pr = floorplanning_ctx.cluster_constraints[blk_id];
+        std::vector<Region> part_region = pr.get_partition_region();
+        if (part_region.size() == 1) {
+            int subtile = part_region[0].get_sub_tile();
+            vtr::Rect<int> rect = part_region[0].get_region_rect();
+
+            if (part_region[0].locked()) {
```
sfkhalid, line 197: get rid of .locked() once you add the utility function because then it is no longer needed.

vpr/src/place/place_constraints.h
```diff
@@ -64,4 +64,10 @@ inline bool floorplan_legal(const t_pl_blocks_to_be_moved& blocks_affected) {
  */
 void load_cluster_constraints();
 
+/*
+ * Marks blocks with a region with xlow = xhigh, ylow = yhigh, and subtile specified as fixed
```
sfkhalid, line 68: Make this comment say that it's for when region specifies exactly one x, y, subtile location as legal -> that's when it gets marked as fixed

vpr/src/place/place_util.cpp
```diff
@@ -427,3 +427,38 @@ void alloc_and_load_legal_placement_locations(std::vector<std::vector<std::vecto
     //avoid any memory waste
     legal_pos.shrink_to_fit();
 }
+
+void mark_block_location(ClusterBlockId blk_id, const t_pl_loc& location) {
```
sfkhalid, line 431: could rename to mark_blk_grid_loc to reflect grid changes too

vpr/src/place/place.cpp
sfkhalid: Number of blocks of that type that can occupy the grid tile.

sfkhalid:
QoR results on this branch: 
[mark_fixed_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6670311/mark_fixed_qor_comparison.xlsx)

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,95 @@ void load_cluster_constraints() {
...
+    for (unsigned int i = 0; i < regions.size(); i++) {
+        reg_size_one = is_region_size_one(regions[i], block_type);
+        if (!reg_size_one) {
+            pr_size_one = false;
+            break;
+        } else {
+            if (i == 0) {
+                intersect_reg = regions[0];
+            } else {
+                intersect_reg = intersection(intersect_reg, regions[i]);
```
sfkhalid, line 331: Comment logic behind this section (why are we getting intersection, etc.)

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,95 @@ void load_cluster_constraints() {
...
+            } else {
+                pr_size_one = true;
+            }
+        }
+    }
+
+    vtr::Rect<int> rect = intersect_reg.get_region_rect();
+    loc.x = rect.xmin();
+    loc.y = rect.ymin();
+    if (intersect_reg.get_sub_tile() == NO_SUBTILE) {
```
sfkhalid, line 345: Comment why this works (dependency on how is_reg_size_one works)

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,95 @@ void load_cluster_constraints() {
...
+    }
+}
+
+bool is_region_size_one(const Region& reg, t_logical_block_type_ptr block_type) {
+    auto& device_ctx = g_vpr_ctx.device();
+    vtr::Rect<int> rb = reg.get_region_rect();
+    bool size_one = false;
+
+    if (rb.xmin() == rb.xmax() && rb.ymin() == rb.ymax()) {
+        if (reg.get_sub_tile() != NO_SUBTILE) {
```
sfkhalid, line 291: Don't need to put an error message, but comment saying that later on the code will catch is subtile is invalid at this location

sfkhalid: check whether the the subtile is compatible at this tile location -> if it is, return true, if not, return false

sfkhalid: essentially, this routine returns false if it is bigger than one or less than one

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,95 @@ void load_cluster_constraints() {
...
+            if (is_tile_compatible(tile, block_type)) {
+                auto cap = tile->capacity;
+                if (cap == 1) {
+                    size_one = true;
+                } else {
+                    size_one = false;
+                }
+            } else {
+                return false;
+                //give an error because somehow tile and block type not compatible;
```
sfkhalid, line 306: don't need error message here

vpr/src/place/place_constraints.cpp
sfkhalid: to differentiate the cases, you could make an enumerated type - reg covers 0, 1, or more than one loc. return this type instead of a bool

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,95 @@ void load_cluster_constraints() {
...
+            //Set block location and grid usage
+            set_block_location(blk_id, loc);
+
+            //Set as fixed
+            place_ctx.block_locs[blk_id].is_fixed = true;
+        }
+    }
+}
+
+bool is_region_size_one(const Region& reg, t_logical_block_type_ptr block_type) {
```
sfkhalid, line 285: could also loop through all grid locations in the region to see whether it covers 0, 1, or more than one tile. This would also then cover the case where a bigger region only essentially covers one legal loc

sfkhalid: given this case -> could also just return int and have 0, 1, 2 as possible return values -> make sure this is commented very clearly in this case

vpr/src/place/place_constraints.h
```diff
@@ -88,4 +88,18 @@ inline bool floorplan_legal(const t_pl_blocks_to_be_moved& blocks_affected) {
...
 
+/*
+ * Marks blocks as fixed if they have a constraint region that specifies exactly one x, y,
+ * subtile location as legal.
+ * Marking them as fixed indicates that they cannot be moved during initial placement and simulated annealing
+ */
+void mark_fixed_blocks();
+
+/*
+ * Number of tiles covered by a region
```
sfkhalid, line 99: Comment doesn't match what is happening right now, also add more detail about what is passed in and what happens (what is returned)

vpr/src/place/place_constraints.h
```diff
@@ -88,4 +88,18 @@ inline bool floorplan_legal(const t_pl_blocks_to_be_moved& blocks_affected) {
...
+ */
+bool is_region_size_one(const Region& reg, t_logical_block_type_ptr block_type);
+
+bool is_pr_size_one(PartitionRegion& pr, t_logical_block_type_ptr block_type, t_pl_loc& loc);
```
sfkhalid, line 103: comment this as well

sfkhalid:
@vaughnbetz Does this look good to merge? 
I moved the initialization of grid_blocks so that the load placement flow and load constraints file can both use the utility function I created, set_block_location. I also refactored the region tile area function to return an int instead of a bool.

sfkhalid:
I'm currently waiting for the QoR results to finish running, once that happens I will post the QoR results for the updated code

sfkhalid:
New QoR comparison: 
[mark_fixed_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6677835/mark_fixed_qor_comparison.xlsx)

vpr/src/base/read_place.cpp
```diff
@@ -185,6 +186,23 @@ void read_place_body(std::ifstream& placement_file,
         place_ctx.block_locs.resize(cluster_ctx.clb_nlist.blocks().size());
     }
 
+    /*
+     * If placement is being loaded (i.e. reading in a place file),
+     * the grid_blocks data structure has not been initialized yet,
+     * so it is initialized here.
+     */
+    if (is_place_file) {
+        auto& grid_blocks = place_ctx.grid_blocks;
```
vaughnbetz, line 195: This should be moved to a helper function that creates grid_blocks and is called in both flows (read place or don't read place).  Could be in place_utils.cpp, or in some other file that deals with grid blocks and related utilities (if such a file exists).

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,149 @@ void load_cluster_constraints() {
...
+    for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
+        if (!is_cluster_constrained(blk_id)) {
+            continue;
+        }
+        PartitionRegion pr = floorplanning_ctx.cluster_constraints[blk_id];
+        auto block_type = cluster_ctx.clb_nlist.block_type(blk_id);
+        t_pl_loc loc;
+
+        if (is_pr_size_one(pr, block_type, loc)) {
+            //Set block location and grid usage
```
vaughnbetz, line 276: Better to give a single higher level comment.  If the block can be placed in exactly one legal (x,y,sub_tile) location, place it now and mark it as fixed.  The current comments don't really add anything useful, as the function called or the data member set makes it clear what that line is doing.

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,149 @@ void load_cluster_constraints() {
...
+            }
+
+            /*
+             * If the region passed has a specific subtile set, increment
+             * the number of tiles set the location using the x, y, subtile
+             * values
+             */
+            if (reg.get_sub_tile() != NO_SUBTILE) {
+                num_tiles++;
+                loc.x = x;
```
vaughnbetz, line 318: I think you need to check if this particular sub_tile is the right type for this block.  The is_tile_compatible call above presumably only checked that *some* sub_tile at this x,y location is compatible with this block type not that *the proposed* sub_tile number is compatible.

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,149 @@ void load_cluster_constraints() {
...
+                 * If the region passed does not have a subtile set, set the
+                 * subtile to zero. The loc that is set will only be used in the
+                 * event that num_tiles is 1, and num_tiles will only be 1 if the
+                 * capacity of the tile type turns out to be zero, thus implying
+                 * that the location subtile will be zero.
+                 */
+            } else if (reg.get_sub_tile() == NO_SUBTILE) {
+                auto& cap = tile->capacity;
+                for (int z = 0; z < cap; z++) {
+                    num_tiles++;
```
vaughnbetz, line 335: Again, don't we have to check if *this* subtile (z) is compatible in type, and only increment num_tiles if it is?  And then we should set sub_tile to the first z value that passes that test.

vpr/src/place/place_constraints.cpp
```diff
@@ -257,3 +258,149 @@ void load_cluster_constraints() {
...
+                pr_size = pr_size + reg_size;
+            } else {
+                intersect_reg = intersection(intersect_reg, regions[i]);
+            }
+            if (intersect_reg.empty()) {
+                pr_size = pr_size + reg_size;
+                if (pr_size > 1) {
+                    break;
+                }
+            } else {
```
vaughnbetz, line 388: This else continue clause is harder to read than it needs to be (more spaghetti code).
Instead, make the whole if (reg_size ==1) condition have an else { clause that has lines 393 - 396 in it (and then delete the continue in 388 - 390).

It may be possible to restructure the ifs more; this is a fairly complicated set of if statements for logic that isn't that fundamentally hard.

vpr/src/place/place_constraints.h
```diff
@@ -88,4 +88,27 @@ inline bool floorplan_legal(const t_pl_blocks_to_be_moved& blocks_affected) {
...
+ * Returns the number of tiles covered by a floorplan region.
+ * The return value of this routine will either be 0, 1, or 2. This
+ * is because this routine is used to check whether the region covers no tile,
+ * one tile, or more than one tile, and so as soon as it is seen that the number of tiles
+ * covered is 2, no further information is needed.
+ */
+int region_tile_cover(const Region& reg, t_logical_block_type_ptr block_type, t_pl_loc& loc);
+
+/*
+ * Returns whether the PartitionRegion covers no tiles, 1 tile, or more than 1 tile.
```
vaughnbetz, line 108: Comment seems wrong.  Returns a bool that specifies if a region covers exactly one compatible location.
Also mention the use block_type and &loc.

vpr/src/place/place_util.cpp
```diff
@@ -427,3 +427,38 @@ void alloc_and_load_legal_placement_locations(std::vector<std::vector<std::vecto
     //avoid any memory waste
     legal_pos.shrink_to_fit();
 }
+
+void set_block_location(ClusterBlockId blk_id, const t_pl_loc& location) {
+    auto& place_ctx = g_vpr_ctx.mutable_placement();
+    auto& device_ctx = g_vpr_ctx.device();
+    auto& cluster_ctx = g_vpr_ctx.clustering();
+
+    std::string block_name = cluster_ctx.clb_nlist.block_name(blk_id);
```
vaughnbetz, line 436: Is this routine speed critical (used in try_swap)?  If so, worth noting that in a comment.

vaughnbetz: Faster to get a reference to block_name via
const std::string& block_name = ...
Avoids a string copy, which is a good idea if this routine could ever be called in a speed-critical routine (and it might be, if try_swap etc. use it).  Better style, so I'd change this even if this routine isn't currently called in speed critical routines.

vpr/src/place/place_util.h
```diff
@@ -218,4 +218,7 @@ void load_grid_blocks_from_block_locs();
 
 ///@brief Builds legal_pos structure. legal_pos[type->index] is an array that gives every legal value of (x,y,z) that can accommodate a block.
 void alloc_and_load_legal_placement_locations(std::vector<std::vector<std::vector<t_pl_loc>>>& legal_pos);
+
+///@brief Marks the location and grid usage for a block that is set at a location
```
vaughnbetz, line 222: Should also mention that it performs error checking on whether this is a legal location or not for a block of that type, and prints and error if it is not legal.

vaughnbetz (changes requested):
Some changes requested.

vpr/src/place/place_util.cpp
sfkhalid: Not currently called by try_swap, but I changed code to const std::string& block_name as suggested.

sfkhalid:
@vaughnbetz Does this look ok to merge once the CI goes green?
I made the changes requested on Friday, like simplifying the if statements in is_pr_size_one, adding the checks for subtile compatibility in region_tile_cover and performing the initialization of placement contexts with a helper routine during the load placement flow.

vaughnbetz:
The changes look good; will merge when CI goes green.

One final feedback: using loop counters that indicate what you are looping over is a bit better than just using i.  E.g. in the code looping over regions, you could use one of ireg/iregion/irect instead of i, and it would make the loop a bit easier to understand.

sfkhalid:
@vaughnbetz The CI has passed, is this PR good to be merged now?

PR #1757, opened 2021-06-02
CI: Drop GCC 5&6 support

Xiretza:
#### Description
See #1754 for the motivation.

GCC 5 and 6 were still used for many CI runs, so the default was updated to GCC 11. This created a few new warnings that didn't exist in the older compilers, causing the `-Werror` job to fail - I've included the fixes in this pull request.

#### Related Issue
Resolves #1754

#### How Has This Been Tested?
All CI jobs work (again).

#### Types of changes
- [x] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1760, opened 2021-06-02
Makefile: fix BUILD_TYPE=*_strict

Xiretza:
#### Description

The `BUILD_TYPE` variable can be used to select the CMake build type when using the wrapper Makefile. Additionally, a `_strict` suffix is supposed to enable VTR's strict compilation mode, but the name of the CMake variable responsible for doing this changed a while ago without the Makefile being updated. Thus, the request for a strict compilation was effectively ignored.

While the actual fix for this is fairly straightforward, I also tried to make sure similar errors may get caught early in the future by adding the `-Werror=dev` flag to CMake invocations in CI. This may eventually also catch unused variable warnings (which [were present](https://github.com/verilog-to-routing/vtr-verilog-to-routing/runs/2724152095#step:6:156) in the CI runs), but they are currently [not covered by the option](https://gitlab.kitware.com/cmake/cmake/-/issues/21412).

#### Motivation and Context

I happened to notice this while figuring out how the `BUILD_TYPE` machinery works for #1757.

`rm -rf build && make BUILD_TYPE=release_strict` used to output:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    VTR_COMPILE_OPTIONS
```

Now the warning is gone and it outputs:

```
-- VPR: building with strict flags
-- 	Adding CXX flag: -Werror
```

#### Types of changes

#### Checklist:


mithro:
@Xiretza - This is really great!

.github/travis/setup.sh
```diff
@@ -1,42 +0,0 @@
-#!/bin/bash
```
mithro, line 1: I think it would still be useful to output the exact version information for the toolchain if we don't do it elsewhere?

Xiretza: CMake outputs the versions of `$CC` and `$CXX` ([example](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1760/checks?check_run_id=2731164444#step:6:15)), do we need anything else? Note that the commands in this script never ran on GHA because `$TRAVIS_BUILD_STAGE_NAME` is empty.

mithro: SGTM!

mithro (approved):

Issue #1748, opened 2021-05-29
[ODIN]  Assertion failure with ripple carry architecture files 


ArashAhmadian:
Synthesis fails with [pipelined_fft_64.v](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/benchmarks/arithmetic/generated_circuits/opencores/verilog/pipelined_fft_64.v) given a ripple carry architecture file . 
#### Expected Behaviour
Should synthesize 
#### Current Behaviour
The following assertion fails: 
```/home/ahmadi55/vtr-verilog-to-routing/ODIN_II/SRC/subtractions.cpp:467: split_adder_for_sub: Assertion not_node[index]->input_pins[0]->net->num_driver_pins == 1 failed
   465 |                      * if the input pin of not gate connects to vcc, replacing the input pin and the not gate with gnd.*/
   466 |                     const int index = i * sizeb + j - offset;
   467 |                     oassert(not_node[index]->input_pins[0]->net->num_driver_pins == 1);
        ^~~~
   468 |                     if (not_node[index]->input_pins[0]->net->driver_pins[0]->node->type == GND_NODE) {
   469 |                         connect_nodes(netlist->vcc_node, 0, node[i], (sizea + j));
Command terminated by signal 6
```
#### Steps to Reproduce
Run vtr_flow with the following parameters: 

- circuit: [pipelined_fft_64.v](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/benchmarks/arithmetic/generated_circuits/opencores/verilog/pipelined_fft_64.v)
- architecture: Any ripple carry architecture e.g- [k6_N8_unbalanced_ripple_chain_gate_boost_0.2V](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/COFFE_22nm/k6_N8_unbalanced_ripple_chain_gate_boost_0.2V_22nm.xml)

#### Context
This issue arose while running the following tasks: [open_cores](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/arithmetic_tasks/open_cores/config/config.txt) & [open_cores_frac](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/arithmetic_tasks/open_cores_frac/config/config.txt) while in the process of updating regression_tests.

[...]

Issue #1738, opened 2021-05-25
VNumber copy constructor with a new length does set defined_size attribute


sdamghan:
#### Expected Behaviour

When a new VNumber is created using a copy constructor with a new length, the `defined_size` value should set to `true`. 

#### Current Behaviour

A new VNumber is created with a new size while the `defined_size` is false. 

#### Possible Solution

Creating a new VNumber using copy constructor with a new length, the `defined_size` should set to `true`.

#### Steps to Reproduce

1. Run Odin_II with the following test case:
`${VTR_ROOT}/ODIN_II/odin_ii -V` [minus_operator.v](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6535961/minus_operator.v.txt)

#### Your Environment

* VTR revision used: 8.1.0-dev+00617ff76
* Operating System and version: 20.04.1-Ubuntu SMP
* Compiler version: GNU 9.3.0 on Linux-5.8.0-53-generic x86_64

ArashAhmadian:
@sdamghan PR #1739 didn't fix the concatenation issue we discussed in email. 
![Capture](https://user-images.githubusercontent.com/70601261/120024758-a4d3cc00-bfbd-11eb-90b7-a1367800d63a.JPG)
It has the same behaviour as prior to the PR.

sdamghan:
@ArashAhmadian That's weird! I've just checked the pipelined_fft_64.v_ with Odin_II, and it worked properly.
Would you compile again and run the following command in the ODIN_II directory? Please let me know the result.
`./odin_II -V ../vtr_flow/benchmarks/arithmetic/generated_circuits/opencores/verilog/pipelined_fft_64.v`

ArashAhmadian:
> @ArashAhmadian That's weird! I've just checked the pipelined_fft_64.v_ with Odin_II, and it worked properly.
> Would you compile again and run the following command in the ODIN_II directory? Please let me know the result.
> `./odin_II -V ../vtr_flow/benchmarks/arithmetic/generated_circuits/opencores/verilog/pipelined_fft_64.v`

@sdamghan I had forgotten to compile ODIN after pulling :) (sorry) . It synthesizes properly based on the results of the command you sent.  However, when running the open_core task an assertion fails: 
```
[...]

PR #1761, opened 2021-06-02
Connecting untouched nets to the pad node in resolving subtraction node

sdamghan:

#### Description

Undriven nets would cause segfault in resolving subtraction nodes. Solution is provided by connecting the to the pad node

#### Related Issue

#1748 #1738

#### Motivation and Context

bug fix

`make test`

#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/SRC/subtractions.cpp
+++ b/ODIN_II/SRC/subtractions.cpp
@@ -45,6 +45,7 @@ int subchaintotal = 0;
 int* sub = NULL;
 
 void init_split_adder_for_sub(nnode_t* node, nnode_t* ptr, int a, int sizea, int b, int sizeb, int cin, int cout, int index, int flag);
+static void cleanup_sub_old_node(nnode_t* nodeo, netlist_t* netlist);
 
 /*---------------------------------------------------------------------------
  * (function: report_sub_distribution)
@@ -405,6 +406,11 @@ void split_adder_for_sub(nnode_t* nodeo, int a, int b, int sizea, int sizeb, int
         for (i = 0; i < b; i++) {
             /* If the input pin of not gate connects to gnd, replacing the input pin and the not gate with vcc;
              * if the input pin of not gate connects to vcc, replacing the input pin and the not gate with gnd.*/
+            /* connecting untouched nets in the netlist creation to the pad node */
+            if (not_node[i]->input_pins[0]->net->num_driver_pins == 0) {
+                /* join untouched net with pad net */
+                join_nets(netlist->pad_net, not_node[i]->input_pins[0]->net);
+            }
             oassert(not_node[i]->input_pins[0]->net->num_driver_pins == 1);
             if (not_node[i]->input_pins[0]->net->driver_pins[0]->node->type == GND_NODE) {
                 connect_nodes(netlist->vcc_node, 0, node[0], (lefta + i));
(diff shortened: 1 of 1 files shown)


ODIN_II/SRC/subtractions.cpp
```diff
@@ -449,6 +459,11 @@ void split_adder_for_sub(nnode_t* nodeo, int a, int b, int sizea, int sizeb, int
                 if (i == count - 1 && flag == 1) {
                     /* If the input pin of not gate connects to gnd, replacing the input pin and the not gate with vcc;
                      * if the input pin of not gate connects to vcc, replacing the input pin and the not gate with gnd.*/
+                    /* connecting untouched nets in the netlist creation to the pad node */
```
vaughnbetz, line 462: Should delete the duplicate comment (repeated line).

ODIN_II/SRC/subtractions.cpp
```diff
@@ -555,15 +575,36 @@ void split_adder_for_sub(nnode_t* nodeo, int a, int b, int sizea, int sizeb, int
     //connect_nodes(node[count - 1], (node[(count - 1)]->num_output_pins - 1), netlist->gnd_node, 0);
     //}
 
-    /* Probably more to do here in freeing the old node! */
-    vtr::free(nodeo->name);
-    vtr::free(nodeo->input_port_sizes);
-    vtr::free(nodeo->output_port_sizes);
+    /* Freeing the old node! */
```
vaughnbetz, line 578: This routine is about 250 lines long ... would be better to shorten where possible.  This code addition looks like it could be a helper function that you call with only nodeo passed in.  That would make it a great candidate for a helper function. Any thoughts?

sdamghan: We already have a `free_nnode(node)` helper function in netlist utils; this situation is pretty specific for the freeing process since there is weird usage of pointers. However, I couldn't agree more with what you said. I will create a short helper function for this specific file.

ODIN_II/SRC/subtractions.cpp
sdamghan: So sorry, I uploaded the code quickly to see whether there is any failure or not.

ODIN_II/SRC/subtractions.cpp
```diff
@@ -654,3 +668,52 @@ void clean_adders_for_sub() {
         processed_adder_list = delete_in_vptr_list(processed_adder_list);
     return;
 }
+
+/**
+ * -------------------------------------------------------------------------
+ * (function: cleanup_sub_old_node)
+ *
+ * @brief Clean up the old minus node based on the
```
vaughnbetz, line 676: I'd add a bit more context in the comment:  nodeo is a subtraction node that has been replaced by .... (logic in split_adder_for_sub?) ... so this function cleans up by removing nodeo and associated net etc. data from the netlist.

vaughnbetz (approved):
Thanks @sdamghan .  I'll merge this, but I have one more suggestion to make the function level comment a bit more detailed / explict.

ODIN_II/SRC/subtractions.cpp
sdamghan: For sure, thank you for such a good suggestions

ODIN_II/SRC/subtractions.cpp
```diff
@@ -654,3 +668,57 @@ void clean_adders_for_sub() {
...
+ * (function: cleanup_sub_old_node)
+ *
+ * @brief <clean up nodeo, a high level MINUS node> 
+ * In split_adder_for_sub function, nodeo is splitted to small adders/subtractors, 
+ * while because of the complexity of input pin connections they have not been 
+ * remapped to new nodes, they just copied and added to new nodes. This function 
+ * will detach input pins from the nodeo. Moreover, it will connect the net of 
+ * unconnected output signals to the GND node, detach the pin from nodeo and 
+ * free the output pins to avoid memory leak.
+ *
```
sdamghan, line 683: @vaughnbetz  Hoping this one is better and more comprehensive :)

vaughnbetz:
Looks good, thanks!

PR #1762, opened 2021-06-03
Add a template submission script for SLURM-managed clusters

MohamedElgammal:
#### Description

This PR adds a template SLURM submission script that can be used to submit a large VTR task run on a SLURM-managed cluster. I have also added some documentation on how to use this script to run easily using slurm.

The PR is tested on computecanada cluster (Niagara system)
#### Types of changes

#### Checklist:

### DIFF
--- a/README.developers.md
+++ b/README.developers.md
@@ -280,6 +280,9 @@ You can also run multiple regression tests together:
 #Run both the basic and strong regression, with up to 4 tests in parallel
 $ ./run_reg_test.py vtr_reg_basic vtr_reg_strong -j4
 ```
+## Running in a large cluster using SLURM
+For the very large runs, you can submit your runs on a large cluster. A template of submission script to 
+Slurm-managed cluster can be found under vtr_flow/tasks/slurm/
 
 ## Odin Functionality Tests
 
--- a/README.developers.md
+++ b/README.developers.md
@@ -282,7 +282,7 @@ $ ./run_reg_test.py vtr_reg_basic vtr_reg_strong -j4
 ```
 ## Running in a large cluster using SLURM
 For the very large runs, you can submit your runs on a large cluster. A template of submission script to 
-Slurm-managed cluster can be found under vtr_flow/tasks/slurm/
+a Slurm-managed cluster can be found under vtr_flow/tasks/slurm/
 
 ## Odin Functionality Tests
 
--- a/vtr_flow/scripts/slurm/submission_template.sh
+++ b/vtr_flow/scripts/slurm/submission_template.sh
@@ -1,5 +1,13 @@
 #! /bin/bash
 
+#======================================================================================
+#Please Make a copy of this template in the same directory and edit it to fit your run.
+#Please do not submit your copy of this template to the repo.
+#======================================================================================
+
+
+#The following section controls the parameters of your run, feel free to edit:
+#=============================================================================
 #SBATCH --nodes=1                                       # Number of nodes in our pool
 #SBATCH --ntasks=80                                     # Number of parallel tasks
 #SBATCH --ntasks-per-node=40                            # maximum number of tasks simultaneously running on a node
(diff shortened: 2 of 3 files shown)


mithro:
PTAL - @QuantamHD

vtr_flow/scripts/slurm/submission_template.sh
```diff
@@ -0,0 +1,27 @@
+#! /bin/bash
+
+#SBATCH --nodes=1                                       # Number of nodes in our pool
+#SBATCH --ntasks=80                                     # Number of parallel tasks
+#SBATCH --ntasks-per-node=40                            # maximum number of tasks simultaneously running on a node
+#SBATCH --job-name=vtr_test                             # Job name
+#SBATCH --mail-type=FAIL,END                            # When to send an email
+#SBATCH --mail-user=XXXX@mail.utoronto.ca               # The user's email
```
QuantamHD, line 8: Is there anyway to template these SBATCH directives with environment variables?

MohamedElgammal: Sure, we can! 
Just by replacing all of them with
`#SBATCH --export=ALL`

and let the user set the following environment variables:
- SLURM_JOB_NUM_NODES instead of --nodes
- SLURM_NTASKS or SLURM_NPROCS instead of --ntasks
- SLURM_NTASKS_PER_NODE instead of --ntasks-per-node
- SLURM_JOB_ID instead of %j
- SBATCH_JOB_NAME instead of --job-name

However, the question is which one will be easier to use .. the environment variables or the script local ones?

QuantamHD: I think editing the file is somewhat more difficult for 2 reasons, but I would understand if you disagree

1. By editing the file users now need to remove any local changes for their clusters before submitting PRs to the repo. It's also more likely that a subtle change will slip through the cracks into the repo.
2. When deploying to different cloud environments you often want to have the ability to change the configuration depending on the environment. Doing this through env variables is easier than doing it through file modifications. 

To me I think a good compromise would be to use the environment mode, but include a comment with the most common/required fields.

MohamedElgammal: Makes sense. I tested it and unfortunately it seems that most of these environment variables are output from SLURM based on the allocated nodes/tasks/tasks-per-node. The only input environment variables I can found are:
- SBATCH_JOB_NAME instead of --job-name
- SBATCH_TIMELIMIT instead of --time

For all the other parameters, I am not able to find an env. variable to overwrite them. Hence, I have added more commenting in the script explaining these options.

QuantamHD: Interesting, this probably means in the long run we would have to dynamically generate a bash file in order to template slurm jobs. This change seems good for now given the current landscape, but I would love to see this become more dynamically configurable.

README.developers.md
```diff
@@ -280,6 +280,9 @@ You can also run multiple regression tests together:
 #Run both the basic and strong regression, with up to 4 tests in parallel
 $ ./run_reg_test.py vtr_reg_basic vtr_reg_strong -j4
 ```
+## Running in a large cluster using SLURM
+For the very large runs, you can submit your runs on a large cluster. A template of submission script to
```
vaughnbetz, line 284: "to a Slurm"

vtr_flow/scripts/slurm/submission_template.sh
```diff
@@ -0,0 +1,40 @@
...
+
+#The following section controls the parameters of your run, feel free to edit:
+#=============================================================================
+#SBATCH --nodes=1                                       # Number of nodes in our pool
+#SBATCH --ntasks=80                                     # Number of parallel tasks
+#SBATCH --ntasks-per-node=40                            # maximum number of tasks simultaneously running on a node
+#SBATCH --job-name=vtr_test                             # Job name
+#SBATCH --mail-type=FAIL,END                            # When to send an email
+#SBATCH --mail-user=XXXX@mail.utoronto.ca               # The user's email
+#SBATCH --output=parallel_%j.log                        # Determine the output log file (%j) is the jobid
```
vaughnbetz, line 17: file; (%j) is the

vaughnbetz:
Good discussion.  
I discussed this with Mohamed some and propose we put the template back, but (once we have multiple clusters & people using them), we make a higher-level script called
sbatch_env   (or some such)
that takes a defined set of environment variables (if they exist) and writes them to the slurm submission script file. If the environment variables don't exist, then the defaults in the template slurm submission script (either in a file or embedded in sbatch_env) are used for any missing variables.  
Then the documentation could describe using sbatch_env as the main way to submit jobs, and explain how to set the proper environment variables (and have a template for setting them :).

vaughnbetz (commented):
Just a few typos to fix ...

PR #1764, opened 2021-06-03
clang-tidy fixes around const std::string references.

mithro:
#### Description

This is the [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/) fixes that @luk036 had in his pull request at https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1078

These changes make the code follow closer to best practices around using const std::string references.

### DIFF
--- a/ODIN_II/SRC/Hashtable.cpp
+++ b/ODIN_II/SRC/Hashtable.cpp
@@ -29,33 +29,33 @@
 #include "vtr_memory.h"
 
 void Hashtable::destroy_free_items() {
-    for (auto kv : my_map)
+    for (const auto& kv : my_map)
         vtr::free(kv.second);
 }
 
-void Hashtable::add(std::string key, void* item) {
-    this->my_map.emplace(key, item);
+void Hashtable::add(const std::string& key, void* item) {
+    this->my_map.insert({key, item});
 }
 
-void* Hashtable::remove(std::string key) {
+void* Hashtable::remove(const std::string& key) {
     void* value = NULL;
     auto v = this->my_map.find(key);
     if (v != this->my_map.end()) {
-        value = v->second;
-        this->my_map.erase(v);
+        value = this->my_map[key];
+        this->my_map.erase(key);
     }
     return value;
 }
 
-void* Hashtable::get(std::string key) {
+void* Hashtable::get(const std::string& key) {
     void* value = NULL;
     auto v = this->my_map.find(key);
     if (v != this->my_map.end())
-        value = v->second;
+        value = this->my_map[key];
 
     return value;
 }
 
 bool Hashtable::is_empty() {
-    return my_map.empty();
+    return (my_map.size() == 0);
 }
--- a/ODIN_II/SRC/ast_util.cpp
+++ b/ODIN_II/SRC/ast_util.cpp
@@ -29,6 +29,7 @@
 #include <stdarg.h>
 #include <math.h>
 #include <algorithm>
+#include <utility>
 #include "odin_globals.h"
 #include "odin_types.h"
 
@@ -520,7 +521,7 @@ void change_to_number_node(ast_node_t* node, VNumber number) {
 
     node->type = NUMBERS;
     node->types.identifier = temp_ident;
-    node->types.vnumber = new VNumber(number);
+    node->types.vnumber = new VNumber(std::move(number));
 }
 
 /*---------------------------------------------------------------------------------------------
--- a/abc/src/base/abci/abcTiming.c
+++ b/abc/src/base/abci/abcTiming.c
@@ -45,7 +45,7 @@ struct Abc_ManTime_t_
 
 #define TOLERANCE 0.001
 
-static inline int          Abc_FloatEqual( float x, float y )     { return fabs(x-y) < TOLERANCE; }
+static inline int          Abc_FloatEqual( float x, float y )     { return fabsf(x-y) < TOLERANCE; }
 
 // static functions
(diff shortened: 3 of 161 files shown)


mithro:
This change isn't quite ready to be merged yet as it currently includes changes to code under `libs/EXTERNAL`. These need to be extracted out of the pull request and sent to the appropriate external projects.

ODIN_II/SRC/Hashtable.cpp
```diff
@@ -29,33 +29,33 @@
...
 }
 
-void* Hashtable::remove(std::string key) {
+void* Hashtable::remove(const std::string& key) {
     void* value = NULL;
     auto v = this->my_map.find(key);
     if (v != this->my_map.end()) {
-        value = v->second;
-        this->my_map.erase(v);
+        value = this->my_map[key];
```
hzeller, line 44: this is not a good strategy, the original `v = v->second` is what we want.
Because otherwise what we do here is to look up the value _again_ which we already did before.

ODIN_II/SRC/Hashtable.cpp
```diff
@@ -29,33 +29,33 @@
...
+        this->my_map.erase(key);
```
hzeller, line 45: also, this also doesn't look like the right strategy. Here _again_ we force the map to look up the value from the key.
But we already have the iterator. So here as well, the original code is better.

ODIN_II/SRC/Hashtable.cpp
```diff
@@ -29,33 +29,33 @@
...
     return value;
 }
 
-void* Hashtable::get(std::string key) {
+void* Hashtable::get(const std::string& key) {
     void* value = NULL;
     auto v = this->my_map.find(key);
     if (v != this->my_map.end())
-        value = v->second;
+        value = this->my_map[key];
```
hzeller, line 54: Same as above, this should be the original `v->second`

ODIN_II/SRC/Hashtable.cpp
```diff
@@ -29,33 +29,33 @@
...
 
     return value;
 }
 
 bool Hashtable::is_empty() {
-    return my_map.empty();
+    return (my_map.size() == 0);
```
hzeller, line 60: This as well should be the original code `my_map.empty()`, not call to size().

ODIN_II/SRC/Hashtable.cpp
```diff
@@ -29,33 +29,33 @@
...
     void* value = NULL;
     auto v = this->my_map.find(key);
     if (v != this->my_map.end())
-        value = v->second;
+        value = this->my_map[key];
 
     return value;
 }
 
 bool Hashtable::is_empty() {
```
hzeller, line 59: We might want to make this `is_empty()` call a const method. As well Hashtable::get(). But I guess this is out of the scope for this change.

ODIN_II/SRC/ast_util.cpp
```diff
@@ -520,7 +521,7 @@ void change_to_number_node(ast_node_t* node, VNumber number) {
 
     node->type = NUMBERS;
     node->types.identifier = temp_ident;
-    node->types.vnumber = new VNumber(number);
+    node->types.vnumber = new VNumber(std::move(number));
```
hzeller, line 524: I'd probably keep changes that involve std::move in a separate change to keep it by  `const std::string &`.

(I'd only use std::move if really needed. It more uglifies code otherwise)

abc/src/base/abci/abcTiming.c
```diff
@@ -45,7 +45,7 @@ struct Abc_ManTime_t_
 
 #define TOLERANCE 0.001
 
-static inline int          Abc_FloatEqual( float x, float y )     { return fabs(x-y) < TOLERANCE; }
+static inline int          Abc_FloatEqual( float x, float y )     { return fabsf(x-y) < TOLERANCE; }
```
hzeller, line 48: The abc/ directory is probably also something that is just imported and we shouldn't touch too much.

vpr/src/power/PowerSpicedComponent.cpp
```diff
@@ -84,7 +85,7 @@ PowerCallibSize* PowerCallibInputs::get_entry_bound(bool lower,
 
 PowerSpicedComponent::PowerSpicedComponent(std::string component_name,
                                            float (*usage_fn)(int num_inputs, float transistor_size)) {
-    name = component_name;
+    name = std::move(component_name);
```
hzeller, line 88: instead of calling std::move here, we should pass a `const std::string &` as parameter, and just use assignment here.

vpr/src/route/clock_connection_builders.cpp
```diff
@@ -131,19 +132,19 @@ int RoutingToClockConnection::create_virtual_clock_network_sink_node(
  */
 
 void ClockToClockConneciton::set_from_clock_name(std::string clock_name) {
-    from_clock = clock_name;
+    from_clock = std::move(clock_name);
```
hzeller, line 135: all these: instead of using std::move from a by-value parameter, we should use assignment and pass in a `const std::string &`

hzeller (changes requested):
I did a first quick scan, but these are a lot of files, and there are many questionable 'improvements' in there.

So I suggest to find the clang-tidy changes that _only_ deal with `std::string` -> `const std::string &` in parameters as first step, which makes it easier to blanket review as it not contains hidden 'gems' of other things in there.

vaughnbetz:
@luk036 : feedback from @hzeller on these clang-tidy changes.  Looks like we'd only want a subset of them.

PR #1766, opened 2021-06-04
Gitpod configuration

sdamghan:
#### Description

1. The origin remote of the repository in Gitpod was the Loksu fork -> changed to the main VTR repo
2. Running tasks in the `gitpod.yml` caused errors at the starting point of the Gitpod -> removed
3. Setting the working directory for the docker file as VTR_ROOT env.

#### Motivation and Context

wrong URL and removing errors

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1771, opened 2021-06-09
Building VTR in debug mode in GitPod Initialization

sdamghan:
signed-off-by: Seyed Alireza Damghani <sdamghan@unb.ca>

#### Description

VTR will be compiled in debug mode in GitPod prebuilt.

#### Related Issue

follow up on #1766

#### Types of changes

#### Checklist:

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,6 +1,10 @@
 image:
   file: .gitpod.Dockerfile
 
+tasks:
+  - init: echo "Starting init" |
+          make BUILD_TYPE=debug
+
 github:
   prebuilds:
     # enable for the master/default branch (defaults to true)


sdamghan:
@mithro Init tasks were missed in the previous PR. Now the GitPod is in PR checks (end of list). 
Do we have any preference for starting the project up? I only set to compile VTR in debug mode. However, we can download titan, ispd and symbiflow benchmarks as well.
@vaughnbetz I think you might have some thoughts on this as well.

vaughnbetz:
I haven't used gitpod yet, so I don't have a very informed opinion.  A few thoughts though:

From this post, it looks like it has a 30 GB disk limit, 12 GB of RAM and 7 vCPUs (which get throttled if you use them a lot).  https://community.gitpod.io/t/how-much-resources-does-one-workspace-have/1677

A lot of work on the titan and ispd designs is going to be QoR work where you want cpu time measurements and fast runs, so they are less essential to get on gitpod, which won't have the cpu to do very controlled experiments.  But if they'll fit in the 30 GB space limit comfortably (and I think they will) I'd pull down the Titan designs.  I'd hold off on ispd for now, as we haven't integrated them into our tuning much yet.

sdamghan:
@vaughnbetz thanks,
It seems compilation is more practical than other options; I will leave the init phase as it is now only to compile the VTR.
I am done with this PR. It is ready to merge.

vaughnbetz:
Sounds good, thanks.

Issue #1752, opened 2021-06-01
Should print <none> for floorplan constraint file if no file is specified (not blank)


vaughnbetz:
If no floorplan constraint file is specified, vpr still echoes out the floorplan constraints file (blank).  It should instead say something like "none".

#### Expected Behaviour
Print out a line that clearly shows no floorplan constraint file is being used.

#### Current Behaviour
Prints:
Vpr floorplanning constraints file:

#### Possible Solution
Print:
Vpr floorplanning constraints file: not specified

#### Steps to Reproduce
Run vpr and don't specify a floorplan constraint file

#### Your Environment
VTR master on May 31
Ubuntu 16.04

sfkhalid:
Was resolved by pull request #1773

PR #1773, opened 2021-06-10
Print not specified for floorplan constraints file when running vpr w…

sfkhalid:
…ithout floorplan constraints

When a floorplan constraints file is not provided when running vpr (i.e. read_vpr_constraints option not used), the VTR log file used print:
Vpr floorplanning constraints file: 

Now, it prints the following
Vpr floorplanning constraints file: not specified

This change makes it clear to the user that no constraints file was passed in when running vpr. This fix is for issue #1752

### DIFF
--- a/vpr/src/base/ShowSetup.cpp
+++ b/vpr/src/base/ShowSetup.cpp
@@ -30,7 +30,11 @@ void ShowSetup(const t_vpr_setup& vpr_setup) {
     VTR_LOG("Circuit placement file: %s\n", vpr_setup.FileNameOpts.PlaceFile.c_str());
     VTR_LOG("Circuit routing file: %s\n", vpr_setup.FileNameOpts.RouteFile.c_str());
     VTR_LOG("Circuit SDC file: %s\n", vpr_setup.Timing.SDCFile.c_str());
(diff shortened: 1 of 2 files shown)


vpr/src/base/ShowSetup.cpp
```diff
@@ -30,7 +30,11 @@ void ShowSetup(const t_vpr_setup& vpr_setup) {
     VTR_LOG("Circuit placement file: %s\n", vpr_setup.FileNameOpts.PlaceFile.c_str());
     VTR_LOG("Circuit routing file: %s\n", vpr_setup.FileNameOpts.RouteFile.c_str());
     VTR_LOG("Circuit SDC file: %s\n", vpr_setup.Timing.SDCFile.c_str());
-    VTR_LOG("Vpr floorplanning constraints file: %s\n", vpr_setup.FileNameOpts.read_vpr_constraints_file.c_str());
+    if (vpr_setup.FileNameOpts.read_vpr_constraints_file == "") {
+        VTR_LOG("Vpr floorplanning constraints file: not specified %s\n");
```
vaughnbetz, line 34: This line has a bug that is crashing when run through the sanitizer regtests.  You have a format code (%s) with no matching argument, which can cause a seg fault.  Please fix and CI will relaunch.

vaughnbetz:
@sfkhalid : this PR needs to be fixed (bad memory access).

PR #1778, opened 2021-06-16
Improve region intersect function

sfkhalid:
The intersection function belonging to the Region class was modified to also consider cases where one region has no subtile specified and the other region has a subtile specified.

#### Description
Previously, the region intersection function would only calculate the intersection when the two regions passed in both had no subtiles specified, or both had the same subtiles specified. However, this meant that the routine would miss out on some cases. 

For example, if region r1 was constrained from region (0,0) to (10,10) with subtile 0 specified, and region r2 was constrained from region (0,0) to (10,10) with no subtile specified, the previous intersection function would say that there is no intersection between these two regions. The new intersection function would return (0,0) to (10,10) with subtile 0 as the intersection between r1 and r2. 

#### Related Issue

#### Motivation and Context
The missing of some intersecting regions becomes an issue especially when trying to intersect a region with the grid dimensions, to ensure that the region does not fall outside of the grid dimensions. If a PartitionRegion is created to match grid dimensions, it does not have a subtile specified since different areas of the grid can have different subtile dimensions. If a region with subtile specified is intersected with the grid PartitionRegion, the old Region intersection will return no intersection, even when the region is within the grid dimensions. The new Region intersection function solves this problem.

Another thing that should be considered is whether there is a way to check that the subtile dimensions of the region are within the subtile dimensions of the grid at the specified x, y values.

### DIFF
--- a/vpr/src/base/region.cpp
+++ b/vpr/src/base/region.cpp
@@ -103,24 +103,17 @@ Region intersection(const Region& r1, const Region& r2) {
         intersect_rect = intersection(r1_rect, r2_rect);
         intersect.set_region_rect(intersect_rect.xmin(), intersect_rect.ymin(), intersect_rect.xmax(), intersect_rect.ymax());
 
-        return intersect;
-
     } else if (r1.get_sub_tile() == NO_SUBTILE && r2.get_sub_tile() != NO_SUBTILE) {
         intersect.set_sub_tile(r2.get_sub_tile());
         intersect_rect = intersection(r1_rect, r2_rect);
         intersect.set_region_rect(intersect_rect.xmin(), intersect_rect.ymin(), intersect_rect.xmax(), intersect_rect.ymax());
 
-        return intersect;
-
     } else if (r1.get_sub_tile() != NO_SUBTILE && r2.get_sub_tile() == NO_SUBTILE) {
         intersect.set_sub_tile(r1.get_sub_tile());
         intersect_rect = intersection(r1_rect, r2_rect);
         intersect.set_region_rect(intersect_rect.xmin(), intersect_rect.ymin(), intersect_rect.xmax(), intersect_rect.ymax());
-
-        return intersect;
     }
 
-    //If none of the above cases are true, an empty intersect region is returned
     return intersect;
 }
 
--- a/vpr/src/base/region.cpp
+++ b/vpr/src/base/region.cpp
@@ -81,23 +81,46 @@ bool do_regions_intersect(Region r1, Region r2) {
 
 Region intersection(const Region& r1, const Region& r2) {
     Region intersect;
+    vtr::Rect<int> r1_rect = r1.get_region_rect();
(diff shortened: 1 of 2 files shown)


vpr/src/base/region.cpp
```diff
@@ -81,23 +81,46 @@ bool do_regions_intersect(Region r1, Region r2) {
...
      */
     if (r1.get_sub_tile() == r2.get_sub_tile()) {
         intersect.set_sub_tile(r1.get_sub_tile());
-        vtr::Rect<int> r1_rect = r1.get_region_rect();
-        vtr::Rect<int> r2_rect = r2.get_region_rect();
-        vtr::Rect<int> intersect_rect;
+        intersect_rect = intersection(r1_rect, r2_rect);
+        intersect.set_region_rect(intersect_rect.xmin(), intersect_rect.ymin(), intersect_rect.xmax(), intersect_rect.ymax());
 
+        return intersect;
```
sfkhalid, line 106: you can just do one return intersect at the end, it won't make the routine any slower

vpr/src/place/initial_placement.cpp
```diff
@@ -47,9 +47,9 @@ static t_physical_tile_type_ptr pick_placement_type(t_logical_block_type_ptr log
 vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
 
 //Sort the blocks according to how difficult they are to place, prior to initial placement
-std::vector<ClusterBlockId> sort_blocks(vtr::vector<ClusterBlockId, t_block_score> block_scores);
+std::vector<ClusterBlockId> sort_blocks(const vtr::vector<ClusterBlockId, t_block_score>& block_scores);
 
-void print_sorted_blocks(std::vector<ClusterBlockId> sorted_blocks, vtr::vector<ClusterBlockId, t_block_score> block_scores);
+void print_sorted_blocks(const std::vector<ClusterBlockId> sorted_blocks, vtr::vector<ClusterBlockId, t_block_score>& block_scores);
```
sfkhalid, line 52: make both of these const references if you are not changing it (didn't put ampersand for sorted_blocks)

vpr/src/place/initial_placement.cpp
```diff
@@ -400,7 +406,7 @@ std::vector<ClusterBlockId> sort_blocks(vtr::vector<ClusterBlockId, t_block_scor
     return sorted_blocks;
 }
 
-void print_sorted_blocks(std::vector<ClusterBlockId> sorted_blocks, vtr::vector<ClusterBlockId, t_block_score> block_scores) {
+void print_sorted_blocks(const std::vector<ClusterBlockId> sorted_blocks, vtr::vector<ClusterBlockId, t_block_score>& block_scores) {
```
sfkhalid, line 409: change both to const reference

sfkhalid:
@vaughnbetz Does this look good to merge?

vaughnbetz:
Looks good.

Issue #1768, opened 2021-06-07
Add the ability to perform detailed routing with different channel widths


bheiner14:
Hi, I am a BYU CCL student working with Jeff Goeders and have been working on an improvement to the detailed routing so that we can specify different channel widths and distributions in the x and y direction.

#### Proposed Behaviour
This change would allow the user to specify a different channel distribution and peak in the x and y directions for detailed routing.

#### Current Behaviour
Currently detailed routing is only supported on FPGAs with all channels of equal width. If you attempt to use channels of different widths in the x and y direction you receive a fatal error due to the checks in CheckSetup.cpp

#### Possible Solution
Removing the check in CheckSetup.cpp seems to allow detailed routing to be performed on an architecture with different x and y channel widths. However, this causes the regression test width calculator to incorrectly pick an odd numbered width for a unidirectional graph. This is fixed by adding in a conditional statement that will only allow odd widths for bidirectional graphs and will round odd widths for unidirectional graphs to the nearest even number.

#### Context
We are trying to create an approximate capture of Xilinx-style FPGA architectures in the VTR architecture description language, and being able to specify channel widths is essential to creating this approximation.

PR #1786, opened 2021-06-21
Added the ability to perform detailed routing with different channel widths

heinerb1:
#### Description
I modified the init_chan function to prevent odd channel widths from being selected during task and regression test runs. In addition, I modified the binary_search_place_and_route function to pass directionality information to the init_chan function so that bidirectional graphs can still choose odd channel widths. Since I had to remove the VTR fatal error in CheckSetup.cpp to allow detailed routing, I also had to remove a function parameter in the CheckSetup function that was used only in the section that I removed.

#### Related Issue
#1768 

#### Motivation and Context
This change is necessary to create an approximate capture of Xilinx FPGA architectures in the VTR architecture description language, and for any other FPGA architectures that utilize different channel widths in the horizontal and vertical directions. This change allows you to utilize the existing VTR global routing distributions to perform detailed routing on FPGAs with different channel widths in the horizontal and vertical directions. Currently, these distributions can only be used for the global route type.

I ran several individual vpr runs and tasks to test my changes. In addition, I ran the vtr_reg_basic and vtr_reg_strong regression tests. I am using the latest version of VTR and performed a clean build before running these tests. All of the tests ran successfully and passed the QoR test except for the QoR test for strong_global_nonuniform. This failed with the error message "x_gaussian_y_uniform.xml/stereovision3.v/common min_chan_width relative value 1.5 outside of range [0.25,1.3] and not equal to golden value: 12.0" However, I was wondering if it is important to pass this test since it is using the less common global route mode instead of the detailed routing mode? Other than this one issue, the code seems to have no adverse affects on other areas of the code. I would appreciate it if someone could let me know if I should run additional tests or recommend additional modifications to my changes.

#### Types of changes

#### Checklist:


heinerb1:
@vaughnbetz I made all of the changes that you requested and it should now pass all CI checks.

vpr/src/base/place_and_route.cpp
```diff
@@ -452,6 +457,26 @@ t_chan_width init_chan(int cfactor, t_chan_width_dist chan_width_dist) {
     return chan_width;
 }
 
+/**
+ * @brief Computes the channel width and adjusts it to be an an even number if unidirectional 
+ *        since unidirectional graphs need to have paired wires.
+ * 
+ *   @param cfactor                 Channel width factor.
```
vaughnbetz, line 464: Could explain what channel width factor is: 
cfactor --> Channel width factor: multiplier on the channel width distribution (i.e. this is usually the number of tracks in the widest channel).

vaughnbetz (approved):
Looks good; thanks.  I just have one minor suggestion for an enhanced comment.

vpr/src/base/place_and_route.cpp
heinerb1: Sure thing. I just updated it to include your description.

vaughnbetz:
Looks like the kokoro tests weren't running as Brandon wasn't in the vtr org.  Just sent 
@heinerb1 an invite to vtr, and force started kokoro.

PR #1791, opened 2021-06-28
RRGraphView::node_direction() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::node_direction()` and `RRGraphView::node_direction_string()` throughout VTR. Every time `device_ctx.rr_nodes[index].direction()` was called has been replaced with `rr_graph.node_direction(RRNodeId(index))`, in addition to the replacements for `node_direction_string()`. In order to do this, I followed a pattern similar to that in the last PR.

### Motivation and Context

### How Has This Been Tested?

I have run the regression tests for `vtr_reg_strong` along with the QoR testing found at `    $ ../scripts/run_vtr_task.py regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain`. Results from these QoR tests can be found below. The file containing all results can be found [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6729548/node_direction.xlsx)

|--|--|--|
|vtr_flow_elapsed_time	|1	|0.979977348|
|odin_synth_time	|1	|0.930196862|
|abc_synth_time	|1	|0.998563829|
|max_vpr_mem	|1	|0.999830052|
|pack_time	|1	|1.008292914|
|place_time	|1	|0.992582016|
|min_chan_width_route_time	|1	|0.989353431|
|crit_path_route_time	|1	|0.982176651|

#### Types of changes

#### Checklist:


vpr/src/device/rr_graph_view.h
```diff
@@ -62,6 +62,16 @@ class RRGraphView {
         return node_storage_.node_capacity(node);
     }
 
+    /* Get the direction of a routing resource node. This function is inlined for runtime optimization. */
```
vaughnbetz, line 65: This is a good time to beef up the comment.  Explain what the four direction types are, either here or in the vpr_types.h where the enum is defined (or both).  
INC_DIRECTION: wire driver is positioned at the low-coordinate end of the wire.
DEC_DIRECTION: wire_driver is positioned at the high-coordinate end of the wire.
BI_DIRECTION: wire has multiple drivers, so signals can travel either way along the wire
NO_DIRECTION: wire does not have a valid direction set 

I also think it would be good to change NO_DIRECTION to INVALID_DIRECTION to make its meaning more clear.

vaughnbetz (commented):
Looks good. 

I have one request for a comment (not really related to your change, but a good opportunity to comment on what direction means).

I also think we should change to std::string for the node_direction_string, in order to gradually move the code to use more std::string and fewer c-style strings.  I am interested in @tangxifan and @hzeller 's opinions on this though -- do you agree we should move toward C++ strings where possible, or should we just use C-strings in new / changed code like this?

vpr/src/device/rr_graph_view.h
ethanroj23: I have added the comments here and in `vpr_types.h`. I have also changed `NO_DIRECTION` to `INVALID_DIRECTION`. I will wait until Xifan and Henner give their opinions about the C-strings v. C++ strings until I do anything there.

vpr/src/device/rr_graph_view.h
```diff
@@ -62,6 +62,16 @@ class RRGraphView {
...
+    inline e_direction node_direction(RRNodeId node) const {
+        return node_storage_.node_direction(node);
+    }
+
+    /* Get the direction string of a routing resource node. This function is inlined for runtime optimization. */
+    inline const char* node_direction_string(RRNodeId node) const {
```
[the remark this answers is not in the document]
tangxifan: Hi @vaughnbetz @hzeller  I do not have a strong opinion here. Personally, I prefer the C++ string object as it is easier to manipulate (e.g., check empty, check value etc.). For internal storage, we can use the ``const char*`` to save memory footprint.  For data exchange, using the string object is fine because it does not cost too much.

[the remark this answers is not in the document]
vaughnbetz: I prefer having the method over having a constant array of strings that we expose (better abstraction). 

 I also have a mild preference for the C++ string object due to easier manipulation; so far that means Xifan and I both have a mild preference to change this method to return a C++ string.

vpr/src/base/vpr_types.h
```diff
@@ -1344,15 +1344,21 @@ struct t_det_routing_arch {
...
+ * DEC_DIRECTION: wire_driver is positioned at the high-coordinate end of the wire.
+ * BI_DIRECTION: wire has multiple drivers, so signals can travel either way along the wire
+ * INVALID_DIRECTION: wire does not have a valid direction set
+ */
 enum e_direction : unsigned char {
     INC_DIRECTION = 0,
     DEC_DIRECTION = 1,
     BI_DIRECTION = 2,
-    NO_DIRECTION = 3,
+    INVALID_DIRECTION = 3,
```
tangxifan, line 1357: Actually, the ``NUM_DIRECTION`` represents the invalid direction. It was made to be last one in the enumeration and considered as the invalid value.
We can consider to remove this one and use ``NUM_DIRECTIONS`` as the invalid value to return. This can reduce the size this enumeration and follow the same convention as other enumerations.

vpr/src/device/rr_graph_view.h
```diff
@@ -62,6 +62,21 @@ class RRGraphView {
         return node_storage_.node_capacity(node);
     }
 
+    /* Get the direction of a routing resource node. This function is inlined for runtime optimization.
+     * INC_DIRECTION: wire driver is positioned at the low-coordinate end of the wire.
+     * DEC_DIRECTION: wire_driver is positioned at the high-coordinate end of the wire.
+     * BI_DIRECTION: wire has multiple drivers, so signals can travel either way along the wire
+     * INVALID_DIRECTION: wire does not have a valid direction set
+     */
+    inline e_direction node_direction(RRNodeId node) const {
```
tangxifan, line 71: I looked at the needs of the node_direction. 
Most of the time, client functions want to know if this is a bi-directional routing track or not. In some cases, it wants to know the directionality for uni-directional tracks. 
As @vaughnbetz suggested, we should avoid exposing the raw data to developers.
Should we think about creating a query method ``node_is_bidirectional()`` and ``node_is_unidirectional()``? For unidirectional wires, we can return the ``INC_DIR`` and ``DEC_DIR`` data to developers.

I am also happy with the current APIs. These are just suggestions.

tangxifan (commented):
@ethanroj23 Looks good to me. Just a few thoughts on how to further improve the APIs.

vpr/src/device/rr_graph_view.h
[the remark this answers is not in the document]
ethanroj23: I have altered the code to use a C++ std::string. Let me know if you see anything that still needs to be adjusted. Thanks!

vpr/src/base/vpr_types.h
ethanroj23: @vaughnbetz do you have any input on this? The code currently has `INVALID_DIRECTION` as the fourth enumeration. I can change that back to `NO_DIRECTION`, or even remove it entirely. (I think that is what Xifan is meaning in this comment)

vpr/src/device/rr_graph_view.h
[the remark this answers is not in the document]
hzeller: So if you want to do a string return, I'd not do it this way as the compiler will have to copy every time (it does return value optimization, but that only means that it copies it in the string on the call-site).

If these methods are only called every now and then for printing something, this is ok. If the strings are very short (I think they are) it probably is also not too big of a deal (as the std::string does not allocate dynamic memory in that case).

Something to consider: if these strings can be pre-allocated as `std:string something[arraysize]`, you can return `const std::string&` references and don't have to worry about copy.

In which context are the returned strings needed ? Are you doing string manipulation on these that require std::string-ness ? Would you have to call `.c_str()` on every string on reception, because it is used in some fprintf()-like context ?

vpr/src/device/rr_graph_view.h
```diff
@@ -62,6 +62,21 @@ class RRGraphView {
...
+        return node_storage_.node_direction(node);
+    }
+
+    /* Get the direction string of a routing resource node. This function is inlined for runtime optimization. */
+    inline std::string node_direction_string(RRNodeId node) const {
+        return node_storage_.node_direction_string(node);
```
hzeller, line 77: Unless really really needed (e.g. if constructed on-the-fly), I'd try to avoid returning strings by value but rather return them by const reference.

vpr/src/base/vpr_types.h
```diff
@@ -1344,15 +1344,21 @@ struct t_det_routing_arch {
...
     NUM_DIRECTIONS
 };
 
-constexpr std::array<const char*, NUM_DIRECTIONS> DIRECTION_STRING = {{"INC_DIRECTION", "DEC_DIRECTION", "BI_DIRECTION", "NO_DIRECTION"}};
+constexpr std::array<const char*, NUM_DIRECTIONS> DIRECTION_STRING = {{"INC_DIRECTION", "DEC_DIRECTION", "BI_DIRECTION", "INVALID_DIRECTION"}};
```
hzeller, line 1361: Can this hold `std::string` instead of `const char*`, so that you can return a const reference to it and don't have to convert it to a std::string every time ? (I presume this is the place where the string conversions come from ?)

vaughnbetz: Good idea; that will remove any (small) performance overhead.

ethanroj23: @hzeller I am unsure if this is possible. When I try to change the `const char*` to a `std::string`, the compiler gives me the following error: 

`error: the type ‘const std::array<const std::__cxx11::basic_string<char>, 4>’ of ‘constexpr’ variable ‘DIRECTION_STRING’ is not literal
 1361 | constexpr std::array<const std::string, static_cast<int>(e_direction::NUM_DIRECTIONS)> DIRECTION_STRING = {{"INC_DIRECTION", "DEC_DIRECTION", "BI_DIRECTION", "NO_DIRECTION"}};`

Also, this [answer](https://stackoverflow.com/questions/27123306/is-it-possible-to-use-stdstring-in-a-constexpr) on stackoverflow seems to say it is not possible to do this. Perhaps I am missing something, though. Do you have any suggestions here?

vpr/src/base/vpr_types.h
```diff
@@ -1344,15 +1344,21 @@ struct t_det_routing_arch {
     std::string write_rr_graph_filename;
 };
 
+/*
+ * INC_DIRECTION: wire driver is positioned at the low-coordinate end of the wire.
+ * DEC_DIRECTION: wire_driver is positioned at the high-coordinate end of the wire.
+ * BI_DIRECTION: wire has multiple drivers, so signals can travel either way along the wire
+ * INVALID_DIRECTION: wire does not have a valid direction set
+ */
 enum e_direction : unsigned char {
```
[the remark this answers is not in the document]
vaughnbetz: That is a good idea.  Also makes the names a bit more explicit (class name :: constant name).  In that case I think we can shorten the enum constant names to INC, DEC, BIDIR, and NONE.  I'd rename the enum type (class) to Direction and usage would be things like

if (dir == Direction::INC)  {  // and so on.

vpr/src/base/vpr_types.h
vaughnbetz: OK with me to change back to NO_DIRECTION (or now, Direction::NONE).  Should comment how that happens / usage though: I believe it is for nodes where direction is not meaningful, such as input or output pins (not wires).

vpr/src/base/vpr_types.h
vaughnbetz: You're right; from stackoverflow it looks like this isn't doable until C++20.  I wouldn't do it in that case; the performance overhead is minimal and forcing people to use a C++20 standard compiler is probably not a good idea for a few years (we'll get a bunch of build error complaints).

vpr/src/device/rr_graph_view.h
vaughnbetz: Making a node_is_bidirectional() call makes sense to me, as the number one use of these calls seems to be to check if the node is bidirectional.  I don't have a really strong opinion though; if you want to land it as is that's OK with me, but adding a node_is_bidirectional method and using it instead of == BI_DIRECTION may add a small amount of abstraction so I'm also fine with changing to that.

vpr/src/route/rr_graph_storage.cpp
```diff
@@ -549,19 +549,19 @@ const char* t_rr_graph_view::node_type_string(RRNodeId id) const {
...
+const std::string& t_rr_graph_storage::node_direction_string(RRNodeId id) const {
+    Direction direction = node_direction(id);
 
-    if (direction == e_direction::INC_DIRECTION) {
-        return "INC_DIR";
-    } else if (direction == e_direction::DEC_DIRECTION) {
-        return "DEC_DIR";
-    } else if (direction == e_direction::BI_DIRECTION) {
-        return "BI_DIR";
+    if (direction == Direction::INC) {
```
vaughnbetz, line 555: I'd rewrite this to get rid of the if's.  The usual way to do this is something like:

int_direction = static_cast<int>(direction);
VTR_ASSERT (int_direction >= 0 && int_direction < Direction::NUM_DIRECTIONS);
return CONST_DIRECTION_STRING[int_direction];

Shorter and scales better if we add more enum values later.

vaughnbetz (commented):
Looks good; just have one more request for a change. (make the node_direction_string method more succinct)

vpr/src/route/rr_graph_storage.cpp
ethanroj23: Yeah, that makes sense. I'll adjust that real quick.

vaughnbetz: Just noticed by syntax is bad:  meant 
int_direction = static_cast<int>(direction); 

and so on :).

vpr/src/base/vpr_types.h
hzeller: Does this array have to be public in the header ? This looks like the original problem exposing such thing in the header.
Usually, I'd just make this a `static std::string foo[] = {"foo", "bar", "baz"};` somewhere in the compilation unit and provide a function to convert the int to the string.

Alternatively as quick hack, the to string function could be to put it in the header as inline function:
```
enum class Dir { FOO, BAR, BAZ };
inline const std::string& direction_string(Dir x) {
  static std::string foo[] = {"foo", "bar", "baz"};
  return foo[(int)x];
}
```

vpr/src/route/rr_graph_storage.cpp
hzeller: switch/case ?

vpr/src/base/vpr_types.h
vaughnbetz: I don't think there's any reason it has to be in the header; it would be better to keep it in the .cpp file as you suggest.

tangxifan:

Issue #1769, opened 2021-06-08
Add option to have wire segments run only vertically and/or horizontally


brycejh:
Hello, I'm a BYU student working with Jeff Goeders, working on routing specifically. Currently any added segments listed in the architecture occur in both horizontal and vertical routing channels. Newer architectures often have certain segments that only occur in one of them, or occur in different frequencies depending upon position.

#### Proposed Behaviour
This would allow a user to specify if a certain segment runs horizontally, vertically, or both. Additionally, with the currently implemented 'frequency' option one could adjust the occurrence in each direction independently.

#### Current Behaviour
Current behaviour allows for segments of different lengths at different frequencies within a given channel, this frequency is constant throughout the board in all directions.

#### Possible Solution
Add an optional parameter in the .xml file, default would be both horizontal and vertical, that allows the user to specify what direction they wish the segments to run. This is then stored as a new type/value in t_segment_inf and used when building the segments. As I understand, the X and Y channels are built, at the same time, separately and could be introduced during this stage. This would not influence channel width as an error would be thrown if the frequency in either x or y does not add up to 100%

#### Context
In the BYU CCL lab we are attempting to integrate and expand the capabilities of VTR to allow for modeling of xillinx 7 series boards, newer generations of both xillinx and alterra/Intel boards appear to have more of these types of heterogeneous routing patterns and adding this feature would lead to better modeling of these newer boards.

PR #1794, opened 2021-07-07
Enable wire segments to only be one direction; X,Y or both

brycejh:
In this draft I added to the type that is made from reading the XML. Also changed the XML reader to accept the new parameter.
#### Description

Made a new enumeration type to include X wires, Y wires, and wires that go in both directions. Included this in the t_segment_inf data type to store the data in. 
Added code to read in this parameter from an XML document. It is processed at the same time as the other aspects of a given wire segment (just after frequency). Parameter is optional and is default to both X and Y if not specified. Throws an error if an incorrect option is entered

#### Related Issue

#1769
#### Motivation and Context

This change will eventually allow make different wires of different length in only X direction, only Y direction or both. Newer FPGA architectures use this in their design. Additionally this will allow for different frequencies in the different directions.

Tested locally by running basic and strong tests. Zero fails.
#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1382,6 +1382,11 @@ enum e_directionality {
     UNI_DIRECTIONAL,
     BI_DIRECTIONAL
 };
+enum e_segment_adjacency {
+    COL_ROW_SEGMENT,
+    COLLUMN_SEGMENT,
+    ROW_SEGMENT
+};
 enum e_switch_block_type {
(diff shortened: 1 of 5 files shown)


ArashAhmadian:
#1836 is a very similar PR. The main different being #1836 is more backwards compatible.

brycejh:
> #1836 is a very similar PR. The main different being #1836 is more backwards compatible.

Agreed, while similar I prefer the look in #1836.

PR #1799, opened 2021-07-12
RRGraphView::node_fan_in() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::node_fan_in()` throughout VTR. Every time `device_ctx.rr_nodes[index].fan_in()` was called has been replaced with `rr_graph.node_direction(RRNodeId(index))`. In order to do this, I followed a pattern similar to that in the last PR.

### Motivation and Context

### How Has This Been Tested?

I have run the regression tests for `vtr_reg_strong` along with the QoR testing found at `    $ ../scripts/run_vtr_task.py regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain`. Results from these QoR tests can be found below. The file containing all results can be found [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6810934/node_fan_in.xlsx)

|---|---|---|
|vtr_flow_elapsed_time	|1|	0.998931189|
|odin_synth_time	|1|	0.989230519|
|abc_synth_time	|1|	0.994648961|
|max_vpr_mem	|1|	0.99915845|
|pack_time	|1|	0.997989839|
|place_time	|1|	0.995451755|
|min_chan_width_route_time	|1|1.001024542|
|crit_path_route_time	|1|	0.999048907|

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/power/power.cpp
+++ b/vpr/src/power/power.cpp
@@ -831,17 +831,20 @@ static void power_usage_routing(t_power_usage* power_usage,
                         case CHANX:
                         case CHANY:
                         case IPIN:
-                            if (next_node_power->net_num == node_power->net_num) {
-                                next_node_power->selected_input = next_node_power->num_inputs;
-                            }
-                            next_node_power->in_dens[next_node_power->num_inputs] = clb_net_density(node_power->net_num);
-                            next_node_power->in_prob[next_node_power->num_inputs] = clb_net_prob(node_power->net_num);
-                            next_node_power->num_inputs++;
-                            if (next_node_power->num_inputs > next_node.fan_in()) {
-                                VTR_LOG("%d %d\n", next_node_power->num_inputs,
-                                        next_node.fan_in());
-                                fflush(nullptr);
-                                VTR_ASSERT(0);
+                            {
+                                if (next_node_power->net_num == node_power->net_num) {
+                                    next_node_power->selected_input = next_node_power->num_inputs;
+                                }
+                                next_node_power->in_dens[next_node_power->num_inputs] = clb_net_density(node_power->net_num);
+                                next_node_power->in_prob[next_node_power->num_inputs] = clb_net_prob(node_power->net_num);
+                                next_node_power->num_inputs++;
+                                const t_edge_size next_node_fan_in = rr_graph.node_fan_in(RRNodeId(next_node_id));
+                                if (next_node_power->num_inputs > next_node_fan_in) {
+                                    VTR_LOG("%d %d\n", next_node_power->num_inputs,
+                                            next_node_fan_in);
+                                    fflush(nullptr);
+                                    VTR_ASSERT(0);
+                                }
                             }
                             break;
                         default:
@@ -865,6 +868,7 @@ static void power_usage_routing(t_power_usage* power_usage,
         int switchbox_fanout;
         //float C_per_seg_split;
         int wire_length;
+        const t_edge_size node_fan_in = rr_graph.node_fan_in(RRNodeId(rr_node_idx));
 
         switch (rr_graph.node_type(RRNodeId(rr_node_idx))) {
             case SOURCE:
@@ -877,13 +881,13 @@ static void power_usage_routing(t_power_usage* power_usage,
                  *  - Driver (accounted for at end of CHANX/Y - see below)
                  *  - Multiplexor */
 
-                if (node.fan_in()) {
+                if (node_fan_in) {
                     VTR_ASSERT(node_power->in_dens);
                     VTR_ASSERT(node_power->in_prob);
 
                     /* Multiplexor */
                     power_usage_mux_multilevel(&sub_power_usage,
-                                               power_get_mux_arch(node.fan_in(),
+                                               power_get_mux_arch(node_fan_in,
                                                                   power_ctx.arch->mux_transistor_size),
                                                node_power->in_prob, node_power->in_dens,
                                                node_power->selected_input, true,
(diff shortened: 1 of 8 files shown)


vpr/src/route/rr_graph.cpp
```diff
@@ -507,6 +507,8 @@ static void build_rr_graph(const t_graph_type graph_type,
     /* get maximum number of pins across all blocks */
     int max_pins = types[0].num_pins;
     for (const auto& type : types) {
+        VTR_LOG("%s\n", type.name);
```
tangxifan, line 510: It seems that this log message was used for debugging. Should remove it once finished debugging.

vpr/src/power/power.cpp
```diff
@@ -1209,16 +1210,17 @@ void power_routing_init(const t_det_routing_arch* routing_arch) {
         int fanout_to_seg = 0;
         auto node = device_ctx.rr_nodes[rr_node_idx];
         t_rr_node_power* node_power = &rr_node_power[rr_node_idx];
+        const t_edge_size node_fan_in = rr_graph.node_fan_in(RRNodeId(rr_node_idx));
```
tangxifan, line 1213: we can use the ``t_edge_size``  when declaring the ``max_fanin``, ``max_IPIN_fanin``, ``max_seg_to_seg_fanout`` and ``max_seg_to_IPIN_fanout``, rather than the currently used ``int``. 
This can save the use of ``static_cast`` in downstream codes.

tangxifan (changes requested):
@ethanroj23 The code modification is mostly clean. Just some minor comments.

vpr/src/power/power.cpp
hzeller: +1

vpr/src/power/power.cpp
```diff
@@ -1361,14 +1363,13 @@ bool power_uninit() {
     bool error = false;
 
     for (size_t rr_node_idx = 0; rr_node_idx < device_ctx.rr_nodes.size(); rr_node_idx++) {
```
hzeller, line 1365: Probably something for later, but if we could
```
for (RRNodeId id : device_ctx.rr_nodes) {
}
```

It would be much more expressive what we mean, and also we wouldn't need the typecast that we have to do now when passing it to `rr_graph.node_fan_in(....)`.

(this is somewhat the idea of working towards a goal that makes the code look like one wishes it looked like, so then towards `device_ctx.rr_nodes` being something that iterates out`RRNodeId`.

hzeller (approved):
LGTM, and ready to merge after @tangxifan's comments are addressed.

vpr/src/power/power.cpp
vaughnbetz: Agreed.  I think that change would work, and would be more expressive (and avoiding casts is good).

vaughnbetz (changes requested):

tangxifan:
@vaughnbetz Just rerun the CI on this PR to find out why the S:Basic test failed. This PR does not contain my latest refactoring effort. But it still failed. I plan to create another branch and revert to the status where no refactoring is done. I will run the Ci on that branch and see what happened. If the test remain failed, I suspect it is a problem of Github CI runners.

vaughnbetz:
Thanks @tangxifan .  Another good thing to do (in retrospect should have done this during the refactoring): run a QoR test with all your refactoring to date vs. a version with no refactoring.  Key things to check:  most things (crit path delay, wirelength) are the same, how much did runtime change, how much did peak memory change.  That is a good check that we didn't degrade on any important metrics (particularly peak memory footprint and runtime).

tangxifan:
> Thanks @tangxifan . Another good thing to do (in retrospect should have done this during the refactoring): run a QoR test with all your refactoring to date vs. a version with no refactoring. Key things to check: most things (crit path delay, wirelength) are the same, how much did runtime change, how much did peak memory change. That is a good check that we didn't degrade on any important metrics (particularly peak memory footprint and runtime).

Yes. I will post QoR results on each of my PR now. Should I provide these data with and without sanity checks?
Even though, the sanity check failures look a bit weird to me now, after some investigation
- I created a branch https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/test_sanity_basic , which is reverted to a commit before refactoring happened. I see sanity check passed. So, I think the refactoring is indeed increasing peak memory usage in sanity checks.
- I see the sanity checks passed in past when we merge the PRs but it starts to turn red this Monday when rerun these tests. So, it seems to me that Github runner has changed memory limits.
- I am trying the reproduce the memory issue on my local machine when compiled with the same flags as sanity tests and run basic regression tests. But memory usage looks normal on my side. 

So I am a bit lost on what exactly causes the issue. It looks not so simple. If you have any insights about why such memory leak happens only in sanity checks, I can spend more effort to spot the issue.
Thanks!

vaughnbetz:
The sanitized basic tests are being killed with bash code 137, which is out of memory. The logs don't seem to show exactly where (I suspect some info in the log was cached and not yet flushed to disk when it died) since the logs generally show the latter parts of placement. I suspect instead that they die at the start of routing.

The sanitizers add overhead to every memory allocation, so if we allocate more small memory items, or fragment memory more so there is more unused data on the free list, it would show up more in the sanitized builds than the regular ones.

There were also some commits cleaning up / adjusting how various CI runners work so it's possible the limits were changed at the same time too.

We're also failing a lot of Odin II basic regtests with SIGABRT (could be out of memory, could be assert, could be memory corruption, ...).
I would have expected the Odin II basic tests to have low memory usage, so not sure what's making them fail but it will be unrelated to Xifan's changes. @sdamghan : are you seeing these Odin II basic regtest failures on other CI runs?  Can you look into them, as it looks like a separate issue?

@tansell : were there any recent changes to github action runner memory limits or set up (maybe unintentional as part of a clean up)?

@tangxifan : I suggest collecting 
(1) peak memory footprint data and runtime before and after the entire refactoring for the VTR and titan benchmarks.   That will show if we've increased resident set size or runtime in general.
(2) I'd also collect peak memory footprint before and after the change on a small set of circuits (could be just a couple) with the sanitizers on.  That will show if there is some bad behaviour that happens only with the sanitizers.
Depending on what is found, there may be more tuning of the memory allocation, or it may clearly show this is a github action set up issue.

vaughnbetz:
Note that the 3rd failure (yosys+odin) is expected as the code to support that is still in pr review.

ethanroj23:
> Note that the 3rd failure (yosys+odin) is expected as the code to support that is still in pr review.

@vaughnbetz I have run the checks again, and `Yosys+ODIN-II Strong Tests (Presubmit)` was the only failing test. (Which I believe is expected) If there are no other issues, this PR should be ready to merge.

PR #1800, opened 2021-07-14
Deploy ``RRGraphBuilder`` in RRGraph Reader and Writer to replace the use of ``rr_node_indices``

tangxifan:
#### Description

This PR focuses on updating RRGraph reader and writer functions, where we use the refactored data structure ``RRGraphBuilder`` to replace the legacy data structure ``rr_node_indices``.
This PR aims to eliminate the use of ``rr_node_indices`` in the RRGraph reader and writer, as one step further in deprecating the legacy data structure.

Checklist:

- [x] Eliminate the use of ``rr_node_indices`` for RRGraph reader
- [x] Eliminate the use of ``rr_node_indices`` for RRGraph writer

#### Motivation and Context

This pull request is a follow-up PR on the routing resource graph refactoring effort https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1779

After the previous PR #1779  , we start reworking all the source files that use the legacy data structure ``rr_node_indices`` in a high priority, in order to deprecate the legacy data structure as soon as possible.
Current statistics on the files that use ``rr_node_indices`` (in total there are 143 lines related):
- ./route/clock_connection_builders.cpp
- ./route/clock_network_builders.cpp
- ./route/clock_network_builders.h
- ./route/router_lookahead_map_utils.cpp
- ./route/rr_graph.cpp
- ./route/rr_graph2.cpp
- ./route/rr_graph2.h
- ./route/rr_graph_clock.cpp
- ./route/rr_graph_clock.h
- ./route/rr_graph_reader.cpp
- ./route/rr_graph_util.cpp
- ./route/rr_graph_uxsdcxx_serializer.h
- ./route/rr_graph_writer.cpp

This PR will remove the use in 
- [X] ./route/rr_graph_writer.cpp
- [X] ./route/rr_graph_uxsdcxx_serializer.h
- [X] ./route/rr_graph_reader.cpp

Considering this is all about refactoring, this PR does not need to create new tests. Expect all the existing tests to pass.

#### Types of changes

#### Checklist:


vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1567,148 +1567,50 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
-                }
-            }
-        }
-
-        int count;
-        /* CHANX and CHANY need to reevaluated with its ptc num as the correct index*/
-        for (size_t inode = 0; inode < rr_nodes_->size(); inode++) {
-            auto node = (*rr_nodes_)[inode];
-            if (rr_graph.node_type(node.id()) == CHANX) {
+            for (int ix = node.xlow(); ix <= node.xhigh(); ix++) {
```
vaughnbetz, line 1584: Could all this be made a utility function in RRGraphBuilder (or a global utility function) and called in multiple places?  This code would handle the 3 call sites with 2D loop nests in the clock rr-graph builder too.  It seems like making this a general utility would be a good idea.  Would be a function like: add_node_to_all_locs () or some such.

vaughnbetz (commented):
Generally looks fine, except it makes the case even more that we should have a utility function for adding a node to all the locations in the spatial lookup it touches, in the proper way for each type.

vaughnbetz:
Note: single CI failure is unrelated/spurious.

vpr/src/route/rr_graph_uxsdcxx_serializer.h
tangxifan: Agree. The node addition is very generic and applicable to all kinds of nodes. I think it should be an API of the RRGraphBuilder because it should be the only way to add an existing ``rr_node`` to RRSpatialLookup data structure.

vpr/src/device/rr_graph_builder.h
```diff
@@ -37,6 +37,15 @@ class RRGraphBuilder {
     t_rr_graph_storage& node_storage();
     /* Return a writable object for update the fast look-up of rr_node */
     RRSpatialLookup& node_lookup();
+    /* Add an existing rr_node in the node storage to the node look-up
```
vaughnbetz, line 40: I think this comment should also say that the node will be added to the lookup for every side it is on (for OPINs and IPINs) and for every (x,y) location at which it exists (for wires that span more than one (x,y).

vaughnbetz:
The changes all look good; just suggesting one more comment addition.  Merging this, but if you can add the comment in another PR that would be great.

tangxifan:
> The changes all look good; just suggesting one more comment addition. Merging this, but if you can add the comment in another PR that would be great.

Thanks @vaughnbetz I start working on the next PR. The comment will be first one to be applied in that PR.

PR #1801, opened 2021-07-14
Deploy ``RRGraphBuilder`` in RRGraph Clock Builder to replace the use of ``rr_node_indices``

tangxifan:
#### Description

This PR focuses on updating RRGraph clock builder functions, where we use the refactored data structure ``RRGraphBuilder`` to replace the legacy data structure ``rr_node_indices``.
This PR aims to eliminate the use of ``rr_node_indices`` in the RRGraph clock builder, as one step further in deprecating the legacy data structure.

Checklist:

- [x] Added a new API ``find_sink_nodes()`` to ``RRSpatialLookup``
- [x] Eliminate the use of ``rr_node_indices`` for RRGraph Clock Builder
- [x] Delete obsoleted functions in ``rr_graph2.cpp`` which are replaced by built-in APIs in ``RRSpatialLookup``

#### Motivation and Context

This pull request is a follow-up PR on the routing resource graph refactoring effort https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1779

After the previous PR #1779  , we start reworking all the source files that use the legacy data structure ``rr_node_indices`` in a high priority, in order to deprecate the legacy data structure as soon as possible.
Current statistics on the files that use ``rr_node_indices`` (in total there are 143 lines related):
- ./route/clock_connection_builders.cpp
- ./route/clock_network_builders.cpp
- ./route/clock_network_builders.h
- ./route/router_lookahead_map_utils.cpp
- ./route/rr_graph.cpp
- ./route/rr_graph2.cpp
- ./route/rr_graph2.h
- ./route/rr_graph_clock.cpp
- ./route/rr_graph_clock.h
- ./route/rr_graph_reader.cpp
- ./route/rr_graph_util.cpp
- ./route/rr_graph_uxsdcxx_serializer.h
- ./route/rr_graph_writer.cpp

This PR will remove the use in 
- [X] ./route/clock_connection_builders.cpp
- [X] ./route/clock_network_builders.cpp
- [X] ./route/clock_network_builders.h
- [X] ./route/rr_graph_clock.cpp
- [X] ./route/rr_graph_clock.h

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/device/rr_spatial_lookup.cpp
+++ b/vpr/src/device/rr_spatial_lookup.cpp
@@ -74,17 +74,18 @@ RRNodeId RRSpatialLookup::find_node(int x,
     return RRNodeId(rr_node_indices_[type][node_x][node_y][node_side][ptc]);
 }
 
-std::vector<RRNodeId> RRSpatialLookup::find_channel_nodes(int x,
-                                                          int y,
-                                                          t_rr_type type) const {
+std::vector<RRNodeId> RRSpatialLookup::find_nodes(int x,
+                                                  int y,
+                                                  t_rr_type type,
+                                                  e_side side) const {
     /* TODO: The implementation of this API should be worked 
      * when rr_node_indices adapts RRNodeId natively!
      */
-    std::vector<RRNodeId> channel_nodes;
+    std::vector<RRNodeId> nodes;
 
     /* Pre-check: the x, y, type are valid! Otherwise, return an empty vector */
-    if ((x < 0 || y < 0) && (type == CHANX || type == CHANY)) {
-        return channel_nodes;
+    if (x < 0 || y < 0) {
+        return nodes;
     }
 
     /* Currently need to swap x and y for CHANX because of chan, seg convention 
@@ -106,76 +107,44 @@ std::vector<RRNodeId> RRSpatialLookup::find_channel_nodes(int x,
      * - Return an empty list if any out-of-range is detected
      */
     if (size_t(type) >= rr_node_indices_.size()) {
-        return channel_nodes;
+        return nodes;
     }
 
     if (node_x >= rr_node_indices_[type].dim_size(0)) {
-        return channel_nodes;
+        return nodes;
     }
 
     if (node_y >= rr_node_indices_[type].dim_size(1)) {
-        return channel_nodes;
+        return nodes;
     }
 
-    /* By default, we always added the channel nodes to the TOP side (to save memory) */
-    e_side node_side = TOP;
-    if (node_side >= rr_node_indices_[type].dim_size(2)) {
-        return channel_nodes;
+    if (side >= rr_node_indices_[type].dim_size(2)) {
+        return nodes;
     }
 
-    for (const auto& node : rr_node_indices_[type][node_x][node_y][node_side]) {
+    for (const auto& node : rr_node_indices_[type][node_x][node_y][side]) {
         if (RRNodeId(node)) {
-            channel_nodes.push_back(RRNodeId(node));
+            nodes.push_back(RRNodeId(node));
         }
     }
 
-    return channel_nodes;
+    return nodes;
+}
+
+std::vector<RRNodeId> RRSpatialLookup::find_channel_nodes(int x,
+                                                          int y,
+                                                          t_rr_type type) const {
+    /* Pre-check: node type should be routing tracks! */
+    if (type != CHANX && type != CHANY) {
+        return std::vector<RRNodeId>();
+    }
+
+    return find_nodes(x, y, type);
 }
 
 std::vector<RRNodeId> RRSpatialLookup::find_sink_nodes(int x,
                                                        int y) const {
-    /* TODO: The implementation of this API should be worked 
-     * when rr_node_indices adapts RRNodeId natively!
-     */
-    std::vector<RRNodeId> sink_nodes;
-
-    /* Pre-check: the x, y, type are valid! Otherwise, return an empty vector */
-    if (x < 0 || y < 0) {
-        return sink_nodes;
-    }
-
-    VTR_ASSERT_SAFE(3 == rr_node_indices_[SINK].ndims());
-
-    /* Sanity check to ensure the x, y, side are in range 
-     * - Return a list of valid ids by searching in look-up when all the parameters are in range
-     * - Return an empty list if any out-of-range is detected
-     */
-    if (size_t(SINK) >= rr_node_indices_.size()) {
-        return sink_nodes;
-    }
-
-    if (size_t(x) >= rr_node_indices_[SINK].dim_size(0)) {
-        return sink_nodes;
-    }
-
-    if (size_t(y) >= rr_node_indices_[SINK].dim_size(1)) {
-        return sink_nodes;
-    }
-
-    /* By default, we always added the sink nodes to the TOP side (to save memory) */
-    e_side node_side = TOP;
-    if (node_side >= rr_node_indices_[SINK].dim_size(2)) {
-        return sink_nodes;
-    }
-
-    /* Only insert valid ids in the returned vector */
-    for (const auto& node : rr_node_indices_[SINK][x][y][node_side]) {
-        if (RRNodeId(node)) {
-            sink_nodes.push_back(RRNodeId(node));
-        }
-    }
-
-    return sink_nodes;
+    return find_nodes(x, y, SINK);
 }
 
 std::vector<RRNodeId> RRSpatialLookup::find_nodes_at_all_sides(int x,
--- a/vpr/src/device/rr_spatial_lookup.cpp
+++ b/vpr/src/device/rr_spatial_lookup.cpp
@@ -162,12 +162,13 @@ std::vector<RRNodeId> RRSpatialLookup::find_sink_nodes(int x,
         return sink_nodes;
     }
 
-    /* By default, we always added the channel nodes to the TOP side (to save memory) */
+    /* By default, we always added the sink nodes to the TOP side (to save memory) */
     e_side node_side = TOP;
     if (node_side >= rr_node_indices_[SINK].dim_size(2)) {
         return sink_nodes;
     }
 
+    /* Only insert valid ids in the returned vector */
     for (const auto& node : rr_node_indices_[SINK][x][y][node_side]) {
         if (RRNodeId(node)) {
             sink_nodes.push_back(RRNodeId(node));
@@ -211,6 +212,11 @@ void RRSpatialLookup::add_node(RRNodeId node,
                                e_side side) {
     VTR_ASSERT(node); /* Must have a valid node id to be added */
     VTR_ASSERT_SAFE(3 == rr_node_indices_[type].ndims());
+
+    /* For non-IPIN/OPIN nodes, the side should always be the TOP side which follows the convention in find_node() API! */
+    if (type != IPIN && type != OPIN) {
+        VTR_ASSERT(side == SIDES[0]);
+    }
 
     resize_nodes(x, y, type, side);
 
--- a/vpr/src/device/rr_spatial_lookup.cpp
+++ b/vpr/src/device/rr_spatial_lookup.cpp
@@ -132,6 +132,51 @@ std::vector<RRNodeId> RRSpatialLookup::find_channel_nodes(int x,
     return channel_nodes;
 }
 
+std::vector<RRNodeId> RRSpatialLookup::find_sink_nodes(int x,
+                                                       int y) const {
+    /* TODO: The implementation of this API should be worked 
+     * when rr_node_indices adapts RRNodeId natively!
+     */
+    std::vector<RRNodeId> sink_nodes;
+
+    /* Pre-check: the x, y, type are valid! Otherwise, return an empty vector */
+    if (x < 0 || y < 0) {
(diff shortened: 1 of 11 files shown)


tangxifan:
@mithro Not sure why the Yosys+ODINII strong tests always failed in a few second. The artifact log file is empty.

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -132,6 +132,51 @@ std::vector<RRNodeId> RRSpatialLookup::find_channel_nodes(int x,
...
+
+    if (size_t(x) >= rr_node_indices_[SINK].dim_size(0)) {
+        return sink_nodes;
+    }
+
+    if (size_t(y) >= rr_node_indices_[SINK].dim_size(1)) {
+        return sink_nodes;
+    }
+
+    /* By default, we always added the channel nodes to the TOP side (to save memory) */
```
vaughnbetz, line 165: This comment looks wrong (talking about channels).  

This code also points out why we should rework the underlying rr_node_indices implementation; it's strange to have a multi-dimensional array matrix where some dimensions must always be accessed with a specific index.  Hopefully that is on your near-term agenda Xifan. Could put a TODO in to refactor that way, if you think such a comment would be a useful reminder.

vpr/src/device/rr_spatial_lookup.cpp
```diff
@@ -132,6 +132,51 @@ std::vector<RRNodeId> RRSpatialLookup::find_channel_nodes(int x,
...
+    e_side node_side = TOP;
+    if (node_side >= rr_node_indices_[SINK].dim_size(2)) {
+        return sink_nodes;
+    }
+
+    for (const auto& node : rr_node_indices_[SINK][x][y][node_side]) {
+        if (RRNodeId(node)) {
```
vaughnbetz, line 172: Could comment this to say you're inserting only the valid ids in the returned vector.

vpr/src/device/rr_spatial_lookup.h
```diff
@@ -81,6 +81,18 @@ class RRSpatialLookup {
...
                                              t_rr_type type) const;
 
+    /**
+     * Returns the indices of the specified routing resource nodes,
+     * representing virtual sinks.  
+     * - (x, y) are the coordinate of the sink nodes within the FPGA
+     *
+     * Note: 
+     * - Return an empty list if there are no sinks at the given (x, y) location
+     * - The node list returned only contain valid ids
```
vaughnbetz, line 91: typo: contain -> contains

vpr/src/route/clock_connection_builders.cpp
```diff
@@ -76,13 +74,13 @@ void RoutingToClockConnection::create_switches(const ClockRRGraphBuilder& clock_
         // Connect to x-channel wires
         unsigned num_wires_x = x_wire_indices.size() * fc;
         for (size_t i = 0; i < num_wires_x; i++) {
-            clock_graph.add_edge(rr_edges_to_create, x_wire_indices[i], clock_index, arch_switch_idx);
+            clock_graph.add_edge(rr_edges_to_create, size_t(x_wire_indices[i]), clock_index, arch_switch_idx);
```
vaughnbetz, line 77: Do we need the size_t cast?  Can we change the clock_graph.add_edge function to take an RRNodeId and get rid of the cast?

vpr/src/route/clock_connection_builders.cpp
```diff
@@ -95,35 +93,38 @@ int RoutingToClockConnection::create_virtual_clock_network_sink_node(
...
+    // Use a generic way when adding nodes to lookup.
+    // However, since the SINK node has the same xhigh/xlow as well as yhigh/ylow, we can probably use a shortcut
+    for (int ix = rr_graph.node_xlow(node_index); ix <= rr_graph.node_xhigh(node_index); ++ix) {
+        for (int iy = rr_graph.node_ylow(node_index); iy <= rr_graph.node_yhigh(node_index); ++iy) {
+            node_lookup.add_node(node_index, ix, iy, rr_graph.node_type(node_index), rr_graph.node_ptc_num(node_index), SIDES[0]);
+        }
+    }
 
-    return node_index;
+    return size_t(node_index);
```
vaughnbetz, line 127: Can we change the return type of this function to RRNodeId and remove the cast?  (May have to change the calling function too some then).

vpr/src/route/clock_connection_builders.cpp
```diff
@@ -95,35 +93,38 @@ int RoutingToClockConnection::create_virtual_clock_network_sink_node(
...
     float C = 0.;
-    rr_nodes[node_index].set_rc_index(find_create_rr_rc_data(R, C));
+    rr_graph.set_node_rc_index(node_index, find_create_rr_rc_data(R, C));
 
-    add_to_rr_node_indices(device_ctx.rr_node_indices, rr_nodes, node_index);
+    // Use a generic way when adding nodes to lookup.
+    // However, since the SINK node has the same xhigh/xlow as well as yhigh/ylow, we can probably use a shortcut
+    for (int ix = rr_graph.node_xlow(node_index); ix <= rr_graph.node_xhigh(node_index); ++ix) {
+        for (int iy = rr_graph.node_ylow(node_index); iy <= rr_graph.node_yhigh(node_index); ++iy) {
+            node_lookup.add_node(node_index, ix, iy, rr_graph.node_type(node_index), rr_graph.node_ptc_num(node_index), SIDES[0]);
```
vaughnbetz, line 123: Looks like we use SIDES[0] to insert the sink node, and look for it later on the TOP side.  Hence should we use TOP consistently or SIDES[0] consistently to make sure this is always in alignment?  I assume SIDES[0] == TOP so this all works, but it still seems dangerous as if anyone ever edits the order of SIDES we'd subtly break code written this way.

vpr/src/route/clock_connection_builders.cpp
```diff
@@ -320,7 +319,7 @@ void ClockToPinsConnection::create_switches(const ClockRRGraphBuilder& clock_gra
 
                     //Create edges depending on Fc
                     for (size_t i = 0; i < clock_network_indices.size() * fc; i++) {
-                        clock_graph.add_edge(rr_edges_to_create, clock_network_indices[i], clock_pin_node_idx, arch_switch_idx);
+                        clock_graph.add_edge(rr_edges_to_create, clock_network_indices[i], size_t(clock_pin_node_idx), arch_switch_idx);
```
vaughnbetz, line 322: Can we change the function prototype to avoid the cast (send in an RRNodeId instead)?

vpr/src/route/clock_network_builders.cpp
```diff
@@ -343,7 +343,15 @@ int ClockRib::create_chanx_wire(int x_start,
...
+    auto& rr_graph = (*rr_nodes);
+    RRNodeId chanx_node = RRNodeId(node_index);
+    for (int ix = rr_graph.node_xlow(chanx_node); ix <= rr_graph.node_xhigh(chanx_node); ++ix) {
+        for (int iy = rr_graph.node_ylow(chanx_node); iy <= rr_graph.node_yhigh(chanx_node); ++iy) {
+            //TODO: CHANX uses odd swapped x/y indices here. Will rework once rr_node_indices is shadowed
+            rr_graph_builder.node_lookup().add_node(chanx_node, iy, ix, rr_graph.node_type(chanx_node), rr_graph.node_ptc_num(chanx_node), SIDES[0]);
+        }
+    }
 
     return node_index;
```
vaughnbetz, line 356: Consider changing return type to RRNodeId, and directly making node_index an RRNodeId (and therefore wouldn't need chanx_node).

vpr/src/route/clock_network_builders.cpp
```diff
@@ -640,7 +648,14 @@ int ClockSpine::create_chany_wire(int y_start,
     }
     node.set_cost_index(CHANX_COST_INDEX_START + num_segments + seg_index);
 
-    add_to_rr_node_indices(*rr_node_indices, *rr_nodes, node_index);
+    /* Add the node to spatial lookup */
```
vaughnbetz, line 651: This code has been repeated almost exactly (minus the weird CHANX/CHANY x/y index swapping) 3 times.  Should be a helper function.  Maybe not so easy to make a helper due to the swapping of x/y, but should have a TODO to make a helper in that case.  Or (better) make the .add_node function do the x/y swapping based on the node type (CHANX swaps I think) so all 3 function calls look the same and you can make a helper function. Plus the weird swapping is then localized in the node_lookup class.

vaughnbetz (changes requested):
Thanks @tangxifan .  Looks generally good, but I have some suggested changes / comments (none too major).

vaughnbetz:
Note: the Yosys+Odin-II failure is an expected, unrelated failure and is not gating.

vpr/src/device/rr_spatial_lookup.cpp
tangxifan: Agree. Added the comment.

vpr/src/device/rr_spatial_lookup.h
tangxifan: Thanks for pointing it out. Fixed.

vpr/src/route/clock_connection_builders.cpp
tangxifan: @vaughnbetz 

Strongly Agree. I did not do it because of the ``clock_graph.add_edge()`` function relies on the data structure ``t_rr_edge_info``

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/c06800fb49d2732a0fcedbb2f542613fedd6a330/vpr/src/route/rr_edge.h#L4-L14

where you can see that the data structure uses the integer as node id.
The ``t_rr_edge_info`` has been used in many functions across ``rr_graph2.cpp``, ``rr_graph.cpp`` as well the clock builder functions, whenever an edge is added.
You can see it will require massive changes to the code base.

To avoid big changes in this PR, I added the TODO comment to the ``t_rr_edge_info``, as a reminder that we must change it before finishing the refactoring.
This should be the topic of my next PR after my current two PRs are merged.
See if this is o.k. for you.

vpr/src/route/clock_connection_builders.cpp
tangxifan: Just checked the dependency of this API. It is used only by another API. It is simple change. Done.

vpr/src/route/clock_connection_builders.cpp
tangxifan: See my reply in a previous comment on the ``t_rr_edge_info``

vpr/src/route/clock_network_builders.cpp
tangxifan: Actually, I believe, we can use the API ``add_node_to_all_locs()`` as you suggested in another PR #1800. These for loops will be replaced by a simple line: 

```
rr_graph_builder.add_node_to_all_locs(chanx_node)
```

I leave a TODO comment here to remind me to finish the API replacement after PR #1800 is merged.
Let me know if this is fine for you.

vpr/src/route/clock_network_builders.cpp
tangxifan: Yes. I think the new API API ``add_node_to_all_locs()`` as you suggested in another PR #1800 will resolve this issue. 
I leave a TODO comment here to remind me to finish the API replacement after PR #1800 is merged.
Let me know if this is fine for you.

vpr/src/route/clock_connection_builders.cpp
tangxifan: Agree. We should not expose the side argument to developers because they may feel confused.
You are right on the SIDE[0] which is actually the TOP side. This convention has been there for a while (I do not know who introduced it). I am o.k. to change it if necessary.

Actually, the side argument for non-IPIN/OPIN nodes will be overwritten inside the API.
As a result, developers cannot mess up the side by giving any invalid values.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/c06800fb49d2732a0fcedbb2f542613fedd6a330/vpr/src/device/rr_spatial_lookup.cpp#L23-L29

However, I think it is not a good idea to force the developers to give a dummy side when using the API. I modified the API so that the side option will have a default value (SIDE[0]) if not specified. In client functions, I remove all the usage of the SIDE[0] for non-IPIN/OPIN nodes.

vpr/src/device/rr_spatial_lookup.cpp
tangxifan: Sorry for the typo. Just fixed it.
I think it will be a good time to rethink the ``rr_node_indices`` data structure once all the related APIs are replaced. 
Actually, it is not too far. I expect only we need another 2 pull requests, before fully shadowing the current ``rr_node_indices``. 

I put a TODO to remind us during the refactoring. I will take a deep look into the ``rr_node_indices`` data structure and propose something.

vpr/src/route/clock_network_builders.cpp
vaughnbetz: Sounds good; thanks.

vpr/src/route/clock_connection_builders.cpp
vaughnbetz: OK, thanks.

vaughnbetz:
Looks good; thanks.

PR #1802, opened 2021-07-15
Cluster attraction groups

sfkhalid:
#### Description
In order to place blocks in user-specified floorplan regions, the packing stage of vpr should pack primitives together when they are part of the same floorplan region. This is especially important if the specified floorplan regions are small, since it will be harder to make sure all the necessary primitives are placed within the region if they are packed into many different clustered blocks.

To achieve this goal during packing, attraction groups were introduced to the clustering process. The attraction groups are made up of atoms that are part of the same partition. During clustering, the molecule gain is increased for molecules that have the same attraction group as the cluster (if the cluster has an associated attraction group). This helps ensure that primitives in the same attraction group get packed together.

#### Related Issue
This pull request is part of the work related to issue #932.

### DIFF
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -27,7 +27,6 @@
 #include "compressed_grid.h"
 #include "metadata_storage.h"
 #include "vpr_constraints.h"
-#include "attraction_groups.h"
 
 /**
  * @brief A Context is collection of state relating to a particular part of VPR
@@ -417,8 +416,6 @@ struct FloorplanningContext : public Context {
      * The constraints on each cluster are computed during the clustering process and can change.
      */
     vtr::vector<ClusterBlockId, PartitionRegion> cluster_constraints;
-
-    AttractionInfo attraction_groups;
 };
 
 /**
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -27,6 +27,7 @@
 #include "compressed_grid.h"
 #include "metadata_storage.h"
 #include "vpr_constraints.h"
+#include "attraction_groups.h"
 
 /**
  * @brief A Context is collection of state relating to a particular part of VPR
@@ -416,6 +417,8 @@ struct FloorplanningContext : public Context {
      * The constraints on each cluster are computed during the clustering process and can change.
      */
     vtr::vector<ClusterBlockId, PartitionRegion> cluster_constraints;
+
+    AttractionInfo attraction_groups;
 };
 
 /**
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -1,0 +1,6 @@
+#include "attraction_groups.h"
+
+
+
+
+
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -11,7 +11,10 @@ AttractionInfo::AttractionInfo(bool attraction_groups_on) {
     atom_attraction_group.resize(num_atoms);
     fill(atom_attraction_group.begin(), atom_attraction_group.end(), AttractGroupId::INVALID());
 
-    //Create an attraction group for each partition in the floorplanning constraints
+    /*
+     * Create an attraction group for each partition in the floorplanning constraints
+     * if the packer option for attraction groups is turned on.
+     */
     if (attraction_groups_on) {
         for (int ipart = 0; ipart < num_parts; ipart++) {
             PartitionId partid(ipart);
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -36,7 +36,7 @@ AttractionInfo::AttractionInfo() {
     }
 }
 
-AttractionGroup AttractionInfo::get_attraction_group_info(const AttractGroupId group_id) {
+const AttractionGroup& AttractionInfo::get_attraction_group_info(const AttractGroupId group_id) {
     return attraction_groups[group_id];
 }
 
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -1,6 +1,38 @@
 #include "attraction_groups.h"
 
+AttractGroupId AttractionInfo::get_atom_attraction_group(AtomBlockId atom_id) {
+	return atom_attraction_group[atom_id];
+}
 
+AttractionGroup AttractionInfo::get_attraction_group_info(AttractGroupId group_id) {
+	return attraction_groups[group_id];
+}
 
+void AttractionInfo::set_atom_attraction_group(AtomBlockId atom_id, AttractGroupId group_id) {
+	atom_attraction_group[atom_id] = group_id;
+}
 
+void AttractionInfo::set_attraction_group_info(AttractGroupId group_id, AttractionGroup group_info) {
+	attraction_groups[group_id] = group_info;
+}
 
+void AttractionInfo::add_attraction_group(AttractionGroup group_info) {
+	attraction_groups.push_back(group_info);
+}
+
+int AttractionInfo::num_attraction_groups() {
+	return attraction_groups.size();
+}
+
+float AttractionInfo::get_attraction_group_gain(AttractGroupId group_id) {
+	return attraction_groups[group_id].gain;
+}
+
+void AttractionInfo::set_attraction_group_gain(AttractGroupId group_id, float new_gain) {
+	attraction_groups[group_id].gain = new_gain;
+}
+
+void AttractionInfo::initialize_atom_attraction_groups(int num_atoms) {
+	atom_attraction_group.resize(num_atoms);
+	fill(atom_attraction_group.begin(), atom_attraction_group.end(), NO_ATTRACTION_GROUP);
+}
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -25,7 +25,7 @@ AttractionInfo::AttractionInfo() {
     //Then, fill in the group id for the atoms that do have an attraction group
     int num_att_grps = attraction_groups.size();
 
-    for (int igroup; igroup < num_att_grps; igroup++) {
+    for (int igroup = 0; igroup < num_att_grps; igroup++) {
         AttractGroupId group_id(igroup);
 
         AttractionGroup att_group = attraction_groups[group_id];
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -36,39 +36,14 @@ AttractionInfo::AttractionInfo() {
     }
 }
 
-AttractGroupId AttractionInfo::get_atom_attraction_group(AtomBlockId atom_id) {
-    return atom_attraction_group[atom_id];
-}
-
-AttractionGroup AttractionInfo::get_attraction_group_info(AttractGroupId group_id) {
+AttractionGroup AttractionInfo::get_attraction_group_info(const AttractGroupId group_id) {
     return attraction_groups[group_id];
 }
 
-void AttractionInfo::set_atom_attraction_group(AtomBlockId atom_id, AttractGroupId group_id) {
-    atom_attraction_group[atom_id] = group_id;
-}
-
-void AttractionInfo::set_attraction_group_info(AttractGroupId group_id, AttractionGroup group_info) {
+void AttractionInfo::set_attraction_group_info(AttractGroupId group_id, const AttractionGroup& group_info) {
     attraction_groups[group_id] = group_info;
 }
 
-void AttractionInfo::add_attraction_group(AttractionGroup group_info) {
+void AttractionInfo::add_attraction_group(const AttractionGroup& group_info) {
     attraction_groups.push_back(group_info);
-}
-
-int AttractionInfo::num_attraction_groups() {
-    return attraction_groups.size();
-}
-
-float AttractionInfo::get_attraction_group_gain(AttractGroupId group_id) {
-    return attraction_groups[group_id].gain;
-}
-
-void AttractionInfo::set_attraction_group_gain(AttractGroupId group_id, float new_gain) {
-    attraction_groups[group_id].gain = new_gain;
-}
-
-void AttractionInfo::initialize_atom_attraction_groups(int num_atoms) {
-    atom_attraction_group.resize(num_atoms);
-    fill(atom_attraction_group.begin(), atom_attraction_group.end(), NO_ATTRACTION_GROUP);
 }
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -1,5 +1,41 @@
 #include "attraction_groups.h"
 
+AttractionInfo::AttractionInfo() {
+    auto& floorplanning_ctx = g_vpr_ctx.mutable_floorplanning();
+    auto& atom_ctx = g_vpr_ctx.atom();
+    int num_parts = floorplanning_ctx.constraints.get_num_partitions();
+
+    //Create an attraction group for each partition in the floorplanning constraints
+    for (int ipart = 0; ipart < num_parts; ipart++) {
+        PartitionId partid(ipart);
+        Partition part = floorplanning_ctx.constraints.get_partition(partid);
+
+        AttractionGroup group_info;
+        group_info.group_atoms = floorplanning_ctx.constraints.get_part_atoms(partid);
+
+        attraction_groups.push_back(group_info);
+    }
+
+    //Initialize every atom to have no attraction group id
+    int num_atoms = atom_ctx.nlist.blocks().size();
+
+    atom_attraction_group.resize(num_atoms);
+    fill(atom_attraction_group.begin(), atom_attraction_group.end(), NO_ATTRACTION_GROUP);
+
+    //Then, fill in the group id for the atoms that do have an attraction group
+    int num_att_grps = attraction_groups.size();
+
+    for (int igroup; igroup < num_att_grps; igroup++) {
+        AttractGroupId group_id(igroup);
+
+        AttractionGroup att_group = attraction_groups[group_id];
+
+        for (unsigned int iatom = 0; iatom < att_group.group_atoms.size(); iatom++) {
+            atom_attraction_group[att_group.group_atoms[iatom]] = group_id;
+        }
+    }
+}
+
 AttractGroupId AttractionInfo::get_atom_attraction_group(AtomBlockId atom_id) {
     return atom_attraction_group[atom_id];
 }
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -20,7 +20,7 @@ AttractionInfo::AttractionInfo() {
     int num_atoms = atom_ctx.nlist.blocks().size();
 
     atom_attraction_group.resize(num_atoms);
-    fill(atom_attraction_group.begin(), atom_attraction_group.end(), NO_ATTRACTION_GROUP);
+    fill(atom_attraction_group.begin(), atom_attraction_group.end(), AttractGroupId::INVALID());
 
     //Then, fill in the group id for the atoms that do have an attraction group
     int num_att_grps = attraction_groups.size();
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -1,20 +1,9 @@
 #include "attraction_groups.h"
 
-AttractionInfo::AttractionInfo() {
+AttractionInfo::AttractionInfo(bool attraction_groups_on) {
     auto& floorplanning_ctx = g_vpr_ctx.mutable_floorplanning();
     auto& atom_ctx = g_vpr_ctx.atom();
     int num_parts = floorplanning_ctx.constraints.get_num_partitions();
-
-    //Create an attraction group for each partition in the floorplanning constraints
-    for (int ipart = 0; ipart < num_parts; ipart++) {
-        PartitionId partid(ipart);
-        Partition part = floorplanning_ctx.constraints.get_partition(partid);
-
-        AttractionGroup group_info;
-        group_info.group_atoms = floorplanning_ctx.constraints.get_part_atoms(partid);
-
-        attraction_groups.push_back(group_info);
-    }
 
     //Initialize every atom to have no attraction group id
     int num_atoms = atom_ctx.nlist.blocks().size();
@@ -22,16 +11,29 @@ AttractionInfo::AttractionInfo() {
     atom_attraction_group.resize(num_atoms);
     fill(atom_attraction_group.begin(), atom_attraction_group.end(), AttractGroupId::INVALID());
 
-    //Then, fill in the group id for the atoms that do have an attraction group
-    int num_att_grps = attraction_groups.size();
+    //Create an attraction group for each partition in the floorplanning constraints
+    if (attraction_groups_on) {
+        for (int ipart = 0; ipart < num_parts; ipart++) {
+            PartitionId partid(ipart);
+            Partition part = floorplanning_ctx.constraints.get_partition(partid);
 
-    for (int igroup = 0; igroup < num_att_grps; igroup++) {
-        AttractGroupId group_id(igroup);
+            AttractionGroup group_info;
+            group_info.group_atoms = floorplanning_ctx.constraints.get_part_atoms(partid);
 
-        AttractionGroup att_group = attraction_groups[group_id];
+            attraction_groups.push_back(group_info);
+        }
 
-        for (unsigned int iatom = 0; iatom < att_group.group_atoms.size(); iatom++) {
-            atom_attraction_group[att_group.group_atoms[iatom]] = group_id;
+        //Then, fill in the group id for the atoms that do have an attraction group
+        int num_att_grps = attraction_groups.size();
+
+        for (int igroup = 0; igroup < num_att_grps; igroup++) {
+            AttractGroupId group_id(igroup);
+
+            AttractionGroup att_group = attraction_groups[group_id];
+
+            for (unsigned int iatom = 0; iatom < att_group.group_atoms.size(); iatom++) {
+                atom_attraction_group[att_group.group_atoms[iatom]] = group_id;
+            }
         }
     }
 }
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -1,38 +1,38 @@
 #include "attraction_groups.h"
 
 AttractGroupId AttractionInfo::get_atom_attraction_group(AtomBlockId atom_id) {
-	return atom_attraction_group[atom_id];
+    return atom_attraction_group[atom_id];
 }
 
 AttractionGroup AttractionInfo::get_attraction_group_info(AttractGroupId group_id) {
-	return attraction_groups[group_id];
+    return attraction_groups[group_id];
 }
 
 void AttractionInfo::set_atom_attraction_group(AtomBlockId atom_id, AttractGroupId group_id) {
-	atom_attraction_group[atom_id] = group_id;
+    atom_attraction_group[atom_id] = group_id;
 }
 
 void AttractionInfo::set_attraction_group_info(AttractGroupId group_id, AttractionGroup group_info) {
-	attraction_groups[group_id] = group_info;
+    attraction_groups[group_id] = group_info;
 }
 
 void AttractionInfo::add_attraction_group(AttractionGroup group_info) {
-	attraction_groups.push_back(group_info);
+    attraction_groups.push_back(group_info);
 }
 
 int AttractionInfo::num_attraction_groups() {
-	return attraction_groups.size();
+    return attraction_groups.size();
 }
 
 float AttractionInfo::get_attraction_group_gain(AttractGroupId group_id) {
-	return attraction_groups[group_id].gain;
+    return attraction_groups[group_id].gain;
 }
 
 void AttractionInfo::set_attraction_group_gain(AttractGroupId group_id, float new_gain) {
-	attraction_groups[group_id].gain = new_gain;
+    attraction_groups[group_id].gain = new_gain;
 }
 
 void AttractionInfo::initialize_atom_attraction_groups(int num_atoms) {
-	atom_attraction_group.resize(num_atoms);
-	fill(atom_attraction_group.begin(), atom_attraction_group.end(), NO_ATTRACTION_GROUP);
+    atom_attraction_group.resize(num_atoms);
+    fill(atom_attraction_group.begin(), atom_attraction_group.end(), NO_ATTRACTION_GROUP);
 }
--- a/vpr/src/pack/attraction_groups.h
+++ b/vpr/src/pack/attraction_groups.h
@@ -11,6 +11,7 @@
 #include "vtr_strong_id.h"
 #include "vtr_vector.h"
 #include "atom_netlist.h"
+#include "globals.h"
 
 /// @brief Type tag for AttractGroupId
 struct attraction_id_tag;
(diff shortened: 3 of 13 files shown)


vpr/src/pack/cluster.cpp
```diff
@@ -2050,6 +2062,13 @@ static void update_total_gain(float alpha, float beta, bool timing_driven, bool
                 cur_pb->pb_stats->sharinggain[blk_id] = 0;
             }
 
+            //Attraction Group stuff
```
sfkhalid, line 2065: Remove or change comment

vpr/src/base/vpr_context.h
```diff
@@ -416,6 +417,8 @@ struct FloorplanningContext : public Context {
      * The constraints on each cluster are computed during the clustering process and can change.
      */
     vtr::vector<ClusterBlockId, PartitionRegion> cluster_constraints;
+
+    AttractionInfo attraction_groups;
```
sfkhalid, line 421: Might be better in the clustering context.

sfkhalid: Clear the attraction groups after clustering and comment that you are doing so. (Would lead to wasted memory if no one else needs it after clustering).

sfkhalid: Actually would be better if you create it early (during prepacking) and then pass it around by reference to the routines that need it during clustering. Can find some different structures that are all passed around during clustering and group them together so that you don't have to add it individually to too many parameter lists. Can call it during cl_optimize or something along those lines.

Can create it in a header file (probably cluster.h). Can just make it a struct with no functions (or maybe have a constructor and destructor if they are all created and destroyed and the same time).

vpr/src/pack/attraction_groups.cpp
```diff
@@ -0,0 +1,38 @@
+#include "attraction_groups.h"
+
+AttractGroupId AttractionInfo::get_atom_attraction_group(AtomBlockId atom_id) {
```
sfkhalid, line 3: for a short function like this that just returns an int, could make it an inline function in the header file

vpr/src/pack/attraction_groups.cpp
```diff
@@ -0,0 +1,38 @@
...
+    return atom_attraction_group[atom_id];
+}
+
+AttractionGroup AttractionInfo::get_attraction_group_info(AttractGroupId group_id) {
```
sfkhalid, line 7: return by reference, not by value

vpr/src/pack/attraction_groups.cpp
sfkhalid: make all the short ones in this file inline in the header file

vpr/src/pack/attraction_groups.cpp
```diff
@@ -0,0 +1,38 @@
...
+    return attraction_groups[group_id];
+}
+
+void AttractionInfo::set_atom_attraction_group(AtomBlockId atom_id, AttractGroupId group_id) {
```
sfkhalid, line 11: see if you should also add the atom in the other data structure (the vector of atoms that are part of an attraction group)

sfkhalid: may not need these accessors, can also just give the callers direct access to the data structures

vpr/src/pack/attraction_groups.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+float AttractionInfo::get_attraction_group_gain(AttractGroupId group_id) {
+    return attraction_groups[group_id].gain;
+}
+
+void AttractionInfo::set_attraction_group_gain(AttractGroupId group_id, float new_gain) {
+    attraction_groups[group_id].gain = new_gain;
+}
+
+void AttractionInfo::initialize_atom_attraction_groups(int num_atoms) {
```
sfkhalid, line 35: make this a default constructor, can get num atoms from that atom context

vpr/src/pack/attraction_groups.h
```diff
@@ -0,0 +1,58 @@
...
+ *      Author: khalid88
+ */
+
+#ifndef VPR_SRC_PACK_ATTRACTION_GROUPS_H_
+#define VPR_SRC_PACK_ATTRACTION_GROUPS_H_
+
+#include "vtr_strong_id.h"
+#include "vtr_vector.h"
+#include "atom_netlist.h"
+
```
sfkhalid, line 14: Add high-level comment. What is the purpose of attraction group, comment each member of the attraction group and say what it is used for. For example, why do I need a bool that tells me whether the region is size one

vpr/src/pack/attraction_groups.h
```diff
@@ -0,0 +1,58 @@
...
+    float gain = 5; //give every attraction group an initial gain of 5
+    bool region_size_one = false;
+};
+
+/// @brief sentinel value for indicating that an attraction group has not been specified
+constexpr AttractGroupId NO_ATTRACTION_GROUP(-1);
+
+class AttractionInfo {
+  public:
+    AttractGroupId get_atom_attraction_group(AtomBlockId atom_id);
```
sfkhalid, line 32: add short comment to all of these, ex. setters and getters below

vpr/src/pack/cluster.cpp
```diff
@@ -300,6 +300,11 @@ void add_cluster_molecule_candidates_by_highfanout_connectivity(t_pb* cur_pb,
                                                                 const std::multimap<AtomBlockId, t_pack_molecule*>& atom_molecules,
                                                                 const int feasible_block_array_size);
 
+void add_cluster_molecule_candidates_by_attraction_group(t_pb* cur_pb,
+                                                         t_cluster_placement_stats* cluster_placement_stats_ptr,
+                                                         const std::multimap<AtomBlockId, t_pack_molecule*>& atom_molecules,
```
sfkhalid, line 305: could put atom_molecules, cluster_placement_stats_ptr, and the attraction groups data structure in a struct together and pass that around

vpr/src/pack/cluster.cpp
```diff
@@ -300,6 +300,11 @@ void add_cluster_molecule_candidates_by_highfanout_connectivity(t_pb* cur_pb,
...
+                                                         const int feasible_block_array_size);
```
sfkhalid, line 306: the feasible block array size could go into t_pb_stats with the actual array

vpr/src/pack/cluster.cpp
```diff
@@ -2041,6 +2048,11 @@ static void update_total_gain(float alpha, float beta, bool timing_driven, bool
...
     auto& atom_ctx = g_vpr_ctx.atom();
     t_pb* cur_pb = pb;
+
+    auto& floorplanning_ctx = g_vpr_ctx.mutable_floorplanning();
+    auto& attraction_info = floorplanning_ctx.attraction_groups;
+    AttractGroupId cluster_att_grp_id = cur_pb->pb_stats->attraction_grp_id;
+
     while (cur_pb) {
         for (AtomBlockId blk_id : cur_pb->pb_stats->marked_blocks) {
             if (cur_pb->pb_stats->connectiongain.count(blk_id) == 0) {
```
sfkhalid, line 2058: comment what this is doing

vpr/src/pack/cluster.cpp
```diff
@@ -2050,6 +2062,13 @@ static void update_total_gain(float alpha, float beta, bool timing_driven, bool
                 cur_pb->pb_stats->sharinggain[blk_id] = 0;
             }
 
+            //Attraction Group stuff
+            AttractGroupId atom_grp_id = attraction_info.get_atom_attraction_group(blk_id);
+            if (atom_grp_id != NO_ATTRACTION_GROUP && atom_grp_id == cluster_att_grp_id) {
+                //increase gain of atom by 1
+                cur_pb->pb_stats->gain[blk_id]++;
```
sfkhalid, line 2069: use the attraction group gain to update

sfkhalid: right now you are updating the blocks that are part of the attraction group multiple times - decide if that is a problem. If it isn't a problem, comment that that is what you are doing. If it is a problem, write code to make sure that you only do this once (i.e. did the cluster attraction group just change) - if you can do this it is good because it can be faster

vpr/src/pack/cluster.cpp
```diff
@@ -2115,6 +2134,8 @@ static void update_cluster_stats(const t_pack_molecule* molecule,
     molecule_size = get_array_size_of_molecule(molecule);
     cb = nullptr;
 
+    auto& floorplanning_ctx = g_vpr_ctx.mutable_floorplanning();
```
sfkhalid, line 2137: will have these passed in as a parameter as a part of the struct I am creating

sfkhalid: can remove the above to-do comment from Vaughn. Or change the comment to a better overall comment for what the function does.

vpr/src/pack/cluster.cpp
```diff
@@ -2128,13 +2149,23 @@ static void update_cluster_stats(const t_pack_molecule* molecule,
...
+
         while (cur_pb) {
             /* reset list of feasible blocks */
             if (cur_pb->is_root()) {
                 cb = cur_pb;
             }
             cur_pb->pb_stats->num_feasible_blocks = NOT_VALID;
             cur_pb->pb_stats->num_child_blocks_in_pb++;
+
+            if (atom_grp_id != NO_ATTRACTION_GROUP) {
```
sfkhalid, line 2164: The StrongId types (like AttractGroupId) will have AttractGroupId::INVALID(). Use that and get rid of NO_ATTRACTION_GROUP

vpr/src/pack/cluster.cpp
```diff
@@ -2418,6 +2449,15 @@ static t_pack_molecule* get_highest_gain_molecule(t_pb* cur_pb,
         }
     }
 
+    add_cluster_molecule_candidates_by_attraction_group(cur_pb, cluster_placement_stats_ptr, atom_molecules, feasible_block_array_size);
+
+    //t_pack_molecule* molecule = nullptr;
+
+    //Get highest gain molecule that also belongs to attraction group
```
sfkhalid, line 2456: delete this commented out code

vpr/src/pack/cluster.cpp
```diff
@@ -2522,6 +2562,50 @@ void add_cluster_molecule_candidates_by_highfanout_connectivity(t_pb* cur_pb,
...
+        AttractionGroup group = attraction_info.get_attraction_group_info(grp_id);
+
+        for (AtomBlockId blk_id : group.group_atoms) {
+            if (atom_ctx.lookup.atom_clb(blk_id) == ClusterBlockId::INVALID()) {
+                auto rng = atom_molecules.equal_range(blk_id);
+                for (const auto& kv : vtr::make_range(rng.first, rng.second)) {
+                    t_pack_molecule* molecule = kv.second;
+                    if (molecule->valid) {
+                        bool success = true;
+                        for (int j = 0; j < get_array_size_of_molecule(molecule); j++) {
```
sfkhalid, line 2586: can change j to indicate that it is an atom within the molecule (i_atom)

vpr/src/pack/cluster.cpp
```diff
@@ -2522,6 +2562,50 @@ void add_cluster_molecule_candidates_by_highfanout_connectivity(t_pb* cur_pb,
...
+    //If the current cluster belongs to an attraction group, add all of the atoms
+    //from that attraction group to the feasible blocks
+    AttractGroupId grp_id = cur_pb->pb_stats->attraction_grp_id;
+    if (grp_id != NO_ATTRACTION_GROUP) {
+        AttractionGroup group = attraction_info.get_attraction_group_info(grp_id);
+
+        for (AtomBlockId blk_id : group.group_atoms) {
+            if (atom_ctx.lookup.atom_clb(blk_id) == ClusterBlockId::INVALID()) {
+                auto rng = atom_molecules.equal_range(blk_id);
+                for (const auto& kv : vtr::make_range(rng.first, rng.second)) {
```
sfkhalid, line 2582: might be adding the molecule multiple times (because different atoms might be part of the same molecule) -> if not a problem, comment it, if it is -> can have an unordered set to check if it has been added.

Also put this in a subroutine and refactor all of the functions that go through this atom_molecules loop to call the subroutine

vpr/src/pack/cluster.cpp
```diff
@@ -2566,6 +2650,49 @@ void add_cluster_molecule_candidates_by_transitive_connectivity(t_pb* cur_pb,
     }
 }
 
+t_pack_molecule* sort_cluster_molecule_candidates_by_attraction(t_pb* cur_pb) {
```
sfkhalid, line 2653: delete this routine

vpr/src/pack/prepack.cpp
```diff
@@ -1684,3 +1685,35 @@ static void print_chain_starting_points(t_pack_patterns* chain_pattern) {
 
     VTR_LOG("\n");
 }
+
+void alloc_attraction_groups() {
```
sfkhalid, line 1689: put this inside the attraction group class. make a default constructor that sets it to invalid/empty. and then put this code in an initialization routine in the class.

sfkhalid: if you have all the info you need though, just put this routine into the default constructor.

sfkhalid: if you can return stuff by reference in these routines, do so

vpr/src/pack/prepack.h
```diff
@@ -22,4 +22,6 @@ t_pack_molecule* alloc_and_load_pack_molecules(t_pack_patterns* list_of_pack_pat
 
 void free_pack_molecules(t_pack_molecule* list_of_pack_molecules);
 
+void alloc_attraction_groups();
```
sfkhalid, line 25: won't need this declaration anymore

vaughnbetz:
Added WIP to the title as this isn't ready for merge yet.
Email from @luujason indicates our understanding of the pb_stats and gains seems correct -- only the top-level (root/cluster) pb_stats is used for gains.  That means @sfkhalid should refactor the code to move the gains-related data out of the pb_stats data structure and into a new cl_optimize data structure (along with the other data we talked about today).  
That could be done in a separate pr if it is easier to get the current code going with the redundant pb_stats loading code for attraction gains -- I suspect that is best as it will minimize the size of each PR.

vpr/src/pack/cluster.cpp
sfkhalid: Does not get added multiple times because add_molecule_to_pb_stats checks whether the molecule is already in feasible blocks

vpr/src/pack/cluster.cpp
sfkhalid: feasible_block_array_size is the packer option that specifies the max size of the array. It is different from the number of feasible blocks currently in the array, which is already part of t_pb_stats

sfkhalid:
QoR comparison of this branch vs. master:
[att_groups_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6862570/att_groups_qor_comparison.xlsx)
No significant changes noted in run time.

vaughnbetz:
This branch was created on Sunday, July 11 (or very close to that date).  That's well before @tangxifan 's recent RRGraphBuilder changes were merged (that was four days ago, or July 18).  Both the sanitized push test and the sanitized pull request test on vtr_reg_basic are failing due to out of memory.  Is the push test running only on Sarah's branch (without pulling the changes into master) while the pull request test is running on the master with Sarah's changes pulled in?  If so, the fact that both are failing suggests @tangxifan 's recent changes are not the culprit.  (I'm not quite sure of how the push vs. pull request tests differ, or if in fact they're just duplicates of each other).

vaughnbetz:
S:basic on the push had 7 failures and S:basic on the pull had 6 failures.  So if they're running on the same code, that indicates some contention between jobs that's leading to out of memory errors as it isn't completely predictable.

vaughnbetz:
Odin basic tests on the pull request failed 6 benchmarks and Odin basic tests on the push failed 7 so again some difference.

vaughnbetz:
The pattern of failures in the main body of the Odin II basic regtests looks the same in the push and pull request versions (same tests fail, just sometimes in a slightly different order, presumably due to parallel jobs).

vaughnbetz:
Odin II basic regtests are configured to use the sanitizers as well.  So that looks like a big clue: the two sets of runs we have that use the sanitizers (Odin II & vtr basic) on github are failing with out of memory.  
The strong sanitized runs are not failing, but they run on kokoro which has bigger machines.  I think we have a configuration issue and need to increase the memory on the github runners for the sanitized basic tests, or move them to kokoro.
@tangxifan : adding your name to make sure you get this :).

vpr/src/pack/attraction_groups.h
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * @file
+ * @brief This file defines the AttractionInfo class, which is used to store atoms in attraction groups, which are
+ *        used during the clustering process.
+ *
+ * Overview
+ * ========
+ * Attraction groups are used during the clustering process to tell the clusterer which atoms should be packed
+ * in the same cluster by virtue of having the same floorplan constraints. The attraction groups may also be
```
sfkhalid, line 24: rephrase more precisely - "atoms in the same att group will be highly desirable to be packed together. Will be seen as having higher gain if in the same att group. Right now the att groups are based on floorplan constraints, atoms in the same partition are in the same att group. Right now used for floorplanning, but could be used for on other concepts in the future

vpr/src/pack/attraction_groups.h
```diff
@@ -0,0 +1,105 @@
...
+ * used to cluster similar atoms together in other ways in the future.
+ */
+
+/// @brief Type tag for AttractGroupId
+struct attraction_id_tag;
+
+/// @brief A unique identifier for a partition
```
sfkhalid, line 31: change this comment to att group

vpr/src/pack/attraction_groups.h
```diff
@@ -0,0 +1,105 @@
...
+     */
+    float gain = 3;
+
+    /*
+     * If the group is made up from a partition of atoms that are confined to a size one spot
+     * (i.e. one x, y grid location), the clusterer will immediately put all atoms in the group
+     * into the same cluster
+     */
+    /* TODO: Add the code in the clusterer that will do the above steps. */
+    bool region_size_one = false;
```
sfkhalid, line 50: potentially rename - must_be_packed_in_one_cluster

sfkhalid: can even just comment this out for now, add it back when something uses it

vpr/src/pack/attraction_groups.h
```diff
@@ -0,0 +1,105 @@
...
+class AttractionInfo {
+  public:
+    //Constructor that fills in the attraction groups based on vpr's floorplan constraints.
+    //If no constraints were specified, then no attraction groups will be created.
+    AttractionInfo();
+
+    //Setters and getters for the class
+    AttractGroupId get_atom_attraction_group(const AtomBlockId atom_id);
+
+    AttractionGroup get_attraction_group_info(const AttractGroupId group_id);
```
sfkhalid, line 62: return by const reference

vpr/src/pack/attraction_groups.h
```diff
@@ -0,0 +1,105 @@
...
+    float get_attraction_group_gain(const AttractGroupId group_id);
+
+    void set_attraction_group_gain(const AttractGroupId group_id, const float new_gain);
+
+    void add_attraction_group(const AttractionGroup& group_info);
+
+    int num_attraction_groups();
+
+  private:
+    //Store each atom's attraction group assuming each atom is in at most one attraction group
```
sfkhalid, line 77: Atoms with no attraction group will have AttractGroupId::INVALID()

vpr/src/pack/cluster.cpp
```diff
@@ -2036,20 +2053,35 @@ static void mark_and_update_partial_gain(const AtomNetId net_id, enum e_gain_upd
 }
 
 /*****************************************/
-static void update_total_gain(float alpha, float beta, bool timing_driven, bool connection_driven, t_pb* pb) {
+static void update_total_gain(float alpha, float beta, bool timing_driven, bool connection_driven, t_pb* pb, AttractionInfo& attraction_groups) {
     /*Updates the total  gain array to reflect the desired tradeoff between*
```
sfkhalid, line 2057: write a better overall comment for this routine

sfkhalid: routine is called when new molecule added to cluster, updates gain for blocks in marked blocks array

vpr/src/pack/cluster.cpp
```diff
@@ -2092,25 +2124,15 @@ static void update_cluster_stats(const t_pack_molecule* molecule,
                                  const bool timing_driven,
                                  const bool connection_driven,
                                  const int high_fanout_net_threshold,
-                                 const SetupTimingInfo& timing_info) {
-    /* Updates cluster stats such as gain, used pins, and clock structures.  */
+                                 const SetupTimingInfo& timing_info,
+                                 AttractionInfo& attraction_groups) {
+    /* Updates cluster stats such as gain, used pins, and clock structures. Also
```
sfkhalid, line 2129: Write better overall comment - a molecule has been added to the cluster, this updates stats on that cluster such as used pins, etc. and makes calls to make sure gains of other atoms are adjusted to reflect the new cluster content

vpr/src/pack/cluster.cpp
```diff
@@ -2128,13 +2150,22 @@ static void update_cluster_stats(const t_pack_molecule* molecule,
...
+
         while (cur_pb) {
             /* reset list of feasible blocks */
             if (cur_pb->is_root()) {
                 cb = cur_pb;
             }
             cur_pb->pb_stats->num_feasible_blocks = NOT_VALID;
             cur_pb->pb_stats->num_child_blocks_in_pb++;
+
+            if (atom_grp_id != AttractGroupId::INVALID()) {
```
sfkhalid, line 2165: To-Do comment - allow more than one attraction group

sfkhalid: otherwise you are overwriting the attraction group that was previously in the cluster

sfkhalid: can have a vector of attraction groups for each cluster

vpr/src/pack/cluster.cpp
```diff
@@ -2566,9 +2593,30 @@ void add_cluster_molecule_candidates_by_transitive_connectivity(t_pb* cur_pb,
...
+bool check_free_primitives_for_molecule_atoms(t_pack_molecule* molecule, t_cluster_placement_stats* cluster_placement_stats_ptr) {
+    auto& atom_ctx = g_vpr_ctx.atom();
+    bool success = true;
+
+    for (int i_atom = 0; i_atom < get_array_size_of_molecule(molecule); i_atom++) {
+        if (molecule->atom_block_ids[i_atom]) {
+            VTR_ASSERT(atom_ctx.lookup.atom_clb(molecule->atom_block_ids[i_atom]) == ClusterBlockId::INVALID());
+            auto blk_id2 = molecule->atom_block_ids[i_atom];
+            if (!exists_free_primitive_for_atom_block(cluster_placement_stats_ptr, blk_id2)) {
+                /* TODO: debating whether to check if placement exists for molecule
```
sfkhalid, line 2605: add Jason Luu's name so people would know who to ask about this

vpr/src/pack/cluster.cpp
```diff
@@ -300,6 +302,12 @@ void add_cluster_molecule_candidates_by_highfanout_connectivity(t_pb* cur_pb,
                                                                 const std::multimap<AtomBlockId, t_pack_molecule*>& atom_molecules,
                                                                 const int feasible_block_array_size);
 
+void add_cluster_molecule_candidates_by_attraction_group(t_pb* cur_pb,
```
sfkhalid, line 305: make the functions you are adding static (add to declaration and definition) as well - makes them file scope

vpr/src/pack/cluster.cpp
```diff
@@ -300,6 +302,12 @@ void add_cluster_molecule_candidates_by_highfanout_connectivity(t_pb* cur_pb,
...
+                                                         t_cluster_placement_stats* cluster_placement_stats_ptr,
+                                                         const std::multimap<AtomBlockId, t_pack_molecule*>& atom_molecules,
+                                                         AttractionInfo& attraction_groups,
+                                                         const int feasible_block_array_size);
+
 void add_cluster_molecule_candidates_by_transitive_connectivity(t_pb* cur_pb,
```
sfkhalid, line 311: these should also technically be static

vpr/src/pack/cluster.cpp
sfkhalid: also add comments in the definitions (to match the other routines)

sfkhalid: add comments for ones you have figured out as well

sfkhalid:
As a note, I ran the sanitized basic tests locally, and they are failing locally as well. I ran the test on both this branch (cluster_attraction_groups) and the master branch, and there are failing tests on both branch. The tests are failing with the exit code 23.

tangxifan:
> As a note, I ran the sanitized basic tests locally, and they are failing locally as well. I ran the test on both this branch (cluster_attraction_groups) and the master branch, and there are failing tests on both branch. The tests are failing with the exit code 23.

@sfkhalid Can you share the log files on the locally failed tests? I can look into it and figure it how to patch. Thanks!

sfkhalid:
> > As a note, I ran the sanitized basic tests locally, and they are failing locally as well. I ran the test on both this branch (cluster_attraction_groups) and the master branch, and there are failing tests on both branch. The tests are failing with the exit code 23.
> 
> @sfkhalid Can you share the log files on the locally failed tests? I can look into it and figure it how to patch. Thanks!

Sure, below are the log files of some of the failing tests as well as summary of which tests failed.

[master_ch_intrinsics_vpr.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6889109/master_ch_intrinsics_vpr.txt)
[master_diffeq1_vpr.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6889110/master_diffeq1_vpr.txt)
[master_mkPktMerge_vpr.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6889114/master_mkPktMerge_vpr.txt)
[master_single_ff_vpr.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6889115/master_single_ff_vpr.txt)
[master_single_wire_vpr.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6889116/master_single_wire_vpr.txt)
[master_run_summary.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6889117/master_run_summary.txt)

These are all from the run done on the master branch. I've compared with the run that was done on the attraction groups branch and so far all the failures seem to be the same.

tangxifan:
@sfkhalid Just checked the reports. It seems to be a TBB issue. I double checked the log file of [S:Basic test](https://pipelines.actions.githubusercontent.com/k4WhX565mV5YoOwLfWUFDkcVVIQ2u8tSLZAJkQP8ExBUWpSugE/_apis/pipelines/1/runs/1827/signedlogcontent/22?urlExpires=2021-07-28T22%3A23%3A22.3594732Z&urlSigningMethod=HMACV1&urlSignature=RhB1mfJxq3v1pkOqcXetbOEyPvSVW%2BDXNdErqG7O5Yc%3D). The TBB is actually disabled in these tests. In my local tests, TBB was off as well. 

If you can turn off TBB, we may bypass this issue and see if we see the same issue as the Github runner shows.
Or we may consider to add some tests for TBB

sfkhalid:
@tangxifan Did your local sanitized basic tests pass with TBB off?

tangxifan:
> @tangxifan Did your local sanitized basic tests pass with TBB off?

@sfkhalid Yes. That is what I see on my side.

sfkhalid:
> > @tangxifan Did your local sanitized basic tests pass with TBB off?
> 
> @sfkhalid Yes. That is what I see on my side.

Ok, I see, thank you. I'll try running without tbb on my side and see what happens

vaughnbetz:
There is also a suppression file for the sanitizers that the regtests use.  Are you using that @sfkhalid ?  I believe there are a few issues in the Intel TBB library (small memory leaks, some object accesses that the sanitizer complains about) that we can't fix, so we suppress those errors.

sfkhalid:
> There is also a suppression file for the sanitizers that the regtests use. Are you using that @sfkhalid ? I believe there are a few issues in the Intel TBB library (small memory leaks, some object accesses that the sanitizer complains about) that we can't fix, so we suppress those errors.

@vaughnbetz No, I wasn't aware of that file. Where is it located/how do I enable it for the reg test?

vaughnbetz:
The leak sanitizer suppression file is at https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/vpr/lsan.supp

But looking at it, it doesn't have TBB suppression, just one suppression for graphics.  There is also a valgrind suppression file, and it also doesn't have TBB suppresion (but perhaps doesn't need one). @kmurray : should we make an lsan.supp file with the TBB memory leaks and invalid object messages suppressed?  If you already know those are spurious warnings it seems like we should make one.

sfkhalid:
A summary of some test cases:
[Attraction Groups Testing.docx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6901327/Attraction.Groups.Testing.docx)
These test cases were done to see whether using attraction groups during packing was giving any added benefit to helping VPR succeed with floorplan constraints.

vpr/src/pack/cluster.cpp
vaughnbetz: This is marked as resolved but I don't see Jason Luu's name added?

vpr/src/pack/cluster.cpp
```diff
@@ -2522,6 +2532,37 @@ void add_cluster_molecule_candidates_by_highfanout_connectivity(t_pb* cur_pb,
     cur_pb->pb_stats->tie_break_high_fanout_net = AtomNetId::INVALID(); /* Mark off that this high fanout net has been considered */
 }
 
+void add_cluster_molecule_candidates_by_attraction_group(t_pb* cur_pb,
```
vaughnbetz: No comment on what this function does.  Should have one.
I think this is a file scope function so should be static (I don't think it is in the header).
Should apply the same logic to all added functions: comment their use, and make them static if their declaration is not in the .h file. Comment their use in the .h file if they are declared there, or in the .cpp file (at the definition) if declared static in the .cpp file.
Should also declare all static functions near the top of the .cpp file (where the other static functions are declared).

vaughnbetz (changes requested):
Seems that not all the comments from our in-zoom review have been addressed, although the comments are marked resolved.  Please address before merging.  
Should update qor spreadsheet too so it checks the final code (after all updates).

vpr/src/pack/cluster.cpp
sfkhalid: Sure, I'll add a comment to this function and the other add_cluster_molecule functions.
I made all of the add_cluster_molecule_candidates_by_x functions static and added the declaration for this new one at the top of the file, I'm not sure why that's not showing here. I can see it on this pull request if I go on under files changed though.

vpr/src/pack/cluster.cpp
sfkhalid: It is added at the end of the comment, I'll move it to the beginning though so it's more clear

vpr/src/pack/cluster.cpp
vaughnbetz: Thanks.  Maybe I clicked on the wrong link and wasn't looking at the latest commit.

sfkhalid:
QoR spreadsheet for code comparison after all changes:
[att_groups_qor_comparison_2.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6909198/att_groups_qor_comparison_2.xlsx)
No significant changes were noted across the categories between the master branch and the attraction groups branch.

vaughnbetz:
Looks good; merging.

sfkhalid:
> Looks good; merging once CI goes green unless you tell me there is a reason to wait @sfkhalid

Nope, should be good to merge once the tests pass.

sfkhalid:
@vaughnbetz The tests passed, I think this is good to merge now

PR #1804, opened 2021-07-18
 ci: move scripts from '.github/gha' and '.github/travis' into '.github/scripts'

### LINKED ISSUE #1755: Remove travis name from CI scripts
Travis CI is no longer used on this project but there still exists a [`.github/travis` directory](https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/.github/travis).

I assume this is because these travis scripts are used by Kokoro or GitHub Actions? If so, they should probably be moved / renamed.

umarcor:
#### Description

This PR moves scripts from `.github/gha` and `.github/travis` into `.github/scripts`. Scripts that use those resources are updated accordingly. BTW, the workflow YAML is slightly simplified.

#### Related Issue

Close #1755

#### Motivation and Context

Travis is non used anymore.

CI is still all green.

#### Types of changes

#### Checklist:

#### Additional context

There are some remaining references to travis:

```
# grep -lr ./ -e "travis"
./abc/README.md
./libs/EXTERNAL/capnproto/README.md
./libs/EXTERNAL/libtatum/README.md
```

However, I assume that those are vendores dependencies, so they handle CI on their own.


dev/check-format-py.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+source $(dirname "$0")/../.github/scripts/common.sh
```
mithro, line 3: I think the goal was to keep these scripts from depending on stuff under the `.github` directory.

umarcor: Yeah. That's what I thought. However, the `common.sh` is currently explicitly sourced. See https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d3449e8744e447bd8d27df1c41ba9b6435918ddc/.github/workflows/test.yml#L37-L59

I tried to do minimal functional modifications in this PR. So, precisely, I preserved common to be sourced before/in `check-format.sh` and `check-formar-py.sh`, but I did not in `pylint_check.py`.

If you or any maintainer/developer can confirm that `common.sh` is not required in `check-format.sh`/`check-formar-py.sh`, I will update them.

Moreover, the git blame belongs to me. So, this is inherited from the previous `.travis.yml` file. There, the common script was source in all the jobs: https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/13a8c9f9b50082b3ef782020a9ab503faf1a31a3#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L243-L254

PR #1811, opened 2021-07-29
Fixed Fc values for designs with different horizontal and vertical channel widths

heinerb1:
#### Description
Work in progress follow up to PR #1786 that added the ability for the user to specify detailed routing for designs with different horizontal and vertical channel widths. I modified the Fc section of the build_rr_graph function and the alloc_and_load_actual_fc function which determines the actual Fc value for a particular pin. Added code in the alloc_and_load_actual_fc function that uses pin location to provide the correct Fc value using the channel width and Fc percentages specified by the user. However, the code is currently split into two sections and uses the channel widths to determine which part to use. I am beginning to work on changing this so that only one set of code is used for both cases. In addition, more documentation will need to be added and the code will need to be reviewed to determine where it needs to be improved, so there is still a decent amount of work that needs to be done.

#### Related Issue
#1768 

#### Motivation and Context
Due to the changes in PR #1786, a user could now specify detailed routing for designs with different horizontal and vertical channel widths. However, due to assumptions in rr_graph that one channel would be used for the entire design, the Fc values were being computed incorrectly for the smaller of the two channel widths. In order to correct this, the rr_graph builder needed to be modified to determine which channel a pin was connecting to and calculate the correct Fc value for that pin.

As mentioned above, since the code is currently split into two sections and uses the channel widths to determine which part to use, the code currently passes all tests. I have tested the new pieces with various different designs that had different and vertical channel widths, and the results seem to be correct so far. However, as additional changes are made to this code such as the unification process, it might not pass the tests and need to be modified.

#### Types of changes

#### Checklist:


vpr/src/route/rr_graph.cpp
```diff
@@ -310,6 +318,8 @@ static void build_rr_graph(const t_graph_type graph_type,
                            int* wire_to_rr_ipin_switch,
                            int* Warnings);
 
+std::vector<t_pin_loc> get_pin_loc(const e_pin_type pin_type, const t_physical_tile_type_ptr Type);
```
heinerb1, line 321: Forgot to remove this. It was used during my testing and is not necessary, so I will fix this in the next commit.

vpr/src/route/rr_graph.cpp
```diff
@@ -572,13 +591,15 @@ static void build_rr_graph(const t_graph_type graph_type,
         }
     }
 
+    // Not sure what perturbation is. This may also need to be modified since it uses sets_per_seg_type.
```
heinerb1, line 594: There are a few comments like this that I used for notes that will be removed in a future commit.

PR #1813, opened 2021-07-30
dev scripts should not depend on ci scripts

umarcor:
As discussed in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1804#discussion_r672426293, maybe the common CI script does not need to be sourced in the `dev/check-format*` scripts.

/cc @mithro

### DIFF
--- a/dev/check-format-py.sh
+++ b/dev/check-format-py.sh
@@ -1,33 +1,3 @@
 #!/usr/bin/env bash
 
-clean=$(git status -s -uno | wc -l) #Short ignore untracked
-
-if [ $clean -ne 0 ]; then
-    echo "Current working tree was not clean! This tool only works on clean checkouts"
-    exit 2
-else
-    echo "Code Formatting Check"
-    echo "====================="
-    make format-py > /dev/null 2>&1
-
-    valid_format=$(git diff | wc -l)
-
-    if [ $valid_format -ne 0 ]; then
-        echo "FAILED"
-        echo ""
-        echo "You *must* make the following changes to match the formatting style"
-        echo "-------------------------------------------------------------------"
-        echo ""
-
-        git diff
-
-        echo ""
-        echo "Run 'make format-py' to apply these changes"
-
-        git reset --hard > /dev/null
-        exit 1
-    else
-        echo "OK"
-    fi
-fi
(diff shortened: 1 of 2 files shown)


umarcor:
Since both scripts are the same, except for the make target, I changed `check-format-py.sh` so that it calls `check-format.sh`.

vaughnbetz:
This looks like a good change to me.  @jgoeders any concerns?
Re-started CI; will merge tomorrow if I don't hear any concerns.

umarcor:
*ping* @vaughnbetz @jgoeders

vaughnbetz:
python lint failure is unrelated (old issue, fixed on master already). No idea why the same python lint test on the pull request is run twice, with what appears to be the same commit, and one fails and one passes.  Will override and approve that though.

kokoro tests don't seem to be running; trying a force restart.

umarcor:
> python lint failure is unrelated (old issue, fixed on master already). No idea why the same python lint test on the pull request is run twice, with what appears to be the same commit, and one fails and one passes. Will override and approve that though.

I did not check it thoroughly, however, if they depend on external dependencies, those might be different even though the vtr sources are the same. I wouldn't worry about it.

I rebased, just in case.

vaughnbetz:
Thanks.  Passes python lint now; just waiting for CI to go green and I'll merge.

vaughnbetz:
kokoro seems stuck; forcing it to relaunch.

vaughnbetz:
strong sanitized failed with no log; kokoro launch issue?
Trying again.

vaughnbetz:
The yosys+OdinII tests should pass now (that was all merged) and the basic tests have been deprecated.  So re-running to see if everything passes this time ... hopefully!

umarcor:
For the record, all CI tests are green, except "ODIN-II Basic Test" and "Yosys+ODIN-II Basic", which are "Waiting for status to be reported". However, since @vaughnbetz commented that the basic tests were deprecated, I believe the CI state of this PR is a result of it not being rebased. Therefore, I rebased and force-pushed now.

vaughnbetz:
Some tests seem to not have launched; trying a kokoro force rerun.

PR #1814, opened 2021-08-01
route_common: Fix cost check assertion

gatecat:
I was looking through the VPR router to compare it to nextpnr's one and noticed a small issue in one of the assertions, base cost was being multiplied twice instead of by the accumulated historical cost. This caused an assertion failure when debug assertions are enabled.


vaughnbetz:
Thanks; definitely looks like a mistake.

vaughnbetz:
Kokoro seems to be stuck; restarting.
python lint issue is not a problem; was an issue on the master due to upgrading python lint version and it's fixed now.

PR #1816, opened 2021-08-02
RRGraphView node_R(), node_C(), node_rc_index() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::node_R()`,  `RRGraphView::node_C()` and `RRGraphView::node_rc_index()` throughout VTR. Every time `device_ctx.rr_nodes[index].R()` was called has been replaced with `rr_graph.node_R(RRNodeId(index))`. In order to do this, I followed a pattern similar to that in the last PR.

### Motivation and Context

### How Has This Been Tested?

I have run the regression tests for `vtr_reg_strong` along with the QoR testing found at `    $ ../scripts/run_vtr_task.py regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain`. Results from these QoR tests can be found below. The file containing all results can be found [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6917797/node_r_c.xlsx).

|---|---|---|
|vtr_flow_elapsed_time	|1|	0.994093231|
|odin_synth_time	|1|	1.028283499|
|abc_synth_time	|1|	1.00525759|
|max_vpr_mem	|1|	0.999672195|
|pack_time	|1|	1.016018952|
|place_time	|1|	0.996246966|
|min_chan_width_route_time	|1|	0.995872138|
|crit_path_route_time	|1|	1.001645744|

#### Types of changes

#### Checklist:


vpr/src/route/connection_router.cpp
```diff
@@ -685,13 +685,16 @@ void ConnectionRouter<Heap>::evaluate_timing_driven_node_costs(t_heap* to,
     float switch_Tdel = rr_switch_inf_[iswitch].Tdel;
     float switch_Cinternal = rr_switch_inf_[iswitch].Cinternal;
 
+    const auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
+
     //To node info
-    auto rc_index = rr_nodes_.node_rc_index(RRNodeId(to_node));
+    auto rc_index = rr_graph.node_rc_index(RRNodeId(to_node));
```
tangxifan, line 692: The connection_router object actually caches the rr_graph, in order to avoid the use of global variables. I suggest to add the RRGraphView to the cache here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/2e0bafe202e3f8cb4e1f099002f7272f4705068d/vpr/src/route/connection_router.h#L27-L44

Then you just use the cached copy ``rr_graph_`` in these functions.

tangxifan (changes requested):
@ethanroj23 It looks good to me in general. Just a few comments to be address for a clean code organization.

vpr/src/route/connection_router.cpp
ethanroj23: @tangxifan I am trying to add the RRGraphView to the cache with 

`const RRGraphView& rr_graph,` and

`, rr_graph_(rr_graph)`

But when I try this, the compiler gives me the following warning

`error: use of deleted function ‘RRGraphView::RRGraphView(const RRGraphView&)’`

Do you know what I am missing? I know that the copy constructors have been disabled in RRGraphView, but perhaps I am supposed to copy to cache in a different way.

tangxifan: Yes. You can consider to use the pointer 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/2e0bafe202e3f8cb4e1f099002f7272f4705068d/vpr/src/route/rr_graph_uxsdcxx_serializer.h#L1853

which works fine in the reader/writer class.

tangxifan: Anyhow, I think the use of ``const RRGraphView&`` is better. But not sure the compiler spits out such error. 
@hzeller If you have time, can you look into this and share your opinion? Thanks!

vaughnbetz: This code snippet looks like a call to the deleted copy constructor (but I am just going by the conversation, so I don't see the code in context):
`, rr_graph_(rr_graph)`

ethanroj23: @vaughnbetz, I think you are correct. I have implemented the `RRGraphView*` as @tangxifan suggested, but we could also modify `RRGraphView.h` so that the copy constructor was no longer deleted if that is what we wanted to do.

vaughnbetz:
No it will be expensive to copy so a pointer is better.

Vaughn

vpr/src/route/rr_graph_timing_params.cpp
```diff
@@ -191,7 +191,7 @@ void add_rr_graph_C_from_switches(float C_ipin_cblock) {
 
     //Create the final flywieghted t_rr_rc_data
     for (size_t inode = 0; inode < device_ctx.rr_nodes.size(); inode++) {
```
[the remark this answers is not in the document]
ethanroj23: Yes, I think that would be useful. We will have to consider that in the future.

ethanroj23:
> No it will be expensive to copy so a pointer is better. Vaughn
> […](#)
> On Aug 10, 2021, at 2:00 PM, ethanroj23 ***@***.***> wrote: ﻿ @ethanroj23 commented on this pull request. In vpr/src/route/connection_router.cpp: > //To node info - auto rc_index = rr_nodes_.node_rc_index(RRNodeId(to_node)); + auto rc_index = rr_graph.node_rc_index(RRNodeId(to_node)); @vaughnbetz, I think you are correct. I have implemented the RRGraphView* as @tangxifan suggested, but we could also modify RRGraphView.h so that the copy constructor was no longer deleted if that is what we wanted to do. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

Ok, sounds good. I will keep it as I have it for now.

tangxifan (approved):
@ethanroj23 Look good to me. Thanks for the effort.

vpr/src/route/route_timing.cpp
```diff
@@ -340,6 +340,7 @@ bool try_timing_driven_route_tmpl(const t_router_opts& router_opts,
         device_ctx.grid,
         *router_lookahead,
```
vaughnbetz, line 341: Couldn't this be the regular (const) device context rr_graph ?

utils/route_diag/src/main.cpp
```diff
@@ -102,6 +102,7 @@ static void do_one_route(int source_node, int sink_node,
             device_ctx.grid,
             *router_lookahead,
             device_ctx.rr_nodes,
+            &g_vpr_ctx.mutable_device().rr_graph,
```
vaughnbetz, line 105: Couldn't / shouldn't this be the regular (const) device context rr_graph?

vpr/src/route/connection_router.cpp
```diff
@@ -967,6 +967,7 @@ std::unique_ptr<ConnectionRouterInterface> make_connection_router(
     const DeviceGrid& grid,
     const RouterLookahead& router_lookahead,
     const t_rr_graph_storage& rr_nodes,
+    RRGraphView* rr_graph,
```
vaughnbetz, line 970: I don't think the RRGraphView should be modified by the connection router, so this should be a const RRGraphView* . 
That would then let you pass a const pointer using
&g_vpr_ctx.device().rr_graph (instead of mutable_device).

vaughnbetz (changes requested):
Looks mostly good, but I think you can pass const RRGraphView* pointers in several places, and then use const contexts (g_vtr_context.device().rr_graph).

vpr/test/test_connection_router.cpp
```diff
@@ -48,6 +48,7 @@ static float do_one_route(int source_node, int sink_node, const t_router_opts& r
         device_ctx.grid,
         *router_lookahead,
         device_ctx.rr_nodes,
+        &g_vpr_ctx.mutable_device().rr_graph,
```
vaughnbetz, line 51: Should be possible to make this a g_vpr_ctx.device (i.e. const).

vpr/src/route/router_delay_profiling.cpp
```diff
@@ -17,6 +17,7 @@ RouterDelayProfiler::RouterDelayProfiler(
           g_vpr_ctx.device().grid,
           *lookahead,
           g_vpr_ctx.device().rr_nodes,
+          &g_vpr_ctx.mutable_device().rr_graph,
```
vaughnbetz, line 20: .device() instead of .mutable_device() should be possible.

vpr/src/route/connection_router.h
```diff
@@ -246,6 +248,7 @@ class ConnectionRouter : public ConnectionRouterInterface {
     const DeviceGrid& grid_;
     const RouterLookahead& router_lookahead_;
     const t_rr_graph_view rr_nodes_;
+    RRGraphView* rr_graph_;
```
vaughnbetz, line 251: Think these should be const RRGraphView*

vpr/src/route/connection_router.h
```diff
@@ -264,6 +267,7 @@ std::unique_ptr<ConnectionRouterInterface> make_connection_router(
     const DeviceGrid& grid,
     const RouterLookahead& router_lookahead,
     const t_rr_graph_storage& rr_nodes,
+    RRGraphView* rr_graph,
```
vaughnbetz, line 270: Think these should be const RRGraphView*

vpr/src/route/connection_router.h
```diff
@@ -28,12 +28,14 @@ class ConnectionRouter : public ConnectionRouterInterface {
         const DeviceGrid& grid,
         const RouterLookahead& router_lookahead,
         const t_rr_graph_storage& rr_nodes,
+        RRGraphView* rr_graph,
```
vaughnbetz, line 31: const RRGraphView*

vpr/src/route/connection_router.h
```diff
@@ -28,12 +28,14 @@ class ConnectionRouter : public ConnectionRouterInterface {
...
         const std::vector<t_rr_rc_data>& rr_rc_data,
         const std::vector<t_rr_switch_inf>& rr_switch_inf,
         std::vector<t_rr_node_route_inf>& rr_node_route_inf)
         : grid_(grid)
         , router_lookahead_(router_lookahead)
         , rr_nodes_(rr_nodes.view())
+        , rr_graph_(rr_graph)
```
vaughnbetz, line 38: const RRGraphView*

ethanroj23:
@nelsobe

PR #1818, opened 2021-08-03
Add tbb suppression

sfkhalid:
#### Description
When sanitizers are turned on, some known but harmless errors from the tbb library come up while running vpr. On this pull request, the tbb library was added to the existing leak sanitizer suppression file as well as a new address sanitizer suppression file to suppress those known errors.

The suppressions are currently working for the leak sanitizer errors. However, some errors from the address sanitizers are still showing in the vpr log files when running with sanitizers.


vaughnbetz:
@sfkhalid: can you write down exactly what output you get from the sanitizers when tbb is on after your changes, and exactly how you ran them?  Are you setting the environment variable:
ASAN_OPTIONS=suppressions=vpr/vpr.supp  
(or some path like that)?  Can you paste in the environment variable?

The documentation I've found on address sanitizer suppression environment variables is for clang; I assume g++ uses the same environment variable but haven't been able to confirm.

sfkhalid:
This is the log file that shows the remaining address sanitizer errors that still come up:
[log_with_add_sanitizer_errors.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/6983642/log_with_add_sanitizer_errors.txt)

sfkhalid:
> @sfkhalid: can you write down exactly what output you get from the sanitizers when tbb is on after your changes, and exactly how you ran them? Are you setting the environment variable:
> ASAN_OPTIONS=suppressions=vpr/vpr.supp
> (or some path like that)? Can you paste in the environment variable?
> 
> The documentation I've found on address sanitizer suppression environment variables is for clang; I assume g++ uses the same environment variable but haven't been able to confirm.

Yes, I set the environment variable using the following command:
ASAN_OPTIONS=suppressions=vpr/asan.supp/
where asan.supp was the file I added with the tbb suppressions. I echoed back the ASAN_OPTIONS environment variable to make sure it was set properly and it was.

This makes me think that the way I am writing in the tbb suppression in asan.supp is not correct somehow, although I tried a few variations to try to get it to work.

vaughnbetz:
1.  Can you try a few path variations?
  - give the absolute path to the suppression file
  - delete the final / at after asan.supp/ (it's unusual to have a / at the end of a filename)
2. Can you put a syntax error (just a random string) at the start of the .supp file to see if the sanitizers then complain about it?  That would let you know if they read it or not.
3. Is there any output from vpr when compiled with sanitizers that say if it has read the .supp file?

vpr/asan.supp
```diff
@@ -0,0 +1,5 @@
+#AddressSanitizer suppression file for VPR
+
+#Leaks from TBB library, which
```
vaughnbetz, line 3: Should update this comment to say you are disabling errors from TBB as there are some errors flagged by the sanitizers that we can't fix and suspect are spurious.
The current comment is incorrect (talks about leaks).

sfkhalid:
> 1. Can you try a few path variations?
> 
> * give the absolute path to the suppression file
> * delete the final / at after asan.supp/ (it's unusual to have a / at the end of a filename)
> 
> 1. Can you put a syntax error (just a random string) at the start of the .supp file to see if the sanitizers then complain about it?  That would let you know if they read it or not.
> 2. Is there any output from vpr when compiled with sanitizers that say if it has read the .supp file?

I tried these things out and it turned out that vpr was not reading the asan.supp file. To fix this, I added the ASAN_OPTIONS environment variable to a couple of scripts, the same way LSAN_OPTIONS was added. 

Now I can confirm that the file is being read, because if I put in a random string at the beginning I get errors from that. However, it is still not suppressing the TBB errors, even though I specified the TBB library in asan.supp. I also tried suppressing ubsan just to see what would happen (as this library comes up in the errors) and suppressing individual functions and currently haven't seen those methods work either.

PR #1819, opened 2021-08-04
RRGraphView node_xlow(), node_xhigh(), node_ylow(), node_yhigh() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::node_xlow()`,  `RRGraphView::node_xhigh()`, `RRGraphView::node_ylow()` and `RRGraphView::node_yhigh()` throughout VTR. Every time `device_ctx.rr_nodes[index].xlow()` was called has been replaced with `rr_graph.node_xlow(RRNodeId(index))`. In order to do this, I followed a pattern similar to that in the last PR.

### Motivation and Context

### How Has This Been Tested?

I have run the regression tests for `vtr_reg_strong` along with the QoR testing found at  `$ ../scripts/run_vtr_task.py regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain`. Results from these QoR tests can be found below. The file containing all results can be found [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7077360/node_xlow_comparison.xlsx)

|---|---|---|
|vtr_flow_elapsed_time	|1|	0.996933617|
|odin_synth_time	|1|	1.014076279|
|abc_synth_time	|1|	0.992281017|
|max_vpr_mem	|1|	1.000259087|
|pack_time	|1|	0.992229957|
|place_time	|1|	0.98888593|
|min_chan_width_route_time	|1|	0.99801215|
|crit_path_route_time	|1|	0.995672431|

## Uses of xlow, xhigh, ylow, yhigh

As there may be cases where new API functions (such as `node_length()`) could be implemented, I have included a list of calls to xlow, xhigh, ylow, yhigh and what the purpose of each call was.

Of the functions listed here, `node_length`, `node_is_initialized`, and `nodes_are_adjacent` have all been implemented in this PR.

| Function | Use Scenario | One Usage (Location) | Count |
| --       | --           | --       | -- |
| xlow | retrieve physical tile from device_ctx.grid | test_fasm.cpp:193 | 16 |
| xlow | Obtain length of wire | stats.cpp:313 | 3 |
| xlow | Determine switchpoint | draw.cpp:1475 | 1 |
| xlow | Draw from one edge to the other | draw.cpp:1934 | 1 |
| xlow | Create bounding box | draw.cpp:2083 | 1 |
| xlow | Check if nodes are adjacent | check_route.cpp:481 | 1 |
| xlow | Check if node is initialized | check_route.cpp:64 | 3 |
| xlow | Include x value in console output | check_route.cpp:258 | 5 |
| xlow | Create a serial num for the wire  x value in console output | check_route.cpp:198 | 3 |
| xlow | Count cblocks | rr_graph_area.cpp:229 | 3 |
| xlow | Return the segment number (distance along the channel) of the connection box from from_rr_type (CHANX or CHANY) to to_node (IPIN) | rr_graph_util.cpp:21 | 1 |
| xlow | Check direction going from one node to another| rr_graph_util.cpp:45 | 1 |
| xlow | Write Coordinates | VprTimingGraphResolver.cpp:345 | 1 |

#### Types of changes

#### Checklist:


ethanroj23:
@vaughnbetz Should I be worried about the Yosys+Odin (Continuous) and Yosys+Odin (Presubmit) failing? I remember that the Yosys+ODIN-II Strong Tests were expected to fail, but not the others...

vaughnbetz:
All the yosys+odin tests will fail until we merge the PR enabling that.  The one test was split into multiple ones, which is why you're seeing multiple failures.  Not a concern.

ethanroj23:
> All the yosys+odin tests will fail until we merge the PR enabling that. The one test was split into multiple ones, which is why you're seeing multiple failures. Not a concern.

Ok that makes sense. Thanks!

vpr/src/route/route_common.cpp
```diff
@@ -1147,27 +1147,27 @@ t_bb load_net_route_bb(ClusterNetId net_id, int bb_factor) {
     const t_rr_node& source_node = device_ctx.rr_nodes[driver_rr];
     VTR_ASSERT(rr_graph.node_type(RRNodeId(driver_rr)) == SOURCE);
 
-    VTR_ASSERT(source_node.xlow() <= source_node.xhigh());
-    VTR_ASSERT(source_node.ylow() <= source_node.yhigh());
+    VTR_ASSERT(rr_graph.node_xlow(source_node.id()) <= rr_graph.node_xhigh(source_node.id()));
```
tangxifan, line 1150: Should we use ``RRNodeId(driver_rr)`` here because it will save our effort later down the road.
The data type in ``net_rr_terminals`` will adopt ``RRNodeId``.

vpr/src/route/route_breadth_first.cpp
```diff
@@ -384,15 +384,16 @@ static void breadth_first_expand_neighbours(BinaryHeap& heap, int inode, float p
...
     auto& route_ctx = g_vpr_ctx.routing();
 
     for (RREdgeId from_edge : device_ctx.rr_nodes.edge_range(RRNodeId(inode))) {
         RRNodeId to_node = device_ctx.rr_nodes.edge_sink_node(from_edge);
 
-        if (device_ctx.rr_nodes.node_xhigh(to_node) < route_ctx.route_bb[net_id].xmin
-            || device_ctx.rr_nodes.node_xlow(to_node) > route_ctx.route_bb[net_id].xmax
-            || device_ctx.rr_nodes.node_yhigh(to_node) < route_ctx.route_bb[net_id].ymin
-            || device_ctx.rr_nodes.node_ylow(to_node) > route_ctx.route_bb[net_id].ymax)
+        if (rr_graph.node_xhigh(to_node) < route_ctx.route_bb[net_id].xmin
```
tangxifan, line 393: This may be an API we want to. Given a bounding box (you can refer to ``vtr::Rect<short>``), we check if the coordinates of a routing resource node is inside or not. This may be used by routers frequently.

vpr/src/route/route_common.cpp
```diff
@@ -1147,27 +1147,27 @@ t_bb load_net_route_bb(ClusterNetId net_id, int bb_factor) {
...
 
     auto net_sinks = cluster_ctx.clb_nlist.net_sinks(net_id);
     for (size_t ipin = 1; ipin < net_sinks.size() + 1; ++ipin) { //Start at 1 since looping through sinks
         int sink_rr = route_ctx.net_rr_terminals[net_id][ipin];
         const t_rr_node& sink_node = device_ctx.rr_nodes[sink_rr];
         VTR_ASSERT(rr_graph.node_type(RRNodeId(sink_rr)) == SINK);
 
-        VTR_ASSERT(sink_node.xlow() <= sink_node.xhigh());
-        VTR_ASSERT(sink_node.ylow() <= sink_node.yhigh());
+        VTR_ASSERT(rr_graph.node_xlow(sink_node.id()) <= rr_graph.node_xhigh(sink_node.id()));
```
tangxifan, line 1164: Similar to the ``source_node.id()`` comment.

vpr/src/route/route_timing.cpp
```diff
@@ -1671,8 +1671,8 @@ static size_t calculate_wirelength_available() {
     for (size_t i = 0; i < device_ctx.rr_nodes.size(); ++i) {
         const t_rr_type channel_type = rr_graph.node_type(RRNodeId(i));
         if (channel_type == CHANX || channel_type == CHANY) {
-            size_t length_x = device_ctx.rr_nodes[i].xhigh() - device_ctx.rr_nodes[i].xlow();
-            size_t length_y = device_ctx.rr_nodes[i].yhigh() - device_ctx.rr_nodes[i].ylow();
+            size_t length_x = rr_graph.node_xhigh(RRNodeId(i)) - rr_graph.node_xlow(RRNodeId(i));
```
tangxifan, line 1674: Should we use the new API ``node_length()`` here?

vpr/src/route/route_util.cpp
```diff
@@ -31,19 +31,19 @@ vtr::Matrix<float> calculate_routing_usage(t_rr_type rr_type) {
 
         if (rr_type == CHANX) {
             VTR_ASSERT(rr_graph.node_type(RRNodeId(inode)) == CHANX);
-            VTR_ASSERT(rr_node.ylow() == rr_node.yhigh());
+            VTR_ASSERT(rr_graph.node_ylow(rr_node.id()) == rr_graph.node_yhigh(rr_node.id()));
```
tangxifan, line 34: We can use ``RRNodeId(inode)`` which is easier for future refactoring.

vpr/src/route/rr_graph.cpp
```diff
@@ -2937,8 +2937,8 @@ static RRNodeId pick_best_direct_connect_target_rr_node(const t_rr_graph_storage
 
         for (RRNodeId to_rr : candidate_rr_nodes) {
             VTR_ASSERT(rr_graph.node_type(to_rr) == IPIN);
-            float to_dist = std::abs(rr_nodes.node_xlow(from_rr) - rr_nodes.node_xlow(to_rr))
-                            + std::abs(rr_nodes.node_ylow(from_rr) - rr_nodes.node_ylow(to_rr));
+            float to_dist = std::abs(rr_graph.node_xlow(from_rr) - rr_graph.node_xlow(to_rr))
```
tangxifan, line 2940: Consider to use the newly added API ``node_length()``

vpr/src/route/rr_graph2.cpp
```diff
@@ -1203,53 +1203,53 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
...
+                        if (y != rr_graph.node_ylow(rr_node.id())) {
                             VPR_ERROR(VPR_ERROR_ROUTE, "RR node y position does not agree between rr_nodes (%d) and rr_node_indices (%d): %s",
-                                      rr_node.ylow(),
+                                      rr_graph.node_ylow(rr_node.id()),
                                       y,
                                       describe_rr_node(size_t(inode)).c_str());
                         }
 
-                        if (x < rr_node.xlow() || x > rr_node.xhigh()) {
+                        if (x < rr_graph.node_xlow(rr_node.id()) || x > rr_graph.node_xhigh(rr_node.id())) {
```
tangxifan, line 1215: This could be an API which judges if a given ``x`` is in range, since it is frequently used.

vpr/src/route/rr_graph2.cpp
```diff
@@ -1203,53 +1203,53 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
...
+                        if (x != rr_graph.node_xlow(rr_node.id())) {
                             VPR_ERROR(VPR_ERROR_ROUTE, "RR node x position does not agree between rr_nodes (%d) and rr_node_indices (%d): %s",
-                                      rr_node.xlow(),
+                                      rr_graph.node_xlow(rr_node.id()),
                                       x,
                                       describe_rr_node(size_t(inode)).c_str());
                         }
 
-                        if (y < rr_node.ylow() || y > rr_node.yhigh()) {
+                        if (y < rr_graph.node_ylow(rr_node.id()) || y > rr_graph.node_yhigh(rr_node.id())) {
```
tangxifan, line 1232: This could be an API which judges if a given ``y`` is in range, since it is frequently used.

vpr/src/route/rr_graph2.cpp
```diff
@@ -1305,8 +1305,8 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
         auto& rr_node = rr_nodes[size_t(inode)];
 
         if (rr_graph.node_type(inode) == SOURCE || rr_graph.node_type(inode) == SINK) {
-            int rr_width = (rr_node.xhigh() - rr_node.xlow() + 1);
-            int rr_height = (rr_node.yhigh() - rr_node.ylow() + 1);
+            int rr_width = (rr_graph.node_xhigh(rr_node.id()) - rr_graph.node_xlow(rr_node.id()) + 1);
```
tangxifan, line 1308: Consider to use the newly added API node_length()

vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -615,16 +615,20 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         return node.ptc_num();
     }
     inline int get_node_loc_xhigh(const t_rr_node& node) final {
-        return node.xhigh();
+        const auto& rr_graph = (*rr_graph_);
```
tangxifan, line 618: We can simplify it to one line, e.g.,
```
return rr_graph_->node_xhigh(node.id());
```

vpr/src/timing/VprTimingGraphResolver.cpp
```diff
@@ -324,18 +324,18 @@ void VprTimingGraphResolver::get_detailed_interconnect_components_helper(std::ve
 
                 arrow = "->"; //we will point the coordinates from start to finish, left to right
 
-                if (rr_graph.node_direction(RRNodeId(node->inode)) == Direction::DEC) {              //signal travels along decreasing direction
-                    start_x = " (" + std::to_string(device_ctx.rr_nodes[node->inode].xhigh()) + ","; //start coordinates have large value
-                    start_y = std::to_string(device_ctx.rr_nodes[node->inode].yhigh()) + ")";
-                    end_x = "(" + std::to_string(device_ctx.rr_nodes[node->inode].xlow()) + ","; //end coordinates have smaller value
-                    end_y = std::to_string(device_ctx.rr_nodes[node->inode].ylow()) + ")";
+                if (rr_graph.node_direction(RRNodeId(node->inode)) == Direction::DEC) {                //signal travels along decreasing direction
```
tangxifan, line 327: We may consider to have an API ``node_coordinate_to_string()`` which prints the string for a given node (depending on the direction) in this format.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,60 @@ class RRGraphView {
...
+    inline short node_ylow(RRNodeId node) const {
+        return node_storage_.node_ylow(node);
+    }
+
+    /* Get the yhigh of a routing resource node. This function is inlined for runtime optimization. */
+    inline short node_yhigh(RRNodeId node) const {
+        return node_storage_.node_yhigh(node);
+    }
+
+    /* Get the length of a routing resource node. This function is inlined for runtime optimization. */
```
tangxifan, line 119: I think the ``node_length()`` API may need a detailed comment. 
- The node length is only applicable to nodes that represent routing tracks. So you need to add
   ``VTR_ASSERT(node_type() == CHANX  || node_type() == CHANY);``
  This also has to be commented so that developers know when to use the API.
- We need to ensure if the ``node_length()`` can be negative or not. Can you quickly check the use of ``set_node_xhigh()``,  ``set_node_yhigh()``, ``set_node_xlow()`` and ``set_node_ylow()``, to see if the ``xhigh/yhigh`` is always larger than ``xlow/ylow``.  I know this is definitely true for any ``INC`` directional wires. But I forgot if the ``DEC`` wires apply the same rule or not. If so, we may consider to add a comment ``VTR_ASSERT_SAFE`` to ensure that the node_length() is always valid. You may also state in the comment that the node_length() is always a positive number.
- I see some codes use the ``node_length() - 1``. For me, the current ``node_length()`` definition is o.k. which shows ``1`` for a length-1 wire. Please double check the definition and see if we need more API or not.
- You may consider to add a figure in ASICII format to illustrate the idea.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,60 @@ class RRGraphView {
...
+    inline int node_length(RRNodeId node) const {
+        return 1 + node_xhigh(node) - node_xlow(node) + node_yhigh(node) - node_ylow(node);
+    }
+
+    /* Check if routing resource node is initialized. This function is inlined for runtime optimization. */
+    inline bool node_is_initialized(RRNodeId node) const {
```
tangxifan, line 125: This API is confusing, which could be misleading. 
- If we want to check a node is initialized or not, it should be limited to the ``SOURCE`` type. If a node is uninitialized, it is more nature to be set its type to ``NUM_RR_TYPE``, rather than ``SOURCE``.
- The all-0 ``xlow/ylow/xhigh/yhigh`` do exist, for example, a PLL may locate at ``(0, 0)``, which definitely has a SOURCE node which is all-0 in coordinates. As for me, the uninitialized coordinates should all ``-1``.

Therefore, this API is quite confusing to me. See if @vaughnbetz has more insights on this.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,60 @@ class RRGraphView {
...
+        return !((node_type(node) == SOURCE)
+                 && (node_xlow(node) == 0) && (node_ylow(node) == 0)
+                 && (node_xhigh(node) == 0) && (node_yhigh(node) == 0));
+    }
+
+    /* Check if two routing resource nodes are adjacent. This function is inlined for runtime optimization. */
+    inline bool nodes_are_adjacent(RRNodeId first_node, RRNodeId second_node) const {
```
tangxifan, line 132: - The definition of adjacent should be well explained here as a code comment. When developers read this API, they know when to use it and how to evaluate the return value
- You may not need to use local variables in this inline function, you can make it much compact.

tangxifan (changes requested):
@ethanroj23 Most changes look good. Raise some concerns about the newly added APIs. Should be addressable. If you are confused by any comment, please feel free to tell me.

vpr/src/route/rr_graph.cpp
ethanroj23: I am not sure if `node_length()` applies here since this is dealing with two distinct nodes (`from_rr`, `to_rr`) rather than one. Am I missing something?

vpr/src/route/rr_graph2.cpp
ethanroj23: I am also unsure if this instance works for `node_length()`. The `rr_width` and `rr_height` each are only equivalent to `node_length()` in certain cases. (eg. `rr_width == node_length` when the node type is `CHANX`

vpr/src/timing/VprTimingGraphResolver.cpp
ethanroj23: This can definitely be done. @vaughnbetz what are your thoughts on adding this API?

vpr/src/route/rr_graph.cpp
tangxifan: @ethanroj23 My bad. The node_length() is not applicable here. Your changes are good.

vpr/src/route/rr_graph2.cpp
tangxifan: Indeed, the ``node_length()`` should be only applicable to ``CHANX`` and ``CHANY``. 
I double checked the coordinates of ``SOURCE`` and ``SINK`` nodes. They do have a different xlow/ylow/xhigh/yhigh for each node.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/eea0df05c079767cb30f41b5160de17ffedf37dd/vpr/src/route/rr_graph.cpp#L1430-L1432

Therefore, we may consider an API ``node_bounding_box_area()``. But I am not sure such API is frequently needed by router or placer. We can ask @vaughnbetz for advice.

If not frequently used, we can keep the current code changes.

vpr/src/timing/VprTimingGraphResolver.cpp
vaughnbetz: I like it (as we discussed today).  Use it at least in the VprTimingGraphResolver, and there may be other uses (error messages,, overused rrnodes when we fail to route etc.) that could be captured now, and likely more in the future.
So I'd add it, and do a quick search to see if you can find routines that should use it now instead of printing ad-hoc RRnode identifiers.

vpr/src/base/read_route.cpp
```diff
@@ -266,7 +267,7 @@ static void process_nodes(std::ifstream& fp, ClusterNetId inet, const char* file
 
             if (tokens[4] == "to") {
                 format_coordinates(x2, y2, tokens[5], inet, filename, lineno);
-                if (node.xlow() != x || node.xhigh() != x2 || node.yhigh() != y2 || node.ylow() != y) {
+                if (rr_graph.node_xlow(node.id()) != x || rr_graph.node_xhigh(node.id()) != x2 || rr_graph.node_yhigh(node.id()) != y2 || rr_graph.node_ylow(node.id()) != y) {
```
vaughnbetz, line 270: Could shorten this code a little by doing:
auto inode = node.id();
and then using inode in the various calls afterward.  I probably have some preference for that, but it isn't critical if you disagree.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,72 @@ class RRGraphView {
         return node_storage_.fan_in(node);
     }
 
+    /* Get the xlow of a routing resource node. This function is inlined for runtime optimization. */
```
vaughnbetz, line 99: 1. It would be good to briefly explain what xlow is (minimum x-coordinate) is here.  Same for the other four coordinates.
2. Make these comments Doxygen-compatible, so we can document this API in the VTR readthedocs documentation.  That could be done in a separate PR though.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,72 @@ class RRGraphView {
...
+        return node_storage_.node_ylow(node);
+    }
+
+    /* Get the yhigh of a routing resource node. This function is inlined for runtime optimization. */
+    inline short node_yhigh(RRNodeId node) const {
+        return node_storage_.node_yhigh(node);
+    }
+
+    /* Get the length of a routing resource node. This function is inlined for runtime optimization. */
+    /* node_length() only applies to CHANX or CHANY and is always a positive number*/
```
vaughnbetz, line 120: Explain what it is -- the number of grid tile units spanned by the wire, including the endpoints.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,72 @@ class RRGraphView {
...
+
+    /* Check if routing resource node is initialized. This function is inlined for runtime optimization. */
+    inline bool node_is_initialized(RRNodeId node) const {
+        return !((node_type(node) == NUM_RR_TYPES)
+                 && (node_xlow(node) == -1) && (node_ylow(node) == -1)
+                 && (node_xhigh(node) == -1) && (node_yhigh(node) == -1));
+    }
+
+    /* Check if two routing resource nodes are adjacent (must be a CHANX and a CHANY). This function is inlined for runtime optimization. */
+    inline bool nodes_are_adjacent(RRNodeId chanx_node, RRNodeId chany_node) const {
```
vaughnbetz, line 136: Expand the comment to say this routine is used for error checking; it checks if two nodes are physically adjacent (could be connected) based on their geometry.  It does not check the routing edges to see if there are in fact possible to connect in the current routing graph.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,72 @@ class RRGraphView {
...
+        if (node_ylow(chany_node) > node_ylow(chanx_node) + 1 || // verifies that chany_node is not more than one unit above chanx_node
+            node_yhigh(chany_node) < node_ylow(chanx_node))      // verifies that chany_node is not more than one unit beneath chanx_node
+            return false;
+        if (node_xlow(chanx_node) > node_xlow(chany_node) + 1 || // verifies that chany_node is not more than one unit to the left of chanx_node
+            node_xhigh(chanx_node) < node_xlow(chany_node))      // verifies that chany_node is not more than one unit to the right of chanx_node
+            return false;
+        return true;
+    }
+
+    /* Check if node is within bounding box. This function is inlined for runtime optimization. */
```
vaughnbetz, line 147: Expand the comment. To return true, the RRNode must be completely contained within the specified bounding box, with the edges of the bounding box being inclusive.

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,72 @@ class RRGraphView {
...
+    inline bool node_is_inside_bounding_box(RRNodeId node, vtr::Rect<int> bounding_box) const {
+        return (node_xhigh(node) <= bounding_box.xmax()
+                && node_xlow(node) >= bounding_box.xmin()
+                && node_yhigh(node) <= bounding_box.ymax()
+                && node_ylow(node) >= bounding_box.ymin());
+    }
+
+    /* Check if x is within node x range. This function is inlined for runtime optimization. */
```
vaughnbetz, line 155: I'd expand this comment. What does within node range mean?  (It means "within the x-range spanned by the node, inclusive of its endpoints").

vpr/src/device/rr_graph_view.h
```diff
@@ -96,6 +96,72 @@ class RRGraphView {
...
+    inline bool x_in_node_range(int x, RRNodeId node) const {
+        return !(x < node_xlow(node) || x > node_xhigh(node));
+    }
+
+    /* Check if y is within node y range. This function is inlined for runtime optimization. */
```
vaughnbetz, line 160: Again, expand comment as in the x directed one.

vpr/src/draw/draw.cpp
```diff
@@ -1448,17 +1448,17 @@ static void draw_rr_chan(int inode, const ezgl::color color, ezgl::renderer* g)
     int coord_min = -1;
     int coord_max = -1;
     if (type == CHANX) {
-        coord_min = device_ctx.rr_nodes[inode].xlow();
-        coord_max = device_ctx.rr_nodes[inode].xhigh();
+        coord_min = rr_graph.node_xlow(RRNodeId(inode));
```
vaughnbetz, line 1451: Could we declare inode to be or type RRNodeId and get rid of a bunch of these casts?

vpr/src/draw/draw.cpp
```diff
@@ -1899,10 +1899,10 @@ static void draw_chanx_to_chanx_edge(int from_node, int to_node, int to_track, s
     y1 = from_chan.bottom();
     y2 = to_chan.bottom();
 
-    from_xlow = device_ctx.rr_nodes[from_node].xlow();
-    from_xhigh = device_ctx.rr_nodes[from_node].xhigh();
-    to_xlow = device_ctx.rr_nodes[to_node].xlow();
-    to_xhigh = device_ctx.rr_nodes[to_node].xhigh();
+    from_xlow = rr_graph.node_xlow(RRNodeId(from_node));
```
vaughnbetz, line 1902: Same comment on from_node, to_node etc. -- could we just declare them to be of type RRNodeId and reduce casting?

vpr/src/draw/draw.cpp
```diff
@@ -2161,9 +2161,10 @@ void draw_get_rr_pin_coords(const t_rr_node& node, float* xcen, float* ycen, con
     float offset, xc, yc, step;
     t_physical_tile_type_ptr type;
     auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
 
-    i = node.xlow();
-    j = node.ylow();
+    i = rr_graph.node_xlow(node.id());
```
vaughnbetz, line 2166: I think we could shorten the code with
auto inode = node.id();
and then using inode below.

vpr/src/route/check_route.cpp
```diff
@@ -211,8 +211,8 @@ static void check_source(int inode, ClusterNetId net_id) {
                         "in check_source: net %d begins with a node of type %d.\n", size_t(net_id), rr_type);
     }
```
vaughnbetz, line 213: Can we change the type of inode to RRNodeID and get rid of a bunch of the casting here?
Same for from_node and to_node.

vpr/src/route/check_route.cpp
```diff
@@ -472,26 +472,14 @@ static int chanx_chany_adjacent(int chanx_node, int chany_node) {
     /* Returns 1 if the specified CHANX and CHANY nodes are adjacent, 0         *
```
vaughnbetz, line 472: Separate enhancement, but it would be good if you check if this routine should really return a bool (and the calling code should check that).

vpr/src/route/check_rr_graph.cpp
```diff
@@ -216,7 +214,7 @@ void check_rr_graph(const t_graph_type graph_type,
                  */
                 bool is_chain = false;
                 if (rr_type == IPIN) {
-                    t_physical_tile_type_ptr type = device_ctx.grid[device_ctx.rr_nodes[inode].xlow()][device_ctx.rr_nodes[inode].ylow()].type;
+                    t_physical_tile_type_ptr type = device_ctx.grid[rr_graph.node_xlow(RRNodeId(inode))][rr_graph.node_ylow(RRNodeId(inode))].type;
```
vaughnbetz, line 217: Can we change the type of inode to RRNodeId and get rid of a bunch of casting?

vpr/src/route/check_rr_graph.cpp
```diff
@@ -300,10 +298,10 @@ void check_rr_node(int inode, enum e_route_type route_type, const DeviceContext&
     const auto& rr_graph = device_ctx.rr_graph;
 
     rr_type = rr_graph.node_type(RRNodeId(inode));
-    xlow = device_ctx.rr_nodes[inode].xlow();
-    xhigh = device_ctx.rr_nodes[inode].xhigh();
-    ylow = device_ctx.rr_nodes[inode].ylow();
-    yhigh = device_ctx.rr_nodes[inode].yhigh();
+    xlow = rr_graph.node_xlow(RRNodeId(inode));
```
vaughnbetz, line 301: Can we change inode to type RRNodeId and remove casting?

vpr/src/route/overuse_report.cpp
```diff
@@ -197,10 +198,10 @@ static void report_overused_chanx_chany(std::ostream& os, RRNodeId node_id) {
     //CHANX/CHANY rr nodes span across several grid locations.
     //Need to print out their starting and ending locations.
     os << "Grid location: " << '\n';
-    os << "Xlow = " << device_ctx.rr_nodes.node_xlow(node_id) << ", ";
-    os << "Ylow = " << device_ctx.rr_nodes.node_ylow(node_id) << '\n';
-    os << "Xhigh = " << device_ctx.rr_nodes.node_xhigh(node_id) << ", ";
-    os << "Yhigh = " << device_ctx.rr_nodes.node_yhigh(node_id) << '\n';
+    os << "Xlow = " << rr_graph.node_xlow(node_id) << ", ";
```
vaughnbetz, line 201: This routine could probably use your new RRNode to string function.  Could still print extra info if the string doesn't give everything.

vpr/src/route/rr_graph.cpp
```diff
@@ -2432,9 +2432,9 @@ std::string describe_rr_node(int inode) {
 
     msg += vtr::string_fmt(" type: %s", rr_node.type_string());
 
-    msg += vtr::string_fmt(" location: (%d,%d)", rr_node.xlow(), rr_node.ylow());
-    if (rr_node.xlow() != rr_node.xhigh() || rr_node.ylow() != rr_node.yhigh()) {
-        msg += vtr::string_fmt(" <-> (%d,%d)", rr_node.xhigh(), rr_node.yhigh());
+    msg += vtr::string_fmt(" location: (%d,%d)", rr_graph.node_xlow(rr_node.id()), rr_graph.node_ylow(rr_node.id()));
```
vaughnbetz, line 2435: Better to move to RRGraphView / use that new function.

vpr/src/route/rr_graph2.cpp
```diff
@@ -1203,53 +1203,53 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
                     }
 
                     if (rr_graph.node_type(inode) == CHANX) {
-                        VTR_ASSERT_MSG(rr_node.ylow() == rr_node.yhigh(), "CHANX should be horizontal");
+                        VTR_ASSERT_MSG(rr_graph.node_ylow(rr_node.id()) == rr_graph.node_yhigh(rr_node.id()), "CHANX should be horizontal");
```
vaughnbetz, line 1206: Could use inode (and make it of type RRNodeId) to shorten the following lines, I think.  rr_node is just the rr_node associated with id inode, so this is a bit of a roundabout way to get back to inode.

vpr/src/route/rr_graph_area.cpp
```diff
@@ -226,24 +224,24 @@ void count_bidir_routing_transistors(int num_switch, int wire_to_ipin_switch, fl
                 /* Now add in the shared buffer transistors, and reset some flags. */
 
                 if (from_rr_type == CHANX) {
-                    for (i = device_ctx.rr_nodes[from_node].xlow() - 1;
-                         i <= device_ctx.rr_nodes[from_node].xhigh(); i++) {
+                    for (i = rr_graph.node_xlow(RRNodeId(from_node)) - 1;
```
vaughnbetz, line 227: Changing the declared type of from_node and to_node to RRNodeId may let you get rid of some casting.

vaughnbetz (changes requested):
Looks good, but I have some changes to consider.

vpr/src/route/check_route.cpp
ethanroj23: I believe this is correct. The code that uses this function is counting the number of adjacent nodes, so it wants a 1 or 0.

`num_adj += chanx_chany_adjacent(from_node, to_node);`

vaughnbetz (approved):

vaughnbetz:
Looks good; thanks.

PR #1821, opened 2021-08-05
[Infra]: fix run_vtr_task lint error

sdamghan:

#### Description

specifying the `common_task_prefix` variable type while initializing it

#### Motivation and Context

fix CI error for python-format

#### Types of changes

#### Checklist:

### DIFF
--- a/vtr_flow/scripts/run_vtr_task.py
+++ b/vtr_flow/scripts/run_vtr_task.py
@@ -227,11 +227,11 @@ def vtr_command_main(arg_list, prog=None):
 
         config_files = [find_task_config_file(task_name) for task_name in task_names]
         configs = []
-        common_task_prefix = None  # common task prefix to shorten task names
+        common_task_prefix = ""  # common task prefix to shorten task names
         for config_file in config_files:
             config = load_task_config(config_file)
             configs += [config]
-            if common_task_prefix is None:
+            if not common_task_prefix:
                 common_task_prefix = config.task_name
             else:
                 match = SequenceMatcher(


vtr_flow/scripts/run_vtr_task.py
```diff
@@ -227,7 +227,7 @@ def vtr_command_main(arg_list, prog=None):
 
         config_files = [find_task_config_file(task_name) for task_name in task_names]
         configs = []
-        common_task_prefix = None  # common task prefix to shorten task names
+        common_task_prefix = ""  # common task prefix to shorten task names
```
vaughnbetz, line 230: It looks like this changes the functionality of the script.  Shouldn't common_task_prefix be initialized to config.task_name instead of ""?  And with that change, I think the if / else checking if common_task_prefix is None can be deleted (just always do the else).

vaughnbetz:
@sdamghan : I think the script functionality is changed by this PR, but there is an alternative change that wouldn't change it (and should simplify the logic).  Please take a look.

vtr_flow/scripts/run_vtr_task.py
sdamghan: I think the definition of `common_task_prefix` initializes by empty string needs to be outside of the for loop since the 'shorten_task_names' function uses this variable. By that, I mean the value of config.task_name is specified inside the loop so we can not initialize the `common_task_prefix` with the value of config.task_name outside of the loop. However, you are right about the if / else expression, the if expression should change to 'if not common_task_prefix' to follow the previous path. Sorry for this mistake

sdamghan: Please correct me if I am wrong

vaughnbetz: Ah, you're right -- didn't read that carefully enough.  Yes, I think the fix will work.

PR #1822, opened 2021-08-09
 Fixed Fc values for designs with different horizontal and vertical channel widths

heinerb1:
#### Description
Work in progress follow up to PR #1786 that added the ability for the user to specify detailed routing for designs with different horizontal and vertical channel widths. I modified the Fc section of the build_rr_graph function and the alloc_and_load_actual_fc function which determines the actual Fc value for a particular pin. Added code in the alloc_and_load_actual_fc function that uses pin location to provide the correct Fc value using the channel width and Fc percentages specified by the user. However, the code is currently split into two sections and uses the channel widths to determine which part to use. I am beginning to work on changing this so that only one set of code is used for both cases. In addition, more documentation will need to be added and the code will need to be reviewed to determine where it needs to be improved, so there is still a decent amount of work that needs to be done.

#### Related Issue
#1768 

#### Motivation and Context
Due to the changes in PR #1786, a user could now specify detailed routing for designs with different horizontal and vertical channel widths. However, due to assumptions in rr_graph that one channel would be used for the entire design, the Fc values were being computed incorrectly for the smaller of the two channel widths. In order to correct this, the rr_graph builder needed to be modified to determine which channel a pin was connecting to and calculate the correct Fc value for that pin.

As mentioned above, since the code is currently split into two sections and uses the channel widths to determine which part to use, the code currently passes all tests. I have tested the new pieces with various different designs that had different and vertical channel widths, and the results seem to be correct so far. However, as additional changes are made to this code such as the unification process, it might not pass the tests and need to be modified.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/rr_graph.cpp
+++ b/vpr/src/route/rr_graph.cpp
@@ -125,6 +125,10 @@ static void build_unidir_rr_opins(RRGraphBuilder& rr_graph_builder,
                                   const DeviceGrid& grid,
                                   const std::vector<vtr::Matrix<int>>& Fc_out,
                                   const int max_chan_width,
+                                  const int* sets_per_seg_type_x,
+                                  const int* sets_per_seg_type_y,
+                                  const t_chan_width& chan_width,
+                                  const enum e_directionality directionality,
                                   const t_chan_details& chan_details_x,
                                   const t_chan_details& chan_details_y,
                                   vtr::NdMatrix<int, 3>& Fc_xofs,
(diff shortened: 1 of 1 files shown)


heinerb1:
The unified code is currently failing just one QoR test with the following message:

regression_tests/vtr_reg_strong/strong_clock_buf...[Fail]
k6_frac_N10_mem32K_40nm_clk_buf.xml/multiclock_buf.blif/common crit_path_delay_mcw relative value 1.189550164181009 outside of range [0.9,1.1] and not equal to golden value: 1.48921

vaughnbetz:
That QoR failure looks spurious; small design, critical path change is just outside the tolerable range.  If everything else looks good, we can just widen the QoR range slightly for that test.

Restarted CI as some tests seem to have not run.

vaughnbetz:
@jgoeders @ArashAhmadian : I think we do have overlapping work here.  @ArashAhmadian can you update on where you stand?  I believe you are about to make a PR that goes further than this one in moving the x/y channel specifications to be independent, so this PR has likely become redundant.

PR #1827, opened 2021-08-13
fix track_memory_usage no use bug

laiyao1:
#### Expected Behaviour
the run_vtr_flow.py get a parameter track_memory_usage and it decides whether to track memory.

#### Current Behaviour
this parameter has not be inputed in vtr command

#### Possible Solution
input into vtr.CommandRunner

#### Steps to Reproduce

1. run run_vtr_flow.py --track_memory_usage false

### DIFF
--- a/vtr_flow/scripts/run_vtr_flow.py
+++ b/vtr_flow/scripts/run_vtr_flow.py
@@ -406,7 +406,7 @@ def vtr_command_main(arg_list, prog=None):
         temp_dir = Path(args.temp_dir)
     # Specify how command should be run
     command_runner = vtr.CommandRunner(
-        track_memory=True,
+        track_memory=args.track_memory_usage,
         max_memory_mb=args.limit_memory_usage,
         timeout_sec=args.timeout,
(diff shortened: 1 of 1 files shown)


vaughnbetz:
Thanks for the fix.  You should change the default for this option to true (it's set earlier in the file) as part of this change though, as usually we want to track memory usage.  If we don't change the default, I think a lot of our CI will lose memory tracking as we probably don't explicitly set it on the run_vtr_flow command line.

sfkhalid:
> @laiyao1 : a gentle nag to see if you can make this change.
> @sfkhalid : adding you in case not (this is a good infrastructure change to get in, but it needs the tweak above and has to be tested after that.

Made this change on a different branch - it is on PR #1838. I will take a look at the CI on that branch and make sure the memory tracking is on as well.

vaughnbetz:
Moved this code to PR #1838

vaughnbetz:
Closing this since we moved it to PR #1838 .  Thanks for the fix @laiyao1 !

PR #1836, opened 2021-08-31
[VPR] allowing for the specification of segment axis in arch file

ArashAhmadian:
#### Description
This is the first of multiple PRs that are working towards allowing VPR to model different horizontal and vertical channels. The implementation is planned to take a ground-up approach by starting with augmenting read_xml_arch_file.cpp to be able to handle this specification (this PR) and gradually changing the functions inside rr_graph.cpp & rr_graph2.cpp to get full functionality. 

NOTE: I changed the input parameters for the functions in rr_graph2.cpp to 1) Completely deprecate e_seg_details_type 2) Add more appropriate variable names for future PRs. (This doesn't change any functionality and was intended) 

#### Motivation and Context
To allow the modelling of different horizontal and vertical channels. 
I modified took `arch/timing/k6_N10_mem32K_40nm.xml` and added the `axis=x/y` option to it's segment(s). I then ran `vtr_reg_basic/basic_timing` with the aforementioned architecture. This did not change any routing metrics and the only change was was in various run-time metrics due to machine noise, which was expected. 

#### Types of changes

#### Checklist:

#### QoR Comparisons:
[results_diff_x_y.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7117236/results_diff_x_y.xlsx)


ArashAhmadian:
@vaughnbetz This PR is ready for code review.

vaughnbetz:
Expected tests passed except one Odin II basic one:
2021-09-01T07:30:15.5706671Z vtr_reg_multiclock/func_multiclock/once...[Fail]
2021-09-01T07:30:15.5707169Z [Fail]
2021-09-01T07:30:15.5709393Z k6_frac_N10_mem32K_40nm.xml/multiclock_reader_writer.v/common pack_time relative value 214.0 outside of range [0.1,10.0], above absolute threshold 2.0 and not equal to golden value: 0.01

Had a high pack time, which is unexpected; I set the jobs to rerun to see if it was spurious.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3799,6 +3824,11 @@ static void ProcessSegments(pugi::xml_node Parent,
         /* Get next Node */
         Node = Node.next_sibling(Node.name());
     }
+    /*if no segment with default axis option isn't provided, atleast 1 segment along each x & y axes is needed */
+    if (!both_axis_segs && !(x_axis_segs && y_axis_segs && true)) {
```
vaughnbetz, line 3828: I think you can simplify this logic, and use more clear variable naming.  This is an x_axis_segs --> x_axis_seg_found (and so on).  Get rid of both_axis_seg and just set x_axis_seg_found = TRUE if an x-segment is found, and both x_axis_seg_found and y_axis_seg_found = TRUE if you have a segment with neither x nor y set.

libs/libarchfpga/src/physical_types.h
```diff
@@ -1382,6 +1382,11 @@ enum e_directionality {
     UNI_DIRECTIONAL,
     BI_DIRECTIONAL
 };
+enum e_parallel_axis {
```
vaughnbetz, line 1385: Comment what this enum is for and what each value in it means.

libs/libarchfpga/src/physical_types.h
```diff
@@ -1432,8 +1437,10 @@ struct t_segment_inf {
     float Rmetal;
     float Cmetal;
     enum e_directionality directionality;
+    enum e_parallel_axis parallel_axis;
```
vaughnbetz, line 1440: Comment what this data member means (other members are commented above; this one should be too).

vaughnbetz (changes requested):
Looks mostly good, but I have a few suggested changes.

libs/libarchfpga/src/read_xml_arch_file.cpp
ArashAhmadian: Correct. I updated the logic accordingly.

libs/libarchfpga/src/physical_types.h
ArashAhmadian: Added comments accordingly.

libs/libarchfpga/src/physical_types.h
ArashAhmadian: Added comments accordingly.

ArashAhmadian:
@vaughnbetz CI looks to be good. Please take a look at the new comments.

libs/libarchfpga/src/physical_types.h
```diff
@@ -1382,6 +1382,14 @@ enum e_directionality {
     UNI_DIRECTIONAL,
     BI_DIRECTIONAL
 };
+/* X_AXIS: Assigned to a segment's  parallel_axis data member if segment is parallel to the x-axis. Defined in arch file as: axis="x"           *
+ * Y_AXIS: Assigned to a segment's  parallel_axis data amember if segment is parallel to the y-axis. Defined in arch file as: axis="y"          *     
+ * BOTH_AXIS: Assigned to a segment's parallel axis data member if segment's parallel axis is not specified (default is it parallel to both).   */
+enum e_parallel_axis {
+    X_AXIS,
```
vaughnbetz, line 1389: I think you can still explain this a bit more, in a less self-referential way.  
X_AXIS: data that describes an x-directed wire segment (CHANX)
Y_AXIS: data that describes a y-directed wire segment (CHANY)
BOTH_AXIS: data that can be applied to both x- and y-directed wire segments.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3639,6 +3639,12 @@ static void ProcessSegments(pugi::xml_node Parent,
 
     /* Load the segments. */
     Node = get_first_child(Parent, "segment", loc_data);
+
+    /*Keep track of the type of axis of the segments loaded*/
```
vaughnbetz, line 3643: Can improve this comment.  Flags to see if we have *any* x- and y-directed segment types specified, respectively.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3799,6 +3825,12 @@ static void ProcessSegments(pugi::xml_node Parent,
         /* Get next Node */
         Node = Node.next_sibling(Node.name());
     }
+    /*if no segment with default axis option isn't provided, atleast 1 segment along each x & y axes is needed */
```
vaughnbetz, line 3828: Can improve this comment.  Should say we need at least one type of segment that applies to each of x- and y-directed wiring.

vaughnbetz (changes requested):
Still room for improvement in the commenting; see my further suggestions to make the comments more precise and understandable by someone who doesn't know your terminology yet.

vaughnbetz:
To be careful, you should also do a QoR comparison on the VTR benchmarks (or the titan benchmarks but I think one or the other will be enough for this PR).  That will show QoR is essentially unchanged for this PR, which is a tighter check than the CI pass/fail.  See the VTR developer guide on the documentation web page for scripts on how to do that if needed.

ArashAhmadian:
@vaughnbetz I addressed your comments in the latest commit. I also added a QoR spread-sheet comparison for the VTR benchmarks. It's ready for you to review.

vaughnbetz:
Thanks; code and QoR look good. Merging.

PR #1838, opened 2021-09-02
fix_track_memory_usage

sfkhalid:
The change in this pull request is the same as PR #1827. Added the change to a new branch because I wasn't able to checkout the other remote branch as it was on a fork off of the main remote repo verilog-to-routing/vtr-verilog-to-routing.

The original change was to make sure the runner looks at the argument track_memory_usage to see whether memory tracking should be turned on or off, instead of just always making it on.
An additional change was made that was requested in a comment on #1827 - the default value of track_memory_usage was set to "True" so that track_memory_usage is always on unless the argument specifies that it should be off.

#### Types of changes

### DIFF
--- a/vtr_flow/scripts/run_vtr_flow.py
+++ b/vtr_flow/scripts/run_vtr_flow.py
@@ -184,7 +184,7 @@ def vtr_command_argparser(prog=None):
 
     house_keeping.add_argument(
         "-track_memory_usage",
-        default=False,
+        default=True,
         action="store_true",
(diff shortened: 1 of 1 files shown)


vaughnbetz:
Thanks @sfkhalid . Code looks good; will merge once CI goes green.

vaughnbetz:
Thanks; change looks good.
Some kokoro tests didn't run, and nightly_test_2 appears to have no failures yet is flagged as a fail. Re-launching kokoro.

PR #1841, opened 2021-09-09
Change place flow in init place

sfkhalid:
During initial placement, all macros were placed before all blocks. On this pull request, that flow was changed. Now, all blocks (whether part of a macro or not) are placed in the order that they were sorted in - blocks are sorted from most difficult to place to least difficult to place. 
Also, initial placement no longer uses the free_locations and legal_pos data structures to find potential locations for the blocks. Instead, it now uses the compressed grids to search for potential locations for each block. It tries up to a maximum number of randomly-chosen locations from within the block's floorplan region (the floorplan region is the whole chip if the block is not constrained) and after that turns to exhaustive placement if the block still has not been placed. 

#### Related Issue
Related to adding placement constraints to vpr #932


vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,214 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
 //Sort the blocks according to how difficult they are to place, prior to initial placement
 static std::vector<ClusterBlockId> sort_blocks(const vtr::vector<ClusterBlockId, t_block_score>& block_scores);
 
-//Sort the macros according to how difficult they are to place, prior to initial placement
-static std::vector<t_pl_macro> sort_macros(const vtr::vector<ClusterBlockId, t_block_score>& block_scores);
+//Get a legal placement position for the block
```
sfkhalid, line 54: improve this comment (look for one random legal position for a block that respects its region constraints)

Add comments for exhaustive placement routines as well

sfkhalid: describe what it returns (true/false based on whether it found a location) and the location by reference

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,214 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
 
-static int get_free_sub_tile(std::vector<std::vector<int>>& free_locations, int itype, std::vector<int> possible_sub_tiles) {
-    for (int sub_tile : possible_sub_tiles) {
-        if (free_locations[itype][sub_tile] > 0) {
-            return sub_tile;
+static void place_the_blocks(const std::vector<ClusterBlockId>& sorted_blocks,
+                             const vtr::vector<ClusterBlockId, t_block_score>& block_scores,
+                             enum e_pad_loc_type pad_loc_type);
+
+static bool get_legal_placement_loc(PartitionRegion& pr, t_pl_loc& loc, t_logical_block_type_ptr block_type) {
```
sfkhalid, line 67: make pr a const reference, put inputs first and then output

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,214 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+    const auto& compressed_block_grid = g_vpr_ctx.placement().compressed_block_grids[block_type->index];
+
+    bool legal;
+
+    int cx_from = -1;
```
sfkhalid, line 72: have some comments here explaining what you're doing

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,214 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+
+    int max_cx = grid_to_compressed_approx(compressed_block_grid.compressed_to_grid_x, rect.xmax());
+    int max_cy = grid_to_compressed_approx(compressed_block_grid.compressed_to_grid_y, rect.ymax());
+
+    int delta_cx = max_cx - min_cx;
+
+    int cx_to;
+    int cy_to;
+
+    legal = find_compatible_compressed_loc_in_range(block_type, min_cx, max_cx, min_cy, max_cy, delta_cx, cx_from, cy_from, cx_to, cy_to, false);
```
sfkhalid, line 100: better form to declare legal closer to where you will use it

vpr/src/place/initial_placement.cpp
```diff
@@ -405,39 +562,40 @@ static std::vector<ClusterBlockId> sort_blocks(const vtr::vector<ClusterBlockId,
...
-}
-
 void print_sorted_blocks(const std::vector<ClusterBlockId>& sorted_blocks, const vtr::vector<ClusterBlockId, t_block_score>& block_scores) {
     VTR_LOG("\nPrinting sorted blocks: \n");
     for (unsigned int i = 0; i < sorted_blocks.size(); i++) {
         VTR_LOG("Block_Id: %zu, Macro size: %d, Num tiles outside floorplan constraints: %d\n", sorted_blocks[i], block_scores[sorted_blocks[i]].macro_size, block_scores[sorted_blocks[i]].tiles_outside_of_floorplan_constraints);
     }
 }
 
+static void place_the_blocks(const std::vector<ClusterBlockId>& sorted_blocks,
```
sfkhalid, line 572: place_all_blocks

vpr/src/place/initial_placement.cpp
```diff
@@ -405,39 +562,40 @@ static std::vector<ClusterBlockId> sort_blocks(const vtr::vector<ClusterBlockId,
...
+            //put VTR assert
+            continue;
+        }
+
+        if (block_scores[blk_id].macro_size > 0) {
+            int imacro;
+            get_imacro_from_iblk(&imacro, blk_id, place_ctx.pl_macros);
+            t_pl_macro pl_macro;
+            pl_macro = place_ctx.pl_macros[imacro];
+            place_a_macro(MAX_NUM_TRIES_TO_PLACE_MACROS_RANDOMLY, pl_macro);
```
sfkhalid, line 589: place_macro

vpr/src/place/initial_placement.cpp
```diff
@@ -405,39 +562,40 @@ static std::vector<ClusterBlockId> sort_blocks(const vtr::vector<ClusterBlockId,
...
+        } else {
+            place_a_block(MAX_NUM_TRIES_TO_PLACE_BLOCKS_RANDOMLY, blk_id, pad_loc_type);
```
sfkhalid, line 591: place_block

vpr/src/place/initial_placement.cpp
```diff
@@ -405,39 +562,40 @@ static std::vector<ClusterBlockId> sort_blocks(const vtr::vector<ClusterBlockId,
...
+    auto& place_ctx = g_vpr_ctx.placement();
+
+    for (auto blk_id : sorted_blocks) {
+        //Check if block has already been placed
+        if (place_ctx.block_locs[blk_id].loc.x != -1) {
+            //put VTR assert
+            continue;
+        }
+
+        if (block_scores[blk_id].macro_size > 0) {
```
sfkhalid, line 584: check if it's in a macro using pl_macros data structure, more clear

vpr/src/place/initial_placement.cpp
```diff
@@ -236,101 +411,83 @@ static void initial_placement_pl_macros(int macros_max_num_tries, std::vector<st
...
-                legal_pos[itype][isub_tile][ipos] = legal_pos[itype][isub_tile][free_locations[itype][isub_tile] - 1]; /* overwrite used block position */
-                free_locations[itype][isub_tile]--;
+        } // Finished all tries
 
-                placed = true;
-            }
-        }
-    }
-}
+        if (placed == false) {
```
sfkhalid, line 472: if (!placed) is a bit more clear

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,214 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+
+    auto& grid = g_vpr_ctx.device().grid;
+
+    VTR_ASSERT_MSG(grid[loc.x][loc.y].width_offset == 0, "Should be at block base location");
+    VTR_ASSERT_MSG(grid[loc.x][loc.y].height_offset == 0, "Should be at block base location");
+
+    return legal;
+}
+
+static bool try_all_part_region_locations(ClusterBlockId blk_id, PartitionRegion& pr, t_logical_block_type_ptr block_type, enum e_pad_loc_type pad_loc_type) {
```
sfkhalid, line 116: add some comments (go through all regions, all dimensions, etc.)

see if you can pull out any helper functions

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,214 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+            for (int dy = 0; dy < y_range && placed == false; dy++) {
+                int cy = (y_lower_iter + dy)->first;
+
+                to_loc.x = compressed_block_grid.compressed_to_grid_x[cx];
+                to_loc.y = compressed_block_grid.compressed_to_grid_y[cy];
+
+                auto& grid = g_vpr_ctx.device().grid;
+                auto tile_type = grid[to_loc.x][to_loc.y].type;
+
+                if (regions[reg].get_sub_tile() != NO_SUBTILE) {
```
sfkhalid, line 147: check_subtile helper function

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,214 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+                    if (place_ctx.grid_blocks[to_loc.x][to_loc.y].blocks[to_loc.sub_tile] == EMPTY_BLOCK_ID && cluster_floorplanning_legal(blk_id, to_loc)) {
+                        set_block_location(blk_id, to_loc);
+                        placed = true;
+
+                        if (is_io_type(pick_physical_type(block_type)) && pad_loc_type == RANDOM) {
+                            place_ctx.block_locs[blk_id].is_fixed = true;
+                        }
+                    }
+                } else {
+                    for (const auto& sub_tile : tile_type->sub_tiles) {
```
sfkhalid, line 160: loops through calling check_subtile for each

sfkhalid:
QoR test comparisons on this pull requests so far:
titan_quick_qor - [titan_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7158816/titan_qor_comparison.xlsx)
vtr_reg_qor_chain - [compressed_grid_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7158817/compressed_grid_qor_comparison.xlsx)

vpr/src/place/initial_placement.cpp
sfkhalid: Can't make it const, if I do, I can't use the get_partition_region() accessors which I use to get the vector of regions.

vpr/src/place/initial_placement.cpp
sfkhalid: Doesn't work in the code - doesn't make exhaustive placement functions any shorter and also cannot be called from both macro and non-macro exhaustive placement functions because of the differences between how subtiles are checked between them.

vpr/src/place/initial_placement.cpp
sfkhalid: Doesn't work in the code - doesn't make exhaustive placement functions any shorter and also cannot be called from both macro and non-macro exhaustive placement functions because of the differences between how subtiles are checked between them.

sfkhalid:
New titan QoR comparison:
[titan_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7253606/titan_qor_comparison.xlsx)

sfkhalid:
new vtr_reg_qor_chain comparison:
[init_place_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7259854/init_place_qor_comparison.xlsx)

sfkhalid:
There were three circuits that had run failures on the CI due to the circuit being unroutable - these failures appeared in the strong, nightly1, and nightly2 tests. It is most likely that these failures were chance failures, as they do not occur when the same tests are run on a different machine, as well as when they are passed different seeds.

The following file summarizes some runs that were done to see whether the tests would fail locally:
[CI_unroutable_circuit_runs.docx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7260055/CI_unroutable_circuit_runs.docx)

The tests did not fail locally, whether run with the default seed of 1, or when passed different seed values. Furthermore, when the seed values on the configuration files of the regression tests were changed, the circuit runs do not fail on the CI either.

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,247 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+ * Returns true if a valid location is found, false otherwise
+ * Returns location (loc) by reference
+ */
+static bool get_legal_placement_loc(PartitionRegion& pr, t_pl_loc& loc, t_logical_block_type_ptr block_type);
+
+/*
+ * A routine to look for a valid placement location for a (non-macro) block exhaustively once the maximum number of random locations
+ * have been tried. Returns true if the block is placed, false otherwise.
+ */
+static bool try_all_part_region_locations(ClusterBlockId blk_id, PartitionRegion& pr, t_logical_block_type_ptr block_type, enum e_pad_loc_type pad_loc_type);
```
vaughnbetz, line 65: Comment inputs, esp. pad_loc_type as it is a bit mysterious.
Think about whether you should pass in block type as it can be found from blk_id.
Evaluate changing this to return a to_loc to see if the calling code would then be simpler / more logical.  
- ask for random loc
- ask for exhaustive loc
- place (and lock down if pad_loc_type == RANDOM)

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,247 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+
+/*
+ * A routine to look for a valid placement location for a block in a macro exhaustively once the maximum number of random locations
+ * have been tried. Returns true if the block is placed, false otherwise.
+ */
+static bool try_all_part_region_locations_macro(t_pl_macro pl_macro, PartitionRegion& pr, t_logical_block_type_ptr block_type);
```
vaughnbetz, line 71: Also align how the exhaustive & random macro block placers are called/return values.

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,247 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
 void print_sorted_blocks(const std::vector<ClusterBlockId>& sorted_blocks, const vtr::vector<ClusterBlockId, t_block_score>& block_scores);
 
-static int get_free_sub_tile(std::vector<std::vector<int>>& free_locations, int itype, std::vector<int> possible_sub_tiles) {
-    for (int sub_tile : possible_sub_tiles) {
-        if (free_locations[itype][sub_tile] > 0) {
-            return sub_tile;
+static void place_all_blocks(const std::vector<ClusterBlockId>& sorted_blocks,
+                             enum e_pad_loc_type pad_loc_type);
+
+static bool is_loc_on_chip(t_pl_loc& loc) {
```
vaughnbetz, line 78: Probably should put this in the grid class. It's a general utility.

vpr/src/place/initial_placement.cpp
```diff
@@ -51,38 +51,247 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores();
...
+                            int st_low = sub_tile.capacity.low;
+                            int st_high = sub_tile.capacity.high;
+
+                            for (int st = st_low; st <= st_high && placed == false; st++) {
+                                to_loc.sub_tile = st;
+                                if (place_ctx.grid_blocks[to_loc.x][to_loc.y].blocks[to_loc.sub_tile] == EMPTY_BLOCK_ID && cluster_floorplanning_legal(blk_id, to_loc)) {
+                                    set_block_location(blk_id, to_loc);
+                                    placed = true;
+
+                                    if (is_io_type(pick_physical_type(block_type)) && pad_loc_type == RANDOM) {
```
vaughnbetz, line 198: Mysterious: comment this.

vpr/src/place/initial_placement.cpp
```diff
@@ -166,63 +371,71 @@ static int try_place_macro(int itype, int ipos, int isub_tile, t_pl_macro pl_mac
...
-        macro_placed = false;
+    auto& place_ctx = g_vpr_ctx.placement();
+
+    auto& floorplanning_ctx = g_vpr_ctx.floorplanning();
+    auto& device_ctx = g_vpr_ctx.device();
+
+    macro_placed = false;
 
+    while (!macro_placed) {
         // Assume that all the blocks in the macro are of the same type
```
vaughnbetz, line 389: Move this comment down to where you get the block type.

vpr/src/place/initial_placement.cpp
```diff
@@ -236,101 +449,83 @@ static void initial_placement_pl_macros(int macros_max_num_tries, std::vector<st
...
-            auto possible_sub_tiles = get_possible_sub_tile_indices(type, logical_block);
-
-            int isub_tile = get_free_sub_tile(free_locations, itype, possible_sub_tiles);
+    bool placed = false;
+    while (!placed) {
+        /* -1 is a sentinel for a non-placed block, which the code in this routine will choose a location for.
+         * If the x value is not -1, we assume something else has already placed this block and we should leave it there.
+         * For example, if the user constrained it to a certain location, the block has already been placed.
+         */
+        if (place_ctx.block_locs[blk_id].loc.x != -1) {
```
vaughnbetz, line 469: Wrap in a little function :  ! (block_placed(blk_id)).  block_placed return true unless .x is INVALID_X (which you can define to -1).

vpr/src/place/initial_placement.cpp
```diff
@@ -236,101 +449,83 @@ static void initial_placement_pl_macros(int macros_max_num_tries, std::vector<st
...
 
-            bool floorplan_good = cluster_floorplanning_legal(blk_id, to);
+        for (int itry = 0; itry < blocks_max_num_tries && placed == false; itry++) {
+            found_legal_place = get_legal_placement_loc(pr, to, logical_block);
 
-            if (floorplan_good) {
-                place_ctx.grid_blocks[to.x][to.y].blocks[to.sub_tile] = blk_id;
-                place_ctx.grid_blocks[to.x][to.y].usage++;
+            if (found_legal_place) {
+                floorplanning_good = cluster_floorplanning_legal(blk_id, to);
```
vaughnbetz, line 495: Should this be done in get_legal_placement_loc? 
- Comment why this is needed (compressed_grid_approximation)
- Move this check to get_legal_placement_loc (with the comment).
- Reorder the parameters to get_legal_placement_loc (and other routines) so inputs come first, and outputs (return by reference) come after.

vpr/src/place/initial_placement.cpp
```diff
@@ -405,39 +600,42 @@ static std::vector<ClusterBlockId> sort_blocks(const vtr::vector<ClusterBlockId,
...
     }
 }
 
+static void place_all_blocks(const std::vector<ClusterBlockId>& sorted_blocks,
+                             enum e_pad_loc_type pad_loc_type) {
+    auto& place_ctx = g_vpr_ctx.placement();
+
+    for (auto blk_id : sorted_blocks) {
+        //Check if block has already been placed
+        if (place_ctx.block_locs[blk_id].loc.x != -1) {
```
vaughnbetz, line 616: Call a helper function instead of checking .x == -1.

vpr/src/place/place.cpp
```diff
@@ -2313,6 +2314,15 @@ static void get_non_updateable_bb(ClusterNetId net_id, t_bb* bb_coord_new) {
 
     ClusterBlockId bnum = cluster_ctx.clb_nlist.net_driver_block(net_id);
     pnum = net_pin_to_tile_pin_index(net_id, 0);
+
+    if (unsigned(pnum) > (physical_tile_type(bnum)->pin_width_offset.size() - 1)) {
```
vaughnbetz, line 2318: Comment issue open for a better fix -- give issue number. Explain why this code is needed.

vpr/src/place/initial_placement.cpp
```diff
@@ -166,63 +371,71 @@ static int try_place_macro(int itype, int ipos, int isub_tile, t_pl_macro pl_mac
     return (macro_placed);
 }
 
-static void initial_placement_pl_macros(int macros_max_num_tries, std::vector<std::vector<int>>& free_locations, const std::vector<t_pl_macro>& sorted_macros) {
+static void place_macro(int macros_max_num_tries, t_pl_macro pl_macro) {
```
vaughnbetz, line 374: I think you can merge the macro and non-macro cases by making a macro on the fly with only one block in it.  Should reduce code duplication by only having one code path.

vaughnbetz: Also merge place_block and place_macro.

vaughnbetz (changes requested):
Mostly good.
Some commenting suggestions and also a significant change: merge the macro and non-macro cases by always creating a macro and passing it to the helper routines.
Also order you parameters as (inputs, outputs) for function call throughout the file (easier to read).

vpr/src/place/initial_placement.cpp
sfkhalid: Made random macro block placer call try_place_macro (i.e. aligned it with the exhaustive way)

vpr/src/place/initial_placement.cpp
sfkhalid: no longer use this routine (place_block) because the macro and non-macro flows were merged

vpr/src/place/initial_placement.cpp
sfkhalid: tried this and it caused problems with the compiler saying that t_pl_loc is not declared.

vaughnbetz: t_pl_loc is defined in vpr_types.h; you should #include it in the grid class header file (presumably you haven't).

sfkhalid: I included it in the header file there, the message came when it was included

vpr/src/place/place.cpp
sfkhalid: added better fix for bug - sync pins if needed in apply_move_blocks and revert_move_blocks

sfkhalid:
new vtr_reg_qor_comparison:
[init_place_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7289309/init_place_qor_comparison.xlsx)

This comparison was after fixing the bug in placement where physical pins were not always synced when necessary. There are significant changes seen in the place times

vaughnbetz:
Thanks @sfkhalid .  I think both placement time and (fixed channel width) critical path routing runtime are down by 2%?  Is that right?  Probably just machine load noise then.
The total flow time changed more, but is probably dominated by the binary search runtime, which is more chaotic.  Please take a look at the place time in particular to confirm it is OK.

sfkhalid:
> Thanks @sfkhalid . I think both placement time and (fixed channel width) critical path routing runtime are down by 2%? Is that right? Probably just machine load noise then. The total flow time changed more, but is probably dominated by the binary search runtime, which is more chaotic. Please take a look at the place time in particular to confirm it is OK.

Yes, that's correct. On the sheet called compare_times, I have a comparison of the place times and they are basically the same. The min_chan_width_route_time is what's making the overall flow time go up, because that value went up by 24% when comparing the changed branch to the master.

vaughnbetz:
OK, that sounds fine.

sfkhalid:
New titan QoR comparison (after fixing bug related to syncing physical pins during placement):
[titan_qor_comparison.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7294485/titan_qor_comparison.xlsx)
 No significant changes in place time

vaughnbetz (commented):
Looks good; just a few suggestions for shorter/more clear code in a few places.

PR #1843, opened 2021-09-09
RRGraphView node_cost_index() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::node_cost_index()` throughout VTR. Every time `device_ctx.rr_nodes[index].cost_index()` was called has been replaced with `rr_graph.cost_index(RRNodeId(index))`. In order to do this, I followed a pattern similar to that in the last PR.

### Motivation and Context

### How Has This Been Tested?

I have run the regression tests for `vtr_reg_strong` along with the QoR testing found at `    $ ../scripts/run_vtr_task.py regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain`. Results from these QoR tests can be found below. The file containing all results can be found [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7137545/cost_index.xlsx)

|---|---|---|
|vtr_flow_elapsed_time	|1|	1.013103428|
|odin_synth_time	|1|	1.004124928|
|abc_synth_time	|1|	1.009521229|
|max_vpr_mem	|1|	0.999794275|
|pack_time	|1|	0.984221145|
|place_time	|1|	1.004760669|
|min_chan_width_route_time	|1|	1.01720273|
|crit_path_route_time	|1|	1.019740612|

#### Types of changes

#### Checklist:


vpr/src/route/rr_graph_uxsdcxx_serializer.h
```diff
@@ -706,23 +706,23 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
             node.set_cost_index(CHANX_COST_INDEX_START + segment_inf_.size() + segment_id);
-            seg_index_[node.cost_index()] = segment_id;
+            seg_index_[rr_graph.node_cost_index(node.id())] = segment_id;
         }
         return inode;
     }
     inline void finish_node_segment(int& /*inode*/) final {}
     inline int get_node_segment_segment_id(const t_rr_node& node) final {
-        return (*rr_indexed_data_)[node.cost_index()].seg_index;
+        return (*rr_indexed_data_)[(*rr_graph_).node_cost_index(node.id())].seg_index;
```
tangxifan, line 718: You may use ``rr_graph->node_cost_index`` to avoid too many brackets

tangxifan (changes requested):
@ethanroj23 Looks good to me. Just minor changes required.

I actually would like to know @vaughnbetz 's and your opinion about creating a strong id for the cost index ``RRIndexedDataId``. It will help us to avoid the misuse of the cost index.

Accordingly, you need to modify the data type of ``rr_indexed_data``

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/77a3df829fdf509a76b9d7b941376f3835c2576a/vpr/src/base/vpr_context.h#L152

to 

```
vtr::vector<RRIndexedDataId, t_rr_indexed_data> rr_indexed_data
```

You can refer to https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/77a3df829fdf509a76b9d7b941376f3835c2576a/vpr/src/device/rr_graph_fwd.h#L16-L24 
about how to create the strong id.

Note: we may still keep the ``short`` data type inside the rr_graph_storage for the sake of memory efficiency. But the API should always return a strong data type.

Let me know what you think. I am open to discussion.

vpr/src/route/segment_stats.cpp
```diff
@@ -50,7 +50,7 @@ void get_segment_usage_stats(std::vector<t_segment_inf>& segment_inf) {
 
     for (size_t inode = 0; inode < device_ctx.rr_nodes.size(); inode++) {
         if (rr_graph.node_type(RRNodeId(inode)) == CHANX || rr_graph.node_type(RRNodeId(inode)) == CHANY) {
-            cost_index = device_ctx.rr_nodes[inode].cost_index();
+            cost_index = rr_graph.node_cost_index(RRNodeId(inode));
```
vaughnbetz, line 53: Here's an example of where I think we should eventually hide the cost_index.
Instead of asking for a cost_index, and then looking up seg_type, we instead should have an api that takes an RRNodeId and returns the seg_type 
seg_type = rr_graph.seg_type(RRNodeId(inode));
Internally, the .seg_type() function would get the cost_index, and then look up the seg_type and return it.

It hides the fact that this seg_type lookup goes through a cost_index (flyweight) for storage efficiency; the caller doesn't care.

vaughnbetz:
I like the idea of returning a StrongId, if we keep the api to return a cost_index.  It should be stored as a short internally though.

Eventually I think the rr-indexed accesses should all be wrapped in an api presented by RRGraphView so that the various data we store using cost_index in a flyweight pattern (store the unique values) is just requested from RRGraphView.  See an example in the comment above; there would be a few functions like that.

So the question is: should we wrap cost_index in an api now, or just skip that step and wrap the rr_indexed data accesses in an api, and hide cost_index?  Either way, we should eventually get to the rr_indexed data accesses being wrapped and directly accessed using RRGraphView.some_func(inode) .

vpr/src/route/segment_stats.cpp
vaughnbetz: It's possible there's some cpu time penalty to that in some code, if we ask for a bunch of cost_index data in rapid succession.  But in that case we could return a struct with the data that tends to be requested together packed into it.

tangxifan: @vaughnbetz I think it is a good idea to streamline the API here. Actually, I am thinking about if we should include the ``rr_segments`` and ``rr_switch_inf`` as part of the ``RRGraphView`` and ``RRGraphBuilder``.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/77a3df829fdf509a76b9d7b941376f3835c2576a/vpr/src/base/vpr_context.h#L173-L177

I said so because these two data structures are modified copies of the original ones in architecture data. They have been updated to accurately model the switches and segments in the context of a specific routing resource graph.
Since they are adhoc to RRGraph, they can be part of the graph and we can provide shortcut APIs rather than force developers to visit several data structures.

However, it may be too complicated to this PR. It can be in a pull request near future (I can take the lead on it; since I have already tried in the [rr_graph_obj.h](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/77a3df829fdf509a76b9d7b941376f3835c2576a/vpr/src/device/rr_graph_obj.h#L824))

Let me know what you think.

ethanroj23: If we decide to add API calls, the following are all a part of t_rr_indexed_data:     `base_cost`, `saved_base_cost`, `ortho_cost_index`, `seg_index`, `inv_length`, `T_linear`, `T_quadratic`, `C_load`.

The `ortho_cost_index` is used to access each of these data points for a different cost index, so we would have to consider how the API would handle that. We could have a second argument to the functions `rr_graph.seg_type(rr_node, is_ortho)` or potentially two separate functions `rr_graph.seg_type()`, `rr_graph.ortho_seg_type()`

vaughnbetz: I agree, I think both those data structures should go in RRGraphView, and be built in RRGraphBuilder.
I also talked to @kmurray about this today in another meeting, and he also thinks hiding the cost_index and rr_indexed_data in RRGraphView api calls is the right way to go.

vaughnbetz:
So the summary:

1. cost_index: use a StrongId
2. should hide cost index behind new methods in RRGraphView that let us directly return the data stored on a cost_index (hide the flyweight).
3. should move rr_segments and rr_switch_inf into the RRGraphView/RRGraphBuilder eventually.

Should #1 be done in this PR or later?
I believe #2 should be done in separate PRs after this one, and the same for #3.  
@ethanroj23 : what is your preference?

vpr/src/route/segment_stats.cpp
vaughnbetz: I think the rr_graph.ortho_seg_type() style would be more clear.
For items that tend to be used together, I'd make the API return a tuple or struct with them packed in (e.g. inv_length, T_linear, T_quadratic and C_load are all probably used just for the classic lookahead and could be returned as a group only).

tangxifan:
It seems that ODIN-II basic tests are stuck constantly (at least on my pull request #1847 and this one). Not sure what the issue is. It is a required test. As a result, it may block pull request merging.

vaughnbetz:
Yes, this is consistently stuck.  I can override and merge.
But we should probably resolve the questions above first: merge as is, or implement 1, 2, or 3 first?

vaughnbetz:
@sdamghan : it seems the odinII basic tests are still showing up in recent PRs?  Is there some config update that needs to happen?

tangxifan:
> Yes, this is consistently stuck. I can override and merge.
> But we should probably resolve the questions above first: merge as is, or implement 1, 2, or 3 first?

I would like to hear from @ethanroj23 which way he prefer.
I can help whenever he needs. 
My view is 
- Option 1 is doable for this PR (it can be in a separated PR if we prefer single objective per PR).
- Option 2,3 can be in later PRs.

sdamghan:
> @sdamghan : it seems the odinII basic tests are still showing up in recent PRs? Is there some config update that needs to happen?

@vaughnbetz 
I do not think there is anything more than what we should specify in the [VTR_ROOT/.github/workflows/test.yml](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d8417a34a78e170c5bc91cbb5fea5912bd29af3a/.github/workflows/test.yml#L169) file to determine the GitHub actions, however, I'll again investigate this.
I think I need to specify here that ODIN-II Basic Tests is deprecated and no longer performs any CI test. I mean it is only an empty label.  So until I find a solution to remove it from the CI list please feel free to ignore it. I should mention that if Odin-II fails it would appear in one of the following CI tests:
#### GitHub Actions:
- Test / ODIN-II Basic Test [link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d8417a34a78e170c5bc91cbb5fea5912bd29af3a/.github/workflows/test.yml#L170)
- Test / Yosys+ODIN-II Basic  [link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d8417a34a78e170c5bc91cbb5fea5912bd29af3a/.github/workflows/test.yml#L215)

#### Kokoro Tests:
- ODIN-II Strong Tests [link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d8417a34a78e170c5bc91cbb5fea5912bd29af3a/.github/kokoro/continuous/odin_strong.cfg)
- Yosys+ODIN-II Strong Tests [link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d8417a34a78e170c5bc91cbb5fea5912bd29af3a/.github/kokoro/continuous/odin_tech_strong.cfg)
- Yosys+Odin [link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d8417a34a78e170c5bc91cbb5fea5912bd29af3a/.github/kokoro/continuous/yosys_odin_test.cfg)

ethanroj23:
@tangxifan I think option 1 is doable as well. I will take a look tomorrow and let you know if I will need any assistance. Thanks!

sdamghan:
@vaughnbetz `ODIN-II Basic Tests` has been hardcoded into the master branch protection rules as a required status. I think from now on, everything should be good since there is no deprecated CI label.

vaughnbetz:
@sdamghan : seems like it now shows up as finished, and has a detailed result, but no "required".  
We can talk about this on Thursday.  If it is not needed (deprecated) it seems like it should be possible to remove it from the CI list.
The master branch protection rules say all basic tests must pass, but I don't see anything that refers to Odin-II basic tests specifically (but I assume they match as a basic test).

sdamghan:
@vaughnbetz `ODIN-II Basic Tests` was showing in the CI list since it was mentioned in master branch rules. Since `ODIN-II Basic Tests` was replaced with `Test/ODIN-II Basic Test`, it was continuously running and the status was pending. I removed `ODIN-II Basic Tests` from branch rules last night. You are right, I forgot to label the `Test/ODIN-II Basic Test` and `Test/Yosys+Odin-II basic` tests as required tests. I have just labelled them. I believe Odin-II and Yosys+Odin-II CI tests should be fine now. I will provide more information on Thursday if required.

vpr/src/route/rr_graph_storage.h
```diff
@@ -185,8 +185,12 @@ class t_rr_graph_storage {
     short node_capacity(RRNodeId id) const {
         return node_storage_[id].capacity_;
     }
-    short node_cost_index(RRNodeId id) const {
-        return node_storage_[id].cost_index_;
+    RRIndexedDataId node_cost_index(RRNodeId id) const {
+        return RRIndexedDataId(node_storage_[id].cost_index_);
+    }
+
+    bool cost_index_is_invalid(RRNodeId id) const {
```
vaughnbetz, line 192: As discussed in the VTR meeting, we can get rid of this and use the StrongID::Invalid() and/or it's bool test  functionality.

vaughnbetz:
Code changes look good, but there are conflicts to resolve @ethanroj23 .  Can you resolve them and then we can merge?

ethanroj23:
> Code changes look good, but there are conflicts to resolve @ethanroj23 . Can you resolve them and then we can merge?

Conflicts are now resolved. We should be good to go.

tangxifan:

Issue #1808, opened 2021-07-29
Runs with sanitization on seems to be running out of memory in github actions CI


vaughnbetz:
Approximately 10 days ago runs that use sanitization in CI (S: vtr_reg_basic, odin_basic_tests) started failing with an error code that indicates out of memory in the CI tests.  We also get some QoR test failures on those same designs that indicate peak memory footprint has increased by 14x. This is affecting multiple pull requests, so we haven't been able to track it to any recent code change.

@tangxifan has run the sanitizers locally, and they complete fine, with no unusually high memory footprint.  @sfkhalid : I believe you have seen the same thing (there isn't a problem when you run these locally).

Initially I suspected that the recent rr-graph refactoring could have triggered a bad memory behaviour (large footprint) with the sanitizers, but that doesn't seem to be the case so far. @tangxifan has tried rolling back various changes locally and so far hasn't seen a correlation between his changes and these sanitizer failures.  Also, the Odin II basic tests are failing, and they do not use the router, but do use sanitization.  

The S: vtr_reg_strong tests are larger designs that also use sanitization, and they are working fine. They run on kokoro though, not on github actions.  

I have gone through the github action configuration and cmake files and so far haven't seen any recent changes that seem related at all (or any recent changes, in most cases).  

So the leading theory at this point is that something changed in the github runners that is causing the problem.  github action runners are supposed to have 7 GB of memory available for two cores.  @tangxifan found that the S:vtr_reg_basic tests only used ~400 MB when run locally (one at a time), so this should be fine, even if we double the 400 MB for two parallel runs.

I am a bit baffled, but this is impacting our ability to merge code, so it's important to figure out.  One hacky fix could be to move all sanitized runs (S:vtr_reg_basic, odin_basic_regression) to kokoro, which appears to have no problems.  That isn't too satisfying a solution though.

@kgugala : Tim suggested you were the best person to debug github actions, so assigning this to you. @tangxifan : please summarize your data to date here, and @sfkhalid and @sdamghan please add any additional information you have. @umarcor : did the removal of travis in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1804 lead to any configuration settings that have already been changed that might have caused this?

https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1802 has some earlier discussion on this, and shows the symptom.

vaughnbetz:
[...]

PR #1857, opened 2021-09-29
[Github Actions] Push artifacts on regression test failure 

ganeshgore:
#### Description
Updated GitHub action `test.yml` to upload artifacts on the test failure. 
- It tracks failure in only the Regression test (not in build or any other process)
- Artifacts are collected  with these regex `vtr_flow/**/*.log` and `vtr_flow/**/*.out`  to upload
- The retention period for artifacts is set to 3 days (assuming the user can keep a local copy in case of failure)

#### Related Issue

Addresses another part of issue #1808

#### Motivation and Context
Simplify debugging on CI regression failures

Created an artificial bug in the basic regression test
(I will revert that before merging)

As an example please check the following failed CI run and uploaded artifacts
https://github.com/ganeshgore/vtr-verilog-to-routing/actions/runs/1288245384

#### Types of changes

#### Checklist:

- [x] My change requires a change to the documentation (In case the debugging process needs to be documented)


vtr_flow/arch/no_timing/memory_sweep/k4_N10_memSize16384_memData64.xml
```diff
@@ -579,7 +579,7 @@
       <output name="out" num_pins="72"/>
       <mode name="two_divisible_mult_18x18">
         <pb_type name="divisible_mult_18x18" num_pb="2">
-          <input name="a" num_pins="18"/>
+          <input name="aa" num_pins="18"/>
```
tangxifan, line 582: Not sure why the artifacts requires some changes in the XML files.

.github/workflows/test.yml
```diff
@@ -137,6 +137,15 @@ jobs:
...
 
+    - name: Upload artifact
+      uses: actions/upload-artifact@v2
+      if: ${{ failure() }}
+      with:
+        name: failed_${{matrix.name}}_regression_log
+        retention-days: 3
+        path: |
+          vtr_flow/**/*.log
+          vtr_flow/**/*.out
```
tangxifan, line 148: @vaughnbetz 

- If you know any files should be uploaded to the artifacts. Please feel free to comment.
- Currently, the retention of artifacts is set to 3 days. See if you need something different

tangxifan (commented):
@ganeshgore I did look into the CI test example that you have posted. But I cannot find any button to download these artifacts. Can you double check and comment where to download them

Also, just to confirm, the artifacts are only uploaded for failed tests?

ganeshgore:
You can download the artifacts from the bottom of that page, and artifacts are created for the failed matrix which is one of these (The artifact label will let you know which one is failed)
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/f86f9ee452cebbb0b595c786d16711a09c067ee8/.github/workflows/test.yml#L96

image

vtr_flow/arch/no_timing/memory_sweep/k4_N10_memSize16384_memData64.xml
ganeshgore: This is the artificial bug created to fail one of the tests so that we can check artifact uploading. I will revert it before merging to master

.github/workflows/test.yml
vaughnbetz: This looks great -- thanks!  I think you've captured the key ones.
To make it faster to re-run / investigate errors, we could also bring back:
*.blif (or *.pre-vpr.blif)
*.p
*.net
*.r

Will increase file sizes quite a bit, but would let us immediately examine the results of a specific compilation stage.  So if they don't cause disk problems I would bring them back (compressed).

tangxifan:
@ganeshgore Thanks for the screenshot. Confirmed on my side as well, although the interface looks a bit different.

image

umarcor (commented):
I recommend uploading the artifacts unconditionally and removing the explicit retention time. The motivation is that CI resources for open source (aka public) GitHub projects are free and "unlimited".

For workflows/jobs, there is a limit of 18 concurrent executions, but all additional scheduled jobs are added to the queue. Therefore, there might be bottlenecks, but everything is executed in the end.

For artifacts, there is some limit per org (not explicitly documented) and it is used as a circular FIFO. That is, whenever the size limit is reached and new artifacts are pushed, the oldest ones are removed. At the same time, artifacts older than 90 days are removed.

Therefore, unless you expect hundreds of workflow executions and artifacts to be generated everyday, I think it is safe no not worry about `if:` or `retention-days:`. Those features are rather meant for companies using GitHub for private repos, which do need to pay for the resources.

umarcor:
> But I cannot find any button to download these artifacts. Can you double check and comment where to download them

@tangxifan, note that, unlike release assets, artifacts are not publicly visible. You need to be logged in.

> although the interface looks a bit different.

You were looking at the "Checks" tab of this PR, while he was looking at the "Actions" tab of the repo. That is because the CI workflows triggered by a PR are checks.

vaughnbetz:
1. We're getting an error in vtr_reg_strong due to a : in a filename (name of the test), which artifact upload considers an invalid character.  Please rename that test to remove the :  .
`Error: Artifact path is not valid: /tasks/regression_tests/vtr_reg_strong/strong_target_pin_util/run001/EArch.xml/styr.blif/common_--target_ext_pin_util_clb:0.1_0.8/EArch.xml. Contains character: ":". Invalid characters include: ",:,<,>,|,*,?.`

2. Given the @umarcor 's comments on file limits being very generous I think we should upload the intermediate result files I listed above too.  Could do that in a separate PR if you want to test for any bad effects of adding those larger files separately.

ganeshgore:
CI is green now, I considered all the above suggestions.  You can download artifacts from the actions pages now. It's around 800 MBs of storage per run. 

image

PR #1860, opened 2021-09-30
Global variable removal

baberali-rs:
#### Description

The purpose of this PR is to remove the use of global variables from any function for modularization of VTR. This effort is started from the describe_rr_node() function defined in rr_graph.cpp in route directory. After removing global variables from routers, we will move on to placers, GUI, timing analyzer etc.

#### Motivation and Context

Use of global variables is not a recommended way. Removing them will benefit in modularization and less usage of memory.

 - [x] All vtr basic and strong tests are passing.
 - [x] VTR and Titan tests are passed.QoR comparison sheets are attached.  
#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/draw/search_bar.cpp
+++ b/vpr/src/draw/search_bar.cpp
@@ -127,7 +127,7 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
 
 bool highlight_rr_nodes(int hit_node) {
     t_draw_state* draw_state = get_draw_state_vars();
-
+    auto& device_ctx = g_vpr_ctx.device();
     char message[250] = "";
 
     if (hit_node != OPEN) {
@@ -147,11 +147,11 @@ bool highlight_rr_nodes(int hit_node) {
                 draw_state->draw_rr_node[node].node_highlighted = false;
             }
             //Print info about all nodes to terminal
-            VTR_LOG("%s\n", describe_rr_node(node).c_str());
+            VTR_LOG("%s\n", describe_rr_node(node,device_ctx).c_str());
         }
 
         //Show info about *only* hit node to graphics
-        std::string info = describe_rr_node(hit_node);
+        std::string info = describe_rr_node(hit_node,device_ctx);
 
         sprintf(message, "Selected %s", info.c_str());
         rr_highlight_message = message;
--- a/vpr/src/draw/search_bar.cpp
+++ b/vpr/src/draw/search_bar.cpp
@@ -147,11 +147,11 @@ bool highlight_rr_nodes(int hit_node) {
                 draw_state->draw_rr_node[node].node_highlighted = false;
             }
             //Print info about all nodes to terminal
-            VTR_LOG("%s\n", describe_rr_node(node,device_ctx).c_str());
+            VTR_LOG("%s\n", describe_rr_node(node, device_ctx).c_str());
         }
 
         //Show info about *only* hit node to graphics
-        std::string info = describe_rr_node(hit_node,device_ctx);
+        std::string info = describe_rr_node(hit_node, device_ctx);
 
         sprintf(message, "Selected %s", info.c_str());
         rr_highlight_message = message;
--- a/vpr/src/route/route_profiling.cpp
+++ b/vpr/src/route/route_profiling.cpp
@@ -214,6 +214,7 @@ void conn_start() {
     conn_start_time = clock();
 }
 void conn_finish(int src_rr, int sink_rr, float criticality) {
+    auto& device_ctx = g_vpr_ctx.device();    
     float route_time = static_cast<float>(clock() - conn_start_time) / CLOCKS_PER_SEC;
     if (route_time > worst_conn_time) {
         worst_src_rr = src_rr;
@@ -223,16 +224,17 @@ void conn_finish(int src_rr, int sink_rr, float criticality) {
     }
 
     VTR_LOG("%s to %s (crit: %f) took %f\n",
-            describe_rr_node(src_rr).c_str(),
-            describe_rr_node(sink_rr).c_str(),
+            describe_rr_node(src_rr,device_ctx).c_str(),
+            describe_rr_node(sink_rr,device_ctx).c_str(),
             criticality,
             route_time);
 }
 void net_finish() {
     if (worst_conn_time > 0.f) {
+        auto& device_ctx = g_vpr_ctx.device(); 
         VTR_LOG("Worst conn was %s to %s (crit: %f) took %f\n",
-                describe_rr_node(worst_src_rr).c_str(),
-                describe_rr_node(worst_sink_rr).c_str(),
+                describe_rr_node(worst_src_rr,device_ctx).c_str(),
+                describe_rr_node(worst_sink_rr,device_ctx).c_str(),
                 worst_crit,
                 worst_conn_time);
     }
--- a/vpr/src/route/route_profiling.cpp
+++ b/vpr/src/route/route_profiling.cpp
@@ -213,7 +213,9 @@ void profiling_initialization(unsigned max_fanout) {
 void conn_start() {
     conn_start_time = clock();
 }
+
 void conn_finish(int src_rr, int sink_rr, float criticality) {
+    
     auto& device_ctx = g_vpr_ctx.device();    
     float route_time = static_cast<float>(clock() - conn_start_time) / CLOCKS_PER_SEC;
     if (route_time > worst_conn_time) {
@@ -231,7 +233,9 @@ void conn_finish(int src_rr, int sink_rr, float criticality) {
 }
 void net_finish() {
     if (worst_conn_time > 0.f) {
+
         auto& device_ctx = g_vpr_ctx.device(); 
+        
         VTR_LOG("Worst conn was %s to %s (crit: %f) took %f\n",
                 describe_rr_node(worst_src_rr, device_ctx).c_str(),
                 describe_rr_node(worst_sink_rr, device_ctx).c_str(),
--- a/vpr/src/route/route_profiling.cpp
+++ b/vpr/src/route/route_profiling.cpp
@@ -224,8 +224,8 @@ void conn_finish(int src_rr, int sink_rr, float criticality) {
     }
 
     VTR_LOG("%s to %s (crit: %f) took %f\n",
-            describe_rr_node(src_rr,device_ctx).c_str(),
-            describe_rr_node(sink_rr,device_ctx).c_str(),
+            describe_rr_node(src_rr, device_ctx).c_str(),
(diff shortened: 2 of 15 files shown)


vpr/src/route/rr_graph.h
```diff
@@ -41,7 +41,7 @@ void create_rr_graph(const t_graph_type graph_type,
 void free_rr_graph();
 
 //Returns a brief one-line summary of an RR node
-std::string describe_rr_node(int inode);
+std::string describe_rr_node(int inode, const DeviceContext& device_ctx);
```
tangxifan, line 44: Detail the comments about what this function does. Please keep in mind, make the comments being compatible with Doxygen format.
For example, 

```
@brief This function generates text in string format which describes detailed attributes for a given routing resource node. The attributes are essential for developers to diagnose when they hit a bug. The print-out message is helpful to find mismatches or unexpected nodes between the built routing resource graph and architecture description.
- In general, it shows the basic information, such as coordinate, number of fan-in and fan-out.
- It also shows more details depending on the node type. For example, as for routing tracks, it shows the directionality, the wire segment that represents the wire as well as track index in a routing channel.
```

vpr/src/route/rr_graph.cpp
```diff
@@ -2420,10 +2420,9 @@ static vtr::NdMatrix<std::vector<int>, 4> alloc_and_load_track_to_pin_lookup(vtr
 
     return track_to_pin_lookup;
 }
-
 /* TODO: This function should adapt RRNodeId */
-std::string describe_rr_node(int inode) {
-    auto& device_ctx = g_vpr_ctx.device();
+std::string describe_rr_node(int inode, const DeviceContext& device_ctx) {
```
tangxifan, line 2424: Actually, I prefer to have the ``DeviceContext`` as the 1st argument while the ``inode`` as the 2nd. In the argument list, usually we want to put big objects first and then followed by indices. Your call.

vpr/src/route/route_timing.cpp
```diff
@@ -1286,11 +1287,11 @@ static bool timing_driven_route_sink(
      * add this branch to the existing route tree and update pathfinder costs and rr_node_route_inf to reflect this */
     auto& route_ctx = g_vpr_ctx.mutable_routing();
     auto& cluster_ctx = g_vpr_ctx.clustering();
-
+    auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 1290: Should we use ``const auto&`` since you are not going to modify anything in device modeling.

vpr/src/route/route_timing.cpp
```diff
@@ -1187,10 +1187,11 @@ static bool timing_driven_pre_route_to_clock_root(
     auto& route_ctx = g_vpr_ctx.mutable_routing();
     auto& cluster_ctx = g_vpr_ctx.clustering();
     auto& m_route_ctx = g_vpr_ctx.mutable_routing();
+    auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 1190: Should we use ``const auto&`` since you are not going to modify anything in device modeling.

vpr/src/route/route_profiling.cpp
```diff
@@ -213,7 +213,9 @@ void profiling_initialization(unsigned max_fanout) {
 void conn_start() {
     conn_start_time = clock();
 }
+
 void conn_finish(int src_rr, int sink_rr, float criticality) {
+    auto& device_ctx = g_vpr_ctx.device();
```
tangxifan, line 218: Should we use ``const auto&`` since you are not going to modify anything in device modeling.

tangxifan (changes requested):
@baberali-rs The changes are straight forward to understand. Some changes are required.

Also be aware to use ``const auto& device_ctx`` since you are not going to modify anything in device modeling.

tangxifan:
@baberali-rs  From your QoR tests, I see a 8% increase in runtime. It looks a significant drawback? Can you look into it? It may be due to the workload of a runner.

vpr/src/draw/search_bar.cpp
```diff
@@ -127,7 +127,7 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
 
 bool highlight_rr_nodes(int hit_node) {
     t_draw_state* draw_state = get_draw_state_vars();
-
+    auto& device_ctx = g_vpr_ctx.device();
```
vaughnbetz, line 130: I don't think you change anything in this context, so better to mark it as const

Reasons:
1. Makes it more clear to someone reading the code that you won't modify this (large) global construct.
2. Can help the compiler optimize, by guaranteeing you will never modify this object through this reference/pointer. This can help it determine if it is safe to keep some of the data accessed through this pointer in registers, rather than always reloading them by accessing main memory via the pointer (i.e. it simplifies pointer alias analysis).

vpr/src/route/rr_graph2.cpp
```diff
@@ -1176,7 +1176,7 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
                             const RRGraphView& rr_graph,
                             const t_rr_graph_storage& rr_nodes) {
     std::unordered_map<RRNodeId, int> rr_node_counts;
-
+    auto& device_ctx = g_vpr_ctx.device();
```
vaughnbetz, line 1179: This should be const I think

vaughnbetz (changes requested):
A few minor comments and one bigger one:
1. We shouldn't be passing the device_context to describe_rr_node.  Instead pass a const RRGraphView& object.  That object has everything we need for this function, and in removing global variables we are trying to pass a little data as possible to leaf functions.  If we pass the whole device context, we are not really moving forward much or at all in decoupling data structures, and the person reading the code can't tell what describe_rr_node accesses (it could access anything in device_context).
2. describe_rr_node currently takes an integer to indicate the rr_node. If possible, change it to an RRNodeId for stronger type checking and to avoid casting in that routine.  If that causes too much casting in the calling routines (which sometimes are using int rr_node_ids) we can defer this.
3. Line 257 of route/router_lookahead_map.cpp looks odd / should be cleaned up.
```
  const auto& temp_rr_graph = device_ctx.rr_graph; //TODO rename to rr_graph once the variable on the next line is unneeded
    auto& rr_graph = device_ctx.rr_nodes;
```
I think we should be able to get rid of the old rr_graph variable now and rename temp_rr_graph.

vpr/src/route/route_timing.cpp
baberali-rs: I think auto will depict it as` const` as `device()`is returning `const`.

tangxifan: @baberali-rs Do you any article to support this conclusion? It may be interpreted differently by different compilers and we do not want to risk.

vaughnbetz:
See this article:  https://stackoverflow.com/questions/5070521/does-const-auto-have-any-meaning

Conclusion: best to do const auto when it is const -- it's more readable at the least, and safer to ensure we don't lose the const based on exactly what's on the other side.

baberali-rs:
> @baberali-rs From your QoR tests, I see a 8% increase in runtime. It looks a significant drawback? Can you look into it? It may be due to the workload of a runner.

I had run the tests at day time. I will run again QoR tests after making suggested changes at lesser workload on server to see if runtime decreases.

baberali-rs:
> A few minor comments and one bigger one:
> 
>     1. We shouldn't be passing the device_context to describe_rr_node.  Instead pass a const RRGraphView& object.  That object has everything we need for this function, and in removing global variables we are trying to pass a little data as possible to leaf functions.  If we pass the whole device context, we are not really moving forward much or at all in decoupling data structures, and the person reading the code can't tell what describe_rr_node accesses (it could access anything in device_context).
> 
>     2. describe_rr_node currently takes an integer to indicate the rr_node. If possible, change it to an RRNodeId for stronger type checking and to avoid casting in that routine.  If that causes too much casting in the calling routines (which sometimes are using int rr_node_ids) we can defer this.
> 
>     3. Line 257 of route/router_lookahead_map.cpp looks odd / should be cleaned up.
> 
> 
> ```
>   const auto& temp_rr_graph = device_ctx.rr_graph; //TODO rename to rr_graph once the variable on the next line is unneeded
>     auto& rr_graph = device_ctx.rr_nodes;
> ```
> 
> I think we should be able to get rid of the old rr_graph variable now and rename temp_rr_graph.

I must pass another argument `const DeviceGrid& obj` for grid? At this line [https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.cpp#L2449](url)

vaughnbetz:
You're right; you'd have to pass a const reference to the grid too.  

I think there are 2 ways we could go on this:

1. Pass the RRGraphView and the grid (both const &) to this routine.  Makes explicit what data it uses, and that it doesn't change it.

or:

2. Leave the code as is (back out of this PR). The current code lets this low-level function grab a read-only copy of the device context so it can print information.  That's not a bad paradigm; it isn't confusing since it is using multiple parts of the device context and it is read only.

I think the current PR is worse than either of those approaches.  It passes the whole device context explicitly.  Hence we haven't cut down the data that is accessible to this function or made what data it depends on much more explicit.  But we are forcing all callers to now pass it a pointer to a big data structure.  I don't think we want to adopt that paradigm throughout the code -- we'll wind up modifying a lot of callers to pass big device contexts to leaf routines, but I don't see any gain to that. It could be more confusing, since leaf routines might wind up with a mix of passed in device contexts and contexts accessed by calling the global g_vtr_context.device() routine.

I think this is worth discussing in the Thursday meeting, as it is good to hash out the paradigm we're following early.

tangxifan:
@vaughnbetz I agree with your opinion. Passing the ``DeviceContext`` is an overkill. Ideally, we only pass ``DeviceGrid`` and ``RRGraphView``. However, currently I see the function also used ``rr_indexed_data`` and ``rr_segments`` which may be merged into ``RRGraphView`` during our refactoring. Considering the number of dependent data structure, it is too many now. We should discuss this Thursday. We may postpone this PR to merged. It is a starting point. We can rework it once ``RRGraphView`` includes   ``rr_indexed_data`` and ``rr_segments``. Then it will be a good time.

vaughnbetz:
That makes sense to me @tangxifan .  RRGraphView includes rr_segments as a private member, but I think there are not accessors to get at the data in it yet.  rr_indexed_data is also there as a private variable, and @ethanroj23 has just put back accessors to get at its data, so I think it is OK.

tangxifan:
@baberali-rs @behzadmehmood-rs We can continue this effort as RRGraph is now a library

behzadmehmood-rs:
> @baberali-rs @behzadmehmood-rs We can continue this effort as RRGraph is now a library

Sure, we will do it now.

PR #1862, opened 2021-09-30
WIP: RRGraphView is_node_on_specific_side() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::is_node_on_specific_side(node, side)` throughout VTR. Every time `device_ctx.rr_nodes[node].is_node_on_specific_side(side)` was called has been replaced with `rr_graph.is_node_on_specific_side(RRNodeId(node), side)`. In order to do this, I followed a pattern similar to that in the last PR.

### Motivation and Context

### How Has This Been Tested?

As this is a WIP, no testing has been completed yet.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/rr_graph.cpp
+++ b/vpr/src/route/rr_graph.cpp
@@ -1380,6 +1380,9 @@ static void build_rr_sinks_sources(RRGraphBuilder& rr_graph_builder,
     int num_pins = type->num_pins;
     const std::vector<int>& pin_class = type->pin_class;
 
+    auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
+
     /* SINK and SOURCE-to-OPIN edges */
     for (int iclass = 0; iclass < num_class; ++iclass) {
         RRNodeId inode = RRNodeId::INVALID();
@@ -1492,7 +1495,7 @@ static void build_rr_sinks_sources(RRGraphBuilder& rr_graph_builder,
                             L_rr_node.add_node_side(inode, side);
 
                             // Sanity check
-                            VTR_ASSERT(L_rr_node.is_node_on_specific_side(inode, side));
+                            VTR_ASSERT(rr_graph.is_node_on_specific_side(RRNodeId(inode), side));
                             VTR_ASSERT(type->pinloc[width_offset][height_offset][side][L_rr_node.node_pin_num(inode)]);
                         }
                     }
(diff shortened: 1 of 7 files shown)


vpr/src/route/rr_graph.cpp
```diff
@@ -2900,8 +2896,7 @@ static std::vector<bool> alloc_and_load_perturb_opins(const t_physical_tile_type
     return perturb_opins;
 }
 
-static RRNodeId pick_best_direct_connect_target_rr_node(const t_rr_graph_storage& rr_nodes,
-                                                        RRNodeId from_rr,
+static RRNodeId pick_best_direct_connect_target_rr_node(RRNodeId from_rr,
```
tangxifan, line 2899: Since this function has been refactored and the old data structure ``rr_nodes`` has been shadowed, I suggest to replace the ``const t_rr_graph_storage& rr_nodes`` with ``const RRGraphView& rr_graph`` as the first argument.

vpr/src/route/rr_graph.cpp
```diff
@@ -2568,7 +2565,7 @@ static void build_unidir_rr_opins(RRGraphBuilder& rr_graph_builder,
         }
 
         /* Add in direct connections */
-        get_opin_direct_connections(rr_graph_builder, i, j, side, pin_index, opin_node_index, rr_edges_to_create, rr_nodes,
```
tangxifan, line 2571: Since the function has no use of ``rr_nodes``, it is time to put ``rr_graph_view`` as one of the argument. I suggest to put it right after the ``rr_graph_builder`` in the argument list.

tangxifan (changes requested):
@ethanroj23 Good progress. I am happy to see a few functions are now fully reworked. The old data structure has been completely shadowed in these functions. Therefore, I suggest to the remove the use of global variables in these functions, by putting the ``RRGraphView`` to the argument list.

vpr/src/route/rr_graph.cpp
```diff
@@ -1281,6 +1278,7 @@ static void build_bidir_rr_opins(RRGraphBuilder& rr_graph_builder,
                                  const int num_directs,
                                  const t_clb_to_clb_directs* clb_to_clb_directs,
                                  const int num_seg_types) {
+    const auto& rr_graph = g_vpr_ctx.device().rr_graph;
```
tangxifan, line 1281: Same comments as you see in the ``build_unidir_rr_opins()`` function. I think we can start propagate the use the of RRGraphView as an argument of these functions until reaching the ``build_rr_graph()`` function (To avoid too many code changes and impact too many files)

tangxifan (changes requested):
@ethanroj23 Just a few changes required due to the chain effect from adding RRGraphView to argument lists. We are almost there.

tangxifan:
@vaughnbetz I merged this since @ethanroj23 has addressed all the comments.
If anything wrong happened, we will actively solve the issues.

PR #1863, opened 2021-09-30
[Infra]: fix missing VTR time in parse_results 

sdamghan:
#### Description

While a new VPR memory usage stat is added to the run_vtr_flow output, the value of VTR elapsed time is no longer read by parse scripts due to a mismatch in the corresponding regex expressions.

#### Types of changes

#### Checklist:

### DIFF
--- a/vtr_flow/parse/parse_config/common/vtr_flow.txt
+++ b/vtr_flow/parse/parse_config/common/vtr_flow.txt
@@ -1,3 +1,3 @@
 #Flow Run-time Metrics
-vtr_flow_elapsed_time;vtr_flow.out;.* \(took (.*) seconds\)
+vtr_flow_elapsed_time;vtr_flow.out;.* \(took (.*) seconds, .*\)
 error;output.txt;error=(.*)


vaughnbetz:
Thanks.  Kokoro didn't run; trying a force restart.  Change looks fine ... could probably merge even without kokoro.

sdamghan:
@vaughnbetz Anytime. BTW, it seems there are continuous failures in all Kokoro runners caused by the following error: 
```
++ sudo apt-key add -
gpg: no valid OpenPGP data found.


[ID: 1004054] Build finished after 27 secs, exit value: 2


Warning: Permanently added 'localhost' (ED25519) to the list of known hosts.
[15:19:51] Collecting build artifacts from build VM
Build script failed with exit code: 2
```
Although I am not sure, I found a related [post](https://stackoverflow.com/questions/21338721/gpg-no-valid-openpgp-data-found) on the StackOverflow.

vaughnbetz:
Thanks.  From the stackoverflow post it seems this may be a permissions issue on kokoro. @mithro : if you can have someone look into it that would be great.
@sdamghan : that stack overflow post also suggests some alternative command formulations that might work.  If you have time and they look like potential solutions it would be good if you could try them out.

vaughnbetz:
Should the regex look for the trailing comma, or just skip it as part of the .* ?
It seems like not explicitly coding the "look for a comma" in the regex could make this more robust to future changes.

sdamghan:
> Should the regex look for the trailing comma, or just skip it as part of the .* ? It seems like not explicitly coding the "look for a comma" in the regex could make this more robust to future changes.

@vaughnbetz - This regex coding only looks for the elapsed time here. In the following, you would find a sample _vtr_flow.out_, since the VPR memory usage is already reported in the _parse_results.txt_, I just write a code to ignore whatever is after the elapsed time.
```
k4_N10_memSize16384_memData64/single_wire		OK (took 0.16 seconds, vpr run consumed 22.85 MiB memory)
```
I can change it to specifically search for a comma if you would like.

vaughnbetz:
Yes, I was asking about removing the comma from the regex (to make things less fragile).

sdamghan:
@vaughnbetz - I see, sorry for the misunderstanding. So, since the comma is connected to the "seconds," with regex is not possible to ignore it. Either we need to ignore entire "seconds," with `\(took (.*) .*\)` or whatever is after "seconds, " (the space after "," is the point here) as the current changes. Could you please let me know which one would you prefer?

vaughnbetz:
Ok let's leave it as is then.

Issue #1866, opened 2021-10-07
Collect parse_results*.txt as artifacts from CI to make it easier to update golden results


vaughnbetz:
Right now when we fail a QoR check in CI due to an expected QoR change, the developer needs to rerun the test on his/her machine and run parse_vtr_task.py --create_golden to make a new expected QoR file (golden_results.txt).

We can simplify this by bringing back the parse_results*.txt files (currently parse_results.txt and parse_results2.txt is also used in some tests), and then the developer could just run parse_vtr_task.py --create_golden using those files.

#### Proposed Behaviour
Add parse_results*.txt as an artifact to bring back from CI.

#### Current Behaviour
We don't bring back those files.

#### Possible Solution
Update the file list for artifact capture.

#### Context
It's common to have to update QoR files; we should streamline the process.  This would also avoid issues where the CI machines are slower or faster or have libraries that take more memory etc. than someone's server, by generating the new results in CI itself.

vaughnbetz:
This was completed ... marking closed.

PR #1867, opened 2021-10-07
Collected parse_results*.txt as artifacts from CI run

ganeshgore:
#### Description
This update collects `parse_results*.txt` from CI regression run, which can be analyzed locally or used to replace existing golden results. 

#### Related Issue

Addressing issue #1866 

#### Motivation and Context

_"It's common to have to update QoR files; we should streamline the process. This would also avoid issues where the CI machines are slower or faster or have libraries that take more memory etc. than someone's server, by generating the new results in CI itself."_ - from issue #1866

To update golden results, you  might be able to do the following
1. Download `*_golden_results` artifacts and uncompress
2. *run `rsync --recursive downloaded_artifact/ local_task_directory/`
3. *run `parse_vtr_task.py --create_golden`
4. Create a new commit with updated golden results files
[*Remember to clean local run directory, `parse_vtr_task` collects results from the most recent run, and Step. 2 will update `run001` by default]

You can check the test CI run here
https://github.com/ganeshgore/vtr-verilog-to-routing/actions/runs/1317429846#artifacts

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -137,23 +137,23 @@ jobs:
         ./.github/scripts/build.sh
         ./run_reg_test.py ${{ matrix.suite }} -show_failures -j2
 
-    - name: Upload artifact
+    - name: Upload regression run files
       uses: actions/upload-artifact@v2
       with:
-        name: ${{matrix.name}}_regression_log
+        name: ${{matrix.name}}_run_files
         path: |
-          vtr_flow/**/*.log
           vtr_flow/**/*.out
           vtr_flow/**/*.blif
           vtr_flow/**/*.p
           vtr_flow/**/*.net
           vtr_flow/**/*.r
 
-    - name: Upload golden results
+    - name: Upload regression results
       uses: actions/upload-artifact@v2
       with:
-        name: ${{matrix.name}}_golden_results
+        name: ${{matrix.name}}_results
         path: |
+          vtr_flow/**/*.log
           vtr_flow/**/parse_results*.txt
 
   Sanitized:


tangxifan:
@ganeshgore Compared to previous PR, what are the new packages here?

image

ganeshgore:
everything named as `*_golden_results`

umarcor:
> I am not sure how to know when
> 
> > ... the golden results need to be updated ...
> 
> My understanding was it depends on the user to update or not.

@ganeshgore, see #1866:

> Right now when we fail a QoR check in CI due to an expected QoR change, the developer needs to rerun the test on his/her machine and run parse_vtr_task.py --create_golden to make a new expected QoR file (golden_results.txt).
> 
> We can simplify this by bringing back the parse_results*.txt files (currently parse_results.txt and parse_results2.txt is also used in some tests), and then the developer could just run parse_vtr_task.py --create_golden using those files.

umarcor:
> I'd like updating golden results to be fast after a failure,

@vaughnbetz can you please be more explicit about this? Assume that the reader is not aware of the internals of VTR. What does it mean "after a failure"? Which tests/jobs do we need to monitor in order to decide that "this is a failure that needs to trigger a golden results update" vs "this is a legit failure of the job/workflow which does not require further action in regard to golden results".

vaughnbetz:
I don't think we can make this decision automatically.  A QoR update happens when:
1. Someone looks at specific failures and concludes the QoR bounds were too tight or lucky (mostly on small circuits) and hence updating them is OK.
2. Someone has changed an algorithm, and expects better QoR or a new trade-off (maybe a new feature for moderately higher CPU, or maybe a CPU reduction that takes us outside the current QoR bounds, ...).  After collecting results, discussing with others, etc. the trade-off is ruled a good one and the QoR data is updated as part of the PR that puts the new code in.  Easiest to update this by simply making a PR, seeing what the QoR failures are (if any) and getting the CI artifact data to make the new golden results.

umarcor:
@vaughnbetz, thanks!

@ganeshgore, according to the last explanations, I take back my previous suggestion. It is ok to have this file as a separated artifact.

PR #1879, opened 2021-10-15
Add a missing operator.

hzeller:
This turned out to be needed with a different std::sort
algorithm.

Signed-off-by: Henner Zeller <hzeller@google.com>


hzeller:
This is already upstream now. Closing.

Issue #1831, opened 2021-08-17
Documentation of Yosys+Odin-II Implementation


sdamghan:
#### Context

Full documentation of Yosys+Odin-II implementation, including new features, instructions to use, added libraries, testing process, etc., should be added to VTR docs.

#### Related PR / Issues
PR #1798 
Issue #1830

PR #1798, opened 2021-07-08
Yosys+Odin

sdamghan:
#### Description

This PR adds the functionality of using Odin's technology mapping for a Yosys generated BLIF file in coarse grain format. Basically, the feature is added according to the following flow:
1. **Yosys**: A hardware design will be given to Yosys. Using the commands mentioned in [synth.tcl](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d4fabdfd764a44379a9358cbb5edfd9cca5d9e7c/ODIN_II/regression_test/tools/synth.tcl) script, Yosys generates a coarsen (not exploded into low-level logic) BLIF file. Technically, the generated BLIF file should have Yosys logic cells. such as $and, $logic_and, $add, $sub, $div, $pow, $mul, $mem and etc.
2. **BLIF Reader**: Odin-II is fed by the generated BLIF file as the main design. At this point, other options like architecture files can also be given to Odin-II as before.
3. **BLIF Elaboration**: The BLIF file will pass to the BLIF Elaboration phase, which basically, makes the netlist compatible with Odin-II's partial mapping. Worth noting to mention that some Yosys logic cells such as simple DFF or bitwise operations will be passed to the partial mapping phase intact. The first critical stage of the Yosys+Odin happens in the BLIF Elaboration phase since some logic cells like $pow, $div or $mul will be exploded into smaller components such as adders or shifters. Moreover, the block memories are being detected at this point; however, the hard-block mapping or memory-related optimization process will happen in the mid-phase between BLIF Elaboration and Partial mapping. The third phase for other logic cells, like logical nodes, will only include input/output ports reordering to make them compatible with the partial mapping structure.

[...]

PR #1885, opened 2021-10-18
Yosys+Odin-II documentation

sdamghan:
#### Description

Providing the documentation for newly integrated the Yosys+Odin-II synthesizer.
Related PR: #1798 

#### Related Issue

Issue: #1831

#### Types of changes

#### Checklist:


sdamghan:
> Generally LGTM.
> 
> There might be some more detailed high level information around:
> 
>     * the rationale behind using yosys combined with ODIN
> 
>     * what tools does what exactly. For instance, it is not clear to me which tool is performing synthesis.


Thanks @acomodi , I have added to the explanation of tools' functionality and their relation to enrich the context. Technically, the synthesis process happens using both Yosys and Odin-II. Yosys perform the first phase, which includes the coarse-grained synthesis and elaboration. The Yosys output is a coarse-grained BLIF file that is not technology mapped. So here, once Odin-II parses the BLIF file, it performs the BLIF elaboration on coarse-grained components, and as the last step, it performs partial mapping with the FPGA architecture awareness on the netlist. 
One of the main ideas behind the combination of these two tools was how Yosys flow works. Basically, it forces the user to decide the discrete circuit implementation manually. Odin-II partial mapper has quite a few options to perform trade-offs on hard/soft logic inference. As a result of this combination, the control over the partial mapping process to the user through runtime parameters became available.

PR #1886, opened 2021-10-19
Update cluster gains once per cluster

sfkhalid:
#### Description
Currently during clustering the gain values are updated once for each level in a t_pb - from the root node down to the child node. However, the gain information only needs to be stored for the root pb, - the gain stored for the child pbs is never used. This pull request fixes the excess updates done during clustering.


vaughnbetz:
@sfkhalid : there are warnings in the code that need to be fixed to pass CI.

sfkhalid:
> @sfkhalid : there are warnings in the code that need to be fixed to pass CI.

Yep, I'm looking into fixing those now.

PR #1904, opened 2021-10-28
RRGraphView rr_segment() Implementation

umariqbal-rs:
#### Description
In this PR, I am  implementing RRGraphView::rr_segments(PRSegmentId Seg_Id ) throughout VTR. Every time device_ctx.rr_segment[Seg_Id]  was called has been replaced with rr_graph.rr_segment(Seg_Id). In order to do this, I followed a pattern similar to that in the last PR.

#### Motivation and Context
These changes are a continuation of the RRGraphView API implementation effort.

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
#### Issues and Errors:
As we discussed  , facing some issues and errors below is shown screenshot of errors and issues.
image

PR #1906, opened 2021-11-03
RRGraphView::rr_segments() Implementation

umariqbal-rs:
#### Description
In this PR, I am implementing `RRGraphView::rr_segments(PRSegmentId Seg_Id)` throughout VTR. Every time `device_ctx.rr_segments[Seg_Id]` was called has been replaced with `rr_graph.rr_segments(Seg_Id)`. In order to do this, I worked in  a similar way as did in previous PRs for `RRGraphView` APIs implementation.

#### Motivation and Context
These changes are a continuation of the `RRGraphView` API implementation effort.
#### Types of changes

#### Checklist:

#### Issues and Errors:
I have created API for `rr_segments` and passed `RRSegmentId` StrongId as argument and replaced in respective files where it is needed but we stuck at  [`Segments = Arch->Segments`](https://github.com/RapidSilicon/vtr-verilog-to-routing/blob/c24d287309b06dce7a70a4204c66c73d38126c52/vpr/src/base/SetupVPR.cpp#L167)   which points to  [`std::vector<t_segment_inf> Segments;`](https://github.com/RapidSilicon/vtr-verilog-to-routing/blob/c24d287309b06dce7a70a4204c66c73d38126c52/libs/libarchfpga/src/physical_types.h#L1748)  where we can't pass `RRSegmentId` StrongId, because that is outside of the vpr/ directory. And that pop out many warnings and errors, below is shown screenshot of errors and warnings. 

image


umariqbal-rs:
@tangxifan  I have changed what you suggested in #1904 and a lot, could you review it and suggest me what I should do, because  I am still facing issues.

tangxifan:
> @tangxifan I have changed what you suggested in #1904 and a lot, could you review it and suggest me what I should do, because I am still facing issues.

@umariqbal-rs I looked into the code changes. Please note that there are two copies of ``segments``

- One is the ``Segments`` in the architecture struct. This is the original copy, where the ``RRSegmentId`` should **NOT** be applied.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/libs/libarchfpga/src/physical_types.h#L1748

- Another is the ``rr_segments`` in the ``DeviceContext``. This is a modified copy where additional segments are added due to the RRGraph changes. This is where should apply ``RRSegmentId``

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/base/vpr_context.h#L177

To avoid large code changes, I think what we should do it to
- Find the functions that creates the modified copy and apply ``RRSegmentId`` there
- Modify the ``rr_segments`` in the ``DeviceContext`` to use ``RRSegmentId`` (You have already done so)
- And APIs to ``RRGraphView`` which can return segments by a given id (You have already done so)
- Adapt other source files related.

Let me know what you think.

umariqbal-rs:
@tangxifan  
   ` device_ctx.rr_segments = segment_inf;`[https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/route/rr_graph.cpp#L448](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/route/rr_graph.cpp#L448)
Here `segment_inf` is assigned to `device_ctx.rr_segments`. `segment_inf`   which is  called from architecture struct, where we can't pass `RRSegmentId`.Where get stuck.

tangxifan:
@umariqbal-rs For this line, you can simply write the following codes

```
// If you want to save memory, use reserve()
device_ctx.rr_segments.reserve(num_segments);
// If you can use range-based loop, it will be better
for (int iseg = 0; iseg < num_segments; ++iseg) {
  device_ctx.rr_segments.push_back(segment_inf[iseg]);
}
```

umariqbal-rs:
@tangxifan 
device_ctx.rr_segments = segment_inf; 
is used at two places
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/route/rr_graph.cpp#L448 
and 
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/route/rr_graph_reader.cpp#L50
You suggested me to use 

// If you want to save memory, use reserve()
device_ctx.rr_segments.reserve(num_segments);
// If you can use range-based loop, it will be better
for (int iseg = 0; iseg < num_segments; ++iseg) {
  device_ctx.rr_segments.push_back(segment_inf[iseg]);
}

I have used it like this but 

1.  How we will get ` num_segments` here?
    I have got it like this    ` size_t num_segments = segment_inf.size();`.
    is it right or not? 
 2.  In function `void load_rr_file`
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/route/rr_graph_reader.cpp#L73
   we have `arch->strings` which refer to arch in   [  `const t_arch* arch;`](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/base/vpr_context.h#L230) which points to architecture struct.
And i got following errors

tangxifan:
@umariqbal-rs 

Num_segments
-----------------
This is correct. 
  ``size_t num_segments = segment_inf.size();``

If you look at the ``segment_inf``,  it is a ``std::vector``.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/libs/libarchfpga/src/physical_types.h#L1748

You can find methods from C++ documentation: https://www.cplusplus.com/reference/vector/vector/

RRGraph Reader
------------------

If you look at the RrGraphSerializer data structure at 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/295393776e11d9f97c762827e96267992e577e8b/vpr/src/route/rr_graph_uxsdcxx_serializer.h#L252-L273

You can see that the ``device_ctx.arch->strings`` is not access the ``t_arch``. It is asking for an string inside the ``t_arch``.
What you should look into is the ``const std::vector<t_segment_inf>& segment_inf`` at LINE 268.
You need to change it to ``RRSegmentId``

PR #1893, opened 2021-10-25
VPR: interchange: add initial interchange architecture frontend

acomodi:
#### Description

This PR adds an initial implementation of the interchange format architecture frontend, that is an implementation that aims to be parallel to the XML architecture reading step, but which reads directly from an FPGA interchange device.

NOTE: this PR is based on top of https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1870, of which, many of the line changes come from the interchange-schema subtree.

Alongside with the initial (and partial) architecture reading step, there is also an addition to the option list to read-in the Logical Netlist and write the Physical Netlists as well. Possibly, these may be removed until the support is complete (at least partially).

#### Motivation and Context

This is the first step to get a wider adoption of the [Interchange Format](https://github.com/chipsalliance/fpga-interchange-schema) and provide yet another tool able to use the format (alongside with nextpnr and RapidWright).

#### How Has This Been Tested?

To speed-up the integration process of the various implementation stages, smaller PRs would be required, and also tested, to verify the added functionality. Therefore, unit tests are being developed as well, to quickly check the contents of the VTR's internal structs.

There is an addition of a test architecture, already serialized and gzipped in the cap'n'proto file format.

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1870, opened 2021-10-12
CMake: add initial interchange schema sources compilation

acomodi:
#### Description

This PR adds a very initial step towards the FPGA interchange adoption and integration within VTR.

This step currently adds the [FPGA interchange schema definitions](https://github.com/chipsalliance/fpga-interchange-schema) as a submodule and, if the cmake cache variable is defined, the various capnp schemas are compiled and their srcs and headers generated, ready to be used in the VTR codebase.

#### How Has This Been Tested?

For now, the build process and cmake are enough to catch regressions when compiling the interchange schema files

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1907, opened 2021-11-03
FPGA Interchange: Initial interchange device reading

acomodi:
#### Description

This PR supersedes both https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1893 and https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1870

It adds the very first usage of the interchange schema, by providing an option to read in the device primitive library filling in the device models.

The PR contains three types of changes/additions:

1. Addition of FPGA interchange device reading (and corresponding C++ unit tests)
2. Addition of switches to transition from normal XML to Interchange format
3. CMake infrastructure changes to allow schema compilation

#### Motivation and Context

Incremental addition of the Interchange format usage capability

C++ tests

#### Types of changes

#### Checklist:


vpr/src/base/read_options.cpp
```diff
@@ -2472,7 +2482,7 @@ void set_conditional_defaults(t_options& args) {
     }
 
     if (args.SDCFile.provenance() != Provenance::SPECIFIED) {
-        //Use the full path specified in the original circuit name,
+        //Use the full path specified in the originial circuit name,
```
tangxifan, line 2485: Typo: originial -> original

vpr/src/base/read_options.cpp
```diff
@@ -1147,6 +1150,13 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         "If the implementation is illegal analysis can be forced by explicitly\n"
         "specifying the --analysis option.");
 
+    auto& FPGAInterchage_grp = parser.add_argument_group("FPGA Interchange options");
```
tangxifan, line 1153: Should we consider an option like ``--arch_format`` like the option ``--circuit_format`` which 
 is more flexible? 

https://docs.verilogtorouting.org/en/latest/vpr/command_line_usage/#cmdoption-vpr-circuit_format

vpr/src/base/SetupVPR.cpp
```diff
@@ -109,11 +110,20 @@ void SetupVPR(const t_options* Options,
...
-                    Arch,
-                    device_ctx.physical_tile_types,
-                    device_ctx.logical_block_types);
+        if (Options->FPGAInterchangeDevice) {
+            FPGAInterchangeReadArch(Options->FPGAInterchangeDeviceFile().value().c_str(),
+                                    TimingEnabled,
+                                    Arch,
+                                    device_ctx.physical_tile_types,
+                                    device_ctx.logical_block_types);
+            VTR_LOG("Use FPGA Interchange device\n");
```
tangxifan, line 119: Should we show the log before calling the ReadArch() function? It will be easier for debugging

libs/libarchfpga/src/read_common_func.h
```diff
@@ -0,0 +1,19 @@
+#ifndef READ_COMMON_FUNC_H
+#define READ_COMMON_FUNC_H
+
+#include "arch_types.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* function declarations */
```
tangxifan, line 10: I think the common functions should be splitted from the large source files.
Also I believe this is a good time to think about what common functions we need.
My suggestion is that we need
- Check functions, such as ``check_models()`` as you have done, as well as other ``check_arch()``. If we agree, can we rename this file to ``check_arch.h``. We can also add comments to these functions so that it is easer for other developers to follow-up

vpr/src/base/read_options.cpp
acomodi: Fixed

vpr/src/base/SetupVPR.cpp
acomodi: Done

libs/libarchfpga/src/read_common_func.h
acomodi: Yeah, this makes completely sense. Given the huge amount of stuff in the XML arch reader that is not strictly corresponding to parsing the file itself, I think as well that it is good to extract generic functions in separate files.

I have started to address this in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1907/commits/e9ba680b7de5c3350183aefa0e7e8a3689b64abb and put all the check functions in an `arch_check` file.

vpr/src/base/read_options.cpp
acomodi: Yes, makes sense, I have added the corresponding change

acomodi:
@tangxifan Thanks for the review, I have addressed the comments and it should be good to go for a second review pass

tangxifan (approved):

acomodi:
CI went green, merging

PR #1914, opened 2021-11-17
Refactor build/dependency documentation + new markdown parser (myst-parser)

jgoeders:
# Overview

This PR makes two changes at once (my apologies, it would be better if they were separate, but they become quite intertwined while getting this to all work):
  1. This PR provides a few updates to ease the environment setup for new users by refactoring the build documentation that was spread across multiple different pages.
  2. Migrates the Sphinx markdown processor from recommonmark (now deprecated) to myst-parser (the newly official markdown processor).

# Refactoring Build Documentation

Multiple changes were made to the instructions on building and setting up VTR.

1.I've moved the required packages into a install_apt_packages.sh file. The list of packages has been changed somewhat, and I've removed old packages that we aren't using anymore.  I've tested on a fresh Ubuntu 18 and 20 system.

2. I've created a `make evn` target to setup a Python environment.

    With the move to using Python for the runner scripts, and the inclusion of a [requirements.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/requirements.txt) file of required Python packages, I think it would be nice if we provided an automated mechanism to create a Python virtual environment and install the necessary packages.  To me this seems like good practice versus having users modify their system Python version, especially since our requirements.txt is requiring specific versions of certain packages.  

3. I've refactored the build documentation. We currently have build documentation in three places, and it's quite inconsistent about what information is where:
  * [The Quick Start guide](https://docs.verilogtorouting.org/en/latest/quickstart/)
  * [Building VTR](https://docs.verilogtorouting.org/en/latest/building/building/)
  * [Installing VTR](https://docs.verilogtorouting.org/en/latest/vtr/install_vtr/)

With these changes:
 * The quick start guide now includes information about setting up necessary dependencies
 * All required build information is moved to a single "Building VTR" page.
 * Optional build information (debug mode, sanitizers, other platforms, etc), is moved to a single "Optional Build Information" page.
  
# Changing Markdown Processor

While refactoring this documentation I discovered issue #1941 where certain internal markdown links were broken.  This seemed to be a byproduct of our efforts to fix links between the Sphinx documentation and our top-level Markdown files (ie CONTRIBUTING.md).

Rather than trying to fix our broken workaround, which was still using recommonmark (deprecated Markdown processor), I migrated our Sphinx build to use the newer (and official) myst-parser.  At first it wasn't obvious how to support links between our top-level Markdown files and our Sphinx docs, but I managed to develop a [solution](https://github.com/executablebooks/MyST-Parser/issues/361#issuecomment-1006873840).

This involved replacing symlinks that we had from our `docs/src` to the top-level Markdown files, and instead use some stub files with myst-parser's *relative-docs* feature.  Example stub file:

    ```{include} ../../BUILDING.md
    :relative-docs: doc/src
    ```

This approach seems to fix all of our linking issues.  I've also updated several broken links as I've cleaned things up.

#### Related Issue
#1941 

I've compared the Sphinx build log outputs before and after this change to ensure no new errors/broken links were created.

#### Types of changes

#### Checklist:


jgoeders:
This fixes #1941

jgoeders:
@vaughnbetz @acomodi 

This is ready for review.

BUILDING.md
```diff
@@ -1,255 +1,49 @@
...
-* GCC/G++: 7, 8, 9, 10, 11
-* Clang/Clang++: 6, 7, 10
 
-Other compilers may work but are untested (your milage may vary).
+You will also need several Python packages.  You can optionally install and activate a Python virtual environment so that you do not need to modify your system Python installation:
 
-## Unix-like ##
-For unix-like systems we provide a wrapper Makefile which supports the traditional `make` and `make clean` commands, but calls CMake behind the scenes.
+    make env
+    source .venv/bin/activate
```
vaughnbetz, line 15: Should the source .venv/bin/activate command be done by make env? (one less command to run).

BUILDING.md
```diff
@@ -1,255 +1,49 @@
...
- * A modern C++ compiler supporting C++14 (such as GCC >= 4.9 or clang >= 3.6)
+    pip install -r requirements.txt
 
-For the VPR GUI you need:
- * Cairo
- * FreeType
- * Xft (libXft + libX11)
- * fontconfig
- * libgtk-3-dev
+**Note:** If you chose to install the Python virtual environment, you will need to remember to activate it on any new terminal window you use, before you can run the VTR flow or regressions tests.
```
vaughnbetz, line 21: Might be good to give the exact command to run to activate the environment here, maybe in brackets.  Especially useful if the activate gets moved into the make env command, so you only have to run this command in other windows.

doc/src/quickstart/index.rst
```diff
@@ -14,6 +14,24 @@ The first step is to `download VTR <https://verilogtorouting.org/download/>`_ an
...
+
+    > ./install_apt_packages.sh
+
+Then, to install the required Python packages within a new virtual environment:
+
+.. code-block:: bash
+
+    > make env
+    > source .venv/bin/activate
+    > pip install -r requirements.txt
```
vaughnbetz, line 32: Should make env run the next two commands too (fewer commands to run)?

vaughnbetz: Or are you trying to show them how to do this system wide or in a virtual envronment?  If so, suggest putting a command on the two virtual environment lines.

make env    # optional: install python virtual environment
source .venv/bin/activate   # optional: activate python virtual environment
pip install -r requirements.txt   # install python packages (in virtual environment if prior commands run, system wide otherwise)

vaughnbetz (commented):
Looks great; just a few comments.

BUILDING.md
jgoeders: Unfortunately there's no way to affect the environment of the caller's process (the user's terminal) from within a Makefile command.  https://stackoverflow.com/a/7508273

What I've seen done in other projects (including several Symbiflow projects) is to activate the environment from within a Makefile target, which would have effect for only that target.  Since we don't run the VTR flow using Makefile targets (but rather have users directly invoke the Python scripts) that doesn't really help.

vaughnbetz: OK, that's a good reason :).

BUILDING.md
jgoeders: Fixed.

doc/src/quickstart/index.rst
jgoeders: Can't collapse commands as noted above, but I did add the comments.  Good idea.

acomodi (approved):
LGTM

vaughnbetz (approved):
LGTM

PR #1916, opened 2021-11-18
RRGraphView::node_first_edge() Implementation

Raza-jafari-RS:
#### Description
In this PR, I have implemented RRGraphView::node_first_edge() throughout VTR. Every time device_ctx.graph.first_edge() was called has been replaced with rr_graph.node_first_edge( RRNodeId (index)). In order to do this, I followed a pattern similar to that in the last PR.

#### Motivation and Context

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/device/rr_graph_view.h
+++ b/vpr/src/device/rr_graph_view.h
@@ -123,7 +123,7 @@ class RRGraphView {
     inline short node_yhigh(RRNodeId node) const {
         return node_storage_.node_yhigh(node);
     }
-       /** @brief Get the first edge of resource node. This function is inlined for runtime optimization. */
+    /** @brief Get the first edge of resource node. This function is inlined for runtime optimization. */
     inline RREdgeId node_first_edge(RRNodeId node) const {
         return node_storage_.first_edge(node);
     }
--- a/vpr/src/device/rr_graph_view.h
+++ b/vpr/src/device/rr_graph_view.h
@@ -123,10 +123,12 @@ class RRGraphView {
     inline short node_yhigh(RRNodeId node) const {
         return node_storage_.node_yhigh(node);
     }
+
     /** @brief Get the first edge of resource node. This function is inlined for runtime optimization. */
     inline RREdgeId node_first_edge(RRNodeId node) const {
         return node_storage_.first_edge(node);
     }
+
     /** @brief Get the length (number of grid tile units spanned by the wire, including the endpoints) of a routing resource node.
      * node_length() only applies to CHANX or CHANY and is always a positive number
      * This function is inlined for runtime optimization.
@@ -144,8 +146,6 @@ class RRGraphView {
                  && (node_xlow(node) == -1) && (node_ylow(node) == -1)
                  && (node_xhigh(node) == -1) && (node_yhigh(node) == -1));
     }
-
-    
 
     /** @brief Check if two routing resource nodes are adjacent (must be a CHANX and a CHANY). 
      * This function is used for error checking; it checks if two nodes are physically adjacent (could be connected) based on their geometry.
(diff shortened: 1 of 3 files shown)


Raza-jafari-RS:
@tangxifan
I have created node_first_edge() API in rr_graph_view.h as below
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/c812542d7a098e614aace3b0b5f16595b3c1b5c4/vpr/src/device/rr_graph_view.h#L127 
And I made changes in respective files.
Can you please review them and suggest me how to move forward.

tangxifan:
@Raza-jafari-RS Please clear the code format issue

vpr/src/device/rr_graph_view.h
```diff
@@ -123,7 +123,10 @@ class RRGraphView {
     inline short node_yhigh(RRNodeId node) const {
         return node_storage_.node_yhigh(node);
     }
-
+    /** @brief Get the first edge of resource node. This function is inlined for runtime optimization. */
```
tangxifan, line 126: Can you check if this API gets the first incoming edge or the outcoming edge of a node? I remember it is for outcoming edges. Please confirm. We should this to the comments.

tangxifan (changes requested):

Raza-jafari-RS:
@tangxifan 
I have created node_last_edge() API in rr_graph_view.h as below
https://github.com/RapidSilicon/vtr-verilog-to-routing/blob/77b5a1375b318e149a3eda9f49ecc4ed9f438d59/vpr/src/device/rr_graph_view.h#L132
And I made changes in respective files.
Can you please review them and suggest me how to move forward.

tangxifan:
@Raza-jafari-RS Looks o.k. now. Please also adapt the comments for API last_edge(). Then I think the PR is ready to go.

vpr/src/device/rr_graph_view.h
Raza-jafari-RS: It will get the first out coming edge of a node.

Raza-jafari-RS:
QoR tests are complete and quick summary is given below. QoR comparison files are also attached.
**vtr_reg_qor_chain_depop test:**
Flow run time is not added in the results.
critical path delay 0 % on average.
Peak memory usage 4.12 % on average
**titan_quick_qor test:**
Flow run time is not added in the results.
Peak memory usage - 0.01% on average
[vtr_reg_qor_chain_depop.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7640759/vtr_reg_qor_chain_depop.xlsx)
[titan_quick_qor.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7640760/titan_quick_qor.xlsx)

vpr/src/route/rr_graph_util.cpp
```diff
@@ -98,6 +98,7 @@ int seg_index_of_sblock(int from_node, int to_node) {
 void reorder_rr_graph_nodes(const t_router_opts& router_opts) {
     auto& device_ctx = g_vpr_ctx.mutable_device();
     auto& graph = device_ctx.rr_nodes;
+    auto& rr_graph = device_ctx.rr_graph;
```
tangxifan, line 101: After a discussion with @vaughnbetz , we may consider to move this function as an internal method of RRGraphView/RRGraphbuilder.

tangxifan:
@Raza-jafari-RS We get the green light to merge. A follow-up PR may be required. Can you investigate
- How many external variables are required by the ``reorder_rr_graph_nodes()`` function? It seems to me that it needs some parameters from router options, but it is mainly manipulating internal data of RRGraph.
- If it is mainly on the internal data. It is more reasonable to move the main body of the function to be a method of the ``RRGraphBuilder`` class. We can keep the ``reorder_rr_graph_nodes()`` here. But inside it, we just call the API like 

```
rr_graph.reorder_nodes(int <router_parameter> ...)
```

Let me know if it makes sense to you.

PR #1925, opened 2021-11-26
libs: arch: interchange: add basic device information

acomodi:
#### Description

This PR adds basic support for the architecture reading:
  - Device
  - Layout
  - Switches

#### Motivation and Context

This is part of the FPGA Interchange support in VTR

Catch2 testing framework

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/read_fpga_interchange_arch.cpp
+++ b/libs/libarchfpga/src/read_fpga_interchange_arch.cpp
@@ -34,6 +34,72 @@ using namespace DeviceResources;
 using namespace LogicalNetlist;
 using namespace capnp;
 
+static float get_corner_value(Device::CornerModel::Reader model, const char* speed_model, const char* value) {
+    bool slow_model = std::string(speed_model) == std::string("slow");
+    bool fast_model = std::string(speed_model) == std::string("fast");
+
+    bool min_corner = std::string(value) == std::string("min");
+    bool typ_corner = std::string(value) == std::string("typ");
+    bool max_corner = std::string(value) == std::string("max");
+
+    if (!slow_model && !fast_model) {
+        archfpga_throw("", __LINE__,
+                       "Wrong speed model `%s`. Expected `slow` or `fast`\n", speed_model);
+    }
+
+    if (!min_corner && !typ_corner && !max_corner) {
+        archfpga_throw("", __LINE__,
+                       "Wrong corner model `%s`. Expected `min`, `typ` or `max`\n", value);
+    }
+
+    bool has_fast = model.getFast().hasFast();
+    bool has_slow = model.getSlow().hasSlow();
+
+    if (slow_model && has_slow) {
+        auto half = model.getSlow().getSlow();
+        if (min_corner && half.getMin().isMin()) {
+            return half.getMin().getMin();
+        } else if (typ_corner && half.getTyp().isTyp()) {
+            return half.getTyp().getTyp();
+        } else if (max_corner && half.getMax().isMax()) {
+            return half.getMax().getMax();
+        } else {
+            if (half.getMin().isMin()) {
+                return half.getMin().getMin();
+            } else if (half.getTyp().isTyp()) {
+                return half.getTyp().getTyp();
+            } else if (half.getMax().isMax()) {
+                return half.getMax().getMax();
+            } else {
+                archfpga_throw("", __LINE__,
+                               "Invalid speed model %s. No value found!\n", speed_model);
+            }
+        }
+    } else if (fast_model && has_fast) {
+        auto half = model.getFast().getFast();
+        if (min_corner && half.getMin().isMin()) {
+            return half.getMin().getMin();
+        } else if (typ_corner && half.getTyp().isTyp()) {
+            return half.getTyp().getTyp();
+        } else if (max_corner && half.getMax().isMax()) {
+            return half.getMax().getMax();
+        } else {
+            if (half.getMin().isMin()) {
+                return half.getMin().getMin();
+            } else if (half.getTyp().isTyp()) {
+                return half.getTyp().getTyp();
+            } else if (half.getMax().isMax()) {
+                return half.getMax().getMax();
+            } else {
+                archfpga_throw("", __LINE__,
+                               "Invalid speed model %s. No value found!\n", speed_model);
+            }
+        }
+    }
+
+    return 0.;
+}
+
 struct ArchReader {
   public:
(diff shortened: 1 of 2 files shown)


libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -163,6 +234,220 @@ struct ArchReader {
...
+            single.owned_meta = std::make_unique<t_metadata_dict>(data);
+            single.meta = single.owned_meta.get();
+            grid_def.loc_defs.emplace_back(std::move(single));
+        }
+
+        arch_->grid_layouts.emplace_back(std::move(grid_def));
+    }
+
+    void process_device() {
+        /*
```
tangxifan, line 283: Make sense. Should we add ``FIXME`` or ``TODO`` to highlight that there are some pending work to do?

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -163,6 +234,220 @@ struct ArchReader {
...
+            for (auto pip : tile_type.getPips()) {
+                wire_names.insert(wires[pip.getWire0()]);
+                wire_names.insert(wires[pip.getWire1()]);
+            }
+        }
+        int num_seg = wire_names.size();
+        arch_->Segments.resize(num_seg);
+        uint32_t index = 0;
+        for (auto i : wire_names) {
+            // Use default values as we will populate rr_graph with correct values
```
tangxifan, line 432: Does it mean that we will always require users to provide a RRGraph file so that we will run the VPR in a correct way?
If we deposit all the bi-directional segments, I am not sure how it will impact the type of RRGraph: 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/f92cfebaff5235224fe2f970c694ebca6daf7e8c/vpr/src/route/rr_graph.h#L12-L18

If it will cause the graph type to be always BIDIR, will it cause any problem when we parse a UNIDIR routing resource graph from a file?

Just need a double check here and identify any expected errors.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -163,6 +234,220 @@ struct ArchReader {
...
+
+        grid_def.width = grid_def.height = 0;
+        for (auto tile : tileList) {
+            grid_def.width = std::max(grid_def.width, tile.getCol() + 1);
+            grid_def.height = std::max(grid_def.height, tile.getRow() + 1);
+        }
+
+        grid_def.grid_type = GridDefType::FIXED;
+        std::string name = std::string(ar_.getName());
+        if (name == "auto") {
```
tangxifan, line 253: Does it means that FPGA interchangeable layout will not support auto-sizing layout as VPR does nowadays? Should we clarify it in documentation? And should we comment here?

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -34,6 +34,72 @@ using namespace DeviceResources;
 using namespace LogicalNetlist;
 using namespace capnp;
 
+static float get_corner_value(Device::CornerModel::Reader model, const char* speed_model, const char* value) {
```
tangxifan, line 37: The corner values are something new. Should we add sufficient comments about it models the ``speedgrade`` and ``process corners``? Also should we comment the strategy to find the timing value under a given speed grade and a corner?

tangxifan (changes requested):
@acomodi Look o.k. on code changes. Just some clarification needed.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: In general, the RR graph is provided already within the device database (which may as well be split in the future from the rest of the architecture data).

Rather than a complete VPR-like RR graph, the interchange rr graph is a collection of nodes and PIPs, and we are working towards an Interchange RR graph reader, to read in the information from the interchange device data and build the VPR RR graph.

The [interchange schema](https://github.com/chipsalliance/fpga-interchange-schema/blob/53e3feda4628e39e939ce8d2e4add72025809d42/interchange/DeviceResources.capnp#L285-L297) can specify the directionality of a PIP and therefore whether some nodes are `UNIDIR` or `BIDIR`.

For the time being, the implementation proposed is to handle a simpler case, which can surely be made more complex to handle different kinds of graphs later on.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: At the moment, the interchange specifies fixed layout only architectures, and allowing for auto-sizing might be doable to allow for experimentation on new archs, but for the time being we may just restrict it to disallow `auto` sizes.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: Added a `FIXME`

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: Added some more comments around this, also redirecting to the schema where there is a description of the timing model structure.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
tangxifan: Make sense to me. I guess you have already tried some RRGraph which have both ``UNIDIR`` and ``BIDIR`` segments before. Are they always defined as ``GRAPH_BIDIR``? If so, I believe it is fine.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
tangxifan: That is o.k. We just need to document it well then.

tangxifan (approved):

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: Yeah, I need to double check, but for instance, the Series7 graph in SymbiFlow is `BIDIR`, but IIRC we treat all the segments as `BIDIR`, actually. We'll experiment on whether adding support also for `UNIDIR` segments in a graph defined as `BIDIR`

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: Ok, added some in-code comments about this

acomodi:
All required comments were added and CI went green on all tests, merging.

PR #1926, opened 2021-11-29
libs: arch: interchange: add logical block and physical tiles support

acomodi:
#### Description

This PR builds on top of #1924 and #1925.

It does the following:
- moves some checking and common functions from the XML reader to the utility source file
- adds basic support for reading in the physical tiles
- adds basic support for reading the logical blocks (e.g. pb_types)

#### Motivation and Context

Improve support for the interchange reading in VTR

#### How Has This Been Tested?

Catch2 tests framework

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

PR #1927, opened 2021-11-29
WIP: libs: arch: interchange: cell bel mapping implementation

acomodi:
#### Description

This PR builds on top of https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1926.

The changes include:
- Improvement of the cell to bel mapping, or in other words the mapping between a model primitive and its pb type, to allow the generation of multiple modes based on the various models that can be implemented in a pb_type
- move the package pad bels information in the architecture reader, rather than having them in the `arch` structure
- refactoring of some parts of the interchange reader, for instance by adopting the string internment to translate from an index to its string value.

#### Motivation and Context

Improvement in the Interchange support

Tests in Catch2 framework

#### Types of changes

#### Checklist:


tangxifan:
@acomodi @vaughnbetz 
To follow up our discussion on placing the IO files, should we try it on the ``libarchfpga`` directory?
Since we have two readers and writers, I am thinking if we can reorganize the ``src`` directory as follows.
There will be four subdirectories, ``base``, ``xml_io``, ``interchange_io`` and ``utils``
- The``base`` subdirectory contains data structures, i.e., ``physical_types.c|h``, ``arch_types.h`` etc.
- The ``xml_io`` subdirectory contains source files for XML reader and writer
- The ``interchange_io`` subdirectory contains source files for FPGA interchange reader and writer
- The ``utils`` contains any utility functions for data structure and IO reader/writers.

Let me know if this makes sense or not.

vaughnbetz:
That seems reasonable to me.

acomodi:
Superseded by https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1937. Closing

In regards of changing the directory structure to separate files in their different scope, I think it would be better to do so after the main interchange architecture reader is in place, so that we can then open a PR which only purpose will be to change the directory structures, without affecting the source file behavior.

Issue #1931, opened 2021-12-05
Update the vqm2blif parser in order to add support for Stratix 10 and Agliex vqm files


kimiatkh:
The vqm2blif parser grammar is defined based on the vqm files dumped for Stratix IV. The vqm files generated by the new version of Quartus Prime targeting Stratix 10 and Agilex have subtle differences compared to the older versions that causes them not to fit into the currently defined grammar. By making small modifications to the grammar, the parser can support both the old and new versions of the vqm file. 

#### Proposed Behaviour
The parser should be able to parse both versions of the vqm files. 

#### Current Behaviour
The vqm files dumped for Stratix 10 and Agliex are being marked as invalid by the parser.

#### Possible Solution
The flex and yacc files available in the libvqm directory define the grammar for the vqm file. By making small modifications to these two files the parser can support both versions of the vqm file. 

#### Context

The vqm2blif tool is a part of the Titan flow. In order to add support for Stratix 10 and Agilex in the Titan flow, the vqm files generated by Quartus targeting these two devices have to successfully go through the vqm2blif tool.

PR #1936, opened 2021-12-14
Update the vqm2blif parser to add support for Stratix 10 and Agliex vqm files

kimiatkh:
#### Description
This PR includes a few minor modifications to the vqm2blif parser.  Small changes have been made to the parser grammar files (vqm_parser.l and vqm_parser.y) in order to achieve the following objectives:

- The synthesis attributes and directives inside (* and *) delimiters should be ignored

- The names that start with \ (called TOKEN_ESCAPEID) should be allowed to be used for declaration of a vector. Currently the name of the vectors can only start with alphabetic characters and _ (TOKEN_REGULARID).

- If a line in the code ends with a TOKEN_ESCAPEID and there is no space between the token and the ; at the end of the line, ; should not be considered part of the TOKEN_ESCAPEID. 

- The ! operator is escaped. Though it should be noted that this results in the lut_mask parameter being incorrect after conversion from VQM to BLIF. 

- Bit strings in hexadecimal format (nh’[0-9|A-F]+) should be accepted by the parser. 

-  If the declarations and statements appear in a mixed order the parser should still be able to successfully parse the VQM file. 

#### Related Issue

This PR is related to the issue #1931 

#### Motivation and Context
The purpose of these modifications is to enable the tool to parse the Quartus VQM files targeting Stratix 10 and Agilex. The current version of the parser grammar is written based on the VQM files targeting Stratix IV devices. The new VQM files cannot successfully go through the parser as they differ slightly in formatting.

In order to be able to test the changes, the architecture description of Stratix 10 or Agilex is required. The tests will be included in a different pull request once the architecture capture of one of these two devices has reached a major milestone. 

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libvqm/vqm_parser.l
+++ b/libs/libvqm/vqm_parser.l
@@ -39,6 +39,7 @@
 %%
 ^[ \t]*\/\/[^\n\r]*(\n|\r\n)                /* skip one-line comments */
 [ \t]+                                      /* skip white spaces */
+!					    /* skip the logical operator ! applied on the input ports of the lut - this results in lut mask not being valid anoymore */
 (\n|\r\n)                                   /* skip empty lines */
 module                                      return TOKEN_MODULE;
 endmodule                                   return TOKEN_ENDMODULE;
--- a/libs/libvqm/vqm_parser.l
+++ b/libs/libvqm/vqm_parser.l
@@ -38,6 +38,7 @@
 
 %%
 ^[ \t]*\/\/[^\n\r]*(\n|\r\n)                /* skip one-line comments */
+^[ \t]*\(\*[^\n\r]*\*\)                     /* skip synthesis attributes and directives */
 [ \t]+                                      /* skip white spaces */
 !					    /* skip the logical operator ! applied on the input ports of the lut - this results in lut mask not being valid anoymore */
 (\n|\r\n)                                   /* skip empty lines */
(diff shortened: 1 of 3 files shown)


libs/libvqm/vqm_parser.l
```diff
@@ -38,7 +38,9 @@
 
 %%
 ^[ \t]*\/\/[^\n\r]*(\n|\r\n)                /* skip one-line comments */
+^[ \t]*\(\*[^\n\r]*\*\)                     /* skip synthesis attributes and directives */
 [ \t]+                                      /* skip white spaces */
+!					    /* skip the logical operator ! applied on the input ports of the lut - this results in lut mask not being valid anoymore */
```
vaughnbetz, line 43: probably want to explain this means inversion in the vqm netlist, but we are ignoring it (hence the LUTmask won't be correct).

utils/vqm2blif/src/base/vqm2blif_util.h
```diff
@@ -197,7 +197,9 @@ extern t_boolean buffd_outs;	//user-set flag that regulates whether to keep buff
 */
 const map<std::string, DeviceInfo> device_parameter_database {
     // stratix 4 device
```
vaughnbetz, line 199: Add a more descriptive comment here (device family, combinational cell name, output name, 1?, dff name, dff output name).

vaughnbetz (approved):
Looks good; let's merge and then you can add these comments for a future PR.

PR #1937, opened 2021-12-15
FPGA Interchange: improve arch reader

acomodi:
#### Description

This PR is to improve and add missing features to be able to correctly read the interchange device, and place and route a design using the automatically generated RR graph.

This is true for all the currently available interchange architectures, namely xilinx series7, lattice nexus and the basic test-architecture.

Given that the RR graph is auto-generated, the route file will be invalid, but we can start adding regression tests to verify that the whole flow is correct.

#### Motivation and Context

This is in the contect of adding support to the FPGA interchange format

Unit C++ tests

#### Types of changes

#### Checklist:


acomodi:
@tangxifan @vaughnbetz I opted to open a bigger PR to include all the changes necessary to have a LUT design to be placed and routed with the interchange format.

The diff is quite big, but all the major changes are only corresponding to the interchange arch reader, to which I have added some high level comments.

libs/libarchfpga/src/arch_util.h
```diff
@@ -100,4 +100,8 @@ bool pb_type_contains_blif_model(const t_pb_type* pb_type, const std::string& bl
 const t_pin_to_pin_annotation* find_sequential_annotation(const t_pb_type* pb_type, const t_model_ports* port, enum e_pin_to_pin_delay_annotations annot_type);
 const t_pin_to_pin_annotation* find_combinational_annotation(const t_pb_type* pb_type, std::string in_port, std::string out_port);
 
+void link_physical_logical_types(std::vector<t_physical_tile_type>& PhysicalTileTypes,
```
tangxifan, line 103: It would be great to add comments to these two APIs. I think they are created when adding the support on physical_tile/sub_tiles. I believe that their life cycle is long. Having comments will be easier for other developers.

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -1098,9 +1098,6 @@ static bool realloc_and_load_pb_graph_pin_ptrs_at_var(const int line_num,
         return false; //clb[9:0]123
     }
     (*token_index)++;
-    if (!checkTokenType(tokens[*token_index], TOKEN_STRING)) {
```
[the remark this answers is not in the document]
acomodi: Yeah, main motivation behind this is to be able to also allow numbers for pin names in pb types' ports. I think this should be allowed as if there is a multi-bit port, the bit is always enclosed in square brackets, avoiding ambiguity. I may add a `checkTokenType(..., TOKEN_DIGIT)` or similar check maybe, so that we exclude all other kinds of tokens.

libs/libarchfpga/src/arch_util.h
```diff
@@ -60,8 +60,8 @@ void free_type_descriptors(std::vector<t_physical_tile_type>& type_descriptors);
 
 t_port* findPortByName(const char* name, t_pb_type* pb_type, int* high_index, int* low_index);
 
-t_physical_tile_type SetupEmptyPhysicalType();
-t_logical_block_type SetupEmptyLogicalType();
+t_physical_tile_type get_empty_physical_type(std::string name);
```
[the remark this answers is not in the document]
acomodi: I have added a constant string and added it to be set as default when calling the `get_empty_...` function

vpr/src/pack/pb_type_graph.cpp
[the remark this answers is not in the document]
acomodi: I have changed this to allow only either INT or STRING tokens

libs/libvtrutil/src/vtr_util.h
```diff
@@ -118,6 +118,8 @@ void uniquify(Container container) {
 
 int get_pid();
 
+char* stringf(const char* format, ...);
```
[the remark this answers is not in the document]
acomodi: This API is actually not really required. The fact is that IMO it is better to make wider use of `std::string` and C++ standards to build strings. I have changed the interchange reader to avoid using this string formatter API

libs/libarchfpga/src/arch_util.h
acomodi: Added a high level comment about this and yes, this function is actually relevant to link the equivalent logical blocks that can be placed at a physical location.

vpr/src/base/SetupVPR.cpp
```diff
@@ -139,8 +139,9 @@ void SetupVPR(const t_options* Options,
     int num_inputs = 0;
     int num_outputs = 0;
     for (auto& type : device_ctx.physical_tile_types) {
-        if (strcmp(type.name, EMPTY_BLOCK_NAME) == 0) {
+        if (type.index == 0) {
```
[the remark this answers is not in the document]
acomodi: Added an `is_empty` attribute to the physical tile, similar to the other attributes such as `is_input_tile`. I think that in the future we might revisit all these data structures and make them more complete, with getters, modifiers, and such, so not to directly access what should be private members.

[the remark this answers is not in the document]
tangxifan: @acomodi Yes. I think we should revisit the data structures. I have proposed a plan last year. If time allowed, I will start refactoring these data structures in Q3-Q4 2022. Creating these APIs will help us to understand what client functions really need.

libs/libarchfpga/src/arch_util.h
tangxifan: Great. Thanks!

libs/libarchfpga/src/arch_util.h
```diff
@@ -15,6 +15,8 @@ void set_arch_file_name(const char* arch);
  */
 const char* get_arch_file_name();
 
+const std::string EMPTY_BLOCK_NAME("EMPTY");
```
tangxifan, line 18: Currently, most constant variable is defined by ``constexpr``. And most of time, we use ``char*`` than ``std::string``, to save memory.
This is a really preference from my side.
@hzeller Would you mind to share your opinions here?

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -2883,7 +2780,7 @@ static void ProcessTiles(pugi::xml_node Node,
     /* Alloc the type list. Need one additional t_type_desctiptors:
      * 1: empty psuedo-type
      */
-    t_physical_tile_type EMPTY_PHYSICAL_TILE_TYPE = SetupEmptyPhysicalType();
+    t_physical_tile_type EMPTY_PHYSICAL_TILE_TYPE = get_empty_physical_type(std::string("EMPTY"));
```
tangxifan, line 2783: Since you have a default input argument, you can simply write

```
get_empty_physical_type()
```

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3556,7 +3453,7 @@ static void ProcessComplexBlocks(vtr::string_internment* strings, pugi::xml_node
     /* Alloc the type list. Need one additional t_type_desctiptors:
      * 1: empty psuedo-type
      */
-    t_logical_block_type EMPTY_LOGICAL_BLOCK_TYPE = SetupEmptyLogicalType();
+    t_logical_block_type EMPTY_LOGICAL_BLOCK_TYPE = get_empty_logical_type(std::string("EMPTY"));
```
tangxifan, line 3456: Since you have a default input argument, you can simply write

```
get_empty_physical_type()
```

libs/libarchfpga/src/physical_types.h
```diff
@@ -653,6 +653,9 @@ struct t_physical_tile_type {
 
     // Does this t_physical_tile_type contain an outpad?
     bool is_output_type = false;
+
+    // Is this t_physical_tile_type an empty type?
+    bool is_empty = false;
```
tangxifan, line 658: When we will refactor the data structure, I think I will not keep a boolean variable. Currently, as far as I see, to identify an empty ``physical_tile``, there are multiple ways. I see in some codes, the name is used as a reference, while in other codes, the index is used as a reference. That will create issues. That is why we need an API ``is_empty()`` to provide a general reference. I do not have a strong opinion whether to use name or index. But we should use an API so that developers have a standard way to follow.

For example:

```
bool is_empty() {
  return std::string(name)== EMPTY_BLOCK_NAME;
} 
```

I am open to discussion. Let me know what you think.

tangxifan (changes requested):
@acomodi Thanks for the effort. Most comments have been well addressed. Just some minor details to cover.

libs/libarchfpga/src/physical_types.h
acomodi: Right, I have actually implemented this change in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/1937/commits/162ca071680bcd720af4421b29cffc5688644383. Also, I think that for now, the Interchange can also make use of the default name for the `EMPTY` blocks, but it is worth noticing that, if this can be a customizable parameter, relying on its name to verify whether a block is empty could not be the best solution.

I am thinking that maybe we can use the same approach as for the [architecture name](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/36789e456ab292038b8518bf2c835c9b9dcaaa0e/libs/libarchfpga/src/arch_util.cpp#L23-L36)?

libs/libarchfpga/src/read_xml_arch_file.cpp
acomodi: Done

libs/libarchfpga/src/read_xml_arch_file.cpp
acomodi: Done

libs/libarchfpga/src/arch_util.h
acomodi: No strong opinion on my side actually, I have set this to be `constexpr`

libs/libarchfpga/src/physical_types.h
tangxifan: @acomodi The approach for the architecture name makes sense to me. We follow this way.

tangxifan (approved):

acomodi:
All CI checks passed. Merging

PR #1932, opened 2021-12-09
RRGraphView nodes() Implementation

ethanroj23:
### Description

In this PR, I have implemented `RRGraphView::nodes()` which is a function that can be used to iterate over every node. Places throughout VTR that were doing this using `for (int id=0; id<nodes.size(); id++)` have been replaced with this function.

### Motivation and Context

These changes are a continuation of the RRGraphView API implementation effort.

### How Has This Been Tested?

I have run the regression tests for `vtr_reg_strong` along with the QoR testing found at `    $ ../scripts/run_vtr_task.py regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain`. Results from these QoR tests can be found below. The file containing all results can be found [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/7682594/nodes_output.xlsx).

Only rows that are different are shown.

|	| VTR Before These Changes	|With Changes in this PR|
|---|---|---|
|vtr_flow_elapsed_time	|1|	1.016294606|
|odin_synth_time	|1|	1.042278333|
|abc_synth_time	|1|	1.008543954|
|max_vpr_mem	|1|	1.00009757|
|pack_time	|1|	1.006138366|
|place_time	|1|	0.998640717|
|min_chan_width_route_time	|1|	1.004259057|
|crit_path_route_time	|1|	0.987037457|

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #1939, opened 2021-12-16
API reorder_rr_graph_node() created in rr_graph_util.cpp 

Raza-jafari-RS:
#### Description
In this PR, I have relocated the  reorder_rr_graph_node() from rr_graph_util.cpp to rr_graph_builder.cpp. 

#### Motivation and Context
As it was devised in the node_first_edge() Pull Request discussion.

#### Types of changes

#### Checklist:


Raza-jafari-RS:
I have replaced the function reorder_rr_graph_nodes() from rr_graph_util.cpp to rr_graph_builder.cpp as below.
https://github.com/RapidSilicon/vtr-verilog-to-routing/blob/17f164c5350aed7c460855e662f5be68bb67bf4f/vpr/src/device/rr_graph_builder.cpp#L57
And I made changes in respective files.
Can you please review them and suggest me how to move forward.

vpr/src/device/rr_graph_builder.h
```diff
@@ -59,6 +59,20 @@ class RRGraphBuilder {
     /** @brief Clear all the underlying data storage */
     void clear();
 
+    // Reorder RRNodeId's using one of these algorithms:
```
tangxifan, line 62: Please follow the comment style of other APIs, which are compatible with Doxygen.
As such, the API comments will be shown in online documentation.

tangxifan (changes requested):
@Raza-jafari-RS Let me know if you are fine with the comments. Feel free to ask questions.

Raza-jafari-RS:
@tangxifan
In my view all other issues has been resolved except removing the globals as the rr_node_metadata and the rr_edge metadata can not be moved outside the function the reason behind this is that they were using the internal data (v_num) in it.
please guide me about how to move them out of this function.
Moreover the API's node_first_edge and the node_last_edge are now useless as they were not called at any location. Should I remove them or keep them for future use?

tangxifan:
@Raza-jafari-RS 
```
size_t v_num = node_storage_.size();
```
This can be achieved by calling existing API
```
device_ctx.rr_nodes.size()
```
When @ethanroj23 's PR #1932 is in, we can use the new API from RRGraphView.

Raza-jafari-RS:
@tangxifan
In my view all other issues has been resolved except removing the globals as the rr_node_metadata and the rr_edge metadata can not be moved outside the function the reason behind this is that they were using the internal data (v_num) in it.
please guide me about how to move them out of this function.
Moreover the API's node_first_edge and the node_last_edge are now useless as they were not called at any location. Should I remove them or keep them for future use?

> @Raza-jafari-RS
> 
> ```
> size_t v_num = node_storage_.size();
> ```
> 
> This can be achieved by calling existing API
> 
> ```
> device_ctx.rr_nodes.size()
> ```
> 
> When @ethanroj23 's PR #1932 is in, we can use the new API from RRGraphView.

Okay I am doing using the device_ctx.rr_nodes.size() and once the PR #1932 is complete I will change it accordingly

Raza-jafari-RS:
@tangxifan,
The globals has been removed from the function but the rr_node_metadata and rr_edge_metadata are not moved out of this function as the basic c test fails after moving out.
please review the commit and guide me about how to move further.

tangxifan:
The changes make sense to me. Good work @Raza-jafari-RS 
Can you provide QoR tests?

umariqbal-rs:
@tangxifan 
QoR tests are complete and quick summary is given below. QoR comparison files are also attached.
**vtr_reg_qor_chain_depop test**
Flow run time -0.46% on average.
Peak memory usage 0.04% on average
**titan_quick_qor test:**
flow run time 2.68% on average.
Peak memory usage 0.02% on average

tangxifan:
@Raza-jafari-RS @umariqbal-rs Thanks for the updates. QoR tests look good. 

@vaughnbetz I propose to merge. However, I believe it would be great that you can take a quick review on this PR, before I merge. The main point is to see if you agree on
- If we should move the ``rr_node_metadata`` and ``rr_edge_metadata`` into the RRGraphBuilder
- It will not break any backward compatibility. If you know any authors of the ``rr_node_metadata`` and ``rr_edge_metadata`` or who will be impacted (@mithro Perhaps symbiflow will be impacted?), feel free to ask those guys for a review.

Also, I believe it would be great to add more comments for this API ``reorder_nodes``, if you know anyone can help.
We appreciate your help and support on this effort.

vpr/src/device/rr_graph_builder.h
```diff
@@ -58,6 +61,22 @@ class RRGraphBuilder {
 
     /** @brief Clear all the underlying data storage */
     void clear();
+    /** @brief reorder all the nodes
+     * Reorder RRNodeId's using one of these algorithms:
```
vaughnbetz, line 65: Here's some high-level context you can add to this comment.

Reordering the rr-graph nodes may be helpful in increasing cache locality during routing, and hence improving compile time. So far, however, the best re-ordering algorithm (DEGREE_BFS) does not speed up the router on most architectures vs. using the node ordering created by the rr-graph builder in VPR, so it is off by default.  The other use of this algorithm is for some unit tests; by changing the order of the nodes in the rr-graph before routing we check that no code depends on the rr-graph node order (as it shouldn't be).

tangxifan: Thanks @vaughnbetz. @Raza-jafari-RS Would you mind adding the comments to the API?

vpr/src/device/rr_graph_builder.h
```diff
@@ -208,6 +227,8 @@ class RRGraphBuilder {
     t_rr_graph_storage& node_storage_;
     /* Fast look-up for rr nodes */
     RRSpatialLookup node_lookup_;
+    MetadataStorage<int>& rr_node_metadata_;
```
vaughnbetz, line 230: Add comment:
Metadata is extra data on rr-nodes and edges, respectively, that is not used by vpr but simply passed through the flow so that it can be used by downstream tools. The main (perhaps only) current use of this metadata is the fasm tool of symbiflow, which needs extra metadata on which programming bits control which switch in order to produce a bitstream.

vpr/src/route/rr_graph_util.h
```diff
@@ -7,7 +7,7 @@ int seg_index_of_cblock(t_rr_type from_rr_type, int to_node);
 
 int seg_index_of_sblock(int from_node, int to_node);
 
-void reorder_rr_graph_nodes(const t_router_opts& router_opts);
+//void reorder_rr_graph_nodes(const t_router_opts& router_opts);
```
vaughnbetz, line 10: Delete this line instead of commenting it out.  It is dead code so better not to see it.

vaughnbetz:
Moving a reference to the metadata into the RRGraphBuilder makes sense to me, as you've done so far.  

@tangxifan: are you proposing to make RRGraphBuilder the owner of the underlying storage too (i.e. move the data, not just a reference, into the RRGraphBuilder)?  That might make sense but I haven't looked at the code in detail.  @acomodi  might know more.

@Raza-jafari-RS : I've added some proposed comments giving high-level context on rr-graph reordering (why would it be used?) and metadata.  Please incorporate those in the final PR.

tangxifan:
@vaughnbetz Thanks a lot on the comments. Happy new year by the way.

This PR has not done so actually. We are still using references in ``RRGraphBuilder``, which means ``DeviceContext`` is still the owner of the metadata. I raise this question because I believe it will impact our future PRs.

Regarding the final owner of the metadata, I think that ``RRGraphBuilder`` can be the owner because we agree that the reordering is an internal business of routing resource graph. We can build APIs to return metadata just as we did for the spatial look-up. 

I do not have a very strong opinion here. If @acomodi can provide more insights, we can discuss.

vaughnbetz:
Happy New Year to you too Xifan!
I have no objection to moving the metadata into the RRGraphBuilder.  It seems more logical than leaving it in the DeviceContext, but I haven't looked in detail.

Raza-jafari-RS:
@tangxifan,
I am adding the comments and I am curious about the two API's  node_first_edge() and node_last_edge() as now they are not in use so should I remove them or keep them for future use?
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/1333ba735f7f90a16797a037fba093389ef4847e/vpr/src/device/rr_graph_view.h#L128-L135

tangxifan:
@Raza-jafari-RS Happy new year.
Please do not touch the APIs in this PR. We will review the APIs  one by one once the refactoring is done.

vpr/src/route/rr_graph_util.h
Raza-jafari-RS: Removed

vpr/src/device/rr_graph_builder.h
Raza-jafari-RS: Comment added

vpr/src/device/rr_graph_builder.h
Raza-jafari-RS: comment added

Raza-jafari-RS:
@tangxifan 
Thanks for the wishes.
I have added the comments and also removed the unwanted lines. please review it

tangxifan:
@Raza-jafari-RS Thanks for the effort. You have applied @vaughnbetz 's comments well. I will wait for @mithro 's inputs before approve on the PR.

acomodi:
@vaughnbetz @tangxifan Sorry for the late response, anyhow, I have looked a bit through the code, as I wasn't too familiar with `rr metadata` code as well, and I think that moving the metadata info in the RR graph builder should not cause any troubles. There also is a C++ test in the catch2 framework that should verify that the RR metadata is properly working, and it passed in this PR.

For clarity though it might be good to move the following lines just below `rr_nodes`: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/base/vpr_context.h#L198-L215

vaughnbetz:
Thanks @acomodi and I agree with you on the comment; we'll want to keep that so it should move if the owner of this data moves.

tangxifan:
@vaughnbetz @acomodi 
Thanks for the inputs. In follow-up PRs, 
- We will move the comment along with the data structure under the ``RRGraphBuilder`` (the owner of the data).
- We will also keep an eye on the C++ unit test so that it will not break any backward compatibility.

Let me know if I misunderstand the message.

If we do not have any concerns, I suggest to merge this PR as is.

tangxifan:
As discussed today with @mithro , one thing we should keep in mind on the rr_node_metadata:

**It should be kept separated inside the RRGraphBuilder from other data structures**, i.e., ``node_storage``, ``node_lookup``,  ``RRSegment`` and ``RRSwitch``

We should also add code comments (warnings) to ``RRGraphBuilder``, so that developers will not merge the metadata into other data structures.

tangxifan (commented):
@Raza-jafari-RS Good work. We decide to merge this PR.
We need a follow-up pull request to move the data into RRGraphBuilder. Some rules have to respected. Please check the code comments.

tangxifan (approved):

PR #1940, opened 2021-12-16
added flags to ignore warnings when compiling unit tests that use the catch2 framework

Srivat97:
added flags to ignore warnings when compiling unit tests that use the catch2 framework

VTR currently uses the catch2 framework for unit tests. But compiling any project with the catch2 framework causes warnings to appear that are within the catch2 project. This PR removes the warnings due to the catch2 framework so they do not appear during compilation.

#### Description

The "-w" flag has been added to all unit test projects so that any warnings related to the catch2 framework are ignored.

#### Related Issue

This PR is related to issue 1935.

#### Motivation and Context

It is preferred to have a warning free build.

This has been tested by compiling VTR locally and verifying that all the warnings related to the catch2 framework do not appear.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/EXTERNAL/CMakeLists.txt
+++ b/libs/EXTERNAL/CMakeLists.txt
@@ -49,3 +49,12 @@ if(${VTR_ENABLE_CAPNPROTO})
       $<INSTALL_INTERFACE:include>
     )
 endif()
+
+# Some catch2 headers generate warnings, so treat them as system headers to suppress warnings
+target_include_directories(Catch2
+  SYSTEM
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/libcatch2/src>
+    $<INSTALL_INTERFACE:include>
+)
+


vaughnbetz:
It would be better if you turn off only the minimum warnings you need in order to not have spurious catch2 warnings that we're not going to fix.  That way we still get as much warning coverage as we can in the code we write to make unit tests.

Srivat97:
I found an easier solution to removing warnings when compiling the unit tests. Previously I was adding compiler tags to each unit test project to remove the warnings. Instead of the previous method, I decided to modify the libcatch2 build files and removed the warnings during its compilation.

Issue:
The problem with my change is that libcatch2 is currently included using git subtrees. So any changes I make to it should not be available to a new user who clones the repo for the first time. I think they would have to manually add the changes themselves.

Solutions:
1. We could change it so that we create a local copy of libcatch2, instead of using git subtrees,
2. Add somewhere in the documentation that a change needs to be made to libcatch2 so that warnings do not appear when building unit tests.

I am not sure which is the better option here.

vaughnbetz:
This is a very big PR; lots of changes that are already in master are listed here.  Did you rebase master onto your branch instead of the other way around?

Srivat97:
> This is a very big PR; lots of changes that are already in master are listed here. Did you rebase master onto your branch instead of the other way around?

Fixed now.

Srivat97:
@vaughnbetz : Found a fix that doesn't need us to modify the catch2 source code. All warnings for catch2 are now ignored and shouldn't clutter build output. This should be ready to be merged.

vaughnbetz (approved):
Looks good; thanks.

PR #1942, opened 2021-12-20
Updated comments to elaborate configurable/non-configurable edges()

behzadmehmood-rs:
#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/device/rr_graph_view.h
+++ b/vpr/src/device/rr_graph_view.h
@@ -259,12 +259,12 @@ class RRGraphView {
         return node_storage_.node_side_string(node);
     }
 
-    /** @brief Get the number of configurable edges. This function is inlined for runtime optimization. */
+    /** @brief Get the number of configurable edges, each configurable edge is controlled by a programmable memory bit. This function is inlined for runtime optimization. */
     inline t_edge_size num_configurable_edges(RRNodeId node) const {
         return node_storage_.num_configurable_edges(node);
     }
 
-    /** @brief Get the number of non-configurable edges. This function is inlined for runtime optimization. */
+    /** @brief Get the number of non-configurable edges, each non-configurable edge is a hard-wired connection. This function is inlined for runtime optimization. */
     inline t_edge_size num_non_configurable_edges(RRNodeId node) const {
         return node_storage_.num_non_configurable_edges(node);
     }
--- a/vpr/src/device/rr_graph_view.h
+++ b/vpr/src/device/rr_graph_view.h
@@ -259,22 +259,29 @@ class RRGraphView {
         return node_storage_.node_side_string(node);
     }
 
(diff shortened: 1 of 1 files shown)


vpr/src/device/rr_graph_view.h
```diff
@@ -259,12 +259,12 @@ class RRGraphView {
         return node_storage_.node_side_string(node);
     }
 
-    /** @brief Get the number of configurable edges. This function is inlined for runtime optimization. */
+    /** @brief Get the number of configurable edges, each configurable edge is controlled by a programmable memory bit. This function is inlined for runtime optimization. */
```
tangxifan, line 262: Can you enrich the code comments by following the reference:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/36cc78fe8ed5eb223e56899b6f67662b0aba2bf0/vpr/src/device/rr_graph_obj.h#L26-L31

We may enrich the comment like 
```
A configurable edge represents a programmable switch between routing resources, which could be 
- a multiplexer
- a tri-state buffer
- a pass gate

File: vpr/src/device/rr_graph_view.h [hunk 2]
```diff
@@ -259,12 +259,12 @@ class RRGraphView {
...
     inline t_edge_size num_configurable_edges(RRNodeId node) const {
         return node_storage_.num_configurable_edges(node);
     }
 
-    /** @brief Get the number of non-configurable edges. This function is inlined for runtime optimization. */
+    /** @brief Get the number of non-configurable edges, each non-configurable edge is a hard-wired connection. This function is inlined for runtime optimization. */
```
tangxifan, line 267: Also we can detailed the comment like
```
A non-configurable edge represents a hard-wired connection between routing resources, which could be 
-  a non-configurable (can not be turned off) buffer
-  a short metal connection that can not be turned off
```

tangxifan (changes requested):
@vaughnbetz I reviewed the code comments on configurable/non-configurable edges. I sugges the changes like this. Please let me know if it is detailed enough or not.

vpr/src/device/rr_graph_view.h
vaughnbetz: The current comment looks pretty good, but the enrichment you suggest would be good to add as well.

vpr/src/device/rr_graph_view.h
vaughnbetz: Agreed, adding the enrichment would be good.

behzadmehmood-rs: Sure.

tangxifan:
Thanks @vaughnbetz for the insights. 
@behzadmehmood-rs Once CI is green, I will propose to merge.

tangxifan (approved):

PR #1945, opened 2021-12-29
RRGraphView rr_switches() Implementation

umariqbal-rs:
#### Description
In this PR, I am implementing RRGraphView::rr_switch_inf(PRSwitchId switch_Id ) throughout VTR. Every time device_ctx.rr_switch_inf[switch_Id] was called has been replaced with rr_graph.rr_switch_inf(switch_id). In order to do this, I followed a pattern similar to that in the last PR.

#### Motivation and Context

#### Types of changes

#### Checklist:


umariqbal-rs:
@tangxifan 
I have created api of `rr_switch_inf` in `rr_graph_view.h`
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/e1acc7e7b6dfcbf82a291de929eb7d73e3ff961f/vpr/src/device/rr_graph_view.h#L355
I have also started passing strong Id.
Can you Please review it ,So I move forward.

vpr/src/device/rr_graph_view.cpp
```diff
@@ -2,9 +2,10 @@
 #include "rr_node.h"
 #include "physical_types.h"
 
-RRGraphView::RRGraphView(const t_rr_graph_storage& node_storage, const RRSpatialLookup& node_lookup, const vtr::vector<RRIndexedDataId, t_rr_indexed_data>& rr_indexed_data, const vtr::vector<RRSegmentId, t_segment_inf>& rr_segments)
+RRGraphView::RRGraphView(const t_rr_graph_storage& node_storage, const RRSpatialLookup& node_lookup, const vtr::vector<RRIndexedDataId, t_rr_indexed_data>& rr_indexed_data, const vtr::vector<RRSegmentId, t_segment_inf>& rr_segments,const vtr::vector<RRSwitchId, t_rr_switch_inf> rr_switch_inf)
```
tangxifan, line 5: You should use ``const vtr::vector<RRSwitchId, t_rr_switch_inf>&``

vpr/src/device/rr_graph_view.h
```diff
@@ -38,7 +38,8 @@ class RRGraphView {
     RRGraphView(const t_rr_graph_storage& node_storage,
                 const RRSpatialLookup& node_lookup,
                 const vtr::vector<RRIndexedDataId, t_rr_indexed_data>& rr_indexed_data,
-                const vtr::vector<RRSegmentId, t_segment_inf>& rr_segments);
+                const vtr::vector<RRSegmentId, t_segment_inf>& rr_segments,
+				const vtr::vector<RRSwitchId, t_rr_switch_inf> rr_switch_inf);
```
tangxifan, line 42: You should use ``const vtr::vector<RRSwitchId, t_rr_switch_inf>&``

vpr/src/device/rr_graph_view.h
```diff
@@ -369,6 +375,8 @@ class RRGraphView {
 
     /* Segment info for rr nodes */
     const vtr::vector<RRSegmentId, t_segment_inf>& rr_segments_;
+     /* switch info for rr nodes */
+    const vtr::vector<RRSwitchId, t_rr_switch_inf> rr_switch_inf_;
```
tangxifan, line 379: You should use ``const vtr::vector<RRSwitchId, t_rr_switch_inf>&``

tangxifan (changes requested):
Please correct all the errors in CI.  
You can have code format issues as well as compilation errors.
After that, we can look into the errors in regression tests.

Double check if you have the RRSwitchId defined as a data type.

umariqbal-rs:
@tangxifan 
I have changed `device_ctx.rr_switch_inf` with `rr_graph.rr_switch_inf(SwitchId)` in files where it needed.
But in `rr_graph.cpp` I am facing an issue.
From line [873 ](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/436a9f442eadf906697a836a79182e888b3d7da7/vpr/src/route/rr_graph.cpp#L873) to [879](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/436a9f442eadf906697a836a79182e888b3d7da7/vpr/src/route/rr_graph.cpp#L879)
Can you Please suggest me how can I go forward?

tangxifan:
@umariqbal-rs You should use RRSwitchId as the index:

```
device_ctx.rr_switch_inf[RRSwitchId(rr_switch_idx)].set_type(device_ctx.arch_switch_inf[arch_switch_idx].type());
```

If you want to modify the ``rr_switch_inf`` through ``rr_graph``, you need to add APIs to ``RRGraphBuilder`` rather than ``RRGraphView``. ``RRGraphView`` is designed to be a read-only object.

umariqbal-rs:
@tangxifan 
I didn't understand it Can you please explain a bit more for me?

umariqbal-rs:
So We need to create API in `rr_graph_builder` and remove it from rr_graph_view.h?

tangxifan:
> So We need to create API in `rr_graph_builder` and remove it from rr_graph_view.h?

No need to delete APIs from ``rr_graph_view.h``. You are doing in the right way.
For ``RRGraphBuilder``, it should provide mutators for rr_switch.
That is why I say ``If you want to modify the rr_switch_inf through rr_graph, you need to add APIs to RRGraphBuilder rather than RRGraphView.``

I have drawn the figure to illustrate the idea
image

Please let me know if you are still confused.

umariqbal-rs:
@tangxifan 
So We need to use RRSwitchId as the index in rr_graph.cpp now ,OR we should create API in rr_graph_builder as well?

tangxifan:
> @tangxifan So We need to use RRSwitchId as the index in rr_graph.cpp now ,OR we should create API in rr_graph_builder as well?

As a first step, I suggest you to use RRSwitchId as the index. Once you can compile with no errors, we think about the creation of APIs in ``RRGraphBuilder``. Let me know if it is clear.

umariqbal-rs:
@tangxifan 
Thanks I will go as you suggest, I will use now as index.After with no compile errors we can move forward.

umariqbal-rs:
@tangxifan 
That is completed with no compile errors.

tangxifan:
@umariqbal-rs Nightly1 test fails with the following error:

```
vtr_reg_nightly_test1/arithmetic_tasks/multless_consts...[Fail]
[Fail]
fixed_k6_frac_N8_22nm.xml/mult_057.v/common min_chan_width_route_time relative value 14.044585987261147 outside of range [0.1,14.0], above absolute threshold 2.0 and not equal to golden value: 1.57
```

Significant QoR shift is seen

vpr/src/device/rr_graph_view.h
```diff
@@ -351,6 +352,10 @@ class RRGraphView {
         return rr_segments_[seg_id];
     }
 
+    inline const t_rr_switch_inf& rr_switch_inf(RRSwitchId switch_id) const {
```
tangxifan, line 355: Please add comments to the API:

```
    /** @brief  Return the switch information that is categorized in the rr_switch_inf with a given id
 * rr_switch_inf is created to minimize memory footprint of RRGraph classs
 * While the RRG could contain millions (even much larger) of edges, there are only
 * a limited number of types of switches.
 * Hence, we use a flyweight pattern to store switch-related information that differs
 * only for types of switches (switch type, drive strength, R, C, etc.).
 * Each edge stores the ids of the switch that implements it so this additional information
 * can be easily looked up.
 *
 * @note All the switch-related information, such as R, C, should be placed in rr_switch_inf
 * but NOT directly in the edge-related data of RRGraph. 
 * If you wish to create a new data structure to represent switches between routing resources,
 * please follow the flyweight pattern by linking your switch ids to edges only!
     */ 
```

vpr/src/route/route_tree_timing.cpp
```diff
@@ -234,8 +234,8 @@ t_rt_node* update_route_tree(t_heap* hptr, int target_net_pin_index, SpatialRout
     if (subtree_parent_rt_node != nullptr) { /* Parent exists. */
         Tdel_start = subtree_parent_rt_node->Tdel;
         iswitch = unbuffered_subtree_rt_root->parent_switch;
-        Tdel_start += device_ctx.rr_switch_inf[iswitch].R * unbuffered_subtree_rt_root->C_downstream;
-        Tdel_start += device_ctx.rr_switch_inf[iswitch].Tdel;
+        Tdel_start += rr_graph.rr_switch_inf(RRSwitchId(iswitch)).R * unbuffered_subtree_rt_root->C_downstream;
```
tangxifan, line 237: Just a note (no action needed for this PR):

In future, we need to consider APIs that returns the Tdel for a routing trace in RRGraphView.

tangxifan (changes requested):

vpr/src/route/route_tree_timing.cpp
umariqbal-rs: Should I add this as comment?

umariqbal-rs:
Below are the QoR Results with **current master**:
**vtr_reg_qor_chain_depop test:**
Flow run time -1.98% on average.
Peak memory usage -0.01% on average
**titan_quick_qor test:**
flow run time 0.61% on average.
Peak memory usage 0% on average

umariqbal-rs:
Below are the QoR Results with **Sanity**:
**vtr_reg_qor_chain_depop test:**
Flow run time 0.36% on average.
Peak memory usage 0.01% on average
**titan_quick_qor test:**
flow run time 2.74% on average.
Peak memory usage 0.01% on average

tangxifan (approved):
@vaughnbetz I propose to merge this pull request.
- Code changes are clean. It adds APIs to ``RRGraphView`` which return ``rr_switch`` data structure.
- QoR tests are good.
- It does **not**  add APIs to ``RRGraphBuilder`` which will be owner of the ``rr_switch_inf``. We are going to do it in a follow-up PR (as discussed with @umariqbal-rs ) 

Feel free to comment.

tangxifan:
@vaughnbetz I merge this PR now as it is blocking other PRs on RRGraph refactoring. 
Your comments are always welcome. We will follow up.

PR #1958, opened 2022-01-27
FPGA Interchange: add constant synthetic tiles

acomodi:
#### Description

This adds the synthetic tiles for the constant network, improving support for the interchange devices.

The idea is to place a synthetic tile with two subtiles, each for one constant. The synthetic tile will be placed on the edge of the grid, so not to conflict with real tiles.

With the followup RR graph interchange reader, the constant tiles will be connected to the real RR graph based on the information present in the device

#### Motivation and Context

Constants need to be handled correctly, and fully supported

#### Types of changes

#### Checklist:


libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -2103,6 +2252,16 @@ struct ArchReader {
                 grid_def.loc_defs.emplace_back(std::move(single));
             }
 
+            // The constant source tile will be placed at (0, 0)
```
mkurc-ant, line 2255: How are we sure that there is no real tile at (0, 0) ?

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -811,14 +885,11 @@ struct ArchReader {
...
-        arch_->vcc_cell = str(consts.getVccCellType());
+        arch_->gnd_cell = std::make_pair(str(consts.getGndCellType()), str(consts.getGndCellPin()));
+        arch_->vcc_cell = std::make_pair(str(consts.getVccCellType()), str(consts.getVccCellPin()));
 
-        if (consts.getGndNetName().isName())
-            arch_->gnd_net = str(consts.getGndNetName().getName());
-
-        if (consts.getVccNetName().isName())
-            arch_->vcc_net = str(consts.getVccNetName().getName());
+        arch_->gnd_net = consts.getGndNetName().isName() ? str(consts.getGndNetName().getName()) : "$__gnd_net";
```
mkurc-ant, line 891: Are these hard-coded const net names special in any way? Maybe we could define them as constants somewhere else in the code?

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: There is an empty outer layer surrounding the grid, as all the tiles have been placed with a `(1, 1)` offset: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/libarchfpga/src/read_fpga_interchange_arch.cpp#L2095-L2096

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: So, in general the const names should be present in the interchange device. This is just a way to allow having them blank. At the moment there is no interchange device validator and there is no rule to determine whether the const `GND` and `VCC` net need to be defined.

We could opt for erroring out if those are not present

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -1968,74 +2034,157 @@ struct ArchReader {
...
-                    int true_physical_blk_pin = sub_tile.sub_tile_to_tile_pin_indices[iblk_pin++];
-                    fc_spec.pins.push_back(true_physical_blk_pin);
-                }
+        auto pb_type = new t_pb_type;
+        block.pb_type = pb_type;
 
-                type.fc_specs.push_back(fc_spec);
-            }
+        pb_type->name = vtr::strdup(const_block_.c_str());
+        pb_type->num_pb = 1;
```
tangxifan, line 2056: I see some attributes of the constant block is hard coded. Is it possible to provide more comments? For example, what are the two ports (VCC & GND)? And what are the naming convention for both pb_type and ports? I believe these information is useful when building a legalizer, e.g., check_arch(). As we support multiple file format, we definitely need a legalizer which tells us if the arch database is good enough for downstream functions.

tangxifan (commented):
Basically, the changes look good. Just need some comments so that it is easier for other developers.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
acomodi: I have added a comment on top of the function

acomodi:
I have added more comments on top of the new functions as well as some other. I may open a new PR to increase the comments coverage on the interchange-related functions as well.

For now I'll proceed with merging this PR and soon after I'll open the one about the RR graph generation.

Issue #1959, opened 2022-01-29
Not compiling on newer TBB versions


choelzl:
Does not compile on newer TBB versions (>2021.1) due to library file changes and feature depreation

#### Expected Behaviour
VTR compiles

#### Current Behaviour
VTR  will not compile (and cmake fails finding some files)

#### Possible Solution
Allow CMake to detect the old or new version and to add flags accordingly.

#### Steps to Reproduce
1. clone VTR
2. compile VTR

#### Context
Compilation of VTR

#### Your Environment
* VTR revision used: master (6aa1d568a82cab25fca831b0e4b08f8dec85aeb8)
* Operating System and version: Arch 5.16.3
* Compiler version: 11.1
* TBB version: 2021.5

PR #1961, opened 2022-01-30
Added Support for newer tbb lib versions

choelzl:
Intel updated their TBB library, changing the paths and deprecating some features. 
In order for VTR to continue working on more modern machines, some adaptation was required.

#### Description
Intel updated TBB. 
To support this, the required changes were:
- Fix the CMake files to detect either the `tbb_stddef.h` or `version.h` files of the library.
- Fix `vpr_api.c` due to `task_scheduler_init` being removed in newer versions
- Detect the version in CMake (in order to allow backward compatibility)

#### Related Issue
#1959

#### Motivation and Context
VTR will not compile on machines with TBB version greater than 2021.1

Compiled on a machine with newer TBB version

#### Types of changes

#### Checklist:

### DIFF
--- a/cmake/modules/FindTBB.cmake
+++ b/cmake/modules/FindTBB.cmake
@@ -184,7 +184,12 @@ if(NOT TBB_FOUND)
   ##################################
 
   if(TBB_INCLUDE_DIRS)
-    file(READ "${TBB_INCLUDE_DIRS}/tbb/tbb_stddef.h" _tbb_version_file)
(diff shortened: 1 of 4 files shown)


vaughnbetz:
@duck2 : can you check if we've already merged in an update to tbb, or if we should merge this?

choelzl:
Something similar has already been done in PR #2166 (dropping support for older versions however).

Issue #1756, opened 2021-06-02
Actually build the Dockerfile on CI


mithro:
Currently there is a Dockerfile @ https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/Dockerfile

I'm pretty sure this Dockerfile is currently broken due to trying to use a trusty base image which hasn't been supported for a very long time.

I suggest we start building and publishing the Docker images using the GitHub Container Registry (ghcr.io) & Google Container Registry (gcr.io).

Take a look at a look at https://github.com/google/skywater-pdk-actions/blob/main/.github/workflows/build-docker-image-run-drc-for-cell-gds-using-magic.yml and https://github.com/mithro/actions-includes/blob/main/.github/workflows/publish-docker-image.yml for examples on how to do this.

mithro:
FYI - @umarcor and @Xiretza

PR #1970, opened 2022-02-08
Fixed root Dockerfile for building vtr container

alirezazd:
Fixed broken Dockerfile script located at the project root.

#### Description

- The target image is now based on Ubuntu 20:04 LTS
- Build packages are now extracted from install_apt_packages.sh & requirements.txt files in VTR root
- Image build process is now in accordance with the latest build instructions available in VTR documenation.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1756

#### Motivation and Context

As an option for building VTR, there is a docker container build script. VTR's fixed Dockerfile allows it to be built on different OSs and architectures effortlessly, by just running one command. This can especially be useful for building VTR on devices with Apple Silicon processors which currently lack native Linux support but support Linux-based docker containers.

The fixed Dockerfile script now builds VTR sucessfully.

#### Types of changes

#### Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,35 +1,29 @@
 FROM ubuntu:20.04
+ARG DEBIAN_FRONTEND=noninteractive
 # set out workspace
 ENV WORKSPACE=/workspace
 RUN mkdir -p ${WORKSPACE}
-VOLUME ${WORKSPACE}
 WORKDIR ${WORKSPACE}
-# Set environment variables
-ARG GIT_SSL_NO_VERIFY=1
-ARG DEBIAN_FRONTEND=noninteractive
+COPY . ${WORKSPACE}
 # Install and cleanup is done in one command to minimize the build cache size
 RUN apt-get update -qq \
-    && apt-get -y install --no-install-recommends \
-# Additional packages not listed in install_apt_packages.sh
-    wget \
-    git \
-    ninja-build \
-    libeigen3-dev \
-    libtbb-dev \
-    python3-pip \
-# Clone VTR repo
-    && git clone https://github.com/verilog-to-routing/vtr-verilog-to-routing.git . \
 # Extract package names from install_apt_packages.sh
     && sed '/sudo/d' install_apt_packages.sh | sed '/#/d' | sed 's/ \\//g' | sed '/^$/d' | sed '/^[[:space:]]*$/d' \
 # Install packages
     | xargs apt-get -y install --no-install-recommends \
+# Additional packages not listed in install_apt_packages.sh
+    && apt-get -y install --no-install-recommends \
+    wget \
+    ninja-build \
+    libeigen3-dev \
+    libtbb-dev \
+    python3-pip \
+# Install python packages
+    && pip install -r requirements.txt \
 # Cleanup
     && apt-get autoclean && apt-get clean && apt-get -y autoremove \
     && rm -rf /var/lib/apt/lists/*
-# Install python packages
-RUN pip install -r requirements.txt
 # Build VTR
-RUN make \
-&& make install
+RUN make && make install
 # Container's default launch command
-CMD [ "/bin/bash" ]
+SHELL ["/bin/bash", "-c"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,88 +1,35 @@
-FROM ubuntu:trusty as builder
-
-RUN apt-get update 
-RUN apt-get install -y \
-    software-properties-common
-
-# add auto gpg key other lauchpad ppa
-RUN add-apt-repository ppa:nilarimogard/webupd8
-RUN apt-get update && apt-get install -y \
-    launchpad-getkeys
-
-# add llvm PPA
-RUN printf "\n\
-deb http://ppa.launchpad.net/george-edison55/precise-backports/ubuntu precise main \n\
-deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu trusty main \n\
-deb https://apt.llvm.org/precise llvm-toolchain-precise-3.6 main \n\
-deb https://apt.llvm.org/trusty llvm-toolchain-trusty-6.0 main \n\
-deb https://apt.llvm.org/trusty llvm-toolchain-trusty-7 main \n\
-deb https://apt.llvm.org/trusty llvm-toolchain-trusty-8 main \n\
-" >> /etc/apt/sources.list
-
-# grab llvm keys
-RUN launchpad-getkeys
-
-RUN apt-get update
-RUN apt-get install -y \
-    ninja \
-    libssl-dev \
-    autoconf \
-    automake \
-    bash \
-    bison \
-    binutils \
-    binutils-gold \
-    build-essential \
-    ctags \
-    curl \
-    doxygen \
-    flex \
-    fontconfig \
-    gdb \
-    git \
-    gperf \
-    libcairo2-dev \
-    libgtk-3-dev \
-    libevent-dev \
-    libfontconfig1-dev \
-    liblist-moreutils-perl \
-    libncurses5-dev \
-    libx11-dev \
-    libxft-dev \
-    libxml++2.6-dev \
-    perl \
-    python \
-    python-lxml \
-    texinfo \
-    time \
-    valgrind \
-    zip \
-    qt5-default \
-    clang-format-7 \
-    g++-7 \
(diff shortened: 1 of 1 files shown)


mithro:
PTAL - @umarcor

mithro (commented):
We should add CI which makes sure that this Docker file continues to work?

alirezazd:
> We should add CI which makes sure that this Docker file continues to work?

Yes, I'll be wroking on it and I'm also intending to Update the related documentaton..

Dockerfile
```diff
@@ -1,88 +1,35 @@
...
+FROM ubuntu:20.04
 # set out workspace
 ENV WORKSPACE=/workspace
 RUN mkdir -p ${WORKSPACE}
 VOLUME ${WORKSPACE}
 WORKDIR ${WORKSPACE}
-
+# Set environment variables
+ARG GIT_SSL_NO_VERIFY=1
+ARG DEBIAN_FRONTEND=noninteractive
```
umarcor, line 9: It is not desirable/necessary to set this as an argument. I'd recommend using it as a hardcoded argument to `apt install`. See e.g. https://github.com/hdl/containers/blob/main/debian-bullseye/vtr.dockerfile#L27.

Dockerfile
```diff
@@ -1,88 +1,35 @@
...
+    && apt-get -y install --no-install-recommends \
+# Additional packages not listed in install_apt_packages.sh
+    wget \
+    git \
+    ninja-build \
+    libeigen3-dev \
+    libtbb-dev \
+    python3-pip \
+# Clone VTR repo
+    && git clone https://github.com/verilog-to-routing/vtr-verilog-to-routing.git . \
```
umarcor, line 21: I'd suggest to clone the tool in a different stage (RUN) than installing the dependencies. That will help iterate faster when working with the dockerfile.

Dockerfile
```diff
@@ -1,88 +1,35 @@
...
+    && sed '/sudo/d' install_apt_packages.sh | sed '/#/d' | sed 's/ \\//g' | sed '/^$/d' | sed '/^[[:space:]]*$/d' \
+# Install packages
+    | xargs apt-get -y install --no-install-recommends \
+# Cleanup
+    && apt-get autoclean && apt-get clean && apt-get -y autoremove \
+    && rm -rf /var/lib/apt/lists/*
+# Install python packages
+RUN pip install -r requirements.txt
+# Build VTR
+RUN make \
```
umarcor, line 32: Ninja and cmake are installed, but they seems not to be used, since `make` is called. Are those used "internally"?

Dockerfile
```diff
@@ -1,88 +1,35 @@
...
+&& make install
+# Container's default launch command
 CMD [ "/bin/bash" ]
```
umarcor, line 35: In order to change the default shell to `bash`, I'd suggest using SHELL instead of CMD. See https://github.com/hdl/containers/blob/main/debian-bullseye/base.dockerfile#L26.

Dockerfile
umarcor, line 21: Since this dockerfile is located in the root of the repo, users/developers do need to clone the repo already. Therefore, cloning it again inside the Dockerfile is redudant. I recomment to use BuildKit's `mount` features instead. See https://hdl.github.io/containers/dev/Contributing.html#buildkit and https://github.com/hdl/containers/blob/main/debian-bullseye/boolector/Dockerfile#L27.

umarcor (commented):
Overall, it is not desirable to install the dependencies and build VTR in the same `docker build` execution. That is particularly relevant if the dockerfile is to be used locally, to work around environment (host OS) limitations.

Instead, I'd suggest to use named stages and split dependency installation from building the tool (VTR here). That will allow to build the image with dependencies in CI and make it available through a registry. Then, users will only need to pull the image and build VTR, without installing all the deps again and again.

You can still use a single Dockerfile, but you need to use named stages.

Nonetheless, this is not specific to this PR, because the previous Dockerfile was already written like that. Hence, it can be addressed later.

umarcor:
> We should add CI which makes sure that this Docker file continues to work?

@mithro building to ensure that the Dockerfile works is trivial:

```yml
  Image:
    runs-on: ubuntu-latest
    env:
      DOCKER_BUILDKIT: 1
    steps:
    - uses: actions/checkout@v2

    - name: Build container image
      run: docker build -t test/vtr -f Dockerfile .
```

My recommendation would be to use ghcr, because it works with the default token. So:

```yml
  Image:
    runs-on: ubuntu-latest
    env:
      DOCKER_BUILDKIT: 1
    steps:
    - uses: actions/checkout@v2

    - name: Build container image
      run: docker build -t ghcr.io/verilog-to-routing/build -f Dockerfile .

    - name: Push container image
      uses: pyTooling/Actions/with-post-step@r0
      with:
        main: |
          echo '${{ github.token }}' | docker login ghcr.io -u GitHub-Actions --password-stdin
          docker push ghcr.io/verilog-to-routing/build
        post: docker logout ghcr.io
```

On the other hand, there is https://github.com/hdl/containers/blob/main/debian-bullseye/vtr.dockerfile#L24-L43 already.
That is based on Debian Bullseye, instead of Ubuntu 20.04.
I could publish that to gcr.io/hdl-containers as `debian/bullseye/build/vtr`, similarly to *nextpnr* or *impl*: https://console.cloud.google.com/gcr/images/hdl-containers/global/debian/bullseye/build.
Then, users might decide to use the one from ghcr based on Ubuntu or the one from gcr based on Debian.

Dockerfile
alirezazd: Yes exactly, The root makefile is just a cmake wrapper and I can confirm that the ODIN's makefile uses ninja if it's installed.

Dockerfile
```diff
@@ -1,88 +1,35 @@
...
+    wget \
+    git \
+    ninja-build \
+    libeigen3-dev \
+    libtbb-dev \
+    python3-pip \
+# Clone VTR repo
+    && git clone https://github.com/verilog-to-routing/vtr-verilog-to-routing.git . \
+# Extract package names from install_apt_packages.sh
+    && sed '/sudo/d' install_apt_packages.sh | sed '/#/d' | sed 's/ \\//g' | sed '/^$/d' | sed '/^[[:space:]]*$/d' \
```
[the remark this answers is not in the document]
alirezazd: I think this should be adressed in another PR. But yes that would make more sense, instead of sudoing for everly line. Moreover It can help with the Dockerfile radability.

alirezazd:
@umarcor Thank you for your time and the detailed review. Most of your comments have been implemented. Currently, I'm reading BuildKit's docs in order to utilize the mount feature. I'll re-open the PR after completing.

> Overall, it is not desirable to install the dependencies and build VTR in the same `docker build` execution. That is particularly relevant if the dockerfile is to be used locally, to work around environment (host OS) limitations.
> 
> Instead, I'd suggest to use named stages and split dependency installation from building the tool (VTR here). That will allow to build the image with dependencies in CI and make it available through a registry. Then, users will only need to pull the image and build VTR, without installing all the deps again and again.
> 
> You can still use a single Dockerfile, but you need to use named stages.
> 
> Nonetheless, this is not specific to this PR, because the previous Dockerfile was already written like that. Hence, it can be addressed later.

Yes, I agree with you. I just fixed the Dockerfile according to its current structure, since I was trying to build VTR on MAC and did not have native Linux support. I'll rewrite the Dockerfile and implement the structure you mentioned in a future PR (if someone doesn't accomplish it sooner).

alirezazd:
@umarcor I applied your reviews as much as I could; however I could not apply following:

1. Could not remove ARG DEBIAN_FRONTEND=noninteractive since I could not hardcode it in xargs (check Line 13).
2. Tried to use build kit's mount feature  instead of cloning the repo, but could not build with cmake nor make because of not having write permissions (tired _-- prefix_ and _cmake -DCMAKE_INSTALL_PREFIX:PATH=/workspace_ to change the build path but did not work). I used COPY instruction as a workaround.

umarcor:
@alirezazd:

1. It's ok. It's not worth fighting with it now; better deal with it when the script is improved.
2. The mount features is read-only, indeed, so you need to first copy the content to e.g. `/tmp` and then work there. Anyway, that's also not critical. Using COPY in this PR is acceptable.

alirezazd:
@mithro This PR is ready to merge, JFYI.

Issue #1965, opened 2022-02-03
Kokoro tests not running reliably


sfkhalid:
Some kokoro tests are getting stuck in the pending results stage, or expecting results stage.

#### Expected Behaviour
All of the kokoro tests should run to completion on pull requests.

#### Current Behaviour
On pull requests, most kokoro tests run to completion while one or two get stuck in a pending stage. Different tests get stuck each time.

For example, on PR #1938, all kokoro tests were passing except for the VtR Yosys test. A kokoro force run label was added, which made all runs go to completion except for two different kokoro tests. A kokoro force run label was added again, after which all tests ran to completion except VtR Nightly tests - Part 2 (which had succeeded during the previous two attempts at getting all of the kokoro tests to run).

vaughnbetz:
@tangxifan is seeing the same behaviour. He often sees 6 / 9 kokoro tests complete, and has to force a re-run, and then a different subset will team.  It is not consistent in terms of what tests hang.

vaughnbetz:
@QuantamHD says kokoro is his domain -- thanks!

vaughnbetz:
#1963 may be related.

mithro:
Recently we have had some issues with the Google Cloud Platform project which contains the kokoro runners due to some automated log4j checking systems.

@acomodi is looking at trying to move everything to GitHub Action runners hosted on Google Cloud Platform which should be a lot more transparent about when failures.

mithro:
With https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1986 - Kokoro has been replaced by GitHub Action Runners and thus this bug is not longer relavent.

PR #1973, opened 2022-02-11
ci: transition to GH custom runners for nightly builds

acomodi:

#### Description

This PR moves the infrastructure from Kokoro to GH actions to run large tests on GH custom runners

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1965

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/scripts/hostsetup.sh
+++ b/.github/scripts/hostsetup.sh
@@ -39,6 +39,7 @@ apt install -y \
   gdb \
   git \
   gperf \
+  gzip \
   libcairo2-dev \
   libcapnp-dev \
   libgtk-3-dev \
--- a/.github/scripts/hostsetup.sh
+++ b/.github/scripts/hostsetup.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+set -e
+echo "========================================"
+echo "Host updating packages"
+echo "----------------------------------------"
+apt update
+echo "----------------------------------------"
+
+
+echo "========================================"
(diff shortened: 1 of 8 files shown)


kgugala:
there is more than 20000 artifacts to upload in some of the jobs (and it is taking ages to do that). We should check if we upload only the files we want. Packing artifacts into a single archive before uploading may also speed up the CI

vaughnbetz:
Thanks @kgugala and @acomodi !  
I'm not worried about the failure in a few of those multless constant runs.  They are small (min chan with route times of ~1s) and that metric is quite noisy.  It looks like the failures went away in the latest run, and if they come back we could widen the acceptable threshold for it for these designs (or all small designs).

tangxifan:
LGTM
I see both self-hosted runner and kokoro in the CI list. Are you going to deprecate kokoro in the CI soon?

acomodi:
@vaughnbetz All right, for now several runs were successful having 8 cores, with no great impact on the whole CI runtime, so we can probably keep it this way.

@tangxifan Sure, the next step is to disable kokoro (cc @mithro) and than remove the kokoro configuration files from the repo.

mithro:
@acomodi - Create a GitHub issue and assign it to me.

vaughnbetz:
Sounds good.  The 3 failures that came and went are almost certainly not due to memory footprint, as they are tiny designs.

PR #1975, opened 2022-02-13
ci/test: remove unused matrix field

umarcor:
#### Description

Since GCC 5 and 6 were removed from the CI workflow 'test', a field in job Compatibility is not used anymore.

#### Related Issue

- https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/eb6f4b9ea234f7f5a20cea24608362c8ece4886b

#### Motivation and Context

It's a cleanup. Instead of an empty ` BUILD_TYPE`, it's set to `release`.

It's a change in CI, so self tested.

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -331,14 +331,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - { name: 'GCC 7 (Ubuntu Bionic - 18.04)',    eval: 'CC=gcc-7 && CXX=g++-7',           build: ''               }
-        - { name: 'GCC 8 (Debian Buster)',            eval: 'CC=gcc-8 && CXX=g++-8',           build: ''               }
-        - { name: 'GCC 9 (Ubuntu Focal - 20.04)',     eval: 'CC=gcc-9 && CXX=g++-9',           build: ''               }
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -304,7 +304,7 @@ jobs:
       env:
         CMAKE_PARAMS: "-DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on"
         MATRIX_EVAL: ${{ matrix.eval }}
-        BUILD_TYPE: ${{ matrix.build }}
+        BUILD_TYPE: release
```
mithro, line 307: We most certainly want to be building both release and debug versions? I'm unsure how we lost that?

umarcor: In several jobs, debug build types is used. These are the compatibility tests, which default to `release` only: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/Makefile#L20. I'm not sure we want to test compatibility and debug...

umarcor:
@vaughnbetz @mithro can we have this merged?

vaughnbetz:
Thanks for the reminder!

PR #1976, opened 2022-02-13
doc: update copyright year

umarcor:
#### Description

The year shown at the footer of the docs is `2016`. This PR updates it to `2012-2022`. @vaughnbetz what should be the proper starting year?

#### Related Issue

*None*

#### Motivation and Context

The docs have been updated since 2016...

It's changing the content of a hardcoded string. There is no functionality change at all. It's tested in CI.

#### Types of changes

#### Checklist:

### DIFF
--- a/doc/src/conf.py
+++ b/doc/src/conf.py
@@ -92,7 +92,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "Verilog-to-Routing"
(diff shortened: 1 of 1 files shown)


vaughnbetz:
We can put a starting date of 2012.  There was documentation before that (for vpr, then vtr, ...) but 2012 is when the first vtr paper came out so let's align with that.

PR #1978, opened 2022-02-13
ci/test: run jobs on ubuntu-20.04, instead of ubuntu-18.04

umarcor:
#### Description

Since the base image of the Dockerfile is to be updated to ubuntu:focal (see #1970), this PR changes the environment used in the jobs of workflow 'test'.

#### Related Issue

- #1970

#### Motivation and Context

Ubuntu 18.04 is maintained until April 2023 (see https://endoflife.software/operating-systems/linux/ubuntu), however, Ubuntu 20.04 is also LTS. Hence, it is sensible to bump the environment, despite 18.04 not being EOL yet.

It's CI, so self tested.

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,7 +66,7 @@ jobs:
 
   Build:
     name: 'B: Building VtR'
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
 
     - uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


mithro (commented):
While we don't have to run everything on it, we should have at least **some** running jobs on the oldest OS version we support?

umarcor:
@mithro, I believe there is an specific job/matrix for that: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.github/workflows/test.yml#L278-L310. So, the purpose of "Compatibility" is to evaluate the build on multiple OS versions. On the other hand, the rest of the jobs are all executed on the same (ubuntu-18.04 now, ubuntu-20.04 proposed).

umarcor:
@vaughnbetz @mithro can we have this merged?

PR #1983, opened 2022-02-17
Optimize floorplan checks during place moves

sfkhalid:
#### Description
During place moves, if a cluster is constrained, we do an additional step of intersecting the cluster's move range limit with its floorplan constraint before we create a move.

#### Related Issue
Related to adding floorplan constraints to VPR - issue #932

#### Motivation and Context
When running with floorplan constraints, placement time was increasing due to an increased number of aborted moves. This increase was a result of moves being created that moved the block out of its floorplan constraints. By adjusting the range limit, we reduce the number of moves that are illegal with respect to floorplanning and so the place time decreases. Although the place time still increases relative to running with no constraints, the increase is minimized.

The reduction in place time on different titan circuits that were run with floorplan constraints can be seen here:
[titan_QoR_results_round2.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/8089351/titan_QoR_results_round2.xlsx)

### DIFF
--- a/vpr/src/pack/attraction_groups.cpp
+++ b/vpr/src/pack/attraction_groups.cpp
@@ -26,17 +26,7 @@ AttractionInfo::AttractionInfo(bool attraction_groups_on) {
         }
 
         //Then, fill in the group id for the atoms that do have an attraction group
-        int num_att_grps = attraction_groups.size();
-
-        for (int igroup = 0; igroup < num_att_grps; igroup++) {
-            AttractGroupId group_id(igroup);
-
-            AttractionGroup att_group = attraction_groups[group_id];
-
-            for (unsigned int iatom = 0; iatom < att_group.group_atoms.size(); iatom++) {
-                atom_attraction_group[att_group.group_atoms[iatom]] = group_id;
-            }
-        }
+        assign_atom_attraction_ids();
 
         att_group_pulls = 1;
     }
@@ -90,21 +80,11 @@ void AttractionInfo::create_att_groups_for_overfull_regions() {
     }
 
     //Then, fill in the group id for the atoms that do have an attraction group
-    int num_att_grps = attraction_groups.size();
-
-    for (int igroup = 0; igroup < num_att_grps; igroup++) {
-        AttractGroupId group_id(igroup);
-
-        AttractionGroup att_group = attraction_groups[group_id];
-
-        for (unsigned int iatom = 0; iatom < att_group.group_atoms.size(); iatom++) {
-            atom_attraction_group[att_group.group_atoms[iatom]] = group_id;
-        }
-    }
+    assign_atom_attraction_ids();
 
     att_group_pulls = 1;
 
-    VTR_LOG("%d clustering attraction groups created. \n", num_att_grps);
+    VTR_LOG("%d clustering attraction groups created. \n", attraction_groups.size());
 }
 
 void AttractionInfo::create_att_groups_for_all_regions() {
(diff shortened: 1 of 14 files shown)


vpr/src/pack/attraction_groups.cpp
```diff
@@ -107,6 +107,57 @@ void AttractionInfo::create_att_groups_for_overfull_regions() {
...
+        AttractionGroup group_info;
+        group_info.group_atoms = floorplanning_ctx.constraints.get_part_atoms(partid);
+
+        attraction_groups.push_back(group_info);
+    }
+
+    //Then, fill in the group id for the atoms that do have an attraction group
+    int num_att_grps = attraction_groups.size();
+
+    for (int igroup = 0; igroup < num_att_grps; igroup++) {
```
sfkhalid, line 146: create a helper function for this loop (146 - 154)

vpr/src/pack/pack.cpp
```diff
@@ -184,12 +184,22 @@ bool try_pack(t_packer_opts* packer_opts,
             attraction_groups.set_att_group_pulls(1);
 
         } else if (pack_iteration >= 2 && pack_iteration < 5 && floorplan_not_fitting) {
-            VTR_LOG("Floorplan regions are overfull: trying to pack again with more attraction groups exploration and higher target pin utilization. \n");
-            attraction_groups.create_att_groups_for_overfull_regions();
-            VTR_LOG("Pack iteration is %d\n", pack_iteration);
-            attraction_groups.set_att_group_pulls(4);
-            t_ext_pin_util pin_util(1.0, 1.0);
-            target_external_pin_util.set_block_pin_util("clb", pin_util);
+            if (pack_iteration == 2) {
```
sfkhalid, line 187: add comments about what our strategy is here: what do we do differently in each iteration

sfkhalid: also might just want to make a loop with a defined number of max packing iterations

vpr/src/place/move_utils.cpp
```diff
@@ -848,6 +874,43 @@ bool find_compatible_compressed_loc_in_range(t_logical_block_type_ptr type, int
...
+
+    auto& floorplanning_ctx = g_vpr_ctx.floorplanning();
+
+    PartitionRegion pr = floorplanning_ctx.cluster_constraints[b_from];
+    std::vector<Region> regions;
+    if (!pr.empty()) {
+        regions = pr.get_partition_region();
+    }
+    Region intersect_reg;
+    if (regions.size() == 1) {
```
sfkhalid, line 896: if region size is > 1, return true, explain why

vpr/src/place/move_utils.cpp
```diff
@@ -707,6 +726,13 @@ bool find_to_loc_centroid(t_logical_block_type_ptr blk_type,
     int cy_to = OPEN;
     bool legal = false;
 
+    if (is_cluster_constrained(b_from)) {
```
sfkhalid, line 729: might want to move the is_cluster_constrained inside the intersect routine -> run this way on one big circuit, and see if it makes any difference in cpu time

sfkhalid: moving this inside the intersect routine makes the place time go up, so i kept it as is

sfkhalid:
QoR comparison on this branch:
[place_moves_qor.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/8143054/place_moves_qor.xlsx)
All runtimes look the same as the master.

sfkhalid:
All checks passed and QoR has been checked, merging this branch.

PR #1992, opened 2022-03-07
Add missing include for <algorithm>.

hzeller:
There is a [change in llvm](https://github.com/verilog-to-routing/vtr-verilog-to-routing.git) that soon won't automatically
surface symbols from a transitive include of `<algorithm>`, so we need to make sure to IWYU.


mithro (approved):
LGTM! Merge if CI says yes.

PR #1997, opened 2022-03-14
Fixed wrong path for current directory

trick2011:
#### Description
Changed path of the basic timing test given in the verification of the installation. 

#### Motivation and Context
This is just a common sense fix. It is the single line that exists to test the installation and the current version requires debugging to understand instead of running the line and seeing it work.

Did it after my installation


AlexandreSinger:
@vaughnbetz I was just poking around old PRs and saw this one. This just changes one line in a documentation file, I think it is safe to merge.

PR #2006, opened 2022-04-12
doc: use symbiflow theme

umarcor:
#### Description

Change from the RTD theme to the SymbiFlow theme (based on Material theme).

#### Related Issue

https://github.com/chipsalliance/f4pga/issues/244

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,26 +1,9 @@
-#Python requirements file for building documentation
-# used by Read The Docs to install python required
-# modules with pip.
-
 sphinx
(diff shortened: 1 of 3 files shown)


vaughnbetz:
Is there a preview of what the documentation would look like that you can point me at?  
I personally prefer the current menu structure in the vtr docs where clicking on the left bar expands sub-menu items, vs. what appears to be the symbiflow style of sub-topics showing up in a right hand menu when you click on an item.  That seems to cut screen real estate, and produces two menus in different parts of the screen.

PR #2009, opened 2022-04-12
doc: copy changelog to avoid warnings

umarcor:
Sphinx produces multiple warnings because https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/doc/src/CHANGELOG.md does not have a title. In this PR, the file is copied to `doc/src` in the `conf.py`, instead of having a "redirection" file.


mithro (commented):
Ideally this should be handled by the `MarkdownSymlinksDomain` / `markdown_code_symlinks` plugin, but I think @acomodi was planning on retiring that?

@acomodi -- I think we are looking to you for the best solution here?

vaughnbetz:
I don't think I have an informed opinion about whether or not this is the right fix.  @acomodi : should we merge this, or would you recommend something else?

acomodi:
I see that [CHANGELOG.md](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/doc/src/CHANGELOG.md) in `doc/src` is actually a symlink to the actual CHANGELOG.md in the root dir, while for the other `.md` files (e.g. `CONTRIBUTING.md`), there is no symlink but rather it defines a reference to the actual CONTRIBUTING.md file:

````
```{include} ../../CONTRIBUTING.md
```
````

We may try using the same approach for the CHANGELOG.md file and see if this solves the issue.

PR #2015, opened 2022-04-15
SystemVerilog file support added to yosys; Added support to provide t…

karanmathur:
…op module from command line

#### Description

Yosys supports SystemVerilog files by providing -sv to its read_verilog command. This change allows for running SystemVerilog files. Also added support for explicitly declaring the top module in the command line for the yosys flow @sdamghan. 

#### Motivation and Context

Yosys supports .sv files and this change enables that support. 

Ran vtr_reg_basic and vtr_reg_strong they passed, however they do not use yosys. Ran vtr_reg_yosys and vtr_reg_yosys_odin, all tests passed for this. 

#### Types of changes

#### Checklist:


aman26kbm:
@karanmathur , looks like there are some failures that need to be looked at.

There are 3 that are formatting errors. You can click on the failed task in the list above and see the reason for the failure. The formatting related failure messages will tell you how to fix them.

There are 2 other test failures also. These are failures from ODIN's regression tests. We only ran the relevant VTR regression tests. From what I see, parts of these tasks have failed:

vtr-verilog-to-routing/ODIN_II/regression_test/benchmark/task/yosys+odin/koios
vtr-verilog-to-routing/ODIN_II/regression_test/benchmark/task/yosys+odin/keywords

Can you please rerun them locally and see the failures? You may want to only run the ones that failed to save time. For example, for the koios task, robot_rl benchmark failed, so you can edit the task.conf file to only have robot_rl in it and run.

You can read about ODIN regression tests here:
https://docs.verilogtorouting.org/en/latest/odin/dev_guide/regression_test/
I think to run a task, we need to run `make test` in the directory of the test, but we can check with @sdamghan if this doesn't work.

ODIN_II/regression_test/tools/run_yosys.sh
```diff
@@ -362,7 +362,7 @@ function run_single_hdl() {
     for circuit in "${_VERILOG_INPUT_LIST[@]}"
     do
         # validate input file
-        if [ "${circuit: -2}" != ".v" ]; then
+        if ["${circuit: -2}" != ".sv" and "${circuit: -2}" != ".v" ]; then
```
sdamghan, line 365: this should be `"${circuit: -3}" != ".sv"`

sdamghan: there is also a script in `$VTR_ROOT/ODIN_II/regression_test/tools/synth.tcl` that needs to be modified

aman26kbm:
I looked at all the failures.

There are two types of issues:

1. In the koios robot_rl design, the systemverilog keyword "bit" is used as an identifier. Now that we enable systemverilog by passing "-sv", the "bit" identifier becomes an error (keywords can't be used as identifiers). For this one, I've fixed the benchmark code.

2. In other small tests which check for things signed, unsigned keywords, multiple declarations, etc., there are failures which are expected. But when systemverilog support is enabled by passing "-sv", those designs pass. Example:

Here is the code for one of the benchmarks:	
```
	  1 module simple_op(a,b);
	  2 input signed [3:0] a;
	  3 output unsigned [5:0] b;
	  4
	  5 assign b = a;
	  6
	  7 endmodule
```
	
Without -sv (behavior before this PR), we get this error in yosys log:
`54 regression_test/benchmark/verilog/keywords/signed_unsigned/signed_to_unsigned.v:3: ERROR: syntax error, unexpected TOK_ID, expecting ',' or '=' or ';'     or '['`
And hence this error in odin log:
`Yosys failed to perform elaboration, Please look at the log file for the failure cause or pass '--show_yosys_log' to Odin-II to see the logs. `

But with this PR's changes, these errors are not there in the logs. So, we just need to update the golden results I think.

@sdamghan , can you opine about these?

sdamghan:
LGTM, thanks for the update.

sdamghan (approved):

aman26kbm:
Merging this PR, since all checks have passed.

PR #2026, opened 2022-05-06
(Bug-Fix) No default search type

SebastianLievano:
#### Description
1) Added a missing line of code to the NO_PICTURE_TO_ROUTING version of the initial setup function to set a default search option
2) Added a check to ensure a type was selected, and create a warning dialog window if one was not selected. 

#### Motivation and Context
When opening vpr graphics, no initial search type is selected. Any attempts to search with no search type selected will crash the program. This fix aims to rectify that.

Searches with all different search settings were attempted before and after the change, and produced identical results. It no longer crashes the program, and it now opens with a default search type selected.

#### Types of changes

- [ X] Bug fix (change which fixes an issue)

#### Checklist:


vaughnbetz (approved):
Looks good .  One suggestion though: add a comment saying why this is needed.  Can do that when you fix this or in a separate PR; no need for me to re-review.

SebastianLievano:
Merging this version. Will add comment in next PR

Issue #2008, opened 2022-04-12
Seemingly incorrect optimization of multipliers in ODIN


aman26kbm:
#### Expected Behaviour
There should be >4096 multipliers in the design.

#### Current Behaviour
I'm running a design with the ODIN+Yosys flow. I was expecting atleast 4096 multipliers in the design. In the yosys log, I do see this at the end in the statistics section:

 $mul                         4160

In the odin log also, I see this:
Total # of multipliers = 4160

But then towards the end of the odin log, it shows:
Number of <MULTIPLY> node:                272

And then the VPR log also shows 272 multipliers.

Somehow the number of multipliers seems to have changed from 4160 to 272. And it seems to have happened in the ODIN stage. 

#### Steps to Reproduce

Arch file: k6FracN10LB_mem20K_complexDSP_customSB_22nm.xml
Design file: [all.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/8475453/all.zip)
Additional args: --const_gen_inference comb --timing_report_detail aggregated --route_chan_width 300 -elaborator yosys -fflegalize

#### Context

This design is one of the new Koios benchmarks.

#### Your Environment

* VTR revision used: master
* Operating System and version: ubuntu 18.04
* Compiler version:

sdamghan:
@aman26kbm - I have checked the design, the number of multipliers is 4160 in the Yosys output and after partial mapping. However, it is reduced to 272 after removing the unused logic phase in Odin-II. This phase actually traverses the netlist from the output nodes and marks every node seeing in a path from an output node to an input node. So technically, nodes that are not in any path that ends in an output node are considered unused logic. I would recommend double-checking the design and making sure all mulptiplication results are connected to at least one output of the top module.

aman26kbm:
I see. Thanks for checking, @sdamghan .
Let me try to find if something isn't connected correctly.
This benchmark is from a tool called DNNWeaver from UCSD. The github page mentions that they ran it through Xilinx Vivado.

sdamghan:
Oh, didn't know that. So we should dive more into this since this optimization is actually an old part of Odin-II. We may need to check from where the Yosys coarse-grained netlist is generated until after the unused logic removal.

[...]

PR #2027, opened 2022-05-09
Odin check netlist discontinuity

alirezazd:
Added a function that traverses the netlist forwards and backwards to find nodes of a specific type and generates a warning in case the results of the traverses do not match.

#### Description

I checked the ODIN optimization, techmap, and cleanup routines and found no problems in the code that could cause the nodes to be incorrectly removed. I added a function to the check_netlist() routine which is called after the elaboration and before the optimization phase. Just like the remove_unused_logic() function, it traverses the netlist forwards and backwards and searches for a specific node type (currently multiply nodes). If the number of the nodes found in the traverses doesn't match, it will generate a warning to indicate discontinuity within the netlist. I'm planning to extend this feature to check the netlist for main nodes types e.g. logical, addition. The feature can notify a user when there is a discontinuity in an input design, which will result in the main nodes being removed in the cleanup stage of Odin.

#### Related Issue

Isuue #2008 [Seemingly incorrect optimization of multipliers in ODIN](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2008#)
#### Types of changes

#### Checklist:

### DIFF
--- a/ODIN_II/SRC/include/odin_types.h
+++ b/ODIN_II/SRC/include/odin_types.h
@@ -67,6 +67,7 @@
 #define GRAPH_CRUNCH 17
 #define STATS 18
 #define SEQUENTIAL_LEVELIZE 19
+#define NETLIST_DISCONTINUITY_TRAVERSE_VALUE 11
 /* a unique mark for the BLIF elaboration DFS traversal */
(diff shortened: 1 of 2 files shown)


alirezazd:
@aman26kbm @sdamghan

PR #2029, opened 2022-05-10
Pb graph bug fixes

amin1377:
1- .gitignore file is changed to accommodate the files generated by Clion IDE.

2- Other than the pb_types defined in xml file, another level of pb_type is added to the structure if the pb_type in the XML file uses a [class](https://docs.verilogtorouting.org/en/latest/arch/reference/#classes).
Normal pb_types are initialized in "ProcessPb_Type" function in "libs/libarchfpga/src/read_xml_arch_file.cpp". However, the newly added pb_types are initialized in "alloc_and_load_default_child_for_pb_type" in "libs/libarchfpga/src/arch_util.cpp". Since they are initialized in two different functions, some pb_type members are missed in the second function.

3- Three members of t_physical_tile_type are not initialized properly. Each tile consists of, potentially, several types of sub_tiles. Each sub_tile has a capacity that is equal to the number of that sub_tile in the tile. In the current implementation, only the number of pins of one instance of a sub_tile is added to the corresponding member in tile data structure. There are two possible solutions for this problem: 1- change the name of num_*_pins to num_inst_*_pins 2- multiply the number of pins by the capacity of the sub_tile (In this PR, this solution is implemented)

@vaughnbetz

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,9 @@ tags
 .history
 #eclipse project
 .project
+
+#
+# CLION
+#
+.idea
+cmake-build-debug
--- a/libs/libarchfpga/src/arch_util.cpp
+++ b/libs/libarchfpga/src/arch_util.cpp
@@ -611,6 +611,7 @@ void alloc_and_load_default_child_for_pb_type(t_pb_type* pb_type,
     copy->num_clock_pins = pb_type->num_clock_pins;
     copy->num_input_pins = pb_type->num_input_pins;
     copy->num_output_pins = pb_type->num_output_pins;
+    copy->num_pins = pb_type->num_pins;
     copy->num_pb = 1;
 
     /* Power */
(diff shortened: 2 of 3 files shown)


vaughnbetz:
Lots of failures getting the required dependencies: 
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gdb/gdb_10.2-0ubuntu1~18.04~2_amd64.deb  Cannot initiate the connection to ppa.launchpad.net:80 (2001:67c:1560:8008::19). - connect (101: Network is unreachable) [IP: 91.189.95.85 80]

Can you relaunch @amin1377

amin1377:
I re-launched the CI tests, and now it passes all tests. [Link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/2302103820)

PR #2032, opened 2022-05-13
Added Auto-Complete Functionality for Block Names (VPR)

SebastianLievano:
Added an auto-complete function for Block Names in VPR. Search bar still only searches on clicking search button, so selection of an option from auto-complete menu only fills it into the search bar. Should still be quite helpful considering how long some of the names get. 

#### Description
Uses GtkListStore to store objects and GtkEntryCompletion to display/auto-complete the different entries. Added two helper functions:
**search_type_changed**: Manages auto-completion models based on search type. Currently sets model to null unless selected type is block name
**load_block_names**: Stores block names in listStore that is set as completion model when appropriate

Also added a few comments on last pull request, where a default search option was set. 

This was tested on some of the larger circuits (mcml.v and mes_noc (titan) ). It did not have a noticeable impact on load times, and was still responsive. If you find that this is slowing vpr down, please contact me. 

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/main.ui
+++ b/vpr/main.ui
@@ -1,13 +1,37 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.16.1 -->
+<!-- Generated with glade 3.22.2 -->
 <interface>
   <requires lib="gtk+" version="3.0"/>
+  <object class="GtkListStore" id="BlockNames">
+    <columns>
+      <!-- column-name gchararray1 -->
+      <column type="gchararray"/>
+    </columns>
+    <data>
+      <row>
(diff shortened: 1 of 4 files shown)


vaughnbetz (approved):
Looks good.  Ideas for further improvements embedded.
1. Make it work for nets.
2. Check if we can select primitives (LUTs, FFs, etc.) by name.  They are also blocks.  Test with the internals of blocks shown and without.
   - Ideally would highlight the primitive block if you select it by name
   - If block internals not shown, would still be convenient if it highlighted a cluster block if you typed the name of a primitive inside it.

PR #2033, opened 2022-05-13
Switching All Mallocs to News in VPR

jmah76:
#### Description
Most mallocs have been switched to news in the vpr folder. Changing some mallocs caused segfaults, happens when allocating memory to some objects of non-primitive types (ex. node_power->in_dens). Next step will be trying to find a fix & documenting what's wrong.


vpr/src/base/echo_files.cpp
```diff
@@ -62,8 +63,8 @@ char* getEchoFileName(enum e_echo_files echo_option) {
 }
 
 void alloc_and_load_echo_file_info() {
-    echoFileEnabled = (bool*)vtr::calloc((int)E_ECHO_END_TOKEN, sizeof(bool));
-    echoFileNames = (char**)vtr::calloc((int)E_ECHO_END_TOKEN, sizeof(char*));
+    echoFileEnabled = new bool[(int)E_ECHO_END_TOKEN]();
```
vaughnbetz, line 66: I think we should explicitly initialize these variables to false & NULL.  Looking up info on when C++ initializes built-in types like this to 0 when you use new gives complex answers: e.g. https://stackoverflow.com/questions/15212261/default-initialization-of-pod-types-in-c
says sometimes they are initialized to 0, sometimes not, depending on the exact syntax of the new call.  Not sure if that's correct or not, but since we rely on these being initialized to false/NULL we should make that explicit.

vpr/src/base/echo_files.cpp
```diff
@@ -62,8 +63,8 @@ char* getEchoFileName(enum e_echo_files echo_option) {
...
+    echoFileNames = new char*[(int)E_ECHO_END_TOKEN]();
```
vaughnbetz, line 68: Again, explicitly initialize to NULL.

vpr/src/base/echo_files.cpp
```diff
@@ -158,9 +160,9 @@ void alloc_and_load_output_file_names(const std::string default_name) {
     char* name;
 
     if (outputFileNames == nullptr) {
-        outputFileNames = (char**)vtr::calloc((int)E_FILE_END_TOKEN, sizeof(char*));
+        outputFileNames = new char*[(int)E_FILE_END_TOKEN]();
```
vaughnbetz, line 163: Again, best to replace calloc with explicit 0 initialization (or find documentation in the standard saying this is guaranteed to set to 0, and comment that and the fact that we rely on it).

vpr/src/base/echo_files.cpp
```diff
@@ -158,9 +160,9 @@ void alloc_and_load_output_file_names(const std::string default_name) {
...
 
-        name = (char*)vtr::malloc((strlen(default_name.c_str()) + 40) * sizeof(char));
+        name = new char[(strlen(default_name.c_str()) + 40)];
```
vaughnbetz, line 165: This "40" magic number makes me nervous.
- Best: can we change this to C++ strings (no size issues anymore)?  If not too much work, that would be good.
- If not, change the 40 to a meaningful defined constant (have to read the code to see what it should be) and make it considerably bigger -- say 1024 -- so we never have an issue with it.

vpr/src/base/stats.cpp
```diff
@@ -342,7 +342,7 @@ void print_wirelen_prob_dist() {
     int prob_dist_size, i, incr;
 
     prob_dist_size = device_ctx.grid.width() + device_ctx.grid.height() + 10;
-    prob_dist = (float*)vtr::calloc(prob_dist_size, sizeof(float));
+    prob_dist = new float[prob_dist_size]();
```
vaughnbetz, line 345: Again, does the standard guarantee 0 initialization of all elements?  If not, write a loop to set to 0.  Even if it does, comment it.

vpr/src/base/stats.cpp
```diff
@@ -364,7 +364,7 @@ void print_wirelen_prob_dist() {
                 VTR_LOG("Realloc'ing to increase 2-pin wirelen prob distribution array.\n");
                 incr = index - prob_dist_size + 2;
                 prob_dist_size += incr;
-                prob_dist = (float*)vtr::realloc(prob_dist, prob_dist_size * sizeof(float));
+                prob_dist = new float[prob_dist_size];
```
vaughnbetz, line 367: This looks like a memory leak.  realloc is a combination of free + malloc; I think you need a new, copy, delete here.

vpr/src/base/stats.cpp
```diff
@@ -377,8 +377,7 @@ void print_wirelen_prob_dist() {
                 VTR_LOG("Realloc'ing to increase 2-pin wirelen prob distribution array.\n");
                 incr = index - prob_dist_size + 2;
                 prob_dist_size += incr;
-                prob_dist = (float*)vtr::realloc(prob_dist,
-                                                 prob_dist_size * sizeof(float));
+                prob_dist = new float[prob_dist_size];
```
vaughnbetz, line 380: Same comment as above.  If realloc is used repeatedly, you could make a utility function to handle the new, copy, delete sequence.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -380,7 +380,7 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
             int type_len = strlen(pb_type->name);
             int name_len = strlen(pb->name);
             int tot_len = type_len + name_len;
-            char* blk_tag = (char*)vtr::malloc((tot_len + 8) * sizeof(char));
+            char* blk_tag = new char[(tot_len + 8)];
```
vaughnbetz, line 383: I don't like the "8" magic number.  Same comment as above: make a C++ string, or make a defined constant (and be sure 8 is always enough, or increase it).

vpr/src/place/place.cpp
```diff
@@ -2590,15 +2592,13 @@ static void alloc_and_load_for_fast_cost_update(float place_cost_exp) {
      * subhigh must be greater than or equal to sublow, we only need to       *
      * allocate storage for the lower half of a matrix.                       */
 
-    chanx_place_cost_fac = (float**)vtr::malloc(
-        (device_ctx.grid.height()) * sizeof(float*));
+    chanx_place_cost_fac = new float*[(device_ctx.grid.height())];
```
vaughnbetz, line 2595: Might be able to change these to a vtr_matrix (I think you can -- they are a 2D array, so a vtr_matrix should be a good replacement). Can leave for a future PR if you wish, but good to file an issue on and schedule a clean up.

vaughnbetz: See the vtr utilities documentation on the vtr website for information on vtr_matrix.

vpr/src/power/power.cpp
```diff
@@ -147,8 +147,8 @@ static void power_usage_primitive(t_power_usage* power_usage, t_pb* pb, t_pb_gra
 
         LUT_size = pb_graph_node->num_input_pins[0];
 
-        input_probabilities = (float*)vtr::calloc(LUT_size, sizeof(float));
-        input_densities = (float*)vtr::calloc(LUT_size, sizeof(float));
+        input_probabilities = new float[LUT_size]();
```
vaughnbetz, line 150: Again, these are calloc --> explicitly set to 0 in a loop, or verify the standard guarantees 0 initialization and comment that.

vpr/src/power/power_callibrate.cpp
```diff
@@ -150,8 +150,7 @@ void power_print_spice_comparison() {
     fprintf(power_ctx.output->out, "Energy of LUT (High Activity)\n");
     for (i = 0; i < (sizeof(LUT_sizes) / sizeof(int)); i++) {
         for (j = 1; j <= LUT_sizes[i]; j++) {
-            SRAM_bits = (char*)vtr::realloc(SRAM_bits,
-                                            ((1 << j) + 1) * sizeof(char));
+            SRAM_bits = new char[((1 << j) + 1)];
```
vaughnbetz, line 153: realloc -> new will create a memory leak (see comments above).

vpr/src/power/power_callibrate.cpp
```diff
@@ -163,8 +162,8 @@ void power_print_spice_comparison() {
             SRAM_bits[1 << j] = '\0';
         }
 
-        dens = (float*)vtr::realloc(dens, LUT_sizes[i] * sizeof(float));
-        prob = (float*)vtr::realloc(prob, LUT_sizes[i] * sizeof(float));
+        dens = new float[LUT_sizes[i]];
```
vaughnbetz, line 165: realloc --> needs more handling than this.

vpr/src/power/power_components.cpp
```diff
@@ -48,7 +48,7 @@ void power_components_init() {
     int i;
     auto& power_ctx = g_vpr_ctx.mutable_power();
 
-    power_ctx.by_component.components = (t_power_usage*)vtr::calloc(POWER_COMPONENT_MAX_NUM, sizeof(t_power_usage));
+    power_ctx.by_component.components = new t_power_usage[POWER_COMPONENT_MAX_NUM]();
```
vaughnbetz, line 51: calloc --> 0 initialize.

vpr/src/power/power_components.cpp
```diff
@@ -222,15 +222,13 @@ void power_usage_lut(t_power_usage* power_usage, int lut_size, float transistor_
     num_SRAM_bits = 1 << lut_size;
 
     /* Initialize internal node data */
-    internal_prob = (float**)vtr::calloc(lut_size + 1, sizeof(float*));
-    internal_dens = (float**)vtr::calloc(lut_size + 1, sizeof(float*));
-    internal_v = (float**)vtr::calloc(lut_size + 1, sizeof(float*));
+    internal_prob = new float*[lut_size + 1]();
```
vaughnbetz, line 225: calloc --> 0 initialize

vpr/src/power/power_components.cpp
```diff
@@ -436,8 +434,8 @@ void power_usage_local_interc_mux(t_power_usage* power_usage, t_pb* pb, t_interc
             /* Many-to-1, or Many-to-Many
              * Implemented as a multiplexer for each output
              * */
-            in_dens = (float*)vtr::calloc(interc->interconnect_power->num_input_ports, sizeof(float));
-            in_prob = (float*)vtr::calloc(interc->interconnect_power->num_input_ports, sizeof(float));
+            in_dens = new float[interc->interconnect_power->num_input_ports]();
```
vaughnbetz, line 437: calloc --> 0 initialize

vpr/src/power/power_components.cpp
```diff
@@ -532,7 +530,7 @@ void power_usage_mux_multilevel(t_power_usage* power_usage,
     bool found;
     PowerSpicedComponent* callibration;
     float scale_factor;
-    int* selector_values = (int*)vtr::calloc(mux_arch->levels, sizeof(int));
+    int* selector_values = new int[mux_arch->levels]();
```
vaughnbetz, line 533: calloc --> 0 initialized by new?

vpr/src/power/power_components.cpp
```diff
@@ -578,7 +576,7 @@ static void power_usage_mux_rec(t_power_usage* power_usage, float* out_prob, flo
         return;
     }
 
-    v_in = (float*)vtr::calloc(mux_node->num_inputs, sizeof(float));
+    v_in = new float[mux_node->num_inputs]();
```
vaughnbetz, line 579: calloc --> 0 initialize

vpr/src/power/power_components.cpp
```diff
@@ -589,8 +587,8 @@ static void power_usage_mux_rec(t_power_usage* power_usage, float* out_prob, flo
         }
     } else {
         /* Higher level of mux - inputs recursive from lower levels */
-        in_prob = (float*)vtr::calloc(mux_node->num_inputs, sizeof(float));
-        in_dens = (float*)vtr::calloc(mux_node->num_inputs, sizeof(float));
+        in_prob = new float[mux_node->num_inputs]();
```
vaughnbetz, line 590: calloc --> 0 initialize

vpr/src/power/power_sizing.cpp
```diff
@@ -140,7 +140,7 @@ static double power_count_transistors_mux(t_mux_arch* mux_arch) {
     float* max_inputs;
 
     /* SRAM bits */
-    max_inputs = (float*)vtr::calloc(mux_arch->levels, sizeof(float));
+    max_inputs = new float[mux_arch->levels]();
```
vaughnbetz, line 143: calloc --> 0 initialize

vpr/src/power/power_sizing.cpp
```diff
@@ -745,10 +745,8 @@ static void power_size_pin_buffers_and_wires(t_pb_graph_pin* pin,
         int* fanout_per_mode;
         float* wirelength_out_per_mode;
 
-        fanout_per_mode = (int*)vtr::calloc(this_pb_type->num_modes,
-                                            sizeof(int));
-        wirelength_out_per_mode = (float*)vtr::calloc(this_pb_type->num_modes,
-                                                      sizeof(float));
+        fanout_per_mode = new int[this_pb_type->num_modes]();
```
vaughnbetz, line 748: calloc --> 0 initialize

vpr/src/route/rr_graph.cpp
```diff
@@ -2392,7 +2392,7 @@ static void check_all_tracks_reach_pins(t_logical_block_type_ptr type,
     VTR_ASSERT(max_chan_width > 0);
 
     int* num_conns_to_track; /* [0..max_chan_width-1] */
-    num_conns_to_track = (int*)vtr::calloc(max_chan_width, sizeof(int));
+    num_conns_to_track = new int[max_chan_width]();
```
vaughnbetz, line 2395: calloc --> 0 initialize

vpr/src/route/rr_graph.cpp
```diff
@@ -2641,14 +2641,14 @@ static t_clb_to_clb_directs* alloc_and_load_clb_to_clb_directs(const t_direct_in
 
     auto& device_ctx = g_vpr_ctx.device();
 
-    clb_to_clb_directs = (t_clb_to_clb_directs*)vtr::calloc(num_directs, sizeof(t_clb_to_clb_directs));
+    clb_to_clb_directs = new t_clb_to_clb_directs[num_directs]();
```
vaughnbetz, line 2644: calloc --> 0 initialize

vpr/src/route/rr_graph_area.cpp
```diff
@@ -145,11 +145,11 @@ void count_bidir_routing_transistors(int num_switch, int wire_to_ipin_switch, fl
         trans_track_to_cblock_buf = 0;
     }
 
-    num_inputs_to_cblock = (int*)vtr::calloc(rr_graph.num_nodes(), sizeof(int));
+    num_inputs_to_cblock = new int[rr_graph.num_nodes()]();
```
vaughnbetz, line 148: calloc --> 0 initialize

vpr/src/route/rr_graph_area.cpp
```diff
@@ -145,11 +145,11 @@ void count_bidir_routing_transistors(int num_switch, int wire_to_ipin_switch, fl
...
 
     maxlen = std::max(device_ctx.grid.width(), device_ctx.grid.height());
-    cblock_counted = (bool*)vtr::calloc(maxlen, sizeof(bool));
-    shared_buffer_trans = (float*)vtr::calloc(maxlen, sizeof(float));
+    cblock_counted = new bool[maxlen]();
+    shared_buffer_trans = new float[maxlen]();
```
vaughnbetz, line 152: calloc --> 0 initialize

vpr/src/route/rr_graph_area.cpp
```diff
@@ -319,7 +319,7 @@ void count_unidir_routing_transistors(std::vector<t_segment_inf>& /*segment_inf*
      * switches of all rr nodes. Thus we keep track of which muxes we have already
      * counted via the variable below. */
     bool* chan_node_switch_done;
-    chan_node_switch_done = (bool*)vtr::calloc(rr_graph.num_nodes(), sizeof(bool));
+    chan_node_switch_done = new bool[rr_graph.num_nodes()]();
```
vaughnbetz, line 322: calloc --> 0 initialize

vpr/src/route/rr_graph_area.cpp
```diff
@@ -349,9 +349,9 @@ void count_unidir_routing_transistors(std::vector<t_segment_inf>& /*segment_inf*
         trans_track_to_cblock_buf = 0;
     }
 
-    num_inputs_to_cblock = (int*)vtr::calloc(rr_graph.num_nodes(), sizeof(int));
+    num_inputs_to_cblock = new int[rr_graph.num_nodes()]();
```
vaughnbetz, line 352: calloc --> 0 initialize

vpr/src/route/rr_graph_timing_params.cpp
```diff
@@ -42,8 +42,8 @@ void add_rr_graph_C_from_switches(float C_ipin_cblock) {
     auto& mutable_device_ctx = g_vpr_ctx.mutable_device();
 
     maxlen = std::max(device_ctx.grid.width(), device_ctx.grid.height());
-    cblock_counted = (bool*)vtr::calloc(maxlen, sizeof(bool));
-    buffer_Cin = (float*)vtr::calloc(maxlen, sizeof(float));
+    cblock_counted = new bool[maxlen]();
```
vaughnbetz, line 45: calloc --> 0 initialize

vpr/src/route/rr_graph_timing_params.cpp
```diff
@@ -172,7 +172,7 @@ void add_rr_graph_C_from_switches(float C_ipin_cblock) {
      * Current structures only keep switch information from a node to the next node and
      * not the reverse.  Therefore I need to go through all the possible edges to figure
      * out what the Cout's should be */
-    Couts_to_add = (float*)vtr::calloc(rr_graph.num_nodes(), sizeof(float));
+    Couts_to_add = new float[rr_graph.num_nodes()]();
```
vaughnbetz, line 175: calloc --> 0 initialize

vpr/src/route/segment_stats.cpp
```diff
@@ -41,13 +41,11 @@ void get_segment_usage_stats(std::vector<t_segment_inf>& segment_inf) {
         max_segment_name_length = std::max(max_segment_name_length, static_cast<int>(segment_inf[seg_type].name.size()));
     }
 
-    seg_occ_by_length = (int*)vtr::calloc((max_segment_length + 1),
-                                          sizeof(int));
-    seg_cap_by_length = (int*)vtr::calloc((max_segment_length + 1),
-                                          sizeof(int));
+    seg_occ_by_length = new int[max_segment_length + 1]();
```
vaughnbetz, line 44: calloc --> 0 initialize

vpr/src/route/segment_stats.cpp
```diff
@@ -41,13 +41,11 @@ void get_segment_usage_stats(std::vector<t_segment_inf>& segment_inf) {
...
+    seg_cap_by_length = new int[max_segment_length + 1]();
 
-    seg_occ_by_type = (int*)vtr::calloc(segment_inf.size(), sizeof(int));
-    seg_cap_by_type = (int*)vtr::calloc(segment_inf.size(), sizeof(int));
+    seg_occ_by_type = new int[segment_inf.size()]();
```
vaughnbetz, line 47: calloc --> 0 initialize

vaughnbetz (changes requested):
1. I believe the conversion of realloc -> new is not correct (see comments).
2. I'm not sure about the calloc -> new conversion without explicit initialization to 0.  May be guaranteed by the standard, but need to look that up and be sure (new initialization with a plain-old data type / built-in data type).
3. There are a few funky char* sequences of code with magic numbers in them. This is a good time to get rid of those fixed size buffers if we can (C++ strings), or at least use defined constants like MAX_EXTRA_LEN or some such and verify they are always big enough (comment why they're big enough, or make bigger and there's only one or a few strings).

vaughnbetz:
This link is helpful in showing a few succinct ways to make new initialize an array to 0:  https://stackoverflow.com/questions/2204176/how-to-initialise-memory-with-new-operator-in-c
It looks like it is not guaranteed to be initialized to 0 with the usual syntax (the one you are using), which matches what I thought.

vpr/src/base/echo_files.cpp
```diff
@@ -63,8 +63,12 @@ char* getEchoFileName(enum e_echo_files echo_option) {
 }
 
 void alloc_and_load_echo_file_info() {
-    echoFileEnabled = new bool[(int)E_ECHO_END_TOKEN]();
-    echoFileNames = new char*[(int)E_ECHO_END_TOKEN]();
+    echoFileEnabled = new bool[(int)E_ECHO_END_TOKEN];
+    echoFileNames = new char*[(int)E_ECHO_END_TOKEN];
+    for (auto i = 0; i < (int)E_ECHO_END_TOKEN; i++) {
+        echoFileEnabled[i] = 0;
```
vaughnbetz, line 69: Better to set the bool to false (cleaner code, although functionally the same).

vpr/src/base/stats.cpp
```diff
@@ -364,9 +366,15 @@ void print_wirelen_prob_dist() {
                 VTR_LOG("Realloc'ing to increase 2-pin wirelen prob distribution array.\n");
                 incr = index - prob_dist_size + 2;
                 prob_dist_size += incr;
+
+                float temp[prob_dist_size];
```
vaughnbetz, line 370: Even better change: we could just make this a std::vector and use vector operations.

vpr/src/base/stats.cpp
vaughnbetz: Or just make it a vector (probably best of all).

vpr/src/power/power_callibrate.cpp
```diff
@@ -150,7 +150,13 @@ void power_print_spice_comparison() {
     fprintf(power_ctx.output->out, "Energy of LUT (High Activity)\n");
     for (i = 0; i < (sizeof(LUT_sizes) / sizeof(int)); i++) {
         for (j = 1; j <= LUT_sizes[i]; j++) {
+            char temp[((1 << j) + 1)];
```
vaughnbetz, line 153: Could use a string to make this shorter/cleaner.

vpr/src/route/rr_graph.cpp
```diff
@@ -2392,7 +2392,9 @@ static void check_all_tracks_reach_pins(t_logical_block_type_ptr type,
     VTR_ASSERT(max_chan_width > 0);
 
     int* num_conns_to_track; /* [0..max_chan_width-1] */
```
vaughnbetz, line 2394: Could recode this as a vector, and just use the 
num_conns_to_track(max_chan_width, 0);  constructor.

vpr/src/route/rr_graph_area.cpp
```diff
@@ -319,7 +325,9 @@ void count_unidir_routing_transistors(std::vector<t_segment_inf>& /*segment_inf*
      * switches of all rr nodes. Thus we keep track of which muxes we have already
      * counted via the variable below. */
     bool* chan_node_switch_done;
-    chan_node_switch_done = new bool[rr_graph.num_nodes()]();
+    chan_node_switch_done = new bool[rr_graph.num_nodes()];
+    for (auto i = 0; i < rr_graph.num_nodes(); i++)
+        chan_node_switch_done[i] = 0;
```
vaughnbetz, line 330: Better to set to false (more clear).
Could recode as vector and use its constructor to set to 0; a bit nicer code as it should get rid of an explicit new and delete too if this is a local variable.

vaughnbetz (commented):
Thanks; the changes look correct.

One tweak: set bools to false instead of 0 (easier to read).
Also, some of these (maybe most) could be changed to strings or vectors and the realloc code would go away and the calloc code could be replaced by a constructor (and no new or delete if they're local variables). That would be a good code clean up to do.

vpr/src/base/stats.cpp
```diff
@@ -367,14 +367,15 @@ void print_wirelen_prob_dist() {
                 incr = index - prob_dist_size + 2;
                 prob_dist_size += incr;
 
-                float temp[prob_dist_size];
+                float* temp = new float[prob_dist_size];
```
vaughnbetz, line 370: We should just make this a vector I think.  The code is basically doing vector.resize() operations to gradually grow an array.  If we just switch the prob_dist to a vector we can get rid of some code.

jmah76: Is there any reason why we do `if (index >= prob_dist_size) { [contents here]}` twice, or should I just leave it?

vpr/src/power/power_components.cpp
vaughnbetz, line 51: Worth commenting that this tricky bit of code initializes all the power usage data (built in types anyway) to 0, or (probably even better) just make a constructor in t_power_usage that does that.  Then the new will call the default constructor and it will be less mysterious.

Issue #2031, opened 2022-05-10
Crash when running yosys+odin flow


aman26kbm:
I'm running yosys+odin flow on a design. Yosys abruptly ends (at the AUTONAME pass). And an error is logged in odin.out saying blif file not found.

#### Expected Behaviour

Yosys+odin should finish normally.

#### Current Behaviour

I've attached a zip file that contains the design file. This is an HLS generated design.

I ran two experiments to dig into what's going on.

First, I enabled debug build. With that I see memory leak in the odin log. The odin.out is in the attached zip file. Also included is the yosys elaboration log.

Second, I tried to see which module in the design is causing the crash.
I ran the flow with all the second level modules under the top module.
All modules passed, except conv_2d_latency_cl_0_0_0_0_0_0_0_0_0_0_0_0_0
Then I ran the flow with all the third level modules under this module.
They all passed as well.
So, the issue is from the logic in this module itself.

I see a giant state machine in this module (1000s of states). May be that's causing the issue.

[bug_may10.zip](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/8665689/bug_may10.zip)

#### Steps to Reproduce

I'm using this architecture: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/COFFE_22nm/k6FracN10LB_mem20K_complexDSP_customSB_22nm.xml 

For debug, I suggest running just this module. You can use a command like this:
../../scripts/run_vtr_flow.py lenet_may10.v ../../arch/COFFE_22nm/k6FracN10LB_mem20K_complexDSP_customSB_22nm.xml --const_gen_inference comb --timing_report_detail aggregated --route_chan_width 300 -top_module conv_2d_latency_cl_0_0_0_0_0_0_0_0_0_0_0_0_0 -temp_dir lenet_may10 -elaborator yosys -fflegalize

#### Your Environment

* VTR revision used: master
* Operating System and version:
* Compiler version:

sdamghan:
[...]

PR #2035, opened 2022-05-13
Performing the Yosys "autoname" pass after netlist optimizations

sdamghan:

#### Description

move the `autoname` pass in the Yosys+Odin-II synthesis steps to avoid giant RAM usage by Yosys for big benchmarks, like the one mentioned in #2031 

#### Related Issue

#2031 

#### Types of changes

#### Checklist:


vaughnbetz:
Some failures.  I looked at the vtr_reg_yosys_odin, and it has two QoR failures, both of which look like they could be due to increased memory block counts.  So maybe they are expected changes and the golden QoR values should be updated?

=======
```

21:28:54 | regression_tests/vtr_reg_yosys_odin/vtr_benchmarks...[Fail]
21:28:54 | [Fail]
21:28:54 | k6_frac_N10_frac_chain_mem32K_40nm.xml/arm_core.v/common device_limiting_resources Task value 'memory' does not match golden 'clb'
21:28:54 | 
21:28:54 | regression_tests/vtr_reg_yosys_odin/vtr_benchmarks...[Fail]
21:28:54 | [Fail]
21:28:54 | k6_frac_N10_frac_chain_mem32K_40nm.xml/sha.v/common device_grid_tiles relative value 1.1377777777777778 outside of range [0.9,1.1] and not equal to golden value: 225.0
```

ODIN_II/SRC/YYosys.cpp
```diff
@@ -259,14 +259,14 @@ void YYosys::execute() {
...
 
-        // Use a readable name convention
-        run_pass(std::string("autoname"));
         // Translate processes to netlist components such as MUXs, FFs and latches
         run_pass(std::string("proc; opt;"));
         // Extraction and optimization of finite state machines
         run_pass(std::string("fsm; opt;"));
         // Collects memories, their port and create multiport memory cells
         run_pass(std::string("memory_collect; memory_dff; opt;"));
+        // Use a readable name convention
```
vaughnbetz, line 268: It seems like it would be good to comment that this step can have a high memory footprint (can reference the issue #), and hence it is best to run it after the optimizations above as the smaller netlist helps.  
 --> Hopefully my understanding is correct, if not please correct in the comment :)

sdamghan: I will certainly add related comments, and yes, your statement is entirely correct.

sdamghan:
@vaughnbetz - thanks for the review. I have been actually inspecting the results. Moving the `autoname` pass, which is only a renaming process (to the best of my knowledge), to execute a few commands later shouldn't result in any change in the number of netlist blocks. Interestingly, it seems Yosys output BLIF has changed. I want to make sure nothing is broken then I will regenerate golden results.

PR #2046, opened 2022-05-30
analytical placement's bugs are fixed

saaramahmoudi:
The analytical placement approach used to fail in some large titan benchmarks due to some logical bugs. The bugs have been resolved in this branch.
Function update_macros did not check whether macro can be placed with the current head position. Hence, in some cases, blk_locs for some macro members remained -1 until the end of the flow and produced an error at the end. 
Some assertions were removed since they were not logically correct, and some extra conditions were added in Cut_Spreader.cpp file to avoid producing errors in large designs. 
@vaughnbetz

### DIFF
--- a/vpr/src/place/analytic_placer.cpp
+++ b/vpr/src/place/analytic_placer.cpp
@@ -410,28 +410,19 @@ void AnalyticPlacer::setup_solve_blks(t_logical_block_type_ptr blkTypes) {
  * when formulating the matrix equations), an update for members is necessary
  */
 void AnalyticPlacer::update_macros() {
-    size_t max_x = g_vpr_ctx.device().grid.width();
-    size_t max_y = g_vpr_ctx.device().grid.height();
     for (auto& macro : g_vpr_ctx.mutable_placement().pl_macros) {
         ClusterBlockId head_id = macro.members[0].blk_index;
-        bool mac_can_be_placed = true;
-        //check all members' locations are within the chip, otherwise macro can not be placed with current head position
-        for (auto member = ++macro.members.begin(); member != macro.members.end(); ++member) {
-            auto member_loc = blk_locs[head_id].loc + member->offset;
-            if (member_loc.x >= int(max_x) || member_loc.x < 0) { //member location is not on the chip
-                mac_can_be_placed = false;
-                break;
-            }
-            if (member_loc.y >= int(max_y) || member_loc.y < 0) { //member location is not on the chip
-                mac_can_be_placed = false;
-                break;
-            }
-        }
+        bool mac_can_be_placed = macro_can_be_placed(macro, blk_locs[head_id].loc, true);
+
         //if macro can not be placed in this head pos, change the head pos
         if (!mac_can_be_placed) {
             size_t macro_size = macro.members.size();
             blk_locs[head_id].loc -= macro.members[macro_size - 1].offset;
         }
+
+        //macro should be placed successfully after changing the head position
+        VTR_ASSERT(macro_can_be_placed(macro, blk_locs[head_id].loc, true));
+
         //update other member's location based on head pos
         for (auto member = ++macro.members.begin(); member != macro.members.end(); ++member) {
             blk_locs[member->blk_index].loc = blk_locs[head_id].loc + member->offset;
--- a/vpr/src/place/analytic_placer.cpp
+++ b/vpr/src/place/analytic_placer.cpp
@@ -410,8 +410,29 @@ void AnalyticPlacer::setup_solve_blks(t_logical_block_type_ptr blkTypes) {
  * when formulating the matrix equations), an update for members is necessary
  */
 void AnalyticPlacer::update_macros() {
+    size_t max_x = g_vpr_ctx.device().grid.width();
+    size_t max_y = g_vpr_ctx.device().grid.height();
     for (auto& macro : g_vpr_ctx.mutable_placement().pl_macros) {
         ClusterBlockId head_id = macro.members[0].blk_index;
+        bool mac_can_be_placed = true;
+        //check all members' locations are within the chip, otherwise macro can not be placed with current head position
+        for (auto member = ++macro.members.begin(); member != macro.members.end(); ++member) {
+            auto member_loc = blk_locs[head_id].loc + member->offset;
+            if (member_loc.x >= int(max_x) || member_loc.x < 0) { //member location is not on the chip
+                mac_can_be_placed = false;
+                break;
+            }
+            if (member_loc.y >= int(max_y) || member_loc.y < 0) { //member location is not on the chip
+                mac_can_be_placed = false;
+                break;
+            }
+        }
+        //if macro can not be placed in this head pos, change the head pos
+        if (!mac_can_be_placed) {
+            size_t macro_size = macro.members.size();
+            blk_locs[head_id].loc -= macro.members[macro_size - 1].offset;
+        }
+        //update other member's location based on head pos
         for (auto member = ++macro.members.begin(); member != macro.members.end(); ++member) {
             blk_locs[member->blk_index].loc = blk_locs[head_id].loc + member->offset;
         }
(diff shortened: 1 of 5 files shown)


vaughnbetz:
@kgugala : still hitting some google cloud resource failures (e.g. this PR is failing due to them). Sara, you will have to try re-running the failed tests later.

vpr/src/place/analytic_placer.cpp
```diff
@@ -410,8 +410,29 @@ void AnalyticPlacer::setup_solve_blks(t_logical_block_type_ptr blkTypes) {
...
+            if (member_loc.x >= int(max_x) || member_loc.x < 0) { //member location is not on the chip
+                mac_can_be_placed = false;
+                break;
+            }
+            if (member_loc.y >= int(max_y) || member_loc.y < 0) { //member location is not on the chip
+                mac_can_be_placed = false;
+                break;
+            }
+        }
+        //if macro can not be placed in this head pos, change the head pos
```
vaughnbetz, line 430: 1. Is this always safe or could the new macro head location also cause a problem?  If there's a chance it could cause a problem, it seems best to check if the macro can be placed again after you make this change.  
2. The code would be easier to read and the routine to check if a macro can be placed would be more reusable if you extracted the "mac_can_be_placed" logic into another helper function (can be called macro_can_be_placed).  Now checking if a macro can be placed after you adjust the head macro position is trivial (one more call).
3. There may also be an existing "macro_can_be_placed" routine -- worth doing a search of the placement code to see. @MohamedElgammal might also know.

vpr/src/place/cut_spreader.cpp
```diff
@@ -193,9 +193,17 @@ int CutSpreader::tiles_at(int x, int y) {
  * * grow merged to include all mergee grids
  */
 void CutSpreader::merge_regions(SpreaderRegion& merged, SpreaderRegion& mergee) {
+    size_t max_x = g_vpr_ctx.device().grid.width();
+    size_t max_y = g_vpr_ctx.device().grid.height();
     for (int x = mergee.bb.xmin(); x <= mergee.bb.xmax(); x++)
         for (int y = mergee.bb.ymin(); y <= mergee.bb.ymax(); y++) {
-            VTR_ASSERT(reg_id_at_grid[x][y] == mergee.id);
+            if (x >= max_x || x < 0 || y >= max_y || y < 0) { //location is not within the chip
```
vaughnbetz, line 200: This sequence of (1) compute max_x, max_y and (2) check if a location is in the grid comes up a few times. Code would be shorter if you made a helper function loc_is_in_grid (x. y).  If you're worried it's performance critical, you can write it in the header file and make it inline (then it won't really result in a function call, but the code will be more readable).
I don't think we have a function for that, but again worth a quick check of the code to see.

vaughnbetz (changes requested):
Generally looks good, but a couple of refactoring suggestions for helper routines, and I think there's still a vulnerability in the macro placement.

vpr/src/place/analytic_placer.cpp
saaramahmoudi: 1. I think it is safe but I can investigate more, and add this if necessary.
2. A function named "macro_can_be_placed" exists in the initial_placement.cpp. However, it also considers placement constraints which is not yet supported in Analytical Placement (I think). Should I go ahead and move the function to some file that is visible to both inital_placement.cpp and analytic_placer.cpp  or write a new one visible only to analytic_placer.cpp?
@vaughnbetz

vaughnbetz:
Best to use that one then.  If it checks placement constraints as well that's fine; we could put an earlier error check in saying the analytic placer doesn't support them and do a vpr_fatal_error if they exist.  That's better than ignoring them anyway.

Looks like there is also an is_loc_on_chip() function too.

You could move both them to place_util.cpp/.h.  That's probably the most clear place for them.  (could also delete the static specifier on them, but cleaner to move them as you suggest).

saaramahmoudi:
Applied all your suggestions to the code. @vaughnbetz

vpr/src/place/place_util.cpp
```diff
@@ -462,3 +463,71 @@ void set_block_location(ClusterBlockId blk_id, const t_pl_loc& location) {
...
+            break;
+        }
+
+        /*
+         * analytical placement approach do not need to make sure whether location could accommodate more blocks
+         * since overused locations will be spreaded by legalizer afterward.
+         * floorplan constraint is not supported by analytical placement yet, 
+         * hence, if macro_can_be_placed is called from analytical placer, no further actions are required. 
+         */
+        if (enable_analytic_placer) {
```
vaughnbetz, line 495: I'd delete this continue and the passing of enable_analytic_placer in here.  It's OK for the analytic placer not to support floorplan constraints for now, but that's best checked with a command line compatibility check, not by putting special skip code in lower-level routines.

vpr/src/place/place_util.h
```diff
@@ -221,4 +222,15 @@ void alloc_and_load_legal_placement_locations(std::vector<std::vector<std::vecto
...
 void set_block_location(ClusterBlockId blk_id, const t_pl_loc& location);
+
+/// @brief check if a specified location is within the device grid
+inline bool is_loc_on_chip(int x, int y) {
+    auto& device_ctx = g_vpr_ctx.device();
+    //return false if the location is not within the chip
+    return (x >= 0 && x < int(device_ctx.grid.width()) && y >= 0 && y < int(device_ctx.grid.height()));
+}
+
+/// @brief  Checks that each macro member location is legal based on the head position and its offset
```
vaughnbetz, line 233: Remove enable_analytic_placer (see earlier comment).

It's also not included in the comment here, and every parameter should be described; the fact that describing it seems odd here is a hint that it's not a great parameter to have.

vaughnbetz (changes requested):
Mostly good, but we shouldn't be passing enable_analytic_placement to a low level routine like macro_can_be_placed.  We can discuss tomorrow if this is unclear.

vpr/src/place/place_util.cpp
```diff
@@ -492,7 +492,7 @@ bool macro_can_be_placed(t_pl_macro pl_macro, t_pl_loc head_pos, bool enable_ana
          * floorplan constraint is not supported by analytical placement yet,
```
vaughnbetz, line 492: This comment should also mention that you're not checking the capacity restriction since the analytic placer checks that separately.

The more I think about it the more I think this parameter should also be renamed to check_all_legality or some such and the comment should explain that the analytic placer doesn't check everything. A parameter called analytic_placer doesn"t really explain to a future user trying to figure out if they can call this or not what the value of this parameter should be (since the parameter name does not describe what the parameter controls)

vpr/src/place/place_util.h
```diff
@@ -230,7 +230,21 @@ inline bool is_loc_on_chip(int x, int y) {
...
+ * If the function is called from initial_placement, it should ensure that the macro placement is legal.
+ * Each macro member should be placed in a location with the right type that can accommodate more blocks.
+ * If the function is called from analytical_placement, it should only ensure that all macro members are
+ * placed within the chip. The overused blocks will be spread by the strict_legalizer function.
+ *  
+ * @param pl_macro
+ *        macro's member can be accessible from pl_macro parameter. 
+ * @param head_pos
+ *        head_pos is the macro's head location.
+ * @param analytic_placer
```
vaughnbetz, line 245: not quite accurate as this parameter also determines if floorplan constraints are checked currently

Should rename the analytic_placement parameter and describe what it controls, and then give current use cases (initial_placement and annealer want fully legal, analytic can set check_all_legality to false).

vaughnbetz (changes requested):

vaughnbetz:
Looks good; thanks. Will merge as soon as CI goes green.

vaughnbetz:
Changes look good; thanks. 
Google has moved vtr runs to a larger cloud, so if you relaunch the failed tests they should hopefully work now.  Can you relaunch the appropriate ones?

saaramahmoudi:
Sure. I relaunch them right now.

Issue #2001, opened 2022-03-19
--graphics command doesn't seem to work


sildipankar:
Trying to use --graphics command on example runs in the document, but command gives error

#### Expected Behaviour
Should save a graphics file

#### Current Behaviour
Command gives error

#### Steps to Reproduce
Use the sample command  with graphics command

$VTR_ROOT/vpr/vpr     $VTR_ROOT/vtr_flow/arch/timing/EArch.xml     $VTR_ROOT/vtr_flow/benchmarks/blif/tseng.blif     --route_chan_width 100     --analysis --save_graphics on ----graphics_commands save_graphics place.png

####Error
Unexpected command-line argument 'place.png'

#### Your Environment
Version: 8.1.0-dev+08ec56b5d
Revision: v8.0.0-5295-g08ec56b5d
Compiled: 2022-03-17T20:40:39
Compiler: GNU 7.5.0 on Linux-4.9.201-tegra aarch64
Build Info: release IPO VTR_ASSERT_LEVEL=2

jmah76:
The documentation for the graphics commands are wrong. I think the correct command should be $VTR_ROOT/vpr/vpr $VTR_ROOT/vtr_flow/arch/timing/EArch.xml $VTR_ROOT/vtr_flow/benchmarks/blif/tseng.blif --route_chan_width 100 --analysis --save_graphics on ----graphics_commands "save_graphics place.png;" . 

If you're still interested, see PR #2055.

PR #2049, opened 2022-06-02
Refactored vpr/src/draw

jmah76:
Refactored the vpr/src/draw folder by splitting draw.cpp into smaller files to reduce file length.

#### Motivation and Context
Draw.cpp was too long (~3000-4000 lines), now much shorter (~1500 lines).

#### How It Was Tested
- VPR Graphics was run once with the changes, then without. The visualizations of both were compared, and were checked if they were relatively the same. 
- Command line scripts in the documentation were tested for expected behaviour. 
   - Note: As in Issue #2001, graphics_commands don't work (throws Unexpected command-line argument, could be that the commands aren't being recognized as graphics commands but normal command line options-- the error message matches that of an options error and not a graphics command error. I think it's because the OP of the options reading code forgot to add [subcommands](https://docs.python.org/3/library/argparse.html#sub-commands) to the --graphics_command argument, which would be needed to collect whatever comes after --graphics_command. This is a bigger issue than I thought, so I'll continue this convo in Issue #2001 ). This can be fixed in a seperate PR after this PR lands.   
- Buttons on the GUI were pressed and tested for expected behaviour.


packing_pin_util.rpt
```diff
@@ -0,0 +1,65 @@
+#Packing pin usage report
```
vaughnbetz, line 1: Is this file accidentally added?  It looks like an output file that shouldn't go in the repo.

vpr/src/draw/draw.cpp
```diff
@@ -34,6 +34,13 @@
 #include "globals.h"
 #include "draw_color.h"
 #include "draw.h"
+#include "draw_basic.h"
```
vaughnbetz, line 37: Would be good to add to the top-level comment at the top of this file: say what kinds of functions should go in this file (or say put new functions in helper files if at all possible).

vpr/src/draw/draw.cpp
```diff
@@ -1043,2427 +672,176 @@ void init_draw_coords(float width_val) {
...
     }
 }
 
-static void draw_congestion(ezgl::renderer* g) {
-    /* Draws all the overused routing resources (i.e. congestion) in various contrasting colors showing congestion ratio.   */
+/* This helper function determines whether a net has been highlighted. The highlighting
+ * could be caused by the user clicking on a routing resource, toggled, or
+ * fan-in/fan-out of a highlighted node.
+ */
+bool draw_if_net_highlighted(ClusterNetId inet) {
```
vaughnbetz, line 712: If there is no corresponding comment for this function in the .h file, duplicate or summarize this comment there (we're moving toward commenting what a function does in the .h file, and how it works in the .cpp file if necessary).

vpr/src/draw/draw.cpp
```diff
@@ -3500,229 +878,8 @@ ezgl::point2d atom_pin_draw_coord(AtomPinId pin) {
...
-            VTR_ASSERT(clb_src_block == clb_sink_block);
-        }
-    }
-
-    points.push_back(atom_pin_draw_coord(atom_sink_pin));
-}
-
 //Returns the set of rr nodes which connect driver to sink
-static std::vector<int> trace_routed_connection_rr_nodes(
+std::vector<int> trace_routed_connection_rr_nodes(
```
vaughnbetz, line 882: Duplicate this comment (or move it) in the .h file.

vpr/src/draw/draw.cpp
```diff
@@ -3787,7 +944,7 @@ bool trace_routed_connection_rr_nodes_recurr(const t_rt_node* rt_node,
 }
 
 //Find the edge between two rr nodes
-static t_edge_size find_edge(int prev_inode, int inode) {
+t_edge_size find_edge(int prev_inode, int inode) {
```
vaughnbetz, line 947: Duplicate this comment in the .h file

vpr/src/draw/draw.h
```diff
@@ -32,20 +33,16 @@ void alloc_draw_structs(const t_arch* arch);
...
-void draw_triangle_along_line(ezgl::renderer* g, ezgl::point2d loc, ezgl::point2d start, ezgl::point2d end, float arrow_size = DEFAULT_ARROW_SIZE);
-void draw_triangle_along_line(ezgl::renderer* g, float xend, float yend, float x1, float x2, float y1, float y2, float arrow_size = DEFAULT_ARROW_SIZE);
-
 const ezgl::color SELECTED_COLOR = ezgl::GREEN;
 const ezgl::color DRIVES_IT_COLOR = ezgl::RED;
 const ezgl::color DRIVEN_BY_IT_COLOR = ezgl::LIGHT_MEDIUM_BLUE;
 
 const float WIRE_DRAWING_WIDTH = 0.5;
 
+t_edge_size find_edge(int prev_inode, int inode);
```
vaughnbetz, line 42: Some of these have comments in the .cpp file --> copy or move those "what do I do" comments here.  If you know what some other functions do, go ahead and comment them here too. (Don't worry if you can't comment them all -- some is better than none).

vpr/src/draw/draw_basic.cpp
```diff
@@ -0,0 +1,1227 @@
+/*draw_basic.cpp contains all functions that draw things on the screen.*/
```
vaughnbetz, line 1: Can you be more specific?  I assume draw things means draw in the main graphics/visualization area?

vpr/src/draw/draw_mux.cpp
```diff
@@ -0,0 +1,133 @@
+/*draw_mux.cpp contains all functions that draw muxes.*/
```
vaughnbetz, line 1: Good to put a space after /* and before */; easier to read and easier to search for.

vpr/src/draw/draw_basic.cpp
vaughnbetz: Also, it seems this comment isn't correct, as there are also a draw_rr and draw_mux files.

vpr/src/draw/draw_searchbar.h
```diff
@@ -0,0 +1,52 @@
...
+#    include "search_bar.h"
+#    include "draw_debug.h"
+#    include "manual_moves.h"
+
+#    include "rr_graph.h"
+#    include "route_util.h"
+#    include "place_macro.h"
+#    include "buttons.h"
+
+ezgl::rectangle draw_get_rr_chan_bbox(int inode);
```
vaughnbetz, line 44: Some of these functions have comments on what they do in the .cpp file --> copy or move the overall what does this function do and how do I call it comment to the .h file where you can (more important in .h file than .cpp).

vpr/src/draw/draw_toggle_functions.cpp
```diff
@@ -0,0 +1,432 @@
+/*draw_toggle_functions.cpp contains functions that toggle things (mostly callback functions)*/
```
vaughnbetz, line 1: Explain what this means.  Control state variables connected to buttons and/or other UI widgets?

vpr/src/draw/draw_toggle_functions.h
```diff
@@ -0,0 +1,62 @@
...
+#    include "search_bar.h"
+#    include "draw_debug.h"
+#    include "manual_moves.h"
+
+#    include "rr_graph.h"
+#    include "route_util.h"
+#    include "place_macro.h"
+#    include "buttons.h"
+
+void toggle_nets(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data*/);
```
vaughnbetz, line 44: Copy or move "what function does" comments from .cpp to the .h file where you can.

vpr/src/draw/draw_triangle.cpp
```diff
@@ -0,0 +1,122 @@
+/* draw_triangle.cpp contains functions that draw triangles.*/
```
vaughnbetz, line 1: Can you say what they are used for?

vaughnbetz: (They are generally used for showing switching in the routing, the direction of signals, and such.)

vpr/src/draw/draw_triangle.h
```diff
@@ -0,0 +1,49 @@
...
+#    include "search_bar.h"
+#    include "draw_debug.h"
+#    include "manual_moves.h"
+
+#    include "rr_graph.h"
+#    include "route_util.h"
+#    include "place_macro.h"
+#    include "buttons.h"
+
+void draw_triangle_along_line(ezgl::renderer* g, ezgl::point2d start, ezgl::point2d end, float relative_position = 1., float arrow_size = DEFAULT_ARROW_SIZE);
```
vaughnbetz, line 44: Move function-level comments from .cpp file to here (the how to call and what do I do part).

vpr/src/draw/draw_types.h
```diff
@@ -163,6 +163,9 @@ typedef struct {
  * save_graphics_file_base: Base of save graphis file name (i.e. before extension)
  * pres_fac: present congestion cost factor
  */
+/* Note: t_draw_struct is used in the same way as a Context, but cannot be a Context because Contexts are
+ * not copyable, while t_draw_struct must be.
```
vaughnbetz, line 167: Say why it must be copied; graphics commands feature uses that to do save an restore of the drawing state.

vpr/src/draw/draw_xtoy.cpp
```diff
@@ -0,0 +1,571 @@
+/*draw_xtoy.cpp contains all functions that draw lines from one point to another.*/
```
vaughnbetz, line 1: I'd rename this file.  I initially read it as being a "toy".  It is really drawing the edges between routing resource (rr) nodes, so a better name would be 
draw_rr_edges.cpp

vaughnbetz (changes requested):
Nice refactoring; I have some further suggestions though.

Issue #1928, opened 2021-11-29
Request: Fail if timing is not met


tcal-x:
The user should know when their implementation does not meet their stated timing goal.   It is not clear from vpr when this is the case.

#### Proposed Behaviour
Perhaps by default or perhaps only by an option, vpr should have a "fail" exit status when it does not meet the timing constraints.   Also it should not print "VPR succeeded" in that case.  If it is an option, our wrapper can turn it on by default.

#### Current Behaviour
The exit status indicates success, and "VPR succeeded" is printed out, even when 

#### Possible Solution
I think the text above describes things well enough.  Please let me know if I'm mis-characterizing current vpr behavior -- we have wrappers on wrappers so I may be misplacing blame.

Perhaps it makes sense to limit this check to intra-domain paths.

#### Context
We had a user with a design that didn't work on the board.   The root issue was failed timing, but the user was given no indication that this was the problem.

https://github.com/enjoy-digital/litex/issues/1110
https://github.com/SymbiFlow/symbiflow-arch-defs/issues/2331

mithro:
I'm sure that we had discussed this previously, but I can't find an existing issue....

tcal-x:
@jgoeders , @vaughnbetz , any comment?

The concern is that the user finishes the build with no indication of any problems, and then the design does not run on the board, because it had failed timing badly.

Our workaround would be to have the SymbiFlow wrapper grep the router output for a negative WNS:
```
Final setup Worst Negative Slack (sWNS): -9.92936 ns
```
but this is not a robust solution.   We'd prefer to have the exit status indicated success or failure, where failure includes timing failure.

vaughnbetz:
It makes sense to return a different status code or new message to clearly indicate timing failed.  We (and others) have a lot of scripts that might break if we change the status code to be non-zero when execution finished normally, so that solution makes me a bit nervous.  @jgoeders: you know the flow script code better than I; how much would break if we returned a non-zero status for timing failure?  The problem is that many users and tests use: "as fast as possible" as their timing constraint, so while an impossible timing constraint isn't met, the results are meaningful and useful and they want the flow to produce results.

Easier would be to print a clear message saying timing failed, which could be examined by downstream tools, ad wouldn't break anything (for a new message) or much (if we change a message).

tcal-x:
[...]

PR #2052, opened 2022-06-07
Fail status if fail timing

jmah76:
Added command line option (--fail-status-if-fail-timing {on | off})to throw a VPR_FATAL_ERROR if a design doesn't meet its timing constraints. Related to Issue #1928 . 

#### Description
If command line option --fail-status-if-fail-timing is on, if negative slack occurs, VPR throws a VPR_FATAL_ERROR right before the flow completes. 

#### Related Issue
Related to Issue #1928 .

#### Motivation and Context
Context: Even if the build fails timing, it finishes with no indication of any problems. The design then does not run on the board.

Ran the command `vpr/vpr vtr_flow/arch/timing/EArch.xml vtr_flow/benchmarks/blif/tseng.blif --fail-status-if-fail-timing on` and check for expected output (for VPR to fail right before the flow completed.)

#### Types of changes
- [ X] New feature (change which adds functionality)

#### Checklist:

- [X ] I have updated the documentation accordingly
- [X ] All new and existing tests passed

### DIFF
--- a/vpr/src/place/analytic_placer.cpp
+++ b/vpr/src/place/analytic_placer.cpp
@@ -412,6 +412,18 @@ void AnalyticPlacer::setup_solve_blks(t_logical_block_type_ptr blkTypes) {
 void AnalyticPlacer::update_macros() {
     for (auto& macro : g_vpr_ctx.mutable_placement().pl_macros) {
         ClusterBlockId head_id = macro.members[0].blk_index;
+        bool mac_can_be_placed = macro_can_be_placed(macro, blk_locs[head_id].loc, true);
+
+        //if macro can not be placed in this head pos, change the head pos
+        if (!mac_can_be_placed) {
+            size_t macro_size = macro.members.size();
+            blk_locs[head_id].loc -= macro.members[macro_size - 1].offset;
+        }
+
+        //macro should be placed successfully after changing the head position
+        VTR_ASSERT(macro_can_be_placed(macro, blk_locs[head_id].loc, true));
+
+        //update other member's location based on head pos
         for (auto member = ++macro.members.begin(); member != macro.members.end(); ++member) {
             blk_locs[member->blk_index].loc = blk_locs[head_id].loc + member->offset;
         }
--- a/vpr/src/place/cut_spreader.cpp
+++ b/vpr/src/place/cut_spreader.cpp
@@ -11,6 +11,7 @@
 #    include "vtr_time.h"
 #    include "globals.h"
 #    include "vtr_log.h"
+#    include "place_util.h"
 
 // sentinel for base case in CutSpreader (i.e. only 1 block left in region)
 constexpr std::pair<int, int> BASE_CASE = {-2, -2};
@@ -173,13 +174,17 @@ void CutSpreader::init() {
     }
 }
 
-int CutSpreader::occ_at(int x, int y) { return occupancy[x][y]; }
+int CutSpreader::occ_at(int x, int y) {
+    if (!is_loc_on_chip(x, y)) {
+        return 0;
+    }
+    return occupancy[x][y];
+}
 
 int CutSpreader::tiles_at(int x, int y) {
-    int max_x = g_vpr_ctx.device().grid.width();
-    int max_y = g_vpr_ctx.device().grid.height();
-    if (x >= max_x || y >= max_y)
+    if (!is_loc_on_chip(x, y)) {
         return 0;
+    }
     return int(subtiles_at_location[x][y].size());
 }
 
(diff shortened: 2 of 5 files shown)


jmah76:
I think I messed up and branhced this branch off of the draw refactor one.

vpr/src/base/read_options.cpp
```diff
@@ -1474,6 +1474,12 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("off")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    gen_grp.add_argument<bool, ParseOnOff>(args.fail_status_if_fail_timing, "--fail-status-if-fail-timing")
+        .help(
+            "During final timing analysis after routing, if a negative slack anywhere is returned and this option is set, \n"
+            "VPR_FATAL_ERROR is called and stops processing.")
```
vaughnbetz, line 1480: "and stops processing" -> "and processing ends."

vpr/src/base/read_options.cpp
```diff
@@ -1474,6 +1474,12 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("off")
```
vaughnbetz, line 1474: Remove tseng.net.post_routing (added by accident I think).

vpr/src/base/read_options.cpp
```diff
@@ -1474,6 +1474,12 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
...
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    gen_grp.add_argument<bool, ParseOnOff>(args.fail_status_if_fail_timing, "--fail-status-if-fail-timing")
```
vaughnbetz, line 1477: I'd rename this.  
1. We use _ to separate words in command line options (vpr convention) so we should stick with that.
2. terminate_if_timing_fails would be a better name, as it indicates what happens if this is "on".

vpr/src/base/vpr_context.h
```diff
@@ -95,6 +95,10 @@ struct TimingContext : public Context {
     std::shared_ptr<tatum::TimingConstraints> constraints;
 
     t_timing_analysis_profile_info stats;
+
+    /* Represents whether or not VPR should fail if timing constraints aren't met. */
+    bool fail_status_if_fail_timing = false;
+    std::string fail_timing_msg = "";
```
vaughnbetz, line 101: Does this message have to be in the context?  Ideally contexts are major items read in many places so it would be better to keep this message out of the context if possible.

vpr/src/main.cpp
```diff
@@ -69,6 +69,8 @@ int main(int argc, const char** argv) {
         auto& timing_ctx = g_vpr_ctx.timing();
         print_timing_stats("Flow", timing_ctx.stats);
 
+        error_if_timing_failed();
```
vaughnbetz, line 72: Could this be moved into print_timing_stats (or another timing routine) so it isn't visible at the top level?  Ideally we'd keep the top level flow pretty short.

vpr/src/timing/timing_util.cpp
```diff
@@ -637,6 +643,21 @@ void print_hold_timing_summary(const tatum::TimingConstraints& constraints, cons
...
+}
+
+/* Checks if slack is negative and if user wants VPR flow to fail if their desisn doesn't meet timing
+ * constraints. If both conditions are true, the function adds details about the negative slack to a string
+ * that will be printed when VPR throws an error.
+ */
+void check_if_failed_timing_constraints(double& slack, std::string slack_name) {
+    auto& timing_ctx = g_vpr_ctx.mutable_timing();
+    if (timing_ctx.fail_status_if_fail_timing && slack < 0) {
+        timing_ctx.fail_timing_msg = timing_ctx.fail_timing_msg + "Final " + std::to_string(slack) + " Negative Slack of " + slack_name.c_str() + " was found to be negative. \n";
```
vaughnbetz, line 658: avoid using fail_timing_msg (passing a variable all the way through a top level context to a single routine -- set up the message here instead).
Message should be more explicit: say you're terminating processing due to a negative slack and option xx set.

vpr/src/util/vpr_utils.cpp
```diff
@@ -2021,3 +2021,15 @@ void print_timing_stats(std::string name,
             current.num_full_hold_updates - past.num_full_hold_updates,
             current.num_full_setup_hold_updates - past.num_full_setup_hold_updates);
 }
+
+void error_if_timing_failed() {
+    auto& timing_ctx = g_vpr_ctx.mutable_timing();
+    // Everytime a negative slack is found, it adds on to timing_ctx.fail_timing_msg where it failed.
+    // If timing_ctx.fail_timing_msg is empty, then no negative slacks were found.
+    if (timing_ctx.fail_timing_msg != "") {
```
vaughnbetz, line 2029: Seems like this could be localized to a timing_something.cpp file, instead of being put in a general utility file.  This probably isn't a utility that many will call.

vaughnbetz:
@tcal-x can you check if this meets your use case?
@jmah76 : you should redo the PR vs. the master so the various draw updates (now merged) don't show up.

vaughnbetz:
Asked @acomodi about the use case: he agrees we should only fail if timing isn't met after routing (not in earlier stages).

PR #2058, opened 2022-06-09
Do not make models with no pin-to-pin relations constant generators

mkurc-ant:
#### Description
Currently for models with no pin-to-pin relations defined VPR assumes that their outputs become constant generators even though such models have inputs and they are connected to active logic.

This PR makes VPR stop implying that.

#### Motivation and Context

Generic blackbox support

#### Types of changes

Not sure though how breaking the change is. Probably a discussion is needed.

#### Checklist:

### DIFF
--- a/vpr/src/base/atom_netlist_utils.cpp
+++ b/vpr/src/base/atom_netlist_utils.cpp
@@ -471,7 +471,7 @@ int mark_undriven_primitive_outputs_as_constant(AtomNetlist& netlist, int verbos
                 }
             }
 
-            if (!has_connected_inputs) {
+            if (!has_connected_inputs && !upstream_ports.empty()) {
                 //The current output port has no inputs driving the primitive's internal
                 //timing edges. Therefore we treat all its pins as constant generators.
                 for (AtomPinId output_pin : netlist.port_pins(output_port)) {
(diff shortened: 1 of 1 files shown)


vpr/src/base/atom_netlist_utils.cpp
```diff
@@ -471,7 +471,7 @@ int mark_undriven_primitive_outputs_as_constant(AtomNetlist& netlist, int verbos
                 }
             }
 
-            if (!has_connected_inputs) {
+            if (!has_connected_inputs && !upstream_ports.empty()) {
```
vaughnbetz, line 474: This needs a comment explaining what !upstream_ports.empty() indicates (why is this not a constant generator if none of its inputs are connected to anything?). An example would be useful.

vaughnbetz (changes requested):
Not quite sure why the code change is as structured; needs a comment. Also, some tests are failing, so there seems to be some issue.

vpr/src/base/atom_netlist_utils.cpp
vaughnbetz: Wouldn't a black box still have some inputs of the atom connected (so upstream_ports wouldn't necessarily be empty)?  I am likely missing the point, but that means a comment with an example is needed I think.

vaughnbetz:
@kmurray : this is a very short PR that would be good for you to take a look at as it's about timing analysis / graph.

mkurc-ant:
@vaughnbetz @kmurray Indeed I see that the change I've made is breaking. Basically what I wanted to achieve is to make VPR not consider output of black-box models as constant sources. This is needed (at least for now) for FPGA interchange where we do not have (currently) pin-to-pin relations defined. With the current VPR we see that all cell outputs are considered constants which makes the placer ignore all nets hence it hangs in an infinite loop. I thought that this change might be beneficial for VPR in general as it makes blackbox models more usable.

PR #2061, opened 2022-06-09
Complete Search Revamp | Working Autocomplete

SebastianLievano:
#### Description
Added auto-complete options to the Block and Net Name search functions, managed by a Min. Search Term value.
The value is the min. number of characters before it searches through all names to find any that contain (not start with) the search term. At that point, the program may experience slight delays (1-2 sec.) as it searches, so its recommended to keep the Min. Value near the length of your search term.

Additionally, the search function is now able to find and highlight internal components. Based on lvl of internals visibility, it will highlight either the AtomBlock or the ClusterBlock that contains it.  

Also added a function able to highlight an Atom Block just from its Id, which was not present before. May be useful to someone. 

#### Motivation and Context

The original search function had no form of automatic completion, required users to enter the full names of the block or net they were searching for, with no room for error, and was also fully based on the Clustered netlist, and thus could not search or highlight internals. It was pretty difficult to use as a result, as you'd need to know the full name of the thing you were searching for, and the names are often synthesized by VTR so you probably would not know that.

That has been rectified through the new features implemented in this PR. 

Tested using PR auto-tester as well as visual tests to determine performance. Originally, auto-complete was too slow so key length selecting feature was implemented to mitigate this. 

#### Types of changes

### DIFF
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -357,7 +357,7 @@ static void initial_setup_NO_PICTURE_to_PLACEMENT(ezgl::application* app,
                                    "Block Name");
     gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "Net ID");
     gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "Net Name");
-    gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "RR Node ID");                                                      
+    gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "RR Node ID");
     //Important to have default option set, or else user can search w. no selected type which can cause crash
     gtk_combo_box_set_active((GtkComboBox*)search_type, 0); // default set to Block ID which has an index 0
     g_signal_connect(search_type, "changed", G_CALLBACK(search_type_changed), app);
@@ -455,7 +455,7 @@ static void initial_setup_NO_PICTURE_to_ROUTING(ezgl::application* app,
                                    "Block Name");
     gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "Net ID");
     gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "Net Name");
-    gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "RR Node ID");                                                      
+    gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(search_type), "RR Node ID");
     //Important to have default option set, or else user can search w. no selected type which can cause crash
     gtk_combo_box_set_active((GtkComboBox*)search_type, 0); // default set to Block ID which has an index 0
     g_signal_connect(search_type, "changed", G_CALLBACK(search_type_changed), app);
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -2891,12 +2891,31 @@ static bool highlight_rr_nodes(float x, float y) {
 }
 
 #    if defined(X11) && !defined(__MINGW32__)
-void act_on_key_press(ezgl::application* /*app*/, GdkEventKey* /*event*/, char* key_name) {
+void act_on_key_press(ezgl::application* app, GdkEventKey* /*event*/, char* key_name) {
     //VTR_LOG("Key press %c (%d)\n", key_pressed, keysym);
     std::string key(key_name);
+    std::cout << "Pressed " << key << std::endl;
+    if(gtk_widget_is_focus(GTK_WIDGET(app->get_object("TextInput")))){
+        std::cout << "Typing" << std::endl;
+        if(key == "Return"){
+            std::cout << "Loading autocomplete results" << std::endl;
+            enable_autocomplete(app);
+        }
+    }
 }
 #    else
-void act_on_key_press(ezgl::application* /*app*/, GdkEventKey* /*event*/, char* /*key_name*/) {
+void act_on_key_press(ezgl::application* app, GdkEventKey* /*event*/, char* key_name) {
+    std::string key(key_name);
+    std::cout << "Pressed " << key << std::endl;
+    GtkWidget* searchBar = GTK_WIDGET(app->get_object("TextInput"));
+    if(gtk_widget_is_focus(searchBar)){
+        std::cout << "Typing" << std::endl;
+        std::string oldText(gtk_entry_get_text(GTK_ENTRY(searchBar)));
+        if(key == "Return"){
+            std::cout << "Loading autocomplete results" << std::endl;
+            enable_autocomplete(app);
+        }
+    }
 }
 #    endif
 
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -369,12 +369,6 @@ static void initial_setup_NO_PICTURE_to_PLACEMENT(ezgl::application* app,
     GtkEntryCompletion* wildcardComp = GTK_ENTRY_COMPLETION(app->get_object("Completion"));
     gtk_entry_completion_set_match_func(wildcardComp, (GtkEntryCompletionMatchFunc)customMatchingFunction, NULL, NULL);
 
-    //Hiding Match length Selecters
-    GtkSpinButton* completionKeyLength = GTK_SPIN_BUTTON(app->get_object("KeyLength"));
-    g_signal_connect(completionKeyLength, "value-changed", G_CALLBACK(key_length_val_changed), app);
-    gtk_widget_hide(GTK_WIDGET(completionKeyLength));
-    GtkWidget* keyLengthLabel = GTK_WIDGET(app->get_object("KeyLengthLabel"));
-    gtk_widget_hide(keyLengthLabel);
     button_for_toggle_nets();
     button_for_net_max_fanout();
     button_for_net_alpha();
@@ -468,9 +462,6 @@ static void initial_setup_NO_PICTURE_to_ROUTING(ezgl::application* app,
     gtk_entry_completion_set_match_func(completion, (GtkEntryCompletionMatchFunc)customMatchingFunction, NULL, NULL);
 
     //Hiding Match length Selecters
-    GtkSpinButton* completionKeyLength = GTK_SPIN_BUTTON(app->get_object("KeyLength"));
-    g_signal_connect(completionKeyLength, "value-changed", G_CALLBACK(key_length_val_changed), app);
-    gtk_widget_hide(GTK_WIDGET(completionKeyLength));
     GtkWidget* keyLengthLabel = GTK_WIDGET(app->get_object("KeyLengthLabel"));
     gtk_widget_hide(keyLengthLabel);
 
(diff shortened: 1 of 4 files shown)


SebastianLievano:
Will fix formatting tmrw

vaughnbetz:
Sounds great.  Going through the code now.
An idea (may not be feasible but thought I'd write it down): if you didn't start the search/autocomplete until the user hit Ctrl+D or enter, would that be more intuitive than having a Min Search Term value?

vpr/src/draw/draw.cpp
```diff
@@ -436,18 +448,31 @@ static void initial_setup_NO_PICTURE_to_ROUTING(ezgl::application* app,
...
-    load_block_names(app);
+
+    load_block_names(app); //Loading block and net names into GtkListStores
+    load_net_names(app);
+
+    //Setting custom matching function for entry completion (searches whole string instead of start)
+    GtkEntryCompletion* completion = GTK_ENTRY_COMPLETION(app->get_object("Completion"));
+    gtk_entry_completion_set_match_func(completion, (GtkEntryCompletionMatchFunc)customMatchingFunction, NULL, NULL);
+
+    //Hiding Match length Selecters
```
vaughnbetz, line 470: Good to comment why you need the minimum match length and how it works.

vpr/src/draw/search_bar.cpp
```diff
@@ -96,14 +98,33 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
...
         std::string block_name = "";
         ss >> block_name;
 
-        highlight_blocks((std::string)block_name);
+        AtomBlockId atom_blk_id = AtomBlockId::INVALID();
+        atom_blk_id = atom_ctx.nlist.find_block(block_name);
+        //If block is found, the CLB containing it is highlighted
+        if (atom_blk_id != AtomBlockId::INVALID()) {
+            ClusterBlockId block_id = atom_ctx.lookup.atom_clb(atom_blk_id);
+            //Highlighting atom block. IF function returns false, highlighting clb that contains
```
vaughnbetz, line 113: Comment is a bit confusing.  
contains -> contains it  (I think)
What does this function returning false mean? That's what the comment should explain.

vpr/src/draw/search_bar.cpp
```diff
@@ -96,14 +98,33 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
...
+            if (!highlight_atom_block(atom_blk_id, block_id, app)) {
+                highlight_cluster_block(block_id);
+            }
+            return;
+        }
+        //Continues if atom block not found (Checking if user searched a clb)
+        ClusterBlockId block_id = ClusterBlockId::INVALID();
```
vaughnbetz, line 120: Do we need to set this to INVALID? Couldn't we just collapse these two lines into one, as the block_id is updated no matter what?

vpr/src/draw/search_bar.cpp
```diff
@@ -96,14 +98,33 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
...
+        highlight_cluster_block((ClusterBlockId)block_id);
     }
 
     else if (search_type == "Block Name") {
         std::string block_name = "";
         ss >> block_name;
 
-        highlight_blocks((std::string)block_name);
+        AtomBlockId atom_blk_id = AtomBlockId::INVALID();
+        atom_blk_id = atom_ctx.nlist.find_block(block_name);
```
vaughnbetz, line 109: Do we need to set this to INVALID? Couldn't we just collapse these two lines into one, as the block_id is updated no matter what?

vpr/src/draw/search_bar.cpp
```diff
@@ -96,14 +98,33 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
             return;
         }
 
-        highlight_blocks((ClusterBlockId)block_id);
+        highlight_cluster_block((ClusterBlockId)block_id);
     }
 
     else if (search_type == "Block Name") {
         std::string block_name = "";
         ss >> block_name;
```
vaughnbetz, line 107: Good to comment the overall logic.  Check if it is an atom name; if not, check if it is a cluster name.

vpr/src/draw/search_bar.cpp
```diff
@@ -123,8 +144,15 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
     else if (search_type == "Net Name") {
         std::string net_name = "";
         ss >> net_name;
+        AtomNetId atom_net_id = AtomNetId::INVALID();
+        atom_net_id = atom_ctx.nlist.find_net(net_name);
```
vaughnbetz, line 148: Can collapse these two lines I think (no need for invalid preset).
Should also comment why you only need to look through the atom net names (all clb nets are a subset of atom net names)?

vpr/src/draw/search_bar.cpp
```diff
@@ -123,8 +144,15 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
...
 
-        highlight_nets((std::string)net_name);
+        if (atom_net_id == AtomNetId::INVALID()) {
+            warning_dialog_box("Invalid Net Name");
+            return; //name not exist
+        }
+        ClusterNetId clb_net_id = atom_ctx.lookup.clb_net(atom_net_id);
+        highlight_nets(clb_net_id);
```
vaughnbetz, line 155: This is a long routine -- suggest making some helper functions for the block names, net names etc. so it's shorter.

vpr/src/draw/search_bar.cpp
```diff
@@ -254,45 +287,82 @@ void highlight_blocks(ClusterBlockId clb_index) {
     application.refresh_drawing();
 }
 
-void highlight_nets(ClusterNetId net_id) {
-    t_trace* tptr;
-    auto& route_ctx = g_vpr_ctx.routing();
+/**
+ * @brief Finds and highlights the atom block. Returns true if block shows, false if not
```
vaughnbetz, line 291: What does "block shows" mean?  I think saying "block found and will be visible"

vpr/src/draw/search_bar.cpp
```diff
@@ -254,45 +287,82 @@ void highlight_blocks(ClusterBlockId clb_index) {
...
+ * @brief Recursively looks through pb graph to find block w. given name
+ * 
+ * @param name name of block being searched for
+ * @param pb current node to be examined
+ * @return t_pb* t_pb ptr of block w. name "name"
+ */
+t_pb* find_atom_block_in_pb(std::string name, t_pb* pb) {
+    //Checking if block is one being searched for
+    std::string pbName(pb->name);
+    if (pbName == name) return pb;
```
vaughnbetz, line 330: Prefer to have simple if statements like this broken over two lines.  Not sure if make format will do that.

vpr/src/draw/search_bar.cpp
```diff
@@ -254,45 +287,82 @@ void highlight_blocks(ClusterBlockId clb_index) {
...
-
-    if (net_id == ClusterNetId::INVALID()) {
-        warning_dialog_box("Invalid Net Name");
-        return; //name not exist
+/**
+ * @brief Recursively looks through pb graph to find block w. given name
+ * 
+ * @param name name of block being searched for
+ * @param pb current node to be examined
+ * @return t_pb* t_pb ptr of block w. name "name"
```
vaughnbetz, line 325: Also say returns nullptr if no pb with a matching name is found.

vpr/src/draw/search_bar.cpp
```diff
@@ -254,45 +287,82 @@ void highlight_blocks(ClusterBlockId clb_index) {
...
 
-    ClusterBlockId block_id = ClusterBlockId::INVALID();
-    block_id = cluster_ctx.clb_nlist.find_block(block_name);
+    t_draw_state* draw_state = get_draw_state_vars();
 
-    if (block_id == ClusterBlockId::INVALID()) {
-        warning_dialog_box("Invalid Block Name");
-        return; //name not exist
-    }
+    if (int(route_ctx.trace.size()) == 0) return;
```
vaughnbetz, line 361: Comment this -- no routing exists  
(I think)

vpr/src/draw/search_bar.cpp
```diff
@@ -254,45 +287,82 @@ void highlight_blocks(ClusterBlockId clb_index) {
...
 
-    highlight_blocks(block_id); //found block
+    for (tptr = route_ctx.trace[net_id].head; tptr != nullptr; tptr = tptr->next) {
```
vaughnbetz, line 363: I don't think you need this loop.  It is setting the colour on the same net over and over (once per wire in the net routing), but you should just be able to set it once.

vpr/src/draw/search_bar.cpp
```diff
@@ -336,34 +406,111 @@ void warning_dialog_box(const char* message) {
...
+    GtkWidget* keyLengthSpinButton = GTK_WIDGET(app->get_object("KeyLength"));
+    GtkWidget* keyLengthLabel = GTK_WIDGET(app->get_object("KeyLengthLabel"));
     //Ensuring a valid type was selected
     if (!type) return;
     if (type[0] == '\0') return;
     std::string searchType(type);
-    //Setting active completion model to blockCompleter if type selected is block Name
+
+    /*
+     * If search type is name, connecting search bar to completion,
```
vaughnbetz, line 420: This might be a good place for a comment about key length and its performance impact (probably better than the earlier place I listed).

vpr/src/draw/search_bar.cpp
```diff
@@ -336,34 +406,111 @@ void warning_dialog_box(const char* message) {
...
+        gtk_widget_show(keyLengthSpinButton);
+        gtk_widget_show(keyLengthLabel);
+    } else { //setting to null if option does not require auto-complete
+        gtk_widget_hide(keyLengthSpinButton);
+        gtk_widget_hide(keyLengthLabel);
         gtk_entry_set_completion(searchBar, nullptr);
     }
 }
 
 /**
```
vaughnbetz, line 441: Nice doxygen comments, but better to put them in the .h file.

vpr/src/draw/search_bar.cpp
```diff
@@ -336,34 +406,111 @@ void warning_dialog_box(const char* message) {
...
+    //Loading net names
+    for (AtomNetId id : atom_ctx.nlist.nets()) {
+        gtk_list_store_append(netStorage, &iter);
+        gtk_list_store_set(netStorage, &iter,
+                           0, (atom_ctx.nlist.net_name(id)).c_str(), -1);
+    }
+}
+
+/**
+ * @brief A new matching function used for the wildcard search options
```
vaughnbetz, line 481: I'd explain this matching function searches for a substring anywhere in the name, rather than just a prefix (which is the default matcher).
I wouldn't call this wildcard since it doesn't do wildcards.
You say this will be slower --> if you have data on relative speed put that in the comment.

vpr/src/draw/search_bar.cpp
```diff
@@ -336,34 +406,111 @@ void warning_dialog_box(const char* message) {
...
+    g_utf8_casefold(text, -1);
+    g_utf8_normalize(text, -1, G_NORMALIZE_DEFAULT);
+    std::string cppText(text);
+    //If substring not found, returning false;
+    return (cppText.find(key, 0) != std::string::npos);
+}
+
+/**
+ * @brief Updates min. key length when the value in spin button changes
+ */
```
vaughnbetz, line 509: I'd give a bit more context on this one -- explain what the key length is for. Can also say this is a callback for when the user clicks on an arrow to update the spin button controlling the length (which I think is the case).

vpr/src/draw/search_bar.h
```diff
@@ -13,19 +13,32 @@
 void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app);
```
vaughnbetz, line 13: Put a comment at the top of the file saying what kinds of functions should go in here (searchbar related, and the search bar is for ...).
Move the function level comments from the .cpp file to this .h file (better convention).
Comments about how the function works go in the .cpp file; comments on what it does and how to call it go in the .h

vaughnbetz (changes requested):
Looks good; some commenting and minor changes suggested.

vaughnbetz:
Not sure if this has been merged into another PR as you mentioned; if not, needs an update.

SebastianLievano:
Closing this one. New PR created w. updated version

PR #2066, opened 2022-06-15
Add Intra-Cluster Pins & Connections to RR Graph

amin1377:
In this PR, the routing resources (pins & connections) inside of the tiles, to which netlist's block has been mapped to it, are added to the RR Graph. To enable this feature, *flat_routing* flag should be set in the command line (--flat_routing true.) The default value of this flag is false. 

#### Description
After placement is done, the location of each block in the netlist is determined. If *flat_routing* flag is set, the RR Graph would be created from the **ground up** when routing starts. The other option was to only add extra resources to the existing graph instead of creating the graph from scratch. We decided to go with the first option to take advantage of the already implemented optimizations and preserve memory contiguity among the resources which are in the same tile.

In XML file, different logical blocks(root-level pb_type) has been defined. One or multiple logical blocks can be mapped to each tile. To be able to uniquely identify the pins inside each tile, we assign a unique number, which we called it physical number, to each pin of these logical blocks. Thus, by having the pin number and the tile, the pb_type associated with that pin can be determined. To make this indexing system compatible with the existing implementation, the pins on the tile are indexed as before, and the physical number for internal pins starts at n=*number of the pins on the tile*. Several helper functions are added to physical_types_util.h/cpp. These functions are used to find the blocks associated with each pin or vice versa. Regarding SINK/SOURCE pins, we add these pins for all levels of pb_blocks. Physical numbers assigned to these pins are also compatible with the existing systems. There are some helper functions in physical_types_util.h/cpp which return the block associated with the given pin number or vice versa.

Another indexing system is also defined. This indexing system assigned a unique number, which is called a logical number, to each pin, and this number is unique at the logical block level(root-level pb_block). The physical number is assigned by adding an offset to the logical number of each pin to make them unique at the tile level.

In this implementation, we create the RR Graph after placement. It is because we need to know the tiles to which netlist's blocks are mapped to. Moreover, since each block can contain several modes, we need to know the block mode to only add the resources which are actually being used.

In the process of adding intra-tile routing resources to RR Graph, we don't need to add the root-level pb_block pins to the RR Graph since the tile's pins are already added. Thus, we skip this level and add the pins of sub-blocks to the RR Graph.

#### Motivation and Context
This is the first step toward implementing a flat-router in VPR. Our final goal is to route the signals from the source primitive to the sink primitives. To do so, we first have to add the required resources to the RR Graph. Then, we need to change the router so that it can work with the new RR Graph.

Two of the biggest benchmark sets were used to examine the changes: *nightly_test2/titan_quick_qor* and *nightly_test3/vtr_reg_qor_chain_large.*

In the case that the flat_routing flag is set, intra-cluster resources are added to the RR Graph. However, intra-cluster routing does not take place although some newly added nodes may be traversed by the router.

You can find the results [here](https://utoronto-my.sharepoint.com/:x:/g/personal/amin_mohaghegh_mail_utoronto_ca/EcMOaclo-L1Dj6VT_z8WKAcB9NJMA9sePLmBXAffROxesg)

#### Types of changes

#### Checklist:

@vaughnbetz


vaughnbetz:
CI failures: make format (just run it)
One QoR failure on titan_qor_other: murax has a peak vpr memory footprint that went up more than the 20% limit (about 25%). It's the only failure, and when re-run on wintermute, it passes.

doc/src/vpr/file_formats.rst
```diff
@@ -411,15 +411,13 @@ For example:
     .param multiplier 0.50
     .param power 001101
 
-Would set the parameters ``feedback``, ``multiplier`` and ``power``  of the above ``pll`` ``.subckt`` to ``"internal"``, ``0.50`` and ``001101`` respectively.
-
-.. warning:: Integers in notation other than binary (e.g. decimal, hexadecimal) are not supported. Occurrence of params with digits other than 1 and 0 for binary words, not quoted (strings) or not separated with dot ``.`` (real numbers) are considered to be illegal.
```
MohamedElgammal, line 416: Did you add support for decimal and hexa? otherwise the warning should be kept in the docs.

libs/libarchfpga/src/physical_types.h
```diff
@@ -450,6 +450,10 @@ struct t_class {
 struct t_class_range {
     int low = 0;
     int high = 0;
+
+    int total() const {
```
MohamedElgammal, line 454: Can you add a small comment to explain it?e.g. "returns the total number of elements in this range". Also, total_num( ) may be a better name

doc/src/vpr/file_formats.rst
amin1377: Actually, I did not make these changes, so I don't know whether the support is added. It seems that @lpawelcz and @mkurc-ant accidentally change my branch. If you think it should be reversed, I will do that.

libs/libarchfpga/src/physical_types.h
```diff
@@ -840,6 +853,10 @@ struct t_logical_block_type {
     std::vector<t_physical_tile_type_ptr> equivalent_tiles; ///>List of physical tiles at which one could
                                                             ///>place this type of netlist block.
 
+    std::unordered_map<int, const t_pb_graph_pin*> pb_pin_num_map; /* pb_pin_num_map[pin logical number] -> pb_graph_pin ptr} */
```
MohamedElgammal, line 856: pb_pin_logical_num_to_pb_graph_pin_mapping  

may be better name?

libs/libarchfpga/src/physical_types.h
```diff
@@ -840,6 +853,10 @@ struct t_logical_block_type {
...
+    std::unordered_map<const t_pb_graph_pin*, int> pb_pin_class_map; /* pb_pin_class_map[pb_graph_pin ptr] -> class logical number */
```
MohamedElgammal, line 857: pb_pin_graph_pin_to_pb_pin_logical_num   may be better name?

libs/libarchfpga/src/physical_types.h
```diff
@@ -840,6 +853,10 @@ struct t_logical_block_type {
...
+    std::vector<t_class> logical_class_inf; /* logical_class_inf[class_logical_number] -> class */
```
MohamedElgammal, line 858: logical classes[class_logical_number] may be better naming?

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -35,35 +36,102 @@ struct t_pin_inst_port {
...
-static std::tuple<int, int, int> get_pin_index_for_inst(t_physical_tile_type_ptr type, int pin_index) {
-    VTR_ASSERT(pin_index < type->num_pins);
+// #TODO: remove the default value for is_flat
+static t_pin_inst_port block_type_pin_index_to_pin_inst(t_physical_tile_type_ptr type, int pin_index, bool is_flat = false);
 
-    int total_pin_counts = 0;
-    int pin_offset = 0;
-    for (auto& sub_tile : type->sub_tiles) {
-        total_pin_counts += sub_tile.num_phy_pins;
+static int get_sub_tile_num_internal_classes(const t_sub_tile* sub_tile);
```
MohamedElgammal, line 44: We need to add comments on the new function explaning what do these function do and how to use them: what are the inputs and what are the outputs.

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -35,35 +36,102 @@ struct t_pin_inst_port {
...
+static std::vector<int> get_connected_child_pins(t_physical_tile_type_ptr physical_type,
+                                                 const t_sub_tile* sub_tile,
+                                                 t_logical_block_type_ptr logical_block,
+                                                 int relative_cap,
+                                                 int mode_num,
+                                                 const t_pb_graph_pin* pin);
+
+static std::vector<const t_pb_graph_node*> get_child_pb_graph_node_mode(const t_pb_graph_node* parent_node, int mode_num);
+
+static std::tuple<int, int, int, int> get_pin_index_for_inst(t_physical_tile_type_ptr type, int pin_index, bool is_flat) {
```
MohamedElgammal, line 91: comment to explain what is the order of the output tuple?

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
 
     return nullptr;
 }
+
+/* Access information related to pin classes */
+
+/** get information given class physical num **/
+
+std::tuple<const t_sub_tile*, int> get_sub_tile_from_class_physical_num(t_physical_tile_type_ptr physical_tile, int physical_class_num) {
```
MohamedElgammal, line 865: All the functions implementations don't have any comments. High-level comments on what is done in the function implementation is very useful

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
...
+    } else {
+        VTR_ASSERT(is_on_tile == true);
+        class_type = physical_tile->class_inf[physical_class_num].type;
+    }
+
+    return class_type;
+}
+
+int get_class_num_pins_from_class_physical_num(t_physical_tile_type_ptr physical_tile, int physical_class_num) {
+    int num_pins = -1;
```
MohamedElgammal, line 929: use INVALID_INT instead

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
...
+
+    return num_pins;
+}
+
+
+int get_pin_physical_num_from_class_physical_num(t_physical_tile_type_ptr physical_tile, int physical_class_num, int pin_logical_num) {
+    bool is_on_tile = is_class_on_tile(physical_tile, physical_class_num);
+    const t_sub_tile* sub_tile;
+    int sub_tile_cap;
+    int pin_physical_num = -1;
```
MohamedElgammal, line 951: INVALID_INT?

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
...
+
+bool is_class_on_tile(t_physical_tile_type_ptr physical_tile, int class_physical_num) {
+    if(class_physical_num < (int)physical_tile->class_inf.size()){
+        return true;
+    } else {
+        return false;
+    }
+}
+
+/** **/
```
MohamedElgammal, line 979: remove

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
...
+                                       logical_block,
+                                       sub_tile_cap,
+                                       pb_pin);
+    }
+
+    return pin_physical_num;
+}
+
+bool is_class_on_tile(t_physical_tile_type_ptr physical_tile, int class_physical_num) {
+    if(class_physical_num < (int)physical_tile->class_inf.size()){
```
MohamedElgammal, line 972: return (class_physical_num < (int)physical_tile->class_inf.size());

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
...
+
+
+/* Access information related to pins */
+
+/** get information given pin physical number **/
+std::tuple<const t_sub_tile*, int> get_sub_tile_from_pin_physical_num(t_physical_tile_type_ptr physical_tile, int physical_num) {
+
+
+    const t_sub_tile* target_sub_tile = nullptr;
+    int target_sub_tile_cap = -1;
```
MohamedElgammal, line 1116: INVALID

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
...
+
+bool is_pin_on_tile(t_physical_tile_type_ptr physical_tile, int physical_num) {
+    if(physical_num < physical_tile->num_pins) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+/** **/
```
MohamedElgammal, line 1218: add comment or remove

libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -490,3 +857,494 @@ const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin) {
...
+
+    return num_pins;
+}
+
+int get_tile_ipin_opin_max_ptc(t_physical_tile_type_ptr tile, bool is_flat) {
+
+    if(is_flat) {
+        return tile->num_pins + get_total_num_tile_internal_pins(tile);
+    } else {
+        VTR_ASSERT(is_flat == false);
```
MohamedElgammal, line 1343: No need for this assertion

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -301,4 +302,93 @@ const t_physical_tile_port* get_port_by_pin(const t_sub_tile* sub_tile, int pin)
...
 const t_port* get_port_by_pin(t_logical_block_type_ptr type, int pin);
 
+
+/************************************ Access to intra-block resources ************************************/
+
+/* Access information related to pin classes */
+
+
+/** get information given class physical num **/
+std::tuple<const t_sub_tile*, int> get_sub_tile_from_class_physical_num(t_physical_tile_type_ptr physical_tile, int physical_class_num);
```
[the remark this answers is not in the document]
MohamedElgammal: If they are related, is it better to define a struct or class?

vpr/src/base/netlist.h
```diff
@@ -874,7 +874,7 @@ class Netlist {
      */
     PinId create_pin(const PortId port_id, BitIndex port_bit, const NetId net_id, const PinType pin_type, bool is_const = false);
 
-    /**
+    virtual /**
```
MohamedElgammal, line 877: ?

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -50,6 +51,17 @@ static void alloc_and_load_pb_graph(t_pb_graph_node* pb_graph_node,
                                     bool load_power_structures,
                                     int& pin_count_in_cluster);
 
+static void set_logical_block_all_pins_indices(t_logical_block_type* logical_block);
+
+static std::vector<const t_pb_graph_node*> get_all_logical_block_pb_graph_nodes(const t_pb_graph_node* pb_graph_node);
+
+static void add_logical_classes(t_logical_block_type* logical_block);
+
+void static add_port_logical_classes(t_logical_block_type* logical_block,
```
MohamedElgammal, line 60: static void?

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -50,6 +51,17 @@ static void alloc_and_load_pb_graph(t_pb_graph_node* pb_graph_node,
                                     bool load_power_structures,
                                     int& pin_count_in_cluster);
 
+static void set_logical_block_all_pins_indices(t_logical_block_type* logical_block);
```
MohamedElgammal, line 54: Add comments for the new functions

vpr/src/pack/pb_type_graph.cpp
MohamedElgammal: it is better to send inputs as const

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -334,6 +350,179 @@ static void alloc_and_load_pb_graph(t_pb_graph_node* pb_graph_node,
...
+    while(!pb_graph_node_to_set_pin_index_q.empty()) {
+        auto curr_pb_graph_node = pb_graph_node_to_set_pin_index_q.front();
+        pb_graph_node_to_set_pin_index_q.pop();
+        auto curr_pb_type = curr_pb_graph_node->pb_type;
+
+        for (int port_type = 0; port_type < 3; port_type++) {
+            int num_ports;
+            int* num_pins;
+            t_pb_graph_pin** pins;
+            if (port_type == 0) {
```
MohamedElgammal, line 378: comment what is port type 0 or use enum

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -334,6 +350,179 @@ static void alloc_and_load_pb_graph(t_pb_graph_node* pb_graph_node,
...
+                    auto child_pb_graph_node = &(curr_pb_graph_node->child_pb_graph_nodes[mode_idx][pb_type_idx][pb_idx]);
+                    pb_graph_node_to_set_pin_index_q.push(child_pb_graph_node);
+                }
+            }
+        }
+    }
+
+}
+
+static std::vector<const t_pb_graph_node*> get_all_logical_block_pb_graph_nodes(const t_pb_graph_node* pb_graph_node) {
```
MohamedElgammal, line 420: Add comments on the function implementations

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -334,6 +350,179 @@ static void alloc_and_load_pb_graph(t_pb_graph_node* pb_graph_node,
...
+
+    logical_block->pb_pin_class_map = std::unordered_map<const t_pb_graph_pin*, int>();
+
+    auto pb_graph_nodes = get_all_logical_block_pb_graph_nodes(logical_block->pb_graph_head);
+    for(auto pb_graph_node : pb_graph_nodes) {
+        for(int port_type = 0; port_type < 3; port_type++) {
+            int num_ports;
+            int* num_pins;
+            t_pb_graph_pin** pb_graph_pins;
+            if(port_type == 0) {
```
MohamedElgammal, line 452: use IN_PORT , OUT_PORT instead of 0,1,2

vpr/src/route/route_common.cpp
```diff
@@ -282,8 +282,23 @@ bool try_route(int width_fac,
                     det_routing_arch,
                     segment_inf,
                     router_opts,
-                    directs, num_directs,
-                    &warning_count);
+                    directs,
+                    num_directs,
+                    &warning_count,
+                    router_opts.flat_routing);
+//    if(router_opts.flat_routing) {
```
MohamedElgammal, line 289: if not used, delete it

vpr/src/route/rr_graph.cpp
```diff
@@ -220,6 +243,62 @@ static void build_rr_sinks_sources(RRGraphBuilder& rr_graph_builder,
                                    const int delayless_switch,
                                    const DeviceGrid& grid);
 
+static void build_internal_rr_sinks_sources_flat(RRGraphBuilder& rr_graph_builder,
```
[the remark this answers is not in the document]
MohamedElgammal: Order the function parameters as input, state, output or vice versa but be consistent

vpr/src/route/rr_graph.cpp
```diff
@@ -1403,6 +1588,27 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
     return update_chan_width;
 }
 
+//static void alloc_and_load_intra_cluster_rr_graph(RRGraphBuilder& rr_graph_builder,
```
MohamedElgammal, line 1591: delete if unused

vpr/src/route/rr_graph.cpp
```diff
@@ -1488,6 +1694,109 @@ void free_rr_graph() {
     invalidate_router_lookahead_cache();
 }
 
+static void add_internal_rr_class(const t_class* class_inf,
```
MohamedElgammal, line 1697: Add comments on the function implementations

vpr/src/route/rr_graph.h
```diff
@@ -37,12 +38,28 @@ void create_rr_graph(const t_graph_type graph_type,
                      const t_router_opts& router_opts,
                      const t_direct_inf* directs,
                      const int num_directs,
-                     int* Warnings);
+                     int* Warnings,
+                     bool is_flat = false);
+
+//void add_intra_cluster_rr_graph(RRGraphBuilder& rr_graph_builder,
```
MohamedElgammal, line 44: Delete if unused

vpr/src/route/rr_graph2.cpp
```diff
@@ -36,7 +36,41 @@ static void load_chan_rr_indices(const int max_chan_width,
 
 static void load_block_rr_indices(RRGraphBuilder& rr_graph_builder,
                                   const DeviceGrid& grid,
-                                  int* index);
+                                  int* index,
+                                  bool is_flat);
+
+static void add_tile_pins_spatial_lookup(RRGraphBuilder& rr_graph_builder,
```
MohamedElgammal, line 42: Add commnets

MohamedElgammal (changes requested):
This PR is very useful and a great step toward flat routing. The overall structure is good. However, the new functions' definitions and implementations need comments.

vpr/src/pack/pb_type_graph.cpp
amin1377: There is no CLOCK_PORT Type. There are also other functions that follow the same convention. If you think it is useful, I can make a separate struct to specify the logical port type.

libs/libarchfpga/src/physical_types.h
amin1377: I changed it to pin_logical_num_to_pb_pin_mapping. If you think of any other name please let me know

libs/libarchfpga/src/physical_types.h
amin1377: Changed to pb_pin_to_class_logical_num_mapping

libs/libarchfpga/src/physical_types.h
amin1377: I wanted it to be consistent with the member (class_info) defined t_physical_tile_type. If you think I don't need to consider that, I think the name you suggested is good.

libs/libarchfpga/src/physical_types_util.h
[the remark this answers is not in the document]
amin1377: I agree that if the number of variables exceeds 2, it could be hard to work with it. I changed the functions which return the tuple object with more than 2 variables. For this particular function, I don't think we really need to change it. However, if you still think it should be changed please let me know.

vaughnbetz:
@amin1377 : some conflicts and failures to resolve now.

amin1377:
> @amin1377 : some conflicts and failures to resolve now.

I tried a lot of things to figure out what is causing the memory increase when flat_routing is **not** enabled. The only change I made in the data structure was changing the "ptc" variable type from "short" to "int." I changed the type back to "short", and it only reduced memory usage by about 0.1%. My best guess is that I have changed some data layout or the newly added data types are causing the increase in memory usage.

Is it possible to increase the referenced memory usage by about 4%? Or do you prefer not to do that?

@vaughnbetz

vaughnbetz:
A 4% memory increase isn't a deal breaker, but it's good to try to find it if it is unexpected. Going from short->int will have some impact, so might be that.  If that's necessary it's OK.

amin1377:
> A 4% memory increase isn't a deal breaker, but it's good to try to find it if it is unexpected. Going from short->int will have some impact, so might be that. If that's necessary it's OK.

I changed "int" back to "short," but it slightly reduced the memory usage (at most 1%). It seems the increase in memory usage is for another reason. Also, since every pin in each mode and sub-block has a unique ptc number, we have to change "short" to "int."

vaughnbetz:
@amin1377 : can you attach the latest QoR results (earlier ones are out of date te?  Also, I believe the 4% memory footprint increase is now gone based on what you mentioned earlier today -- can you confirm that?

amin1377:
> @amin1377 : can you attach the latest QoR results (earlier ones are out of date te? Also, I believe the 4% memory footprint increase is now gone based on what you mentioned earlier today -- can you confirm that?

Sorry for the late response. I was waiting to get titan_quick_qor results. [Here](https://utoronto-my.sharepoint.com/:x:/g/personal/amin_mohaghegh_mail_utoronto_ca/EbOkMr10isZHnrofkNkJQsgBXfjeNPXedTVewit1CJvdHg?e=gVy5Q8) are the QoR results of titan_other and tian_quick_qor.

The memory increase is also gone as you said because none of the data structures related to internal routing resources will not be initialized if flat_routing is not enabled.

Moreover, as is shown in the spreadsheet, the run time has increased. The reason is that when I was running this benchmark Wintermute Server was under heavy load.

vaughnbetz:
Looks good, I'll merge.  But please add big picture comments in a follow-up PR.

PR #2053, opened 2022-06-07
Example of Xilinx architecture description

WhiteNinjaZ:
#### Description

This PR adds an example of an approximate capture of the Xilinx architecture to VTR. 

#### Motivation and Context
Several recent changes to VTR remain to be tested in the reg_tests. This PR attempts to remedy this as well as to provide an example of a simplified Xilinx architecture. The plan is to continue adding descriptions in this file that test the functionality of new features (i.e. diagonal wires, partial crossbars, different architecture width/height, etc.) For now we only provide the description for a simplified CLB that matches the Xilinx arch as well as correct wire distributions/lengths in the horizontal/vertical directions (a feature newly added by PR #1883). Care has been taken to only include functionality in this PR that is fully supported by the current flow. In cases where functionality is currently lacking, comments are present. 

#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #2067, opened 2022-06-15
Addition of switch_override feature to custom SB

### LINKED ISSUE #2043: Unable to Describe Diagonal Wires with Current VPR architecture description
#### Current Behaviour
Several current FPGA's utilize diagonal routing in their architectures (specifically Xilinx parts). Currently, VPR has no way to implement this functionality. The proposal has been made that diagonal wires could be implemented by using a combination of shorted segments and custom switch blocks. However, there is a major issue with this: 
The issue is that switchpoints in vtr are currently referenced as 0->1->2->...->0 (i.e. both ends have a 0 index). Custom SB allow for unique specification of interconnect for each switch-point, but because both ends are specified with a zero index there is no way to uniquely specify an interconnect for the two ends. To demonstrate why this is an issue, lets assume we want to create a diagonal wire that goes up/down 4 and over 2 (i.e. vertical length 4 wires and horizontal length 2 wires). The desired functionality would be to have each end of vertical L4 wires connect to each begining of horizontal L2 wires via a short and for the beginning of the L4 wire and the end of the L2 wire to have "normal" interconnect (i.e. able to connect to other wire types), as shown in the drawing bellow:
image0
However, with the current iteration of VTR it is impossible to differentiate between the start and end of each wire segment in the architecture description since both the start and end are only accessible via their switch-points at the 0th index. The most you can do with the current architecture description is to short multiple diagonal segments together.

 

#### Possible Solution
This problem could be remedied by just changing how wire segments are represented (i.e. for a length _y_ wire the start switchpoint has index 0 and the end switch point has index _y_), however, the assumption that both wire ends have index 0 is pretty deeply ingrained in the code used to build switch blocks and parse wire segments as well as in previously created architecture descriptions. The use case is also unique enough that I wonder if it would just be better to create a new diagonal segment type. I am currently going through the code base to determine the best way to implement diagonal wires and will post here once I have a better understanding of a good approach. The main purpose for this issue is to provide a place to discuss ideas for how diagonal wires should be implemented and how they should be represented in the xml if we go that route.  

__Update:__ after thinking about diagonal wires for a while and working through some proofs using set theory, it appears that there is actually a way to build a rough approximation for diagonal wires using only custom SBs and a very small change to the code base (I believe the issue stated above was an error on my part and also due to me trying to implement diagonal wires in a bi-directional architecture). In the case of unidirectional architectures at least, all that is needed is to specify that the wires being driven by the vertical component of a diagonal segment have a arch_wire_switch that is of type "short" (i.e. in a L4->L2 diagonal wire, the L2 wire wire switch is overridden to be a short). However, this technique for specifying a diagonal wire in all four Intercardinal directions (NE,NW,SE,SW) requires a high level of verbosity. At least eight segments are needed and four custom switch blocks to specify a single diagonal wire. Additionally, there are certain aspects of diagonal wires that make their routing and interconnection unique and difficult or impossible to describe in the current architecture description. In light of this, I still think it would be a good idea to create a special way to describe diagonal wires in the architecture.

WhiteNinjaZ:
#### Description

This PR allows a user to specify a switch_override within the wireconn of a custom SB. The given switch will be used to override the wire_switch of the wire in the `to` set of the switch block. This PR will allow for more complicated wire shapes such as L and T shapes and provides the functionality needed to address issue #2043.  

#### Motivation and Context
Allows for more complicated wire types in architecture description. Will close #2043 by allowing for diagonal interconnect as well as more complicated wire structures.

Many tests have been run on a variety of custom architectures to ensure that the functionality of switch_override works properly. Tests have also been conducted to ensure that this new functionality does not interfere with older architectures. 

I plan to update the Xilinx arch in #2053 to test diagonal wires after we have gotten different x/y channel distributions figured out with custom SB's. 

#### Types of changes

#### Checklist:


WhiteNinjaZ:
@vaughnbetz ready for review.

vaughnbetz:
1. Can you confirm that the rr-graph created (when this feature is not used) is identical to what it used to be, and the QoR is also unchanged?  I'd expect that to be the case, but it's good to confirm on the stratixiv arch used for the titan run.  Easiest way to do this: write out the rr_graph.xml or rr_graph.echo before and after this change on that arch (which uses custom switch blocks) and confirm they are the same.  This is a stronger test than the CI tests, which only check we're in a reasonable range for various metrics.
2. Should add a test for the new feature (architecture file that uses it and at least one small design that runs on it).  That could be done in a later PR if you prefer.  If it's fast enough, you could add the test to vtr_reg_strong; otherwise add it to one of the nightly ones.

WhiteNinjaZ:
@vaughnbetz @amin1377 I have added a test case for the diagonal wires. Also, here are the [results of the titan_quick_qor](https://docs.google.com/spreadsheets/d/1tdUTZpEYfipoqTWwdfcIY1i4NKBiEAa_3_TmRrVjsjM/edit?usp=sharing). I did do a comparison of the rr_graph using the stratixiv arch and the directrf_stratixiv_arch_timing.blif (one of the larger tests). I can confirm that the two are equivalent and that the change has not changed the rr_graph.

doc/src/arch/reference.rst
```diff
@@ -2401,6 +2401,12 @@ The full format is documented below.
 
         .. note:: See ``from_switchpoint_order`` for value descritpions.
 
+    :opt_param switch_override:
+
+        Specifies the name of a switch to be used to override the wire_switch of the segments in the ``to`` set. Can be used to create more complicated wire structures such as T and L shaped segments.
+        
+        **Default:** If no override is specified, the normal interconnect will be used.
```
vaughnbetz, line 2408: I'd change "the normal interconnect will be used" to "the usual wire_switch that drives the ``to`` wire" will be used.

doc/src/arch/reference.rst
```diff
@@ -2401,6 +2401,12 @@ The full format is documented below.
 
         .. note:: See ``from_switchpoint_order`` for value descritpions.
 
+    :opt_param switch_override:
+
+        Specifies the name of a switch to be used to override the wire_switch of the segments in the ``to`` set. Can be used to create more complicated wire structures such as T and L shaped segments.
```
vaughnbetz, line 2406: I'd change the "Can be used" sentence to: "Can be used to create switch patterns where different switches are used for different types of connections; by using a zero-delay and zero-resistance switch one can also create T and L shaped wire segments."

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -331,6 +341,24 @@ void read_sb_switchfuncs(pugi::xml_node Node, t_switchblock_inf* sb, const pugiu
     return;
 }
 
+static void parse_switch_override(const char* switch_override, t_wireconn_inf& wireconn, const t_arch_switch_inf* switches, int num_switches) {
+    // check to see if user did not specify a switch_override
+    if (switch_override == std::string("")) {
+        wireconn.switch_override_indx = DEFAULT_SWITCH; //Default
```
vaughnbetz, line 347: Say "sentinel value to use default driving switch for the receiving wire type"

libs/libarchfpga/src/physical_types.h
```diff
@@ -1688,6 +1688,7 @@ struct t_wireconn_inf {
     std::vector<t_wire_switchpoints> to_switchpoint_set;               //The set of segment/wirepoints representing the 'to' set (union of all t_wire_switchpoints in vector)
     SwitchPointOrder from_switchpoint_order = SwitchPointOrder::FIXED; //The desired from_switchpoint_set ordering
     SwitchPointOrder to_switchpoint_order = SwitchPointOrder::FIXED;   //The desired to_switchpoint_set ordering
+    int switch_override_indx = DEFAULT_SWITCH;                         // index in switch array of the switch used to override wire_switch of the 'to' set.
```
vaughnbetz, line 1691: Explain that DEFAULT_SWITCH is a sentinel value meaning you will use the usual driving switch from a wire for the receiving wire type.

vaughnbetz (approved):
Looks good.  I suggest a few commenting changes.  None affect functionality, so we can merge this and you can do the commenting changes in a separate PR if you prefer -- your call.

vaughnbetz:
Looks like there are some conflicts to resolve too, so may as well make the commenting changes will fixing them.

WhiteNinjaZ:
@vaughnbetz all merge conflicts and requests resolved. Ready to merge.

WhiteNinjaZ:
@vaughnbetz CI is green, no conflicts, and all suggested changes have been implemented. Ready for merge.

PR #2082, opened 2022-06-30
Noc traffic flows parser final

Srivat97:
- Added a parser to read in traffic flow information between NoC routers in a design
- Added a datastructure to store and retrieve the parsed in traffic flow information 

#### Description

libvpr
- Added a xml parser that could read in NoC traffic flow files (.flows format)
- Created a class called NocTrafficFlows that stored the parsed traffic flow information from a .flows file
- Added another command line option to provide the name of the NoC traffic flows file
- Modified the ClusteredNetlist class to include a function that could find a block name by matching a regex string
- Added sample benchmark designs to verify this new feature

#### Motivation and Context

During placement, to optimize the placement of the NoC routers, we need information about how routers are communicating with each other. Some example things we should know are which routers are communicating with each other, what is the transmitted data size, are there any constraints on how fast the data should arrive at the destination. Using the previous information, we can tell whether a given placement of routers in the NoC is optimal or not.  We can check whether any of the links between the two routers are overused, we can also determine whether the data will be received at the destination within the given constraints. The important note here is that additional information is needed about the NoC. This PR allows designers to provide this additional information to VPR.

- Unit tests were created to verify individual components
- A test circuit and traffic flows file was also used to verify the entire feature

#### Types of changes

#### Checklist:


vaughnbetz (commented):
Initial comment.

PR #2100, opened 2022-07-20
Remaining malloc to new

jmah76:
#### Description
Changed remaining malloc/calloc/reallocs to new.

Passes all CI tests. 

#### Types of changes

- [ X] Bug fix (change which fixes an issue)

#### Checklist:


vaughnbetz:
document build failure is infrastructure issue so can remove draft.
Are all free/malloc/calloc/realloc gone from related libraries like libarchfpga?
Can you test runtime with QoR run on titan or vtr benchmarks?
See https://docs.verilogtorouting.org/en/latest/README.developers/?highlight=QoR%20#collecting-qor-measurements for procedure.

vaughnbetz:
All looks good except placed_wirelength_est looks too good (predicting a 6% reduction in wirelength, which shouldn't happen). Is that one a typo?

vaughnbetz:
Looked at the results some more.  Everything looks good except placed_wirelength_est looks too good in both comparisons.  Was there a change in the master code?  Is there a bug introduced in that computation (but nothing else, which would mean it is a bug in the bounding box -> wirelength estimation correction factors probably)?  That code does have an array used for lookup and it might be 1-indexed (shifted pointer) so maybe it got messed up?  Messing up that code wouldn't affect much or anything else other than that wirelength estimate, so that's a possible root cause.

Filing an issue on the parsing qor not working directly with the downloaded CI artifacts is a good idea.  If you know how to fix the scripts go ahead and fix them; otherwise we can probably get Jeff G. or Seyed to fix.

vaughnbetz:
Documentation failure is an infrastructure issue; ignore it.

vpr/src/power/power_util.cpp
```diff
@@ -426,8 +421,9 @@ static void alloc_and_load_mux_graph_recursive(t_mux_node* node,
     node->starting_pin_idx = starting_pin_idx;
 
     if (level != 0) {
-        node->children = (t_mux_node*)vtr::calloc(node->num_inputs,
-                                                  sizeof(t_mux_node));
+        node->children = new t_mux_node[node->num_inputs];
+        for (int i = 0; i < node->num_inputs; i++)
```
vaughnbetz, line 425: This for loop is unnecessary, as the default constructor will automatically be called on each element of the new-allocated array.

jmah76:
In the artifacts of GitHub's tests, every placed_wirelength_est value is -1... (see below)
image
I've checked over multiple tests and multiple actions, and they're all -1. I think I need a second set of eyes on this to make sure I'm not going insane.

Parsing QoR results doesn't work with parse_results produced when running tests from the command line (./run_reg_test.py vtr_reg_XYZ), either. I've made an Issue (Issue #2107 ) & PR (PR #2108 ) to address this.

vpr/src/route/bucket.cpp
```diff
@@ -141,10 +142,15 @@ void Bucket::free_all_memory() {
...
-    heap_ = (BucketItem**)vtr::realloc((void*)(heap_),
-                                       heap_size_ * sizeof(BucketItem*));
-    std::fill(heap_ + old_size, heap_ + heap_size_, nullptr);
+    int i;
+
+    std::vector<BucketItem*> temp(heap_, heap_ + old_size);
+    delete[] heap_;
+    heap_ = new BucketItem*[heap_size_];
+    for (i = 0; i < (int)old_size; i++)
+        heap_[i] = temp[i];
```
vaughnbetz, line 151: Same comment: better to do this without the temp vector.  Just allocate heap_new, copy over from heap_, delete heap_, then set heap_ = heap_new.

vaughnbetz (commented):
Looks good overall.  A few unnecessary loops and a couple of unnecessary copies.
Also worth checking how extensively we use the hash.h functions; a follow up PR could get rid of it and use std::unordered_map.

vpr/src/route/binary_heap.cpp
```diff
@@ -143,12 +140,20 @@ void BinaryHeap::sift_up(size_t leaf, t_heap* const node) {
...
+//expands heap by "realloc"
 void BinaryHeap::expand_heap_if_full() {
     if (heap_tail_ > heap_size_) { /* Heap is full */
-        heap_size_ *= 2;
-        heap_ = (t_heap**)vtr::realloc((void*)(heap_ + 1),
-                                       heap_size_ * sizeof(t_heap*));
-        heap_--; /* heap goes from [1..heap_size] */
+
+        // copy elements from heap [1 ... heap_size_] to temp [0 ... heap_size_ -1]
+        std::vector<t_heap*> temp(heap_ + 1, heap_ + 1 + heap_size_);
```
vaughnbetz, line 148: Better to do this without the temp vector. Just allocate heap_new, copy over from heap_, delete heap_, then set heap_ = heap_new. Right now there is an extra copy in a routine that is called in some cpu-intensive router code.

Bigger change (optional): could change the heap to a vector<t_heap*> and mess around with the indexing (need to keep convert between 1-based and 0-based indexing). Don't have to do that if you don't want to though, and if you do that, it should be in a separate PR.

vaughnbetz: (This comment got lost in the original review somehow)

vaughnbetz:
Not sure about the -1 place_wirelength_est in the github artifacts; maybe not parsed out in the CI tests?  The regression and CI scripts and conf files are fairly involved and have the ability to specify what gets parsed out.

jmah76:
- for future PR: vector size vs heap_size_

vpr/src/route/binary_heap.cpp
jmah76: Changed heap_ to a vector w/ an empty element in [0].

jmah76:
It turns out the QoR results I included in my wiki were wrong-- I editted the files manually, and messed up so that I shifted the results over by 1 column. My QoR_compare.py script edit is actually correct, and placed_wirelength_est is -1. Manually converting the .txt files to a spreadsheet shows placed_wirelength_est is -1 as well. I don't know what to do, so I'll send an email with relavent files.

vpr/src/power/power.cpp
```diff
@@ -763,7 +763,7 @@ static void power_usage_clock_single(t_power_usage* power_usage,
 /* Frees a multiplexer graph */
 static void dealloc_mux_graph(t_mux_node* node) {
     dealloc_mux_graph_rec(node);
-    free(node);
+    delete (node);
```
hzeller, line 766: Don't parenthesize parameters for `delete`.

vpr/src/util/hash.cpp
```diff
@@ -83,7 +84,7 @@ t_hash* insert_in_hash_table(t_hash** hash_table, const char* name, int next_fre
     h_ptr = hash_table[i];
 
     while (h_ptr != nullptr) {
-        if (strcmp(h_ptr->name, name) == 0) {
+        if (strcmp(h_ptr->name.c_str(), name) == 0) {
```
hzeller, line 87: If name is nw a std::string, then this comparison can be made more readable
```
  if (h_ptr->name == name) { ...
```

hzeller (commented):
Mostly some stylistic suggestions such as not having parenthesis in `delete` parameters.
... but do have parenthesis when calling a constructor `new foo()`.

vaughnbetz:
Thanks @hzeller .  Good suggestions.
@jmah76 : if placed_wirelength_est is -1 in the master and your code, then it is fine.  If it is -1 in your code but not master, you should rebase this PR onto (or merge in) the latest master so you get the updated placed_wirelength_est code.  If everything else matches there is almost certainly no issue in any case though.

vaughnbetz:
If you update your spreadsheets, please attach them here so we have a record of them (better than emailing them).

vaughnbetz:
Getting close: please confirm all review suggestions addressed (and push latest code), resolve conflicts, and upload the latest QoR spreadsheet and summarize. Then we can merge.

vaughnbetz:
Looks like there are a few errors in the QoR tests -- I think results were not generated for a few runs, like:
00:13:31 | Warning: task includes result for arch.timing.xml/picosoc_basys3_full_100.eblif missing in golden results
00:13:31 | Warning: task includes result for arch.timing.xml/picosoc_basys3_full_50.eblif missing in golden results
00:13:31 | Warning: task includes result for arch.timing.xml/linux_arty.eblif missing in golden results
00:13:31 | Warning: task includes result for arch.timing.xml/minilitex_arty.eblif missing in golden results
00:13:31 | Warning: task includes result for arch.timing.xml/minilitex_ddr_arty.eblif missing in golden results
00:13:31 | Warning: task includes result for arch.timing.xml/minilitex_ddr_eth_arty.eblif missing in golden results

vaughnbetz:
@jmah76 : can you take a look?  Not sure why warnings passed on this one, and not on Sebastian's PR (maybe he needs to update the branch)?

vaughnbetz:
6 failures (missing QoR data?) in nightly_test1.  One slightly higher vpr_max_mem on a small design in nightly_test2:
stratixiv_arch.timing.xml/murax_stratixiv_arch_timing.blif/common max_vpr_mem relative value 1.202197661475418 outside of range [0.8,1.2], above absolute threshold 102400.0 and not equal to golden value: 812136.0

The nightly test 2 failure is spurious; you can just update the golden results to have the new memory footprint number (it's not a big change -- just slightly more than the allowed 20%).  Or you can update the test to allow a [0.8, 1.25] range.

jmah76:
It's not missing QoR data-- one of the ASSERTs in bucket.cpp is failing because of the shift from using a vector to temporarily store the heap_ array for the resize to changing the heap_ array to a vector. I don't know why this is happening.

vaughnbetz:
If you can't sort it you can revert that one (just go back).  If you look at it closely hopefully you can figure out the bug, but if not, revert.

The bucket heap is only used for the Artix7 (Symbiflow) compiles, so it's not really the default flow.

jmah76:
I reverted only bucket.cpp back, and vtr_reg_nightly 1 seems to be working now. If it's okay, to fix vtr_neg_nightly 2, I'll do what you suggested here:
> The nightly test 2 failure is spurious; you can just update the golden results to have the new memory footprint number (it's not a big change -- just slightly more than the allowed 20%).

vaughnbetz:
Yup, that's fine.

vaughnbetz:
CI passed.  @jmah76 : can you attach the QoR comparison data so we see all is well (should be, but I'd like to check). There was some earlier data attached but I believe it had parse errors and wasn't correct.

vaughnbetz:
QoR generally looks good (unchanged critical paths, wirelength etc. on the vtr benchmarks, and roughly tied cpu time).  Memory footprint is unchanged on the larger designs, but goes up by 1.6x or even 2.2x on the smaller designs; suspect some small data structures is consuming more memory due to being turned into a vector but it has no impact on larger designs since not dominant / at peak.

To be careful @jmah76 will attach the vtr QoR spreadsheet and run and attach the Titan one.

jmah76:
@vaughnbetz Full Titan QoR results are on my wiki. Nothing looks concerning...
image

jmah76:
Here's the vtr QoR chain comparison:
image
Full on my wiki.

vaughnbetz:
Thanks.  Looks good, so merging.

PR #2102, opened 2022-07-21
Search Update + NoC GUI Update

SebastianLievano:
**Search **
Updated search functionality to include Block and Net Name suggestions/autocomplete, toggleable by hitting the "Enter" or "Return" key.  
Additionally, switched search to include Cluster and Atom netlist for Blocks, and the Atom netlist for nets (Cluster netlist is a subset of atom). The autosuggest algorithm/set up is meant to only run once when you hit enter, so you may need to rehit the key to reactivate it. Upon hitting enter, the program will search through all of the names and display all of the names that **contain** the given substring. This is obviously a large operation, and will take a bit of time. 

Furthermore, searching a subblock now highlights the block if it is visible at the current zoom lvl, and if not, highlights the containing cluster block. Additionally, a new subroutine which highlights an Atom Block from its id has been added, could be useful. 

**NoC Graphics Additions**
Added an NoC graphics combo box in the Block Settings dropdown. This will only display if the user uses the "--noc on" option on startup. 

#### Motivation and Context

The original search function had no form of automatic completion, required users to enter the full names of the block or net they were searching for, with no room for error, and was also fully based on the Clustered netlist, and thus could not search or highlight internals. It was pretty difficult to use as a result, as you'd need to know the full name of the thing you were searching for, and the names are often synthesized by VTR so you probably would not know that.

The new functionality will automatically parse through all names to find those that contain the given string. This allows users to search for a specific string even if they don't know its full name, and they can enable auto-suggestion only when they need it by pressing enter. 

All new GUI features have been tested and ensured to perform as expected. Search functionality has also been tested a lot on different architectures of varying sizes. 

#### Types of changes

### DIFF
--- a/vpr/src/draw/search_bar.cpp
+++ b/vpr/src/draw/search_bar.cpp
@@ -1,3 +1,18 @@
+/**
+ * @file search_bar.cpp
+ * @author Sebastian Lievano
+ * @brief Contains search/auto-complete related functions
+ * @version 0.1
+ * @date 2022-07-20
+ * 
+ * This file essentially follows the whole search process, from searching, finding the match,
+ * and finally highlighting the searched for item. Additionally, auto-complete related stuff is found
+ * here.
+ * 
+ * @copyright Copyright (c) 2022
+ * 
+ */
+
 #ifndef NO_GRAPHICS
 #    include <cstdio>
 #    include <sstream>
@@ -94,20 +109,25 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
     }
 
     else if (search_type == "Block Name") {
+        /* If the block exists in atom netlist, proceeding with highlighting process. 
+         * if highlight atom block fn returns false, that means that the block can't be highlighted
+         * We've already confirmed the block exists in the netlist, so that means that at this zoom lvl,
+         * the subblock is not shown. Therefore highlight the clb mapping.
+         * 
+         * If the block does not exist in the atom netlist, we will check the CLB netlist to see if
+         * they searched for a cluster block*/
         std::string block_name = "";
         ss >> block_name;
 
-        AtomBlockId atom_blk_id = AtomBlockId::INVALID();
-        atom_blk_id = atom_ctx.nlist.find_block(block_name);
-        //If block is found, the CLB containing it is highlighted
+        AtomBlockId atom_blk_id = atom_ctx.nlist.find_block(block_name);
         if (atom_blk_id != AtomBlockId::INVALID()) {
             ClusterBlockId block_id = atom_ctx.lookup.atom_clb(atom_blk_id);
-            //Highlighting atom block. IF function returns false, highlighting clb that contains
             if (!highlight_atom_block(atom_blk_id, block_id, app)) {
                 highlight_cluster_block(block_id);
             }
             return;
         }
+
         //Continues if atom block not found (Checking if user searched a clb)
         ClusterBlockId block_id = ClusterBlockId::INVALID();
         block_id = cluster_ctx.clb_nlist.find_block(block_name);
@@ -134,10 +154,11 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
     }
 
     else if (search_type == "Net Name") {
+        //in this case, all nets (clb and non-clb) are contained in the atom netlist
+        //So we only need to search this one
         std::string net_name = "";
         ss >> net_name;
-        AtomNetId atom_net_id = AtomNetId::INVALID();
-        atom_net_id = atom_ctx.nlist.find_net(net_name);
+        AtomNetId atom_net_id = atom_ctx.nlist.find_net(net_name);
 
         if (atom_net_id == AtomNetId::INVALID()) {
             warning_dialog_box("Invalid Net Name");
@@ -314,7 +335,7 @@ bool highlight_atom_block(AtomBlockId atom_blk, ClusterBlockId cl_blk, ezgl::app
  * 
  * @param name name of block being searched for
  * @param pb current node to be examined
- * @return t_pb* t_pb ptr of block w. name "name"
+ * @return t_pb* t_pb ptr of block w. name "name". Returns nullptr if nothing found
  */
 t_pb* find_atom_block_in_pb(std::string name, t_pb* pb) {
     //Checking if block is one being searched for
@@ -350,6 +371,7 @@ void highlight_nets(ClusterNetId net_id) {
 
     t_draw_state* draw_state = get_draw_state_vars();
 
+    //If routing does not exist return
     if (int(route_ctx.trace.size()) == 0) return;
 
     for (tptr = route_ctx.trace[net_id].head; tptr != nullptr; tptr = tptr->next) {
@@ -422,8 +444,8 @@ void search_type_changed(GtkComboBox* self, ezgl::application* app) {
 }
 
 /**
- * @brief A new matching function used for the wildcard search options
- * Will be slower
+ * @brief A non-default matching function. As opposed to simply searching for a prefix(default),
+ * searches string for presence of a substring. Ignores cases.
  * 
  * @param completer the GtkEntryCompletion being used
  * @param key a normalized and case-folded key representing the text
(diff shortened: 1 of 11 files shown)


vaughnbetz:
1. Please add cpu time data on a couple of cases for posterity.
2. The build the docs error is an unrelated infrastructure issue so can ignore.

vaughnbetz:
Needs a make format too.
@duck2 : may need this PR as he is hitting net search issues in the old UI.

vaughnbetz:
Read the docs error was an infrastructure issue (now fixed). Can ignore.
Need to do a 'make format' though

vpr/src/draw/search_bar.cpp
```diff
@@ -257,45 +300,83 @@ void highlight_blocks(ClusterBlockId clb_index) {
...
+    if (pb->child_pbs == nullptr) return nullptr;
+
+    int num_child_types = pb->get_num_child_types();
+    //Iterating through all child types
+    for (int i = 0; i < num_child_types; ++i) {
+        if (pb->child_pbs[i] == nullptr) continue;
+        int num_children_of_type = pb->get_num_children_of_type(i);
+        //Iterating through all of pb's children of given type
+        for (int j = 0; j < num_children_of_type; ++j) {
+            t_pb* child_pb = &pb->child_pbs[i][j];
```
vaughnbetz, line 354: Slightly simpler to write pb.child_pbs[i][j]

vpr/src/draw/search_bar.cpp
```diff
@@ -257,45 +300,83 @@ void highlight_blocks(ClusterBlockId clb_index) {
...
 
-    if (block_id == ClusterBlockId::INVALID()) {
-        warning_dialog_box("Invalid Block Name");
-        return; //name not exist
-    }
+    //If routing does not exist return
+    if (int(route_ctx.trace.size()) == 0) return;
 
-    highlight_blocks(block_id); //found block
+    for (tptr = route_ctx.trace[net_id].head; tptr != nullptr; tptr = tptr->next) {
```
vaughnbetz, line 377: This loop is redundant; it sets the draw_state->net_color[net_id] many times (once per routing resource in its traceback) but it can just do it once.

vpr/src/draw/search_bar.cpp
```diff
@@ -339,34 +420,135 @@ void warning_dialog_box(const char* message) {
...
+    } else { //setting to null if option does not require auto-complete
+        gtk_entry_completion_set_model(completion, NULL);
         gtk_entry_set_completion(searchBar, nullptr);
     }
 }
 
 /**
- * @brief loads block names into gtk list store item used for completion
+ * @brief A non-default matching function. As opposed to simply searching for a prefix(default),
+ * searches string for presence of a substring. Ignores cases.
```
vaughnbetz, line 448: Ignores cases --> Case insensitive

vpr/src/draw/search_bar.cpp
```diff
@@ -339,34 +420,135 @@ void warning_dialog_box(const char* message) {
...
  * 
- * @param app ezgl application
+ * @param completer the GtkEntryCompletion being used
+ * @param key a normalized and case-folded key representing the text
+ * @param iter GtkTreeIter pointing at the current entry being compared
+ * @param user_data null
+ * @return gboolean
```
vaughnbetz, line 454: When is true vs. false returned?

vaughnbetz: My guess of return value comment:
Returns true if the (single) text entry pointed at by GtkTreeIter contains key as a case-insensitive substring; false otherwise.

vpr/src/draw/search_bar.cpp
```diff
@@ -339,34 +420,135 @@ void warning_dialog_box(const char* message) {
...
+    g_utf8_casefold(text, -1);
+    g_utf8_normalize(text, -1, G_NORMALIZE_DEFAULT);
+    std::string cppText(text);
+    //If substring not found, returning false;
+    return (cppText.find(key, 0) != std::string::npos);
+}
+
+/**
+ * @brief Creates a GdkEvent that simulates user pressing key "key"
+ *
```
vaughnbetz, line 475: Might be useful to say what this is currently used for.
Also, I think it'll be better if the function level Doxygen comments move to the .h file.

vpr/src/draw/search_bar.cpp
```diff
@@ -339,34 +420,135 @@ void warning_dialog_box(const char* message) {
...
+    new_event.key.group = 0;
+    return new_event;
+}
+
+/**
+ * @brief Turns on autocomplete
+ * 
+ * This function enables the auto-complete fuctionality for the search bar. 
+ * Normally, this is pretty simple, but the idea is to have auto-complete appear as soon
+ * as the user hits the "Enter" key. To accomplish this, a fake Gdk event is created
```
vaughnbetz, line 502: Good to explain why this design choice was made (speed).

vpr/src/draw/search_bar.cpp
```diff
@@ -339,34 +420,135 @@ void warning_dialog_box(const char* message) {
...
+}
+
+std::string get_search_type(ezgl::application* app) {
+    GObject* combo_box = (GObject*)app->get_object("SearchType");
+    gchar* type = gtk_combo_box_text_get_active_text(GTK_COMBO_BOX_TEXT(combo_box));
+    //Checking that a type is selected
+    if (type && type[0] == '\0') {
+        warning_dialog_box("Please select a search type");
+        app->refresh_drawing();
+        return "Failed lol";
```
vaughnbetz, line 548: Not sure this is the best failure code ... empty string instead?

vpr/src/draw/search_bar.h
```diff
@@ -13,19 +13,38 @@
...
+bool highlight_atom_block(AtomBlockId atom_blk, ClusterBlockId cl_blk, ezgl::application* app);
+
+//Turns on autocomplete/suggestions
+void enable_autocomplete(ezgl::application* app);
+
+//Simulates key press event 
+GdkEvent simulate_keypress(char key, GdkWindow* window);
+
+//Returns current search type
+std::string get_search_type(ezgl::application* app);
```
vaughnbetz, line 47: Can fail so comment that (empty string for no valid search type, or with current code "failed_lol")!

vpr/src/draw/ui_setup.cpp
```diff
@@ -80,7 +80,7 @@ void net_button_setup(ezgl::application* app) {
  * @brief sets up block related buttons, connects their signals
  *
```
vaughnbetz, line 81: Nice doxygen comments, but move to .h file.

vpr/src/draw/ui_setup.cpp
```diff
@@ -169,4 +191,43 @@ void hide_widget(std::string widgetName, ezgl::application* app) {
     gtk_widget_hide(widget);
 }
 
+/**
+ * @brief loads block names into gtk list store item used for completion
```
vaughnbetz, line 195: Should say "atom and cluster level block names"
(and put in .h file)

vpr/src/draw/ui_setup.cpp
```diff
@@ -169,4 +191,43 @@ void hide_widget(std::string widgetName, ezgl::application* app) {
...
+    }
+    for (AtomBlockId id : atom_ctx.nlist.blocks()) {
+        gtk_list_store_append(blockStorage, &iter);
+        gtk_list_store_set(blockStorage, &iter,
+                           0, (atom_ctx.nlist.block_name(id)).c_str(), -1);
+    }
+}
+
+/**
+ * @brief loads net names into gtk list store item used for completion
```
vaughnbetz, line 217: net names -> atom net names

vaughnbetz (changes requested):
Some commenting changes suggested and a few minor restructurings. Looks very good and clean overall!

vpr/src/draw/search_bar.cpp
SebastianLievano: Not sure what you mean here. If the gchar* is empty and we attempt to create a c++ std::string, the program will crash. Therefore, we need to check if it is empty before we construct the string. I added a check to ensure type is not nullptr.

SebastianLievano: Oh wait the failed_lol. True.

SebastianLievano: oh oops forgot to get rid of this. Set it to empty string and added checks where it is called

vaughnbetz:
One warning in vpr that should be fixed. 
/home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/vpr/src/draw/search_bar.cpp:372:14: error: unused variable ‘tptr’ [-Werror=unused-variable]
  372 |     t_trace* tptr;

vaughnbetz:
@jmah76 @SebastianLievano : this looks good to merge except the "error on warning" compiler check is failing for vpr.
One warning is in Sebtasian's new code (listed above, get rid of an unused tptr).
A bunch are warnings on code of the type 
delete (pin_criticality+1)
Those are likely cropping up due to the free->delete changes. We probably suppressed those warnings with free (you can #pragma off specific warnings in specific files) but haven't suppressed the delete versions. Jennifer, can you either suppress them or (even better!) just allocate one more entry in the appropriate array (e.g. pin_criticality) and get rid of the pointer adjustment (e.g. pin_criticality--) and the +1 in the delete call. Maybe you've already done that and we just need to do an update branch operation?

jmah76:
I _think_ the pin_criticality + 1 warnings were there before I started (not 100% sure, but I haven't touched route_timing.cpp, the file where the warnings are popping up). Those "errors" are suppressed, but I can try to look into this on a different branch.

vaughnbetz:
OK, if those errors are suppressed and the test passes once the tptr error is gone, I'm OK with it.  Getting rid of them would be even cleaner if you can though!  If you expand the array/vector to have a 0'th entry, please set that 0'th entry to invalid values (NULL for a pointer, -1 for an integer, etc.) so we are likely to crash if we accidentally use it.

jmah76:
I made the changes [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/rem_malloc_to_new). I was able to get rid of the warnings by adding an entry in the arrays that were throwing the warning and setting their first element to 0. Once the CI tests pass, I'll make a PR.

vaughnbetz:
Great, thanks Jennifer.

SebastianLievano:
Added final modifications to toggle completion off after first showing. However, some limitation either in EZGL or GTK (probably some bug with focuses) means some key presses do not register in the function, making the toggling a bit inconsistent. Not worth pursuing imho. Will merge once checks are done @vaughnbetz

vaughnbetz:
One failure, but looks like an unrelated infrastructure issue. Can you rerun that test and see if it passes now (instructions are in the developer guide in the web documentation)?

Issue #2107, opened 2022-07-25
Script to Compare QoR Doesn't Work


jmah76:
./vtr_flow/scripts/qor_compare.py doesn't work on parse_results files from GitHub artifacts without editing the parse_results file. 
./vtr_flow/scripts/qor_compare.py requires QoR metrics in .txt files to be seperated by a tab ONLY (i.e. \t), but parse_results files output are formatted such that QoR results are seperated by a tab and two spaces (i.e. \t )

amin1377:
The script is still not working for me. I was wondering whether the issue is solved for you.
@jmah76

amin1377:
I just reverted the changes made in [PR#2108](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2108), and it works now!

vaughnbetz:
@jmah76 : seems the fix in PR #2108 is not robust and needs to be reverted to changed?

jmah76:
\s* greedy

jmah76:
@amin1377 Bumping this ^

amin1377:
@jmah76: It is working (at least for me.) Thank!
@vaughnbetz: I think it is safe to merge it.

jmah76:
Great, thank you so much! See PR #2122 .

amin1377:
> Great, thank you so much! The relavent PR is #2122 .

Yeah, I also meant PR #2122

PR #2108, opened 2022-07-25
Changed sep to be '\t  ' in QoR comparison script

jmah76:
Fix to Issue #2107. Changed the seperator between QoR metrics in parse_results.txt files to be '\t  ' instead of '\t' so users don't have to change the parse_results.txt files in order to run the comparison script, QoR_compare.py.


vaughnbetz:
Thanks!

Issue #2111, opened 2022-08-01
Mistakes in VPR Placement Constraints docs


TaoBi22:
#### Expected Behaviour
VPR should succeed when provided with the example placement constraints in the 'VPR Placement Constraints' documentation.

#### Current Behaviour
VPR throws the following: `Error parsing start element tag (line: 9 col: 10)`. This is due to a missing closing bracket - adding this bracket still results in `Start-end tags mismatch (line: 9 col: 13)` being thrown as the self-closing tags are missing backslashes.

Also worth noting that the note under constraints file format has an unintended em-dash before the command line argument and colons around the word option, where it should be formatted as an option - I think this is just due to an erroneous space.

#### Possible Solution
Just requires the addition of an angle bracket and some backslashes for the XML - I've updated this locally already so I'll submit it in a PR. 
I believe the incorrect section in the note should read ``:vpr:option:`--read_vpr_constraints` ``  in the source, as in other doc files - I'll add this in the PR too.

#### Steps to Reproduce
1. Copy the placement constraints provided (here)[https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/7989197469e9650e45a00e035f67139edc296831/doc/src/vpr/placement_constraints.rst] into a constraint file
2. Run VPR, providing that constraint file with the `--read_vpr_constraints`

vaughnbetz:
Fixed in the mentioned PR.

PR #2112, opened 2022-08-01
Fix mistakes in VPR Placement Constraints documentation

TaoBi22:
#### Description
This fixes a couple of mistakes in the VPR Placement Constraints documentation - the example constraints XML is now valid and accepted by VPR, and the `--read_vpr_constraints` option is formatted as as a VPR option, whereas before an erroneous space led to unintended characters appearing.

#### Related Issue
#2111 
#### Motivation and Context
The example XML in the docs previously caused VPR to fail - if you're used to XML the cause is pretty obvious but otherwise it might cause a bit of confusion, so hopefully this can avoid that. 

The docs successfully build, and the HTML page renders as it should.

#### Types of changes

#### Checklist:

### DIFF
--- a/doc/src/vpr/placement_constraints.rst
+++ b/doc/src/vpr/placement_constraints.rst
@@ -15,16 +15,16 @@ A Constraints File Example
 	<vpr_constraints tool_name="vpr">
      		<partition_list>
 	  	<partition name="Part0">
-	       		<add_atom name_pattern="li354">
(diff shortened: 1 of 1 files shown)


vaughnbetz:
Good catch -- thanks!
@jmah76 : FYI.

jmah76:
Just saw this haha. Thanks for the fix-- it's been giving me grief as well!

PR #2114, opened 2022-08-03
Addressing route_timing Offset

jmah76:
Referenced [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2102#issuecomment-1199941863) 

#### Description
Fixed warnings caused by arrays in route_timing.cpp's delete calls being offset by one because they were 1-based arrays. Fixed by adding an element to the arrays and ensuring the 0th element is invalid & will cause an error if used.


vpr/src/route/route_timing.cpp
```diff
@@ -908,11 +913,11 @@ void free_timing_driven_route_structs(float* pin_criticality, int* sink_order, t
     /* Frees all the structures needed only by the timing-driven router.        */
 
     // coverity[offset_free : Intentional]
-    delete[](pin_criticality + 1); /* Starts at index 1. */
+    delete[](pin_criticality); /* Starts at index 1. */
```
vaughnbetz, line 916: remove starts at index 1 comment

vaughnbetz:
Just one comment to remove -- looks good.

PR #2119, opened 2022-08-07
Build on M1

meme:
This allows vpr to be built on M1 MacBooks

#### Description

For this to work, you need `tbb@2020` from Homebrew.

I was able to build the software using this CMake command:

    TBBROOT=$(brew --prefix)/opt/tbb@2020/ cmake -DCMAKE_BUILD_TYPE=Release -DVPR_USE_EZGL=OFF -G Ninja -S . -B build

#### Types of changes

#### Checklist:


vaughnbetz:
Thanks for the contribution.
The changes look OK to me, and I've started CI.
@duck2 : there are some tbb build changes here to be compatible with apple.  I know you're also planning some changes to update our tbb version; could you take a look at these build changes and see if they interact with your update or if you have any thoughts.

vaughnbetz:
Not sure why vtr_reg_nightly_test1 was cancelled; appeared everything in it passed already anyway.

meme:
> Not sure why vtr_reg_nightly_test1 was cancelled; appeared everything in it passed already anyway.

It looks like it was cancelled because it hit the maximum timeout of 360 minutes. Not sure if that's related to my changes or not?

duck2:
The current `FindTBB.cmake` doesn't work with newer TBB versions anyway, so it should be OK to hack on it. The rest of the changes look good to me as well

duck2 (approved):

vaughnbetz:
Thanks.  Can't think of a reason why these changes would make us hit the 360 minute timeout (esp. after everything was done) so I'm assuming that was some transient infrastructure or load issue.  Merging.

PR #2129, opened 2022-08-14
Fixed Bug where VPR UI was not changing when switching mode

SebastianLievano:
There was a bug where as you proceeded through placement and routing of a circuit, the UI options/buttons available would not update to reflect the new state of vpr. This was caused by two things:

1. In the new initial setup functions, a guard if statement was placed that would return if a new window was not being created, which worked in the old nested format, but now that we are not nesting initial setup functions, we need to instead check if it is a new window, and only run the default_setup function if it is new. 
2. When hiding menu options, we were using gtk_widget_hide, but were not using gtk_widget_show when showing menu options.

These bugs have been fixed in this PR.

### DIFF
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -286,19 +286,16 @@ static void default_setup(ezgl::application* app) {
     search_setup(app);
 }
 
+// Initial Setup functions run default setup if they are a new window. Then, they will run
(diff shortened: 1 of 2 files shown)


SebastianLievano:
@kimiatkh @vaughnbetz Here is the fix for the bug Kimia brought up on Thursday. I've run through a few circuits and it seems to work. Please let me know if you find any other issues. Sorry about that.

PR #2134, opened 2022-08-16
Draw Partitions for Floorplanning

jmah76:
#### Description
Given a constraints file, VPR can now:
- Toggle on/off partition drawing
- Highlight partitions included in the constraints file
- Highlight primitives constrained in partitions
- Label highlighted partitions and constrained primitives

image
image

#### Tests
1. More partitions than number of colours available:
image
image
Number of possible colours for partitions was limited to 5 colours. Number of partitions in the constraint file was set to 9. Having moe partitions than colours available does not crash the program. Pictures above.

2. Checking different architectures:
Ran `vpr/vpr (insert architecture here) vtr_flow/benchmarks/blif/tseng.blif --route_chan_width 100 --read_vpr_constraints constraints.xml` then `vpr/vpr (insert architecture here) vtr_flow/benchmarks/blif/tseng.blif --route_chan_width 100 --read_vpr_constraints constraints.xml --analysis --disp on` with EArch.xml, k6_frac_N10_40nm.xml, and k6_frac_N10_mem32K_40nm.xml. All draw the partitions properly and their atoms inside their respective partitions properly.

3. Valgrind
Even on the master branch, running Valgrind with graphics results in [memory leaks](https://stackoverflow.com/questions/19389238/why-do-the-following-program-leaks-memory). Because of this, I'm not sure how to test if I added more memory leaks. Valgrind was run with the command `valgrind --leak-check=full --suppressions=./vpr/valgrind.supp --suppressions=/usr/share/glib-2.0/valgrind/glib.supp --suppressions /usr/share/gtk-3.0/valgrind/gtk.supp vpr/vpr (insert architecture here) vtr_flow/benchmarks/blif/tseng.blif --route_chan_width 100 --read_vpr_constraints constraints.xml --analysis --disp on`. Without graphics it produces no leaks.


jmah76:
I'm not crazy about how the partition names are displayed on the screen. If anyone else agrees, I might change the colour of the text to black or gray. The only downside of that would be that if there is overlap between two partitions, the colour coding isn't there anymore to indicate which partition belongs to which label.

jmah76:
Add comments & add doxygen comments. rst file

jmah76:
add how it was tested, check for different architechures, different #s of partitions & primitives

jmah76:
Paste output w/ Valgrind, confirm if runs w/o leaks

Issue #2145, opened 2022-08-27
RR Graph Direct Link Problem


amin1377:
[This function](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.cpp#L2888) adds the edges that correspond to direct links between blocks.  In this [line](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.cpp#L2965) it is assumed that adding the *sub_tile_offset* to *z* whould result in the sub_tile index which is used to access to the corresponding sub_tile in this [line](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph.cpp#L2966). However, this assumption is not true. Neither *sub_tile_offset* nor *z* have anything to do with the sub_tile index. They are related to the capacity number of the intended sub_tile.

@vaughnbetz

PR #2146, opened 2022-08-27
Fix rr_graph direct link

amin1377:
This PR is created to fix issue #2145
@vaughnbetz

### DIFF
--- a/vpr/src/route/rr_graph.cpp
+++ b/vpr/src/route/rr_graph.cpp
@@ -2962,12 +2962,20 @@ static int get_opin_direct_connections(RRGraphBuilder& rr_graph_builder,
                             }
                         }
 
-                        int target_sub_tile = z + directs[i].sub_tile_offset;
-                        if (relative_ipin >= target_type->sub_tiles[target_sub_tile].num_phy_pins) continue;
+                        int target_cap = z + directs[i].sub_tile_offset;
+                        const t_sub_tile* target_sub_tile = nullptr;
+                        for (const auto& sub_tile : target_type->sub_tiles) {
+                            if (sub_tile.capacity.is_in_range(target_cap)) {
+                                target_sub_tile = &sub_tile;
+                                break;
+                            }
+                        }
+                        VTR_ASSERT(target_sub_tile != nullptr);
+                        if (relative_ipin >= target_sub_tile->num_phy_pins) continue;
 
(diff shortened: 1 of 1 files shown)


vpr/src/route/rr_graph.cpp
```diff
@@ -2962,12 +2962,20 @@ static int get_opin_direct_connections(RRGraphBuilder& rr_graph_builder,
                             }
                         }
 
-                        int target_sub_tile = z + directs[i].sub_tile_offset;
-                        if (relative_ipin >= target_type->sub_tiles[target_sub_tile].num_phy_pins) continue;
+                        int target_cap = z + directs[i].sub_tile_offset;
```
vaughnbetz, line 2965: Can you comment what the logic is here?

vpr/src/route/rr_graph.cpp
```diff
@@ -2962,12 +2962,20 @@ static int get_opin_direct_connections(RRGraphBuilder& rr_graph_builder,
...
+                                break;
+                            }
+                        }
+                        VTR_ASSERT(target_sub_tile != nullptr);
+                        if (relative_ipin >= target_sub_tile->num_phy_pins) continue;
 
                         //If this block has capacity > 1 then the pins of z position > 0 are offset
                         //by the number of pins per capacity instance
-                        int ipin = get_physical_pin_from_capacity_location(target_type, relative_ipin, target_sub_tile);
+                        int ipin = get_physical_pin_from_capacity_location(target_type, relative_ipin, target_cap);
```
vaughnbetz, line 2978: Is the comment above still correct?  I'm not seeing the logic of how the pins wrap around (pins of z position > 0 offset by number of pins per capacity instance).

vaughnbetz (commented):
Nice job finding this!  Can you update the comments in the code to explain the logic though and make sure the existing comment is still correct?

vpr/src/route/rr_graph.cpp
amin1377: "relative_pin" shows the pin number in its capacity. So, to get the number of the pin at tile-level, "relative_ipin" should be added by the number of pins on prior sub_tiles and capacities.

vpr/src/route/rr_graph.cpp
amin1377: Done

PR #2158, opened 2022-09-10
router lookahead: fix criticality multiplication bug

amin1377:
In router_lookahead_map ([this line](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/a7cd8f99740d64acb92c470fd5b9069c1f0e04c5/vpr/src/route/router_lookahead_map.cpp#L329)), the expected delay is multiplied by (1-criticality). It seems backward to me. I think the delay should be multiplied by the criticality factor, and congestion should be multiplied by (1-criticality.)

@vaughnbetz


vaughnbetz:
Nightly test 1 results has a few QoR failures (8) but they look benign:
- one small circuit (mult113) has a wider channel width
- one small mcnc circuit has a smaller channel width
- another small circuit (tseng) has a wider channel width
Updating the QoR golden results for all these (or slightly widening the acceptable QoR range for them) would be OK.

You should do a QoR comparison spreadsheet on both Titan and the VTR benchmarks though to make sure QoR is OK.

vaughnbetz:
Updating the branch so CI is going to kick off again ....

amin1377:
> Nightly test 1 results has a few QoR failures (8) but they look benign:
> 
> * one small circuit (mult113) has a wider channel width
> * one small mcnc circuit has a smaller channel width
> * another small circuit (tseng) has a wider channel width
>   Updating the QoR golden results for all these (or slightly widening the acceptable QoR range for them) would be OK.
> 
> You should do a QoR comparison spreadsheet on both Titan and the VTR benchmarks though to make sure QoR is OK.

I did the QoR comparison for the Titan benchmark. It improved the CPD for titan_quick_qor by almost 1% and didn't have a significant effect on other metrics (less that 1%).
Once VTR tests are done, I will attach the spreadsheet which shows the QoR comparison for each circuit.

vaughnbetz:
Both strong and nightly failures are a few large QoR changes on small designs (some better, some worse): OK to just update the golden results for them.

vaughnbetz:
Can update the few numbers by hand, or use https://docs.verilogtorouting.org/en/latest/README.developers/?highlight=golden#generating-new-qor-golden-result

amin1377:
@vaughnbetz :  I think this PR is finally ready to land :)

PR #2163, opened 2022-09-23
Custom Switch block with different x and y channel widths

kimiatkh:
#### Description
This PR is an extension of #1883 that allows for modelling of unequal horizontal and vertical channel widths in unidirectional architectures with custom switch blocks. 

#### Motivation and Context
Most modern FPGA devices have different horizontal and vertical channel widths due to imbalance of routing congestion between x and y channels. This new feature is being added to allow for a more accurate capture of such routing architectures. 

For the first test, the rr_graph of the Stratix IV capture with equal x and y channel widths has been generated with the new changes introduced and the output is compared to the base rr_graph generated by the original code. The two code bases output the exact same rr_graph.

The code has been tested on titan benchmarks and slightly modified captures of Stratix IV device. The target architectures have y-x channel width ratio of 0.7, 0.9, and 1. The input channel width for each of these architectures has been set such that the average x-y channel width is equal to 300. Below is the geomean of wirelength and CPD across titan benchmarks. 

| y-x ratio | wirelength | critical path delay|
| ------------- | ------------- | ------------- |
| 0.7  | 3204560.54  | 17.63 |
| 0.9 | 3176448.33 | 17.41 |
|1 | 3175743.80  | 17.18 |

As can be seen there is a slight decline in QOR as the ratio gets smaller which is as expected.  

Eventually a new regression test has been added that runs four of the small titan designs on a modified Stratix IV capture where the y-x channel width ratio is set to 0.75. 
#### Types of changes


vpr/src/route/build_switchblocks.cpp
```diff
@@ -321,12 +297,6 @@ t_sb_connection_map* alloc_and_load_switchblock_permutations(const t_chan_detail
 
     /* get a single number for channel width. 
      * AA: Note that this needs be changed to support different horizontal and vertical channels. Future action item ... */
```
vaughnbetz: Should delete this comment. (out of date)

vtr_flow/misc/yosyslib/synthesis.tcl
```diff
@@ -37,6 +37,9 @@ if {$env(PARSER) == "surelog" } {
 # read the custom complex blocks in the architecture
 read_verilog -lib CCC
 
+# read the custom complex blocks in the architecture
```
vaughnbetz, line 40: Repeated lines -- seems like the bottom ones should be deleted.

vaughnbetz (commented):
A couple of things to change, then do make format, then push.

vaughnbetz:
Looks good -- thanks.

PR #2165, opened 2022-09-30
Adding Koios 2.0 benchmarks

aman26kbm:
#### Description
Koios 1.0 benchmark suite is being updated to Koios 2.0. This includes modifications to old benchmarks and addition of new benchmarks.

#### Types of changes

#### Checklist:

- [ x] I have added tests to cover my changes


aman26kbm:
As a part of checking in Koios 2.0 benchmarks, I am adding two regression tests (one in nightly and one in weekly) that run VTR with the new Yosys+ODIN flow. The benchmarks for these tests don't run with ODIN-only flow. The task configs I have added do enable the Yosys+ODIN flow, but I see an error in the CI that says Yosys was not built/installed.

Looks like we are not building with Yosys enabled in the CI scripts. Should we update the VTR build command for all CI runs to enable Yosys (CMAKE_PARAMS="-DODIN_USE_YOSYS=ON")? Or, may be it's better to build with Yosys for only the regression tests that need it. 

Also, can someone point me to documentation on how to update build commands for regression tests?

Tagging @vaughnbetz , @sdamghan , @ArashAhmadian

vaughnbetz:
Thanks @aman26kbm .  It's a coverage gap if we have no YOSYS + ODINII tests; maybe there are some small ones already?  In any case running koios with YOSYS+ODINII seems like a good idea.
I suspect having some CI tests run without (CMAKE_PARAMS="-DODIN_USE_YOSYS=ON") is a good idea to make sure the CMAKE variable still works, and those who don't want YOSYS don't have to build it.

sdamghan:
@aman26kbm - As the Yosys+Odin-II has an additional Yosys building process, which may not be required for all tests under the vtr_nightly regression tests, we added a separate regression test and a CI job named `vtr_reg_yosys_odin`, for testing this front-end (the same goes for Yosys standalone as well). So basically, users can take advantage of the Yosys+Odin-II front-end by compiling the VTR flow using the `CMAKE_PARAMS="-DODIN_USE_YOSYS=ON"` parameter. So, if you would like to have the Koios 2 task under the `vtr_reg_nightly_test4` regression test, you would require to add the above-mentioned CMake parameter to the GitHub action configuration for this test.

Adding "-DODIN_USE_YOSYS=ON" in the cmake attribute at line 32 :
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/47a1a58997348068ebe84f2a0be700695f86191d/.github/workflows/test.yml#L28-L38

Other than that, if you would add Koios 2.0 tests under the `vtr_reg_yosys_odin` regression test, everything should be fine and there should be no CI failure.

aman26kbm:
Thanks guys. I've added the extra CMAKE arg to vtr_reg_nightly_test4

aman26kbm:
I updated the build command to build with YOSYS enabled. With that I do see the build succeeding, but the test run doesn't complete. It says "The operation was canceled." I don't see any other error. 

I thought it was just temporary, so I re-ran the failing test (nightly, 4). But it is failing again with the same error.

@vaughnbetz , can you please point me to the right direction?

vaughnbetz:
Hmmm ... not sure what happened. It looks like vtr_nightly_test_4 just hung after building, and right when it started running tests (I don't see any tests that had output, and the cancelled is probably a timeout being hit).  I'm going to hit the update branch button which will sync up the source and relaunch the test, in case this is a transient issue.

If that doesn't fix it I think you'll need to run vtr_nightly_test_4 on your machine and watch the output to see which test hangs (might be the first, but harder to tell with CI as sometimes it has some buffering etc.).

aman26kbm:
Thanks, @vaughnbetz.
It failed again with the same message.
The failure happened at 6 hour 26 seconds. I believe that is indeed because of a timeout. 
I've added more tests to this task (nightly, 4). So I expect it will take longer. Not sure how much though. I will kick off a run on my local machine to see how long it takes.
Do we want any nightly regression runs to finish within 6 hours? Or can we extend it some? Also, where are the timeout settings mentioned? I can try to increase them temporarily to check how long this task now takes on the CI machine.

sdamghan:
@aman26kbm - The issue is actually the timeout. GitHub runners have [usage limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits). One of them is that the job execution should not exceed 6 hours. So, you have two options here:
1. Adding the Koios 2.0 to a pre-existing regression test, run by a single GitHub runner, in which the total execution time will be less than 6 hours. For instance, the `vtr_reg_yosys_odin` took less than two hours in the latest run. It could be a good option among regression tests as it already has an extra Yosys compilation process, giving you more time to run tests.
2. You can create a new regression test from scratch, specifically allocated to Koios 2.0 tests (something similar to `vtr_reg_nightly4` or `vtr_reg_yosys_odin`). In this case, you would need to create a new folder, including tasks and a `task_list.txt` file and lets name it `vtr_reg_nightly5`, under the _VTR_ROOT/vtr_flow/tasks/regression_tests/_ directory. Then, a new record similar to pre-existing tests should be added to the _test.yml_ file. (see below)

Adding `- {test: "vtr_reg_nightly5",    cores: "16", options: "",          cmake: "-DODIN_USE_YOSYS=ON -DYOSYS_SV_UHDM_PLUGIN=ON"}` at line 33. below
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/47a1a58997348068ebe84f2a0be700695f86191d/.github/workflows/test.yml#L28-L38

In this way, a new GitHub runner will be created, and allocated to this run. It will compile the VTR flow using the given CMake options and run `run_reg_test.py vtr_reg_nighlty5 -j 16` in the VTR root directory in the runner terminal.

vaughnbetz:
I'm not sure where the timeouts are set; they may be on the google cloud machines we use for long running regressions (and hence not directly available for us to edit). @mithro may know better.

Making a new CI run is probably best if these are long running jobs, as it will also avoid lengthening CI (each top-level regression test gets a different machine).

aman26kbm:
Hi @vaughnbetz , @sdamghan , can you please review the changes in this PR?

One thing that is pending is to checkin the golden results file for the weekly regression. I have kicked it off locally and I can get golden results file from there.

sdamghan:
I have looked at some infrastructural files, and overall the PR looks good to me. Thanks to everybody who contributes to this, especially you @aman26kbm for finishing this up.
I have a suggestion; we have gone through many issues/bugs/new features during the Koios 2.0 development that you already pointed out in the article. I believe if we can have all those issues, either fixed ones or the ones we are still waiting to be fixed, outlined somewhere like a readme file, it will be great. This way, developers can easily spot the issue and the corresponding benchmark file, and the fixing process will be much more straightforward. Although it should be specifically added to this PR, we can work on it in another PR as well.

aman26kbm:
> I have looked at some infrastructural files, and overall the PR looks good to me. Thanks to everybody who contributes to this, especially you @aman26kbm for finishing this up. I have a suggestion; we have gone through many issues/bugs/new features during the Koios 2.0 development that you already pointed out in the article. I believe if we can have all those issues, either fixed ones or the ones we are still waiting to be fixed, outlined somewhere like a readme file, it will be great. This way, developers can easily spot the issue and the corresponding benchmark file, and the fixing process will be much more straightforward. Although it should be specifically added to this PR, we can work on it in another PR as well.

Thanks for reviewing, @sdamghan . I'll add a list of github issues (open and closed) in the README file in the Koios benchmarks directory.

aman26kbm:
Merging this PR

PR #2166, opened 2022-10-10
migrate to oneTBB

duck2:
TBB is now called oneTBB and has a bunch of [breaking changes](https://oneapi-src.github.io/oneTBB/main/tbb_userguide/Migration_Guide.html). This change enables VTR to use it.

#### Types of changes

#### Checklist:


duck2:
Just realized none of the tests actually run with TBB 🤦 because _it may increase memory usage_, but that comment seems to be at least 2 years old. Let's see what happens when we enable it.

duck2:
I tested this setup on a custom 20.04 container and it works. What I didn't notice is that all CI tasks are actually running on Bionic even though their names contain Focal or Hirsute. Is this a bug or is it intended?

duck2:
I think all checks are OK now, but there is still an "expected" check. Looks like it's called "B: Building VtR" and isn't connected to any workflow, so I can't rerun it? (I force-pushed to the branch many times and that doesn't seem to be the fix.) @vaughnbetz

vaughnbetz:
Yes, I'm seeing that B:Building VtR never starting on other PRs too, so it is unrelated and not gating.  Let's discuss at the vtr meeting today and get an owner for that CI issue.

vaughnbetz:
Adding @kmurray since this affects parallel tatum / testing.
The code changes look good and safe.
From talking to @duck2, we also now have a CI test for building with TBB (the container image has TBB installed).
We don't appear to have a CI run test of parallel tatum + TBB unfortunately.
I think this is OK to merge, but we should get a CI test of tatum + TBB going soon.  
I think the path forward is for @duck2 to update CI to use Ubuntu 20.04 in CI and see if everything works.

vaughnbetz:
Merging.  To do list though:
- Check if we have a parallel test.  Can make vpr exit if --num_workers > 1.
- Update CI to Ubuntu 20.4 and install oneTBB as a dependency so this test then runs all the time.

PR #2175, opened 2022-10-17
Centroid placement

saaramahmoudi:
Implement another initial placement algorithm called centroid placement. It tries to place blocks in a smarter location compared to random placement based on their logical connections. 

#### Description

Initial placement tries to place each block close to its connected blocks (that are already placed) to produce better quality initial placement in comparison with random initial placement. Since the initial placement quality is better now, we do not need to set starting temperature for the SA algorithm to a high value to destroy our initial placement. Hence, we can cut the first part of the SA algorithm, which tries to find a better placement solution. 
Here are the changes that are made to implement what I described earlier:

1. Instead of sorting blocks in initial placement, we now use a heap data structure to place blocks with the most placed connection earlier. Since updating the heap every time that a block is placed increases place_time, we update the heap frequently based on how big our design is. 
2. The try_centroid_placement suggests a more convenient location for each block. If the location is occupied or does not match the block type, another function tries to find a nearby location to replace the suggestion. 
3. Starting temperature in place.cpp is set to a very low value to avoid destroying our smart initial placement.
4. The starting_t function is also changed, and set the initial temperature to the standard deviation of accepted moves divided by 64, which is an input to the SA algorithm.
5. The place_effort_scaling default value is also set to device_circuit instead of the circuit.

It is notable that if the centroid algorithm can not find a legal location for a block, the algorithm would use the existing random placement to place the block. 

The result for these three benchmarks is attached:
1. [koios Benchmark](https://docs.google.com/spreadsheets/d/1zwQwpHp0K4jU4fy7P6DwMioSSzqyO00fdbe6fO00xSg/edit?usp=sharing) , all circuits in this design are passed.
2. [titan_other Benchmark](https://docs.google.com/spreadsheets/d/1R31MYYnDkUrp0sn6mzb6pZVojdbWSWubn27dJ8Lbmcc/edit?usp=sharing), all circuits in this design are passed
3. [titan_quick_qor](https://docs.google.com/spreadsheets/d/1n_5mcrrWalVJA0xcI745VmlfekDYJE2GL6ONUVjCda4/edit?usp=sharing), all circuits except **"bitcoin_miner"** are passed. The result sheet exclude this design, the vpr output for the failed design is also attached later.

Failed design in titan_quick_qor:
- [centroid placement bitcoin miner vpr output](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/9803634/vpr.txt)
- [master branch bitcoin miner vpr output](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/9803642/vpr.txt)

#### Types of changes

@vaughnbetz


vaughnbetz:
I looked at nightly_test_1 and the QoR failures all look OK; mostly much faster minimum channel width route times for small "multless_constants" designs.  Also slower placement times for small ISPD designs in big devices, which is OK (natural effect of this change).

vaughnbetz:
Thanks.  QoR data shows some increase in placement time on the largest benchmark sets (Titan quick QoR and koios) but also a small quality increase.  Can you also try device_circuit scaling with a change to inner_num (to 0.4) to see if that gets us back to even or ahead on placement time, but hopefully maintains the small quality gain?

saaramahmoudi:
Sure, will run and attach results as soon as possible.

vpr/src/place/initial_placement.cpp
```diff
@@ -236,6 +257,191 @@ static bool is_block_placed(ClusterBlockId blk_id) {
...
+}
+
+static bool find_centroid_neighbor(t_pl_loc centroid_loc, t_logical_block_type_ptr block_type, int& cx_to, int& cy_to) {
+    const auto& compressed_block_grid = g_vpr_ctx.placement().compressed_block_grids[block_type->index];
+
+    //Determine centroid location in the compressed space of the current block
+    int cx_centroid = grid_to_compressed_approx(compressed_block_grid.compressed_to_grid_x, centroid_loc.x);
+    int cy_centroid = grid_to_compressed_approx(compressed_block_grid.compressed_to_grid_y, centroid_loc.y);
+
+    //Determine rlim in each direction
```
vaughnbetz, line 290: comment what rlim is here.

vpr/src/place/initial_placement.cpp
```diff
@@ -236,6 +257,191 @@ static bool is_block_placed(ClusterBlockId blk_id) {
...
+    int min_cy, max_cy;
+
+    min_cx = std::max(0, cx_centroid - rlim_x);
+    max_cx = std::min<int>(compressed_block_grid.compressed_to_grid_x.size() - 1, cx_centroid + rlim_x);
+
+    min_cy = std::max(0, cy_centroid - rlim_y);
+    max_cy = std::min<int>(compressed_block_grid.compressed_to_grid_y.size() - 1, cy_centroid + rlim_y);
+
+    delta_cx = max_cx - min_cx;
+    bool legal = false;
```
vaughnbetz, line 306: Don't need legal = false; routine below always sets it.  Better to delete this line.

vaughnbetz: Context on what this routine is doing, why cx_from and cy_from are -1 etc. would be helpful.

vpr/src/place/initial_placement.cpp
```diff
@@ -236,6 +257,191 @@ static bool is_block_placed(ClusterBlockId blk_id) {
...
+
+static std::vector<ClusterBlockId> find_centroid_loc(t_pl_macro pl_macro, t_pl_loc& centroid) {
+    auto& cluster_ctx = g_vpr_ctx.clustering();
+
+    int x, y;
+    float acc_weight = 0;
+    float acc_x = 0;
+    float acc_y = 0;
+
+    ClusterBlockId b_from = pl_macro.members.at(0).blk_index;
```
vaughnbetz, line 322: b_from --> head_blk (or head_block)

vpr/src/place/initial_placement.cpp
```diff
@@ -236,6 +257,191 @@ static bool is_block_placed(ClusterBlockId blk_id) {
...
+            ClusterPinId sink_pin = *cluster_ctx.clb_nlist.net_sinks(net_id).begin();
+            ClusterBlockId sink = cluster_ctx.clb_nlist.pin_block(sink_pin);
+            if (sink == source) {
+                continue;
+            }
+        }
+
+        //if the pin is driver iterate over all the sinks
+        if (cluster_ctx.clb_nlist.pin_type(pin_id) == PinType::DRIVER) {
+            //continue on ignored nets
```
vaughnbetz, line 342: Delete comment or go to a higher level (ignored nets are globally routed clocks etc; can skip them).

vpr/src/place/initial_placement.cpp
```diff
@@ -236,6 +257,191 @@ static bool is_block_placed(ClusterBlockId blk_id) {
...
+        return false;
+    }
+
+    //centroid suggestion was either occupied or does not match block type
+    //try to find a near location that meet these requirements
+    if (!is_loc_legal(centroid_loc, pr, block_type)) {
+        find_centroid_neighbor(centroid_loc, block_type, centroid_loc.x, centroid_loc.y);
+    }
+
+    //no neighbor were found that meet all out requirements, should be placed with random placement
```
vaughnbetz, line 406: out -> our

vpr/src/place/initial_placement.cpp
```diff
@@ -592,6 +802,7 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores() {
 
     //go through all blocks and store floorplan constraints score
```
vaughnbetz, line 803: Can update this comment (not just floorplan constraints) or move your initialization into its own loop.

vpr/src/place/initial_placement.cpp
```diff
@@ -611,50 +822,23 @@ static vtr::vector<ClusterBlockId, t_block_score> assign_block_scores() {
...
     auto& device_ctx = g_vpr_ctx.device();
-
+    auto blocks = cluster_ctx.clb_nlist.blocks();
     int number_of_unplaced_blks_in_curr_itr;
+
     //keep tracks of which block types can not be placed in each iteration
     std::unordered_set<int> unplaced_blk_type_in_curr_itr;
 
+    auto criteria = [&block_scores, &cluster_ctx](const ClusterBlockId& lhs, const ClusterBlockId& rhs) {
+        int lhs_score = 10 * block_scores[lhs].macro_size + block_scores[lhs].number_of_placed_connections + block_scores[lhs].tiles_outside_of_floorplan_constraints + SORT_WEIGHT_PER_FAILED_BLOCK * block_scores[lhs].failed_to_place_in_prev_attempts;
```
vaughnbetz, line 836: check magnitude of tiles_outside_of_floorplan_constraints.
Add a comment on relative weighting of the 4 factors here.

vpr/src/place/initial_placement.cpp
```diff
@@ -669,18 +853,37 @@ static void place_all_blocks(enum e_pad_loc_type pad_loc_type, const char* const
...
-
         //resize the vector to store unplaced block types empty locations
         blk_types_empty_locs_in_grid.resize(device_ctx.logical_block_types.size());
 
         number_of_unplaced_blks_in_curr_itr = 0;
 
-        for (auto blk_id : sorted_blocks) {
+        //calculate heap update frequency based on number of blocks in the design
+        int update_heap_freq = blocks.size() / 100;
+        if (update_heap_freq == 0) { //if update_heap_freq is zero, update heap after each block get placed
```
vaughnbetz, line 863: can just use max (blocks.size()/100, 1)

vpr/src/place/initial_placement.cpp
```diff
@@ -754,7 +957,8 @@ static void clear_block_type_grid_locs(std::unordered_set<int> unplaced_blk_type
 
 bool place_one_block(const ClusterBlockId& blk_id,
                      enum e_pad_loc_type pad_loc_type,
-                     std::vector<t_grid_empty_locs_block_type>* blk_types_empty_locs_in_grid) {
+                     std::vector<t_grid_empty_locs_block_type>* blk_types_empty_locs_in_grid,
```
vaughnbetz, line 960: Probably cleaner to pass by reference using & instead.

vaughnbetz: Especially because the lower level helper functions are passing by reference with &

vpr/src/place/initial_placement.h
```diff
@@ -3,6 +3,24 @@
...
 
+/**
+ * @brief Used to assign each block a score for how difficult it is to place. 
+ * The higher numbers indicate a block is expected to be more difficult to place.
+ * Hence, initial placement tries to place blocks with higher scores earlier.
+ */
+struct t_block_score {
+    int macro_size = 0; //How many members does the macro have, if the block is part of one - this value is zero if the block is not in a macro
+
+    //The number of tiles NOT covered by the block's floorplan constraints.
```
vaughnbetz, line 14: Is this normalized or not?

vaughnbetz: Should normalize if it isn't right now (could divide by number of tiles in device, could also multiple after that by some constant in your comparison lambda).

vpr/src/place/place.cpp
```diff
@@ -1252,9 +1251,11 @@ static float starting_t(const t_annealing_state* state, t_placer_costs* costs, t
     VTR_LOG("std_dev: %g, average cost: %g, starting temp: %g\n", std_dev, av, 20. * std_dev);
 #endif
 
-    /* Set the initial temperature to 20 times the standard of deviation */
+    /* Set the initial temperature to the standard of deviation divided by 64 */
```
vaughnbetz, line 1254: and so that the initial placement is not completely destroyed (lets us have a smarter initial placement and fine-tune it with the anneal).

vaughnbetz (changes requested):

vpr/src/pack/re_cluster_util.cpp
```diff
@@ -82,7 +82,7 @@ void commit_mol_move(const ClusterBlockId& old_clb,
         g_vpr_ctx.mutable_placement().block_locs.resize(g_vpr_ctx.placement().block_locs.size() + 1);
         get_imacro_from_iblk(&imacro, old_clb, g_vpr_ctx.placement().pl_macros);
         set_imacro_for_iblk(&imacro, new_clb);
-        place_one_block(new_clb, device_ctx.pad_loc_type, NULL);
+        place_one_block(new_clb, device_ctx.pad_loc_type, NULL, NULL);
```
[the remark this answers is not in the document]
saaramahmoudi: @MohamedElgammal  can you briefly tell me what does place_one_block do in here? Does this only care to find a location for a specific block? Right now, place_one_block uses the centroid_placement as the primary initial placement algorithm. Does it create any problem here? We can directly call place_macro instead of place_one_block.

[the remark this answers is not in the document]
MohamedElgammal: @saaramahmoudi I am calling this function to find a place for a new cluster --if created. If the place_one_block function is now running centroid placement, I agree that we can either try to place the new block in its centroid location (keep using place_one_block) or just place it in a random location using place_macro directly.

vpr/src/place/initial_placement.cpp
saaramahmoudi: Can't pass NULL when it is passed by reference.

vpr/src/place/initial_placement.cpp
```diff
@@ -236,6 +257,191 @@ static bool is_block_placed(ClusterBlockId blk_id) {
...
+    auto& place_ctx = g_vpr_ctx.mutable_placement();
+    bool legal = false;
+
+    //Check if the location is within its constraint region
+    for (auto reg : pr.get_partition_region()) {
+        if (reg.get_region_rect().contains(vtr::Point<int>(loc.x, loc.y))) {
+            //Check if blk_id type and the location physical types match
+            if (grid[loc.x][loc.y].type->index == block_type->index) {
+                //Check if the location has enough capacity left to place blk_id
+                if (place_ctx.grid_blocks[loc.x][loc.y].usage < grid[loc.x][loc.y].type->capacity) {
```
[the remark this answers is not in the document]
saaramahmoudi: I checked and found that we should check the capacity after determining the sub tiles, and it already exists in "macro_can_be_placed" function in place_util.cpp. So, I removed the condition because I don't need to check it beforehand.

vpr/src/pack/re_cluster_util.cpp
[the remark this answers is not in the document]
saaramahmoudi: @MohamedElgammal Is it okay if I change this to place_macro?

saaramahmoudi:
Yes, initial placement would go through centroid placement by default by the scaling_effort option is back to its default option which is circuit.

vaughnbetz:
Thanks.

PR #2179, opened 2022-10-22
Print WL Information Based on Direction and Length

amin1377:
- Print the number of wiring segments for each direction and length
- Print the utilization rate for each wiring segment length in each direction

To calculate the WL for a particular WL and direction, one can use the following formula:
WL = [length] * [utilization] * [number of wiring segment]

To be consistent with the current way that VPR prints information about wiring segment usage, I decided not to print the wire length directly and, instead, print the utilization rate.

@vaughnbetz


vpr/src/route/segment_stats.cpp
```diff
@@ -68,10 +80,54 @@ void get_segment_usage_stats(std::vector<t_segment_inf>& segment_inf) {
...
+        }
+
+    }
+
+
+    VTR_LOG("\n");
+    VTR_LOG("X - Directed Wiring Segment usage by length: length utilization\n");
+    VTR_LOG("                                             ------ -----------\n");
+
+    for (length = 1; length <= max_segment_length; length++) {
```
vaughnbetz, line 116: Couldn't this be shortened to doubly-nested loop (for loop over ax) just like the loop above?

vpr/src/route/segment_stats.cpp
```diff
@@ -68,10 +80,54 @@ void get_segment_usage_stats(std::vector<t_segment_inf>& segment_inf) {
...
+    VTR_LOG("\n");
+    VTR_LOG("Total Number of Wiring Segments by Direction: direction length number\n");
+    VTR_LOG("                                              --------- ------ -------\n");
+    for (length = 1; length <= max_segment_length; length++) {
+        for(auto ax : {X_AXIS, Y_AXIS}) {
+            std::string ax_name = (ax==X_AXIS) ? "X" : "Y";
+            if (directed_cap_by_length[ax][length] != 0) {
+                VTR_LOG("                                              %s %4d %6d\n", ax_name.c_str(),
+                        length,
+                        directed_cap_by_length[ax][length]);
```
vaughnbetz, line 105: Probably want to add code to print LONGLINE usage (index is the defined constant LONGLINE, which is index 0) if it exists.  Code is later in the routine; can just copy.  Longline isn't used much anymore, but probably good to keep it working since it isn't hard.

vpr/src/route/segment_stats.cpp
```diff
@@ -68,10 +80,54 @@ void get_segment_usage_stats(std::vector<t_segment_inf>& segment_inf) {
...
+    VTR_LOG("\n");
+    VTR_LOG("Y - Directed Wiring Segment usage by length: length utilization\n");
+    VTR_LOG("                                             ------ -----------\n");
+
+    for (length = 1; length <= max_segment_length; length++) {
+        if (directed_cap_by_length[Y_AXIS][length] != 0) {
+            utilization = (float)directed_occ_by_length[Y_AXIS][length] / (float)directed_cap_by_length[Y_AXIS][length];
+            VTR_LOG("                                             %6d %11.3g\n", length, utilization);
         }
     }
```
vaughnbetz, line 133: I think we can delete the segment usage by type below (old code, not directional).  The directional is a superset of it, so probably no need to keep both.

vaughnbetz: Would also be nice to refactor the "segment usage by type" code to use vectors, just like you've done for the new code.  Your call ...

vpr/src/route/segment_stats.cpp
```diff
@@ -41,25 +39,28 @@ void get_segment_usage_stats(std::vector<t_segment_inf>& segment_inf) {
         max_segment_name_length = std::max(max_segment_name_length, static_cast<int>(segment_inf[seg_type].name.size()));
     }
 
-    seg_occ_by_length = new int[max_segment_length + 1];
-    seg_cap_by_length = new int[max_segment_length + 1];
+    std::vector<int> seg_occ_by_length(max_segment_length+1, 0);
```
vaughnbetz, line 42: Could probably get rid of this entirely, and compute from the ax and ay data structures.

vaughnbetz: I think we can get rid of the total (x+y) usage by length.
Can keep the total (x+y) usage by segment type (aggregates data for electrically/topologically equivalent wires already).

amin1377:
@vaughnbetz: This branch is ready to land.

PR #2187, opened 2022-11-03
Initialized enums that might not be assigned depending on NO_GRAPHICS macro

hzeller:
small cleanup.


vpr/src/place/place.cpp
```diff
@@ -1462,7 +1462,7 @@ static e_move_result try_swap(const t_annealing_state* state,
         rlim = state->rlim;
     }
 
-    e_create_move create_move_outcome;
+    e_create_move create_move_outcome = e_create_move::ABORT;
```
MohamedElgammal, line 1466: I think we should move the #ifndef NO_GRAPHICS here and the #endif after "} else" and before the "if" in line 1472 so that the manual_move_enabled will only be checked if the graphics are enabled (NO_GRAPHICS isn't defined). Then, create_move_outcome will be initialized in all the cases.

vpr/src/place/place.cpp
```diff
@@ -1481,7 +1481,7 @@ static e_move_result try_swap(const t_annealing_state* state,
     ++move_type_stat.num_moves[(int)move_type];
     LOG_MOVE_STATS_PROPOSED(t, blocks_affected);
 
-    e_move_result move_outcome = ABORTED;
+    e_move_result move_outcome = e_move_result::ABORTED;
```
MohamedElgammal, line 1484: Initialized for any value of the macro

MohamedElgammal (approved):
Makes sense to initialize these values. However, I think it may be even cleaner to move the #ifndef NO_GRAPHICS macro in line 1469 to only check the manual_move_enabled flag when graphics are enabled.

vpr/src/place/place.cpp
vaughnbetz: @hzeller : any thoughts on Mohamed's comment?  I'm OK with merging as is, but if you and @MohamedElgammal have a better fix please make the change now so we fix this one way or the other.

vaughnbetz:
OK, made an issue to track that.

PR #2200, opened 2022-11-11
Partial tag

xanjohns:
#### Description

Adds a new `<partial>` tag available for use in xml architecture files. Like `<direct>, <complete>, and <mux>`, this is a tag to be used under the `<interconnect>` blocks within `<pb_types>`. 

#### Motivation and Context

Before this pull request, any interconnect that used a partial (sparse) crossbar, had to be defined manually by explicitly listing out a fraction of total possible connections using `<complete>` or `<direct>` tags. With this new feature, the many `<complete>` tags needed to simulate a partial crossbar can be replaced by a single `<partial>` tag. Using this tag's `fout` attribute, one can specify how many outputs each input should connect to, allowing a sparse crossbar to be created using only one tag.

Several xml arch files that already exist in VTR using partial crossbars were run through vtr_reg_basic, this time swapping the old style crossbars for the new `<partial>` tag. Channel widths were very similar using the new feature.

See results of this test [here](https://byu-my.sharepoint.com/:x:/g/personal/xanjohns_byu_edu/ESf74t4ydDFLsocuTFYDpd4Bs_LtoMxe70YflILs2RK_tA?e=i1L9pd)

#### Types of changes

#### Checklist:


libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -1634,28 +1634,33 @@ static void ProcessInterconnect(vtr::string_internment* strings, pugi::xml_node
...
 
     i = 0;
-    for (L_index = 0; L_index < 3; L_index++) {
+    for (L_index = 0; L_index < 4; L_index++) {
         if (L_index == 0) {
             Cur = get_first_child(Parent, "complete", loc_data, ReqOpt::OPTIONAL);
         } else if (L_index == 1) {
             Cur = get_first_child(Parent, "direct", loc_data, ReqOpt::OPTIONAL);
-        } else {
+        } else if (L_index == 2) {
```
jgoeders, line 1649: Is this magic number related to enum e_interconnect?  That enum has values 1,2,3,4, but this code is checking for values 0,1,2,3, so they are different, but they look related.  Should there be an enum for the arch file reading here?

jgoeders: Okay, I see what's going on now.  It's just looping through the number of items in the enum, but the actual L_index variable isn't really used.

To clean up the code a bit, perhaps we should add something like this:
```
std::map<enum e_interconnect, std::string> e_interconnect_to_arch_name = {{COMPLETE_INTERC, "complete"}, {DIRECT_INTERC, "direct"} ,...};
```

Then you can change this loop to:
```
for (auto e : e_interconnect_to_arch_name) {
}
```

vaughnbetz:
@amin1377 : this looks relevant to you -- can you take a look?

vaughnbetz:
Had some failures ... @jgoeders can you take a look and drive the merge of this one?

jgoeders:
Yes, I'll take care of this one.

jgoeders:
@vaughnbetz I think this is good to go now.

doc/src/arch/reference.rst
```diff
@@ -1537,9 +1537,29 @@ The following describes the tags that are accepted in the ``<interconnect>`` tag
 
         Mux interconnect example.
 
+.. arch:tag:: <partial name="string" input="string" output="string" fout="string">
+
+    :req_param name: Identifier for the interconnect.
+    :req_param input: Pins that are inputs to this interconnect. Different data lines are separated by a space.
+    :req_param output: Pins that are outputs of this interconnect.
+    :req_param fout: Number of outgoing edges for desired for each input pin.
```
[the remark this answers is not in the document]
vaughnbetz: Need to explain this a bit more.  Does it produce partial crossbar with fout edges from each input pin to a subset of the output pins, where the pattern is unspecified but chosen to roughly balance the number of inputs that can reach each output?  If so, say that.  If not, say what it does ...

vaughnbetz (changes requested):

jgoeders:
@WhiteNinjaZ I think we need to improve the pin assignment algorithm.  Right now the algorithm is doing this (within `alloc_and_load_partial_interc_edges`):

```c
int output_idx = 0;
starting_output_pin = 0;
for (input_pin in input_pins)
            output_idx = starting_output_pin++;

            /* Fill cross_bar_con map */
            for (int i = 0; i < fout; i++) {
                output_idx = output_idx % num_output_pins;
                // connect input_pin to output_idx
                output_idx = output_idx + (num_output_pins / fout);
            }
        }
    }
```

However, this does indeed create the pattern in the image from this PR and your slide presentation, where if you connect 4 input pins to 4 output pins with a fout = 2, you end up with the connectivity pattern [(0,2), (1,3), (0,2), (1,3)].   This isn't great.  We would like the pattern to be more scrambled and not have duplicate connection patterns to different output pins.

Can you think about how we could improve this and see if you can propose something?

WhiteNinjaZ:
Sure I will look into it.

vaughnbetz:
@WhiteNinjaZ @jgoeders : should this PR be merged or has it been superceded and should be closed?  I'd like to get everything for the 7-series done.

WhiteNinjaZ:
@vaughnbetz I believe this was built by another student a while ago and attempted to create a more succinct way to describe a partial crossbar. As I recall this feature had a few bugs in it and we determined that describing the 7-series interconnect manually created better results anyway. I will close this PR.

Issue #2202, opened 2022-11-15
vpr does not build with -std=c++17


heshpdx:
Some small edits can allow us to conform to the modern standard. Pull request to fix is coming shortly.
#### Steps to Reproduce

1. 
2. 

#### Your Environment

* VTR revision used: 3c1ea0885, Nov 14 2022
* Operating System and version: Ubuntu 20.04
* Compiler version: gcc-12.2

PR #2203, opened 2022-11-15
vpr: changes to conform to c++17 standard

heshpdx:
A few small edits to allow building with -std=c++17
#### Description

- std::bind was removed from c++17, as seen here: https://en.cppreference.com/w/cpp/utility/functional/bind12
- std::random_shuffle was removed from c++17, https://en.cppreference.com/w/cpp/algorithm/random_shuffle
- Replaced code with modern equivalents.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2202

#### Motivation and Context

vpr is being considered as a benchmark for the next SPEC CPU benchmark suite. We discovered this issue as part of that vetting process. Thus we are feeding back the changeset here.

I tested a handful of commandlines before and after the change, and there was no change in the output.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/simpleRL_move_generator.cpp
+++ b/vpr/src/place/simpleRL_move_generator.cpp
@@ -245,8 +245,8 @@ void SoftmaxAgent::set_action_prob() {
     // normalize all the action probabilities to guarantee the sum(all action probs) = 1
     float sum_prob = std::accumulate(action_prob_.begin(), action_prob_.end(), 0.0);
     std::transform(action_prob_.begin(), action_prob_.end(), action_prob_.begin(),
-                   [sum_prob, this](float x){ return x + ((1.0 - sum_prob) / this->num_available_actions_); });
-    
+                   [sum_prob, this](float x) { return x + ((1.0 - sum_prob) / this->num_available_actions_); });
+
     // calculate the accumulative action probability of each action
     // e.g. if we have 5 actions with equal probability of 0.2, the cumm_action_prob will be {0.2,0.4,0.6,0.8,1.0}
     float accum = 0;
--- a/vpr/src/place/simpleRL_move_generator.cpp
+++ b/vpr/src/place/simpleRL_move_generator.cpp
@@ -242,12 +242,12 @@ void SoftmaxAgent::set_action_prob() {
         }
     }
 
-    // normalize all the action probabilities to guarantee the sum(all actyion probs) = 1
+    // normalize all the action probabilities to guarantee the sum(all action probs) = 1
     float sum_prob = std::accumulate(action_prob_.begin(), action_prob_.end(), 0.0);
     std::transform(action_prob_.begin(), action_prob_.end(), action_prob_.begin(),
-                   bind2nd(std::plus<float>(), (1.0 - sum_prob) / num_available_actions_));
-
-    //calulcate the accumulative action probability of each action
+                   [sum_prob, this](float x){ return x + ((1.0 - sum_prob) / this->num_available_actions_); });
+    
+    // calculate the accumulative action probability of each action
     // e.g. if we have 5 actions with equal probability of 0.2, the cumm_action_prob will be {0.2,0.4,0.6,0.8,1.0}
     float accum = 0;
     for (size_t i = 0; i < num_available_actions_; ++i) {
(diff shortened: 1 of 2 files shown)


vaughnbetz:
@heshpdx : Thanks; the code looks good.  If you'd like to have a discussion on how to make vpr do a consistent amount of work (as there are some adaptive algorithms in it that can have slightly different amount of work depending on early round-off dependent decisions) let me know.  I did that for the spec98 version of vpr.

heshpdx:
Yes! That would be beneficial. We have the code building and running in the harness, but some ISAs have verification errors and I didn't know where to start debugging.
The search program is still open, so if you'd like you can submit an entry for vpr. I think 90% of the work is already done, but you would get 100% of the credit. We can use help on verification and choosing good cmdlines.
Enter here: https://www.spec.org/cpuv8/
Thanks!

vaughnbetz:
I'm not worried about credit -- happy to help if I can although time is tight.  If you paste in the verification errors I can give some ideas; I'll also ask my PhD students if anyone wants to help out.  

Sounds like they want runtimes in the ~10 minute range on a modern machine.  Is that right?  We have a very wide range of input circuit sizes and desired optimization quality so vpr can hit pretty much any cpu time target with a realistic test case.

heshpdx:
> I'm not worried about credit -- happy to help if I can although time is tight. If you paste in the verification errors I can give some ideas; I'll also ask my PhD students if anyone wants to help out.
>

@vaughnbetz The credit is more about the benchmark prize money if vpr is selected. As a committee member, I cannot claim it. But you can!

On verification, the main problem we see is that the results change based on compiler optimization level. We make sure to build with -fno-fast-math so that the std::isnan's in the code work properly, but even then we see differences between -O0 and -O3. I have crafted a tarball with some simple cmdlines along with the inputs {.blif, .xml} and outputs {.net, .place, .route} for both of those opt levels. We would really appreciate help in debugging this!

Attaching the tarball here. Please start with the README.txt inside. 
[vpr_verification.tar.gz](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/10263996/vpr_verification.tar.gz)

I am using gcc-12 however others have reproduced the issue with gcc-8 and gcc-11.

heshpdx:
@vaughnbetz I have some more info that can help narrow down the work. I set up to directories side-by-side, one built with -O0 and one built with -Ofast. I then took the link line and manually bisected the list of .o files to use either the O0 version or Ofast version. I ran the frankenstein'd binary and observed the output and checked for differences in the output files.

An afternoon of manual bisect/run uncovered that these five files are the ones which create a behavioral difference based on optimization level. 
```
vpr/src/pack/cluster_router.o
vpr/src/pack/cluster_util.o
vpr/src/place/directed_moves_util.o
vpr/src/place/place_util.o
vpr/src/route/route_tree_timing.o
``` 
All the remaining files in vpr and the libraries are agnostic to optimization level. I haven't dug further into this code yet, but I figured I would share to see if you have insight. Thanks!

vaughnbetz:
Thanks, that makes sense.

Basically it's making optimization decisions on floating point cost
functions; one different decision and the clustering/placement/routing
diverges and you get a different result.

What I did for spec cpu2000 was choose command line options that made the
algorithms a bit less adaptive so their work was more predictable, even if
they got a different answer.  Then I wrote out a metric of how much work
the placer and router performed, and checked it was within a small
tolerance (1% or 2% or so).  The program self checks its output is legal
for each algorithm, so other checking isn't very necessary; just need to
look for the "placement check succeeded" etc. output and make sure they
were there.

That same basic approach should work here as well.  If that's acceptable
myself or one of my students can sync up to choose the command line options
& what output to check in the new year.

Vaughn

vaughnbetz:
Thanks, that makes sense.

Basically it's making optimization decisions on floating point cost functions; one different decision and the clustering/placement/routing diverges and you get a different result.

What I did for spec cpu2000 was choose command line options that made the algorithms a bit less adaptive so their work was more predictable, even if they got a different answer.  Then I wrote out a metric of how much work the placer and router performed, and checked it was within a small tolerance (1% or 2% or so).  The program self checks its output is legal for each algorithm, so other checking isn't very necessary; just need to look for the "placement check succeeded" etc. output and make sure they were there. 

That same basic approach should work here as well.  If that's acceptable myself or one of my students can sync up to choose the command line options & what output to check in the new year.

heshpdx:
That approach sounds perfect, and the timing is great. Look forward to next steps!

vaughnbetz:
Adding @amin1377 to the thread since he's agreed to help out (thanks!).

PR #2205, opened 2022-11-15
RLPlace WL-driven placement error fix

MohamedElgammal:
#### Description

Adding error check to make sure that a correct reward function is chosen for WLdriven placement

#### Related Issue

trying to run experiments with wire length optimization only for placement (timing off). It says that it’s using the RL softmax agent however, only uniform random moves are getting used (median and centroid moves have 0% usage). 

#### Types of changes


MohamedElgammal:
@ArashAhmadian Thanks for highlighting the issue. I have added a check to make sure that the correct options are chosen in case of WL-driven placements. The program now will throw a meaningful error to ask the user to choose one of the supported reward functions.

MohamedElgammal:
@vaughnbetz I changed this to throw a warning and set the reward function basic and continue the run. If that's ok, we can land it.

PR #2208, opened 2022-11-18
Add missing instruction to run tests to the documentation

MohamedElgammal:
As vtr_nightly_test1 contains some symbiflow tests, we need to get_symbiflow_benchmarks to be able to run it. This is not included in the docs. Only titan and ISPD benchmarks are listed. In this PR, I have added the instruction of getting symbiflow benchmarks to the running tests section of the docs


vaughnbetz:
Looks good; golden results need an update and then can be merged.

PR #2220, opened 2023-01-05
Fix atom lookup

MohamedElgammal:
In this PR, I am using the atom-cluster lookup in the context everywhere in the flow instead of rebuilding a lookup every time we need it. I am also changing all the queries to use the cluster_to_atoms function. The lookup is not built first time itis queried.


MohamedElgammal:
@amin1377 This PR should speed up your atom lookup in routing. I will land it once the CI tests are done

PR #2225, opened 2023-01-13
Add operator-= as random access iterators need to provide that feature.

hzeller:
An updated std::sort algorithm is relying on it.

Context:
https://github.com/llvm/llvm-project/commit/4eddbf9f10a6d1881c93d84f4363d6d881daf848

Signed-off-by: Henner Zeller <hzeller@google.com>


hzeller:
There are build failures in the CI, but they [look like a set-up issue in the CI, but unrelated to this 4-line code-change](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/3912722040/jobs/6687726219#step:6:19).

PR #2227, opened 2023-01-14
This is a fix to avoid division by zero in initial_placement.

hermanschmit:
#### Description
We successfully adapted VPR to a different build system and compiler, but the execution failed when the code attempted to do a division by zero. Illegal instruction. 

#### Motivation and Context
The prior code worked, but it relied on a floating point division by zero resulting in an "inf" and then being converted to an int. This was sloppy, and relies on particular treatment of a SIGFPE. I think it is much cleaner, and more portable to avoid division if the divisor is zero.

Built in our system and tested.


saaramahmoudi:
Thanks for catching that, I think we should merge this. @vaughnbetz

saaramahmoudi:
@vaughnbetz before merging I need to change the condition afterward to make sure we don't break anything.

saaramahmoudi:
@hermanschmit @hzeller I don't want to make any changes to the branch before checking with you but we should set centroid loc in a else condition to a negative number, so the "try_centroid_placement" function would know we haven't found any legal location for the current block. Do you want me to make the change and make a pull request?

PR #2233, opened 2023-01-26
Support route constraint.

Tulong4Dev:
#### Description

- Allow router to route or skip specific signals.

#### Related Issue

- Currently, VPR ignores or routes a global signal through clock modeling options; it does not allow a specific signal to be treated seperately as desired.

#### Motivation and Context

- Allow route constraints applied to specific signals.

#### Types of changes

#### Checklist:

- [X] My change requires a change to the documentation

### DIFF
--- a/vpr/src/base/route_constraint.cpp
+++ b/vpr/src/base/route_constraint.cpp
@@ -1,0 +1,43 @@
+#include "route_constraint.h"
+
+RouteConstraint::RouteConstraint() {
+    net_name_ = std::string("");
+    net_type_ = std::string("");
+    route_method_ = std::string("");
+    is_valid_ = false;
+}
+
+void RouteConstraint::set_net_name(std::string name) {
+    net_name_ = name;
+    return;
+}
+
+std::string RouteConstraint::get_net_name() const {
+    return net_name_;
+}
+
+void RouteConstraint::set_net_type(std::string type) {
+    net_type_ = type;
+    return;
+}
+
+std::string RouteConstraint::get_net_type() const {
+    return net_type_;
+}
+
+void RouteConstraint::set_route_model(std::string route_method) {
+    route_method_ = route_method;
+    return;
+}
+
+std::string RouteConstraint::get_route_model() const {
+    return route_method_;
+}
+
+void RouteConstraint::set_is_valid(bool value) {
+    is_valid_ = value;;
+}
+
+bool RouteConstraint::get_is_valid() const {
+    return is_valid_;
+}
--- a/vpr/src/base/route_constraint.h
+++ b/vpr/src/base/route_constraint.h
@@ -1,0 +1,65 @@
+#ifndef ROUTE_CONSTRAINT_H
+#define ROUTE_CONSTRAINT_H
+
+#include "vpr_types.h"
+
+/**
+ * @file
+ * @brief This file defines the RouteConstraint class.
+ */
+
+class RouteConstraint {
+  public:
+    /**
+     * @brief Constructor for the RouteConstraint class, sets member variables to invalid values
+     */
+    RouteConstraint();
+
+    /**
+     * @brief get net name
+     */
+    std::string get_net_name() const;
+
+    /**
+     * @brief set net name
+     */
+    void set_net_name(std::string);
+
+    /**
+     * @brief get net type
+     */
+    std::string get_net_type() const;
+
+    /**
+     * @brief set net type
+     */
+    void set_net_type(std::string);
+
+    /**
+     * @brief get route model
+     */
+    std::string get_route_model() const;
+
+    /**
+     * @brief set route model
+     */
+    void set_route_model(std::string);
+
+    /**
+     * @brief set is valid 
+     */
+    void set_is_valid(bool);
+
+    /**
+     * @brief get is valid 
+     */
+    bool get_is_valid() const;
+
+  private:
+    std::string net_name_;
+    std::string net_type_;
+    std::string route_method_;
+    bool is_valid_;
+};
+
+#endif /* ROUTE_CONSTRAINT_H */


tangxifan:
Hi @vaughnbetz we are contributing new features to VTR these days. @Tulong4Dev is the EDA director who will lead the development. Can you add him as a contributor to VTR project? We will introduce our technical feature in details during the VTR meeting next week. Thanks!

tangxifan:
[golden_results.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/10998009/golden_results.txt)

This is what I got through my local run. I think you turned on the VTR_ASSERT_LEVEL=3 which may cause the issues.

tangxifan:
I think the santizer option is also on. It causes the issue.

vpr/src/base/vpr_constraints.cpp
```diff
@@ -58,6 +60,70 @@ PartitionRegion VprConstraints::get_partition_pr(PartitionId part_id) {
...
+
+RouteConstraint VprConstraints::get_route_constraint_by_net_name(std::string net_name) {
+    RouteConstraint rc;
+    auto const& rc_itr = route_constraints_.find(net_name);
+    if (rc_itr == route_constraints_.end()) {
+        // try regexp
+        bool found_thru_regex = false;
+        for (auto constraint : route_constraints_) {
+            if (std::regex_match(net_name, std::regex(constraint.first))) {
+                {
```
tangxifan, line 76: Not sure why we need an extra code block here.

vpr/src/base/vpr_constraints.cpp
```diff
@@ -58,6 +60,70 @@ PartitionRegion VprConstraints::get_partition_pr(PartitionId part_id) {
...
+    } else {
+        rc = rc_itr->second;
+    }
+    return rc;
+}
+
+RouteConstraint VprConstraints::get_route_constraint_by_idx(std::size_t idx) const {
+    RouteConstraint rc;
+    if ((route_constraints_.size() == 0) || (idx > route_constraints_.size() - 1)) {
+        rc.set_net_name("INVALID");
```
tangxifan, line 107: It seems to me that the ``set_net_name()``, ``set_net_type()`` and ``set_route_model()`` are redundant here. If the ``is_valid()`` return false, it does not need to be considered.
Maybe I miss some details, open to discussion.

vpr/src/util/vpr_utils.cpp
```diff
@@ -2274,6 +2273,22 @@ bool is_inter_cluster_node(t_physical_tile_type_ptr physical_tile,
     }
 }
 
+void apply_route_constraints(VprConstraints& vpr_constraint) {
```
tangxifan, line 2276: Not sure why we cannot use const &

vpr/src/base/vpr_constraints.cpp
```diff
@@ -58,6 +60,70 @@ PartitionRegion VprConstraints::get_partition_pr(PartitionId part_id) {
     return pr;
 }
 
+void VprConstraints::add_route_constraint(RouteConstraint rc) {
```
tangxifan, line 63: It looks to me that the current data structure is not efficient.
We have duplicated information which is the name of each routing constraint.
I suggest to merge and flatten the class ``RouteConstraint`` into ``VprConstraints``.
See an example at: 

https://github.com/lnis-uofu/OpenFPGA/blob/6adec23271b63ba112157b1f53afabf8e7bc7a20/libs/libbusgroup/src/bus_group.h#L108-L133

vpr/src/base/route_constraint.cpp
```diff
@@ -0,0 +1,43 @@
...
+std::string RouteConstraint::get_net_type() const {
+    return net_type_;
+}
+
+void RouteConstraint::set_route_model(std::string route_method) {
+    route_method_ = route_method;
+    return;
+}
+
+std::string RouteConstraint::get_route_model() const {
```
tangxifan, line 33: Let's follow our API naming convention. For instance ``get_route_model`` -> ``route_model``. All the accessors do not need any prefix ``get_``. Mutators do need ``set_`` or ``add_``. 
See example at 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/librrgraph/src/base/rr_graph_view.h

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/libs/librrgraph/src/base/rr_graph_builder.h

tangxifan (approved):
@Tulong4Dev Thanks for the contribution. Code changes look good in general. Need some tuning. Open to discussion.

tangxifan (changes requested):
Sorry. Wrong review conclusion.

vpr/src/util/vpr_utils.cpp
Tulong4Dev: fixed

vpr/src/base/route_constraint.cpp
Tulong4Dev: fixed.

vpr/src/base/route_constraint.h
```diff
@@ -0,0 +1,65 @@
+#ifndef ROUTE_CONSTRAINT_H
+#define ROUTE_CONSTRAINT_H
+
+#include "vpr_types.h"
+
+/**
+ * @file
+ * @brief This file defines the RouteConstraint class.
```
vaughnbetz, line 8: I think we should have a bigger comment on what this class is for, what it can currently do, what kind of code belongs here (so people know in the future if there is some new feature code that would best be put here), etc.

vaughnbetz:
@duck2 : can you take a look at the xml parser part of this pull request and give any feedback you think is appropriate?  It is building on some of your code.

tangxifan:
@Tulong4Dev I believe you are merging the codes in ``openfpga`` branch to the master in this PR. Please do not do so. You can see the code changes now contains 9k lines. A way to clean up is to create a new branch on your fork which only contains the routing constraint -related code changes.

vaughnbetz:
@kimiatkh is hoping this one can be merged soon, as she'd like to build some functionality on top of it.  @Tulong4Dev : any ETA?  If you're around for the VTR meeting this Thursday at 1:30 pm EST, this could be a good topic to discuss if there is some interaction with you and kimia that could help land this.

PR #2239, opened 2023-02-10
RLplace move mismatch

saaramahmoudi:
#### Description

RL agent move generator had an order dependency with e_move_type data structure defined in the "moves_util.h" file. The previous code uses the enum structure as an index of an array to access different move generators. However, the order of enum variables and avail_moves (defined in the "simpleRL_move_generator.cpp" file) array were different.

Hence, placement output for centroid and weighted centroid move was inaccurate. This change expects no performance change but removes the order dependency between these structures.

Also, the centroid move in e_move_type should be the third option (it is used to be the fourth) since if the placement algorithm is not timing, it is only used three first moves (Uniform, Median, Centroid). Having the weighted centroid move as the third move type in the enum mentioned above will result in a segmentation fault since criticalities would be NULL.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/simpleRL_move_generator.cpp
+++ b/vpr/src/place/simpleRL_move_generator.cpp
@@ -15,25 +15,29 @@ static float scaled_clipped_exp(float x) { return std::exp(std::min(1000000 * x,
  *                                     *
  *                                     */
 SimpleRLMoveGenerator::SimpleRLMoveGenerator(std::unique_ptr<SoftmaxAgent>& agent) {
-    avail_moves.emplace_back(std::make_unique<UniformMoveGenerator>());
-    avail_moves.emplace_back(std::make_unique<MedianMoveGenerator>());
-    avail_moves.emplace_back(std::make_unique<CentroidMoveGenerator>());
-    avail_moves.emplace_back(std::make_unique<WeightedCentroidMoveGenerator>());
-    avail_moves.emplace_back(std::make_unique<WeightedMedianMoveGenerator>());
-    avail_moves.emplace_back(std::make_unique<CriticalUniformMoveGenerator>());
-    avail_moves.emplace_back(std::make_unique<FeasibleRegionMoveGenerator>());
+    avail_moves.resize((int)e_move_type::NUMBER_OF_MOVES);
+
+    avail_moves[(int)e_move_type::UNIFORM] = std::make_unique<UniformMoveGenerator>();
+    avail_moves[(int)e_move_type::MEDIAN] = std::make_unique<MedianMoveGenerator>();
+    avail_moves[(int)e_move_type::CENTROID] = std::make_unique<CentroidMoveGenerator>();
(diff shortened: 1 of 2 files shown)


saaramahmoudi:
@vaughnbetz

vpr/src/place/simpleRL_move_generator.cpp
```diff
@@ -15,25 +15,29 @@ static float scaled_clipped_exp(float x) { return std::exp(std::min(1000000 * x,
...
+
+    avail_moves[(int)e_move_type::UNIFORM] = std::make_unique<UniformMoveGenerator>();
+    avail_moves[(int)e_move_type::MEDIAN] = std::make_unique<MedianMoveGenerator>();
+    avail_moves[(int)e_move_type::CENTROID] = std::make_unique<CentroidMoveGenerator>();
+    avail_moves[(int)e_move_type::W_CENTROID] = std::make_unique<WeightedCentroidMoveGenerator>();
+    avail_moves[(int)e_move_type::W_MEDIAN] = std::make_unique<WeightedMedianMoveGenerator>();
+    avail_moves[(int)e_move_type::CRIT_UNIFORM] = std::make_unique<CriticalUniformMoveGenerator>();
+    avail_moves[(int)e_move_type::FEASIBLE_REGION] = std::make_unique<FeasibleRegionMoveGenerator>();
 
     karmed_bandit_agent = std::move(agent);
```
vaughnbetz, line 28: There is one change in the code: before the size of "avail_moves" would have been 7 and there wouldn't have been an entry for Manual Moves.  Now the size is 8 and there is a default-initialized entry for Manual Moves.  Will that cause any problem in the move generator?  
We should figure out if Manual Moves should have a properly initialized value in this vector, or (more likely) should not have one.  If the latter, then change  the end of the enum definition to:

    FEASIBLE_REGION,
    NUMBER_OF_AUTO_MOVES,
    MANUAL_MOVE = NUMBER_OF_AUTO_MOVES

Also, the comma after the last enum constant seems unusual; I'd delete it.

vaughnbetz:
@saaramahmoudi @MohamedElgammal : I think the code needs a change; see the comment above.

saaramahmoudi:
@vaughnbetz Applied suggestions and also updated the branch to be able to merge it to master.

vaughnbetz:
To be very cautious, it would be best to attach a QoR spreadsheet (e.g. on the large VTR designs, or Titan) showing no quality change after this (or that the results are identical on a  circuit, which would also be strong confirmation of equivalent QoR).

saaramahmoudi:
@vaughnbetz @MohamedElgammal : Attached qor comparison for titan_quick_qor benchmark. I think we are good to merge!
[titan_quick_qor](https://docs.google.com/spreadsheets/d/1-EUid14Ca5qR6i8LWjJkS3ol4ujTg6pxCjdIzSDVRks/edit?usp=sharing)

PR #2254, opened 2023-03-14
Spec Rand

amin1377:
Add support to use SPEC random number generator if SPEC_CPU is defined.

@heshpdx 
@vaughnbetz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ option(VTR_ENABLE_PROFILING "Enable performance profiler (gprof)" OFF)
 option(VTR_ENABLE_COVERAGE "Enable code coverage tracking (gcov)" OFF)
 option(VTR_ENABLE_DEBUG_LOGGING "Enable debug logging" OFF)
 option(VTR_ENABLE_VERBOSE "Enable increased debug verbosity" OFF)
+option(SPEC_CPU "Enable SPEC CPU v8 support" OFF)
 
 #Allow the user to decide whether to compile the graphics library
 set(VPR_USE_EZGL "auto" CACHE STRING "Specify whether vpr uses the graphics library")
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -298,6 +299,16 @@ if(VTR_ENABLE_VERBOSE)
     message(STATUS "Enabling increased debugging verbosity")
 endif()
 
+#
+# Build for SPEC CPU Benchmark v8
+#
+set(SPEC_CPU_FLAGS "")
+if(SPEC_CPU)
+    #Enable gcov
```
vaughnbetz, line 307: Why does this say enable gcov?

vaughnbetz: Comment on what this define is for ... what has to change / what's the purpose?  (spec benchmarking version of vtr; maximizes portability of code and mimimizing the variation in how much work we do on different platforms).

libs/libvtrutil/CMakeLists.txt
```diff
@@ -83,7 +83,7 @@ add_custom_command(OUTPUT ${VTR_VERSION_FILE_OUT}
 #
 # Source files and library
 #
-file(GLOB_RECURSE LIB_SOURCES src/*.cpp)
+file(GLOB_RECURSE LIB_SOURCES src/*.cpp src/*.c)
```
vaughnbetz, line 86: Probably easier to make the new specrand.c file into specrand.cpp --> ask Mahesh if that's OK.  If not, we can do this.

libs/libvtrutil/src/specrand.c
```diff
@@ -0,0 +1,197 @@
...
+   NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+   SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+   Any feedback is very welcome.
+   http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/emt.html
+   email: m-mat @ math.sci.hiroshima-u.ac.jp (remove space)
+*/
+/* Slightly modified for use in SPEC CPU by Cloyce D. Spradling (5 Nov 2009)
+ */
```
vaughnbetz, line 46: Comment why this is in vtr --> for inclusion in the SPEC cpu benchmarks, and it helps ensure we get the random numbers on different platforms.

vaughnbetz (commented):
A few commenting suggestions.

libs/libvtrutil/CMakeLists.txt
heshpdx: It's up you how you want to pull in these files. If you'd like to convert to cpp, please go ahead. There is no requirement from my side :-)

Issue #2197, opened 2022-11-10
Generalize sub_tile in the grid so sub_tiles can have different width & height


vaughnbetz:
#### Proposed Behaviour
We'd like to model multi-die stack FPGAs. This is needed for the crossroads project, and will likely be useful to others.

#### Current Behaviour
The sub_tile feature almost has what we need in order to model one block under another, but it has the restriction that all sub_tiles in a tile must be the same size.  We'd like to model a large router (or large accelerator block) that is underneath multiple LABs (or maybe RAMs, etc.).  Right now we can't do this -- if we make the router large (e.g. width = 3, height = 3), then any LABs in the architecture become wide and tall.

#### Possible Solution
I think if we make the width, height, x_offset and y_offset in the grid a function of the sub_tile instead of a tile, we can have this feature.  The capacity of grid[i][j] would then give the number of sub_tiles overlapping a certain (x,y) location, and you'd go through all those sub_tiles to see where their anchor points where, what their widths and heights are etc.

We should also use this change as an opportunity to wrap some member functions around access to the sub_tiles to ask questions about them and anchor points etc. (I'm assuming we have code directly accessing this data now).

#### Context
Impacting exploring routers with logic over them for the Crossroads project.

#### Alternative proposal
We could add a z dimension to the grid, making the layers explicit. But I believe this would involve more code updates, and probably a higher burden on more code to think about / set the z-coordinate, which would always be 0 for conventional planar architectures.

vaughnbetz:
@kmurray @tangxifan : FYI.  Sara is going to dig into this, with an eye to discussing in a future vtr meeting.

tangxifan:
@vaughnbetz Thanks for the info. Will do.

saaramahmoudi:
Different diagrams for our proposed solutions to this problem are shown below. Blocks with green are the ones that we should add to the existing code.

1. The first diagram is a straightforward approach in which we don't need to change sub-tile and tile structures or even how the grid exists. But, we need to store sub-tile attributes multiple times (as many as LAB existed on top of the NoC) 

![Screenshot from 2022-11-21 11-01-26](https://user-images.githubusercontent.com/29703355/203101459-67a0472d-57ab-4fa7-b57c-6e5fdec3a75e.png)

2. The second diagram Vaughn mentioned in the previous comment as adding a new explicit z dimension.

[...]

PR #2259, opened 2023-03-22
Modeling multi die stack 

saaramahmoudi:
#### Description
To support Multi-die FPGAs, we need to be able to support new xml tag called <layer>, each layer tag will specify a die. This PR will modify the XML parser to find these tags (default to 1 in case unspecified) and build a 3D or multi-die grid. The modification is backward compatible, hence existing architecture will not need to change. 

The architecture modification is compatible with the existing architecture files and it doesn't require any modification to the existing code.  I confirmed the parser functionality for multi-die FPGA with two architectures that I wrote with two dice (one very small architecture and the other is modification to the existing Stratix-IV) 

To land this PR, we first need to land [grid_3d](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2246) PR since they have dependency to each other. 

#### Related Issue
Further discussion of how we modify the XML file and parser is given on [issue ##2197](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2197).

#### Motivation and Context
Supporting Multi-die FPGAs in VPR.

#### Types of changes

#### Checklist:

- [] My change requires a change to the documentation


saaramahmoudi:
@amin1377

amin1377:
@vaughnbetz : As discussed, I changed the grid from a vector of 2D matrices to a 3D matrix.

vaughnbetz (changes requested):
Looks good -- thanks for the accessors!

amin1377:
@vaughnbetz : Your suggestions are applied, and this PR is ready to merge. [Here](https://utoronto-my.sharepoint.com/:x:/g/personal/amin_mohaghegh_mail_utoronto_ca/Ea5Sg1uCPJBKkEgv-eIDOaQBlNZ0tA9rpwAwuXeEWtRPnQ?e=J2hefC) is the QoR comparison for Titan and VTR benchmarks.

vaughnbetz:
I think this is now ready to be merged, so the WIP should be removed from the title.  Will come check before merging ...

PR #2298, opened 2023-04-26
Make delete calls not look like function calls.

hzeller:
Minor purely textual cleanup.


vaughnbetz:
Needs a make format.  It's an odd formatting request, but best to do it anyway:

-        delete[] *pb_graph_pins;
+        delete[] * pb_graph_pins;

hzeller:
Done.

PR #2304, opened 2023-05-05
[WIP] Thread-safe re-clustering API and the connection-driven IIP

MohamedElgammal:
In this PR, the re-clustering API is extended to be thread-safe. Moreover, some bugs in the API are fixed.

Then, the API is utilized to optimize connection-driven packing using iterative improvement packing IIP. Currently working on extending it to be timing driven.


MohamedElgammal:
@amin1377

vaughnbetz:
@MohamedElgammal : should these thread-safe improvements be merged?

PR #2306, opened 2023-05-08
fix readthedocs build

duck2:
See https://github.com/readthedocs/readthedocs.org/issues/10290


duck2:
there's a failing check but that's not related to this PR

PR #2311, opened 2023-05-11
Placement move type and block type print information

saaramahmoudi:
#### Description
After combining RLPlace 2.0 and NoC placement, details about placement statistics (which block has been moved, how many moves were accepted/rejected, ..) was printed twice for NoCs. Since RLPlace 2.0 code was general and contained all block types, I have removed noc code related to printing information about moves to avoid duplication. 

In addition to that, we needed block type information regardless of using RL agent as placement algorithm. Hence, I have moved the block type extraction function to execute regardless of the placer algorithm, so we can have information printed even if we are using StaticMoveGenerator. 

@vaughnbetz 

#### Types of changes

#### Checklist:


vaughnbetz:
Looks good -- nice code cleanup!

vaughnbetz (approved):

PR #2314, opened 2023-05-15
Min Chan W Number of Router Iterations

amin1377:
This pull request updates the behavior of the "run_vtr_flow.py" script to correctly handle the "max_router_iterations" argument. Currently, the script assumes that this argument is only for the router iterations on relaxed channel width. Thus, the VPR's default value of 50 iterations was used when attempting to find the minimum channel width. With this update, "max_router_iterations" determines the number of iterations for finding the minimum channel width, while "crit_path_router_iteration" is used for the relaxed channel width.

For example, running the script with the following command:

`run_vtr_flow.py ch_intrinsics.v k6_frac_N10_frac_chain_mem32K_40nm.xml -start odin -max_router_iterations 100 -crit_path_router_iteration 150`

Will set the maximum number of iterations for finding the minimum channel width to 100, and for the relaxed channel width to 150.

@vaughnbetz


amin1377:
@vaughnbetz: This is ready to land

PR #2338, opened 2023-06-12
turn off tatum's TBB by default

duck2:
#### Description
Running parallel tatum currently yields nondeterministic results. This is not documented anywhere, so disable it by default to prevent surprises.

#### Types of changes

#### Checklist:


libs/EXTERNAL/libtatum/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.16)
 
 project("tatum")
 
-set(TATUM_EXECUTION_ENGINE "auto" CACHE STRING "Specify the framework for (potential) parallel execution")
+set(TATUM_EXECUTION_ENGINE "serial" CACHE STRING "Specify the framework for (potential) parallel execution")
```
vaughnbetz, line 5: Should comment this -- default set to serial for now as we're getting nondeterministic results from Tatum with TBB.

PR #2339, opened 2023-06-13
Vpr graphics update

Meet-Patel2580:
#### Description

Fixed segmentation faults caused by the user clicking features that should only be available during routing. Updated Critical path arrows to automatically resize based on the current zoom level.

#### Types of changes

#### Checklist:


vpr/src/draw/draw.cpp
```diff
@@ -312,7 +312,7 @@ static void initial_setup_NO_PICTURE_to_PLACEMENT(ezgl::application* app,
 
     //Hiding unused functionality
     hide_widget("RoutingMenuButton", app);
-    hide_crit_path_button(app);
+    // hide_crit_path_button(app);
```
vaughnbetz, line 315: Delete this line instead of commenting out -- we still have it in git.

vpr/src/draw/draw.cpp
```diff
@@ -326,7 +326,8 @@ static void initial_setup_NO_PICTURE_to_PLACEMENT_with_crit_path(
 
     //Showing given functionality
     crit_path_button_setup(app);
-
+    // hide crit path routing / routing delays
```
vaughnbetz, line 329: Why?  Routing data isn't available, so hide the options to display. (say why, not just that you're hiding this).

Suggested comment.

// Routing hasn't been done yet, so hide the display options that show routing as they don't make sense and would crash if clicked on.

vpr/src/draw/draw.cpp
```diff
@@ -340,7 +341,9 @@ static void initial_setup_PLACEMENT_to_ROUTING(ezgl::application* app,
         default_setup(app);
 
     routing_button_setup(app);
-    hide_crit_path_button(app);
+    // hide_crit_path_button(app);
```
vaughnbetz, line 344: delete this line

vpr/src/draw/draw.cpp
```diff
@@ -353,7 +356,9 @@ static void initial_setup_ROUTING_to_PLACEMENT(ezgl::application* app,
 
     //Hiding unused functionality
     hide_widget("RoutingMenuButton", app);
-    hide_crit_path_button(app);
+    // hide_crit_path_button(app);
```
vaughnbetz, line 359: delete commented line

vpr/src/draw/draw.cpp
```diff
@@ -365,7 +370,9 @@ static void initial_setup_NO_PICTURE_to_ROUTING(ezgl::application* app,
         default_setup(app);
 
     routing_button_setup(app);
-    hide_crit_path_button(app);
+    // hide_crit_path_button(app);
```
vaughnbetz, line 373: delete commented line

vpr/src/draw/draw_triangle.cpp
```diff
@@ -54,14 +54,45 @@
...
+ * The zoom level is calculated as the ratio of the visible world width to the visible screen width.
+ * The zoom level can be adjusted within the specified range and scaled by a default zoom factor.
+ *
+ * @param g The renderer object.
+ * @param default_zoom The default zoom factor for the default zoom level (1.0 by default).
+ * @param min_zoom The minimum allowed zoom level (0.5 by default).
+ * @param max_zoom The maximum allowed zoom level (2.3 by default).
+ * @return The calculated zoom level.
+ */
+double get_zoom_level(ezgl::renderer* g) {
```
vaughnbetz, line 68: Rename or flip the number (0.5 currently means 2x zoom).
I think a better name would be draw_scaling_factor_from_zoom
Comments should say that for some annotations (e.g. the triangles on critical path display) we should scale them some as we zoom in or out. This routine figures out how much. A value < 1 means the triangle will be drawn with smaller world coordinates.

vaughnbetz: Usually put the usage comments in the .h file only.

vpr/src/draw/draw_triangle.h
```diff
@@ -48,6 +48,18 @@
 #    include "place_macro.h"
 #    include "buttons.h"
 
+/**
+ * Retrieves the current zoom level based on the visible world and screen dimensions.
```
vaughnbetz, line 52: See comment changes in the .cpp, but move them here.

vpr/src/draw/ui_setup.cpp
```diff
@@ -174,7 +174,21 @@ void hide_crit_path_button(ezgl::application* app) {
     hide_widget("CritPathLabel", app);
     hide_widget("ToggleCritPath", app);
 }
-
+/*
+ * @brief hides or displays critical path routing / routing delay
```
vaughnbetz, line 178: Add at end: "UI elements,  Use to ensure we don't show inactive buttons etc. when routing data doesn't exist."

vaughnbetz: Can just put this in the .h file.

vaughnbetz (changes requested):

vaughnbetz:
Please resolve the conflicts and I'll merge this.

vaughnbetz:
Nag, nag ... conflicts ...

PR #2341, opened 2023-06-15
Fixing the pb_pin_fixup for hardbocks_review

behnam-rs:
#### Description
We have fixed the bug via pb_pin_fixup raised via hard block like DSPs and BRAMs.

#### Motivation and Context
There was a bug when we call pb_pin_fixup.

We have tested it via our FPGA architecture and several testcases

#### Types of changes
- [X ] Bug fix (change which fixes an issue)

#### Checklist:


vpr/src/pack/cluster_placement.cpp
```diff
@@ -127,6 +127,41 @@ bool get_next_primitive_list(t_cluster_placement_stats* cluster_placement_stats,
     for (i = 0; i < cluster_placement_stats->num_pb_types; i++) {
         if (!cluster_placement_stats->valid_primitives[i].empty()) {
             t_cluster_placement_primitive* cur_cluster_placement_primitive = cluster_placement_stats->valid_primitives[i].begin()->second;
+                
+            auto& atom_ctx = g_vpr_ctx.atom();
```
vaughnbetz, line 131: This block of code should have a comment ... what is being done here?  If possible, it would also be good to make this a helper function as this is a long routine.

vpr/src/pack/post_routing_pb_pin_fixup.cpp
```diff
@@ -153,49 +153,55 @@ static void update_cluster_pin_with_post_routing_results(const Netlist<>& net_li
...
-                                      pb_graph_pin->to_string().c_str(),
-                                      net_list.net_name(routing_net_id).c_str(),
-                                      net_list.net_name(rr_node_nets[rr_node]).c_str());
-                    }
-                    VTR_ASSERT(routing_net_id == rr_node_nets[rr_node]);
-                }
-                routing_net_id = rr_node_nets[rr_node];
-                valid_routing_net_cnt++;
-                visited_rr_nodes.push_back(rr_node);
+        int addToY = 0, addToX = 0;
```
vaughnbetz, line 156: Again, good to comment why this is being done (what is the loop over x and y for)?
Consider factoring this into a helper routine.

vpr/src/pack/post_routing_pb_pin_fixup.cpp
```diff
@@ -153,49 +153,55 @@ static void update_cluster_pin_with_post_routing_results(const Netlist<>& net_li
...
+        addToY = physical_tile->height;
+        addToX = physical_tile->width;
```
vaughnbetz, line 158: Better coding practice to avoid the dummy initialization to 0, and instead just directly say
int addToY = physical_tile->height;

(Shorter and more idiomatic).

vpr/src/pack/post_routing_pb_pin_fixup.cpp
```diff
@@ -581,6 +587,19 @@ static void rec_remove_downstream_pb_routes(t_pb_routes& curr_pb_routes,
     }
 }
 
+static void rec_remove_downstream_pb_routes_ignore_used_items(t_pb_routes& curr_pb_routes,
```
vaughnbetz, line 590: Add a declaration of this function at the top of the file in the local function section.
Also add a comment of what the function does above that declaration, and comment what each input is.

vaughnbetz (changes requested):
Some stylistic changes requested.  Also, is there a regtest to check this behaviour continues to work? If not, it would be good to make one and add it to an appropriate regtest suite (likely one of the vtr_reg_nightly tests; could be vtr_reg_strong if the test case is very small and runs very fast, i.e. a few seconds).

hzeller:
+1 to adding a test. Something that fails before and passes after.

vpr/src/pack/cluster_placement.cpp
behnam-rs: stash the change for this file as its a local modification.

vpr/src/pack/post_routing_pb_pin_fixup.cpp
behnam-rs: We apply your comments.

vpr/src/pack/post_routing_pb_pin_fixup.cpp
behnam-rs: We apply your comments.

vpr/src/pack/post_routing_pb_pin_fixup.cpp
behnam-rs: We apply your comments.

behnam-rs:
> +1 to adding a test. Something that fails before and passes after.

Our test is architecture dependent and hence, not possible to include it here.

vpr/src/pack/post_routing_pb_pin_fixup.cpp
vaughnbetz: Thanks

doc/requirements.txt
```diff
@@ -2,7 +2,7 @@
 # used by Read The Docs to install python required
 # modules with pip.
 
-sphinx
+sphinx<7
```
vaughnbetz, line 5: Good to comment why we need spinx < 7 (breaking change in 7?  rtd them not supported in 7+ ?)

vaughnbetz (approved):

vaughnbetz:
Please rebase onto the latest master branch to make the CI errors go away; @duck2 fixed an issue where the github runners decided the abc and capnproto cmake files were too old a version and wouldn't build them. @behnam-rs : I think if you accept the vtr invitation I sent you that the CI tests will start running for you automatically, without any approval from me.

PR #2343, opened 2023-06-26
Read flat router .route file

saaramahmoudi:
#### Description
Once we used flat router to route the design, newly generated .route file couldn't be read by VPR anymore. We couldn't load any .route file generated by flat router. This PR address this issue and add support for reading any .route file regardless of how it is generated (normal or flat routing). 

#### Types of changes

#### Checklist:


vaughnbetz:
The code looks good.  
1. Please also make a regtest for this.
2. What happens if you try to read a flat routing file when flat routing isn't specified or vice versa?  Please try it out, and if at all possible, give a good error message (instead of a seg fault or baffling assert).

saaramahmoudi:
@vaughnbetz It wouldn't be seg-fault, it will be out with an assertion which tell the user that rr-graph node in the .route file doesn't exist.

vaughnbetz:
We should replace the assert with an error message that gives them an idea of what went wrong (rr-node does not exist. Running without flat routing; if this file was created with flat routing, re-run vpr with the --flat_routing option
(not sure of the exact option so please check that).

@amin1377 : It doesn't seem the flat routing option is documented in the vpr command line options. Can you add it?

amin1377:
> We should replace the assert with an error message that gives them an idea of what went wrong (rr-node does not exist. Running without flat routing; if this file was created with flat routing, re-run vpr with the --flat_routing option (not sure of the exact option so please check that).
> 
> @amin1377 : It doesn't seem the flat routing option is documented in the vpr command line options. Can you add it?

There is a brief description available if you run `vpr -h`. You can find the corresponding code [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/base/read_options.cpp#L2367). However, if you are referring to the online documentation, you are right. I haven't included any information about the flat-router there yet. I will make sure to add it as soon as possible.

vaughnbetz:
Thanks. Yes, we should put it in the online documentation too.

saaramahmoudi:
@vaughnbetz Three places can give us error if we read .route file that is generated with flat router, I added message for all three that maybe the user should try --flat_routing true option.
Can we please merge this one and I will make another PR for regtest if that is possible.

saaramahmoudi:
@vaughnbetz can you please merge this?

PR #2344, opened 2023-06-26
fix rtd build

duck2:
looks like this issue to me? https://github.com/readthedocs/readthedocs.org/issues/10279


vaughnbetz:
Thanks for figuring out what looks like a nasty issue!

PR #2345, opened 2023-07-02
Recompte noc costs when restorting a checkpoint

soheilshahrouz:
#### Description

When a placement checkpoint is restored, wirelength and timing-driven costs are computed from scratch. However, NoC costs were left intact, meaning that a mismatch between router locations in the restored checkpoint and the placement before restoring the checkpoint causes the NoC costs to be no longer valid. Changes in this pull request make sure that NoC-related costs and internal data structures are re-initialized when a placement checkpoint is restored.

#### Motivation and Context

If NoC-related costs are not re-computed after restoring a checkpoint, NoC costs may not be valid.

Placement checkpoints are only stored during quench. I could not find a reproducible case where routers are swapped during the quench, and as result, the restored checkpoint ended up having a different router placement. Since checkpoints are only stored during the quench and swapping routers during quench rarely improves WL and TD costs, I needed to artifically produce cases where the restored checkpoint had a different router placement. To do this, I used manual moves and temporarily changed the code to store a checkpoint when a router was manually moved.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/noc/xy_routing.h
+++ b/vpr/src/noc/xy_routing.h
@@ -140,7 +140,7 @@ class XYRouting : public NocRouting {
      * that is currently being visited on the FPGA
      * @return RouteDirection The direction to travel next
      */
-    RouteDirection get_direction_to_travel(int sink_router_x_position, int sink_router_y_position, int curr_router_x_position, int curr_router_y_position);
+    static RouteDirection get_direction_to_travel(int sink_router_x_position, int sink_router_y_position, int curr_router_x_position, int curr_router_y_position);
 
     /**
      * @brief Given the direction to travel next, this function determines
@@ -166,7 +166,7 @@ class XYRouting : public NocRouting {
      * @return true A suitable link was found that we can traverse next
      * @return false No suitable link was found that could be traversed
      */
-    bool move_to_next_router(NocRouterId& curr_router_id, int curr_router_x_position, int curr_router_y_position, RouteDirection next_step_direction, std::vector<NocLinkId>& flow_route, std::unordered_set<NocRouterId>& visited_routers, const NocStorage& noc_model);
+    static bool move_to_next_router(NocRouterId& curr_router_id, int curr_router_x_position, int curr_router_y_position, RouteDirection next_step_direction, std::vector<NocLinkId>& flow_route, std::unordered_set<NocRouterId>& visited_routers, const NocStorage& noc_model);
 };
 
 #endif
--- a/vpr/src/noc/xy_routing.h
+++ b/vpr/src/noc/xy_routing.h
@@ -140,7 +140,7 @@ class XYRouting : public NocRouting {
      * that is currently being visited on the FPGA
      * @return RouteDirection The direction to travel next
      */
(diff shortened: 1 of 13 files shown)


vpr/src/noc/xy_routing.h
```diff
@@ -140,7 +140,7 @@ class XYRouting : public NocRouting {
      * that is currently being visited on the FPGA
      * @return RouteDirection The direction to travel next
      */
-    RouteDirection get_direction_to_travel(int sink_router_x_position, int sink_router_y_position, int curr_router_x_position, int curr_router_y_position);
+    static RouteDirection get_direction_to_travel(int sink_router_x_position, int sink_router_y_position, int curr_router_x_position, int curr_router_y_position);
```
vaughnbetz, line 143: Is there any reason to make this static?  I guess it doesn't need the calling object and hence can be made static, but if there's no real advantage to making it static (we have an object we can call it with) I'd lean toward keeping it a normal method for simplicity and in case some future change means you need to look something up in the calling object.

vpr/src/noc/xy_routing.h
```diff
@@ -166,7 +166,7 @@ class XYRouting : public NocRouting {
      * @return true A suitable link was found that we can traverse next
      * @return false No suitable link was found that could be traversed
      */
-    bool move_to_next_router(NocRouterId& curr_router_id, int curr_router_x_position, int curr_router_y_position, RouteDirection next_step_direction, std::vector<NocLinkId>& flow_route, std::unordered_set<NocRouterId>& visited_routers, const NocStorage& noc_model);
+    static bool move_to_next_router(NocRouterId& curr_router_id, int curr_router_x_position, int curr_router_y_position, RouteDirection next_step_direction, std::vector<NocLinkId>& flow_route, std::unordered_set<NocRouterId>& visited_routers, const NocStorage& noc_model);
```
vaughnbetz, line 169: Same question on static

vpr/src/place/place_checkpoint.cpp
```diff
@@ -43,6 +44,11 @@ void restore_best_placement(t_placement_checkpoint& placement_checkpoint, std::s
                                    timing_info.get(),
                                    &costs);
 
+        // If NoC is enabled, re-compute NoC costs and re-initialize NoC internal data structures
```
vaughnbetz, line 47: Good to say why (the NoC placement may have changed, and hence the routing of traffic flows and NoC costs need to be recomputed).

vaughnbetz (changes requested):
Code looks very clean (in fact you cleaned up some related code) -- thanks!
A few minor suggestions on not making class functions static unless necessary and a comment.

vpr/src/noc/xy_routing.h
soheilshahrouz: My aim was to make it clear that this member function does not use any non-static member variables/functions. The only advantage that I can think of is that no "this" pointer is passed to the function, and as a result, the function is slightly more memory efficient, which is not that important. I'll change it back to a normal function.

vpr/src/noc/xy_routing.h
soheilshahrouz: Removed static in a new commit.

PR #2348, opened 2023-07-12
Enhanced 3D Placement Cost and Router Lookahead with Cross-Die Delay

amin1377:
In this PR, I have introduced a new move to the placement stage, enabling the modification of a block's layer (in the second state of the agent). Additionally, the first three moves of the placement stage, which do not target the critical paths, have been enhanced in the following way. After calculating the x-y location, they now check whether the current layer has available space. If not, they search across layers to identify any free blocks at the designated x-y location. In case a free block is found, it is relocated to that layer; otherwise, the block remains in its current layer.

Furthermore, the router lookahead mechanism has been updated to account for the cost associated with crossing layers.


amin1377:
@vaughnbetz This PR is ready for review.

amin1377:
> Let's go over this today. Does this mean the router lookahead and placer delay matrix have been extended to 3D with no significant runtime cost?

The changes related to making the router lookahead and placer delay lookup 3D have already been merged. In this PR, I updated the router lookahead to be aware of connections with delay (the previous PR assumed perfect connections). Additionally, I identified and resolved some bugs related to 3D P&R. The bounding box has also been updated, so if a 3D architecture is used, there will be a separate bounding box for each layer.

vaughnbetz:
1. Please file an issue or add code to auto-detect whether we have wire-to-wire layer  changing switches and choose 3D vs. multiple bounding boxes based on that.  Comment this well, and set a placer option based on its state so it's also easy to add a command line option in the future if necessary.

vaughnbetz:
CI failing, apparently due to experimental code. Will need to take that out for the PR to be mergable.

amin1377:
Titan QoR on latest commit:
Screenshot from 2023-10-19 16-09-12

libs/librrgraph/src/base/rr_graph_utils.h
```diff
@@ -48,4 +48,5 @@ vtr::vector<RRNodeId, std::vector<RREdgeId>> get_fan_in_list(const RRGraphView&
 int seg_index_of_cblock(const RRGraphView& rr_graph, t_rr_type from_rr_type, int to_node);
 int seg_index_of_sblock(const RRGraphView& rr_graph, int from_node, int to_node);
 
+bool inter_layer_connections_limited_to_opin(const RRGraphView& rr_graph);
```
vaughnbetz, line 51: Put a comment on what this does (Doxygen format).

vpr/src/base/SetupVPR.cpp
```diff
@@ -622,6 +622,8 @@ static void SetupPlacerOpts(const t_options& Options, t_placer_opts* PlacerOpts)
 
     PlacerOpts->constraints_file = Options.constraints_file;
 
+    PlacerOpts->write_initial_place_file = Options.write_initial_place_file;
```
vaughnbetz, line 625: Also add this to the documentation

vpr/src/base/place_and_route.cpp
```diff
@@ -355,7 +355,7 @@ int binary_search_place_and_route(const Netlist<>& placement_net_list,
                     auto& cluster_ctx = g_vpr_ctx.clustering();
                     // Cluster-based net_list is used for placement
                     print_place(filename_opts.NetFile.c_str(), cluster_ctx.clb_nlist.netlist_id().c_str(),
-                                filename_opts.PlaceFile.c_str());
+                                filename_opts.PlaceFile.c_str(), false);
```
vaughnbetz, line 358: The false should be gone ...  just want print_place. No need for it to know if it is the initial placement or not.

vpr/src/base/read_options.cpp
```diff
@@ -430,6 +430,41 @@ struct ParsePlaceAlgorithm {
     }
 };
 
+struct ParsePlaceBoundingBox {
```
vaughnbetz, line 433: Add to online documentation.

vpr/src/base/read_options.cpp
```diff
@@ -2007,6 +2047,21 @@ argparse::ArgumentParser create_arg_parser(std::string prog_name, t_options& arg
         .default_value("10")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    place_grp.add_argument<e_place_bounding_box_mode, ParsePlaceBoundingBox>(args.place_bounding_box_mode, "--place_bounding_box_mode")
+        .help(
+            "Specifies the type of bounding box to be used in 3D architectures.\n"
```
vaughnbetz, line 2052: Specifies the type of the wirelength estimator used during placement. For single layer architectures, cube_bb (a 3D bounding box) is always used (and is the same as per_layer_bb). For 3D architectures, cube_bb is appropriate if you can cross between layers at switch blocks, while if you can only cross between layers at output pins per_layer_bb (one bouding box per layer) is more accurate and appropriate.

vpr/src/base/read_place.cpp
```diff
@@ -306,10 +306,14 @@ void read_place_body(std::ifstream& placement_file,
  *
  * The architecture and netlist files used to generate this placement are recorded
  * in the file to avoid loading a placement with the wrong support file later.
+ *
+ * The is_initial_place bool indicates whether this is the initial placement or not. It is used to determine the
```
vaughnbetz, line 310: remove is_initial_place parameter (re-factoring: this logic should be handled in the calling routine).

vpr/src/place/centroid_move_generator.cpp
```diff
@@ -34,10 +34,12 @@ e_create_move CentroidMoveGenerator::propose_move(t_pl_blocks_to_be_moved& block
                                     placer_opts.place_dm_rlim};
 
     t_pl_loc to, centroid;
+    to.layer = from.layer;
 
     /* Calculate the centroid location*/
     calculate_centroid_loc(b_from, false, centroid, nullptr);
 
+    to.layer = (centroid.layer < 0) ? from.layer : centroid.layer;
```
vaughnbetz, line 42: Should this be an assert?  Maybe an assert_debug (moderately hot loop).

vpr/src/place/feasible_region_move_generator.cpp
```diff
@@ -33,6 +33,7 @@ e_create_move FeasibleRegionMoveGenerator::propose_move(t_pl_blocks_to_be_moved&
 
     /* Calculate the feasible region */
     t_pl_loc to;
+    to.layer = from.layer;
```
vaughnbetz, line 36: Maybe comment:  just keeping the blocks on their current layer, but maybe in the future the feasible region computation should try to figure out a good layer.

vpr/src/place/median_move_generator.cpp
```diff
@@ -240,77 +266,84 @@ static void get_bb_from_scratch_excluding_block(ClusterNetId net_id, t_bb* bb_co
  * the pins always lie on the outside of the bounding box.            *
  * The x and y coordinates are the pin's x and y coordinates.         */
 /* IO blocks are considered to be one cell in for simplicity.         */
-static bool get_bb_incrementally(ClusterNetId net_id, t_bb* bb_coord_new, int xold, int yold, int xnew, int ynew) {
+static bool get_bb_incrementally(ClusterNetId net_id, t_bb& bb_coord_new, int xold, int yold, int xnew, int ynew, int /* layer */) {
```
vaughnbetz, line 269: Watch that there are no loops that go over every terminal on a high fanout net and can be called once per move.

vaughnbetz: Delete layer and comment that this is updating only the x and y parts of the bounding box.

vpr/src/place/move_utils.cpp
```diff
@@ -1299,3 +1290,148 @@ std::string e_move_result_to_string(e_move_result move_outcome) {
...
+    } else {
+        layer_num = compatible_layers[vtr::irand(compatible_layers.size() - 1)];
+    }
+
+    return layer_num;
+}
+
+t_bb union_2d_bb(const std::vector<t_2D_bb>& bb_vec) {
+    t_bb merged_bb;
+
```
vaughnbetz, line 1335: Standard technique: 
assign first entry to xmin, xmax, etc. and then loop over rest doing min and max.

vaughnbetz: Or comment if there is some assumption / encoding of invalid bounding boxes here.

vpr/src/place/move_utils.cpp
```diff
@@ -1299,3 +1290,148 @@ std::string e_move_result_to_string(e_move_result move_outcome) {
...
+        }
+        if (merged_bb.layer_max == OPEN || layer_bb.layer_num > merged_bb.layer_max) {
+            merged_bb.layer_max = layer_bb.layer_num;
+        }
+    }
+
+    return merged_bb;
+}
+
+std::pair<t_bb, t_bb> union_2d_bb_incr(const std::vector<t_2D_bb>& num_edge_vec,
```
vaughnbetz, line 1367: What is this doing?  Needs commenting.  (I think it's keeping track of the bounding box and the number of terminals on each edge, on each layer).

vpr/src/place/move_utils.h
```diff
@@ -297,11 +296,10 @@ std::vector<t_bb> get_compressed_grid_target_search_range(const t_compressed_blo
  * @param num_layers
```
vaughnbetz, line 296: Delete num_layers param in the comments.

vpr/src/place/move_utils.h
```diff
@@ -328,6 +326,15 @@ bool intersect_range_limit_with_floorplan_constraints(t_logical_block_type_ptr t
 
 std::string e_move_result_to_string(e_move_result move_outcome);
 
+int find_free_layer(t_logical_block_type_ptr logical_block, const t_pl_loc& loc);
```
vaughnbetz, line 329: Comment all these with doxygen.

vaughnbetz: Utilities: important to comment well!

vpr/src/place/place.cpp
```diff
@@ -148,7 +148,9 @@ static vtr::NdMatrix<float, 2> chany_place_cost_fac({0, 0}); //[0...device_ctx.g
 
 /* The following arrays are used by the try_swap function for speed.   */
 /* [0...cluster_ctx.clb_nlist.nets().size()-1] */
-static vtr::vector<ClusterNetId, t_bb> ts_bb_coord_new, ts_bb_edge_new;
+static vtr::vector<ClusterNetId, t_bb> ts_bb_edge_new, ts_bb_coord_new;
```
vaughnbetz, line 151: Comment what each of these is.  ts_ is for cube_bb (used by both 2D architectures and 3D where you can switch layers at switch bloks).
layer_ts is for per layer bounding box. Used with per-layer_bb (useful when you can only switch layers at OPINs).

While a vector of 3D bounding boxes could work in both cases, it slows the single cube_bb code down by 20% or so. Hence we handle the special case of cube_bb with its own (non-vector) data structures for speed so we don't slow down the 2D case.

Comment when ts_layer_sink_pin_count is used.  (only per layer).

vaughnbetz: Also comment ts_nets_to_update.

vpr/src/place/place.cpp
```diff
@@ -252,6 +254,9 @@ std::unique_ptr<FILE, decltype(&vtr::fclose)> f_move_stats_file(nullptr,
 void print_clb_placement(const char* fname);
 #endif
 
+static bool is_cube_bb(const e_place_bounding_box_mode place_bb_mode,
```
vaughnbetz, line 257: Comment.

vpr/src/place/place.cpp
```diff
@@ -269,6 +274,8 @@ static void free_fast_cost_update();
 
 static double comp_bb_cost(e_cost_methods method);
 
+static double comp_layer_bb_cost(e_cost_methods method);
```
vaughnbetz, line 277: Comment this.

vpr/src/place/place.cpp
```diff
@@ -326,14 +348,72 @@ static float analyze_setup_slack_cost(const PlacerSetupSlacks* setup_slacks);
 
 static e_move_result assess_swap(double delta_c, double t);
 
-static void get_non_updateable_bb(ClusterNetId net_id, t_bb* bb_coord_new);
-
-static void update_bb(ClusterNetId net_id, t_bb* bb_coord_new, t_bb* bb_edge_new, int xold, int yold, int xnew, int ynew);
+static void get_non_updateable_bb(ClusterNetId net_id,
```
vaughnbetz, line 351: Comment every one of these.  Also comment why so many variations (give the big picture).

vpr/src/place/place.cpp
```diff
@@ -345,6 +425,13 @@ static void update_net_bb(const ClusterNetId net,
                           int iblk,
                           const ClusterBlockId blk,
                           const ClusterPinId blk_pin);
+
+static void update_net_layer_bb(const ClusterNetId net,
```
vaughnbetz, line 429: Comment all these new ones.

vpr/src/place/place.cpp
```diff
@@ -356,11 +443,26 @@ static void update_placement_cost_normalization_factors(t_placer_costs* costs, c
 
 static double get_total_cost(t_placer_costs* costs, const t_placer_opts& placer_opts, const t_noc_opts& noc_opts);
 
-static double get_net_cost(ClusterNetId net_id, t_bb* bb_ptr);
+static double get_net_cost(ClusterNetId net_id, const t_bb& bbptr);
```
vaughnbetz, line 446: Comment, comment, comment

vpr/src/place/place.cpp
```diff
@@ -1375,13 +1516,26 @@ static void update_move_nets(int num_nets_affected) {
...
+    int num_layers = g_vpr_ctx.device().grid.get_num_layers();
+
     for (int inet_affected = 0; inet_affected < num_nets_affected;
          inet_affected++) {
         ClusterNetId net_id = ts_nets_to_update[inet_affected];
 
-        place_move_ctx.bb_coords[net_id] = ts_bb_coord_new[net_id];
-        if (cluster_ctx.clb_nlist.net_sinks(net_id).size() >= SMALL_NET)
-            place_move_ctx.bb_num_on_edges[net_id] = ts_bb_edge_new[net_id];
+        if (num_layers == 1) {
```
vaughnbetz, line 1525: Explain why the special case (speed for 2D).

vaughnbetz (commented):
Partly done. (place.cpp).

vpr/src/place/place.cpp
```diff
@@ -578,7 +696,12 @@ void try_place(const Netlist<>& net_list,
     /* Gets initial cost and loads bounding boxes. */
 
     if (placer_opts.place_algorithm.is_timing_driven()) {
-        costs.bb_cost = comp_bb_cost(NORMAL);
+        if (num_layers == 1) {
+            costs.bb_cost = comp_bb_cost(NORMAL);
```
vaughnbetz, line 700: When you make the cost object, it's possible you could make a template wirelength cost object that automatically makes one of two types of bb_cost objects, and then you invoke the same method on it. Worth discussing with Fahri who has done similar things for the router @duck2 . But not necessary if it is hard to do or understand.

vpr/src/place/place.cpp
```diff
@@ -578,7 +696,12 @@ void try_place(const Netlist<>& net_list,
...
+        } else {
+            VTR_ASSERT_SAFE(num_layers > 1);
```
vaughnbetz, line 702: If you use cube_bb, aren't the layer = 1 and layer > 1 code paths basically the same?  If so, maybe they can be merged and this test should be on cube_bb instead?

vpr/src/place/place.cpp
```diff
@@ -733,7 +856,7 @@ void try_place(const Netlist<>& net_list,
         std::string filename = vtr::string_fmt("placement_%03d_%03d.place", 0,
                                                0);
         VTR_LOG("Saving initial placement to file: %s\n", filename.c_str());
-        print_place(nullptr, nullptr, filename.c_str());
+        print_place(nullptr, nullptr, filename.c_str(), false);
```
vaughnbetz, line 859: Get rid of "false" argument.

vpr/src/place/place.cpp
```diff
@@ -968,7 +1093,7 @@ void try_place(const Netlist<>& net_list,
         std::string filename = vtr::string_fmt("placement_%03d_%03d.place",
                                                state.num_temps + 1, 0);
         VTR_LOG("Saving final placement to file: %s\n", filename.c_str());
-        print_place(nullptr, nullptr, filename.c_str());
+        print_place(nullptr, nullptr, filename.c_str(), false);
```
vaughnbetz, line 1096: get rid of false.

vpr/src/place/place.cpp
```diff
@@ -1375,13 +1516,26 @@ static void update_move_nets(int num_nets_affected) {
...
         ClusterNetId net_id = ts_nets_to_update[inet_affected];
 
-        place_move_ctx.bb_coords[net_id] = ts_bb_coord_new[net_id];
-        if (cluster_ctx.clb_nlist.net_sinks(net_id).size() >= SMALL_NET)
-            place_move_ctx.bb_num_on_edges[net_id] = ts_bb_edge_new[net_id];
+        if (num_layers == 1) {
+            place_move_ctx.bb_coords[net_id] = ts_bb_coord_new[net_id];
+        } else {
+            place_move_ctx.layer_bb_coords[net_id] = layer_ts_bb_coord_new[net_id];
+            place_move_ctx.num_sink_pin_layer[net_id] = ts_layer_sink_pin_count[net_id];
```
vaughnbetz, line 1529: Does cube_bb need a bounding box per layer, or couldn't you just update one 3D bb (and use the other, non-vector, data structure)?

vpr/src/place/place.cpp
```diff
@@ -1375,13 +1516,26 @@ static void update_move_nets(int num_nets_affected) {
...
+        }
+
+        if (cluster_ctx.clb_nlist.net_sinks(net_id).size() >= SMALL_NET) {
+            if (num_layers == 1) {
+                place_move_ctx.bb_num_on_edges[net_id] = ts_bb_edge_new[net_id];
+            } else {
+                place_move_ctx.layer_bb_num_on_edges[net_id] = layer_ts_bb_edge_new[net_id];
```
vaughnbetz, line 1536: It seems you might have done cube_bb in a bit of an odd way, where you compute a bb per layer and then go through and compute an overall bb.  More in common with the 2D case and probably a bit faster to just directly compute one 3D bounding box, which I think the scalar ts_bb* data structures can do.

vpr/src/place/place.cpp
```diff
@@ -2480,15 +2789,131 @@ static void get_bb_from_scratch(ClusterNetId net_id, t_bb* coords, t_bb* num_on_
...
+    num_on_edges.xmax = xmax_edge;
+    num_on_edges.ymin = ymin_edge;
+    num_on_edges.ymax = ymax_edge;
+}
+
+/* This routine finds the bounding box of each net from scratch (i.e.   *
+ * from only the block location information).  It updates both the       *
+ * coordinate and number of pins on each edge information.  It           *
+ * should only be called when the bounding box information is not valid. */
+static void get_layer_bb_from_scratch(ClusterNetId net_id,
```
vaughnbetz, line 2807: Should say this is for multi-layer bounding boxes.

vpr/src/place/place.cpp
```diff
@@ -2480,15 +2789,131 @@ static void get_bb_from_scratch(ClusterNetId net_id, t_bb* coords, t_bb* num_on_
...
+                                      std::vector<t_2D_bb>& num_on_edges,
+                                      std::vector<t_2D_bb>& coords,
+                                      std::vector<int>& layer_pin_sink_count) {
+    auto& device_ctx = g_vpr_ctx.device();
+    const int num_layers = device_ctx.grid.get_num_layers();
+    num_on_edges.resize(num_layers, t_2D_bb());
```
vaughnbetz, line 2813: Why is there a resize in an inner loop?  This should have been allocated to the right size long ago.

vpr/src/place/place.cpp
```diff
@@ -2480,15 +2789,131 @@ static void get_bb_from_scratch(ClusterNetId net_id, t_bb* coords, t_bb* num_on_
...
+    coords.resize(num_layers, t_2D_bb());
```
vaughnbetz, line 2814: Another resize. Why?

vpr/src/place/place.cpp
```diff
@@ -2480,15 +2789,131 @@ static void get_bb_from_scratch(ClusterNetId net_id, t_bb* coords, t_bb* num_on_
...
+    layer_pin_sink_count.resize(num_layers, 0);
+    std::vector<int> xmin(num_layers, OPEN);
```
vaughnbetz, line 2816: Constructing these vectors will take some time.  I don't think you should be doing this.  Just directly compute the results in the layer_bbs, saving vector creation and destruction and value copying.

vpr/src/place/place.cpp
```diff
@@ -2519,14 +2944,46 @@ static double get_net_wirelength_estimate(ClusterNetId net_id, t_bb* bbptr) {
...
+    double ncost = 0.;
+    double crossing = 0.;
+    int num_layers = g_vpr_ctx.device().grid.get_num_layers();
+
+    for (int layer_num = 0; layer_num < num_layers; layer_num++) {
+        VTR_ASSERT(layer_pin_sink_count[layer_num] != OPEN);
+        if (layer_pin_sink_count[layer_num] == 0) {
+            continue;
+        }
+        crossing = wirelength_crossing_count(layer_pin_sink_count[layer_num] + 1);
```
vaughnbetz, line 2969: This may be overkill; if it is making you keep per-layer bb data and sink counts when computing a 3D bounding box, you could simplify it to use the crossing count based on the fanout of the whole net.  It will make little difference since most nets are low fanout, and it isn't clear which approach is more accurate anyway.

vpr/src/place/place.cpp
```diff
@@ -2611,13 +3103,88 @@ static void get_non_updateable_bb(ClusterNetId net_id, t_bb* bb_coord_new) {
...
-static void update_bb(ClusterNetId net_id, t_bb* bb_coord_new, t_bb* bb_edge_new, int xold, int yold, int xnew, int ynew) {
+static void get_non_updateable_layer_bb(ClusterNetId net_id,
+                                        std::vector<t_2D_bb>& bb_coord_new,
+                                        std::vector<int>& num_sink_layer) {
+    //TODO: account for multiple physical pin instances per logical pin
+
+    auto& device_ctx = g_vpr_ctx.device();
+    int num_layers = device_ctx.grid.get_num_layers();
+    num_sink_layer = std::vector<int>(num_layers, 0);
+    std::vector<int> xmin(num_layers, OPEN);
```
vaughnbetz, line 3120: Why the xmin, ymin temporaries again?  Unnecessary allocations, deletes and copying.

vpr/src/place/place.cpp
```diff
@@ -2636,177 +3203,525 @@ static void update_bb(ClusterNetId net_id, t_bb* bb_coord_new, t_bb* bb_edge_new
...
+        bb_edge_new.ymin = curr_bb_edge->ymin;
+        bb_edge_new.ymax = curr_bb_edge->ymax;
     }
 
     if (bb_updated_before[net_id] == NOT_UPDATED_YET) {
         bb_updated_before[net_id] = UPDATED_ONCE;
     }
 }
 
+static void update_layer_bb(ClusterNetId net_id,
```
vaughnbetz, line 3377: These new *layer* routines are under-commented; there isn't any overall comment talking about the per-layer approach vs. the 3D approach -- there should be one.

vpr/src/place/place.cpp
```diff
@@ -2636,177 +3203,525 @@ static void update_bb(ClusterNetId net_id, t_bb* bb_coord_new, t_bb* bb_edge_new
...
+            bb_edge_new[layer_num].ymax = 1;
+            bb_coord_new[layer_num].ymax = y_new;
+        } else if (y_new == curr_bb_coord[layer_num].ymax) {
+            bb_edge_new[layer_num].ymax = curr_bb_edge[layer_num].ymax + 1;
+            bb_coord_new[layer_num].ymax = curr_bb_coord[layer_num].ymax;
+        }
+    }
+}
+
+static inline void update_bb_layer_changed(ClusterNetId net_id,
```
vaughnbetz, line 3578: It seems like update_bb_layer_changed and update_bb_same_layer might be mergable if you wrote the code as "remove from old location" and "insert in new location". But maybe that's less efficient; just something to think about ...

vpr/src/place/place.cpp
```diff
@@ -2636,177 +3203,525 @@ static void update_bb(ClusterNetId net_id, t_bb* bb_coord_new, t_bb* bb_edge_new
...
+    }
+
+    add_block_to_bb(pin_new_loc,
+                    curr_bb_edge[new_layer_num],
+                    curr_bb_coord[new_layer_num],
+                    bb_edge_new[new_layer_num],
+                    bb_coord_new[new_layer_num]);
+}
+
+static void update_bb_pin_sink_count(ClusterNetId /* net_id */,
```
vaughnbetz, line 3653: VB to here.

vpr/src/place/simpleRL_move_generator.h
```diff
@@ -211,10 +212,12 @@ class SimpleRLMoveGenerator : public MoveGenerator {
      *
      *   @param agent std::unique_ptr to the agent. Only EpsilonGreedyAgent and SoftmaxAgent types are accepted
      *   by the constructor. If other types are passed, a compile error would be thrown.
+     *
+     *   @param is_multi_layer A boolean value to indicate whether the placement is multi-layer or not
      */
     template<class T,
              class = typename std::enable_if<std::is_same<T, EpsilonGreedyAgent>::value || std::is_same<T, SoftmaxAgent>::value>::type>
-    explicit SimpleRLMoveGenerator(std::unique_ptr<T>& agent);
+    explicit SimpleRLMoveGenerator(std::unique_ptr<T>& agent, bool is_multi_layer);
```
vaughnbetz, line 220: Do you need to pass is_multi_layer around, or could you store it in the architecture or placement context?

vaughnbetz (commented):
A bit more ...

vpr/src/route/router_lookahead_map.cpp
```diff
@@ -255,6 +255,11 @@ static void store_min_cost_to_sinks(std::unordered_map<int, std::unordered_map<i
  */
 static void min_global_cost_map(vtr::NdMatrix<util::Cost_Entry, 3>& internal_opin_global_cost_map);
 
+static std::pair<float, float> get_cost_from_src_opin(const std::map<int, util::t_reachable_wire_inf>& src_opin_delay_map,
```
vaughnbetz, line 258: Add comment (use doxygen format)

vpr/src/route/router_lookahead_map.cpp
```diff
@@ -295,6 +300,10 @@ static void print_wire_cost_map(int layer_num, const std::vector<t_segment_inf>&
 static void print_router_cost_map(const t_routing_cost_map& router_cost_map);
 
 /******** Interface class member function definitions ********/
+MapLookahead::MapLookahead(const t_det_routing_arch& det_routing_arch, bool is_flat)
```
vaughnbetz, line 303: Comment

vpr/src/route/router_lookahead_map.cpp
```diff
@@ -444,66 +454,31 @@ std::pair<float, float> MapLookahead::get_expected_delay_and_cong(RRNodeId from_
...
-
-                float this_delay_cost = (params.criticality) * (reachable_wire_inf.delay + wire_cost_entry.delay);
-                float this_cong_cost = (1. - params.criticality) * (reachable_wire_inf.congestion + wire_cost_entry.congestion);
-
-                expected_delay_cost = std::min(expected_delay_cost, this_delay_cost);
-                expected_cong_cost = std::min(expected_cong_cost, this_cong_cost);
-            }
+        // Currently, we assume inter-layer connections are only from a block output pin to another layer. Thus, if the from and to layers are different,
+        // We use src_opin_inter_layer_delays.
+        if (from_layer_num == to_layer_num) {
```
vaughnbetz, line 467: Is the if statement for speed?  Otherwise a degenerate look-up with both the from to layer numbers being the same would work.  If there's no speed advantage, I suggest combining the code (always go through the more general data structure) to minimize the number of code paths.

vpr/src/route/router_lookahead_map_utils.cpp
```diff
@@ -535,12 +563,20 @@ static void dijkstra_flood_to_wires(int itile, RRNodeId node, util::t_src_opin_d
...
-            if (!src_opin_delays[node_layer_num][itile][ptc].count(seg_index)
-                || curr.delay < src_opin_delays[node_layer_num][itile][ptc][seg_index].delay) {
+            if ((!src_opin_delays[node_layer_num][itile][ptc].count(seg_index)
+                 || curr.delay < src_opin_delays[node_layer_num][itile][ptc][seg_index].delay)
+                && curr_layer_num == node_layer_num) {
                 src_opin_delays[node_layer_num][itile][ptc][seg_index].wire_rr_type = curr_rr_type;
                 src_opin_delays[node_layer_num][itile][ptc][seg_index].wire_seg_index = seg_index;
                 src_opin_delays[node_layer_num][itile][ptc][seg_index].delay = curr.delay;
                 src_opin_delays[node_layer_num][itile][ptc][seg_index].congestion = curr.congestion;
+            } else if (is_multi_layer && (!src_opin_inter_layer_delays[node_layer_num][itile][ptc][curr_layer_num].count(seg_index) || curr.delay < src_opin_inter_layer_delays[node_layer_num][itile][ptc][curr_layer_num][seg_index].delay)
```
vaughnbetz, line 573: Comment what this else if for. What's the high-level thinking?

vpr/src/route/router_lookahead_map_utils.h
```diff
@@ -288,7 +291,8 @@ typedef std::vector<std::unordered_map<int, Cost_Entry>> t_ipin_primitive_sink_d
 // and the tile's IPIN. If there are many connections to the same IPIN, the one with the minimum delay is selected.
 typedef std::vector<std::vector<std::vector<t_reachable_wire_inf>>> t_chan_ipins_delays;
 
-t_src_opin_delays compute_router_src_opin_lookahead(bool is_flat);
+std::pair<t_src_opin_delays, t_src_opin_inter_layer_delays> compute_router_src_opin_lookahead(bool is_flat);
```
vaughnbetz, line 294: Add comment if you can.

vpr/src/util/vpr_utils.cpp
```diff
@@ -2508,3 +2508,21 @@ void add_pb_child_to_list(std::list<const t_pb*>& pb_list, const t_pb* parent_pb
         }
     }
 }
+
+float get_min_cross_layer_delay(const std::vector<t_arch_switch_inf>& arch_switch_inf,
+                                const std::vector<t_segment_inf>& segment_inf,
+                                const int wire_to_ipin_arch_sw_id) {
+    float min_delay = std::numeric_limits<float>::max();
+
+    if (wire_to_ipin_arch_sw_id != OPEN) {
```
vaughnbetz, line 2517: Explain the logic here in a comment.

vpr/src/util/vpr_utils.h
```diff
@@ -311,4 +311,8 @@ t_arch_switch_inf create_internal_arch_sw(float delay);
 
 void add_pb_child_to_list(std::list<const t_pb*>& pb_list, const t_pb* parent_pb);
 
+float get_min_cross_layer_delay(const std::vector<t_arch_switch_inf>& arch_switch_inf,
```
vaughnbetz, line 314: Comment what this does.  (Also good to comment why it is useful, if you can).

vtr_flow/arch/multi_die/aman_3d_coffe.xml
```diff
@@ -0,0 +1,1597 @@
+<architecture>
```
vaughnbetz, line 1: Add comment to top of architecture. Where did it come from, what's it for, any caveats?
(Can add a reference to the fpt paper if it has info on this architecture).
Rename to something other than aman (describe what it is, not who wrote it).
Does his have delays between dice?  If not, update it to have some reasonable switch delay between dice.

vtr_flow/arch/multi_die/aman_3d_limited.xml
```diff
@@ -0,0 +1,1579 @@
+<architecture>
```
vaughnbetz, line 1: Add comment to top of architecture. Where did it come from, what's it for, any caveats?
Add a reference to the fpt paper as it definitely has info on the limited out put architecture, and results for it.
Rename to something other than aman (describe what it is, not who wrote it).
Does his have delays between dice? If not, update it to have some reasonable switch delay between dice.

vaughnbetz (changes requested):
Done my review.

Lots of additional commenting needed.
Some code paths / data structures that have two versions (2D and 3D) may be mergeable, and that is preferable where possible to minimize code size and complexity. See my detailed comments for some ideas.
There are some temporary vectors in hot bounding box loops that should be eliminated for speed (see detailed comments).
There are a lot of variations on the bounding box routines (2d, 3d, per layer, incremental, from scratch, ...). Can any be merged / eliminated?
Check if the 3D bounding box code for multi layers is as incremental as possible; I think there may be some unnecessary code splitting between 2D and 3D when cube_bb is used, and we seem to be keeping some per layer data in the 3D case with cube_bb that seems unnecessary.

vpr/src/place/place.cpp
```diff
@@ -356,11 +443,26 @@ static void update_placement_cost_normalization_factors(t_placer_costs* costs, c
...
+                                t_bb& num_on_edges);
 
-static void get_bb_from_scratch(ClusterNetId net_id, t_bb* coords, t_bb* num_on_edges);
+static void get_layer_bb_from_scratch(ClusterNetId net_id,
+                                      std::vector<t_2D_bb>& num_on_edges,
+                                      std::vector<t_2D_bb>& coords,
+                                      std::vector<int>& layer_pin_sink_count);
 
-static double get_net_wirelength_estimate(ClusterNetId net_id, t_bb* bbptr);
+static double get_net_wirelength_estimate(ClusterNetId net_id, const t_bb& bbptr);
```
duck2, line 461: `bbptr` doesn't look like a ptr/output reference to me, so probably better to call it `bb`. Applies to all cases of `bbptr`, even when it's a ptr, since that would be visible from the type/usage.

vpr/src/place/place.cpp
duck2: Probably better to make every bb a cube. I saw a comment about it becoming 20% slower though? No idea why that would happen...

With the current version, I don't think there's a point templating here: there are only 2 cases and the code is organized as non-class fns so it would be a pain to propagate all the templates.

However as it's the case with `is_flat` I think it's best to hoist `cube_bb` in the global context as soon as possible to avoid adding a param to every fn. That would simplify things a lot

vpr/src/place/place.cpp
```diff
@@ -1746,6 +1901,33 @@ static e_move_result try_swap(const t_annealing_state* state,
     return move_outcome;
 }
 
+static bool is_cube_bb(const e_place_bounding_box_mode place_bb_mode,
+                       const RRGraphView& rr_graph) {
+    bool cube_bb;
+    const int number_layers = g_vpr_ctx.device().grid.get_num_layers();
```
duck2, line 1910: Does `place_bb_mode` change during a placer run? If not, it's best to calculate and hoist something like `g_vpr_ctx.placement().use_cube_bb` as soon as possible instead of having this fn and the extra parameters in placer fns.

If `place_bb_mode` can change while the placer runs (I don't think it does? seems to come from `placer_opts`), it would be better to precompute `inter_layer_...` (looks like it's a property of the RRG) and then expose `is_cube_bb` as a global inline/constexpr fn.

vpr/src/place/place.cpp
amin1377: I will comment the functions related to calculating net cost in the other PR. In that PR, I have moved all these functions in another file.

vpr/src/place/place.cpp
amin1377: In another PR...

vpr/src/place/place.cpp
amin1377: In another PR...

vpr/src/place/place.cpp
amin1377: :))

vpr/src/place/median_move_generator.cpp
```diff
@@ -5,9 +5,9 @@
 #include "placer_globals.h"
 #include "move_utils.h"
 
-static bool get_bb_incrementally(ClusterNetId net_id, t_bb* bb_coord_new, int xold, int yold, int xnew, int ynew);
+static bool get_bb_incrementally(ClusterNetId net_id, t_bb& bb_coord_new, int xold, int yold, int xnew, int ynew, int layer);
```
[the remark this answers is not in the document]
amin1377: The function in "place.cpp" that updates the bounding box incrementally is named "update_bb". We cannot reuse this function because it checks specific static variables within the "place.cpp" file to determine whether the bounding box has already been updated or not.

vpr/src/place/median_move_generator.cpp
amin1377: I have rechecked and ensured that I am not iterating over all terminals. In this move, I have implemented the use of a 3D bounding box, regardless of the type chosen by the user, to simplify the process (The bounding box type specified by the user is utilized by the WL estimator.)
Since this implementation relies on the bounding box stored in the placement context, I need to create a 3D bounding box if the type stored there is "per-layer." This is why the function "union_2d_bb_incr" is called when the bounding box type is "per-layer." In that function, I iterate over all layers, not terminals, to create a 3D bounding box from the "per-layer" bounding box.

vpr/src/place/place.cpp
amin1377: These comments are added to the next PR which I moved the net cost, etc. into another file

vpr/src/place/place.cpp
amin1377: Thanks, Vaughn and Fahri. I relocated the `cube_bb` to the global context. If I'm able to encapsulate them within a class, I believe templating it would be a cleaner solution than duplicating all the code.

Regarding the performance slowdown with per-layer BB, the issue arises because we are essentially converting all integer variables in cost functions to vectors. These functions reside inside the inner loop of the placement process. I've also profiled the usage of "per-layer" BB for the 2D architecture (which basically means have a vector of size on instead of an integer) and confirmed that the slowdown is caused by the functions responsible for updating the BB.

vpr/src/place/place.cpp
amin1377: I am storing the number of sinks that each net has on each layer. The computation doesn't require many operations. For low fanout nets, we already iterate over all sinks, and for high-fanout ones, I update it incrementally.

vpr/src/place/place.cpp
amin1377: The reason for creating `xmin`, `ymin`, etc. is because I need to include the source block in all layers, so I initialize them with the source block coordinates. Later, if there is no sink on a particular layer, I want to assign `OPEN` to the bounding box coordinates of that layer. Additionally, at the end of the function, I perform additional checks to ensure the bounding box coordinates are inside the grid. These reasons led me to create temporary values to obtain the bounding box coordinates on each layer and store the valid coordinates in the main data structure.

vpr/src/place/place.cpp
amin1377: I will have this overall comments in the next PR that I have done the refactoring

vpr/src/place/placer_context.h
```diff
@@ -91,11 +91,20 @@ struct PlacerRuntimeContext : public Context {
...
 
     // [0..cluster_ctx.clb_nlist.nets().size()-1]. Store the number of blocks on each of a net's bounding box (to allow efficient updates)
-    vtr::vector<ClusterNetId, t_bb> bb_num_on_edges;
+    vtr::vector<ClusterNetId, std::vector<t_2D_bb>> layer_bb_num_on_edges;
+
+    // [0..cluster_ctx.clb_nlist.nets().size()-1]. Store the bounding box coordinates of a net's bounding box
+    vtr::vector<ClusterNetId, std::vector<t_2D_bb>> layer_bb_coords;
+
+    // [0..cluster_ctx.clb_nlist.nets().size()-1]. Store the number of blocks on each layer ()
+    vtr::vector<ClusterNetId, std::vector<int>> num_sink_pin_layer;
```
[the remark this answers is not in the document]
amin1377: This is data structure keeps track of how many sinks each net have on each layer

vpr/src/place/uniform_inter_layer_move_generator.h
```diff
@@ -0,0 +1,18 @@
+#ifndef VTR_UNIFORM_INTER_LAYER_MOVE_GENERATOR_H
+#define VTR_UNIFORM_INTER_LAYER_MOVE_GENERATOR_H
+
+#include "move_generator.h"
+#include "timing_place.h"
+
+/**
+ * @brief Uniform inter-layer move generator
+ *
+ * randomly picks a from_block with equal probabilities for all blocks, and then moves it randomly within
```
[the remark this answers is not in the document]
amin1377: We decided to not use this new move; thus, I removed the files related to it.

vpr/src/place/weighted_median_move_generator.cpp
```diff
@@ -72,6 +77,15 @@ e_create_move WeightedMedianMoveGenerator::propose_move(t_pl_blocks_to_be_moved&
         place_move_ctx.X_coord.insert(place_move_ctx.X_coord.end(), ceil(coords.xmax.criticality * CRIT_MULT_FOR_W_MEDIAN), coords.xmax.edge);
         place_move_ctx.Y_coord.insert(place_move_ctx.Y_coord.end(), ceil(coords.ymin.criticality * CRIT_MULT_FOR_W_MEDIAN), coords.ymin.edge);
         place_move_ctx.Y_coord.insert(place_move_ctx.Y_coord.end(), ceil(coords.ymax.criticality * CRIT_MULT_FOR_W_MEDIAN), coords.ymax.edge);
+        if (is_multi_layer) {
+            for (int layer_num = 0; layer_num < num_layers; layer_num++) {
+                layer_blk_cnt[layer_num] += place_move_ctx.num_sink_pin_layer[net_id][layer_num];
+            }
+            if(cluster_ctx.clb_nlist.pin_type(pin_id) != PinType::DRIVER) {
+                VTR_ASSERT(layer_blk_cnt[from.layer] > 0);
+                layer_blk_cnt[from.layer]--;
```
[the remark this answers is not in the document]
amin1377: We shouldn't consider the moving block when calculating the median location. Therefore, if a pin of the block that is about to move is of type sink, I will decrement the number of sinks on that layer to exclude that pin from the calculation.

vpr/src/place/place.cpp
vaughnbetz: You will pay a runtime price for this, so if it's a key loop it's worth thinking about whether you need the vector temporaries or could recode without them.

amin1377: I will attempt to rewrite it without utilizing these temporary vectors. Although I initially chose this approach for its cleanliness, I agree that given the inner loop context, it is worth experimenting with alternatives.

vpr/src/route/connection_router.cpp
```diff
@@ -5,6 +5,40 @@
 #include "bucket.h"
 #include "rr_graph_fwd.h"
 
+static inline bool has_path_to_sink(const t_rr_graph_view& rr_nodes,
```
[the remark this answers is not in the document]
amin1377: When inter-layer connections are solely from OPINs, there's no need to push back the routing path for a connection on another layer into the heap. This function determines whether the node, about to be pushed back into the heap, has an actual path to the sink. For instance, if inter-layer connections are only from OPINs and the node under consideration is of type CHANX/Y, it would only return true if that node is on the same layer as SINK.

vpr/src/route/router_lookahead_extended_map.h
```diff
@@ -19,6 +19,9 @@ class ExtendedMapLookahead : public RouterLookahead {
     ///<Look-up table from SOURCE/OPIN to CHANX/CHANY of various types
     util::t_src_opin_delays src_opin_delays;
 
+    ///< Lookup table from SOURCE/OPIN to CHANX/CHANY of the another layer
```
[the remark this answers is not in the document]
amin1377: In the update to the router lookahead (to support inter-layer connection from SW block) this data structure is deleted since there will be no use to it.

vpr/src/route/router_lookahead_map.cpp
amin1377: That is for speed. In the updated router lookahead (6d array) a general data structure is used.

vpr/src/route/router_lookahead_map_utils.cpp
amin1377: This is gone in the 6d router lookahead PR.

vtr_flow/arch/multi_die/aman_3d_coffe.xml
amin1377: I have to fix a bunch of 3D architectures and add to them. Also, in another branch, I am working on developing some more tests on 3D architecture too. These comments are added on that branch.

vtr_flow/arch/multi_die/aman_3d_limited.xml
amin1377: I have to fix a bunch of 3D architectures and add to them. Also, in another branch, I am working on developing some more tests on 3D architecture too. These comments are added on that branch.

vpr/src/place/place.cpp
amin1377: Done in the PR that I refactored the place.cpp file

vpr/src/place/place.cpp
amin1377: You are correct; it doesn't change during placement. The issue arises when the `cube_bb` type is set to `auto`; in this case, I iterate over the RR graph to ensure all inter-layer connections are from OPINs (I don't rely on the architecture file to be able to support custom RR graph). To address this, I moved the `cube_bb` to the placement context and set it at the beginning of placement. Do you have any suggestions for an alternative approach?

vaughnbetz: I think that's OK and matches Fahri's suggestion (set as soon as possible, which in this case is at the start of placement).

vaughnbetz:
Needs format.
Try matrix.
Get QoR.
Post this PR: could try grabbing some cpu time by tuning SMALL_NET for incremental BB, and by defining a new fanout threshold at which we ignore bb costs/computations (would be a background task).

amin1377:
QoR:
The first column represents the master branch. In the second column, the data structure storing the number of sinks on each layer per net is implemented as a vector of arrays. The third column represents the case where this structure is a 2D matrix. Considering that a 2D matrix offers slight improvements, I have modified the implementation to use it.
Screenshot from 2023-11-10 09-38-05

vaughnbetz:
Great!  Matrix it is (more maintainable as its a dynamic data structure, and faster too).

amin1377:
@vaughnbetz This PR is ready to merge.

PR #2349, opened 2023-07-19
RR Graph Reader - Backward Compatiblity

amin1377:
In this pull request, I have made changes to the RR Graph reader to enable it to read RR graph files that do not specify the layer number of nodes and grid.

After updating VPR to support 3D architecture, the layer number is now included in both the RR node description and the grid description of the XML file. However, some RR graph files generated by older runs may not have this information. As a result, if the reader cannot find the layer number, it would raise an error, causing issues for users using these older RR graph files.

With the updated RR Graph reader, we ensure that it can handle both cases seamlessly, allowing users to work with RR graph files from both older and newer runs without encountering any errors.

### DIFF
--- a/libs/librrgraph/src/io/gen/rr_graph_uxsdcxx.h
+++ b/libs/librrgraph/src/io/gen/rr_graph_uxsdcxx.h
@@ -6,7 +6,7 @@
  *
  * Cmdline: uxsdcxx/uxsdcxx.py /home/amin/vtr-verilog-to-routing/libs/librrgraph/src/io/rr_graph.xsd
  * Input file: /home/amin/vtr-verilog-to-routing/libs/librrgraph/src/io/rr_graph.xsd
- * md5sum of input file: 8672cb3951993f7e0ea3433a02507672
+ * md5sum of input file: 8a13fa50b87a91c2baab7c6ced02f573
  */
 
 #include <functional>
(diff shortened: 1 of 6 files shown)


amin1377:
@vaughnbetz : Files under `libs/librrgraph/src/io/gen` and `libs/libvtrcapnproto/gen` are generated automatically.

vaughnbetz:
Looks good.  It looks like you added some more error output too, which is nice.

One spurious QoR failure in vtr_reg_nightly_test1: k6_frac_2ripple_N8_22nm.xml/fir_pipe_15.v/common min_chan_width_route_time relative value 15.092783505154639 outside of range [0.1,14.0], above absolute threshold 2.0 and not equal to golden value: 2.91
(small design, min chan width route time varies a lot, so ignore).

vaughnbetz:
Looks good -- the one failure is a spurious min width route time cpu time qor failure on a small design (OK).  
Can you resolve the conflict so we can merge?

vaughnbetz:
@amin1377 : needs conflicts resolved and then should pass CI.  If we get a spurious failure on min_width_route_time on small designs I suggest we expand the legal range to [0.1 to 16] for the failing group of (small) tests (or so) as we have had failures like this with moderate frequency.

amin1377:
@vaughnbetz: CI tests are green.

PR #2353, opened 2023-07-24
Fix Seg Fault error when reading in an RR Graph with different x-y segment types

kimiatkh:
Description

The pull request is to fix the segmentation fault that occurred when reading in the rr graph that had different segment types in the X and Y directions. The error was caused by some parts of the code that set the node cost index, which still relied on the assumption that the X and Y axes have the same segment types. The issue has been fixed by updating the relevant code.

How Has This Been Tested?

The changes have been tested using the Stratix 10 architecture. Before the fix, the rr graph could not be read in properly, resulting in the segmentation fault. After applying the fix, the issue has been resolved, and the rr graph is now correctly processed without any errors.

#### Types of changes

#### Checklist:


vaughnbetz:
Looks good, thanks.

PR #2364, opened 2023-08-10
Read Tileable rr_graph back into the flow

saaramahmoudi:
Since openfpga branch and vpr has different ptc number assignment, when we try to read back rr-graph generated with tileable=true, we get a node mismatch since some data structures are over-written in the process. This PR changes the read functions and try to resolve this mismatch. 

#### Types of changes


alaindargelas:
@tangxifan @ganeshgore can one of you review and approve this PR?

ganeshgore (approved):

ganeshgore:
I reviewed the code it looks good to me. Sara also tried this locally it results in the same netlist and GSB generation. 
@tangxifan  Can you review and merge, I am planning to make PR with the test task/CI on the OpenFPGA repo after this.

tangxifan:
> I reviewed the code it looks good to me. Sara also tried this locally it results in the same netlist and GSB generation. @tangxifan Can you review and merge, I am planning to make PR with the test task/CI on the OpenFPGA repo after this.

Ok. Will do it today.

libs/librrgraph/src/base/rr_graph_builder.cpp
```diff
@@ -37,12 +37,36 @@ vtr::vector<RRNodeId, std::vector<short>>& RRGraphBuilder::node_ptc_storage() {
...
     t_rr_type node_type = node_storage_.node_type(node);
-    short node_ptc_num = node_storage_.node_ptc_num(node);
     short node_layer = node_storage_.node_layer(node);
+    short node_ptc_base = node_storage_.node_ptc_num(node);
+    Direction node_dir = (node_type == CHANX || node_type == CHANY) ? node_storage_.node_direction(node): Direction::NONE;
+
+    int offset = 0;
     for (int ix = node_storage_.node_xlow(node); ix <= node_storage_.node_xhigh(node); ix++) {
         for (int iy = node_storage_.node_ylow(node); iy <= node_storage_.node_yhigh(node); iy++) {
+            short node_ptc_num = calculate_node_ptc(graph_type, node_ptc_base, node_dir , offset);
```
tangxifan, line 68: This may be dangerous, potentially leading to some issues in OpenFPGA, e.g., netlist error. I need to double check.

ganeshgore: Can you list the scenarios where this might fail? 
Certainly, there is a scope for improvement, but for now, it generated exact same netlist for a couple of architectures I tested.

alaindargelas:
Can someone retrigger also the failed Action? There is only one failed action, but just to be sure.

saaramahmoudi:
> Can someone retrigger also the failed Action? There is only one failed action, but just to be sure.

Done

libs/librrgraph/src/base/rr_graph_builder.cpp
tangxifan:
```
grep "add_node_to_all_locs" */*.cpp ../../vtr-verilog-to-routing/vpr/src/*/*.cpp ../../vtr-verilog-to-routing/libs/librrgraph/src/*/*.cpp
../../vtr-verilog-to-routing/vpr/src/route/clock_network_builders.cpp:    /* TODO: Will replace these codes with an API add_node_to_all_locs() of RRGraphBuilder */
../../vtr-verilog-to-routing/vpr/src/route/clock_network_builders.cpp:    /* TODO: Will replace these codes with an API add_node_to_all_locs() of RRGraphBuilder */
../../vtr-verilog-to-routing/libs/librrgraph/src/base/rr_graph_builder.cpp:void RRGraphBuilder::add_node_to_all_locs(RRNodeId node) {
```

Actually, the API ``add_node_to_all_locs()`` is not used by tileable rr graph builder. So the changes here may not work on tileable rr graph builder. Can you clarify your motivation? If it is just for the native rr graph builder of VPR?

tangxifan (changes requested):
@saaramahmoudi 
Actually, in tileable routing resource builder, I use API ``add_node_track_num()``
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/ec634df07a5be186a0e7b66f2789c4daf2e4b44c/vpr/src/tileable_rr_graph/tileable_rr_graph_node_builder.cpp#L772

to keep tracking the the track ids of staggered routing wires. 
You can see that I have reworked the storage of ptc_num for rr_nodes

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/ec634df07a5be186a0e7b66f2789c4daf2e4b44c/libs/librrgraph/src/base/rr_graph_builder.cpp#L236-L251

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/ec634df07a5be186a0e7b66f2789c4daf2e4b44c/libs/librrgraph/src/base/rr_graph_builder.h#L415-L432

So my suggestion is to avoid creating your own function that manage/calculate track ids. In tileable routing resource graph, it is pre-calculated when building routing channels. 

For the I/O support, I believe you just need to output all the track num to file and read these track num from files.

libs/librrgraph/src/base/rr_graph_builder.cpp
saaramahmoudi: The add_node_to_all_locs() belongs to the process that tries to read rr_graph back in not while generating it. This function works fine while you try to read an rrgraph generated without tileable option. But if you have tileable = true in the architecture and generate an rrgraph, you can't use --read_rr_graph anymore with the generated rr graph, you will get a node mismatch error since some data structure in that process don't know about ptc assignment in tileable true. Hence, the update goes to reading the rr graph functions not the one that are generating it. 
I am not sure why you think it wouldn't work on tileable rr graph builder; since we have generated many netlists and GSB files with tileable=true and checked the rr graph and read it back in with the same option. If you clarify your question, I can probably give you a better answer.

tangxifan: @saaramahmoudi I see the pain point: you have no track information when reading a rr_graph file because the track information is built on-fly during rr_graph builder functions. 
I think it is better not to do the calculation in the I/O as it may bring more restrictions in future. My opinion is that such calculation should be handled by the engineer which crafts rr graph files. I/O functions just need to do read-and-assign or get-and-write jobs. 
I believe a better way is to extend the writer of rr graph files which can output a list of track numbers. And we just read back these numbers through reader.

I believe your current changes may work on some specific testcases. However, since these testcases are not open to me, I can judge based on the code changes that I see. If you can more provide some testcases or try some existing testcases on OpenFPGA. It will be a better proof.

saaramahmoudi: @tangxifan Your suggestion sounds good to me. However, if we change the write functions, we may ended up not supporting existing rr-graphs for both vpr baseline and openfpga (tileable = true) options. 
I almost test all of circuit and architectures available in openfpga_flow with new function. Plus, VTR CI tests have some example of reading rr graph as well (like symbiflow test case) and those are also passing, meaning that baseline is working as expected.  
If you have any testcase in mind, I would be happy to test it out or you can think of test that It won't work, please let me know

tangxifan: @saaramahmoudi The changes on writer function should be fairly straightforward. 

image

- I/O function can accept a rr_node with multiple ptc number. No checks will be enforced for now (unless we see some demands). For instance, ``ptc="1,2,3,4"``
- In regular routing resource graph, there is only one ptc number. So this will not change anything. When parsing, you can assign the same ptc number to all the node coordinates. For example, ``ptc="1"``
- When parsing tileable routing resource graph, there could be a number of ptc number. They will be all parsed and assigned, just like what you did in the ``add_node_to_all_locs()``.

I agree that your codes change may work in some architectures and netlists. However, it may force restrictions when building routing resource graphs. When we design the ``RRGraphView`` and ``RRGraphBuilder``, we made the APIs as doer functions, which means they just do as builder functions want. These APIs should be general purpose, while builder functions do the calculation and customization. Similarly, I/O functions are also doer functions. It should be user who decide how their routing resource graphs look like.

I have suggested some tests in my reply to @ganeshgore . Feel free to try.

ganeshgore: I agree with @tangxifan here, it is better to change the rrgraph writer so that the generated rrgraph doesn't have any special properties/interpretation and it is entirely compatible with the rrgraph generated by the VPR.

PR #2378, opened 2023-08-30
Added Optional Attribute to RRGraph 

saaramahmoudi:
#### Description
Using tileable option with openfpga branch, we have different ptc assignment in CHANX/CHANY nodes. To generate the same format rr graph while having tileable="true" and tileable="false", We added a new tag to <node> tag while writing the rr graph.

The picture below shows the difference in ptc assignment in CHANX/CHANY nodes. The ptc number for each part of wire increase linearly with starting point ptc number (0) and a twist number (in this case would be 1) 

image

So, the new generated rr graph looks like this:
`<node dir="" id="" type="CHANX"><loc ptc="0" twist="1" layer="" xlow="" xhigh="" ylow="" yhigh">`

To be backward compatible with previously generated rrgraphs for master branch (where the ptc number doesn't change across the CHANX/CHANY nodes), the twist attribute is not required value (it is optional). 

#### Related Issue

@ganeshgore 

#### Types of changes

#### Checklist:


saaramahmoudi:
@vaughnbetz

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -709,6 +724,11 @@ class t_rr_graph_storage {
     // in a separate vector.
     vtr::vector<RRNodeId, short> node_layer_;
 
+    //Twist number is defined for CHANX/CHANY nodes, it is useful when different parts of same wire (with length > 1) have different ptc numbers.
```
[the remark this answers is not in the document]
vaughnbetz: First line: the comma should be a semicolon.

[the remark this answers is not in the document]
vaughnbetz: 3. Twist is only meaningful for CHANX and CHANY nodes; it is 0 for other node types. We also don't bother allocating this storage if the FPGA is not specified to be tileable; instead in that case the twist for all nodes will always be returned as 0.

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -164,6 +164,11 @@ class RRGraphView {
     inline short node_layer(RRNodeId node) const {
         return node_storage_.node_layer(node);
     }
+    
+     /** @brief Get the ptc number twist of a routing resource node. This function is inlined for runtime optimization. */
+    inline short node_ptc_twist(RRNodeId node) const{
+        return node_storage_.node_ptc_twist(node);
```
vaughnbetz, line 170: Check if not allocated; return 0 if not.

vpr/src/route/router_lookahead_map.cpp
```diff
@@ -1413,12 +1413,14 @@ static void compute_tile_lookahead(std::unordered_map<int, util::t_ipin_primitiv
...
                                    const int delayless_switch) {
     RRGraphBuilder rr_graph_builder;
+    int ptc_twist = 0;
     int layer = 0;
     int x = 1;
     int y = 1;
     build_tile_rr_graph(rr_graph_builder,
                         det_routing_arch,
                         physical_tile,
+                        ptc_twist,
```
vaughnbetz, line 1423: Change to not allocate the vector when we don't need it.

vpr/src/route/rr_graph.cpp
```diff
@@ -275,6 +277,7 @@ static void connect_src_sink_to_pins(RRGraphBuilder& rr_graph_builder,
 static void alloc_and_load_tile_rr_graph(RRGraphBuilder& rr_graph_builder,
                                          std::map<int, t_arch_switch_inf>& arch_sw_inf_map,
                                          t_physical_tile_type_ptr physical_tile,
+                                         int twsit,
```
vaughnbetz, line 280: misspelt.
Also, use ptc_twist_increment as the name or ptc_twist_incr

vpr/src/route/rr_graph.cpp
```diff
@@ -2099,7 +2111,8 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
             for (size_t j = 0; j < grid.height() - 1; ++j) {
                 if (i > 0) {
                     int tracks_per_chan = ((is_global_graph) ? 1 : chan_width.x_list[j]);
-                    build_rr_chan(rr_graph_builder, layer, i, j, CHANX, track_to_pin_lookup_x, sb_conn_map, switch_block_conn,
+                    /* In case of having different ptc number in CHANX/CHANY nodes, we should twist the "ptc_twist" number to the change factor*/
```
vaughnbetz, line 2114: clarify using twist increment terminology.

vpr/src/route/rr_graph.cpp
```diff
@@ -2117,7 +2130,8 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
                 }
                 if (j > 0) {
                     int tracks_per_chan = ((is_global_graph) ? 1 : chan_width.y_list[i]);
-                    build_rr_chan(rr_graph_builder, layer, i, j, CHANY, track_to_pin_lookup_y, sb_conn_map, switch_block_conn,
+                    /* In case of having different ptc number in CHANX/CHANY nodes, we should twist the "ptc_twist" number to the change factor*/
```
vaughnbetz, line 2133: Update comment as above.

vaughnbetz (changes requested):
I think you can store this more efficiently, since a tileable architecture is fundamentally a flyweight. 
(x,y,layer) -> rr_tile_data
rr_tile_data stores the ptc_twist_incr for each ptc_num.

We can land this in the interim if you like, but we should explore this flyweight.  It would allow efficient storage of additional openFPGA data, as well as saving storage for this entry.

saaramahmoudi:
@vaughnbetz all suggestions applied, ready to be merged.

saaramahmoudi:
@vaughnbetz 

The vector is defined but won't be allocated unless the graph type is UNIDIR_TILEABLE (rr_graph.cpp line 1371). I just checked for the normal flow and the memory foot print is the same. I will attach the titan benchmark shortly. 

I agree that we can move to fly-weighting at some point, but it can take time (I will create an issue for it to track it)

libs/librrgraph/src/base/rr_graph_view.h
vaughnbetz: Shouldn't this check that the storage is allocated and error out if it isn't?  (Seems important since we don't always allocate it).

vaughnbetz:
OK. In that case I think it's important to check the storage is allocate before accessing a member -- see my comment above.

vaughnbetz:
Thanks for adding the check. This is good to merge once CI passes.

rs-dhow:
Thanks for adding twist="#". Your picture at the start of this convo says 0,1,2,3,4, but your code comment says:
//For example, an L4 wire would change tracks 4 times with metal shorts [e.g. 0, 2, 4, 6] and track 6 would drive a switch -- together this implements an L4 wire with only one layout tile.
The latter is consistent with current twisted ptc numbers (for INC_DIR) so we're OK.

Since Mahdi's original change altered e.g.
node_lookup_.add_node(node, ix, iy, node_type, node_ptc_num, SIDES[0]);
to
node_lookup_.add_node(node, ix, iy, node_type, node_ptc_num + (i*twist), SIDES[0]);
I had assumed that (i) each *coordinate* already had a ptc number and (ii) we didn't need to save twist in the node.

I noticed the convo about "flyweight" attributes and saving memory. Since twist = (ptc of adjacent coordinate) - (ptc of xlow,ylow), there are reduction possibilities someday.

Thanks again.

saaramahmoudi:
@vaughnbetz updated with master branch, ready to merge

PR #2380, opened 2023-08-30
3D FPGA block select fix, and rr-node display fix for 3D FPGAs

Meet-Patel2580:
#### Types of changes


vpr/src/draw/draw.cpp
```diff
@@ -985,6 +985,7 @@ static void draw_router_expansion_costs(ezgl::renderer* g) {
 
 static void highlight_blocks(double x, double y) {
```
vaughnbetz, line 986: 1. Put a comment on what this function does in the .h file.  It highlights the block that was clicked on, looking from the top layer downwards for 3D devices (choose the top visible layer for overlapping blocks). It highlights the block, its fanin and fanout, in various colours by updating the draw_state variables.

2. Could consider factoring this routine to have a helper function that returns the blk_id (or -1) of the clicked on block.  That would shorten this function by moving the top part to another routine, and make commenting it easier.

vaughnbetz (changes requested):
A couple of changes to consider.

vpr/src/draw/draw_rr.cpp
Meet-Patel2580, line 1: highlighting of RR nodes has just been changed to ensure that only nodes on currently visible layers are checked to determine whether they should be highlighted or not. This fixes the bug of the user clicking on a node, but not seeing it highlight due to the code mistakenly changing the highlight setting for a node on an invisible layer.

vpr/src/draw/draw.h
```diff
@@ -155,6 +155,17 @@ bool rgb_is_same(ezgl::color color1, ezgl::color color2);
  */
 t_draw_layer_display get_element_visibility_and_transparency(int src_layer, int sink_layer);
 
+/**
+ * @brief takes in the x and y coordinates of where the user clicked on the screen and returns the corresponding clusterBlockId that represents
+ *         the clb clicked upon by the user on a currently visible FPGA layer. Search for the clb begins from the top layer to ensure to
+ *         return the clusterBlockId of a clb on a higher layer during instances of overlap between clb blocks.
+ * @param x
```
vaughnbetz, line 162: Say in world coordinates (for both x and y)

vaughnbetz (commented):
One suggested comment update. Can update that before or after merge, based on whatever is more convenient.

vaughnbetz:
@Meet-Patel2580 : are the clicked x,y coordinates in world coordinates?  If so we should comment that; I think after that it can be merged.

vaughnbetz:
Looks good; can be merged as soon as CI finishes.

PR #2381, opened 2023-08-30
Routing Util 3d

alanbu1:
Implemented routing util for multi layered architectures.


vaughnbetz (changes requested):
Please check .h file comments before merging.

vaughnbetz:
CI failed hard ... updating to the master to see if they work now.

vpr/src/route/route_util.h
```diff
@@ -1,9 +1,19 @@
...
+#include "draw_global.h"
 
 vtr::Matrix<float> calculate_routing_avail(t_rr_type rr_type);
-vtr::Matrix<float> calculate_routing_usage(t_rr_type rr_type, bool is_flat);
+
+/**
+ * @brief Calculates routing usage over entire grid
+ * Collects all in-use nodes and records number of used resources
+ * in each x/y channel. Takes into consideration visible layers
+ * for multi-layered architectures. Also takes into consideration
```
vaughnbetz, line 13: I think this comment needs updating.

1. @brief: Calculates and returns the usage over the entire grid for the specified type of rr_node to the usage array. The usage is recorded at each (x,y) location. 

@param rr_type: Type of rr_node that we are calculating the usage of; can be CHANX or CHANY.
@param is_flat: Is the flat router being use or not?
@param only_visible: If true, only record the usage of rr_nodes on layers that are visible according to the current drawing settings.

rr_type: type of rr_node to update usage for (currently supports CHANX or CHANY)

vpr/src/route/route_util.cpp
```diff
@@ -27,6 +29,13 @@ vtr::Matrix<float> calculate_routing_usage(t_rr_type rr_type, bool is_flat) {
 
     //Record number of used resources in each x/y channel
     for (RRNodeId rr_node : rr_nodes) {
+        if (!is_print) {
+            t_draw_state* draw_state = get_draw_state_vars();
```
vaughnbetz, line 33: Wrap in a 
#ifndef NO_GRAPHICS

#endif

for this whole if statement.  Otherwise it won't compile when we're trying to compile with NO_GRAPHICS as all the drawing functions and gtk interface have been stripped out.

PR #2385, opened 2023-08-31
Partition based packer

behnam-rs:
#### Description
Our approach combines the benefits of partitioning-based and seed-based packing algorithms (VPR). It consists of two main steps 1) a partitioning step to hierarchically divide the circuit into subcircuits, and 2) a seed-based packing step that clusters all subcircuits concurrently.

An optimal clustering solution should have a low number of clusters, high intra-cluster connections, and low inter-cluster connections. To achieve this, we employ partitioning algorithms designed for hypergraphs, where hyperedges connect more than two vertices. Logic circuits are an example of hypergraphs, with logic blocks as vertices and nets as hyperedges.

The hypergraph partitioning algorithms split the hypergraph into configurable partitions, minimizing interconnecting hyperedges between them. One well-known algorithm for this purpose is hmetis.

To adapt hmetis for molecule clustering in VPR, we split the circuit into partitions such that the number of molecules in each partition is equal to the clb (configurable logic block) capacity. However, packing molecules into a clb has more constraints than just the clb capacity. To address this, we increase the number of molecules in each partition and then use the seed-based packing algorithm inside VPR to pack the molecules within each partition. The molecules in each partition are packed only to the molecules present in that partition.
#### Motivation and Context
The current packer in VPR employs a greedy algorithm by sorting molecules into a seed array based on a configurable option (--cluster_seed_type). In each iteration, it selects the best unpacked molecule from the seed array, generates a new cluster for it, and then adds other unpacked molecules to this new cluster using a greedy algorithm. However, due to its greedy nature, the algorithm fails to explore the search space effectively, leading to the problem of getting stuck in local minima.

Via our bench-marking, this feature bring better QOR.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/SetupVPR.cpp
+++ b/vpr/src/base/SetupVPR.cpp
@@ -597,6 +597,9 @@ void SetupPackerOpts(const t_options& Options,
     PackerOpts->timing_update_type = Options.timing_update_type;
     PackerOpts->pack_num_moves = Options.pack_num_moves;
     PackerOpts->pack_move_type = Options.pack_move_type;
+    PackerOpts->use_partitioning_in_pack = Options.use_partitioning_in_pack;
+    PackerOpts->partitioner_path = Options.partitioner_path;
+    PackerOpts->number_of_molecules_in_partition = Options.number_of_molecules_in_partition;
 }
 
 static void SetupNetlistOpts(const t_options& Options, t_netlist_opts& NetlistOpts) {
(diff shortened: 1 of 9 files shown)


vaughnbetz:
Interesting PR!  Can you add QoR data on the VTR, Titan and Koios benchmark suites?  See https://docs.verilogtorouting.org/en/latest/README.developers/#evaluating-quality-of-result-qor-changes for a detailed how-to guide.

Issue #2356, opened 2023-07-31
CI failures due to old library versions


duck2:
We get CI failures on the github runners with this error message (and not the google cloud runners?):

```
CMake Deprecation Error at abc/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

This happens with abc and capnproto. We can update them or tell CMake to ignore this for now. (but we should remember to update eventually)

plan:

- [x] Short term: I'll tell cmake to ignore this
- [ ] Short term: I'll update capnproto?
- [ ] Medium term: @alirezazd can you update abc?

ept221:
This may be related, but when I attempt to build VTR, I get a failure right after libabc is built:

`[ 76%] Building C object abc/CMakeFiles/libabc.dir/src/bdd/llb/llb4Image.c.o
[ 76%] Building C object abc/CMakeFiles/libabc.dir/src/bdd/llb/llb4Sweep.c.o
[ 76%] Linking CXX static library libabc.a
[ 76%] Built target libabc
make[1]: *** [Makefile:141: all] Error 2
make: *** [Makefile:94: all] Error 2
`

PR #2388, opened 2023-09-11
ABC Version Update

navidjafarof:
#### Description

This PR is made to update ABC to the latest version and make sure everything is stable.

Source repository: https://github.com/berkeley-abc/abc
#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2356#event-9966028236
#### Types of changes

#### Checklist:


navidjafarof:
@alirezazd This is the PR regarding the ABC update.

alirezazd:
@navidjafarof Please remove the merge commit as its not allowed in this codebase.

alirezazd:
@vaughnbetz FYI

navidjafarof:
Here are the spreadsheets regarding the Koios and VTR results.
[ABC_update_koios.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/12745068/ABC_update_koios.xlsx)
[ABC_update_VTR.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/12745069/ABC_update_VTR.xlsx)

vaughnbetz:
@navidjafarof : the abc results all appear to be identical (num_pre_packed_blocks, abc_max_depth etc. are the same for every circuit in both VTR and Koios).  It looks like the same version of abc was used in both runs, as any update to abc should change some of these metrics for at least some circuits.  Can you double check that the QoR test was run correctly?

navidjafarof:
@vaughnbetz I will double check the results and let you know. I will attach new results as soon as it is ready.

navidjafarof:
Dear @vaughnbetz,
Can you please let me know what do you expect to see in the results after updating the ABC? What parameters should change? Can you give me more details so I would know what to look into?
Thank you
Navid

vaughnbetz:
I'd expect the metrics due solely to synthesize, like the num_pre_packed_blocks and abc_max_depth, to change. That will lead to knock on effects in downstream metrics like critical_path_delay and so on.

It seems unlikely that a much more recent version of abc would get identical results; I'd expect at least some change in QoR, even with the same synthesis commands.  It's also possible that the claimed abc improvements in the http://people.eecs.berkeley.edu/~alanmi/publications/2023/date23_gap.pdf are due to running different synthesis commands so that is also worth looking into.

alirezazd:
@vaughnbetz I'm re-checking the QoR, but as you might know this paper is another version of ABC which is called (ABC9) it is not the official version that we've been usning in VTR, it is the ABC which Yosys uses.

vaughnbetz:
I think they just gave ABC9 as the name to the new ABC ... probably best to clarify what new code is in this PR then.

alirezazd:
@vaughnbetz I just Updated the PR with the link of the ABC that we've used. We can also try to use the latest ABC9.
According to the paper: "specifically, we describe how ABC9 — a set of feature additions to ABC — was integrated into Yosys upstream and available in the latest version."

By doing this we can remove the ABC directory and just use the ABC which available in the Yosys directory of the codebase.

navidjafarof:
Dear @vaughnbetz,
I wanted to ensure the accuracy of my comparison between the results of the master branch and my own branch. To accomplish this, I downloaded the artifacts from GitHub. Specifically, I compared the results of "koios" from "qor_results_vtr_reg_nightly_test6" with "vtr_reg_qor_chain" from "vtr_reg_nightly_test3," as we discussed during the meeting. The expected parameters were still the same and no differences were noticed. However, I also compared the tasks for which I updated the golden results, and here I did observe differences. As a specific example, I compared the results for "multless_consts" from "vtr_reg_nightly_test1_odin," and I have attached a spreadsheet detailing this comparison for your reference.
[comparison_output.xlsx](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/12755648/comparison_output.xlsx)

navidjafarof:
Dear @vaughnbetz,
This PR is ready to be merged.
Thanks,
Navid

PR #2389, opened 2023-09-11
Placement Debug Messages 

amin1377:
In this pull request (PR), we've made some enhancements related to debugging messages in placement, aligning them with those found in the router. These additional details are printed only when the `VTR_ENABLE_DEBUG_LOGGING` option is enabled.

Two new command-line arguments, "--placer_debug_net" and "--placer_debug_block" are added. When either of them is set to "-1," the placer will output debug information concerning all blocks and nets. If the value is greater than or equal to zero, the placer will print information when the moving block matches the specified block ID or when one of the connections of the given net id is connected to the block being considered for movement.


amin1377:
@vaughnbetz I think this PR is ready to merge

vpr/src/base/vpr_types.h
```diff
@@ -1199,6 +1199,9 @@ struct t_placer_opts {
     int floorplan_num_horizontal_partitions;
     int floorplan_num_vertical_partitions;
 
+    int placer_debug_block;
```
vaughnbetz, line 1202: Please add a brief comment for each of these; it can refer to the command line parser help for more detail if you want to avoid duplication.

vpr/src/place/move_utils.cpp
```diff
@@ -15,6 +15,8 @@
 //Note: The flag is only effective if compiled with VTR_ENABLE_DEBUG_LOGGING
 bool f_placer_breakpoint_reached = false;
 
+bool f_placer_debug = false;
```
vaughnbetz, line 18: Comment this variable.

vpr/src/place/move_utils.h
```diff
@@ -9,6 +9,8 @@
 /* To turn off incremental bounding box updates, set this to a huge value */
 #define SMALL_NET 4
 
+extern bool f_placer_debug;
```
vaughnbetz, line 12: 1. Comment.
2. Why not put this in the placement context or placer opts?  We're trying to get away from naked global variables.

vaughnbetz (changes requested):
I'd like to get rid of the naked global variable; prefer putting it in an existing context.
Global variables and structure members should also be commented.

Finally, just to be safe please run at least one circuit with default compile / cmake settings and confirm no slowdown.  There shouldn't be any since the logging is conditionally compiled, but it is best to always be paranoid and check.

vaughnbetz:
In this CI, it seems the tests that ran on github runners worked, but all the google cloud ones failed with this message:

Error: Failed to resolve action download info.

If this is still happening please file an issue ASAP and we'll discuss in the vtr meeting today.  If we need something from Google, we can ask Herman in the meeting.

vpr/src/place/initial_placement.cpp
```diff
@@ -963,6 +977,16 @@ static void place_all_blocks(vtr::vector<ClusterBlockId, t_block_score>& block_s
             heap_blocks.pop_back();
 
             auto blk_id_type = cluster_ctx.clb_nlist.block_type(blk_id);
+
+            const auto& cluster_blk_pb_type = cluster_ctx.clb_nlist.block_type(blk_id)->pb_type;
```
vaughnbetz, line 981: 1. This should be a subroutine I think (cleaner, less cognitive pollution for the main code path).
2. It should be guarded by your #ifdef VTR_ENABLE_DEBUG_LOGGING

vpr/src/place/move_utils.cpp
```diff
@@ -520,6 +560,16 @@ ClusterBlockId propose_block_to_move(int& logical_blk_type_index, bool highly_cr
         }
     }
 
+    if (b_from) {
```
vaughnbetz, line 563: This should be a subroutine I think (cleaner, less cognitive pollution for the main code path).
It should be guarded by your #ifdef VTR_ENABLE_DEBUG_LOGGING

vaughnbetz (changes requested):

vaughnbetz:
If the router debug messages are similar, they should also be refactored to use a helper function and have conditional compilation guards.

vaughnbetz:
@amin1377 tells me the router code is not in an inner loop so we can leave it alone.

amin1377:
I tested this PR on the VTR large benchmarks. The QoR was exactly the same, and the run times were less ~1% different.

vpr/src/place/initial_placement.cpp
```diff
@@ -963,6 +977,12 @@ static void place_all_blocks(vtr::vector<ClusterBlockId, t_block_score>& block_s
             heap_blocks.pop_back();
 
             auto blk_id_type = cluster_ctx.clb_nlist.block_type(blk_id);
+
+#ifndef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 981: Shouldn't this be #ifdef?

vpr/src/place/move_utils.cpp
```diff
@@ -497,7 +497,62 @@ std::set<t_pl_loc> determine_locations_emptied_by_move(t_pl_blocks_to_be_moved&
     return empty_locs;
 }
 
-ClusterBlockId propose_block_to_move(int& logical_blk_type_index, bool highly_crit_block, ClusterNetId* net_from, int* pin_from) {
+#ifndef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 500: Shouldn't this be #ifdef?

vpr/src/place/move_utils.cpp
```diff
@@ -519,6 +574,9 @@ ClusterBlockId propose_block_to_move(int& logical_blk_type_index, bool highly_cr
             b_from = pick_from_block(logical_blk_type_index);
         }
     }
+#ifndef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 577: #ifdef

vpr/src/place/move_utils.h
```diff
@@ -324,4 +328,9 @@ bool intersect_range_limit_with_floorplan_constraints(t_logical_block_type_ptr t
 
 std::string e_move_result_to_string(e_move_result move_outcome);
 
+#ifndef VTR_ENABLE_DEBUG_LOGGING
```
vaughnbetz, line 331: 1. Should be #ifdef, not $ifndef
2. Needs a comment saying what the routine does.

vaughnbetz (changes requested):
#ifndef looks backwards.
Needs a comment on the new helper routine.

amin1377:
> #ifndef looks backwards. Needs a comment on the new helper routine.

You are right. It should be #ifdef. Apparently, it didn't have the run-time effect that I thought it had!

vpr/src/place/move_utils.h
```diff
@@ -324,4 +328,17 @@ bool intersect_range_limit_with_floorplan_constraints(t_logical_block_type_ptr t
...
 std::string e_move_result_to_string(e_move_result move_outcome);
 
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+/**
+ * @brief If the block ID passed to the placer_debug_net parameter of the command line is equal to blk_id, or if any of the nets
+ * connected to the block share the same ID as the net ID passed to the placer_debug_net parameter of the command line,
+ * then debugging information should be printed.
+ *
+ * @param placer_opts
+ * @param blk_id The if of the block that is considered to be moved
```
vaughnbetz, line 338: if -> id

amin1377:
@vaughnbetz CIs are green

PR #2397, opened 2023-09-15
Re-clustering API bug fixes

MohamedElgammal:
#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/pack/re_cluster.cpp
+++ b/vpr/src/pack/re_cluster.cpp
@@ -166,8 +166,7 @@ bool swap_two_molecules(t_pack_molecule* molecule_1,
                         int verbosity,
                         t_clustering_data& clustering_data) {
     auto& cluster_ctx = g_vpr_ctx.mutable_clustering();
-    auto& helper_ctx = g_vpr_ctx.mutable_cl_helper();
-
+    
     //define local variables
     PartitionRegion temp_cluster_pr_1, temp_cluster_pr_2;
 
--- a/vpr/src/pack/re_cluster.cpp
+++ b/vpr/src/pack/re_cluster.cpp
@@ -71,6 +71,14 @@ bool move_mol_to_new_cluster(t_pack_molecule* molecule,
     //Commit or revert the move
     if (is_created) {
         commit_mol_move(old_clb, new_clb, during_packing, true);
+        // Update the clb-->atoms lookup table
+        helper_ctx.atoms_lookup.resize(helper_ctx.total_clb_num);
+        for (int i_atom = 0; i_atom < molecule_size; ++i_atom) {
+            if (molecule->atom_block_ids[i_atom]) {
+                helper_ctx.atoms_lookup[new_clb].insert(molecule->atom_block_ids[i_atom]);
+            }
+        }
+
         VTR_LOGV(verbosity > 4, "Atom:%zu is moved to a new cluster\n", molecule->atom_block_ids[molecule->root]);
     } else {
         revert_mol_move(old_clb, molecule, old_router_data, during_packing, clustering_data);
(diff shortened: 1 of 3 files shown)


vaughnbetz:
@MohamedElgammal some tests failed due to download issues --> relaunch?
Adding @AlexandreSinger as a potential reviewer.

vpr/src/pack/re_cluster.cpp
```diff
@@ -157,6 +165,8 @@ bool swap_two_molecules(t_pack_molecule* molecule_1,
                         bool during_packing,
                         int verbosity,
                         t_clustering_data& clustering_data) {
+    auto& cluster_ctx = g_vpr_ctx.mutable_clustering();
```
AlexandreSinger, line 168: I put this on Amin's MR, but why is this declared so far away from where it is used? Is this a style thing?

vpr/src/pack/re_cluster.cpp
```diff
@@ -193,6 +203,11 @@ bool swap_two_molecules(t_pack_molecule* molecule_1,
         return false;
     }
 
+    t_pb* clb_pb_1 = cluster_ctx.clb_nlist.block_pb(clb_1);
+    std::string clb_pb_1_name = (std::string)clb_pb_1->name;
+    t_pb* clb_pb_2 = cluster_ctx.clb_nlist.block_pb(clb_2);
+    std::string clb_pb_2_name = (std::string)clb_pb_2->name;
```
AlexandreSinger, line 209: I also put this on Amins PR lol, but I think these should use the std::string's constructor instead of using the c-style casting

vpr/src/pack/re_cluster_util.cpp
```diff
@@ -110,14 +110,19 @@ t_lb_router_data* lb_load_router_data(std::vector<t_lb_type_rr_node>* lb_type_rr
...
     }
     return (router_data);
 }
 
 bool start_new_cluster_for_mol(t_pack_molecule* molecule,
                                const t_logical_block_type_ptr& type,
-                               const int mode,
-                               const int feasible_block_array_size,
+                               const int& mode,
+                               const int& feasible_block_array_size,
```
AlexandreSinger, line 125: Why were these made pass by reference? Did you see any performance increase after changing this?

vaughnbetz: Passing an int by constant reference isn't going to be faster and might be slower as a pointer is 64-bits and an int is 32, and pointer chasing is slow. After compiler optimization the whole pointer/reference stuff might go away, but generally it is better to pass things smaller than or equal to pointer size by value; use const & for bigger things that aren't changed in a function.

vaughnbetz:
@MohamedElgammal : should this PR be landed?
@KA7E may be interested in these fixes.

AlexandreSinger:
I am also interested in these fixes.

MohamedElgammal:
looking good to me .. We need to have a second look on the failing tests ... let me rerun them and update you

AlexandreSinger:
I think the 4 LLVM CI failures are caused by this branch being behind VTR master by some time. I think rebasing it onto VTR master should resolve the issue. @MohamedElgammal could you rebase the branch? I see a button to do so, but I am not sure if doing it from the GitHub UI would cause any issues.

AlexandreSinger:
I have rebased the branch using the GitHub UI. Hopefully the tests should pass now and this can be merged in.

vpr/src/pack/re_cluster_util.cpp
vaughnbetz: So I suggest putting these back to const int, although it isn't a huge deal.

vaughnbetz:
Great, thanks.  @MohamedElgammal : any barrier to merging this if CI passes?

vaughnbetz:
@AlexandreSinger : I believe this is the branch that has an example of how to use the re-clustering API.  
https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/packing-multithreading
(see pack_utils.cpp)

MohamedElgammal:
@AlexandreSinger : Thanks for rebasing it.
@vaughnbetz : this branch should be good to be merged if it passes (I think it will pass anyway as the API isn't used in the master branch). My plan is to add unit tests for the different use cases of the API functions but never got to this task.

vaughnbetz:
Thanks Mohamed.  Merging!

Issue #2399, opened 2023-09-22
Segment Usage by Type does not return the values that the title would suggest it should


WhiteNinjaZ:
In vpr.out one of the reports is "Segment usage by type (Index)". I would expect from the description/title (usage by type (index)) that this report should give the percentage of segments of each type in the RR that are actually used in the net-list for the design. However, Looking at the code (shown bellow) it appears that what is actually happening is that each wire of a given length being used in the net list is being counted up and that this value is being used to give a percentage of wires used. This creates an issue if wires with different properties have the same length and a user wants to see how many wires with specific properties are used in the design. 

#### Where the Code Producing this issue can be found
The following image is lines 103-122 of ${VPR_ROOT}/vpr/src/route/segment_stats.cpp:
![image](https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/83309027/b63f995b-d827-4f50-a2a0-37d400fb4e96)

#### Expected Behaviour
The description/title of this section should either reflect that we are looking at wire counts by length and index or the code base should be changed to count wire usage using the indexes of those wires in the RR.

Also the title of this section should probably be changed to Segment **Occupancy** By type  since Usage could be interpreted as occurrences (which is different).

amin1377:
@WhiteNinjaZ : Whenever you have time, please take a look at the PR related to this issue and let me know if it would solve the problem.

PR #2408, opened 2023-10-04
Fix Segment Usage Message

amin1377:
This PR addresses issue[#2399](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2399)

Sample output:
Screenshot from 2023-10-04 15-11-38
Screenshot from 2023-10-04 15-13-06


WhiteNinjaZ:
Superb, This looks great and should fix the issue.

vaughnbetz (commented):
Looks OK; slightly odd to use both a map and a vector (two different paradigms for storing the data per length/axis).  Could just use a map or maps, and get rid of the "seen_lengths" set.  Your call if you want to change it or not.

vaughnbetz:
Looks good. Will merge once CI finishes.

PR #2418, opened 2023-10-11
Add 'not found' check for finding file extension of arch xml

coolbreeze413:
#### Description
While checking for a file extension with `.find_last_of()` the 'not found' scenario also needs to be checked.
While using FOEDAG + VPR at Quicklogic, we already have an internal encryption-decryption mechanism which is in use and that uses different file extensions (neither `.xml` nor `.xmle`), and in many cases no extension at all.
These kind of XMLs with no extension, cause the check for finding the extension to throw `std::out_of_range` as the return value for `position` is actually `std::string::npos` which needs to be checked before using it in the the `substr()` call.

Additionally a typo fix while checking to warn users of unrecognized extension.

#### Related Issue
As this change is specific to `openfpga` branch, I have not raised an issue for this.
Please let me know if I need to raise the issue.

#### Motivation and Context
Fix file extension check process

Regular testcases of VPR suite.

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/scripts/hostsetup.sh
+++ b/.github/scripts/hostsetup.sh
@@ -71,7 +71,7 @@ apt install -y \
   gcc-9 \
   wget \
   openssl \
-  libssl-dev
+  libssl-dev \
   libtbb-dev
 
 # installing the latest version of cmake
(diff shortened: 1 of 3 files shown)


coolbreeze413:
@tangxifan @tpagarani @alaindargelas
Please review the changes if they look ok to you.

tangxifan (approved):
LGTM

alaindargelas:
LGTM

Issue #2420, opened 2023-10-19
run_vtr_flow currently does not properly handle sdc files


WhiteNinjaZ:
I have noticed that the current run_vtr_flow.py script does not properly handle the `--sdc_file` option. Whenever this option is used VPR complains that "SDC file '<name of SDC file>' not found" but continues through the flow and just uses default timing constraints. It looks like this is just an issue from the script not copying the file to the provided temp directory as doing so manually works without any issues.  

#### Expected Behaviour

Running vtr_flow with the --sdc parameter should work properly and correctly pass timing constraints to the flow.

#### Your Environment

* VTR revision used:
Running on the most recent version of VTR as of writing (15c534068)

Issue #2302, opened 2023-05-04
Using Parmys For Xilinx-Like Architectures


amin1377:
I tried to implement `stereovision0.v` on `simple-7series.xml`. I ran this command: `run_vtr_flow.py stereovision0.v simple-7series.xml` and got this error message:
`simple-7series/stereovision0		Error: Executable yosys failed full command:  /usr/bin/env time -v /home/mohagh18/vtr-verilog-to-routing/build/bin/yosys -c synthesis.tcl returncode  :  1 log file    :  temp/parmys.out failed: Executable yosys failed (took 0.27 seconds, overall memory peak 11.65 MiB consumed by parmys run)`

[Here](https://utoronto-my.sharepoint.com/:u:/g/personal/amin_mohaghegh_mail_utoronto_ca/EVCYULZBos9KrZRaubFQob4BnaAz5DP78VRQon3lNrTPSQ?e=XlHIYm) is the `parmys.out` file.

@WhiteNinjaZ @vaughnbetz

WhiteNinjaZ:
It looks like this is a synthesis problem with the Parmys[Yosys] front end not an arch issue. I am getting the same problem on multiple old example architectures including the example arch from the documentation (i.e. `$VTR_ROOT/vtr_flow/scripts/run_vtr_flow.py $VTR_ROOT/doc/src/quickstart/blink.v $VTR_ROOT/vtr_flow/arch/timing/EArch.xml -temp_dir . --route_chan_width 100`). Odin works fine, but when trying to use yosys front end I am getting the same issue as amin.

vaughnbetz:
Thanks Joshua. @alirezazd any ideas?

poname:
I just tested `stereovision0.v` on `simple-7series.xml` and it was OK.
The command history:
```
  504  git clone https://github.com/verilog-to-routing/vtr-verilog-to-routing.git vtr
  505  cd vtr
  506  make -j10
  507  cd vtr_flow/scripts/
  508  ./run_vtr_flow.py ../benchmarks/verilog/stereovision0.v ../arch/xilinx/simple-7series.xml
```

and the output:
```
simple-7series/stereovision0            OK (took 1250.60 seconds, overall memory peak 597.01 MiB consumed by vpr run)

```

`Yosys` not getting along with `temp_dir` was reported previously in #2152.

poname:
[...]

PR #2427, opened 2023-11-02
fixed sdc file copy and parmys abs path to temp folder issue

WhiteNinjaZ:
This PR resolves issue #2420 and provides a work around to #2302. 
#### Description

The run_vtr_flow.py script was modified to properly copy a specified sdc file into a temp directory regardless of if a relative or absolute path is given to the sdc file. Also, this PR provides a work around for the new requirement by parmys that the path specified for the temp directory be an absolute path. The updated script provides the flow with an absolute path whether an absolute or relative path is specified. Additionally, several command line options are specified in the documentation as having a double `--` rather than a single `-` (namely `fixed_pins` and `sdc_file`). The script was modified to provide consistency between how these options are used in both the vpr executable and the script.

Several designs were run through the flow using `-temp` and `--sdc_file` files with both relative and absolute paths.

#### Types of changes


vaughnbetz:
Not sure why CI failed -- @WhiteNinjaZ could you relaunch this?

vtr_flow/scripts/run_vtr_flow.py
```diff
@@ -525,6 +525,8 @@ def vtr_command_main(arg_list, prog=None):
 
     assert args.temp_dir
     temp_dir = Path(args.temp_dir)
+    if not str(temp_dir).startswith("/"):
```
vaughnbetz, line 528: Does parmys still need an absolute path?  If not, we can remove this line I think.  If so, this should be commented (say why we're making an absolute path).

vaughnbetz (commented):
@WhiteNinjaZ : I have a question about the absolute path requirement.

vaughnbetz:
@WhiteNinjaZ : It would be good to move this PR along ... see questions and CI failures above.

vtr_flow/scripts/run_vtr_flow.py
WhiteNinjaZ: As far as I can tell, parmys still fails when an absolute path to the temp folder is not included. I have added a comment to this if statement mentioning this.

WhiteNinjaZ:
Will do, it looks like the logs expired for most of the CI tests so re-running now.

vaughnbetz:
Some jobs timed out ... maybe due to busy CI lately. Re-launching the failed jobs.

vaughnbetz:
Not sure why we have some CI failures; relaunching.

vaughnbetz:
It seems like some tests timed out again, and I can't find the usual button to relaunch the tests.  @WhiteNinjaZ @AlexandreSinger : any ideas?  This change looks pretty safe (scripting update) given that most CI tests passed and others seem to have timed out.  If you can figure out how to relaunch that would be safest, but if @WhiteNinjaZ is pretty sure this works with all tests I could also just merge.

vaughnbetz:
Ah, didn't think of that.  Thanks @AlexandreSinger

AlexandreSinger:
@vaughnbetz I also saw this PR and was curious why I could not rerun the CI. I believe the reason is because the CI clears runs after a certain period of time (I think a couple of weeks). If a CI run fails for a PR, the CI cannot be rerun the usual way if its been too long (for this PR, the last run was in June).

Normally when this happens I rebase the branch onto Master; however for some reason that button does not work for me on this PR. It may be because the branch is under a different fork (that I do not have the right to modify). So the button to rebase / merge with master was not here for me.

To get around this issue I just closed and reopened the PR. That relaunched the CI. Kinda goofy, but it works. If it fails CI again, I recommend @WhiteNinjaZ to rebase his branch. I am a bit worried about this change in that maybe some CI tests rely on the CLI options to be a certain way and its causing timeouts. The CI will tell.

vaughnbetz:
Thanks @AlexandreSinger .  Yes, I was a little worried in case there was something unexpected happening in the tests due to the scripting changes.

AlexandreSinger:
@vaughnbetz Yeah there is something in this PR that is causing the strong tests to time out. My guess is the use of `os.getcwd()`. These scripts are often run all over the place, so I am not sure if we can guarantee that the current working directory is where we think it is. Looking through the strong tests there are a lot of failures going on. Perhaps this should be rebased onto master and tested locally.

vaughnbetz:
Thanks @AlexandreSinger . @WhiteNinjaZ : can you try this out locally for the strong tests as Alex suggests?  Something is going wrong with the strong tests.

PR #2430, opened 2023-11-06
Separate flat router test

amin1377:
In this pull request, a new test directory has been created specifically for the run-flat tests. Previously, these tests were located under vtr-nightly-test2/3, which sometimes resulted in timeouts, especially for nightly test 2.

By establishing a dedicated nightly test directory for run-flat tests, they now run on a separate machine. This segregation ensures that the run-time of tests under other nightly tests remains unaffected.


amin1377:
I had to comment out the tests when the minimum channel width is not specified. I believe it would be better to merge this PR now to ensure that CI tests are running correctly. I will create another PR to add tests for cases where the flat router is used to find the minimum channel width.

amin1377:
@vaughnbetz This PR is ready to merge. It is only failing on Clan tests which there is an open issue on it.

vaughnbetz:
Thanks @amin1377 .  @duck2 : Clang is failing on the unit test library catch2.  Any ideas?  I need somebody to figure out what's going on (seems like clang doesn't like uint8?).

PR #2433, opened 2023-11-08
replace assert with non-fatal check for mode conflict

KA7E:
Replace assert with non-fatal check for mode conflict.  Conflict returns BLK_FAILED_FEASIBLE.

#### Description

Changed an assertion to an if statement.

#### Motivation and Context

The legalizer encounters mode conflicts, since the external placer does not model modes and may not differentiate between .names and lcell luts.  The legalizer needs to be able to continue when it encounters a mode mismatch.

CI tests pass (except known spurious Clang failures)
Code with this change has successfully run place and route on all Titan23s.

#### Types of changes

#### Checklist:


vaughnbetz:
Updating branch with master changes & re-launching CI.  Clang errors should be gone now.
If you know what the error condition is you could write a comment; if you're not sure what the comment should be we can leave it (or if it is obvious so no comment needed).

PR #2435, opened 2023-11-09
add invalid cluster id checks 

KA7E:
Add cluster id checks before accessing data structure indexed by cluster id.

#### Description

In init_clb_atoms_lookup and ClusterAtomsLookup::init_lookup, added if statement so that  data structure is only accessed if the cluster id to be used as the index is valid.

#### Motivation and Context

Since the legalizer works with incomplete data structures, it needs this check to avoid out of bounds accesses.

passes vtr_reg_basic and vtr_reg_strong, has been run on all Titan23s

#### Types of changes

#### Checklist:


vaughnbetz:
Updating branch and re-launching CI.
Would be good to add a comment of why an invalid index could occur unless it's super obvious from the context code.

PR #2437, opened 2023-11-09
Update catch2 and move to submodule. Add dependabot for submodules

### LINKED ISSUE #2436: catch2 not including <stdint.h>, causing compiler errors with Clang
Recently Clang has failed to compile the catch2 library.  This appears to be a valid error due to catch2 not including <stdint.h>

#### Expected Behaviour
VTR should compile under Clang.

#### Current Behaviour
```
[ 65%] Building CXX object libs/EXTERNAL/libcatch2/src/CMakeFiles/Catch2.dir/catch2/catch_test_case_info.cpp.o
In file included from /home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/libs/EXTERNAL/libcatch2/src/catch2/catch_test_case_info.cpp:8:
/home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/libs/EXTERNAL/libcatch2/src/catch2/../catch2/catch_test_case_info.hpp:39:37: error: unknown type name 'uint8_t'
    enum class TestCaseProperties : uint8_t {
                                    ^
```

#### Possible Solution
This appears to be fixed in the current catch2 repo. The cleanest fix would be to do a git subtree pull to get the latest code I think; there might be a few issues so it would be best if someone else did this so he/she had the time to fix any integration issues.

A quick and dirty fix would be to add the missing include. 

#### Steps to Reproduce
Run CI.

#### Context
All our CI tests have failures.

jgoeders:
#### Description
Replace catch2 with a submodule, which is an updated version of catch2.  This PR also contains some fixes for the newer catch2 version.

This is the first time submodules are being introduced into the VTR code base; however, when linking against an externally developed library like catch2, they are more appropriate than subtrees (subtrees bring in all files and history into the repo, which isn't what we want here).

One issue I encountered is we have one subtree project (parmys) that has submodules within it.  This breaks the normal submodule initialization (https://stackoverflow.com/questions/45422693/how-do-i-init-submodules-nested-in-a-subtree).  To correct this you would need to either remove these submodules from the subtree, or re-initialize these submodules in the parent repo in the top-level `.gitmodules`.  Since it doesn't seem like we are using these parmys submodules in VTR (they would have had to be configured at the top-level to use them), I've simply removed them from the parmys subtee.

This also configures a github actions dependabot to create PRs when submodules have updates.  This is run monthly.

#### Related Issue
Fixes #2436 

#### Motivation and Context
PR is currently broken due to out-of-date dependency.

CI tests

#### Types of changes

#### Checklist:


jgoeders:
@duck2 Does this look ok? https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2437/commits/1c610624d83abbd7f4d61c50918d0a8ea50197f2 (sorry not familiar with this code and can't tell if the missing bracket was accidentally or intentionally part of the test case)

duck2:
Looks OK

PR #2447, opened 2023-11-27
Change the base cost for IPIN for SPEC

amin1377:
In this PR, the base cost of IPIN is multiplied by a number that can be perfectly represented in binary format, replacing the previous one (0.95).


libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp
```diff
@@ -351,7 +351,13 @@ static void load_rr_indexed_data_base_costs(const RRGraphView& rr_graph,
     rr_indexed_data[RRIndexedDataId(SOURCE_COST_INDEX)].base_cost = delay_normalization_fac;
     rr_indexed_data[RRIndexedDataId(SINK_COST_INDEX)].base_cost = 0.;
     rr_indexed_data[RRIndexedDataId(OPIN_COST_INDEX)].base_cost = delay_normalization_fac;
+    // IF the SPEC_CPU flag is set, we need to make sure that all floating point numbers are perfectly representable in
```
vaughnbetz, line 354: I'd expand this comment.  '... are perfectly representable in a binary mantissa (without round-off) so we can get the same routing result on different platforms. Since the router cost calculations and heap use floating point numbers, normally we get slightly different round off with different compiler settings, leading to different heap sorts and hence different routings. To make result validation for SPEC easier, we choose all router parameters to result in calculations that fit perfectly in a 24-bit binary mantissa.  .875 = 1/2 + 1/4 + 1/8 can be perfectly represented in a binary mantissa with only the first 3 bits set.'

vaughnbetz (changes requested):

libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp
vaughnbetz: Also "IF" -> "If"

Issue #2453, opened 2023-11-28
Clean up / control behind a variable detailed routing algorithm output


vaughnbetz:
These lines come out late in routing:
[...]

PR #2456, opened 2023-11-29
Routing Stats

amin1377:
In this pull request, the routing status message has been modified. With this update, details such as the count of internal/external nodes pushed to/popped from the heap, breakdown of the node types pushed to/popped from the heap, and other extraneous messages will now be displayed only if `VTR_ENABLE_DEBUG_LOGGING` is defined. In case it is not defined, the output will be limited to the total number of routed nets and connections, along with the overall count of nodes pushed to and popped from the heap.

The PR would address the issue #2453

### DIFF
--- a/vpr/src/route/connection_router.cpp
+++ b/vpr/src/route/connection_router.cpp
@@ -25,11 +25,17 @@ static bool relevant_node_to_target(const RRGraphView* rr_graph,
                                     RRNodeId node_to_add,
                                     RRNodeId target_node);
 
-static void update_router_stats(const DeviceContext& device_ctx,
-                                const RRGraphView* rr_graph,
-                                RouterStats* router_stats,
-                                RRNodeId rr_node_id,
-                                bool is_push);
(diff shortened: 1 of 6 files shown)


vpr/src/route/connection_router.cpp
```diff
@@ -1113,17 +1100,25 @@ static inline bool relevant_node_to_target(const RRGraphView* rr_graph,
     return false;
 }
 
-static inline void update_router_stats(const DeviceContext& device_ctx,
-                                       const RRGraphView* rr_graph,
+#ifdef VTR_ENABLE_DEBUG_LOGGING
+static inline void update_router_stats(const RRGraphView* rr_graph,
```
vaughnbetz, line 1104: Consider changing the parameter order to mandatory ones first, then ones that are only needed with debug logging.

amin1377:
@vaughnbetz This PR is ready to merge. Thanks!

PR #2461, opened 2023-12-01
StrongIdIterator Operator Overloading

amin1377:
In this PR, the operators that don't use private methods are moved outside of StrongIdIterator class. The "friend" identifier of other operators inside of the class is removed to avoid redefinition of them when an operator is used.

### DIFF
--- a/libs/libvtrutil/src/vtr_strong_id_range.h
+++ b/libs/libvtrutil/src/vtr_strong_id_range.h
@@ -90,7 +90,7 @@ class StrongIdIterator {
 
     ///@brief ~ operator
     template<typename IdType>
-    ssize_t operator-(const StrongIdIterator<IdType>& other) {
+    ssize_t operator-(const StrongIdIterator<IdType>& other) const {
         VTR_ASSERT_SAFE(bool(id_));
         VTR_ASSERT_SAFE(bool(other.id_));
 
@@ -101,19 +101,19 @@ class StrongIdIterator {
 
     ///@brief == operator
     template<typename IdType>
-    bool operator==(const StrongIdIterator<IdType>& other) {
+    bool operator==(const StrongIdIterator<IdType>& other) const {
         return id_ == other.id_;
     }
 
     ///@brief != operator
     template<typename IdType>
-    bool operator!=(const StrongIdIterator<IdType>& other) {
+    bool operator!=(const StrongIdIterator<IdType>& other) const {
         return id_ != other.id_;
     }
 
     ///@brief < operator
     template<typename IdType>
-    bool operator<(const StrongIdIterator<IdType>& other) {
+    bool operator<(const StrongIdIterator<IdType>& other) const {
         return id_ < other.id_;
     }
 
--- a/libs/libvtrutil/src/vtr_strong_id_range.h
+++ b/libs/libvtrutil/src/vtr_strong_id_range.h
@@ -88,26 +88,6 @@ class StrongIdIterator {
         return StrongId(size_t(id_) + offset);
     }
 
-    ///@brief + operator
-    template<typename IdType>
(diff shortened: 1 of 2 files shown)


libs/libvtrutil/src/vtr_strong_id_range.h
```diff
@@ -88,61 +88,59 @@ class StrongIdIterator {
...
+        ssize_t ret = size_t(id_);
+        ret -= size_t(other.id_);
         return ret;
     }
 
     ///@brief == operator
     template<typename IdType>
-    friend bool operator==(const StrongIdIterator<IdType>& lhs, const StrongIdIterator<IdType>& rhs) {
-        return lhs.id_ == rhs.id_;
+    bool operator==(const StrongIdIterator<IdType>& other) {
```
vaughnbetz, line 104: Put a const on this to say the calling object is const.  Then I think the catch2 call will compile.

libs/libvtrutil/src/vtr_strong_id_range.h
```diff
@@ -88,61 +88,59 @@ class StrongIdIterator {
...
+        return id_ == other.id_;
     }
 
     ///@brief != operator
     template<typename IdType>
-    friend bool operator!=(const StrongIdIterator<IdType>& lhs, const StrongIdIterator<IdType>& rhs) {
-        return lhs.id_ != rhs.id_;
+    bool operator!=(const StrongIdIterator<IdType>& other) {
```
vaughnbetz, line 110: Put const on the object 
... other) const {

libs/libvtrutil/src/vtr_strong_id_range.h
```diff
@@ -88,61 +88,59 @@ class StrongIdIterator {
...
+        return id_ != other.id_;
     }
 
     ///@brief < operator
     template<typename IdType>
-    friend bool operator<(const StrongIdIterator<IdType>& lhs, const StrongIdIterator<IdType>& rhs) {
-        return lhs.id_ < rhs.id_;
+    bool operator<(const StrongIdIterator<IdType>& other) {
```
vaughnbetz, line 116: Put a const on the object

libs/libvtrutil/src/vtr_strong_id_range.h
```diff
@@ -88,61 +88,59 @@ class StrongIdIterator {
...
     template<typename IdType>
-    friend ssize_t operator-(
-        const StrongIdIterator<IdType>& lhs,
-        const StrongIdIterator<IdType>& rhs) {
-        VTR_ASSERT_SAFE(bool(lhs.id_));
-        VTR_ASSERT_SAFE(bool(rhs.id_));
-
-        ssize_t ret = size_t(lhs.id_);
-        ret -= size_t(rhs.id_);
+    ssize_t operator-(const StrongIdIterator<IdType>& other) {
```
vaughnbetz, line 93: Mark object as const

libs/libvtrutil/test/test_strong_id.cpp
```diff
@@ -82,7 +82,7 @@ TEST_CASE("StrongIdIterator", "[StrongId/StrongIdIterator]") {
     REQUIRE(c_iter[-4] == b);
     REQUIRE(c_iter[-5] == a);
 
-    REQUIRE((a_iter + 5) == c_iter);
+    REQUIRE(c_iter == (a_iter + 5));
```
vaughnbetz, line 85: This flip should no longer be needed ... good way to test the const.

vaughnbetz (commented):
A few const to add.
Thanks for fixing!

vaughnbetz (commented):
A few const to add.
Thanks for fixing!

amin1377:
@vaughnbetz This PR is ready to merge

PR #2467, opened 2023-12-08
Resolve Bug in Setting Bounding for High Fan-Out Nets

amin1377:
In this pull request, the issue associated with defining the bounding box for high fan-out nets has been addressed. The bounding box for a high fan-out net is now set to encompass the layers through which the net is stretched. This enhancement provides the router with increased flexibility to reach the SINK without the necessity of pushing the entire routing tree to the heap.


vaughnbetz:
vtr_nightly_test7 has a "too good" critical path delay :).  Just needs a golden result update.
Those golden results also seem to be the conflict, which has to be resolved.

vaughnbetz:
You can merge as soon as this passes CI.  Code looks fine.

PR #2472, opened 2023-12-20
Fix: Address warnings in Koios 3D architectures and add additional test

amin1377:
In Koios 3D architectures, there are blocks of the type "tsv_hole" that are not connected to any channel. However, during the construction of the router lookahead, warnings are generated because the pins on these blocks are not connected to any channel. To resolve this issue, the type of these blocks has been changed to "empty."

Additionally, an extra test has been included for the 3D architecture. When placement bounding box is set to"auto" mode, the Titan 3D architecture utilizes per-layer bounding boxes. The newly added test will assess the same architecture with a cube bounding box.


amin1377:
VB: please add a comment in the arch file that the empty blocks are modelling TSVs. They have power or gnd comments, but that is still a bit mysterious without more detail.

PR #2483, opened 2024-02-01
Ability to act as a server using a socket communication link, handle critical path list requests, and highlight certain critical path requests

w0lek:
[vpr_it1_pullrequest.pdf](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/14128999/vpr_it1_pullrequest.pdf)

#### Types of changes

#### Checklist:


vaughnbetz:
Thanks for developing what looks like a great new feature.
1. Please remove the OpenFPGA tilable rr-graph files from the PR; we should focus on the GUI here.
2. Add end user documentation (document the command line option, and this feature in the graphics section (probably a new tab).
3. We should have Doxygen comments on the server etc. code (which I think you already have) and add the resulting hyperlinked documentation to read the docs, probably under VPR API.  There is guidance on how to do that in the VTR Developer Guide (it's a short amount of code to write once you have the Doxygen comments, and the Guide tells you where to write it).

vpr/test/test_server_telegramparser.cpp
soheilshahrouz, line 1: Cover cases where extraction fails

vpr/test/test_server_telegramoptions.cpp
soheilshahrouz, line 1: Add cases where getting values fail

soheilshahrouz (commented):
Thanks for this interesting PR. I suggested a few changes to improve readablity.

Some methods write to standard output and error. It would nice to add a verbosity parameter to control how much detail is printed. I am also not sure accessing std::cout and std::cerr from the network thread is thread-safe.

Since this PR includes some chagnes from OpenFPGA project, I might have missed some files.

soheilshahrouz:
Superseded by #2526.

PR #2492, opened 2024-02-20
Router lookahead file extension

amin1377:
In this PR, the ".bin" extension has been added as an acceptable file extension for the router lookahead file.


amin1377:
@vaughnbetz

Issue #2484, opened 2024-02-05
VPR Placer runtime issue when all design clusters have fixed locations


rachelselinar:
VPR Placer has a large runtime when all the input design clusters have fixed locations.

#### Expected Behaviour
When all clusters are fixed, VPR's place stage should complete very fast as all the clusters are already placed.

#### Current Behaviour
When all clusters have fixed locations, VPR's placer takes very long time to complete. In contrast, when only IO/PLL clusters are fixed, VPR's placer using annealing completes much faster.

#### Example / Steps to Reproduce
Used the Titan23 benchmarks and all designs have similar behavior. Sharing example from 'gaussianblur' design to show how large the runtime difference can be.

(i) All cluster locations are fixed

Command:
_

> vpr gaussianblur.400x296.stratixiv_arch.timing.xml gaussianblur.blif --place --route --timing_analysis on --route_chan_width 300 --max_router_iterations 400 --astar_fac 1 --verify_file_digests off --timing_report_detail aggregated --device 400x296 --sdc_file gaussianblur.sdc --fix_clusters gaussianblur.fix_clusters --net_file gaussianblur.net

_ 

Placement Snippet from vpr_stdout.log:

> Moves per temperature: 2498181
> Warning 567: Starting t: 0 of 105710 configurations accepted.
> 
> ---- ------ ------- ------- ---------- ---------- ------- ---------- -------- ------- ------- ------ -------- --------- ------
> Tnum   Time       T Av Cost Av BB Cost Av TD Cost     CPD       sTNS     sWNS Ac Rate Std Dev  R lim Crit Exp Tot Moves  Alpha
>       (sec)                                          (ns)       (ns)     (ns)                                                 
> ---- ------ ------- ------- ---------- ---------- ------- ---------- -------- ------- ------- ------ -------- --------- ------
>    1 106145.3 0.0e+00   1.000   90098.80 0.00054943 794.866  -7.11e+07 -793.866   0.000  0.0000  399.0     1.00   2498181  0.200
>    2 66540.4 0.0e+00   1.000   90098.80 0.00054943 794.866  -7.11e+07 -793.866   0.000  0.0000  399.0     1.00   4996362  0.950
> ## Placement Quench took **66540.41** seconds (max_rss 22758.9 MiB)
> post-quench CPD = 794.866 (ns) 
> 
> BB estimate of min-dist (placement) wire length: 27029640
> 
> Completed placement consistency check successfully.

[...]

PR #2495, opened 2024-02-27
Fix fixed clusters issue

soheilshahrouz:
#### Description
When a large number of clustered blocks are fixed, placement takes a very long time. This is because for picking a random block, we exhaust all blocks until we find a movable block. If most clustered blocks are fixed, it would take a long time to exhaust fixed blocks by random selection. This PR stores movable blocks in a vtr::vector so that a movable random block can be selected more quickly. It also makes reading the placement file more robust.

#### Related Issue
[Issue 2484](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2484)

#### Types of changes

#### Checklist:


vaughnbetz:
1. @soheilshahrouz tells me there is a bug to fix.
2. Please add a comparison to master (normalize results to master)
3. Check runtime / QoR on the original issue and summarize the result.

vpr/src/base/vpr_context.h
```diff
@@ -386,6 +386,11 @@ struct PlacementContext : public Context {
     ///@brief The pl_macros array stores all the placement macros (usually carry chains).
     std::vector<t_pl_macro> pl_macros;
 
+    ///@brief Stores ClusterBlockId of all movable clustered blocks.
+    std::vector<ClusterBlockId> movable_blocks;
```
vaughnbetz, line 390: Can we remove the blocks_per_type data structures now? It seems these new structures replace them.

vpr/src/place/initial_placement.cpp
```diff
@@ -204,6 +204,11 @@ static void place_all_blocks(const t_placer_opts& placer_opts, vtr::vector<Clust
  */
 static void check_initial_placement_legality();
 
+/**
+ * @brief Fills movable_blocks in global PlacementContext
+ */
+static void fill_movable_blocks();
```
vaughnbetz, line 210: Usually we call this "alloc_and_load_movable_blocks();"

vpr/src/base/vpr_context.h
vaughnbetz: I'd briefly define movable (not locked down to a single location).

vaughnbetz (changes requested):

vpr/src/base/vpr_context.h
soheilshahrouz: > Can we remove the blocks_per_type data structures now? It seems these new structures replace them.

Romoved blocks_per_type and the helper function to access it.



> I'd briefly define movable (not locked down to a single location).

Updated the comment

soheilshahrouz:
@vaughnbetz 
I applied your comments. Ready for review.

vaughnbetz:
Please summarize the relative runtime, cpd, wirelength etc. of the master vs. this fix.
Please run the original issue and post the new runtime. 
I also have a few suggestions on the code; happy to have a zoom meeting if you'd like to discuss.

soheilshahrouz:
| branch | Place time | Place WL | Place CPD |
|----------|----------|----------|----------|
| this PR | 0.992  | 1.00    | 1.00   |
| master  | 1.000  | 1.00    | 1.00   |

vaughnbetz:
Issue runtime: was a few days, now is less than 5 min so it is fixed.

vaughnbetz:
@soheilshahrouz : need to resolve a conflict.

vaughnbetz:
Looks like many seg faults in CI.

soheilshahrouz:
@vaughnbetz 
CI failures are gone now. The issue was that move generators needed to be crated after initial placement, where movable block are determined, but merging master into this branch reversed the order.

Issue #2497, opened 2024-03-05
Designs with many different wire types fail at certain channel widths with an arithmetic exception


WhiteNinjaZ:
When running the ch_intrinsics and diffeq2 benchmarks through the flow on the Xilinx 7 series I am getting a ``Command terminated by signal 8`` error on a channel width of 58. All other channel widths and designs seem to run fine, it is only those two designs at a channel width of 58 where the design terminates pre-maturely. Using Valgrind it looks like this issue is due to certain CB connection patterns getting a zero value for ``max_chan_width`` in ``load_uniform_connection_block_pattern`` from rr_graph.cpp. Since ``max_chan_width`` is used as the modulus of several if statements a division by zero is happening. I believe the cause of this issue is due to VPR trying to assign too many output pins to single wire types at low channel widths. For arches like the 7-Series approximation where 30+ wire segments are used this becomes an issue when only one or 2 types of a wires appear per channel while other wire types are more prevalent. 

#### Expected Behaviour

There should be an if statement in ``load_uniform_connection_block_pattern`` checking if ``max_chan_width`` is zero. 

#### Steps to Reproduce

1. I have prepared a simplified version of the Xilinx-7 Series architecture to demonstrate this issue. All of the blocks used in this arch are copy pasted from k6_frac_N10_frac_chain_mem32K_htree0short_40nm.xml. The only thing different from k6 and this test arch is a Xilinx-like routing structure. The use of a working architecture blocks is done to remove variables other than routing from the equation.
[channel_assignment.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/14502525/channel_assignment.txt)
2. Run this architecture on either ch_intrinsics or diffeq2 using the latest VPR version and default settings with the exception of setting a fixed channel width of 58.
3. Observe the error in vpr.out. A copy of my output is included here:
[vpr.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/14502557/vpr.txt)

PR #2498, opened 2024-03-07
Fix division by zero on arches with many different channel widths

WhiteNinjaZ:
#### Description

Fixes issue #2497 by checking if max_chan_width is zero when assigning segments to connection block.

Updated code run on various problem designs discussed in issue. Update allows flow to successfully complete.

#### Types of changes


vpr/src/route/rr_graph.cpp
```diff
@@ -3723,6 +3723,11 @@ static void load_uniform_connection_block_pattern(vtr::NdMatrix<int, 6>& tracks_
         /* Bi-directional treats each track separately, uni-directional works with pairs of tracks */
         for (int j = 0; j < (Fc / group_size); ++j) {
             int max_chan_width = (((side == TOP) || (side == BOTTOM)) ? x_chan_width : y_chan_width);
+
+            // if the number of tracks we can assign is zero break from the loop
```
vaughnbetz, line 3727: Suggest saying why you are breaking out of the loop (no switches to allocate)

Issue #2504, opened 2024-03-14
CI Test Failures on Master


AlexandreSinger:
Currently the CI is failing on Master without changes on the VTR side. The errors appear to all be related to ODIN II, since the CI tests that are failing are ODIN II regression tests.

See Test #8008

https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/8273631785

Subsequent PRs are also failing with the same test failures.

#### S: Basic

This test fails when running the `vtr_reg_basic_odin` VTR task. I have run this task on my local development machine, and the test failures on the CI do not appear.

The failures appear to be a mix of ODIN II failures and VPR failures.

#### ODIN-II Basic Test

This test fails when running `odin_ii/regression_test/benchmark/suite/light_suite`.

There are a large number of test failures in this file. Some of them comparing a signed, 32-bit 0 to a 1-bit 0 and saying that is incorrect.

soheilshahrouz:
Apparently, this is a github actions [issue](https://github.com/actions/runner-images/issues/9491), and [the fix](https://github.com/actions/runner-images/pull/9513/files) has already been merged into master.

soheilshahrouz:
Closing the issue as the PR was merged.

AlexandreSinger:
Thank you so much Soheil!

PR #2506, opened 2024-03-16
Fix CI failures in master

soheilshahrouz:
#### Description
CI  has been failing in master since two days ago. This PR fixes the problem. This is not the cleanest fix, but the alternative is to wait until next Friday when Github deploys the fix into their Ubuntu image.

#### Related Issue
[Issue](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2504)

#### Motivation and Context
Preventing CI test from failing in master.

Two CI tests fail in the master branch, but not in this one.

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -303,6 +303,17 @@ jobs:
         # We skip QoR since we are only checking for errors in sanitizer runs
         ./run_reg_test.py ${{ matrix.suite }} -show_failures -j2 -skip_qor
 
+    - name: Upload regression run files
+      uses: actions/upload-artifact@v3
+      with:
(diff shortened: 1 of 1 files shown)


vaughnbetz:
Nice detective work -- thanks!
Wow, that is one weird fix. Can you add a comment saying why this is needed / what it does.  Even if you don't fully understand why this is needed (and I certainly don't) writing down what you do know would be useful.

soheilshahrouz:
@vaughnbetz 
I added a comment explaining the problem and how the fix works. It's ready to be merged.

vaughnbetz:
Thanks!

Issue #2507, opened 2024-03-16
 Failed to build target 'libarchfpga'


imagan1:
#### Expected Behaviour

When I execute the make command, it should be 100% successful
#### Current Behaviour

when using the make command to build to 83% progress, the following error message appeared:
[ 83%] Built target libarchfpga
Consolidate compiler generated dependencies of target read_arch
[ 83%] Linking CXX executable read_arch
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/8/libstdc++fs.a(std-path.o): in function `std::filesystem::__cxx11::filesystem_error::~filesystem_error()':
(.text._ZNSt10filesystem7__cxx1116filesystem_errorD2Ev+0x48): undefined reference to `std::filesystem::__cxx11::path::~path()'
/usr/bin/ld: (.text._ZNSt10filesystem7__cxx1116filesystem_errorD2Ev+0x90): undefined reference to `std::filesystem::__cxx11::path::~path()'
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/8/libstdc++fs.a(std-path.o): in function `std::filesystem::__cxx11::path::_M_trim()':
...
make[2]: *** [libs/libarchfpga/CMakeFiles/read_arch.dir/build.make:105：libs/libarchfpga/read_arch] error 1
make[1]: *** [CMakeFiles/Makefile2:1850：libs/libarchfpga/CMakeFiles/read_arch.dir/all] error 2
make: *** [Makefile:146：all] error 2
#### Possible Solution

The above error seems to suggest an issue with my c++library libstd++fs. Therefore, I checked my gcc and g++, version 8.3.0, which should support the C++17 language standard. In addition, I created a separate main.cpp file that called a function in the namespace "filesystem" and used the command "g++- stdc++=c++17 main.cpp - lstdc++fs" to build it, and the result was successful. This indicates that my system should meet the conditions for the successful construction of the VTR project, but why did the above problem occur? How to solve it?
maybe the CMakeLists.txt in the path (VTR ROOT)/libs/libarchfpga  needs improvents for linking the C++standard library for gcc or g++ witch version is lower than gcc 9.0
#### Steps to Reproduce

1. my os is [deepin 20.9 for linux](https://distrowatch.com/table.php?distribution=deepin), with python 3.7.3, gcc 8.3, g++ 8.3
2. source files were download last week.
3. building process: according to README.md in the root dir.

#### Context

[...]

PR #2508, opened 2024-03-17
Link stdc++fs for older gcc versions

soheilshahrouz:
Fixes [this issue](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2507).

According to [filesystem library documentation](https://en.cppreference.com/w/cpp/filesystem), gcc versions older than 9.1 and clang prior to 9.0 need filesystem library to be explicitly linked. This PR adds the compiler option when it is necessary.

When trying to reproduce the compilation failure, I realized gcc-8.4 does not need the additional compiler option. However, to be consistent with the official documentation, I add the compiler flag for versions older than 9.1.

Current Github Ubuntu images do not have gcc-8 installed by default. When I manually installed gcc-8.4, VTR was built without any problem. It seems that it is only gcc-8.3 and older version sthat need the link flag. I reproduced the compilation error on my local machine by building gcc-8.3 from source. Adding the compiler flag fixed the error. CI tests show that the flag is not applied for newer gcc versions.

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -404,6 +404,11 @@ jobs:
         submodules: 'true'
     - run: ./.github/scripts/install_dependencies.sh
 
+    - uses: egor-tensin/setup-gcc@v1
+      with:
+        version: 8
+        platform: x64
+
     - uses: hendrikmuhs/ccache-action@v1.2
 
     - name: Test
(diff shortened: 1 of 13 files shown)


vaughnbetz:
Looks good, thanks!

PR #2511, opened 2024-03-20
[CAPNPROTO][CMake] Fixed Build When CAPNPROTO is Disabled

### LINKED ISSUE #2510: Disabling CAPNPROTO Crashes Build
When building VTR using the following command:

```
make CMAKE_PARAMS="-DVTR_ENABLE_CAPNPROTO=OFF"
```

The build crashes in libarchfpga

There appears to be a dependency in `read_fpga_interchange_arch.cpp` on CAPNPROTO that does not handle the case when it is not available.

#### Expected Behaviour
Build should finish without issue.

#### Current Behaviour
image

#### Possible Solution
In libs/CMakeLists.txt the libvtrcapnproto library is not even being compiled when the VTR_ENABLE_CAPNPROTO option is set to OFF.
image

This in itself is confusing in my opinion. The code to disable creating the library should really be in libvtrcapnproto/CMakeLists.txt. But also I am not sure why we would not even want to declare building the library. I believe that CMake would not build the library unless an executable that is being built depends on it. But I would need to verify this.

However, in libarchfpga/CMakeLists.txt libvtrcapnproto library is declared as a dependency:
image

And within libarchfpga, files such as read_fpga_interchange_arch.h include CAPNPROTO code without any IFDEFs protecting it when VTR_ENABLE_CAPNPROTO is OFF.

image

This comes together to yield that when VTR_ENABLE_CAPNPROTO is OFF, libvtrcapnproto is never created; which leads to issues when libarchfpga is being built.

In order to fix this, libarchfpga should only link libvtrcapnproto if VTR_ENABLE_CAPNPROTO is ON, and any files in libarchfpga that require capnp should handle the case when CAPNPROTO is off.

#### Your Environment

* VTR revision used: d9644e87e0dc22235c431e75f65813bf1622e9a6
* Operating System and version:
* Compiler version:

AlexandreSinger:
There were many build errors that arose when CAPNPROTO was disabled. It looks like there was a dependency in libarchfpga on libvtrcapnproto even though the library is disabled if the user passes a flag to disable it. Removed this dependency when the flag is disabled.

VPR also had a similar issue. This appears to have been handled in the past, but has since regressed. We should consider adding this build to CI to prevent further regressions (if this feature is needed).

#### Related Issue

resolves #2510 

Built using:
```
make CMAKE_PARAMS="-DVTR_ENABLE_CAPNPROTO=OFF"
```
and
```
make CMAKE_PARAMS="-DVTR_ENABLE_CAPNPROTO=ON"
```

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/CMakeLists.txt
+++ b/libs/libarchfpga/CMakeLists.txt
@@ -22,8 +22,12 @@ target_link_libraries(libarchfpga
                         libvtrutil
                         libpugixml
                         libpugiutil
-                        libvtrcapnproto
 )
+
+if(${VTR_ENABLE_CAPNPROTO})
+    target_link_libraries(libarchfpga libvtrcapnproto)
+    target_compile_definitions(libarchfpga PRIVATE VTR_ENABLE_CAPNPROTO)
+endif()
 
 # Using filesystem library requires additional compiler/linker options for GNU implementation prior to 9.1
 # and LLVM implementation prior to LLVM 9.0;
--- a/libs/libarchfpga/src/read_fpga_interchange_arch.cpp
+++ b/libs/libarchfpga/src/read_fpga_interchange_arch.cpp
@@ -1,3 +1,10 @@
+
+
+#include "read_fpga_interchange_arch.h"
+#include "vtr_error.h"
+
+#ifdef VTR_ENABLE_CAPNPROTO
+
 #include <algorithm>
 #include <kj/std/iostream.h>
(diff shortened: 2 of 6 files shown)


AlexandreSinger:
@vaughnbetz This passes CI, please review when you get the chance!

vaughnbetz:
Thanks Alex.  Yes, if you can add a build test that would be good so we can prevent future regressions.

Issue #2499, opened 2024-03-13
Clean up rr_node_route_inf


vaughnbetz:
1. We have a dead member variable (target_flag) in rr_node_route_inf (not used anymore since we deleted the breadth-first router). We should delete it. 
2. The comment for the prev_edge says it is the index within the switch list of the predecessor node used to reach the current node. I'm pretty sure that is not correct anymore, since we now have globally unique edge ids (do we use that large id instead)? We should check and correct the comment if needed.
3. If we use globally unique edge ids we could use that to find the predecessor node, and not store it anymore. This should be investigated to see if it saves memory / speeds up the router (or if it slows it down, don't do it).
```

/**
 * @brief Extra information about each rr_node needed only during routing
 *        (i.e. during the maze expansion).
 *
 *   @param prev_node  Index of the previous node (on the lowest cost path known
 *                     to reach this node); used to generate the traceback.
 *                     If there is no predecessor, prev_node = NO_PREVIOUS.
 *   @param prev_edge  Index of the edge (from 0 to num_edges-1 of prev_node)
 *                     that was used to reach this node from the previous node.
 *                     If there is no predecessor, prev_edge = NO_PREVIOUS.
 *   @param acc_cost   Accumulated cost term from previous Pathfinder iterations.
 *   @param path_cost  Total cost of the path up to and including this node +
 *                     the expected cost to the target if the timing_driven router
 *                     is being used.
 *   @param backward_path_cost  Total cost of the path up to and including this
 *                     node.
 *   @param target_flag  Is this node a target (sink) for the current routing?
 *                     Number of times this node must be reached to fully route.
 *   @param occ        The current occupancy of the associated rr node
 */
struct t_rr_node_route_inf {
    RRNodeId prev_node;
    RREdgeId prev_edge;

    float acc_cost;
    float path_cost;
    float backward_path_cost;

    short target_flag;

  public: //Accessors
    short occ() const { return occ_; }

  public: //Mutators
    void set_occ(int new_occ) { occ_ = new_occ; }

  private: //Data
    short occ_ = 0;
};



```

AlexandreSinger:
The `acc_cost` and `occ_` members appear to not be used in the maze expansion (against what the comment above this class is saying). We should separate out the per-iteration route info from the maze-expansion route info. Beyond organization, this would reduce the amount of data being stored in the cache during maze expansion and also reduce the number of global variables the maze expansion touches.

[...]

PR #2512, opened 2024-03-20
[Route][Heap] Removed prev_node From t_heap

AlexandreSinger:
A previous commit removed prev_node from t_rr_node_route_inf. This was because the prev_node can be uniquely identified by the prev_edge (but not the other way around).

The same methedology can be applied to the t_heap type. This was only being set, but it was never being used (with the exception of one debug message). Therefore this is only good for performance and memory overhead.

#### Related Issue

issue #2499 

#### Motivation and Context

This cleans up the t_heap struct some, which reduces memory overhead. This makes it easier to work with the heap and also may improve performance slightly.

#### Types of changes

#### Checklist:


AlexandreSinger:
@vaughnbetz Please review when you have a moment. This only removes and does things from the heap expansion so I do not think its runtime needs to be measured.

vaughnbetz:
Looks good.  Always possible it sped things up or saved noticable memory, so you could run an experiment if you wanted to see (but probably wouldn't speed up much). Merging and I'll let you decide if you want to measure for curiosity's sake ...

PR #2513, opened 2024-03-20
[CI][Capnproto] Added Basic Regression Test with CAPNPROTO Disabled

AlexandreSinger:
There has been some regressions with regard to CAPNPROTO being disabled since all of the CI tests have CAPNPROTO enabled.

Added a basic regression test to the CI which will run the vtr_reg_basic tests with CAPNPROTO disabled. This will test both that everything builds and basic VPR tests will run.

In the future we should test direct features that disabling CAPNPROTO would affect (such as writing and reading routing files); however, for now this will prevent the regressions previous seen.

#### Related Issue

issue #2510 

#### Motivation and Context

In a previous PR (PR #2511 ) fixed regressions to when CAPNPROTO is disabled. In the PR I mentioned that there should be regression tests in the CI to prevent this in the future. Added basic regression tests to the CI.

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -198,6 +198,11 @@ jobs:
             suite: 'vtr_reg_basic'
           },
           {
+            name: 'Basic with CAPNPROTO disabled',
+            params: '-DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on -DVTR_ENABLE_CAPNPROTO=off',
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
@vaughnbetz This passed CI, and I confirmed that it is running the new test suite. Please review and merge when you are available.

vaughnbetz:
Thanks!

PR #2514, opened 2024-03-21
Bug fix: asymmetric grid with bidirectional routing

ThomasJagielski:
Replace grid.width() check with grid.height() for top perimeter channels

#### Description
For FPGA fabrics with non-square grids and bidirectional routing the check to determine if you were on a perimeter tile was incorrect. The check compared the (x,y) location against the grid.width() for both dimensions. The vertical dimension should be compared against the grid.height().

#### Motivation and Context
Before this change, asymmetric grid dimensions with bidirectional routing would result in a "node_index" assertion error and a "core dump" which aborted vpr execution while creating the routing resource graph. This change will now allow you to run vpr on non-square grids with bidirectional routing.

All regression tests were rerun and passed locally. A simple example with an asymmetric grid and bidirectional routing that previously resulted in a core dump and node_index assertion error was also run through the vpr flow successfully on my local machine. 

#### Types of changes

#### Checklist:


vaughnbetz:
Thanks!  Will merge once CI passes.

PR #2515, opened 2024-03-22
Fixed a Couple Simple Warnings

AlexandreSinger:
During the CI build process of the new "R: Basic with CAPNPROTO disabled" job, there were a couple of basic warnings that I thought I would resolve.

The first were some warnings directly caused by CAPNPROTO being disabled:
image
These were easy to fix by commenting out the unused variables.

Next were some CMake warnings which appear to be affecting all builds:
image
It looks like the [exec_program](https://cmake.org/cmake/help/latest/command/exec_program.html) command was deprecated and the suggested replacement was to use the [execute_process](https://cmake.org/cmake/help/latest/command/execute_process.html#command:execute_process) command instead. The syntax was similar, but required a few extra arguments in order to properly get the Git version.

#### Motivation and Context

Some build warnings that showed up that caught my eye and I knew were simple to solve.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/route/router_lookahead_extended_map.cpp
+++ b/vpr/src/route/router_lookahead_extended_map.cpp
@@ -606,10 +606,10 @@ float ExtendedMapLookahead::get_expected_cost(
 
 #ifndef VTR_ENABLE_CAPNPROTO
 
-void ExtendedMapLookahead::read(const std::string& file) {
+void ExtendedMapLookahead::read(const std::string& /*file*/) {
     VPR_THROW(VPR_ERROR_ROUTE, "MapLookahead::read not implemented");
 }
-void ExtendedMapLookahead::write(const std::string& file) const {
+void ExtendedMapLookahead::write(const std::string& /*file*/) const {
     VPR_THROW(VPR_ERROR_ROUTE, "MapLookahead::write not implemented");
 }
 
--- a/vpr/src/route/router_lookahead_extended_map.cpp
+++ b/vpr/src/route/router_lookahead_extended_map.cpp
@@ -604,26 +604,25 @@ float ExtendedMapLookahead::get_expected_cost(
     }
 }
 
-#ifndef VTR_ENABLE_CAPNPROTO
-
-void ExtendedMapLookahead::read(const std::string& /*file*/) {
-    VPR_THROW(VPR_ERROR_ROUTE, "MapLookahead::read not implemented");
-}
-void ExtendedMapLookahead::write(const std::string& /*file*/) const {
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
@vaughnbetz Please review when you have a moment.

vpr/src/route/router_lookahead_extended_map.cpp
```diff
@@ -606,10 +606,10 @@ float ExtendedMapLookahead::get_expected_cost(
 
 #ifndef VTR_ENABLE_CAPNPROTO
 
-void ExtendedMapLookahead::read(const std::string& file) {
+void ExtendedMapLookahead::read(const std::string& /*file*/) {
     VPR_THROW(VPR_ERROR_ROUTE, "MapLookahead::read not implemented");
 }
-void ExtendedMapLookahead::write(const std::string& file) const {
+void ExtendedMapLookahead::write(const std::string& /*file*/) const {
```
soheilshahrouz, line 612: Instead of having two method definitions and disabling one of them, we can write a single definition and use `#ifdef` to disable specific parts of the code within the function body. We can suppress unused argument warnings by employing the `[[maybe_unused]]` attribute.

vaughnbetz: That seems like a good idea to me. @AlexandreSinger  what do you think?

AlexandreSinger: I agree completely with putting the IFDEFs within the function declarations, as this would help prevent regressions. I can transition the code in this file to do that.

I do not think we should use the [[maybe_unused]] attribute. We still want to know if an argument goes unused when we do not intend it to. We can simply just void the arguments within the function like this:
```
void foo(int arg1, int arg2) {
#IFNDEF
  // Original implementation
#ELSE
  (void)arg1;
  (void)arg2;
  VPR_THROW(VPR_ERROR_ROUTE, "...");
#ENDIF
}
```

This way the arguments are explicitly thrown away and the warning will still be generated in the original implementation (signifying there is likely an error).

AlexandreSinger: Updated. @soheilshahrouz Let me know if this is what you had in mind!

soheilshahrouz: I think void casting is a better practice because we can catch cases where the parameter is actually ununsed.

AlexandreSinger:
@vaughnbetz I have resolved Soheil's comments and it passed CI again. Please review.

Issue #2518, opened 2024-03-27
Remove Warnings in VTR CI Builds


AlexandreSinger:
There are some warnings in the VTR build which can make it challenging to develop since its sometimes hard to tell if they are new or not. Using this issue to collect the different warnings currently in VTR and track progress on removing them.

To look for warnings, I am using the CI runs to tell where warnings currently show up. The ultimate goal is to remove all warnings in the CI. I got a lot of them, but if people find more, feel free to add to the list! If one is resolved, feel free to check it off!

Most recently checked CI run: https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/8444847760

### Known warnings in most CI builds:

- [x] warning: -jN forced in submake: disabling jobserver mode.
![Screenshot from 2024-03-29 16-02-42](https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/e11200eb-99d9-42ea-85b6-452f286f4820)
This comes up in builds that use multiple threads to make VTR. This seems to be an issue with the way we are using the make files.

- [x] fatal: not a git repository
![Screenshot from 2024-03-27 11-47-46](https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/23250e5a-9f31-4c3a-8ea9-9064c2f01f02)
This one is showing up in a lot of the builds and is showing up all over the place. I cannot trace where this is coming from but I think this is comming from CMake somewhere. We should investigate what is causing it and try to resolve this issue.
Maybe mentioned in issue #2402 but not explicitly the cause of the problem. May be a side-effect.

- [ ] libsdcparse: Parser Deprecated and "fix-its can be applied"
![Screenshot from 2024-03-27 11-52-22](https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/61ffcf09-7ec7-48ba-a472-4ec88f7428b1)
This looks like a very simple fix, but because it is in an external library we may have to fix it in a weird way by contributing to the upstream somehow. The second warning "fix-its can be applied" may be the key to resolving this quickly. Perhaps running "--update" would fix this automatically?

- [ ] libblifparse: Parser Deprecated and "fix-its can be applied"
![Screenshot from 2024-03-27 11-53-36](https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/b5cd5852-5061-475d-9a80-5da26471a66f)
[...]

PR #2520, opened 2024-03-29
[Yosys][Warnings] Fixed Fatal Not Git Repo Warning

AlexandreSinger:
Fixed a warning where "fatal: not a git repository: './/.git' was being printed during builds of VPR.

This is caused by Yosys being included as a subtree instead of a git submodule. The Makefile in Yosys is trying to get the current version of git for Yosys but it cannot do it since it is not a git repo.

Just commented out the line causing the warning for now.

#### Description

This was actually a very interesting warning to debug.

I started by debugging the make command without using multiple threads to see what CMake command was causing this warning:
```
make -d &> temp.dump
```

Looking into the temp.dump file, I found that the warning was being printed when generating "yosys-bin":
image
Not only that, but it was being printed twice.

I then opened up yosys' CMakeList to find this beauty:
image
It looks like all this CMake is doing is running another Makefile, twice. I should note that the "-f" argument should be set here since this is, in-fact, an "out-of-tree" build.

Looking into Yosys' Makefile, we find the culprit:
image
`YOSYS_SRC` is a variable that should point to the directory of the current Makefile as defined here:
image
However, since the "-f" argument was not set before, this gets initialized to "./"; hence the warning. Setting the "-f" argument does not fix the warning however, it just makes it better:
image
Obviously, we still get the same warning since this is not a git repository, since it was included into VTR as a git subtree. To fix this issue I just commented out the line causing this issue since all it was doing was failing to find the git version. Perhaps the Yosys infrastructure should be revisited if we plan to support this long term.

#### Related Issue

See issue #2518

### DIFF
--- a/yosys/Makefile
+++ b/yosys/Makefile
@@ -149,7 +149,8 @@ YOSYS_VER := 0.32
 # back to calling git directly.
 TARBALL_GIT_REV := $(shell cat $(YOSYS_SRC)/.gitcommit)
 ifeq ($(TARBALL_GIT_REV),$$Format:%h$$)
-GIT_REV := $(shell GIT_DIR=$(YOSYS_SRC)/.git git rev-parse --short=9 HEAD || echo UNKNOWN)
+#GIT_REV := $(shell GIT_DIR=$(YOSYS_SRC)/.git git rev-parse --short=9 HEAD || echo UNKNOWN)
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
@vaughnbetz I have verified that this removed the warning from the builds. Please review when you have a moment.

vaughnbetz:
Thanks Alex!

Issue #2517, opened 2024-03-27
clang/LLVM-17 build


heshpdx:
Hello friends,
Can you check that LLVM-17 and LLVM-18 are are able to build VPR? @amin1377 and I are working on the SPEC CPU benchmark version of VPR and the next step is to show it works on a variety of systems and compilers. We are running into issues with the latest LLVM.

I've attached a log file below that shows our build issues. Although our source base is a little behind the mainline, I think the right thing to do is to fix the mainline and let that trickle down to us in SPEC CPU.

Majority of errors look like this:
```
In file included from vtr-vpr/libs/librrgraph/src/base/rr_graph_storage.cpp:2:
In file included from vtr-vpr/libs/libarchfpga/src/arch_types.h:11:
In file included from vtr-vpr/libs/libarchfpga/src/logic_types.h:14:
In file included from /spec/compilers/llvm-17.0.6/bin/../include/c++/v1/vector:321:
In file included from /spec/compilers/llvm-17.0.6/bin/../include/c++/v1/__format/formatter_bool.h:17:
In file included from /spec/compilers/llvm-17.0.6/bin/../include/c++/v1/__format/concepts.h:17:
In file included from /spec/compilers/llvm-17.0.6/bin/../include/c++/v1/__format/format_parse_context.h:16:
In file included from /spec/compilers/llvm-17.0.6/bin/../include/c++/v1/string_view:1059:
In file included from /spec/compilers/llvm-17.0.6/bin/../include/c++/v1/algorithm:1803:
In file included from /spec/compilers/llvm-17.0.6/bin/../include/c++/v1/__algorithm/nth_element.h:15:
/spec/compilers/llvm-17.0.6/bin/../include/c++/v1/__algorithm/sort.h:577:5: error: indirection requires pointer operand ('const edge_sort_iterator' invalid)

  577 |     *__begin = _Ops::__iter_move(__pivot_pos);
      |     ^~~~~~~~
 ```

[CPUv8.026.log](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/14775849/CPUv8.026.log)

Thank you!

AlexandreSinger:
Hi @heshpdx ,

[...]

PR #2522, opened 2024-04-01
[RR_GRAPH] Fixed Bug in LLVM17 Build

AlexandreSinger:
The std::sort function requires operands to be "pointer-like", this requires that de-referencing a const pointer should be allowed.

This was not an issue in the past, but in LLVM17 this property is required. The edge_sort_iterator did not follow this property. This has been fixed.

#### Description
See issue #2517 

The LLVM17 build failed due to a const edge_sort_iterator unable to be de-referenced. This PR fixes that issue.

The problem is that this required the `swapper_` member to be a pointer to allow this to work. This means that the `swapper_` object needs to be declared on the heap. This may have affects on the performance of this implementation. To prevent regression in performance, this needs to be profiled to ensure we do not slow down the current high-priority builds for future builds.

#### Related Issue
issue #2517 

Ran the quick and hard regression tests.

Still need to profile to ensure the performance does not regress.

### DIFF
--- a/libs/librrgraph/src/base/rr_graph_storage.cpp
+++ b/libs/librrgraph/src/base/rr_graph_storage.cpp
@@ -155,6 +155,13 @@ class edge_sort_iterator {
     using pointer = edge_swapper*;
     using difference_type = ssize_t;
 
+    // In order for this class to be used as an iterator within the std library,
+    // it needs to "act" like a pointer. One thing that it should do is that a
+    // const variable of this type should be de-referenceable. Therefore, this
+    // method should be const method; however, this requires modifying the class
+    // and may yield worst performance. For now the std::stable_sort allows this
+    // but in the future it may not. If this breaks, this is why.
+    // See issue #2517 and PR #2522
     edge_swapper& operator*() {
         return this->swapper_;
     }
@@ -419,7 +426,7 @@ size_t t_rr_graph_storage::count_rr_switches(
     // values.
     //
     // This sort is safe to do because partition_edges() has not been invoked yet.
-    std::sort(
+    std::stable_sort(
         edge_sort_iterator(this, 0),
         edge_sort_iterator(this, edge_dest_node_.size()),
         edge_compare_dest_node());
@@ -527,7 +534,7 @@ void t_rr_graph_storage::partition_edges(const vtr::vector<RRSwitchId, t_rr_swit
     //    by assign_first_edges()
     //  - Edges within a source node have the configurable edges before the
     //    non-configurable edges.
-    std::sort(
+    std::stable_sort(
         edge_sort_iterator(this, 0),
         edge_sort_iterator(this, edge_src_node_.size()),
         edge_compare_src_node_and_configurable_first(rr_switches));


heshpdx:
Thanks for this! I have not yet checked if this patch fixes the LLVM-17 build issue, but I will send it over to the tester very soon. But it does build and run for gcc-12 and llvm-15, which is what I have on my system. So, I checked the performance with this patch on the cmdlines that @amin1377 and I came up with for SPEC CPUv8. Indeed, there is about a 10% perf loss across a variety of configs/builds.

Here is the data reported in seconds, on an aarch64 based Ampere Altra 3.0 Ghz system running Ubuntu 22.04.

```
O3:  gcc-12.3     -g -O3    -mcpu=native -fno-fast-math
Of:  gcc-12.3     -g -Ofast -mcpu=native -fno-fast-math -flto=32
L15: clang-15.0.7 -g -O3                 -fno-fast-math

baseline  O3    Of    L15
--------------------------
place1    63    59     65
route1    98    84    100
place2    79    71     79
route2    64    53     63

w/patch   O3    Of    L15
--------------------------
place1    76    63     75
route1   108    88    107
place2    87    75     86
route2    69    57     68

ratio     O3    Of    L15
--------------------------
place1   1.21  1.07  1.15
route1   1.10  1.05  1.07
place2   1.10  1.06  1.09
route2   1.08  1.08  1.08

```

AlexandreSinger:
I spent the last day running the large Titan benchmarks, and confirmed what @heshpdx has found:
image

An 11% increase in overall runtime, no change in QoR. Unless we can find a more efficient way to make the de-reference operator a const method in this class, I do not think this can be merged into VPR master.

We could IFDEF the old version with the new version based on the version of C++ being used to compile, as an idea.

heshpdx:
We confirmed that the patch does allow compilation on llvm-17, and also confirmed a perf slowdown. 

Can you try reverting the patch and then converting all of the instances of `std::sort` to `std::stable_sort`? Because that also allows build with llvm-17 and doesn't incur the performance slowdown. It is arguably better since it would provide same results between standard library implementations, which is a good thing.

AlexandreSinger:
@heshpdx I have reverted the changes and converted the std::sorts to std::stable_sorts. I have also rerun the profiling of the large Titan benchmarks and have shown only a minor slowdown of around 1%.
image

@vaughnbetz What do you think about this change?

vaughnbetz:
Thanks. This change seems OK to me.

heshpdx:
FYI, for SPEC CPUv8 we changed ALL of the sort's into stable_sort's, everywhere in the source. Maybe it was overkill, but it's a defensive move.

vaughnbetz:
@amin1377 : could you try running a cpu time experiment with the remaining sorts turned into stable sorts?  If it doesn't hurt cpu time, we can convert them all.  I'd like to keep the master in sync with SPEC CPUv8, with #ifdef SPEC or some such guarding any minor changes we can't turn on for the master (e.g. making sure all the base_costs are integral).

heshpdx:
I believe the patch below is the real fix. I can confirm that this builds for LLVM17 (and 18 and 19.1.0-rc3). BTW we have found which sorts need to be stable and which ones do not; when I get time I can share a patch that recovers performance.
```
--- a/src/vtr-vpr/libs/librrgraph/src/base/rr_graph_storage.cpp
+++ b/src/vtr-vpr/libs/librrgraph/src/base/rr_graph_storage.cpp
@@ -154,7 +154,7 @@ class edge_sort_iterator {
     using pointer = edge_swapper*;
     using difference_type = ssize_t;
 
-    edge_swapper& operator*() {
+    edge_swapper& operator*() const {
         return this->swapper_;
     }
 
@@ -229,7 +229,7 @@ class edge_sort_iterator {
     }
 
   private:
-    edge_swapper swapper_;
+    mutable edge_swapper swapper_;
 };
 
 class edge_compare_src_node_and_configurable_first {
```

PR #2523, opened 2024-04-02
Set the maximum pres_fac

amin1377:
This pull request introduces a new command line argument to set the maximum `pres_fac`, which is now defaulted to 1000.

The `pres_fac` coefficient is used to adjust the cost of an overused net by multiplying it with the number of nets overusing it. This adjustment encourages the router to explore alternative routes. Previously, the maximum value for this coefficient was set to an extremely large number, resulting in strange router behavior such as excessive ripping up connections toward the end of routing. To address this issue, the default maximum value is now set to 1000, providing more predictable router behavior. Additionally, users have the flexibility to adjust this value as needed via the command line.

### DIFF
--- a/README.developers.md
+++ b/README.developers.md
@@ -1330,7 +1330,7 @@ Instead changes should be made in the relevant up-stream repository, and then sy
 
     For example to update the `libtatum` subtree:
     ```shell
-    ./dev/external_subtrees.py --update libtatum
+    ./dev/external_subtrees.py --update libtatum -m "commit message describing why component is being updated"
     ```
 
 ## Adding a new Subtree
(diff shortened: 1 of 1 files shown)


vaughnbetz:
QoR data?
The one CI failure is a small circuit (alu4) that has a higher critical path delay. Not a real concern so OK.

doc/src/vpr/command_line_usage.rst
```diff
@@ -1163,6 +1163,12 @@ VPR uses a negotiated congestion algorithm (based on Pathfinder) to perform rout
 
     **Default:** ``1.3``
 
+.. option:: --max_pres_fac <float>
+
+    Sets the maximum present overuse penalty factor.
```
vaughnbetz, line 1168: that can ever result during routing. Should always be less than 1e25 or so to prevent overflow. Smaller values may help prevent circuitous routing in difficult routing problems, but may increase the number of routing iterations needed and hence runtime.

vpr/src/route/route.cpp
```diff
@@ -441,8 +441,8 @@ bool route(const Netlist<>& net_list,
         } else {
             pres_fac *= router_opts.pres_fac_mult;
 
-            /* Avoid overflow for high iteration counts, even if acc_cost is big */
-            pres_fac = update_draw_pres_fac(std::min(pres_fac, static_cast<float>(HUGE_POSITIVE_FLOAT / 1e5)));
+            /* Set the maximum pres_fac to the value passed by the command line argument */
+            pres_fac = update_draw_pres_fac(std::min(pres_fac, router_opts.max_pres_fac));
```
vaughnbetz, line 445: Let's rename this function to something a bit more descriptive. update_pres_fac_and_store_for_draw ().
Or move pres_fac to the routing context if it's cleaner.

vaughnbetz (changes requested):
Small change again.

vaughnbetz:
@amin1377 : please merge once CI passes.

vaughnbetz:
Thanks Amin. The code all looks good.  Just for safety, can you run a VTR minimum channel width regtest to confirm QoR (min chan width, routing time, etc.) look OK for it?

vaughnbetz:
Looks good, thanks. I'll merge as soon as CI is green.

PR #2524, opened 2024-04-02
Fix the crash in graphics when run-flat is used

amin1377:
In this PR, the graphic is updated to not crash when run-flat is used. This PR **doesn't** show the routing inside cluster, but the user can still interact with the GUI and the congestion, routing tracks, etc of the inter-cluster resources.


vaughnbetz (changes requested):

vaughnbetz (changes requested):
A couple changes.

PR #2529, opened 2024-04-10
[Bison] Raised Minimum Bison Version to 3.3

AlexandreSinger:
Raised the minimum Bison version from 3.0 to 3.3 to handle deprecation warnings.

#### Description

In recent versions of Bison, the `parser_class_name` directive has been deprecated for `api.parser.class`.

See issue #2518 

The issue is that `api.parser.class` was not introduced until Bison 3.3. Currently the minimum Bison version for libblifparse and libsdcparse is Bison 3.0 (which is the same version we use on our development machines). To properly fix these warnings we would need to raise the minimum Bison version to Bison 3.3 and resolve the deprecation.


AlexandreSinger:
@vaughnbetz What do you think about this? To prevent having to raise the minimum Bison version to Bison 3.3, I have suppressed the deprecation warning only in the two affected yacc files. I have verified that this fixes the warnings.

vaughnbetz:
This is a good question for the Thursday meeting in my opinion (if you can bring up a quick summary).  It is a bit dangerous to suppress a deprecation warning, but I guess it will error if they ever fully deprecate it ...

AlexandreSinger:
@vaughnbetz I agree, I think this is a great topic to bring up! However, I will not be at the one this week. I could bring it up next week! I do not think this is very time sensitive.

vaughnbetz:
Two weeks is fine.  You could bring up any thorny ones there, plus overall status on getting warning clean.

AlexandreSinger:
Now that all CI and development machines are on a recent enough version of Ubuntu, rerunning CI on this PR and will merge when it passes.

PR #2530, opened 2024-04-11
[Yosys][Warnings] Fixed Make Being Called With Multiple Threads

AlexandreSinger:
CMake was calling another Make command, but was explicitly passing the number of threads to use. Make files should not pass the number of threads to use since it may cause more threads to be consumed than what the user intended.

For example, if the user used `make -j10`, the parent make would use 10 threads, and the child make would also use 10 threads (20 threads in total); but this is not what the user wanted. I think CMake recognized this and was suppressing this behaviour anyways. I believe that CMake will pass the number of threads for the child make to use anyways, but in a way that would prevent the above issue.

Removed the ability to set the number of threads for Yosys builds to handle the warning.

See issue #2518

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -418,11 +418,13 @@ if(${WITH_PARMYS}) # define cmake params to compile Yosys
         set(MAKE_PROGRAM "make")
     endif()
 
-    if(NOT DEFINED "${CMAKE_BUILD_PARALLEL_LEVEL}")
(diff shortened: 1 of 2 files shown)


vaughnbetz:
Seems reasonable to me.  @alirezazd @navidjafarof : any concerns with this change?

navidjafarof:
@vaughnbetz No concerns.

PR #2531, opened 2024-04-11
[CI][XML] Added XMLLINT to CI Dependencies

AlexandreSinger:
A warning was being generated by EZGL when pre-processing the resources for main.ui. This could not be removed since the library we use to preprocess always tries to use XMLLINT even when it is not used.

If the system does not have XMLLINT installed, it will always give a warning saying it is not installed.

As a quick fix, we can make it so the CI has it installed, since this is technically the use case we want tested.

See issue #2518


AlexandreSinger:
@vaughnbetz Please review when you have a moment.

PR #2542, opened 2024-04-20
Specify available moves explicitly

soheilshahrouz:
Currently, we pass the number of available moves to the KArmedBanditAgent's constructor to specify which moves are available. This interface assumes that move types have a specific order. This PR changes the interface so that available moves can be specified explicitly.

The main motivation for this PR is to add NoC-specific move types that are disabled for conventional architectures.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/device_grid.cpp
+++ b/libs/libarchfpga/src/device_grid.cpp
@@ -1,14 +1,16 @@
 #include "device_grid.h"
 
+#include <utility>
+
 DeviceGrid::DeviceGrid(std::string grid_name, vtr::NdMatrix<t_grid_tile, 3> grid)
-    : name_(grid_name)
-    , grid_(grid) {
+    : name_(std::move(grid_name))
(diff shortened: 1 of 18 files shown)


vaughnbetz:
@vaughnbetz  : what do you think of this?  (testing)
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/c9fcf98dd880be1f996281ae42805936b17acf4f/vpr/src/base/CheckArch.cpp#L11

vaughnbetz (commented):
One suggested clean up (use a vector of names / properties) for the move types.

soheilshahrouz:
@vaughnbetz 

I applied your comments. Here is [the link](https://docs.google.com/spreadsheets/d/1vKspFxHDAFFv-80apVzBddpAIP62kzOODyjQ20IS_8U/edit?usp=sharing) to QoR for Titan that shows the new interface for specifying availale move types does not degraged the runtime.

Issue #2548, opened 2024-05-08
VTR Quick Start Guide Inconsistencies


robluo:
While following the VTR quick start guide, I faced several issues that cannot be resolved by only reading the quick start guide's instructions. 
#### Issues encountered:

- [x] `make env` did not work. Some issue with Python libraries
Setting the Python black library to 22.8.0 fixed the issue. (requirements.txt)
See: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2527/files
Wintermute cannot support 24.3.0 (its version of Ubuntu is too old).
Would require upgrading python
We would need to figure out what version of black to set this to if we want to fix this issue.
Preferably the minimum version to get this to work. However, 24.3.0 contains security updates we may want. Perhaps we should enforce an upgrade? -Alex

- [ ] The quickstart guide uses Odin as an example; however, Parmys is now the default tool. The quickstart should be updated to use Parmys or should say how to build VTR with Odin. The current quick fix is to build Odin by explicitly setting the flag in CMake. 
- [ ] Run_vtr_flow is having some issues.
Parmys is unable to parse the architecture files…
https://docs.verilogtorouting.org/en/latest/quickstart/#automatically-running-the-vtr-flow 

Error produced after running as instructed: 
<img width="1433" alt="截屏2024-05-09 下午3 02 51" src="https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/96634753/78cdaeac-9b01-4bd8-b557-0257c0e96ed3">

@navidjafarof Could you please take a look. This errror would happen if one strickly follows the quick start guide in the above link. 

- [x] In https://docs.verilogtorouting.org/en/latest/tutorials/flow/basic_flow/, the script parse_vtr_task.py is not needed since it is already part of the run_vtr_task.py script.

vaughnbetz:
@robluo : I think this was subsequently resolved. Can it be closed?

robluo:
@vaughnbetz For the second checkbox, @nedsels also raised an issue for it in #2551. For the third checkbox, it can be resolved by using run_vtr_task. I will close the issue.

PR #2550, opened 2024-05-09
Quickstart, requirements, and flow tutorial update

nedsels:
1. Quickstart now shows how to build for use with ODIN II
2. Updated basic design flow tutorial to remove redundant instructions to run ``parse_vtr_task.py``
3. Changed version of black in ``requirements.txt`` to 21.4b0 so that ``make env`` works

#### Motivation and Context
1. The quickstart guide did not tell users how to build to that ODIN II could be used for synthesis. They would have to go to the ODIN II page and see to navigate to the ``odin_ii`` directory and ``make build``. The new command in the quickstart guide allows them to build ODIN II when they build VTR.
2. There was an instruction to run ``parse_vtr_task.py`` in the Basic Design Flow Tutorial, the reason being it created a ``parse_results.txt`` file, but this is already created by the earlier ``run_vtr_task.py``
3. ``make env`` did not work with the previous version of black, 20.8b1

3. I ran ``make env`` with the new version of black, and there are no errors.

#### Types of changes

#### Checklist:

### DIFF
--- a/doc/src/quickstart/index.rst
+++ b/doc/src/quickstart/index.rst
@@ -48,6 +48,12 @@ On most unix-like systems you can run:
 
     > make
 
+The default front-end for VTR is :ref:`Parmys<parmys>`, but you can build with ODIN II instead using the command below. This is required to run :ref:`Synthesizing with ODIN II<synthesizing_with_odin_ii>`.
+
+.. code-block:: bash
+
+    > make CMAKE_PARAMS="-DWITH_ODIN=on"
+
 from the VTR root directory (hereafter referred to as :term:`$VTR_ROOT`) to build VTR.
 
 .. note:: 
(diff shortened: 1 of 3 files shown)


AlexandreSinger:
@vaughnbetz What do you think about these changes? This resolves the black version which was causing `make env` not to work properly and resolves some inconsistencies with how run_vtr_task.py is introduced.

See issue #2548

doc/src/quickstart/index.rst
```diff
@@ -48,6 +48,12 @@ On most unix-like systems you can run:
 
     > make
 
+or, if you will be using :ref:`ODIN II for synthesis<synthesizing_with_odin_ii>`,
+
+.. code-block:: bash
+
+    > make CMAKE_PARAMS="-DWITH_ODIN=on"
```
AlexandreSinger, line 55: I think that it should be made more clear that Parmys is the default front-end for VTR, but people can build VTR with ODIN by using this build command. Maybe mention that this is required to run "Synthesizing with ODIN II".

AlexandreSinger (commented):
This is a good start. Please see my comment below.

doc/src/quickstart/index.rst
nedsels: This should be updated now

nedsels:
@vaughnbetz Please review

vaughnbetz:
Looks good and merging.
One enhancement we should do next though -- we should change to synthesizing with parmys, as it is now the default synthesis flow.  Basically replace the Odin II section with the parmys equivalent.  The Odin II part can be moved to an optional section at the end of the quick start or (even better) a quick start at the beginning of the Odin II documentation.

Issue #2449, opened 2023-11-28
Segment fault


narutozxp:
It seems that grid_layouts.size() is always larger than 1, even  if  my architecture has just one layer. Therefore, the following code will result in segment fault.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/17a816ac90a21ca197cd14b783f36236cc214f61/libs/libarchfpga/src/echo_arch.cpp#L234-L238

vaughnbetz:
Fixed in PR.

PR #2552, opened 2024-05-14
Architecture Seg Fault Issue

saaramahmoudi:
In the current codebase, when we activate --echo_file, we might get a segmentation file while writing the architecture file. The reason is that we look at the number of layouts instead of layers. The architecture might have different layouts but still not be 3D using multiple <fixed_layout> options.  

This PR addresses the following issue:
[Segmentation Fault](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2449)

#### Types of changes

@vaughnbetz

### DIFF
--- a/libs/libarchfpga/src/echo_arch.cpp
+++ b/libs/libarchfpga/src/echo_arch.cpp
@@ -13,6 +13,10 @@
 
 using vtr::t_linked_vptr;
 
+/// @brief indices to lookup IPIN connection block switch name
+constexpr int ipin_cblock_switch_index_within_die = 0;
+constexpr int ipin_cblock_switch_index_between_dice = 1;
+
 void PrintArchInfo(FILE* Echo, const t_arch* arch);
 static void PrintPb_types_rec(FILE* Echo, const t_pb_type* pb_type, int level);
 static void PrintPb_types_recPower(FILE* Echo,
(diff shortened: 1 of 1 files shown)


libs/libarchfpga/src/echo_arch.cpp
```diff
@@ -232,9 +232,13 @@ void PrintArchInfo(FILE* Echo, const t_arch* arch) {
     }
 
     fprintf(Echo, "\tInput Connect Block Switch Name Within a Same Die: %s\n", arch->ipin_cblock_switch_name[0].c_str());
```
vaughnbetz, line 234: I think it's worth commenting the [0] and [1] indices: [0] is the ipin_cblock_switch_name for within die and [1] is between dice?  This might be more clear with constexpr that gives names to 0 and 1 for the lookups.

saaramahmoudi: @vaughnbetz fixed it

vaughnbetz:
Changes look good but CI failed hard….VaughnOn May 14, 2024, at 5:12 PM, sara_mahmoudi ***@***.***> wrote:﻿
@saaramahmoudi commented on this pull request.



In libs/libarchfpga/src/echo_arch.cpp:
> @@ -232,9 +232,13 @@ void PrintArchInfo(FILE* Echo, const t_arch* arch) {
     }
 
     fprintf(Echo, "\tInput Connect Block Switch Name Within a Same Die: %s\n", arch->ipin_cblock_switch_name[0].c_str());

@vaughnbetz fixed it

saaramahmoudi:
@vaughnbetz Sorry about that, I accidentally had an extra [, which causes compile issue. Fixed now!

PR #2553, opened 2024-05-16
[Warnings] Suppressed Bison Deprecation Warnings

AlexandreSinger:
Suppressed the Bison deprecation warnings for SDCParse and BlifParse since the fix for the deprecation requires Bison 3.3, but the current minimum version of Bison is 3.0

Some development machines using VPR cannot be upgraded to Bison 3.3 easily, so for now the deprecation warnings are just being suppressed until all machines are on Bison 3.3

PR #2529 tracks this issue and contains the code to fix the deprecations.

This represents the last warnings in the CI detailed in issue #2518

### DIFF
--- a/libs/EXTERNAL/libblifparse/CMakeLists.txt
+++ b/libs/EXTERNAL/libblifparse/CMakeLists.txt
@@ -27,7 +27,9 @@ file(GLOB_RECURSE PARSER_SOURCES src/blif*.y)
 #Make the flex and bison targets
 flex_target(BlifLexer ${LEXER_SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/blif_lexer.gen.cpp
         COMPILE_FLAGS --header-file=${CMAKE_CURRENT_BINARY_DIR}/blif_lexer.gen.hpp)
-bison_target(BlifParser ${PARSER_SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/blif_parser.gen.cpp)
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
@vaughnbetz With these suppressions the CI should now be warning free for all builds with the current version of GCC (GCC9). I will leave the PR raising the minimum Bison version alive so that we can easily upgrade in the future when needed. I add the removal of these suppressions to that PR.

vaughnbetz:
@AlexandreSinger : should we suppress these, or given the discussion today, do you think we can leave them and still have warnings as errors in CI?  I'm OK either way ... let's talk tomorrow about whether this should be merged.

AlexandreSinger:
> @AlexandreSinger : should we suppress these, or given the discussion today, do you think we can leave them and still have warnings as errors in CI? I'm OK either way ... let's talk tomorrow about whether this should be merged.

@vaughnbetz Agreed. It does depend on how we implement the fail on warning. I'll show you the CI changes tomorrow.

vaughnbetz:
@AlexandreSinger : should we just close without merging?

AlexandreSinger:
@vaughnbetz Yes. Since the fail on warnings do not fail for these deprecation warnings, we luckily do not have to suppress them. This is good since suppressing these warnings could cause issues in the future.

I will keep PR #2529 open until we raise our dev machines to a more recent version of Ubuntu which will fully resolve the issue.

PR #2555, opened 2024-05-16
[CI] Made the Regression Tests Error on Warning

AlexandreSinger:
To prevent warnings from showing up in the build in the future, made the
CI error on any warnings during the regression tests.
    
It looks like VPR originally had the ability to error on warning;
however, it was specific to only VPR. This CMAKE method will now work
for all executables compiled during the tests.

With this merged, this should completely resolve issue #2518

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -184,17 +184,17 @@ jobs:
         include: [
           {
             name: 'Basic',
-            params: '-DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on',
+            params: '-DCMAKE_COMPILE_WARNING_AS_ERROR=on -DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on',
             suite: 'vtr_reg_basic'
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
To demonstrate the importance of these changes, a commit was merged yesterday which introduced more warnings:
- https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/b207bb9011fee4ed55c682c16f935c0106c11408

The commit removed the use of a variable when debug logging is enabled,.

Hence why two of the tests failed. These warnings build up fast I guess.

vaughnbetz:
Ah, a warning snuck back in already!
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/9217393097/job/25359296613?pr=2555

AlexandreSinger:
@vaughnbetz It looks like we were correct. Even though the Bison deprecation warnings are still there, this method of failing on warning does not fail the CI in this case. This is actually a good thing. As we discussed this may mean that we can keep the deprecations without suppressing them.

Issue #2557, opened 2024-05-24
Remove Uses of std::iterator from VTR Libraries


AlexandreSinger:
The std::iterator class will be deprecated in C++17:
- https://stackoverflow.com/questions/43268146/why-is-stditerator-deprecated

In CI, the GCC 12 and CLANG 14 build show the deprecation warnings. There are other warnings in these builds, but the std::iterator warnings are blocking most of them.

This is fine right now since the CI is currently on Ubuntu 22.04, which comes with GCC 11.4 built-in; however, in the near future we should upgrade the CI to Ubuntu 24.04, which comes with GCC 13.3 built-in. GCC 13 is not even being tested by CI yet, but it will have the same deprecation warnings (if not these features will actually be deprecated).

We should begin to transition the use of std::iterator into their own classes. The process seems to be pretty standard, it would just require some work since some of the uses of std::iterator are in external libraries.

The current users of std::iterator in VTR are:
- [x] librrgraph/src/base/rr_node_impl.h <img width="704" alt="image" src="https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/39ab2764-9c9a-4fdf-a2f5-6e18470f900a">
- [x] librrgraph/src/base/rr_node_types.h <img width="677" alt="image" src="https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/fbcc8249-002b-45b9-ba29-03b6f17bf54c">
- [x] libvtrutil/src/vtr_array_view.h <img width="615" alt="image" src="https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/2baa6516-123b-4c04-8267-9197263598b9">
- [x] libvtrutil/src/vtr_ragged_matrix.h <img width="640" alt="image" src="https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/3ef19e87-c0d9-4c98-a6f2-ad718ad45239">
- [x] libvtrutil/src/vtr_vector.h <img width="612" alt="image" src="https://github.com/verilog-to-routing/vtr-verilog-to-routing/assets/49374526/85f37757-1333-4a41-9201-1de7e605bf38">
[...]

PR #2558, opened 2024-05-25
Removed Simple Uses of STD Iterator

AlexandreSinger:
See issue #2557 for more context.

Removed the uses of std::iterator that were very simple to remove. These just required declaring some types that the users of the iterators expect and removing the inheritance on the std::iterator.

The other cases in VTR which use std::iterator are a bit more complicated and may require some work to remove correctly.


AlexandreSinger:
@vaughnbetz This has passed CI, and the warnings pertaining to these specific uses of std::iterator are gone from the GCC12 build. What do you think about these changes?

vaughnbetz:
Thanks!

PR #2561, opened 2024-05-27
Removed Less-Simple Uses of STD Iterator

AlexandreSinger:
See issue #2557 for context.

As with PR #2558 , removed slightly more complicated uses of std::iterator after discussing this more.

This PR modifies Tatum, I will upstream these changes back to the Tatum library after the PR is merged.

After this PR, the only use of std::iterator would be in the Capnproto library, which may be quite a bit more complicated to resolve.


AlexandreSinger:
@vaughnbetz This has passed CI, I have verified that the warnings for these uses of std::iterator are now gone. These are all of the uses of std::iterator that VTR is responsible for. The rest of the uses can be resolved by upgrading capnproto and Yosys.

vaughnbetz:
Thanks!

PR #2562, opened 2024-05-27
Support for link/router specific bandwidth/latency in architecture file

soheilshahrouz:
This PR adds support for heterogeneous NoCs description in the architecture file. It is now possible to specify the latency of each NoC link and NoC router seperately. The user can also specify the bandwidth of each individual NoC link.

The architecture file reference needs to be updated to explaing how <overrides> tag can be used. I think the code is done and the review process can begin. In meantime, I update the architecture file reference.

#### Description
The default latency and bandwidth for NoC links and routers are specified under <noc> tag with link_bandwidth, link_latency, and router_latency attributes. The user can override these default values for certain routers and links by adding an optional <overrides> tag under <noc> tag.

Before this PR, some NoC-related cost calcualtion functions assumed the latency and bandwidth is the same for all NoC components. Now, these functions check if the latnecy or bandwidth is the same for all NoC components. When this not the case, thse functions query the bandwidth or the latency of NoC components separately.

#### Motivation and Context
Make the architecture description file syntax support heterogeneous NoC topologies.

Has been tested with a toy architecture file.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/place/noc_place_utils.cpp
+++ b/vpr/src/place/noc_place_utils.cpp
@@ -146,10 +146,7 @@ void find_affected_noc_routers_and_update_noc_costs(const t_pl_blocks_to_be_move
     }
 
     // Iterate over all affected links and calculate their new congestion cost and store it
-    for (const auto& link_id : affected_noc_links) {
-        // get the affected link
-        const auto& link = noc_ctx.noc_model.get_single_noc_link(link_id);
-
+    for (const NocLink& link : noc_ctx.noc_model.get_noc_links(affected_noc_links)) {
         // calculate the new congestion cost for the link and store it
         proposed_link_congestion_costs[link] = calculate_link_congestion_cost(link);
 
@@ -174,10 +171,7 @@ void commit_noc_costs() {
     }
 
     // Iterate over all the NoC links whose bandwidth utilization was affected by the proposed move
-    for(auto link_id : affected_noc_links) {
-        // get the affected link
-        const auto& link = noc_ctx.noc_model.get_single_noc_link(link_id);
-
+    for (const NocLink& link : noc_ctx.noc_model.get_noc_links(affected_noc_links)) {
         // commit the new link congestion cost
         link_congestion_costs[link] = proposed_link_congestion_costs[link];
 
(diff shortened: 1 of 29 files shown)


vpr/src/place/noc_place_utils.cpp
```diff
@@ -552,19 +546,47 @@ double calculate_traffic_flow_aggregate_bandwidth_cost(const std::vector<NocLink
...
+    double noc_link_latency_component = 0.0;
+    if (noc_model.get_detailed_link_latency()) {
+        for (const NocLink& link : noc_model.get_noc_links(traffic_flow_route)) {
+            double link_latency = link.get_latency();
+            noc_link_latency_component += link_latency;
+        }
+    } else {
+        auto num_of_links_in_traffic_flow = (double)traffic_flow_route.size();
+        double noc_link_latency = noc_model.get_noc_link_latency();
+        noc_link_latency_component = noc_link_latency * num_of_links_in_traffic_flow;
```
vaughnbetz, line 559: If you stored the latency as part of the routing computation, then you could avoid this if statement / two methods. Consider if that is easier to understand (gets rid of a state control variable).

Also can get rid of the detailed bandwidth variable. (if it controls nothing, I'd get rid of it).

vaughnbetz (commented):
Looks good. Only substantial thing to consider is if you can unify the code paths for the "all same latency" and detailed latency cases.

vaughnbetz:
Should also do a QoR test on the router suite just to be sure it is OK.

vpr/src/place/noc_place_utils.cpp
soheilshahrouz: I think we should keep all NoC-related cost compuation code in a single file. We can't move congestion computation into the routing algorithm. So, moving the latency and aggregate bandwidth computation to the routing algorithm splits the NoC-realted cost calcuation code across files, which is not ideal.

I also think that if we move aggregate bandwidth and latency calculation from here to routing algrotihms, we need to implement the calculation process multiple times as NoCRouting::route_flow(...) method is pure virtual.

vaughnbetz: Valid points.  OK, let's leave as is.

PR #2565, opened 2024-05-28
[CI] Added GCC-13 as a Compatability Test

AlexandreSinger:
Ubuntu 24.04 uses GCC-13 as its default GCC version. The CI should ensure that VTR builds for this version of GCC.


AlexandreSinger:
Found that the Ubuntu 22.04 image does not have GCC-13 since the Ubuntu 24.04 image has it. Closing the PR.

PR #2566, opened 2024-05-28
Sync Tatum with VTR

AlexandreSinger:
libtatum was behind upstream Tatum by a few commits. These commits had no affect on the code used by VTR, so that is likely the reason they were not merged in.

I recently raised a PR on Tatum bringing VTR's changes back to Tatum: https://github.com/verilog-to-routing/tatum/pull/9 

This is the reverse, bringing Tatum's changes in.

When both of these PRs are merged, both libtatum in VTR and Tatum should be synced.


AlexandreSinger:
@vaughnbetz These are just some of the changes in Tatum that were not in VTR. None of them have a direct affect on the code, but its nice to keep the two repos in sync.

PR #2568, opened 2024-05-29
[CI] Upgraded Actions

AlexandreSinger:
The Annotations section of the CI was full of deprecation warnings:
image

Node.js 16 actions were deprecated. Upgraded setup-python, checkout, and upload-artifact to their most recent version. See: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

v3 of upload-artifact was also deprecated. v4 claims to be 98% faster, however, may come with some breaking changes. Pertaining to us, artifacts cannot have the same name now and there is a limit of 500 artifacts. See:
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

Upgrade summary:
- actions/setup-python@v4 -> actions/setup-python@v5
- actions/checkout@v3 -> actions/checkout@v4
- actions/upload-artifact@v3 -> actions/upload-artifact@v4

NOTE: Had to leave the nightly tests behind since the self-hosted
machine does not support node20. Need to upgrade the machine in order to
upgrade these actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,8 @@ jobs:
 
     steps:
 
+      # TODO: This runnner is running on a self-hosted CPU. In order to upgrade
+      #       to v4, need to upgrade the machine to support node20.
     - uses: actions/checkout@v3
       with:
         submodules: 'true'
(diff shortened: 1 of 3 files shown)


AlexandreSinger:
@vaughnbetz What do you think about these changes? This has removed most of the warnings except for 1, which I put a comment in the CI code with a TODO regarding upgrading the self-hosted runners:
image

This also made it easier to download artifacts for the nightly tests since it now separates them out into their own artifacts:
image

.github/workflows/test.yml
```diff
@@ -71,9 +73,12 @@ jobs:
         VTR_CMAKE_PARAMS: ${{ matrix.cmake }}
         NUM_CORES: ${{ matrix.cores }}
 
-    - uses: actions/upload-artifact@v3
-      if: ${{ always() }}
```
soheilshahrouz, line 75: I think this line should not have been removed.

vaughnbetz: @AlexandreSinger : what do you think? Should this go back in?

AlexandreSinger: @soheilshahrouz  @vaughnbetz My bad... I misunderstood what this was trying to do. Perhaps in the future we should consider a better keyword to use:
https://docs.github.com/en/actions/learn-github-actions/expressions#always

In the meantime I have raised a PR reverting this removal: PR #2583

PR #2572, opened 2024-05-30
Merge master to openfpga feature branch

Tulong4Dev:
Merge master to openfpga feature branch.
There are test failures which look like to be caused by RSA settings.


AlexandreSinger:
@Tulong4Dev The test failures you are seeing are likely caused by a recent change in VTR master which made the CI fail on warning for those tests. If your branch has warnings, it will fail those tests. See PR #2555 which introduced this change.

Since you are not merging into master, and you would prefer to keep the warnings, you can opt to remove `-DCMAKE_COMPILE_WARNING_AS_ERROR=on` from these tests in `.github/workflows/test.yml`. It will run the CI as normal, but will allow warnings to pass through.

On master we opted to fail on warning to prevent mistakes and deprecations in the master branch of VTR.

PR #2577, opened 2024-06-04
[CI] Added GCC-13 as a Compatability Test

AlexandreSinger:
Ubuntu 24.04 uses GCC-13 as its default GCC version. The CI should ensure that VTR builds for this version of GCC.

Also added the version of Clang that would come with Ubuntu 24.04

Versions of tools were found from here:
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

### DIFF
--- a/.github/scripts/install_noble_dependencies.sh
+++ b/.github/scripts/install_noble_dependencies.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+
+sudo apt update
+
+sudo apt install -y \
+  autoconf \
+  automake \
+  bash \
+  bison \
+  binutils \
+  binutils-gold \
+  build-essential \
+  capnproto \
+  exuberant-ctags \
+  curl \
+  doxygen \
+  flex \
+  fontconfig \
+  gdb \
+  git \
+  gperf \
+  libcairo2-dev \
+  libcapnp-dev \
+  libgtk-3-dev \
+  libevent-dev \
+  libfontconfig1-dev \
+  liblist-moreutils-perl \
+  libncurses5-dev \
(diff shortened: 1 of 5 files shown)


vaughnbetz:
Thanks @AlexandreSinger. Seems tcl.h included by yosys is not found with Gcc 13:

[ 66%] Generating yosys-bin
In file included from kernel/driver.cc:20:
./kernel/yosys.h:81:12: fatal error: 'tcl.h' file not found
   81 | #  include <tcl.h>
      |            ^~~~~~~
1 error generated.

AlexandreSinger:
@vaughnbetz I believe tcl is included as an apt package in install_dependencies.sh; reading through the CI logs, I think the apt packages are failing to load. I think certain versions of packages do not exist for Ubuntu 24.04 which is killing the load. I am suprised that the CI continued to run after this happened. I will investigate. I am also surprised that Yosys does not detect that the tcl package is missing; so perhaps there is a deeper problem.

AlexandreSinger:
@vaughnbetz This PR adds a new workflow which is not part of the other CI tests and can only be dispatched (or run automatically daily), which will run the Ubuntu 24.04 build with GCC-13. Along with a small patch to fix a small include issue and a new script to install the necessary apt packages, GCC-13 and Clang-18 should build VTR.

There are many warnings in the build, which may need more work to resolve. As discussed during the VTR industry sync, I do plan on combining the `install_dependencies.sh` script with the new noble one. But I want to make sure the build is all correct before doing so. This extra workflow will make it easier to upgrade to Ubuntu 24.04, since it will make testing it easier to debug.

PR #2579, opened 2024-06-04
Use top level pb name as cluster name when printing a .place file.

KA7E:
Use top level pb name instead of clb name when printing a .place file..

Changed one line in print_place to print clb_nlist.block_pb[iblk]->name, instead of clb_nlist.block_name[iblk].

#### Motivation and Context
 When VPR prints a .place file it uses cluster names from clb_nlist.block_name(); when it prints a  a .net file, it names cluster-level blocks using pb names from clb_nlist.block_pb(). While these names both begin as the name of the cluster seed, pb names are modified during clustering (e.g. when adding a carry chain) and by the re-clustering API. By the time .net and .place files are printed, the cluster names may not match. When reading in a .net file, VPR populates clb_nlist.block_name() from the .net file (i.e. the top level pb name), and so when subsequently reading in a .place file the names may not match.  VPR get around this by matching .place file names against all atom names in the cluster.  However, the re-clustering API sometimes adds a suffix to a new cluster name (for uniqueness); in this case the name in the .place file will not match any atom name, and so if it does not match the top level pb name, the .place file entry is skipped. 

#### Types of changes

#### Checklist:


rs-dhow:
Question: from the .place file, we build a correspondence block name <=> x,y,z coordinate.
The block name always appears in the .net file (packer output) on a top-level block (immediately under the block for the entire array). This relationship is unchanged? Block names at all levels (except full array) come from the first output of the contained object(s). Thanks.

KA7E:
This change will ensure that the name in the .place file will match the block name in the .net file.  Currently it is possible to end up with a mismatch.

> Question: from the .place file, we build a correspondence block name <=> x,y,z coordinate. The block name always appears in the .net file (packer output) on a top-level block (immediately under the block for the entire array). This relationship is unchanged? Block names at all levels (except full array) come from the first output of the contained object(s). Thanks.

PR #2584, opened 2024-06-05
add detailed routing stage parameter to reclustering utils

KA7E:
Added detailed routing stage parameter to pack_mol_in_existing_cluster and start_new_cluster_for_mol; both functions pass this parameter to try_pack_molecule.

#### Motivation and Context
Without this option, the legalizer runtime is untenable.

#### Types of changes

#### Checklist:


vpr/src/pack/re_cluster_util.h
```diff
@@ -107,7 +108,8 @@ bool pack_mol_in_existing_cluster(t_pack_molecule* molecule,
                                   std::unordered_set<AtomBlockId>* new_clb_atoms,
                                   bool during_packing,
                                   t_clustering_data& clustering_data,
-                                  t_lb_router_data*& router_data);
+                                  t_lb_router_data*& router_data,
+                                  const int detailed_routing_stage = 1);
```
AlexandreSinger, line 112: This argument of the function does not match the declaration. The implementation passes the int by reference and this is saying to pass by value. Line 91 above also has this issue. This is causing warnings in the CI and we fail on warning now:
image

This *should* actually crash; however, since there is no code currently in VTR using this method, it does not cause an issue. But the warning is saving us here. This should be fixed.

KA7E: Fixed.  Thanks for finding this.

KA7E:
@vaughnbetz could you please review?  Subsequent legalizer PRs depend on this.

KA7E:
Re: QoR.  This option is key to speeding up the legalizer, since it allows it to use pin counting legality checking instead of runtime intensive intra-cluster routing.  I do not think it has any impact on the rest of VPR, since it only impacts re-clustering API functions (which are not currently called anywhere else); for those functions, it only adds an optional const int& parameter.

vpr/src/pack/re_cluster_util.h
```diff
@@ -88,7 +88,8 @@ bool start_new_cluster_for_mol(t_pack_molecule* molecule,
                                t_clustering_data& clustering_data,
                                t_lb_router_data** router_data,
                                PartitionRegion& temp_cluster_pr,
-                               NocGroupId& temp_cluster_noc_grp_id);
+                               NocGroupId& temp_cluster_noc_grp_id,
+                               const int& detailed_routing_stage = 1);
```
vaughnbetz, line 92: This new parameter needs to be commented here (add to the doxygen comment above).  Also comment why it defaults to 1.
Why is it being passed as a const int&?  An int is smaller than a reference, so it is simpler and slightly faster to just pass the int rather than a const int&.
@soheilshahrouz : I can also see that the temp_cluster_noc_grp_id is not commented in the function doxygen -- please add a comment for that in another PR.

vpr/src/pack/re_cluster_util.h
```diff
@@ -107,7 +108,8 @@ bool pack_mol_in_existing_cluster(t_pack_molecule* molecule,
                                   std::unordered_set<AtomBlockId>* new_clb_atoms,
                                   bool during_packing,
                                   t_clustering_data& clustering_data,
-                                  t_lb_router_data*& router_data);
+                                  t_lb_router_data*& router_data,
+                                  const int& detailed_routing_stage = 1);
```
vaughnbetz, line 112: Comment detailed_routing_stage parameter in the Doxygen block.
I think this would be better implemented as an int, not a const int &.

vaughnbetz (changes requested):

vpr/src/pack/re_cluster_util.h
vaughnbetz: Is detailed_routing_stage on or off?  If so, a bool would be better.  If it can take a limited set of enumerated values, declaring it of the appropriate enumerated type is better. And if it can be arbitrary integers it needs to be carefully commented what that means.

vpr/src/pack/re_cluster_util.h
KA7E: I did it that way because I saw mode being passed as a const int& in one of the function.  Happy to change it.
detailed_routing_stage is an enum, not a bool; it might be more clear to #include the header that defines the enum so that the default value can be E_DETAILED_ROUTE_FOR_EACH_ATOM; I made this the default so that all the function calls in re_cluster.cpp don't need to be changed.

vaughnbetz: We should make the variable type the enum type then, for clarity and to avoid potential (and subtle) possible later bugs if someone adds another enum value and the integer 1 is no longer the mode you want.  Including the header you need is fine.

Better to pass the enum by value than const ref, so that should be changed too.

vpr/src/pack/re_cluster_util.h
```diff
@@ -108,7 +113,9 @@ bool pack_mol_in_existing_cluster(t_pack_molecule* molecule,
                                   std::unordered_set<AtomBlockId>* new_clb_atoms,
                                   bool during_packing,
                                   t_clustering_data& clustering_data,
-                                  t_lb_router_data*& router_data);
+                                  t_lb_router_data*& router_data,
+                                  NocGroupId& temp_cluster_noc_grp_id,
+                                 enum e_detailed_routing_stages detailed_routing_stage = E_DETAILED_ROUTE_FOR_EACH_ATOM);
```
vaughnbetz, line 118: Comment the last two arguments (same as in the other comment).

vpr/src/pack/re_cluster_util.h
```diff
@@ -72,11 +72,15 @@ void remove_mol_from_cluster(const t_pack_molecule* molecule,
...
- * @param clustering_data: A data structure containing helper data for the clustering process 
+ * @param clustering_data: A data structure containing helper data for the clustering process
  *                          (is updated if this function is called during packing, especially intra_lb_routing data member).
  * @param router_data: returns the intra logic block router data.
  * @param temp_cluster_pr: returns the partition region of the new cluster.
  * @param temp_cluster_noc_grp_id returns the NoC group ID of the new cluster
+ * @param detailed_routing_stage: options are E_DETAILED_ROUTE_FOR_EACH_ATOM (default) and E_DETAILED_ROUTE_AT_END_ONLY
+ *                                 specifies whether or not to run intra-cluster routing-based legality checking
+ *                                 after adding the molecule to the cluster; default is the more conservative option,
+ *                                 which is used in the top level re-clustering API functions
```
vaughnbetz, line 83: Suggest changing "which is used in the top level re-clustering API functions" to something like:
"This option is passed down to lower-level routines; if it is E_DETAILED_ROUTE_AT_END_ONLY then detailed routing checks will not be done (this is useful if one detailed check can later be done on the completed cluster)."

vaughnbetz:
It looks like the prior PR I merged on (another one of your PRs) resulted in a conflict with this one.  Some line of code got changed twice so git can't figure out which one to use; they are likely the same though as both PRs are closely related.  @KA7E : you'll have to pull and resolve the conflict (it should be an easy one) before this can be merged.  

It's probably best to have one PR for a function (or a full file) that you're changing at a time so you don't get self conflicts like this. Smaller PRs are easier to merge, but not if they overlap on the affected code, and PRs with up to a couple of hundred lines of code would still be considered pretty small and aren't very hard to review. Some PRs will also wind up being bigger than that since they're larger changes that are coupled together.

Issue #2586, opened 2024-06-05
Issue with git submodule update --init Command


amin1377:
#### Expected Behavior
When I run `git submodule update --init`, I expect to initialize and update all of the submodules without any errors.

#### Current Behavior
When I run the mentioned command, I encounter the following error with the `sockpp` submodule:
```
Cloning into 'libs/EXTERNAL/sockpp'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:w0lek/sockpp.git' into submodule path 'libs/EXTERNAL/sockpp' failed
```
I also tried a clean clone (removing the VTR repo and cloning it again), and I faced the same issue. To work around this, when I want to build, I set `VPR_USE_SERVER` to `OFF`.

soheilshahrouz:
Closing as the PR was merged.

PR #2588, opened 2024-06-05
Fix submodule init issue

soheilshahrouz:
Solves [this issue](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2586)

Since sockpp uses ssh to clone the repo, ssh keys need to be set up. Alternatively, we can use https for cloning the repo.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,5 +3,4 @@
 	url = https://github.com/catchorg/Catch2.git
 [submodule "libs/EXTERNAL/sockpp"]
 	path = libs/EXTERNAL/sockpp
-	#url = git@github.com:fpagliughi/sockpp.git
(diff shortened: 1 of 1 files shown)


soheilshahrouz:
Related to this PR is [this stackoverflow answer](https://stackoverflow.com/a/34081606)

AlexandreSinger:
Ah I see what happened here. When you use a git@ url for the repo, it requires that you have a git shh key setup (which is a good idea btw). If you use the https:// version you do not; however if you wish to push to the repo you will not be able to login. @soheilshahrouz I think you have a good point turning this into a https:// since the general user will just be pulling (not pushing anything); and so we would not require people to have an ssh key if they want to build VTR.

I am excited to see what gnarly things this will do once people rebase this change. But it must be done.

vaughnbetz:
Seems logical since we do this for other submodules.  @w0lek let us know if you have any concerns.

PR #2589, opened 2024-06-05
[CI] Changed the Workflow Triggers

AlexandreSinger:
Changed the workflow triggers for the test and container workflows such that they only run when there is a push to master.

Also added concurrency such that if a CI was called on a particular PR or branch that is already running the CI, it will cancel the old run. This will prevent useless runs of the CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,11 +1,23 @@
 name: Test
 
 on:
+  # We want to run the CI when anything is pushed to master.
+  # Since master is a protected branch this only happens when a PR is merged.
+  # This is a double check in case the PR was stale and had some issues.
   push:
+    branches:
+      - master
   pull_request:
   workflow_dispatch:
   schedule:
   - cron: '0 0 * * *' # daily
+
+# We want to cancel previous runs for a given PR or branch / ref if another CI
+# run is requested.
+# See: https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 env:
   # default compiler for all non-compatibility tests
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1,11 +1,23 @@
 name: Containers
 
 on:
+  # We want to run the CI when anything is pushed to master.
+  # Since master is a protected branch this only happens when a PR is merged.
+  # This is a double check in case the PR was stale and had some issues.
   push:
+    branches:
+      - master
   pull_request:
   workflow_dispatch:
   schedule:
   - cron: '0 0 * * 0' # weekly
+
+# We want to cancel previous runs for a given PR or branch / ref if another CI
+# run is requested.
+# See: https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
 


.github/workflows/test.yml
```diff
@@ -1,12 +1,24 @@
...
   workflow_dispatch:
   schedule:
   - cron: '0 0 * * *' # daily
 
+# We want to cancel previous runs for a given PR or branch / ref if another CI
+# run is requested.
+# See: https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
soheilshahrouz, line 20: I don't think canceling previous CI runs is a good idea. When the CI runs with every push, it's easier to find the commit where the first CI failure occurred.

To reduce the frequency of triggering CI, we can configure it to run only when the commit message includes a specific substring, like [run-ci]. This way, the user has more control over when CI is run, and previous CI runs are not cancelled if the user has requested them.

AlexandreSinger: @soheilshahrouz This will not cancel previously completed runs (that finished to completion), this will only cancel runs that are in progress. It is designed to catch the following situation:
1. A person pushes a change to a PR
2. The person 10 minutes later realizes that there is a mistake (maybe someone commented on the PR)
3. They push a new change
Since 1 and 3 are both changes to a PR they will both run the CI (which will run at the exact same time). But we do not care if 1 passes CI, only that 3 passes CI.

In the case you are proposing, I assume the CI will run to completion. So we assume that the time between pushes is long enough such that the entire CI completes. 

The issue is that the CI takes several hours to complete, so if a person were to want every single one of their commits to be labelled with "passes CI" or not, it woud take tons of resources for a person who pushes many commits.

AlexandreSinger: @soheilshahrouz This change was mainly used for the tests running on github's provided runners (as you can tell by this PR, it is currently not running any of those tests since the CI is SO backlogged). The self-hosted runners (which run the vtr_reg_nightly tests) seem to be keeping up better (however, they are also going over quota). We could separate the tests into their own workflows and only apply the concurrency to the github provided runners. But this change is due to the fact that the CI currently takes an incredibly long amount of time to complete and we are starting to get errors due to quota issues.

soheilshahrouz: I understand that it won't canel completed runs (I didn't know you can cancel a compted run. What happens when you cancel an already finished CI run?), but I still think that running CI more frequently helps with debugging cases where a developer accidentally changes the code in a way that causes a CI failure.

> In the case you are proposing, I assume the CI will run to completion. So we assume that the time between pushes is long enough such that the entire CI completes.

Not necessarily. We can allow the user to cancel previous CI runs by adding a specific string like [canel-prev-ci] to the commit message.


> The issue is that the CI takes several hours to complete, so if a person were to want every single one of their commits to be labelled with "passes CI" or not, it woud take tons of resources for a person who pushes many commits.

By allowing the user to decide which commits need a CI run, and after which commits previous CI runs should be cancelled, we can avoid the resource overutilization that we have now.

AlexandreSinger: I am a little confused since the CI should be used to prevent bad code from entering Master, its supposed to be used for PRs to ensure that the code is working and will not cause any issues. I do not believe that the CI should be used for debugging. My goal with this change is to make it more challenging for people to call the CI in cases when it isn't supposed to be (when the code is not *just* about to be merged into master).

Even with this change, it is still possible to dispatch the CI runs manually (and also cancel them manually as well). This change just makes it more difficult to accidentally push 15 commits without thinking and bringing down the CI for people who actually need it.

My issue with what you propose is that, while its great for the person who is using it for debugging, it takes CI resources away from people who are trying to merge code into master (which is its intended use-case).

In my opinion this change only prevents people from running multiple CI runs on the same branch at the same time. They can still run multiple CI runs on the same branch, but they would need to manually stagger when the call CI if they are using it for the case you are talking about.

AlexandreSinger:
@vaughnbetz I think we should merge this in. Currently the CI is full again, this time its someone's PR:
image

This is exactly what the concurrency argument is trying to resolve. I will kill these runs since I don't think this person intends on calling the CI 10 times for the same PR...

vaughnbetz:
Thanks Alex!

.github/workflows/test.yml
soheilshahrouz, line 20: @AlexandreSinger This cancels previous CI runs for the master branch. I think we should avoid canceling these runs on master because we don't push to master as often, and multiple pushes are usually from multiple PRs merged in a short time. If something goes wrong, having CI results for each push to the master branch helps us pinpoint issues more easily. What do you think about this?

AlexandreSinger: I actually did this on purpose since it is a common scenario that we would merge 3-4 PRs at one time. This causes a storm of traffic on the CI when this occurs which can cause problems. Above this section, I specifically allowed the main branch to trigger the CI on push since, I agree, that the main branch should test on merge. I personally just find running the CI on every PR merge a bit much.

If we decide to allow the CI to trigger on every push to main (and run concurrently), this resource discusses how to do that: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-on-specific-branches

Its very simple to do; but I am just trying to prevent too much traffic from hitting the CI when things get merged in. We can discuss in the group meeting.

PR #2590, opened 2024-06-06
add flat idx to pb_graph_node , populate in alloc_and_load_pb_graph

KA7E:
Added a flat_site_idx to pb_graph_node struct, along with code to populate it in alloc_and_load_pb_graph.

#### Description

As alloc_and_load_pb_graph builds the graph from the top down, each parent computes its child's flat index and passes it to them as an argument to alloc_and_load_ph_graph.  
The child's flat site is computed by the parent as:
(parent's flat site)*(total children of this type) + child's sibling index.

#### Motivation and Context
This is used to help the legalizer place a molecule on a user specified site. It is better to store this value in the node than to recompute it over and over.

Compared this value to locally computed value in the legalizer, which is then printed to a flat placement file and compared via a script.  I tested on both StratixIV and ISPD Ultrascale architectures.  In all cases, computed sites and sites stored in the pb_graph_node match.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1225,7 +1225,12 @@ struct t_pin_to_pin_annotation {
  *      parent_pb_graph_node  : parent pb graph node
  *      total_primitive_count : Total number of this primitive type in the cluster. If there are 10 ALMs per cluster
  *                              and 2 FFs per ALM (given the mode of the parent of this primitive) then the total is 20.
- *      flat_site_index       : index of this primitive within its primitive type in this cluster; in [0,...,total_primitive_count]
+ *                              This member is only used by nodes corresponding to primitive sites.
+ *      flat_site_index       : Index of this primitive site within its primitive type within this cluster type.
+ *                              Values are in [0...total_primitive_count-1], e.g. if there are 10 ALMs per cluster, 2 FFS
+ *                              and 2 LUTs per ALM, then flat site indices for FFs would run from 0 to 19, and flat site
+                                indices for LUTs would run from 0 to 19. This member is only used by nodes corresponding
+                                to primitive sites. It is used when reconstructing clusters from a flat placement file.
  *      illegal_modes         : vector containing illegal modes that result in conflicts during routing
  */
 class t_pb_graph_node {
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1225,6 +1225,7 @@ struct t_pin_to_pin_annotation {
  *      parent_pb_graph_node  : parent pb graph node
  *      total_primitive_count : Total number of this primitive type in the cluster. If there are 10 ALMs per cluster
  *                              and 2 FFs per ALM (given the mode of the parent of this primitive) then the total is 20.
+ *      flat_site_index       : index of this primitive within its primitive type in this cluster; in [0,...,total_primitive_count]
  *      illegal_modes         : vector containing illegal modes that result in conflicts during routing
  */
 class t_pb_graph_node {
@@ -1281,6 +1282,8 @@ class t_pb_graph_node {
     int num_output_pin_class;   /* number of output pin classes that this pb_graph_node has */
 
     int total_primitive_count; /* total number of this primitive type in the cluster */
+     int flat_site_index;       /* index of this primitive within sites of its type in this cluster  */
+
 
     /* Interconnect instances for this pb
      * Only used for power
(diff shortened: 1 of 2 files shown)


KA7E:
@vaughnbetz could you please review?  Subsequent legalizer PRs depend on this.

libs/libarchfpga/src/physical_types.h
```diff
@@ -1225,6 +1225,7 @@ struct t_pin_to_pin_annotation {
  *      parent_pb_graph_node  : parent pb graph node
  *      total_primitive_count : Total number of this primitive type in the cluster. If there are 10 ALMs per cluster
  *                              and 2 FFs per ALM (given the mode of the parent of this primitive) then the total is 20.
+ *      flat_site_index       : index of this primitive within its primitive type in this cluster; in [0,...,total_primitive_count]
```
vaughnbetz, line 1228: An example in the comment would be good that makes clear how this is used and how it differs from placement_index. (e.g. if you have 10 LUTs and 10 FFs in a hierarchy level would flat_site_index goes from 0 to 19?).  Explain what this is used for -- I believe it is only for reading in external flat placments so explain that and what it is useful for in that case.

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -349,11 +353,14 @@ static void alloc_and_load_pb_graph(t_pb_graph_node* pb_graph_node,
                                                                                 sizeof(t_pb_graph_node*));
         for (j = 0; j < pb_type->modes[i].num_pb_type_children; j++) {
             pb_graph_node->child_pb_graph_nodes[i][j] = (t_pb_graph_node*)vtr::calloc(pb_type->modes[i].pb_type_children[j].num_pb, sizeof(t_pb_graph_node));
+            int base = flat_index*(pb_type->modes[i].pb_type_children[j].num_pb);
```
vaughnbetz, line 356: 1. Explain this -- what is this mapping?
2. Should this be a helper function?  It seems this may be a specific mapping function/style so putting it in a function will give it a name (helps readability), give a natural place to comment it, and make it easier to change or generalize the mapping function in the future if that ever becomes necessary.

KA7E: flat_index*(pb_type->modes[i].pb_type_children[j].num_pb) is the flattened index, at the level of the child, up until the parent.  

For example, say there are four adults and each has three kids; then the base value (in terms of kid count) at the third adult is 6 (because the first and second adults each have three kids),. The third adult's kids have indices (6+0), (6+1), and (6+2).  

It would be most readable to call some sort of "get child's flat index" function once per child, but then you'd end up computing the same parent base multiple times; I do it outside the k-loop so it only needs to be computed once. 

How would you like me to handle it?  I think either is fine; this function is not called all that many times, so a slight inefficiency for the sake of legibility is probably ok.

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -1914,3 +1929,22 @@ const t_pb_graph_edge* get_edge_between_pins(const t_pb_graph_pin* driver_pin, c
 
     return nullptr;
 }
+
+/* Date:June 8th, 2024
+ * Author: Kate Thurmer
+ * Purpose: This subroutine computes the index of a pb graph node at its
+            level of the pb hierarchy; it is computed by the parent and
+            passed to each child of each child pb type. When the child is
+            a primitive, the computed indes is its flat site index.
```
vaughnbetz, line 1938: indes -> index

vaughnbetz: Slightly better to put this comment above the function declaration instead of the implementation.

KA7E:
QoR check running neuron:

Master VPR:
BB estimate of min-dist (placement) wire length: 609607
Placement estimated critical path delay (least slack): 7.36338 ns
Total wirelength: 778820, average net length: 15.3064
Final critical path delay (least slack): 7.63097 ns
The entire flow of VPR took 603.12 seconds (max_rss 2827.0 MiB)

Master VPR with legalizer-related modifications (this PR and others):
BB estimate of min-dist (placement) wire length: 609607
Placement estimated critical path delay (least slack): 7.36338 ns
Total wirelength: 778820, average net length: 15.3064
Final critical path delay (least slack): 7.63097 ns  
The entire flow of VPR took 547.78 seconds (max_rss 2825.2 MiB)

vaughnbetz:
Thanks, that looks good.  It's odd cpu time sped up by about 10%; probably a load issue I guess.

KA7E:
The modified version was compiling while the baseline was running, but nothing else was happening while the modified version was running.  I thought I had enough resources that this wouldn't matter, but maybe it impacted the runtime.

KA7E:
It is not a laptop, it's a repurposed gaming server (that I like to use so I can control what else is running).  If runtime is a concern I'm happy to re-run each all alone.

vaughnbetz:
I think it's OK ... the odds were low of a slowdown and this pushes them lower still.

PR #2594, opened 2024-06-06
[Docs] Updated Subtree Docs for Upgrading Versions

AlexandreSinger:
Added documentation on how to upgrade the version of a subtree. This is non-trivial since many repos store their versions on different branches. Change the branch of a subtree is basically creating a new subtree, so the old subtree needs to be completely remove from VTR first.


AlexandreSinger:
@vaughnbetz Please review whenever you have a moment. This is a non functional change which just adds documentation on how to upgrade from one version of a subtree to another when the subtree is on a different branch. I do not know if this is the most efficient way of doing it; however, I do think that this is the easiest to write down.

vaughnbetz:
Looks good; thanks.

PR #2597, opened 2024-06-07
[CI] Cleaned Up Artifacts and Resolved Deprecation

AlexandreSinger:
When artifacts were being saved for the regression tests, it was saving
all of the benchmarks as blif files. This is incredibly wasteful (72 MB
per regression test or around 720 MB per CI run). These tests will no
longer save .blif files as artifacts anymore. If anyone needed these
files they could generate them themselves or can add a condition to add
the exact .blif files they may need.
    
The CI was also always generating artifacts (at least for the nightly
    tests) even when the run was cancelled. Made it so the artifacts are
    only saved if the run is not cancelled. This would mean an artifact is
    generated if the run succeeds or fails.

save-output was deprecated. Resolved. See:
    https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,9 +86,10 @@ jobs:
         NUM_CORES: ${{ matrix.cores }}
 
     - name: Upload test results
-      # We always want the test results to be uploaded, even when cancelled.
+      # If the job was not cancelled, we want to save the result (this includes
+      # when the job fails). See warning here:
       # https://docs.github.com/en/actions/learn-github-actions/expressions#always
-      if: ${{ always() }}
+      if: ${{ !cancelled() }}
       # TODO: This runnner is running on a self-hosted CPU. In order to upgrade
       #       to v4, need to upgrade the machine to support node20.
       uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
@soheilshahrouz Are you ok with this change with when artifacts are saved. This change will make it so when a run in cancelled it will not save the artifacts, but it will save the artifacts if the run succeeds or fails. I anticipate there will be more cancelled runs in the future, so this will prevent saving artifacts when they are not needed.

.github/workflows/test.yml
```diff
@@ -86,9 +86,10 @@ jobs:
         NUM_CORES: ${{ matrix.cores }}
 
     - name: Upload test results
-      # We always want the test results to be uploaded, even when cancelled.
+      # If the job was not cancelled, we want to save the result (this includes
+      # when the job fails). See warning here:
       # https://docs.github.com/en/actions/learn-github-actions/expressions#always
-      if: ${{ always() }}
+      if: ${{ !cancelled() }}
```
soheilshahrouz, line 92: Let's say vtr_reg_strong fails after 30 minutes, and I push a new commit to fix the failure (or I cancell the workflow manually). Is the artifact for the failed vtr_reg_strong uploaded?

AlexandreSinger: @soheilshahrouz Thats a great question! I just tested it! Each run (i.e. vtr_reg_strong for example) runs independently of each other. If this run fails it will generate its artifact and that will exist until its deleted (right now thats 90 days from now). If you push another commit, or cancel the run, this artifact will still exist but any test that did not complete yet (neither failed nor succeeded) will not generate an artifact.

Right now when you cancel a run part-way through execution it will stop running the test, then try to generate the artifacts; which in my opinion is a bit strange since the run did not complete.

AlexandreSinger: Here is a CI run for this PR which I cancelled which demonstrates what I have changed:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/9421995802

As you can see vtr_reg_strong's (funny enough) artifact was generated and saved; but the tests that did not complete's aritfacts were not saved.

AlexandreSinger:
@vaughnbetz This has passed CI. I have verified that the artifacts are no longer being collected on cancellation and the overall size of the artifacts decreased from 1000 MBs per run to 250 MBs per run. 120 MBs is being taken up from vtr_reg_nightly_test1, so if we have issues in the future we can see what we need to be collecting for that test.

I have also verified that the deprecation warning is now gone.

PR #2601, opened 2024-06-08
[Route] Added astar_offset Parameter

AlexandreSinger:
Using an astar_offset can help better tune the ordering heuristic for the search used to find the shortest path in the routing graph. It is also necessary to ensure that the heuristic is an underestimate (without setting the astar_fac to 0.0).

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -2499,6 +2499,14 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("1.2")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument(args.astar_offset, "--astar_offset")
+        .help(
+            "Controls the directedness of the timing-driven router's exploration."
+            " It is a subtractive adjustment to the lookahead heuristic."
+            " Values between 0 and 1e-9 are resonable; higher values may increase quality at the expense of run-time.")
+        .default_value("0.0")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
     route_timing_grp.add_argument(args.router_profiler_astar_fac, "--router_profiler_astar_fac")
         .help(
             "Controls the directedness of the timing-driven router's exploration"
--- a/vpr/src/base/vpr_types.h
+++ b/vpr/src/base/vpr_types.h
@@ -1343,6 +1343,8 @@ struct t_placer_opts {
  *             an essentially breadth-first search, astar_fac = 1 is near   *
  *             the usual astar algorithm and astar_fac > 1 are more         *
  *             aggressive.                                                  *
+ * astar_offset: Offset that is subtracted from the lookahead (expected     *
+ *               future costs) in the timing-driven router.                 *
  * max_criticality: The maximum criticality factor (from 0 to 1) any sink   *
  *                  will ever have (i.e. clip criticality to this number).  *
  * criticality_exp: Set criticality to (path_length(sink) / longest_path) ^ *
@@ -1431,6 +1433,7 @@ struct t_router_opts {
     enum e_router_algorithm router_algorithm;
     enum e_base_cost_type base_cost_type;
     float astar_fac;
+    float astar_offset;
     float router_profiler_astar_fac;
     float max_criticality;
     float criticality_exp;
--- a/vpr/src/route/connection_router.cpp
+++ b/vpr/src/route/connection_router.cpp
@@ -1,6 +1,7 @@
 #include "connection_router.h"
-#include "rr_graph.h"
 
+#include <algorithm>
+#include "rr_graph.h"
 #include "binary_heap.h"
 #include "four_ary_heap.h"
 #include "bucket.h"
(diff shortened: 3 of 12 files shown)


vpr/src/route/connection_router.cpp
```diff
@@ -911,12 +914,10 @@ void ConnectionRouter<Heap>::add_route_tree_node_to_heap(
...
     if (!rcv_path_manager.is_enabled()) {
         // tot_cost = backward_path_cost + cost_params.astar_fac * expected_cost;
+        float expected_cost = router_lookahead_.get_expected_cost(inode, target_node, cost_params, R_upstream);
         float tot_cost = backward_path_cost
                          + cost_params.astar_fac
-                               * router_lookahead_.get_expected_cost(inode,
-                                                                     target_node,
-                                                                     cost_params,
-                                                                     R_upstream);
+                               * std::max(0.f, expected_cost - cost_params.astar_offset);
```
soheilshahrouz, line 920: Now that get_expected_cost() is not called here, is it necessary to break this down into 3 lines?

vpr/src/route/connection_router.cpp
```diff
@@ -1,4 +1,5 @@
 #include "connection_router.h"
+#include <algorithm>
```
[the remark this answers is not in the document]
AlexandreSinger: Ah! But it is best practice to put the header file of the file being implemented first, and then follow your proposed style after:

```
<header_for_file>
<STL headers>
<VPR specific headers>
```

I tried looking throughout the route folder, but somehow none of them include from the standard library... thats kinda insane... so I was unable to find "what everyone else does" lol.

I added a return carriage, since at least one file did something similar, to make things more clear:
```
<header_for_file>

<STL headers>
<VPR specific headers>
```

vpr/src/base/read_options.cpp
```diff
@@ -2475,6 +2475,14 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("1.2")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument(args.astar_offset, "--astar_offset")
+        .help(
+            "Controls the directedness of the timing-driven router's exploration."
+            " Acts as an offset for the heuristic."
+            " Values between 0 and 1e-9 are resonable; higher values may increase quality at the expense of run-time.")
```
vaughnbetz, line 2482: Good to say what it does: it is a subtractive adjustment to the lookahead heuristic.

vpr/src/base/vpr_types.h
```diff
@@ -1330,6 +1330,8 @@ struct t_placer_opts {
  *             an essentially breadth-first search, astar_fac = 1 is near   *
  *             the usual astar algorithm and astar_fac > 1 are more         *
  *             aggressive.                                                  *
+ * astar_offset: factor used to offset the expected future costs to         *
```
vaughnbetz, line 1333: "offset that is subtracted from the lookahead (expected future costs) in the timing-driven router."

vaughnbetz:
Should also do a QoR check to compare router runtimes and make sure there isn't a noticeable increase.

AlexandreSinger:
@vaughnbetz FYI, this was the section I was concerned about regarding the RCV path manager:
image

The issue I see here is that the astar_fac is used to scale the raw delay and congestion costs independently (not the weighted sum as its done in the default router). I am not sure the affects this may have on the results but it may cause more critical nets to be biased toward / away from congestion / delay way more than intended.

For this PR I just followed the example and offset at the same point; but I wonder how this may affect RCV's results.

vaughnbetz:
For RCV we need to keep the delay separate from congestion for longer, as we're not targeting min delay anymore but a delay window instead.
RCV also won't work unless A* fac is essentially 1 on the delay part; to hit a delay window you need to estimate pretty precisely, rather than overestimating to get a directed search.  To make the search more directed I think we'd need to add an additional cost outside the delay window computation that was something like (A*fac-1)*remaining delay to make the search more directed.  Or perhaps the resource cost being multiplied by A*fac would be enough to drive toward the target.

I think we should avoid manipulating the delay part of the cost with RCV at all (no multiply by A*fac, no offset, etc.).

AlexandreSinger:
QoR check for these changes:

  | baseline_results.txt | parse_results.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.999587653
num_LAB | 1 | 1
num_DSP | 1 | 1
num_M9K | 1 | 1
num_M144K | 1 | 1
max_vpr_mem | 1 | 1.000038815
num_pre_packed_blocks | 1 | 1
device_grid_tiles | 1 | 1
pack_time | 1 | 1.004174461
placed_wirelength_est | 1 | 1
place_time | 1 | 1.00860537
placed_CPD_est | 1 | 1
routed_wirelength | 1 | 0.999503305
critical_path_delay | 1 | 0.994528741
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 0.994283369
crit_path_route_time | 1 | 0.959547268

Baseline results are from the common commit from Master and this branch. Seems to be no affect on runtime. Rerunning CI.

AlexandreSinger:
@vaughnbetz This PR has passed CI. I have made sure that the documentation is updated with this new CLI parameter. QoR results are shown above on Titan. Seems to be no loss in runtime or quality.

vaughnbetz:
Looks good, thanks.

PR #2603, opened 2024-06-10
Fix high rejection rate for constrained blocks during annealing

soheilshahrouz:
Most proposed moves for blocks with tight floorplan constraints are rejected because the proposed location for a constrained block is not withing the floorplan region. This PR tries to fix this problem.

Running vtr_reg_nightly_test5/vpr_tight_floorplan regression test

#### Types of changes

#### Checklist:


soheilshahrouz:
Results for vtr_reg_nightly_test5/vpr_tight_floorplan test
Circuit: neuron_stratixiv_arch_timing.blif
Architectur: stratixiv_arch_neuron.timing.xml

| Constraint file | Total swaps | Aborted swaps | Placed WL | Placed CPD | Place time |
|----------|----------|----------|----------|----------|----------|
| sixteenth.xml | 1.06 | 0.44 | 0.98 | 1.01 | 1.44 |
| half_blocks_half.xml | 0.98 | 0.54 | 0.90 | 0.89 | 1.31 |
| one_big_partition.xml | 1.01 | 0.99 | 0.99 | 1.02 | 1.29 |

soheilshahrouz:
one_big_partition.xml defines a single partition covering the entire grid. Other constraint files divide the grid into two and sixteen equally-sized regions. For these two constraint files, the number of aborted swaps is halved. This is because intersect_range_limit_with_floorplan_constraints() updates the search range of the target location, making sure tha the selected  location is within the floorplan region. Without this change, the move generator might propose a target location outside the floorplanning region, causing the proposed move be aborted before cost difference evaluation starts.

The placement runtime increases because proposed swaps are not aborted prematurely, and more swaps result in cost difference evalution.

vaughnbetz:
@soheilshahrouz is investigating why one_big_partition slows things down -- there doesn't seem to be a good reason.

vpr/src/place/compressed_grid.h
```diff
@@ -55,6 +55,46 @@ struct t_compressed_block_grid {
         return {cx, cy, layer_num};
     }
 
+    inline t_physical_tile_loc grid_loc_to_compressed_loc_approx_round_up(t_physical_tile_loc grid_loc) const {
+        int cx = OPEN;
+        int cy = OPEN;
+        int layer_num = grid_loc.layer_num;
+
+        auto itr_x = std::upper_bound(compressed_to_grid_x[layer_num].begin(), compressed_to_grid_x[layer_num].end(), grid_loc.x);
```
vaughnbetz, line 63: Comment how this works

vaughnbetz: And you just told me it doesn't work :).

vpr/src/place/compressed_grid.h
```diff
@@ -55,6 +55,46 @@ struct t_compressed_block_grid {
...
+        auto itr_y = std::upper_bound(compressed_to_grid_y[layer_num].begin(), compressed_to_grid_y[layer_num].end(), grid_loc.y);
+        if (itr_y == compressed_to_grid_y[layer_num].begin())
+            cy = 0;
+        else
+            cy = std::distance(compressed_to_grid_y[layer_num].begin(), itr_y - 1);
+
+        return {cx, cy, layer_num};
+    }
+
+    inline t_physical_tile_loc grid_loc_to_compressed_loc_approx_round_down(t_physical_tile_loc grid_loc) const {
```
vaughnbetz, line 78: Doxygen comment.

vpr/src/place/grid_tile_lookup.h
```diff
@@ -31,11 +31,11 @@ class GridTileLookup {
 
     void fill_type_matrix(t_logical_block_type_ptr block_type, vtr::NdMatrix<int, 3>& type_count);
 
-    int region_tile_count(const Region& reg, t_logical_block_type_ptr block_type);
+    int region_tile_count(const Region& reg, t_logical_block_type_ptr block_type) const;
```
vaughnbetz, line 34: Should move the comments from the .cpp file to .h file, and make Doxygen compatible too.

vaughnbetz (commented):
Looks good!  From our discussion the round_up and round_down functions may have weaknesses with large blocks and possibly a bug in general. So check those ones carefully.

soheilshahrouz:
Results for vtr_reg_nightly_test5/vpr_tight_floorplan after fixing grid_loc_to_compressed_loc_approx_round_up()
Circuit: neuron_stratixiv_arch_timing.blif
Architectur: stratixiv_arch_neuron.timing.xml

| Constraint file | Total swaps | Aborted swaps | Placed WL | Placed CPD | Place time |
|----------|----------|----------|----------|----------|----------|
| sixteenth.xml | 1.06 | 0.46 | 0.99 | 1.006 | 1.09 |
| half_blocks_half.xml | 0.97 | 0.48 | 0.92 | 0.907 | 0.93 |
| one_big_partition.xml | 1.01 | 0.99 | 0.99 | 1.023 | 0.95 |

vaughnbetz:
Excellent!  Glad it looks like it is working now.

soheilshahrouz:
It's unclear to me what the expected behavior is when there is a partial overlap between a region and a block larger than 1x1. I'll explain my understanding of the code. Please correct me if I'm wrong.

cluster_floorplanning_legal() checks if a block is within its partition region. This function calls ParitionRegion::is_loc_in_part_reg(), which then calls Region::is_loc_in_reg() for all regions in the PartitionRegion.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/9438f73722a7e5b8baf943ffb85afc46e845f76d/vpr/src/place/place_constraints.cpp#L213-L240

Based on Region::is_loc_in_reg(), a block location is legal if its bottom-left corner is within the floorplanning constraint. Therefore, if only the upper or right part of a block larger than 1x1 overlaps with a region, it doesn't belong to that region.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/9438f73722a7e5b8baf943ffb85afc46e845f76d/vpr/src/base/region.cpp#L45-L69

vaughnbetz:
Hmmm ... looks like I was wrong on this one.  In Quartus I made floorplan regions include a block if any part was touched, but in VTR it looks like we didn't. I am now dimly recalling that this was due to Sarah finding the code more complex to handle that case, so we just decided to document it.  I think we should just add this detail to the documentation.

vaughnbetz:
Update: this looks good, but @soheilshahrouz wants to add a unit test to it before merging.

Issue #2570, opened 2024-05-30
Unreachable code in vpr/src/route/connection_router.cpp


mchuahua:
#### Expected Behaviour

Hi, in [connection_router.cpp](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/connection_router.cpp), line 119-122 has unreachable code. 
#### Current Behaviour

I believe line 116 should return the correct tuple instead, which is reachable code.
#### Possible Solution

Remove lines 119-122 because line 124 returns what is expected when cheapest is found. I may be wrong though, which is why I didn't do a pr.
#### Steps to Reproduce

in connection_router.cpp, lines 100-122 below:
```
   if (cheapest == nullptr) {
        // No path found within the current bounding box.
        //
        // If the bounding box is already max size, just fail
        if (bounding_box.xmin == 0
            && bounding_box.ymin == 0
            && bounding_box.xmax == (int)(grid_.width() - 1)
            && bounding_box.ymax == (int)(grid_.height() - 1)
            && bounding_box.layer_min == 0
            && bounding_box.layer_max == (int)(grid_.get_num_layers() - 1)) {
            VTR_LOG("%s\n", describe_unrouteable_connection(source_node, sink_node, is_flat_).c_str());
            return std::make_tuple(false, nullptr);
        }

        // Otherwise, leave unrouted and bubble up a signal to retry this net with a full-device bounding box
        VTR_LOG_WARN("No routing path for connection to sink_rr %d, leaving unrouted to retry later\n", sink_node);
        return std::make_tuple(true, nullptr);
    }

    if (cheapest == nullptr) {
        VTR_LOG("%s\n", describe_unrouteable_connection(source_node, sink_node, is_flat_).c_str());
        return std::make_tuple(false, nullptr);
    }

```

#### Your Environment

* VTR revision used:
* Operating System and version:
* Compiler version:

vaughnbetz:
Thanks!  Fixed and closing.

PR #2604, opened 2024-06-10
[Route] Removed Unreachable Code

AlexandreSinger:
In an old version of the router, a catch was made to exit the router with no solution and not to retry at max bounding box; however, the function was rewritten to explicitly do this. The old catch statement was left in the code and was unreachable.

Removed the unreachable code.

This resolves issue #2570


AlexandreSinger:
@vaughnbetz This is a very small commit which removes some unreachable code from the router. I noticed this a while back, but did not raise a fix for it. Someone raised an issue on it, so I thought I would raise a quick PR to resolve it.

Issue #2598, opened 2024-06-07
Prevent running the CI if only the docs are updated


ueqri:
#### Proposed Behaviour

We should prevent the CI running (particularly, the [`.github/workflows/test.yml`](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.github/workflows/test.yml)) if only the documentation files are updated.

#### Current Behaviour

Any push or PR that includes ONLY documentation changes (e.g., changes to the developer guide or README) will trigger all current CI workflows, which could be unnecessary and a waste of CI resources.

#### Possible Solution

Add a `path-ignore` filter to the GitHub Actions workflow triggers to filter out the docs changes, https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

> When all the path names match patterns in `paths-ignore`, the workflow will not run. If any path names do not match patterns in `paths-ignore`, even if some path names match the patterns, the workflow will run.

Specifically, we can add `path-ignore` filters to the `push` and `pull_request` triggers (using use push as an example below):
```yaml
push:
  paths-ignore: # Prevent from running if only docs are updated
    - 'doc/**'
    - '**/*README*'
    - '**.md'
    - '**.rst'
```

**Small limitations**: in the PR scenario, the filter will consider all commits (not just a single, or the latest commit) of a PR. In this case, any commit that includes a file not matching the patterns, along with any subsequent commits, will trigger the workflow.

#### Note

Currently, the [VTR online documentation](https://docs.verilogtorouting.org/en/latest/) is built automatically based on the [`.readthedocs.yaml`](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.readthedocs.yaml) through [Read the Docs Services](https://about.readthedocs.com/?ref=readthedocs.com). The build service (independent from GitHub Actions CI) is triggered whenever there is a push to the master branch.

Therefore, skipping all the current GitHub CI workflows won't affect the current online documentation build.

AlexandreSinger:
I am 100% in support of this change! I also agree that the CI should not run if there are non-functional changes.

A potential issue you may run into is that there are required tests in our PR flow (sets of tests which must succeed in order for the code to be merged in). If the CI does not run for documentation changes, the PR may never be merged since it did not pass the required tests. I am not sure the best way to resolve this issue, but it should be resolved.

vaughnbetz:
[...]

PR #2605, opened 2024-06-11
[Docs] Updated Profiling VTR Section in Developer Guide

### LINKED ISSUE #2545: Document how to use Perf for profiling with vpr
The VTR developer guide explains how to use gprof to profile vpr/vtr, but not perf.

#### Proposed Behaviour
We should also document how to run perf and visualize its output.

#### Possible Solution
Add a subsection here.  https://docs.verilogtorouting.org/en/latest/README.developers/#profiling-vtr

ueqri:
Close #2545.

Rewrote the existing Profiling VTR section, specifically the one using GNU `gprof` tool.

Added another subsection to explain how to use the Linux `perf` tool to profile VPR and visualize its output.

### DIFF
--- a/README.developers.md
+++ b/README.developers.md
@@ -1266,47 +1266,96 @@ make CMAKE_PARAMS="-DVTR_IPO_BUILD=off" -j8 vpr
 
 # Profiling VTR
 
-1. Install `gprof`, `gprof2dot`, and `xdot`. Specifically, the previous two packages require python3, and you should install the last one with `sudo apt install` for all the dependencies you will need for visualizing your profile results.
+## Use GNU Profiler gprof
+
+1. **Installation**: Install `gprof`, `gprof2dot`, and `xdot` (optional).
+   1. `gprof` is part of [GNU Binutils](https://www.gnu.org/software/binutils/), which is a commonly-installed package alongside the standard GCC package on most systems. `gprof` should already exist. If not, use `sudo apt install binutils`.
+   2. `gprof2dot` requires python3 or conda. You can install with `pip3 install gprof2dot` or `conda install -c conda-forge gprof2dot`.
+   3. `xdot` is optional. To install it, use `sudo apt install`.
     ```
-    pip3 install gprof
+    sudo apt install binutils
     pip3 install gprof2dot
-    sudo apt install xdot
+    sudo apt install xdot # optional
     ```
 
     Contact your administrator if you do not have the `sudo` rights.
 
-2. Use the CMake option below to enable VPR profiler build.
+2. **VPR build**: Use the CMake option below to enable VPR profiler build.
     ```
     make CMAKE_PARAMS="-DVTR_ENABLE_PROFILING=ON" vpr
     ```
 
-3. With the profiler build, each time you run the VTR flow script, it will produce an extra file `gmon.out` that contains the raw profile information. 
-    Run `gprof` to parse this file. You will need to specify the path to the VPR executable.
+3. **Profiling**:
+   1. With the profiler build, each time you run the VTR flow script, it will produce an extra file `gmon.out` that contains the raw profile information. Run `gprof` to parse this file. You will need to specify the path to the VPR executable.
+        ```
+        gprof $VTR_ROOT/vpr/vpr gmon.out > gprof.txt
+        ```
+
+   2. Next, use `gprof2dot` to transform the parsed results to a `.dot` file (Graphviz graph description), which describes the graph of your final profile results. If you encounter long function names, specify the `-s` option for a cleaner graph. For other useful options, please refer to its [online documentation](https://github.com/jrfonseca/gprof2dot?tab=readme-ov-file#documentation).
+        ```
+        gprof2dot -s gprof.txt > vpr.dot
+        ```
+
+   - Note: You can chain the above commands to directly produce the `.dot` file:
+        ```
+        gprof $VTR_ROOT/vpr/vpr gmon.out | gprof2dot -s > vpr.dot
+        ```
+
+4. **Visualization**:
+   - **Option 1** (Recommended): Use the [Edotor](https://edotor.net/) online Graphviz visualizer.
+     1. Open a browser and go to [https://edotor.net/](https://edotor.net/) (on any device, not necessarily the one where VPR is running).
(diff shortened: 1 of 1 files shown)


ueqri:
#### Expectations:

- The "Test" and "Container" CI workflows are skipped as expected, according to https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2605/checks.
- The Read the Docs are also built successfully, https://vtr-verilog-to-routing--2605.org.readthedocs.build/en/2605/

image

#### Remaining Issues:

"Some checks haven’t completed yet": It appears to be the issue Alex mentioned in https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2598#issuecomment-2155633380.
> A potential issue you may run into is that there are required tests in our PR flow (sets of tests which must succeed in order for the code to be merged in). If the CI does not run for documentation changes, the PR may never be merged since it did not pass the required tests. I am not sure the best way to resolve this issue, but it should be resolved.

Moreover, here is the exact [troubleshoot reference](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks) from GitHub.

> If a workflow is skipped due to [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) or a [commit message](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.
> 
> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success". For more information, see "[Using conditions to control job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution)."
#### Potential Solutions:

- Solution 1: Maybe we can fix it by disabling the 'Require status checks to pass' in the Branch Protection Rule Settings.
- Solution 2: Instead of using path filtering on the CI trigger, we can use job conditions to prevent CI job execution for unnecessary tasks. However, this requires using a third-party [`dorny/paths-filter@v3`](https://github.com/dorny/paths-filter) component as the first job step, rather than just relying on a GitHub CI primitive as we did before.

README.developers.md
```diff
@@ -1266,47 +1266,96 @@ make CMAKE_PARAMS="-DVTR_IPO_BUILD=off" -j8 vpr
 
 # Profiling VTR
 
-1. Install `gprof`, `gprof2dot`, and `xdot`. Specifically, the previous two packages require python3, and you should install the last one with `sudo apt install` for all the dependencies you will need for visualizing your profile results.
+## Use GNU Profiler gprof
+
+1. **Installation**: Install `gprof`, `gprof2dot`, and `xdot` (optional).
+   1. `gprof` is part of [GNU Binutils](https://www.gnu.org/software/binutils/), which is a commonly-installed package alongside the standard GCC package on most systems. `gprof` should already exist. If not, use `sudo apt install binutils`.
+   2. `gprof2dot` requires python3 or conda. You can install with `pip3 install gprof2dot` or `conda install -c conda-forge gprof2dot`.
+   3. `xdot` is optional. To install it, use `sudo apt install`.
```
vaughnbetz, line 1274: Should be `sudo apt install xdot`

README.developers.md
```diff
@@ -1266,47 +1266,96 @@ make CMAKE_PARAMS="-DVTR_IPO_BUILD=off" -j8 vpr
...
-    gprof $VTR_ROOT/vpr/vpr gmon.out | gprof2dot -s > vpr.dot
-    ```
+3. **Profiling**: `perf` needs to know the process ID (i.e., pid) of the running VPR you want to monitor and profile, which can be obtained using the Linux command `top -u <username>`.
+   - **Option 1**: Real-time analysis
+        ```
+        sudo perf top -p <vpr pid>
+        ```
+   - **Option 2** (Recommended): Record and offline analysis
+
+        Use `perf record` to record the profile data and the call graph. (Note: The argument `lbr` for `--call-graph` only works on Intel platforms. If you encounter issues with call graph recording, please refer to the [`perf record` manual](https://perf.wiki.kernel.org/index.php/Latest_Manual_Page_of_perf-record.1) for more information.)
```
vaughnbetz, line 1341: If you are on a non-Intel platform what should you do?  Just leave out --call-graph lbr ? Also describe what leaving it out does -- I believe perf still works but becomes more resource intensive.

vaughnbetz (changes requested):

vaughnbetz:
Thanks!  Looks good -- just a couple of suggestions.

README.developers.md
ueqri: Fixed! 🤣

README.developers.md
ueqri: > If you are on a non-Intel platform what should you do?

For non-Intel platforms, the argument can be set to `fp` which utilizes frame pointer to produce call graph (sometimes might be inaccurate) or `dwarf` using debugging information generated by compiler (resource-consuming to generate call graphs from this during profiling).

Q: Would the following changes work? I was worried that it might be too long to read.
#### Edited:
Use `perf record` to record the profile data and the call graph. (Note: By default, `perf` uses the frame pointer to generate call graphs, which might produce inaccurate results if the program is highly optimized by the compiler. On Intel platforms, it is recommended to specify `lbr` for `--call-graph`, as it is less affected by compiler optimizations, does not require specific compiler options, and uses fewer resources, e.g., less disk storage for storing profiling results. On other platforms, use `--call-graph dwarf` if available. This requires the compiler to produce debugging information in DWARF format and is resource-intensive. For more information on call graph recording, please refer to the [`perf record` manual](https://perf.wiki.kernel.org/index.php/Latest_Manual_Page_of_perf-record.1) and [StackOverflow discussion](https://stackoverflow.com/questions/57430338/what-do-the-perf-record-choices-of-lbr-vs-dwarf-vs-fp-do).)

```bash
sudo perf record --call-graph -p <vpr pid> # use `--call-graph lbr` on Intel platforms or `--call-graph dwarf` on other platforms
```

PR #2607, opened 2024-06-12
Redundant Physical Type

amin1377:
`get_physical_tile_type` is a function that returns a pointer to the physical tile corresponding to a `ClusterId`. There is a similar function in `vpr_utils` with similar functionality (`physical_tile_type`). The only difference is that `get_physical_tile_type` will return a tile type even if the cluster is not placed (for drawing purposes). Instead of having a separate function, I call that function directly if the block is not placed.


vaughnbetz:
Good to also open the graphics and check that everything works. Once you confirm all is good, go ahead and merge.

amin1377:
I tested the graphics, and it seems to be working as before.

PR #2608, opened 2024-06-12
Use Simple Place Delay Model By Default

amin1377:
In this PR, the default place delay mode is changed from `delt` (which runs the router for each `dx` and `dy` and stores the minimum cost across sample locations) to `simple` (which gets the minimum cost for each `dx` and `dy` from the information stored in the router lookahead).


amin1377:
VTR Large: 
image

amin1377:
Titan:
screenshot_2024-06-06_092317

amin1377:
QoR on **all** koios circuits:
image

vaughnbetz:
QoR data for Koios is actually for the placement refactoring ... ignore for this PR.

vaughnbetz:
@amin1377 : is this good to merge (still a few CI failures)? @nedsels will benefit from the larger QoR tolerances in another PR.

PR #2610, opened 2024-06-12
Enable VTR to flush stdout and stderr before assertion

robluo:
When debugging, I find that sometimes when the program hits an assertion, some log would not be printed as expected. Manually flushing before abort() brings more consistent printing.

### DIFF
--- a/libs/libvtrutil/src/vtr_assert.cpp
+++ b/libs/libvtrutil/src/vtr_assert.cpp
@@ -16,6 +16,8 @@ void handle_assert(const char* expr, const char* file, unsigned int line, const 
         fprintf(stderr, " (%s)", msg);
     }
     fprintf(stderr, ".\n");
+    fflush(stdout);
+    fflush(stderr);
     std::abort();
 }
 


libs/libvtrutil/src/vtr_assert.cpp
```diff
@@ -16,6 +16,8 @@ void handle_assert(const char* expr, const char* file, unsigned int line, const
         fprintf(stderr, " (%s)", msg);
     }
     fprintf(stderr, ".\n");
+    fflush(stdout);
+    fflush(stdin);
```
soheilshahrouz, line 20: What does it mean to flush the input stream?

libs/libvtrutil/src/vtr_assert.cpp
```diff
@@ -16,6 +16,8 @@ void handle_assert(const char* expr, const char* file, unsigned int line, const
         fprintf(stderr, " (%s)", msg);
     }
     fprintf(stderr, ".\n");
+    fflush(stdout);
```
soheilshahrouz, line 19: Isn't it better to flush both stdout and stderr?

libs/libvtrutil/src/vtr_assert.cpp
robluo: that have been fixed, I meant stdout and stderr.

PR #2615, opened 2024-06-13
legalizer frontend

KA7E:
Frontend modifications to incorporate legalizer.

#### Description

vpr_api.cpp:
1. Added code to invoke the legalizer in vpr_pack_flow as follows: 
   if doPacking=STAGE_LOAD and packer_opts.load_flat_placement=true, this code:
    a. builds the device grid per a specified fixed device layout
    b. calls load_flat_place (impl coming soon), which will:
         - load and legalize the input flat placement file
         - print a clustered netlist
         - print a fixed clusters (partial .place) file
    c. clears the device grid
    d. if placement will run next, specifies the fixed clusters file as a constraints file.
    e. resumes normal load packing behavior (load .net file, do various checks)
2.  Added a function call to print a flat placement file after VPR placement

SetupGrid.h:
Added default arguments to create_device_grid so that it can be called with a fixed device layout without other args (that are only used when auto-sizing).

read_options.h/cpp:
Added the following command line inputs: 
--legalize, enables the legalizer
--flat_place_file <fname>, legalizer input file
--write_fix_clusters <fname> (default: fix_clusters.out), for legalizer placed clusters output
--write_flat_place <fname> (default: flat_place.out), for post-placement flat placement output

SetupVPR.cpp:
Added the following fields to the FileNameOpts data structure: 
 - FlatPlaceFile
 - write_constraints_file
 - write_flat_place_file
Added a do_legalize option, which sets doPacking to STAGE_LOAD and sets a flag (packer_opts.load_flat_placement) instructing the packer to load from a flat placement file.

CheckSetup.cpp:
1. Added a check that errors out if a fixed device layout is not specified for legalization.
2. Added a check that errors out if an input constraints file is specified (using --fix_clusters <filename>) when running legalization, since the legalizer will provide cluster placement info.
 
read_place.h/cpp:
Added an optional boolean input to print_place which defaults to true (in which case the function's behavior does not change). If false, the header is not printed, unplaced blocks (loc.x = -1) are skipped, and the net_file and net_id parameters are unused and can be set to nullptr. This is useful if the output will be used as a constraints (fix clusters) file.
 
added load_flat_place.h/cpp:
Function to print a flat placement file.
Function to load flat placement (impl coming soon).

#### Motivation and Context

Incorporates the legalizer.

Verified that, if legalizer option is not enabled, vpr runs as usual with no QoR change.
The legalization capability has been tested on all Titan23 benchmarks.

#### Types of changes

- [X ] New feature (change which adds functionality)

#### Checklist:


vpr/src/base/load_flat_place.cpp
```diff
@@ -0,0 +1,55 @@
...
+    t_pl_loc loc = g_vpr_ctx.placement().block_locs[iblk].loc;
+    size_t bnum = size_t(iblk);
+
+    for (auto atom : atoms) {
+        t_pb_graph_node* atom_pbgn = atom_ctx.lookup.atom_pb(atom)->pb_graph_node;
+        fprintf(fp, "%s  %d %d %d %d #%zu %s\n", atom_ctx.nlist.block_name(atom).c_str(),
+                                                loc.x, loc.y, loc.sub_tile,
+                                                atom_pbgn->flat_site_index,
+                                                bnum,
+                                                atom_pbgn->pb_type->name);
```
vaughnbetz, line 23: I'd just do the size_t() case here so you don't create a new temporary variable bnum that is just an alias for iblk.  I imagine the size_t is just so you can print this number, so doing the size_t cast in the print is slightly easier to understand in my opinion.

vpr/src/base/load_flat_place.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include "vpr_types.h"
+
+/**
+ * @brief A function that prints a flat placement file
+ */
+void print_flat_placement(const char* flat_place_file);
+
+/**
+ * @brief A function that loads and legalizes a flat placement file
```
vaughnbetz, line 12: It would be good to briefly explain here or point at to a high-level comment elsewhere for what a flat placement is (an x,y,sub_tile and primitive_index_in_cluster) file that implies both a packing and a placement).

vaughnbetz: layer is for 3D devices; when you implement this function you an assert that there is only one layer for now if you don't support multi-layer devices (and you probably don't). The grid data structure knows how many layers there are.

vaughnbetz: Should comment the parameters; I'm not sure of what exactly archi_id is. vpr_setup is pretty clear but still good to have a basic comment for consistency.

vpr/src/base/read_options.cpp
```diff
@@ -1310,6 +1310,11 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .action(argparse::Action::STORE_TRUE)
         .default_value("off");
 
+    stage_grp.add_argument<bool, ParseOnOff>(args.do_legalize, "--legalize")
+        .help("Legalize a flat placement")
```
vaughnbetz, line 1314: Good to expand this a bit. Briefly explain what a flat placement is (atom-level placement that controls packing and placement).

vpr/src/base/read_options.cpp
```diff
@@ -1627,6 +1636,18 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .help("Writes out new floorplanning constraints based on current placement to the specified XML file.")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    file_grp.add_argument(args.write_constraints_file, "--write_fix_clusters")
+        .help(
+            "Output file containing fixed locations of legalized input clusters.")
```
vaughnbetz, line 1641: Is this usable only with an input flat placement?  If so, good to mention that in the .help.

vpr/src/base/read_place.h
```diff
@@ -17,8 +17,15 @@ void read_place(
  */
 void read_constraints(const char* constraints_file);
 
+/**
+ * This function prints out a place file.
+ * @param is_place_file: defaults to true. If false, does not print file header; this is useful if
+ *                       the output will be used as a constraints file. If is_place_file is false,
+ *                       net_file and net_id parameters are not used and can be set to nullptr.
```
vaughnbetz, line 24: Should also explain that if is_place_file is false, a partial placement can be printed (only placed block will have their placement output).

vpr/src/base/read_place.cpp
```diff
@@ -318,15 +319,21 @@ void print_place(const char* net_file,
...
+                net_id);
+        fprintf(fp, "Array size: %zu x %zu logic blocks\n\n", device_ctx.grid.width(), device_ctx.grid.height());
+        fprintf(fp, "#block name\tx\ty\tsubblk\tlayer\tblock number\n");
+        fprintf(fp, "#----------\t--\t--\t------\t-----\t------------\n");
+    }
 
     if (!place_ctx.block_locs.empty()) { //Only if placement exists
         for (auto blk_id : cluster_ctx.clb_nlist.blocks()) {
+            // if block is not placed, skip (useful for printing legalizer output)
+            if (!is_place_file && (place_ctx.block_locs[blk_id].loc.x == -1)) {
```
vaughnbetz, line 334: I suggest moving 
static constexpr int INVALID_X = -1;

from initial_placement.cpp to vpr_types.h (right above the definition of t_place_loc, and adding a comment saying blocks that are unplaced/have an invalid placement can test for this value. 
Then change the == -1 test to == INVALID_X

vpr/src/base/vpr_api.cpp
```diff
@@ -604,10 +611,41 @@ bool vpr_pack_flow(t_vpr_setup& vpr_setup, const t_arch& arch) {
             vpr_load_packing(vpr_setup, arch);
         } else {
             VTR_ASSERT(packer_opts.doPacking == STAGE_LOAD);
-            //Load a previous packing from the .net file
-            vpr_load_packing(vpr_setup, arch);
-            //Load cluster_constraints data structure here since loading pack file
-            load_cluster_constraints();
+
```
vaughnbetz, line 614: This should be turned into a helper function that does the overall flat placement loading control flow (basically all this code); it keeps what is already a big overall control routine a bit higher-level and understandable.

vpr/src/base/vpr_api.cpp
```diff
@@ -410,6 +413,7 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
             std::cout << "failed placement" << std::endl;
             return false; //Unimplementable
         }
+        print_flat_placement(vpr_setup.FileNameOpts.write_flat_place_file.c_str());
```
vaughnbetz, line 416: 1. It looks like we unconditionally print a flat placement file (which may be OK; just checking). What is the default filename?
2. I think this print_flat_placement should be pushed down into the vpr_place_flow routine (right by wherever the print_placement call is) since it keeps the high level control routine more lean (easier to understand) and also it's more logical to print all placement files together.

vaughnbetz (changes requested):
Some suggested changes.
One more thing: we'll need a documentation topic for the new file format and command line options. The high level documentation can be in the new file format description (that matches what we do for other things, like floorplanning constraints).

Between the documentation and some high level commenting, the overall flow & use case of a flat placement should be clear (right now the big picture isn't documented).  The separate documentation can be done in a separate PR if more convenient.

We should also define if a flat placement needs to include a placement for every atom or if only some is OK (in that high level documentation).

vpr/src/base/load_flat_place.h
KA7E: I've added this definition in comments and will also include in the documentation.

I've added a check for num layers > 1 in vpr_api just before attempting to load the flat placement.

arch_id was a placeholder (to avoid a CI failure for unused args); it is now t_arch& arch.  I added a new function, vpr_load_flat_place(vpr_setup, arch).  These args are the same as those for vpr_pack_flow, vpr_place_flow, and vpr_route_flow.

vpr/src/base/vpr_api.cpp
KA7E: I changed it so you have to request it to print a flat placement file after placement; I also added a post-legalizer echo file to print the flat placement (before VPR places orphan clusters).

KA7E:
@AlexandreSinger can you help me figure out why these keep failing?

KA7E:
Silly me, forgot to delete the old (unconditional) call to print_flat_placement when I moved it inside a function and added a condition.  @vaughnbetz can you take a look once CI has finished?

vaughnbetz:
Looks good; thanks.

PR #2618, opened 2024-06-17
Fix segfault in atom_cluster_noc_group_check()

soheilshahrouz:
g_vpr_ctx.cl_helper().atom_noc_grp_id is not initialized when try_pack() is not called. This causes a segfault when re-clustering API is used as atom_cluster_noc_group_check() tries to access an uninitialized vector. This PR fixes this problem by check whether the vector is initilzided.

### DIFF
--- a/vpr/src/pack/cluster_util.cpp
+++ b/vpr/src/pack/cluster_util.cpp
@@ -1384,7 +1384,8 @@ bool atom_cluster_noc_group_check(AtomBlockId blk_id,
                                   ClusterBlockId clb_index,
                                   int verbosity,
                                   NocGroupId& temp_cluster_noc_grp_id) {
-    const NocGroupId atom_noc_grp_id = g_vpr_ctx.cl_helper().atom_noc_grp_id[blk_id];
+    const auto& atom_noc_grp_ids = g_vpr_ctx.cl_helper().atom_noc_grp_id;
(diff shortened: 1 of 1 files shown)


vpr/src/pack/cluster_util.cpp
```diff
@@ -1384,7 +1384,8 @@ bool atom_cluster_noc_group_check(AtomBlockId blk_id,
                                   ClusterBlockId clb_index,
                                   int verbosity,
                                   NocGroupId& temp_cluster_noc_grp_id) {
-    const NocGroupId atom_noc_grp_id = g_vpr_ctx.cl_helper().atom_noc_grp_id[blk_id];
+    const auto& atom_noc_grp_ids = g_vpr_ctx.cl_helper().atom_noc_grp_id;
```
vaughnbetz, line 1387: I suggest a comment saying why this is necessary. Can be added after CI completes / in another PR so we get this one in quickly if you like.

vaughnbetz:
Thanks for the fast action @soheilshahrouz .  
Tagging @KA7E as she has a dependency on this one.
Do we have any NoC checks in the fast to run tests (vtr_strong etc.)?  This has already passed the fast checks, which is good, but I'm not sure if we have a (tiny) NoC test in those ones.  If not it would be good to add one @soheilshahrouz (in another PR) so we have a very fast NoC test.

vaughnbetz:
Merging since CI passed, but I still think this should have a comment added @soheilshahrouz

PR #2623, opened 2024-06-20
3D Placement Constraints

soheilshahrouz:
This PR enhances the supoort for 3D placement constraints to allow the user to specify a layer range for each floorplanning constraint.

#### Motivation and Context
Each floorplan constraint is limited to a single layer. If the user wants to define a cubic floorplan constraint, she/he has to define multiple rectangular floorplan constraint with the same (x,y) coordinates but for different layers. This PR lets user defne a single floorplan region spanning across multiple layers.

#### Types of changes

#### Checklist:


vaughnbetz:
The FASM unit test failure seems to be due to a change in the default layers returned for placement constraints on a 2D device:
2024-06-24T20:23:40.5141020Z [0m[0;37m/home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/vpr/test/test_vpr_constraints.cpp:51: [0m[1;31mFAILED:
2024-06-24T20:23:40.5143053Z [0m[0;36m  REQUIRE( def_region.get_region_bounds().get_layer_range() == std::make_pair(-1, -1) )
2024-06-24T20:23:40.5144059Z [0mwith expansion:
2024-06-24T20:23:40.5144514Z [1;33m  {?} == {?}

Updating the REQUIRE to pair(0,0) should work.

@soheilshahrouz also tells me that if we make a custom printer class/method for std::pair catch2 will print objects instead of {?} in the failure.

vpr/src/base/vpr_constraints.xsd
```diff
@@ -33,6 +33,8 @@
     <xs:attribute name="y_low" type="xs:int" use="required"/>
     <xs:attribute name="x_high" type="xs:int" use="required"/>
     <xs:attribute name="y_high" type="xs:int" use="required"/>
+    <xs:attribute name="layer_low" type="xs:int"/>
```
vaughnbetz, line 36: Write a validator somewhere after the line or file is read in that layer_low <= layer_high, and both are >= 0. If you know the grid size at that point, also check both are in the grid.

Check that the xlow, etc. values are also validated in the same way.

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -144,7 +143,7 @@ void setup_vpr_floorplan_constraints_cutpoints(VprConstraints& constraints, int
 
             Region reg;
             // This function has not been tested for multi-layer grids. An assertion is used earlier to make sure that the grid has only one layer
-            reg.set_region_rect({xmin, ymin, xmax, ymax, 0});
+            reg.set_region_bounds({xmin, ymin, xmax, ymax, 0, n_layers-1});
```
vaughnbetz, line 146: Should test this and remove this comment.

vpr/src/base/vpr_context.h
```diff
@@ -520,6 +520,17 @@ struct FloorplanningContext : public Context {
...
 
+    /**
+     * @brief Floorplanning constraints in the compressed grid coordinate system.
+     *
+     * Each clustered block has a logical type with a corresponding compressed grid.
+     * Compressed floorplanning constraints are calculated by translating the grid locations
+     * of floorplanning regions to compressed grid locations. To ensure regions do not enlarge:
+     * - The bottom left corner is rounded up to the nearest compressed location.
+     * - The top right corner is rounded down to the nearest compressed location.
+     */
```
vaughnbetz, line 531: Update the comment to say how this works.  Also, should we make the first index the ClusterBlockId (should be slightly faster for 3D, multi-layer case, although for 2D case having layer be the first index could be faster)?
Also, should we just do this intersection in normal grid space?
Comment indices [0..grid.num_layers-1][0..numClusters-1]

vpr/src/draw/draw_floorplanning.cpp
```diff
@@ -100,16 +100,17 @@ static void highlight_partition(ezgl::renderer* g, int partitionID, int alpha) {
...
     for (int region = 0; (size_t)region < regions.size(); region++) {
-        const auto reg_coord = regions[region].get_region_rect();
+        const vtr::Rect<int>& reg_coord = regions[region].get_region_bounds().get_rect();
+        const auto [layer_begin, layer_end] = regions[region].get_region_bounds().get_layer_range();
 
         //TODO: 0 should be replaced with the actual z value of the region when graph is 3D
-        ezgl::rectangle top_right = draw_coords->get_absolute_clb_bbox(reg_coord.layer_num,
-                                                                       reg_coord.xmax,
-                                                                       reg_coord.ymax,
+        ezgl::rectangle top_right = draw_coords->get_absolute_clb_bbox(layer_begin,
```
vaughnbetz, line 107: Should draw what is on the selected layers --> for each region check if it intersects the layer range that is visible.  If so, draw.  If not, don't draw.

vaughnbetz: Or just loop through all visible layers, drawing every region that intersects that layer with the specified transparency. (This is even better I think).

vpr/src/place/grid_tile_lookup.h
```diff
@@ -1,57 +1,89 @@
...
+    * @param block_type The type of logical block.
+    * @param type_count The matrix to be filled with cumulative subtiles count.
+    */
+   void fill_type_matrix(t_logical_block_type_ptr block_type, vtr::NdMatrix<int, 3>& type_count);
+
+   /**
+    * @brief Stores the cumulative total of subtiles available at each location in the grid for each block type.
+    *
+    * Therefore, the length of the vector will be the number of logical block types. To access the cumulative
+    * number of subtiles at a location, you would use block_type_matrices[iblock_type][x][y] - this would
```
vaughnbetz, line 75: Update these comments to include layer.

vaughnbetz (commented):
Half done ... getting slow.

vpr/src/place/place_constraints.cpp
```diff
@@ -114,20 +111,21 @@ PartitionRegion update_macro_head_pr(const t_pl_macro& pl_macro, const Partition
...
+                                       const PartitionRegion& grid_pr) {
     const std::vector<Region>& block_regions = head_pr.get_regions();
     PartitionRegion macro_pr;
 
-    for (const auto& block_region : block_regions) {
+    for (const Region& block_region : block_regions) {
         Region modified_reg;
 
-        const auto block_reg_coord = block_region.get_region_rect();
+        vtr::Rect<int> block_reg_rect = block_region.get_region_bounds().get_rect();
```
vaughnbetz, line 124: Similar code twice -- could make a helper function.

vaughnbetz (commented):
Looks good.
Should also add a region constrained case to one of our 3D architecture tests.  Small design, A few different kinds of regions (multi-layer, single location, single location with subtile, etc.)

vpr/src/pack/constraints_report.cpp
soheilshahrouz, line 1: The function `floorplan_constraints_regions_overfull()` compares the number of blocks assigned to a region with the number it can accommodate. If a region has too many blocks, the packing process is repeated to reduce the number of clustered blocks assigned to it by packing atoms more aggressively.

When a partition has multiple regions, the function checks if all regions can accommodate all the blocks assigned to the partition. This can cause packing to fail if a tight partition has multiple regions. We need to check partitions as a whole rather than individual regions, but regions within a partition must not overlap. This does not reduce the expressiveness of placement constraints, however, it may increase the number required regions to define irregularly-shaped partitions.

Even if we fix the multi-region partition problem, partitions might still overlap. For example, if partitions A and B can each hold 100 LABs but overlap by 50%, they can only accommodate 150 blocks together. The packing algorithm may assign each of them 80 LABs assuming that the physical blocks within a partition are owned by it.

Checking for a legal assignment of blocks to partitions is difficult when partitions overlap. We can prevent overlapping, but it would reduce the flexibility of placement constraints. If each grid location is included in at most two partitions, we can calculate shared physical blocks and check if the total assigned blocks are less than the combined capacity of the overlapping partitions.

vaughnbetz: We should document that it is best practice to define region constraints such that the regions do not overlap as that increases the probability of placement failures.  I'd probably leave it at that; we'll detect the failure in placement instead of clustering, which is not too bad.  Possibly the clusterer could have resolved the failure if we had detected it earlier, but we'd be getting into some pretty weird cases and writing quite a bit of code to try to handle them.

The first issue of checking if a region can accommodate all the blocks of a partition sounds like a clear bug though, which should be fixed by summing up the blocks in all the parts of a partition.  We could ignore overlaps in that (again relying on the documentation to discourage that), or we could make a more accurate union of regions routine that marks locations on the grid as they are covered, and then sums them up.  We could call that more accurate routine only if a partition has multiple regions; that would stop bad CPU behaviour if you had many small partitions (i.e. nearly N, where N is the number of tiles in the device, as in that case straightforward grid-filling code could lead to an O(N^2) loop to check the occupancy of all partitions).

Issue #2612, opened 2024-06-13
high fanout routing to heap code -- only count CHANX and CHANY nodes to determine if enough has been added.


vaughnbetz:
The high fanout net routing code currently counts both the total nodes_added and chan_nodes_added in different tests to see if we need to add more of the partial routing.  This looks wrong: we should always be using chan_nodes_added for all tests and the nodes_added variable should be deleted.  

I think this code is messing up the parallel router with net decomposition and flat routing (although that has not been confirmed yet). In any case, this change would simplify the code slightly and make it more logical.

#### Expected Behaviour
Make this change, test QoR is OK.
@duck2 should then check flat routing with net decomposition again.

#### Context
It has potentially caused @duck2 mental anguish :).

mirjanastojilovic:
Thanks, Vaughn!

mirjanastojilovic:
I think it makes sense to use chan_nodes_added in the tests (line 1057) and delete nodes_added variable.

It appears (line 1033) that only the flat router checks if the node is a wire before incrementing chan_nodes_added. Unless I am missing something, I would expect this check to be performed irrespective of whether the router is flat or not. What do you think?

vaughnbetz:
Agreed.

PR #2624, opened 2024-06-20
CHAN Node Count - High Fanout RT

amin1377:
This PR addresses the issue #2612


vaughnbetz:
Code looks good and CI is good. If the QoR check also looks good, go ahead and merge.

amin1377:
QoR on Titan:
image
The number of heap operations is also similar (on average less that 1% difference)

Issue #2611, opened 2024-06-13
Remove has_path_to_sink


vaughnbetz:
The router has a function that was added for the initial 3D routing (from OPINs) that checks some high-level architecture variables and decides whether or not to expand a node. We shouldn't do this; the lookahead should (and according to @amin1377, does) instead highly cost nodes that are on a layer such that they don't have good paths to the target sink.

Hence this function is bad -- we should figure out the routing path solely from the rr-graph, not high-level architecture variables.

#### Expected Behaviour
We should be able to delete this function and still achieve good 3D routing runtime.

#### Current Behaviour
Things are working for the OPIN crossing case, but not the 3D switch block case.  Maybe this code is involved. In any case, it is messy.

#### Possible Solution
Delete this function and test.

vaughnbetz:
@amin1377 : can we close this now as completed?

PR #2626, opened 2024-06-21
Remove `has_path_to_sink`

amin1377:
This PR addresses the Issue #2611

### DIFF
--- a/vpr/src/route/connection_router.cpp
+++ b/vpr/src/route/connection_router.cpp
@@ -835,9 +835,6 @@ void ConnectionRouter<Heap>::add_route_tree_to_heap(
     /* Pre-order depth-first traversal */
     // IPINs and SINKS are not re_expanded
     if (rt_node.re_expand) {
-        if (target_node.is_valid()) {
-            return;
-        }
         add_route_tree_node_to_heap(rt_node,
                                     target_node,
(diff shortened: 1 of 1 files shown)


vpr/src/route/connection_router.cpp
```diff
@@ -851,7 +835,7 @@ void ConnectionRouter<Heap>::add_route_tree_to_heap(
     /* Pre-order depth-first traversal */
     // IPINs and SINKS are not re_expanded
     if (rt_node.re_expand) {
-        if (target_node.is_valid() && !has_path_to_sink(rr_nodes_, rr_graph_, RRNodeId(rt_node.inode), RRNodeId(target_node), only_opin_inter_layer)) {
+        if (target_node.is_valid()) {
```
vaughnbetz, line 838: What is this check & code doing?

vaughnbetz: Seems like we would always just return?

vaughnbetz (commented):
There is some strange looking code here ... please take a look at add_route_tree_to_heap.  The first condition with target_node.is_valid() looks backwards?

vaughnbetz:
Looks good; will merge after some QoR data has been checked and posted.

PR #2630, opened 2024-06-25
RR graph edge creation uses a incorrect switch type for decremental tracks

saaramahmoudi:
#### Description
While creating RR edge, incremental and decremental tracks use the same seg_details of inc_track. Although this is incorrect, it doesn't affect the quality of results because we don't have an architecture that uses different muxes for incremental and decremental tracks. So, I don't expect anything to change by this fix. 

#### Types of changes


vaughnbetz:
Thanks Sara.  Will merge once CI is green, although as you say this looks quite safe.

PR #2644, opened 2024-07-13
[vpr][place] refactoring variables in net_cost_handler

robluo:
net_cost, proposed_net_cost, and bb_updated_before grouped in a struct. bb_updated_before renamed to bb_update_status.

performance data is compared in:
[https://docs.google.com/spreadsheets/d/1VVvTxhVbGLO6Hq9d2ZmZqi7IWLRDFhMkFIhwDXQtYcw/edit?usp=sharing]

Geomean of vtr flow runtime on titan quick qor increased by 3%. Routed wire length remains unchanged.

### DIFF
--- a/vpr/src/place/net_cost_handler.cpp
+++ b/vpr/src/place/net_cost_handler.cpp
@@ -55,10 +55,12 @@ static vtr::NdMatrix<float, 2> chanx_place_cost_fac({0, 0}); // [0...device_ctx.
 static vtr::NdMatrix<float, 2> chany_place_cost_fac({0, 0}); // [0...device_ctx.grid.height()-2]
 
 /**
- * @brief For the first two element in the struct: Cost of a net, and a temporary cost of a net used during move assessment. 
+ * @brief For each of the vector in this struct, there is one entry per cluster level net: 
+ * [0...cluster_ctx.clb_nlist.nets().size()-1].
+ * net_cost and proposed_net_cost: Cost of a net, and a temporary cost of a net used during move assessment. 
  * We also use negative cost values in proposed_net_cost as a flag to indicate that 
  * the cost of a net has not yet been updated.
- * For the last one element in the struct: Flag array to indicate whether the specific bounding box has been updated
+ * bb_update_status: Flag array to indicate whether the specific bounding box has been updated
  * in this particular swap or not. If it has been updated before, the code    
  * must use the updated data, instead of the out-of-date data passed into the 
  * subroutine, particularly used in try_swap(). The value NOT_UPDATED_YET     
@@ -70,12 +72,12 @@ static vtr::NdMatrix<float, 2> chany_place_cost_fac({0, 0}); // [0...device_ctx.
  * bounding box is got from scratch, so the bounding box would definitely be
  * right, DO NOT update again.                                                   
  */
-struct NetInfo {
+struct PLNetCost {
     vtr::vector<ClusterNetId, double> net_cost;
     vtr::vector<ClusterNetId, double> proposed_net_cost;
-    vtr::vector<ClusterNetId, NetUpdateState> bb_update_status; // [0...cluster_ctx.clb_nlist.nets().size()-1]
-} NetInfo;
-static struct NetInfo net_info;
+    vtr::vector<ClusterNetId, NetUpdateState> bb_update_status;
+};
+static struct PLNetCost pl_net_cost;
 
 /* The following arrays are used by the try_swap function for speed.   */
 
@@ -161,7 +163,7 @@ static void record_affected_net(const ClusterNetId net, int& num_affected_nets);
  * @param num_affected_nets A pointer to the first free element of ts_nets_to_update. If a new net is added, the pointer should be increamented.
  * @param is_src_moving Is the moving pin the source of a net.
  */
-static void update_net_info_on_pin_move(const t_place_algorithm& place_algorithm,
+static void update_pl_net_cost_on_pin_move(const t_place_algorithm& place_algorithm,
                                                const PlaceDelayModel* delay_model,
                                                const PlacerCriticalities* criticalities,
                                                const ClusterBlockId& blk_id,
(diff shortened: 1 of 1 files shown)


vpr/src/place/net_cost_handler.cpp
```diff
@@ -55,14 +55,10 @@ static vtr::NdMatrix<float, 2> chanx_place_cost_fac({0, 0}); // [0...device_ctx.
 static vtr::NdMatrix<float, 2> chany_place_cost_fac({0, 0}); // [0...device_ctx.grid.height()-2]
 
 /**
- * @brief Cost of a net, and a temporary cost of a net used during move assessment. 
+ * @brief For the first two element in the struct: Cost of a net, and a temporary cost of a net used during move assessment.
```
vaughnbetz, line 58: You should use the member names (net_cost etc.) instead of the order in the struct, in case a member is added or deleted or reordered in the future. It also makes the comment more self-contained.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -55,14 +55,10 @@ static vtr::NdMatrix<float, 2> chanx_place_cost_fac({0, 0}); // [0...device_ctx.
...
- * @brief Cost of a net, and a temporary cost of a net used during move assessment. 
+ * @brief For the first two element in the struct: Cost of a net, and a temporary cost of a net used during move assessment. 
  * We also use negative cost values in proposed_net_cost as a flag to indicate that 
  * the cost of a net has not yet been updated.
- */
-static vtr::vector<ClusterNetId, double> net_cost, proposed_net_cost;
-
-/**                                              *
- * @brief Flag array to indicate whether the specific bounding box has been updated
+ * For the last one element in the struct: Flag array to indicate whether the specific bounding box has been updated
```
vaughnbetz, line 61: Change to "bb_update_status: flag array ..."

vpr/src/place/net_cost_handler.cpp
```diff
@@ -74,7 +70,12 @@ static vtr::vector<ClusterNetId, double> net_cost, proposed_net_cost;
  * bounding box is got from scratch, so the bounding box would definitely be
  * right, DO NOT update again.                                                   
  */
-static vtr::vector<ClusterNetId, NetUpdateState> bb_updated_before; // [0...cluster_ctx.clb_nlist.nets().size()-1]
+struct NetInfo {
```
vaughnbetz, line 73: I suggest a somewhat more descriptive name, e.g. PLNetCost.

vpr/src/place/net_cost_handler.cpp
vaughnbetz: Also comment that there is one entry per cluster level net [0...cluster_ctx.clb_nlist.nets().size()-1] for each of these vectors. (Mention this early in the comment).

vpr/src/place/net_cost_handler.cpp
```diff
@@ -74,7 +70,12 @@ static vtr::vector<ClusterNetId, double> net_cost, proposed_net_cost;
...
+    vtr::vector<ClusterNetId, double> net_cost;
+    vtr::vector<ClusterNetId, double> proposed_net_cost;
+    vtr::vector<ClusterNetId, NetUpdateState> bb_update_status; // [0...cluster_ctx.clb_nlist.nets().size()-1]
+} NetInfo;
```
vaughnbetz, line 77: Isn't this making a variable NetInfo as well as a struct called NetInfo?

vpr/src/place/net_cost_handler.cpp
```diff
@@ -74,7 +70,12 @@ static vtr::vector<ClusterNetId, double> net_cost, proposed_net_cost;
...
+static struct NetInfo net_info;
```
vaughnbetz, line 78: Can just write 
static NetInfo net_info;
(or better yet, a renamed 
static PLNetCost pl_net_cost;
)

vaughnbetz (commented):
Some changes requested for commenting and naming.
3% slowdown in place time: there are similar slowdowns in pack and route time, so this just looks like a machine load bounce. So overall the CPU time looks very similar.

robluo:
All comments addressed and pushed again.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -55,14 +55,12 @@ static vtr::NdMatrix<float, 2> chanx_place_cost_fac({0, 0}); // [0...device_ctx.
 static vtr::NdMatrix<float, 2> chany_place_cost_fac({0, 0}); // [0...device_ctx.grid.height()-2]
 
 /**
- * @brief Cost of a net, and a temporary cost of a net used during move assessment. 
+ * @brief For each of the vector in this struct, there is one entry per cluster level net:
```
vaughnbetz, line 58: vector -> vectors

vaughnbetz (approved):
Looks good. Some conflicts need to be resolved, and then CI should launch (not sure why only four tests launched so far but hopefully they all launch after the conflicts are updated).

AlexandreSinger:
> Looks good. Some conflicts need to be resolved, and then CI should launch (not sure why only four tests launched so far but hopefully they all launch after the conflicts are updated).

@vaughnbetz I think the reason the CI did not launch is because this PR is being merged into the `placement_move_primitive` branch, not master. If you want the CI to run, it would have to be manually dispatched from the action tab; however, @robluo was this target branch a mistake? Based on the commit history, this branch is in fact based off `placement_move_primitive` not master.

vaughnbetz:
Ah, that makes sense.  Yes, this should be merged into master.

robluo:
This work is based on `placement_move_primitive`, should that be merged first?

amin1377:
I think Robert is right. We first need to merge the placement refactoring PR before this one. Vaughn has reviewed that PR, and I have addressed his comments. The only remaining task is for me to run the QoR tests one more time. I have discussed this with @AlexandreSinger, and it will be done this week (before Wednesday).

@robluo I have a question regarding the QoR test. Is the baseline the VTR master branch or the placement refactoring PR? Thanks!

robluo:
@amin1377 The base line is `placement_move_primitive` in my QoR test.

PR #2653, opened 2024-07-18
Placement move primitive refactoring assignment 2

robluo:
In placement, we use two methods to update the bounding box: the 3D bounding box (also used for 2D architecture) and the per-layer bounding box. When a swap happens, [multiple variables](https://can01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fverilog-to-routing%2Fvtr-verilog-to-routing%2Fblob%2Fplacement_move_primitive%2Fvpr%2Fsrc%2Fplace%2Fnet_cost_handler.cpp%23L91-L98&data=05%7C02%7Cyulang.luo%40mail.utoronto.ca%7C5e26bbdc343c4119a05f08dc9fb5ddf1%7C78aac2262f034b4d9037b46d56c55210%7C0%7C0%7C638560850665920141%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=GjyyvhqPAvipLmtoMdc4Gr%2BPstuZ0wNKpgQZ8VOYN%2FY%3D&reserved=0) are updated, indicated by the `ts` (temporary swap) prefix. We can put all these variables into a class. Since only a subset is used depending on the bounding box type, you can examine how these variables are updated throughout the file and write methods for this class to update the appropriate ones.

After some discussion, it was decided to not use inheritance or the type eraser due to potential performance degradation and the lack of readability. In this PR, the control flow that decides 3d or 2d placement are wrapped into a class.

### DIFF
--- a/vpr/src/place/net_cost_handler.cpp
+++ b/vpr/src/place/net_cost_handler.cpp
@@ -55,10 +55,12 @@ static vtr::NdMatrix<float, 2> chanx_place_cost_fac({0, 0}); // [0...device_ctx.
 static vtr::NdMatrix<float, 2> chany_place_cost_fac({0, 0}); // [0...device_ctx.grid.height()-2]
 
 /**
- * @brief For the first two element in the struct: Cost of a net, and a temporary cost of a net used during move assessment. 
+ * @brief For each of the vector in this struct, there is one entry per cluster level net: 
+ * [0...cluster_ctx.clb_nlist.nets().size()-1].
+ * net_cost and proposed_net_cost: Cost of a net, and a temporary cost of a net used during move assessment. 
  * We also use negative cost values in proposed_net_cost as a flag to indicate that 
  * the cost of a net has not yet been updated.
- * For the last one element in the struct: Flag array to indicate whether the specific bounding box has been updated
+ * bb_update_status: Flag array to indicate whether the specific bounding box has been updated
  * in this particular swap or not. If it has been updated before, the code    
  * must use the updated data, instead of the out-of-date data passed into the 
  * subroutine, particularly used in try_swap(). The value NOT_UPDATED_YET     
@@ -70,12 +72,12 @@ static vtr::NdMatrix<float, 2> chany_place_cost_fac({0, 0}); // [0...device_ctx.
  * bounding box is got from scratch, so the bounding box would definitely be
  * right, DO NOT update again.                                                   
  */
-struct NetInfo {
+struct PLNetCost {
     vtr::vector<ClusterNetId, double> net_cost;
     vtr::vector<ClusterNetId, double> proposed_net_cost;
-    vtr::vector<ClusterNetId, NetUpdateState> bb_update_status; // [0...cluster_ctx.clb_nlist.nets().size()-1]
-} NetInfo;
-static struct NetInfo net_info;
+    vtr::vector<ClusterNetId, NetUpdateState> bb_update_status;
+};
+static struct PLNetCost pl_net_cost;
 
 /* The following arrays are used by the try_swap function for speed.   */
 
@@ -161,7 +163,7 @@ static void record_affected_net(const ClusterNetId net, int& num_affected_nets);
  * @param num_affected_nets A pointer to the first free element of ts_nets_to_update. If a new net is added, the pointer should be increamented.
  * @param is_src_moving Is the moving pin the source of a net.
  */
-static void update_net_info_on_pin_move(const t_place_algorithm& place_algorithm,
+static void update_pl_net_cost_on_pin_move(const t_place_algorithm& place_algorithm,
                                                const PlaceDelayModel* delay_model,
                                                const PlacerCriticalities* criticalities,
                                                const ClusterBlockId& blk_id,
@@ -478,7 +480,7 @@ static void update_net_bb(const ClusterNetId& net,
     if (cluster_ctx.clb_nlist.net_sinks(net).size() < SMALL_NET) {
         //For small nets brute-force bounding box update is faster
 
-        if (net_info.bb_update_status[net] == NetUpdateState::NOT_UPDATED_YET) { //Only once per-net
+        if (pl_net_cost.bb_update_status[net] == NetUpdateState::NOT_UPDATED_YET) { //Only once per-net
             get_non_updatable_bb(net,
                                  ts_bb_coord_new[net],
                                  ts_layer_sink_pin_count[size_t(net)]);
@@ -516,7 +518,7 @@ static void update_net_layer_bb(const ClusterNetId& net,
     if (cluster_ctx.clb_nlist.net_sinks(net).size() < SMALL_NET) {
         //For small nets brute-force bounding box update is faster
 
-        if (net_info.bb_update_status[net] == NetUpdateState::NOT_UPDATED_YET) { //Only once per-net
+        if (pl_net_cost.bb_update_status[net] == NetUpdateState::NOT_UPDATED_YET) { //Only once per-net
             get_non_updatable_layer_bb(net,
                                        layer_ts_bb_coord_new[net],
                                        ts_layer_sink_pin_count[size_t(net)]);
@@ -652,17 +654,17 @@ static void update_td_delta_costs(const PlaceDelayModel* delay_model,
 static void record_affected_net(const ClusterNetId net,
                                 int& num_affected_nets) {
     /* Record effected nets. */
-    if (net_info.proposed_net_cost[net] < 0.) {
+    if (pl_net_cost.proposed_net_cost[net] < 0.) {
         /* Net not marked yet. */
         ts_nets_to_update[num_affected_nets] = net;
         num_affected_nets++;
 
         /* Flag to say we've marked this net. */
-        net_info.proposed_net_cost[net] = 1.;
+        pl_net_cost.proposed_net_cost[net] = 1.;
     }
 }
 
-static void update_net_info_on_pin_move(const t_place_algorithm& place_algorithm,
+static void update_pl_net_cost_on_pin_move(const t_place_algorithm& place_algorithm,
                                                const PlaceDelayModel* delay_model,
                                                const PlacerCriticalities* criticalities,
                                                const ClusterBlockId& blk_id,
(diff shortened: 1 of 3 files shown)


vpr/src/place/net_cost_handler.cpp
```diff
@@ -88,14 +94,33 @@ static vtr::vector<ClusterNetId, NetUpdateState> bb_updated_before; // [0...clus
...
+    vtr::Matrix<int> ts_layer_sink_pin_count;
+    /* [0...num_afftected_nets] -> net_id of the affected nets */
+    std::vector<ClusterNetId> ts_nets_to_update;
+};
+static struct TSInfo ts_info;
+
+/**
+ * @brief This class is used to hide control flows needed to distinguish 2d and 3d placement
+ */
+class BB2D3DControlFlow {
```
AlexandreSinger, line 112: The classes and structs defined in this file are polluting the namespace. They either need to be put in the anonymous namespace or made static.

Although it may sound crazy, in the future (in a completely separate cpp file) someone may may a TSInfo struct (they just happen to use the same name) and they will get awful linker errors. Whenever a class/struct/function are declared in a cpp file they should be made static / put in the anonymous namespace to prevent this issue.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -88,14 +94,33 @@ static vtr::vector<ClusterNetId, NetUpdateState> bb_updated_before; // [0...clus
...
+public: 
+    void init(size_t num_nets, bool cube_bb);
+    void get_non_updatable_bb(const ClusterNetId& net);
+    bool is_driver(const t_physical_tile_type_ptr& blk_type, const ClusterPinId& blk_pin);
+    void update_bb(ClusterNetId net_id, t_physical_tile_loc pin_old_loc, t_physical_tile_loc pin_new_loc, bool is_driver); 
+    double get_net_cost(const ClusterNetId& net_id);
+    void set_ts_bb_coord(const ClusterNetId& net_id);
+    void set_ts_edge(const ClusterNetId& net_id);
+};
+static BB2D3DControlFlow bb_2d_3d_control_flow;
```
AlexandreSinger, line 123: I do not like the name of this class. I would use something like "BBManipulator" or something. Ideally you want to write code in a way that is extensible and self explanatory. BB2D3DControlFlow tells me "There is a problem with the control flow between 2D and 3D BB, and I made this class to encapsulate it" (which honestly is what we are doing), but I would recommend changing the name to something that says "This is a class that will manipulate the BB (whether it is 2D or 3D or any D in the future)".

vpr/src/place/net_cost_handler.cpp
```diff
@@ -88,14 +94,33 @@ static vtr::vector<ClusterNetId, NetUpdateState> bb_updated_before; // [0...clus
...
+    /* [0...num_afftected_nets] -> net_id of the affected nets */
+    std::vector<ClusterNetId> ts_nets_to_update;
+};
+static struct TSInfo ts_info;
+
+/**
+ * @brief This class is used to hide control flows needed to distinguish 2d and 3d placement
+ */
+class BB2D3DControlFlow {
+    bool cube_bb = false;
```
AlexandreSinger, line 113: Ew global state. I know why this is here, it prevents passing parameters around. @amin1377 this is your call. Do you ever foresee cube_bb changing during execution of the VPR flow? Or is it truly a static variable that never changes.

AlexandreSinger (commented):
Hi @robluo , I agree with this change over using inheritance. I left some comments to improve the PR.

vpr/src/place/net_cost_handler.cpp
amin1377: The bounding box type should be determined at the beginning of the flow, and I don't expect it to change during execution.

vpr/src/place/net_cost_handler.cpp
amin1377: Well put :)

vpr/src/place/net_cost_handler.cpp
amin1377: We can put all the functions in `net_cost_handler` into a separate namespace. We are also following the same practice for router lookahead utilities. If you think this is a standard practice, we can at least enforce it for the new code being added to VTR.

AlexandreSinger: I think it should be enforced for new code. and It does not have to be a separate namespace; it need only be static / put in the anonymous namespace. This is common practice for very very large code bases where polluting the global space can be a real problem.

vpr/src/place/net_cost_handler.cpp
AlexandreSinger: I see. Just wanted to check.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -88,14 +94,33 @@ static vtr::vector<ClusterNetId, NetUpdateState> bb_updated_before; // [0...clus
  * layer_ts_bb_coord_new are used.
```
vaughnbetz, line 94: Also mention: layer_sink_pin_count is only used in the layer bounding box case.
Add @brief at the start of this.

vpr/src/place/net_cost_handler.cpp
vaughnbetz: For file scope variables, static makes sense. For struct / class definitions: probably put in anonymous namespace if you want to hide them.

vpr/src/place/net_cost_handler.cpp
vaughnbetz: BBUpdater?  Or BBandCostUpdater

vpr/src/place/net_cost_handler.cpp
```diff
@@ -448,7 +453,88 @@ static double wirelength_crossing_count(size_t fanout);
...
+    if (cube_bb) {
+        ts_info.ts_bb_edge_new.resize(num_nets, t_bb());
+        ts_info.ts_bb_coord_new.resize(num_nets, t_bb());
+    } else {
+        VTR_ASSERT_SAFE(!cube_bb);
+        ts_info.layer_ts_bb_edge_new.resize(num_nets, std::vector<t_2D_bb>(num_layers, t_2D_bb()));
+        ts_info.layer_ts_bb_coord_new.resize(num_nets, std::vector<t_2D_bb>(num_layers, t_2D_bb()));
+    }
+
+    /*This initialize the whole matrix to OPEN which is an invalid value*/
```
vaughnbetz, line 470: typo: intializes

vpr/src/place/net_cost_handler.cpp
```diff
@@ -448,7 +453,88 @@ static double wirelength_crossing_count(size_t fanout);
 static void set_bb_delta_cost(const int num_affected_nets, double& bb_delta_c);
 
 /******************************* End of Function definitions ************************************/
+// Initialize the ts vectors
+void BB2D3DControlFlow::init(size_t num_nets, bool cube_bb_in) {
+    const int num_layers = g_vpr_ctx.device().grid.get_num_layers();
+
+    cube_bb = cube_bb_in;
```
vaughnbetz, line 460: Could comment: we'll use either the 3D BB or per layer BB data structures, but not both.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -448,7 +453,88 @@ static double wirelength_crossing_count(size_t fanout);
...
+        ts_info.ts_bb_edge_new.resize(num_nets, t_bb());
+        ts_info.ts_bb_coord_new.resize(num_nets, t_bb());
+    } else {
+        VTR_ASSERT_SAFE(!cube_bb);
+        ts_info.layer_ts_bb_edge_new.resize(num_nets, std::vector<t_2D_bb>(num_layers, t_2D_bb()));
+        ts_info.layer_ts_bb_coord_new.resize(num_nets, std::vector<t_2D_bb>(num_layers, t_2D_bb()));
+    }
+
+    /*This initialize the whole matrix to OPEN which is an invalid value*/
+    ts_info.ts_layer_sink_pin_count.resize({num_nets, size_t(num_layers)}, OPEN);
```
vaughnbetz, line 471: I think this one is only used for per layer BB, so it should be moved inside the if statement.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -448,7 +453,88 @@ static double wirelength_crossing_count(size_t fanout);
...
+// Initialize the ts vectors
+void BB2D3DControlFlow::init(size_t num_nets, bool cube_bb_in) {
+    const int num_layers = g_vpr_ctx.device().grid.get_num_layers();
+
+    cube_bb = cube_bb_in;
+    if (cube_bb) {
+        ts_info.ts_bb_edge_new.resize(num_nets, t_bb());
+        ts_info.ts_bb_coord_new.resize(num_nets, t_bb());
+    } else {
+        VTR_ASSERT_SAFE(!cube_bb);
```
vaughnbetz, line 465: I don't have a strong opinion, but usually I don't write asserts that check the else is the opposite of the if condition.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -448,7 +453,88 @@ static double wirelength_crossing_count(size_t fanout);
...
+        ::get_non_updatable_layer_bb(net,
+                                   ts_info.layer_ts_bb_coord_new[net],
+                                   ts_info.ts_layer_sink_pin_count[size_t(net)]);
+}
+
+bool BB2D3DControlFlow::is_driver(const t_physical_tile_type_ptr& blk_type, const ClusterPinId& blk_pin) {
+    auto& cluster_ctx = g_vpr_ctx.clustering();
+    if(cube_bb)
+        return cluster_ctx.clb_nlist.pin_type(blk_pin) == PinType::DRIVER;
+    else
```
vaughnbetz, line 491: I think we can do the if part all the time and it is simpler.  If so, we can just get rid of the routine.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -477,71 +563,27 @@ static void update_net_bb(const ClusterNetId& net,
...
         }
     } else {
         //For large nets, update bounding box incrementally
         int iblk_pin = tile_pin_index(blk_pin);
-        bool src_pin = cluster_ctx.clb_nlist.pin_type(blk_pin) == PinType::DRIVER;
 
         t_physical_tile_type_ptr blk_type = physical_tile_type(blk);
         int pin_width_offset = blk_type->pin_width_offset[iblk_pin];
         int pin_height_offset = blk_type->pin_height_offset[iblk_pin];
+        bool is_driver = bb_2d_3d_control_flow.is_driver(blk_type, blk_pin);
```
vaughnbetz, line 576: Could leave the function if you like it, but I think only the if part is needed (no else).

vpr/src/place/net_cost_handler.h
```diff
@@ -83,10 +83,8 @@ double comp_layer_bb_cost(e_cost_methods method);
 /**
  * @brief update net cost data structures (in placer context and net_cost in .cpp file) and reset flags (proposed_net_cost and bb_updated_before).
  * @param num_nets_affected The number of nets affected by the move. It is used to determine the index up to which elements in ts_nets_to_update are valid.
- * @param cube_bb True if we should use the 3D bounding box (cube_bb), false otherwise.
  */
-void update_move_nets(int num_nets_affected,
-                      const bool cube_bb);
+void update_move_nets(int num_nets_affected);
```
vaughnbetz, line 87: Do we need num_nets_affected?  Can be deleted.

vaughnbetz (changes requested):
Nice work!

PR #2659, opened 2024-07-22
[vpr][place] Placement move primitive refactoring

robluo:
1. In "[net_cost_handler.cpp](https://can01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fverilog-to-routing%2Fvtr-verilog-to-routing%2Fblob%2Fplacement_move_primitive%2Fvpr%2Fsrc%2Fplace%2Fnet_cost_handler.cpp&data=05%7C02%7Cyulang.luo%40mail.utoronto.ca%7C5e26bbdc343c4119a05f08dc9fb5ddf1%7C78aac2262f034b4d9037b46d56c55210%7C0%7C0%7C638560850665905503%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=58ozr4ygx6cDyqFr5NL0%2BYAXmebPJsq5xthyQtC28iA%3D&reserved=0)", there are [three](https://can01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fverilog-to-routing%2Fvtr-verilog-to-routing%2Fblob%2Fplacement_move_primitive%2Fvpr%2Fsrc%2Fplace%2Fnet_cost_handler.cpp%23L62-L77&data=05%7C02%7Cyulang.luo%40mail.utoronto.ca%7C5e26bbdc343c4119a05f08dc9fb5ddf1%7C78aac2262f034b4d9037b46d56c55210%7C0%7C0%7C638560850665913899%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=n1SBtPnW2y4RgGmenIt6RbSk71qlkJRqcGUOXwUBdMA%3D&reserved=0) related global variables: `net_cost`, `proposed_net_cost`, and `bb_updated_before`. We can organize these by putting them into a struct called `t_net_cost_data`. You may also rename `bb_updated_before` to `bb_update_status`.

2. In placement, we use two methods to update the bounding box: the 3D bounding box (also used for 2D architecture) and the per-layer bounding box. When a swap happens, [multiple variables](https://can01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fverilog-to-routing%2Fvtr-verilog-to-routing%2Fblob%2Fplacement_move_primitive%2Fvpr%2Fsrc%2Fplace%2Fnet_cost_handler.cpp%23L91-L98&data=05%7C02%7Cyulang.luo%40mail.utoronto.ca%7C5e26bbdc343c4119a05f08dc9fb5ddf1%7C78aac2262f034b4d9037b46d56c55210%7C0%7C0%7C638560850665920141%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=GjyyvhqPAvipLmtoMdc4Gr%2BPstuZ0wNKpgQZ8VOYN%2FY%3D&reserved=0) are updated, indicated by the `ts` (temporary swap) prefix. We can put all these variables into a class. Since only a subset is used depending on the bounding box type, you can examine how these variables are updated throughout the file and write methods for this class to update the appropriate ones.

3. For runtime reasons, we initialize `ts_nets_to_update` to the number of nets in the netlist and use a variable `num_affected_nets` to indicate how many entries are valid. You can experiment to see how runtime is affected if we remove this variable and just reserve elements in this vector and use `.size()`.

4. The same approach is used for `std::vector<t_pl_moved_block>& moved_blocks`, which has as many entries as the total number of blocks, and we pass an integer to show how many entries are valid. You can also run an experiment to see how much runtime is affected if you remove this.

QoR  bitcoin_miner_stratixiv_arch_timing (final wire length remains the same):
baseline: packing: 354.64, placement: 3706.12, routing 409.20, vpr flow: 4930.11
modified: packing: 356.83, placement: 3777.65, routing 412.23, vpr flow: 4994.61

### DIFF
--- a/vpr/src/place/move_transactions.cpp
+++ b/vpr/src/place/move_transactions.cpp
@@ -1,8 +1,6 @@
 #include "move_utils.h"
 
 #include "globals.h"
-// Is this needed?
-#include "place_util.h"
 
 t_pl_blocks_to_be_moved::t_pl_blocks_to_be_moved(size_t max_blocks){
         moved_blocks.reserve(max_blocks);
--- a/vpr/src/place/move_transactions.cpp
+++ b/vpr/src/place/move_transactions.cpp
@@ -1,6 +1,7 @@
 #include "move_utils.h"
 
 #include "globals.h"
+#include "place_util.h"
 #include "vtr_assert.h"
 
 t_pl_blocks_to_be_moved::t_pl_blocks_to_be_moved(size_t max_blocks){
--- a/vpr/src/place/move_transactions.cpp
+++ b/vpr/src/place/move_transactions.cpp
@@ -1,6 +1,7 @@
 #include "move_utils.h"
 
 #include "globals.h"
+#include "vtr_assert.h"
 
 t_pl_blocks_to_be_moved::t_pl_blocks_to_be_moved(size_t max_blocks){
         moved_blocks.reserve(max_blocks);
@@ -8,7 +9,7 @@ t_pl_blocks_to_be_moved::t_pl_blocks_to_be_moved(size_t max_blocks){
 }
 
 size_t t_pl_blocks_to_be_moved::get_size_and_increment() {
-    VTR_ASSERT(moved_blocks.size() < moved_blocks.capacity());
+    VTR_ASSERT_SAFE(moved_blocks.size() < moved_blocks.capacity());
     moved_blocks.resize(moved_blocks.size() + 1);
     return moved_blocks.size() - 1;
 }
@@ -49,11 +50,11 @@ void apply_move_blocks(const t_pl_blocks_to_be_moved& blocks_affected) {
 
     //Swap the blocks, but don't swap the nets or update place_ctx.grid_blocks
     //yet since we don't know whether the swap will be accepted
-    for (size_t iblk = 0; iblk < blocks_affected.moved_blocks.size(); ++iblk) {
-        ClusterBlockId blk = blocks_affected.moved_blocks[iblk].block_num;
+    for (const auto& block : blocks_affected.moved_blocks) {
+        ClusterBlockId blk = block.block_num;
 
-        const t_pl_loc& old_loc = blocks_affected.moved_blocks[iblk].old_loc;
-        const t_pl_loc& new_loc = blocks_affected.moved_blocks[iblk].new_loc;
+        const t_pl_loc& old_loc = block.old_loc;
+        const t_pl_loc& new_loc = block.new_loc;
 
         // move the block to its new location
         place_ctx.block_locs[blk].loc = new_loc;
@@ -76,11 +77,11 @@ void commit_move_blocks(const t_pl_blocks_to_be_moved& blocks_affected) {
     auto& place_ctx = g_vpr_ctx.mutable_placement();
 
     /* Swap physical location */
-    for (size_t iblk = 0; iblk < blocks_affected.moved_blocks.size(); ++iblk) {
-        ClusterBlockId blk = blocks_affected.moved_blocks[iblk].block_num;
+    for (const auto& block : blocks_affected.moved_blocks) {
+        ClusterBlockId blk = block.block_num;
 
-        const t_pl_loc& to = blocks_affected.moved_blocks[iblk].new_loc;
-        const t_pl_loc& from = blocks_affected.moved_blocks[iblk].old_loc;
+        const t_pl_loc& to = block.new_loc;
+        const t_pl_loc& from = block.old_loc;
 
         //Remove from old location only if it hasn't already been updated by a previous block update
         if (place_ctx.grid_blocks.block_at_location(from) == blk) {
@@ -106,11 +107,11 @@ void revert_move_blocks(const t_pl_blocks_to_be_moved& blocks_affected) {
     auto& device_ctx = g_vpr_ctx.device();
 
     // Swap the blocks back, nets not yet swapped they don't need to be changed
-    for (size_t iblk = 0; iblk < blocks_affected.moved_blocks.size(); ++iblk) {
-        ClusterBlockId blk = blocks_affected.moved_blocks[iblk].block_num;
+    for (const auto& block : blocks_affected.moved_blocks) {
+        ClusterBlockId blk = block.block_num;
 
-        const t_pl_loc& old_loc = blocks_affected.moved_blocks[iblk].old_loc;
-        const t_pl_loc& new_loc = blocks_affected.moved_blocks[iblk].new_loc;
+        const t_pl_loc& old_loc = block.old_loc;
+        const t_pl_loc& new_loc = block.new_loc;
 
         // return the block to where it was before the swap
         place_ctx.block_locs[blk].loc = old_loc;
--- a/vpr/src/place/move_transactions.cpp
+++ b/vpr/src/place/move_transactions.cpp
@@ -1,3 +1,4 @@
+#include "move_transactions.h"
 #include "move_utils.h"
 
 #include "globals.h"
@@ -6,7 +7,6 @@
 
 t_pl_blocks_to_be_moved::t_pl_blocks_to_be_moved(size_t max_blocks){
         moved_blocks.reserve(max_blocks);
-        moved_blocks.resize(0);
 }
 
 size_t t_pl_blocks_to_be_moved::get_size_and_increment() {
@@ -16,9 +16,9 @@ size_t t_pl_blocks_to_be_moved::get_size_and_increment() {
 }
 
 //Records that block 'blk' should be moved to the specified 'to' location
-e_block_move_result record_block_move(t_pl_blocks_to_be_moved& blocks_affected, ClusterBlockId blk, t_pl_loc to) {
-    auto res = blocks_affected.moved_to.emplace(to);
-    if (!res.second) {
+e_block_move_result t_pl_blocks_to_be_moved::record_block_move(ClusterBlockId blk, t_pl_loc to) {
+    auto [to_it, to_success] = moved_to.emplace(to);
+    if (!to_success) {
         log_move_abort("duplicate block move to location");
         return e_block_move_result::ABORT;
     }
@@ -27,8 +27,9 @@ e_block_move_result record_block_move(t_pl_blocks_to_be_moved& blocks_affected, 
 
     t_pl_loc from = place_ctx.block_locs[blk].loc;
 
-    auto res2 = blocks_affected.moved_from.emplace(from);
-    if (!res2.second) {
+    auto [_, from_success] = moved_from.emplace(from);
+    if (!from_success) {
+        moved_to.erase(to_it);
         log_move_abort("duplicate block move from location");
         return e_block_move_result::ABORT;
     }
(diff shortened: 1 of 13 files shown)


vpr/src/place/move_transactions.h
```diff
@@ -25,24 +25,21 @@ struct t_pl_moved_block {
...
-        : moved_blocks(max_blocks) {}
+    explicit t_pl_blocks_to_be_moved(size_t max_blocks);
 
-    int num_moved_blocks = 0;
     std::vector<t_pl_moved_block> moved_blocks;
     std::unordered_set<t_pl_loc> moved_from;
     std::unordered_set<t_pl_loc> moved_to;
 
     std::vector<ClusterPinId> affected_pins;
+    size_t get_size_and_increment();
```
soheilshahrouz, line 42: please add doxygen documentation for this method.

vpr/src/place/move_transactions.h
```diff
@@ -25,24 +25,21 @@ struct t_pl_moved_block {
...
  * moved blocks: a list of moved blocks data structure with     *
  *               information on the move.                       *
  *               [0...max_blocks-1]                       *
  * affected_pins: pins affected by this move (used to           *
  *                incrementally invalidate parts of the timing  *
  *                graph.                                        */
 struct t_pl_blocks_to_be_moved {
-    explicit t_pl_blocks_to_be_moved(size_t max_blocks)
-        : moved_blocks(max_blocks) {}
+    explicit t_pl_blocks_to_be_moved(size_t max_blocks);
```
soheilshahrouz, line 35: it is probably a good idea to delete the default constructor if the objects of this type are always expected to be initialzed with a non-zero capacity

vpr/src/place/move_transactions.cpp
```diff
@@ -2,6 +2,18 @@
 
 #include "globals.h"
 #include "place_util.h"
+#include "vtr_assert.h"
+
+t_pl_blocks_to_be_moved::t_pl_blocks_to_be_moved(size_t max_blocks){
+        moved_blocks.reserve(max_blocks);
+        moved_blocks.resize(0);
```
soheilshahrouz, line 9: is calling resize() necessary?

vpr/src/place/move_transactions.cpp
```diff
@@ -2,6 +2,18 @@
...
+}
+
+size_t t_pl_blocks_to_be_moved::get_size_and_increment() {
+    VTR_ASSERT_SAFE(moved_blocks.size() < moved_blocks.capacity());
+    moved_blocks.resize(moved_blocks.size() + 1);
+    return moved_blocks.size() - 1;
+}
 
 //Records that block 'blk' should be moved to the specified 'to' location
 e_block_move_result record_block_move(t_pl_blocks_to_be_moved& blocks_affected, ClusterBlockId blk, t_pl_loc to) {
```
soheilshahrouz, line 19: this function does not access or modify the global state and only changes the content of blocks_affected. It would be more readable if this functon is implemeneted as a method under t_pl_blocks_to_be_moved.

vpr/src/place/move_transactions.cpp
```diff
@@ -40,11 +51,11 @@ void apply_move_blocks(const t_pl_blocks_to_be_moved& blocks_affected) {
 
     //Swap the blocks, but don't swap the nets or update place_ctx.grid_blocks
     //yet since we don't know whether the swap will be accepted
-    for (int iblk = 0; iblk < blocks_affected.num_moved_blocks; ++iblk) {
-        ClusterBlockId blk = blocks_affected.moved_blocks[iblk].block_num;
+    for (const auto& block : blocks_affected.moved_blocks) {
```
soheilshahrouz, line 54: I think if the element type (t_pl_moved_block) is explicitly specified instead of using auto keyword, the code would be more readable, especially that the type name is not that long.

vpr/src/place/move_transactions.cpp
```diff
@@ -126,10 +137,10 @@ void clear_move_blocks(t_pl_blocks_to_be_moved& blocks_affected) {
     blocks_affected.moved_to.clear();
     blocks_affected.moved_from.clear();
 
-    //For run-time, we just reset num_moved_blocks to zero, but do not free the blocks_affected
+    //For run-time, we just reset size of blocks_affected.moved_blocks to zero, but do not free the blocks_affected
     //array to avoid memory allocation
 
-    blocks_affected.num_moved_blocks = 0;
+    blocks_affected.moved_blocks.resize(0);
```
soheilshahrouz, line 143: clear_move_blocks() can also be implemented as a method of t_pl_blocks_to_be_moved.

soheilshahrouz: if clear_move_blocks() and clear_move_blocks() are implemented as member functions of t_pl_blocks_to_be_moved, we can make all its member variable private.

vpr/src/place/move_transactions.h
soheilshahrouz: Also delete the copy constructor to avoid unintended copies.

vpr/src/place/move_transactions.cpp
soheilshahrouz: res.second and res2.second are not clear enough in this code. Now that you have touched this function, can you please make them more readable by using the following syntax?

```
auto [it, success] = blocks_affected.moved_to.emplace(to);

if (!success) {
}
```

vpr/src/place/move_utils.cpp
```diff
@@ -488,12 +488,12 @@ std::set<t_pl_loc> determine_locations_emptied_by_move(t_pl_blocks_to_be_moved&
     std::set<t_pl_loc> moved_from;
     std::set<t_pl_loc> moved_to;
 
-    for (int iblk = 0; iblk < blocks_affected.num_moved_blocks; ++iblk) {
+    for (const auto& block : blocks_affected.moved_blocks) {
```
soheilshahrouz, line 491: using the type name explicitly makes the code more readable

vpr/src/place/move_transactions.cpp
soheilshahrouz: I guess there is a subtle bug in this function. If the move is aborted because `res.second `is false, we need to remove `to` from `blocks_affected.moved_to`.

vpr/src/place/move_utils.cpp
```diff
@@ -488,12 +488,12 @@ std::set<t_pl_loc> determine_locations_emptied_by_move(t_pl_blocks_to_be_moved&
     std::set<t_pl_loc> moved_from;
```
soheilshahrouz, line 488: blocks_affected can be passed by const reference

soheilshahrouz: this function can be implemented as member function in t_pl_blocks_to_be_moved

vpr/src/place/place_constraints.h
```diff
@@ -100,16 +100,16 @@ void print_macro_constraint_error(const t_pl_macro& pl_macro);
 inline bool floorplan_legal(const t_pl_blocks_to_be_moved& blocks_affected) {
     bool floorplan_legal;
 
-    for (int i = 0; i < blocks_affected.num_moved_blocks; i++) {
-        floorplan_legal = cluster_floorplanning_legal(blocks_affected.moved_blocks[i].block_num,
-                                                      blocks_affected.moved_blocks[i].new_loc);
+    for (const auto& block : blocks_affected.moved_blocks) {
```
soheilshahrouz, line 103: explicit type name make the code more readable

vpr/src/place/move_transactions.cpp
soheilshahrouz: I also suggest that the element name changes from "block" to "moved_block"

vpr/src/place/place.cpp
```diff
@@ -1378,7 +1378,7 @@ static e_move_result try_swap(const t_annealing_state* state,
         //
         //Also find all the pins affected by the swap, and calculates new connection
         //delays and timing costs and store them in proposed_* data structures.
-        int num_nets_affected = find_affected_nets_and_update_costs(
+        find_affected_nets_and_update_costs(
```
soheilshahrouz, line 1381: I think if this function call is written in two lines, it would be more readable:

```
find_affected_nets_and_update_costs(place_algorithm, delay_model, criticalities,
                                    blocks_affected, bb_delta_c, timing_delta_c);
```

vpr/src/place/place.cpp
```diff
@@ -1569,7 +1568,7 @@ static e_move_result try_swap(const t_annealing_state* state,
         calculate_reward_and_process_outcome(placer_opts, move_outcome_stats,
                                              delta_c, timing_bb_factor, move_generator);
     } else {
-//        std::cout << "Group move delta cost: " << delta_c << std::endl;
+        //        std::cout << "Group move delta cost: " << delta_c << std::endl;
```
soheilshahrouz, line 1571: remove the else statement
I used it for debugging, but it somehow found its way into master

soheilshahrouz (commented):
I added some comments for all files other than net_cost_handler.cpp/.h

vpr/src/place/move_transactions.cpp
robluo: No, it is not necessary, the vector default constructor will set size to 0.

vpr/src/place/move_transactions.cpp
robluo: > I guess there is a subtle bug in this function. If the move is aborted because `res.second `is false, we need to remove `to` from `blocks_affected.moved_to`.

Should this apply to res2?

soheilshahrouz: > > I guess there is a subtle bug in this function. If the move is aborted because `res.second `is false, we need to remove `to` from `blocks_affected.moved_to`.
> 
> Should this apply to res2?

Yes. When we abort the move because `res.second is false`, `to` has already been inserted into `locks_affected.moved_to`.

vpr/src/place/move_transactions.cpp
soheilshahrouz: > No, it is not necessary, the vector default constructor will set size to 0.

So I think it is better to remove it

vpr/src/place/move_transactions.cpp
robluo: > > > I guess there is a subtle bug in this function. If the move is aborted because `res.second `is false, we need to remove `to` from `blocks_affected.moved_to`.
> > 
> > 
> > Should this apply to res2?
> 
> Yes. When we abort the move because `res.second is false`, `to` has already been inserted into `locks_affected.moved_to`.

Would this also apply to `res2` and `from`?

soheilshahrouz: > I guess there is a subtle bug in this function. If the move is aborted because `res.second `is false, we need to remove `to` from `blocks_affected.moved_to`.

I made a mistake here. I meant `res2.second`.
If `res.second` is false, `to` has not been inserted to the set.

vaughnbetz:
Thanks!  I'm going to merge this, but I have one more comment you can address in another PR:

(line 32 of move_transactions.cpp)
It would be good to comment that we need to remove the block from moved_to, and why.

Issue #2641, opened 2024-07-05
Passing only Fc max to pin to track map function


saaramahmoudi:
#### Current Behaviour
While we want to create connection blocks (pin->track and track->pin connections), we first loop over segments and calculate the maximum Fc over all pins of one block and this segment type (rr_graph.cpp, line 3253):

```
/* determine the maximum Fc to this segment type across all pins */
  int max_Fc = 0;
  for (int pin_index = 0; pin_index < Type->num_pins; ++pin_index) {
      int pin_class = Type->pin_class[pin_index];
      if (Fc[pin_index][seg_inf[iseg].seg_index] > max_Fc && Type->class_inf[pin_class].type == pin_type) {
          max_Fc = Fc[pin_index][seg_inf[iseg].seg_index];
      }
  }
```  
Afterward, we use max_Fc calculated to pass through the pin_to_seg_type_map to create connections based on max_Fc:
```
/* get pin connections to tracks of the current segment type */
auto pin_to_seg_type_map = alloc_and_load_pin_to_seg_type(pin_type, num_seg_type_tracks, max_Fc, Type, type_layer, perturb_switch_pattern[seg_inf[iseg].seg_index], directionality);

```
Although we use the maximum Fc value, we can override some pins (such as the carry chain) with Fc 0. However, assuming one pin has 4 connections to this segment and the other has 6 connections to this segment, we connect both pins 6 times since the max_Fc would be 6. Here is the comment for the mentioned function (rr_graph.cpp, line 3298):

```
/* Note: currently a single value of Fc is used across each pin. In the future
     * the looping below will have to be modified if we want to account for pin-based
     * Fc values 
*/
```

@vaughnbetz If you agree that it is better to account for pin-based Fc values, I can implement and keep this issue to track the progress. If there is a specific reason for this implementation, I can add more comments to the function to avoid confusion.

vaughnbetz:
Thanks for the detailed analysis Sara.  I agree it would be better to change this; I can't really see what the reason for it is.

WhiteNinjaZ:
@saaramahmoudi I have actually been working on this issue for a while now and didn't see that there was already someone working on it till recently. I have applied a temporary fix that seems to work well. 
Here is what I did: alloc_and_load_pin_to_seg_type() should only be called once per segment type since it keeps track of which pins connect to which segments. What I ended up doing was just making a very simple change to the nested for loop as follows:
```
[...]

PR #2661, opened 2024-07-23
Pass actual fc instead of fc_max to connect to pins

saaramahmoudi:
#### Description
This PR addresses the issue mentioned in [Issue #2641](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2641). 

I tested the branch with my examples with fc_override with different pins. @WhiteNinjaZ has also tested the branch on a couple more benchmarks and confirmed that it works properly. 

#### Types of changes

#### QoR: 
We expect the RR graph to change based on this change. Thus, I have initiated Titan and VTR benchmarks to compare QoR with the master branch. Will attach the results as soon as it is ready.


saaramahmoudi:
@vaughnbetz 

These links are the results for this branch for VTR and Titan benchmarks:

- [VTR benchmark](https://docs.google.com/spreadsheets/d/1w8RWgOSlby4v6JSrNrT5hb8pZl61-FiUtqL9arLPbdE/edit?usp=sharing)
- [Titan benchmark](https://docs.google.com/spreadsheets/d/1x1TXc-ahFkMdJlPq_t-kzOi_lmip_0XB2sNYf_P2EW0/edit?usp=sharing)

vaughnbetz:
A few QoR failures; they seem to be higher routing areas, maybe expected due to the Fc handling change.  I checked vtr_reg_basic and it is routing area increases (37%, above the 30% tolerance) but only on the diffeq circuit (which is small). That seems likely to be OK and just needing a golden result update; other failures are hopefully similar.

vpr/src/route/rr_graph.cpp
```diff
@@ -106,8 +106,10 @@ static vtr::NdMatrix<std::vector<int>, 5> alloc_and_load_pin_to_track_map(const
                                                                           const int* sets_per_seg_type);
 
 static vtr::NdMatrix<int, 6> alloc_and_load_pin_to_seg_type(const e_pin_type pin_type,
```
vaughnbetz, line 108: It would be good to add a doxygen comment saying what this routine does and describing the parameters.  If you don't know what all the parameters do, at least describing the new ones will be a step in the right direction.

Commenting what the Fc matrix stores: how many dimensions, what is in each dimension, and this is an absolute Fc value (I think).  Comment wht max_Fc is too (and the other parameters if you can).

vpr/src/route/rr_graph.cpp
```diff
@@ -284,16 +286,18 @@ static void alloc_and_load_tile_rr_graph(RRGraphBuilder& rr_graph_builder,
 static float pattern_fmod(float a, float b);
 static void load_uniform_connection_block_pattern(vtr::NdMatrix<int, 6>& tracks_connected_to_pin,
```
vaughnbetz, line 287: Add Doxygen comment of purpose and parameters (as much info as you know at least).

vpr/src/route/rr_graph.cpp
```diff
@@ -284,16 +286,18 @@ static void alloc_and_load_tile_rr_graph(RRGraphBuilder& rr_graph_builder,
...
                                                   const std::vector<t_pin_loc>& pin_locations,
+                                                  const vtr::Matrix<int>& Fc,
+                                                  const int seg_index,
                                                   const int x_chan_width,
                                                   const int y_chan_width,
-                                                  const int Fc,
                                                   const enum e_directionality directionality);
 
 static void load_perturbed_connection_block_pattern(vtr::NdMatrix<int, 6>& tracks_connected_to_pin,
                                                     const std::vector<t_pin_loc>& pin_locations,
```
vaughnbetz, line 296: Add Doxygen comment of purpose and parameters (as much info as you know at least).

vaughnbetz (commented):
Looks good; I just suggest some Doxygen comments for the modified routines (which didn't have them before, but adding them would make you a force for good @saaramahmoudi !).

vaughnbetz:
QoR looks good; roughly a tie for VTR and actually a bit better for Titan.
Can you also compare the routing area per tile on Titan (can just do one circuit) to confirm it is very close before and after the change? That will confirm we're getting the same number of switches when you don't change the architecture.

saaramahmoudi:
> QoR looks good; roughly a tie for VTR and actually a bit better for Titan. Can you also compare the routing area per tile on Titan (can just do one circuit) to confirm it is very close before and after the change? That will confirm we're getting the same number of switches when you don't change the architecture.

The spreadsheet I attached for Titan has the area per tile for all circuits (parse_results_master.txt and parse_results.txt tabs). Geomean averaged over all circuits; the routing area per tile for the master branch is "18519," and for my branch, it is "18518", which confirms that we are getting a close number in both branches.

vaughnbetz:
Thanks, that looks good.

vaughnbetz:
If you're able to add the requested comments we can merge. If you can't add those in the near future we could merge too (although it would be nice to get them in while we're looking at the code).

saaramahmoudi:
> If you're able to add the requested comments we can merge. If you can't add those in the near future we could merge too (although it would be nice to get them in while we're looking at the code).

I am adding the comments right now. After I push to the branch with updated golden results, I will ping you.

vpr/src/route/rr_graph.cpp
```diff
@@ -104,7 +121,24 @@ static vtr::NdMatrix<std::vector<int>, 5> alloc_and_load_pin_to_track_map(const
                                                                           const e_directionality directionality,
                                                                           const std::vector<t_segment_inf>& seg_inf,
                                                                           const int* sets_per_seg_type);
-
+/**
+ * @brief This routine calculate pin connections to tracks for a specific type and a specific segment based on the Fc value
```
vaughnbetz, line 125: calculate -> calculates

vpr/src/route/rr_graph.cpp
```diff
@@ -104,7 +121,24 @@ static vtr::NdMatrix<std::vector<int>, 5> alloc_and_load_pin_to_track_map(const
...
                                                                           const int* sets_per_seg_type);
-
+/**
+ * @brief This routine calculate pin connections to tracks for a specific type and a specific segment based on the Fc value 
+ * defined for each pin in the architecture file. This routine is called twice for each combination of block type and segment 
+ * type: 1) connecting tracks to input pins 2) connecting output pins to tracks.  
+ * 
+ *   @param pin_type Specifies whether the routine should connect tracks to *INPUT* pins or connect *OUTPUT* pins to tracks.
+ *   @param Fc Actual Fc value described in the architecture file for all pins of the specific phyiscal type ([0..number_of_pins-1][0..number_of_segments-1]).
+ *   @param seg_type_tracks Number of tracks that is avaliable for the specific segment.
```
vaughnbetz, line 131: segment -> segment type
(I think)

vpr/src/route/rr_graph.cpp
```diff
@@ -104,7 +121,24 @@ static vtr::NdMatrix<std::vector<int>, 5> alloc_and_load_pin_to_track_map(const
...
+ *   @param seg_index The segment index that the function is trying to connect to pins.
+ *   @param max_Fc Used to allocate max possible space for simplicity.
+ *   @param Type Physical type information, such as total number of pins, block width, block height, and etc.
```
vaughnbetz, line 134: Rename Type to tile_type

vpr/src/route/rr_graph.cpp
```diff
@@ -94,8 +94,25 @@ std::set<int> get_layers_pin_is_connected_to(const t_physical_tile_type_ptr type
 ///@brief given a specific layer number and type, it returns layers which have same pin_index connected to the given layer
 std::set<int> get_layers_connected_to_pin(const t_physical_tile_type_ptr type, int to_layer, int pin_index);
 
+///@brief checks whether the channel width has been changed or not
 bool channel_widths_unchanged(const t_chan_width& current, const t_chan_width& proposed);
 
+/**
+ * @brief This routine calculate pin connections to tracks for a specific type on the Fc value defined for each pin in the architecture file.
```
vaughnbetz, line 101: What is the difference between this routine and the one below it?  From their comments they both seem to do the same thing. One returns a 5D matrix of vectors, and the other a 6D matrix, but other than that they don't seem different.  Could one be deleted (don't worry about it if that is too much work, but we could put a TODO in if that's the better long term path).

vpr/src/route/rr_graph.cpp
```diff
@@ -305,6 +353,18 @@ static void load_uniform_connection_block_pattern(vtr::NdMatrix<int, 6>& tracks_
...
                                                   const enum e_directionality directionality);
 
+/**
+ * @brief Loads the tracks_connected_to_pin array with an unevenly distributed set of switches across the channel.
+ * 
+ *   @param tracks_connected_to_pin The funtion loads up this data structure with a track index for each pin ([0..num_pins-1][0..width-1][0..height-1][0..layer-1][0..sides-1][0..Fc-1]]).                        
+ *   @param pin_locations Physical pin informations, such as pin_index in the physical type, side, and etc.
+ *   @param x_chan_width Number of tracks in x-axis.
+ *   @param y_chan_width Number of tracks in y-axis.
+ *   @param Fc Actual Fc value described in the architecture file for all pins of the specific phyiscal type ([0..number_of_pins-1][0..number_of_segments-1]).
```
vaughnbetz, line 363: number_of_segments --> should this be number_of_segment_types?

vpr/src/route/rr_graph.cpp
```diff
@@ -305,6 +353,18 @@ static void load_uniform_connection_block_pattern(vtr::NdMatrix<int, 6>& tracks_
...
+ *   @param seg_index The segment index that the function is trying to connect to pins.
```
vaughnbetz, line 364: seg_index --> this comment doesn't really tell me what this is.  Is it the index of the segment type to which we want to connect pins?

vaughnbetz (changes requested):
Some commenting improvements and variable name improvements suggested.

vpr/src/route/rr_graph.cpp
saaramahmoudi: The one that returns the 6D matrix is written as a helper function for the other one. Basically, we first calculate the number of connections should be made for each pin for the current tile and save a relative track indices. Then, we return the 6D matrix and post process it to calculate the absolute track number and keep a vector of connection per each pin. Note that basically they return the same thing (6D matrix of int and 5D matrix of integer vector) but the logic within each function are different.

saaramahmoudi:
@vaughnbetz applied requested changes, ready to merge!

AlexandreSinger:
@vaughnbetz This PR failed the strong tests. After merging, master is now failing the strong tests...

AlexandreSinger:
@saaramahmoudi Do you think this is a problem? Or do you think the golden solutions just need to be regenerated?

saaramahmoudi:
@AlexandreSinger @vaughnbetz I think we just need to generate the golden results for those.

AlexandreSinger:
@saaramahmoudi Perfect! Thats good news. Would you be able to do that in a PR? If not I can raise a PR tomorrow if I find time.

saaramahmoudi:
@AlexandreSinger will raise a PR tomorrow!

Issue #2652, opened 2024-07-18
[CI] Investigating CI Runners


AlexandreSinger:
The self-hosted runners were down for the last couple of days and has only now gotten back up. I wanted to investigate any anomolies in the logs of the CIs to see if we have any issues in the testcases we are running which my have caused it.

The motivation behind this investigation is this message produced by the CI when the self-hosted runners were not working:
> The self-hosted runner: gh-actions-runner-vtr_XX lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

I went through the logs of the last working nightly test on the master branch ( https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/9932067866 ) and here are the results for the jobs run on the self-hosted runners (this data was collected from the figures at the bottom of the logs). I also collected their run time since I thought it may be valuable.
| Job Name      | Average CPU Usage (%) | Max RAM Usage (GB) | Max /dev/sda2 Usage (GB) | Max eth0 Data Received (Mb/s) | Max eth0 Data Sent (Mb/s) | Test Run Time |
|---------------|------------------------|---------------------|--------------------------|-------------------------------|---------------------------|---------------------------|
| "Capacity" | 100 | 125.82 | 492.0 | - | - | - |
| Run-tests (vtr_reg_nightly_test1, 16) | 32.58  | 6.74  | 32.4 | 577.73 | 69.72 | 2h 24m 25s |
| Run-tests (vtr_reg_nightly_test1_odin, 16, -DWITH_ODIN=ON) | 43.56  | 7.57  | 40.89 | 546.61 | 64.28 | 3h 5m 12s |
| Run-tests (vtr_reg_nightly_test2, 16) | 48.09  | **100.03**  | 97.97 | 630.83 | 33.65 | **4h 20m 20s** |
| Run-tests (vtr_reg_nightly_test2_odin, 16, -DWITH_ODIN=ON) | 54.27  | **98.35**  | 97.88 | 789.33 | 64.23 | **3h 39m 22s** |
| Run-tests (vtr_reg_nightly_test3, 16) | 64.54  | 16.66  | 33.2 | 551.44 | 69.02 | 2h 0m 3s |
[...]

PR #2662, opened 2024-07-23
[CI] Separated Out NightlyTests

AlexandreSinger:
Separated out the NightlyTests from the regular tests. This separation makes sense since the NightlyTests take so much longer to run and are run on different machines. This makes it easier to configure when these tests are run in the future (especially if we need to change the testing machines and do not want these tests to run as often).

Also made them run nightly instead of on push to master or PRs since they are currently not working. Once they start working again, we can re-enable this feature if we want.

This builds off issue #2652 . This should help bring the CI back up (at least getting some of the tests running).


AlexandreSinger:
@vaughnbetz This PR separates out the NightlyTests from the regular Tests. This also turns off the NightlyTests from running on PR for now while we fix the issue with CI (so people can more easily tell when their code is not working). Please review.

vaughnbetz:
Seems OK to me...

PR #2663, opened 2024-07-24
[CI] Added a Workflow Dispatch Trigger for NightlyTest

AlexandreSinger:
Added a workflow dispatch trigger so the NightlyTests can be run manually for testing purposes.

### DIFF
--- a/.github/workflows/nightly_test.yml
+++ b/.github/workflows/nightly_test.yml
@@ -1,6 +1,7 @@
 name: NightlyTest
 
 on:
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
@vaughnbetz This just adds the ability for the NightlyTests to be run manually instead of waiting for them to be run nightly. This will make it easier to test as we fix the self-hosted runners.

vaughnbetz:
Thanks!

PR #2666, opened 2024-07-26
Read Initial Placement

amin1377:
In this PR, I have added an option to the CLI to read the initial placement. It was tested on the Titan Murax circuit, and the results from reading the initial placement from a file were the same as those from running the initial placement during the flow.

### DIFF
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -1187,12 +1187,13 @@ void initial_placement(const t_placer_opts& placer_opts,
     clear_all_grid_locs();
 
     /* Go through cluster blocks to calculate the tightest placement
-    * floorplan constraint for each constrained block
-    */
+     * floorplan constraint for each constrained block
+     */
     propagate_place_constraints();
 
     /*Mark the blocks that have already been locked to one spot via floorplan constraints
-    * as fixed, so they do not get moved during initial placement or later during the simulated annealing stage of placement*/
+     * as fixed, so they do not get moved during initial placement or later during the simulated annealing stage of placement
+     */
     mark_fixed_blocks();
 
     // Compute and store compressed floorplanning constraints
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -1187,12 +1187,12 @@ void initial_placement(const t_placer_opts& placer_opts,
     clear_all_grid_locs();
 
     /* Go through cluster blocks to calculate the tightest placement
-     * floorplan constraint for each constrained block
-     */
+    * floorplan constraint for each constrained block
+    */
     propagate_place_constraints();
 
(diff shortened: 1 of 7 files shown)


vpr/src/place/initial_placement.cpp
```diff
@@ -1204,17 +1204,22 @@ void initial_placement(const t_placer_opts& placer_opts,
...
     }
 
-    if (noc_opts.noc) {
-        // NoC routers are placed before other blocks
-        initial_noc_placement(noc_opts, placer_opts);
-        propagate_place_constraints();
-    }
+    if(!placer_opts.read_initial_place_file.empty()) {
+        const auto& grid = g_vpr_ctx.device().grid;
+        read_place(nullptr, placer_opts.read_initial_place_file.c_str(), false, grid);
```
soheilshahrouz, line 1209: What happens if both constraint and initial placement files are specified?

I think read_constraints() initializes the location of some blocks and mark them as fixed. The location in the initial placement would override the location in the constraint file. We probably need to handle this case

vpr/src/place/place.cpp
```diff
@@ -462,7 +462,7 @@ void try_place(const Netlist<>& net_list,
     if (!placer_opts.write_initial_place_file.empty()) {
         print_place(nullptr,
                     nullptr,
-                    (placer_opts.write_initial_place_file + ".init.place").c_str());
+                    (placer_opts.write_initial_place_file).c_str());
```
soheilshahrouz, line 465: parentheses around placer_opts.write_initial_place_file are not needed

vpr/src/place/initial_placement.cpp
soheilshahrouz: I think it is better to print something to show that the initial placement is read from a file.

amin1377: You are right. The locations in the initial placement file will override the assigned locations in the constraint file. However, at the end of the initial placement, we call a function to check the legality of the placement, and any issues can be caught there.

Regarding printing that the initial placement is read from a file, the `read_place` function prints a message indicating that the placement is read from a file, along with the file name.

vpr/src/place/place.cpp
amin1377: Good eyes :)

vpr/src/place/initial_placement.cpp
soheilshahrouz: Are you referring to `check_initial_placement_legality()` or another function?

amin1377: Yes, I meant this function. I thought it would check whether the location of the fixed blocks matched the one specified in the constraint file, but upon reviewing the implementation, I realized that's not the case.

To avoid slowing down `set_block_location`, which is on the critical path of the placement code, I added the check when the placement file is read in the `read_place.cpp` file.

PR #2668, opened 2024-07-29
update fracturable LUT connectivity based on known legal external placements

KA7E:
Updated fracturable LUT connectivity in order to successfully reconstruct known legal clusters.

#### Description

Added complete interconnect from fracturable LUT data inputs to data inputs for each half LUT.

#### Motivation and Context

Unable to reconstruct known legal RippleFPGA-generated clusters without this change.

Successfully reconstructed RippleFPGA solutions for all 12 ISPD benchmarks.

#### Types of changes

#### Checklist:

### DIFF
--- a/vtr_flow/arch/ispd/ultrascale_ispd.xml
+++ b/vtr_flow/arch/ispd/ultrascale_ispd.xml
@@ -504,11 +504,12 @@
           </pb_type>
           <interconnect>
             <!-- Fracturable LUT connectivity: Based on Fig 3-1 of 'UltraScale Architecture CLB User Guide' UG574 (v1.5) -->
-            <direct name="I0" input="LUT6_2.I0" output="LUTX[0].I0 LUTX[1].I0"/>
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
@vaughnbetz One of the three PRs merged yesterday is causing vtr_reg_nightly_test5 to timout after 7 hours (it usually takes 3). I think this PR may be the culprit since this nightly test contains the `vpr_ispd` test and the other two PRs have nothing to do with vtr_reg_nightly_test5. Something seems to be causing these tests to take WAY longer than expected.

KA7E:
I'll take a look.

KA7E:
It shouldn't make them harder since it adds connectivity, but it may make
packing slower? I don't mind if you change it back for now.  It's only
needed for legalizing ispd placer generated placements which is not
something anyone but me needs to do right now.

vaughnbetz:
Thanks.  Reverting while investigating makes sense.
If necessary, we can make a different version of the ispd arch (ispd_flexible or some such) for the legalizer tests.

PR #2670, opened 2024-07-30
Revert "update fracturable LUT connectivity based on known legal external placements"

AlexandreSinger:
Reverts verilog-to-routing/vtr-verilog-to-routing#2668

vtr_reg_nightly_test5 now times out after 7 hours when it used to take around 3 hours. Only 3 PRs were merged yesterday, and I think this PR may have been the cause. Reverting the PR to see if the timeout goes away.


AlexandreSinger:
@vaughnbetz This has passed CI. Please when whenever you can so we can resolve the issue with CI.

PR #2672, opened 2024-08-01
Cong placement

behzadmehmood-rs:
#### Types of changes

#### Checklist:


soheilshahrouz:
A similar PR (#2384) was submitted a year ago. If this PR supersedes #2384, please close the old one.

serge-rs (approved):

behzadmehmood-rs:
> A similar PR (#2384) was submitted a year ago. If this PR supersedes #2384, please close the old one.

Yes, it supersedes PR (#2384).

vaughnbetz:
I suggest looking at the comments in the earlier PR and addressing them -- https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2384 
Some of those comments are pretty important; for example there are hard-coded limits on data structures that would lead to crashes on grids of size > 400 x 400, and I think those issues are still in this PR.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -54,6 +54,9 @@ enum class NetUpdateState {
 
 const int MAX_FANOUT_CROSSING_COUNT = 50;
 
+double cong_matrix[400][400];
```
vaughnbetz, line 57: Needs comments on what these are, what the overall congestion modeling approach is.
The approach seems to be very similar to a placement cost function that I wrote in my PhD and called NON_LINEAR_CONGESTION.  At the time I measured that code it had little gain in routability for architectures with all channels the same width (in various chip regions) and the CPU time was high.  Not sure if the trade-off is different with this code, but adding QoR data on some larger circuits would be important to check.

vaughnbetz (changes requested):

vpr/src/place/net_cost_handler.cpp
vaughnbetz: Also, hard-coding 400 is not good.

behzadmehmood-rs:
> I suggest looking at the comments in the earlier PR and addressing them -- #2384 Some of those comments are pretty important; for example there are hard-coded limits on data structures that would lead to crashes on grids of size > 400 x 400, and I think those issues are still in this PR.

I will address those.

PR #2674, opened 2024-08-02
Fix for Handling Missing Switch Types in RR Graph Builder

amin1377:
Previously, when the run-flat option was used and the RR graph was read from a file, the assumption was that all switch types (both inter-cluster and intra-cluster) were listed under the switch tag in the RR graph file. Consequently, when adding intra-cluster edges, if the RR graph builder encountered an edge that didn't have a corresponding switch type, it would throw an error.

In this PR, we removed this assumption. Now, if the RR graph builder cannot find a switch type, it will add a new RR switch type.

To test this PR, I used the Titan Murax circuit. The first run was with run-flat enabled, and the graph was built from scratch. For the second run, I first wrote the RR graph with run-flat disabled, so the RR graph file did not contain any of the intra-cluster switches. Then, I ran VPR with run-flat enabled again and read in the aforementioned RR graph. The routing results between these two runs were the same.

### DIFF
--- a/vpr/src/route/rr_graph.cpp
+++ b/vpr/src/route/rr_graph.cpp
@@ -272,7 +272,7 @@ static void connect_src_sink_to_pins(RRGraphBuilder& rr_graph_builder,
                                      t_rr_edge_info_set& rr_edges_to_create,
                                      const int delayless_switch,
                                      t_physical_tile_type_ptr physical_type_ptr,
-                                     bool is_remapped);
+                                     bool switches_remapped);
 
 static void alloc_and_load_tile_rr_graph(RRGraphBuilder& rr_graph_builder,
                                          std::map<int, t_arch_switch_inf>& arch_sw_inf_map,
@@ -381,7 +381,7 @@ static void add_pb_edges(RRGraphBuilder& rr_graph_builder,
                          int layer,
                          int i,
                          int j,
-                         bool is_remapped);
+                         bool switches_remapped);
 
 /**
  * Edges going in/out of collapse nodes are not added by the normal routine. This function add those edges
@@ -2039,7 +2039,7 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
     *Fc_clipped = false;
 
     /* This function is called to build the general routing graph resoruces. Thus, the edges are not remapped yet.*/
-    bool is_remapped = false;
+    bool switches_remapped = false;
 
     int num_edges = 0;
     /* Connection SINKS and SOURCES to their pins - Initializing IPINs/OPINs. */
@@ -2074,7 +2074,7 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
                                              rr_edges_to_create,
                                              delayless_switch,
                                              physical_tile,
-                                             is_remapped);
+                                             switches_remapped);
 
                     //Create the actual SOURCE->OPIN, IPIN->SINK edges
                     uniquify_edges(rr_edges_to_create);
@@ -2477,7 +2477,7 @@ static void connect_src_sink_to_pins(RRGraphBuilder& rr_graph_builder,
                                      t_rr_edge_info_set& rr_edges_to_create,
                                      const int delayless_switch,
                                      t_physical_tile_type_ptr physical_type_ptr,
-                                     bool is_remapped) {
+                                     bool switches_remapped) {
     for (auto class_num : class_num_vec) {
         const auto& pin_list = get_pin_list_from_class_physical_num(physical_type_ptr, class_num);
         auto class_type = get_class_type_from_class_physical_num(physical_type_ptr, class_num);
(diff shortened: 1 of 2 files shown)


vaughnbetz:
One routing failure in vtr_reg_strong when ASSERT_LEVEL is 3. spree doesn't route (either serial or parallel, so makes two failures and a bunch of QoR failures).  Maybe an assert fired?
regression_tests/vtr_reg_strong/strong_flat_router...[Fail]
2024-08-02T14:00:49.4647643Z [32;1m14:00:49[0m | [Fail]
2024-08-02T14:00:49.4649126Z [32;1m14:00:49[0m | k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common vpr_status Task value 'exited with return code 1' does not match golden 'success'
2024-08-02T14:00:49.4650354Z [32;1m14:00:49[0m | [Fail]
2024-08-02T14:00:49.4652150Z [32;1m14:00:49[0m | k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common routed_wirelength relative value -9.344921035417251e-05 outside of range [0.6,1.5], above absolute threshold 5.0 and not equal to golden value: 10701.0
2024-08-02T14:00:49.4653787Z [32;1m14:00:49[0m | [Fail]
2024-08-02T14:00:49.4655506Z [32;1m14:00:49[0m | k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common logic_block_area_total relative value -1.0868915445272294e-07 outside of range [0.8,1.3] and not equal to golden value: 9200550.0
2024-08-02T14:00:49.4657081Z [32;1m14:00:49[0m | [Fail]
2024-08-02T14:00:49.4658690Z [32;1m14:00:49[0m | k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common logic_block_area_used relative value -1.877042456823331e-07 outside of range [0.8,1.3] and not equal to golden value: 5327530.0
2024-08-02T14:00:49.4660240Z [32;1m14:00:49[0m | [Fail]
2024-08-02T14:00:49.4661907Z [32;1m14:00:49[0m | k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common min_chan_width_routing_area_total relative value -6.770893283950952e-07 outside of range [0.7,1.3] and not equal to golden value: 1476910.0
2024-08-02T14:00:49.4663420Z [32;1m14:00:49[0m | [Fail]
2024-08-02T14:00:49.4665071Z [32;1m14:00:49[0m | k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common min_chan_width_routing_area_per_tile relative value -0.0001327099504593755 outside of range [0.7,1.3] and not equal to golden value: 7535.23

vpr/src/route/rr_graph.cpp
```diff
@@ -271,7 +271,8 @@ static void connect_src_sink_to_pins(RRGraphBuilder& rr_graph_builder,
                                      const int j,
                                      t_rr_edge_info_set& rr_edges_to_create,
                                      const int delayless_switch,
-                                     t_physical_tile_type_ptr physical_type_ptr);
+                                     t_physical_tile_type_ptr physical_type_ptr,
+                                     bool is_remapped);
```
vaughnbetz, line 275: is_remapped doesn't seem like a good name.  Suggest switches_remapped.
Add a Doxygen comment above this function; at least describe switches_remapped but better to describe it all.

vpr/src/route/rr_graph.cpp
```diff
@@ -741,11 +744,25 @@ void create_rr_graph(const t_graph_type graph_type,
     }
 
     if (is_flat) {
+        short delayless_switch = OPEN;
```
vaughnbetz, line 747: Could make this a helper function: get_delayless_sw_index ().
Then add a Doxygen comment (right now the block off code is uncommented).

vpr/src/route/rr_graph.cpp
```diff
@@ -2021,6 +2038,9 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
     /* If Fc gets clipped, this will be flagged to true */
     *Fc_clipped = false;
 
+    /* This function is called to build the general routing graph resoruces. Thus, the edges are not remapped yet.*/
+    bool is_remapped = false;
```
vaughnbetz, line 2042: is_remapped -> switches_remapped.
Change comment to explain that the rr_graph will use switch indices from the architecture file at this point (not remapped).
Or change the name to switches_are_arch and negate all the boolean conditions (and still comment it!).

vpr/src/route/rr_graph.cpp
```diff
@@ -2960,19 +2979,10 @@ static void add_chain_node_fan_in_edges(RRGraphBuilder& rr_graph_builder,
                                                                       is_rr_sw_id,
                                                                       delay);
```
vaughnbetz, line 2981: Seems like you need a comment explaining why this data structure sometimes needs to be populated and sometimes doesn't.

vpr/src/route/rr_graph.cpp
```diff
@@ -2773,23 +2800,15 @@ static void add_pb_edges(RRGraphBuilder& rr_graph_builder,
...
-                            break;
-                        }
-                    }
-                }
-                // If the graph is loaded from a file, we expect that all sw types are already listed there since currently, we are not doing any further
-                // Optimization. If the optimization done when the rr graph file was generated is different from the current optimization, in the case that
-                // these optimizations create different RR switches, this VTR ASSERT can be removed.
-                VTR_ASSERT(found);
+                bool is_new_sw;
+                std::tie(is_new_sw, sw_idx) = find_create_intra_cluster_sw(rr_graph_builder,
```
vaughnbetz, line 2806: Check that this function has a doxygen comment and add one if it doesn't.

vaughnbetz: Should also add or improve Doxygen comments for related functions like add_intra_tile_edges_rr_graph which don't have Doxygen comments, or have only placeholder ones that don't really describe the parameters etc.

vaughnbetz (changes requested):

vpr/src/route/rr_graph.cpp
```diff
@@ -741,11 +744,25 @@ void create_rr_graph(const t_graph_type graph_type,
     }
 
     if (is_flat) {
+        short delayless_switch = OPEN;
+        if (load_rr_graph) {
+            const auto& rr_switches = device_ctx.rr_graph.rr_switch();
+            for (int switch_id = 0; switch_id < static_cast<int>(rr_switches.size()); switch_id++){
```
AlexandreSinger, line 750: Small nit-pick; instead of static casting the bound to an int, you can make switch_id a `size_t`. There is no usage of switch_id that requires it to be an int as far as I can tell.

vaughnbetz:
Thanks for addressing the changes ... now waiting for CI to run.  In the interim, can you confirm the spree tests pass when you run them by hand?

amin1377:
@vaughnbetz After the fix, I ran spree and it was working on my local machine.

PR #2680, opened 2024-08-08
if stopping after packing skip unnecessary placement setup

KA7E:
If running packing but not placement, return before unnecessary placement data structure setup.

#### Description

If packing returns success, added check to see if placement will run; if not, return.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/vpr_api.cpp
+++ b/vpr/src/base/vpr_api.cpp
@@ -383,6 +383,9 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
         if (!pack_success) {
             return false; //Unimplementable
         }
+        if (!vpr_setup.PlacerOpts.doPlacement) {
+            return true; // skip unnecessary placer setup
+        }
     }
 
     // For the time being, we decided to create the flat graph after placement is done. Thus, the is_flat parameter for this function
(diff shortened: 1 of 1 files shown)


vpr/src/base/vpr_api.cpp
```diff
@@ -383,6 +383,9 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
         if (!pack_success) {
             return false; //Unimplementable
         }
```
vaughnbetz, line 385: Will this work if --route is specified?  In that case packing is skipped (read from file), placement and skipped (read from file) but routing is performed.  Same question is --analysis is specified.  This change looks like it could break those flows.

KA7E:
Oh right it should check all the flags not just placement.

vpr/src/base/vpr_api.cpp
KA7E: doPlacement is an enum:

enum e_stage_action {
    STAGE_SKIP = 0,
    STAGE_LOAD,
    STAGE_DO,
    STAGE_AUTO
};

So the value checked on line 386 (!vpr_setup.PlacerOpts.doPlacement) is only true when doPlacement is set to STAGE_SKIP.  If Analysis or Routing is run but Placement is not, doPlacement is set to STAGE_LOAD, which does not trigger the early return in line 387.

PR #2683, opened 2024-08-12
[WIP] Router Lookahead Profiler

nedsels:
#### Types of changes

#### Checklist:


vaughnbetz:
@soheilshahrouz was also interested in reviewing this code.  We'll eventually need a QoR run to show no slowdown (good to run both VTR and Titan suites and attach spreadsheets).

vpr/src/route/lookahead_profiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+    VTR_ASSERT_SAFE(cluster_block_types.find(sink_inode) != cluster_block_types.end());
+    VTR_ASSERT_SAFE(tile_dimensions.find(sink_inode) != tile_dimensions.end());
+
+    std::string block_name = atom_block_names[sink_inode];
+    std::string atom_block_model = atom_block_models[sink_inode];
+    std::string cluster_block_type = cluster_block_types[sink_inode];
+    auto [tile_width, tile_height] = tile_dimensions[sink_inode];
+
+    /* Iterate through the given path and record information for each node */
+    for (size_t i = 2; i < branch_inodes.size(); ++i) { // Distance one node away is always 0. (IPIN->SINK)
```
soheilshahrouz, line 107: a more meaningful variable for the iteration number variable

vpr/src/route/lookahead_profiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+        lookahead_verifier_csv << delta_x << ",";              // delta x
+        lookahead_verifier_csv << delta_y << ",";              // delta y
+        lookahead_verifier_csv << actual_cost << ",";          // actual cost
+        lookahead_verifier_csv << actual_delay << ",";         // actual delay
+        lookahead_verifier_csv << actual_congestion << ",";    // actual congestion
+        lookahead_verifier_csv << lookahead_cost << ",";       // predicted cost
+        lookahead_verifier_csv << lookahead_delay << ",";      // predicted delay
+        lookahead_verifier_csv << lookahead_congestion << ","; // predicted congestion
+        lookahead_verifier_csv << cost_params.criticality;     // criticality
+        lookahead_verifier_csv << std::endl;
```
soheilshahrouz, line 171: std::endl flushes the buffer
avoid using it if it isn't necessary to flush the buffer

vpr/src/route/lookahead_profiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+            << ",delta x"
+            << ",delta y"
+            << ",actual cost"
+            << ",actual delay"
+            << ",actual congestion"
+            << ",predicted cost"
+            << ",predicted delay"
+            << ",predicted congestion"
+            << ",criticality"
+            << std::endl;
```
soheilshahrouz, line 56: is flushing the buffer necessary?

vpr/src/route/lookahead_profiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+    auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
+    auto& route_ctx = g_vpr_ctx.routing();
+
+    // If csv file hasn't been opened, open it and write out column headers
+    if (is_empty) {
+        lookahead_verifier_csv.open("lookahead_verifier_info.csv", std::ios::out);
+
+        if (!lookahead_verifier_csv.is_open()) {
+            VTR_LOG_ERROR("Could not open lookahead_verifier_info.csv", "error");
```
soheilshahrouz, line 31: what is the second argument?

vpr/src/route/lookahead_profiler.h
```diff
@@ -0,0 +1,58 @@
+#ifndef VTR_LOOKAHEAD_PROFILER_H
+#define VTR_LOOKAHEAD_PROFILER_H
+
+#include <fstream>
+#include <thread>
```
soheilshahrouz, line 5: why is thread needed?

vpr/src/route/route_common.cpp
```diff
@@ -835,9 +839,15 @@ void reserve_locally_used_opins(HeapInterface* heap, float pres_fac, float acc_f
...
-                                 0., 0.);
+                add_node_to_heap(heap,
+                                 route_ctx.rr_node_route_inf,
+                                 to_node,
+                                 cost,
+                                 RREdgeId::INVALID(),
+                                 0.,
+                                 0.,
+                                 0.,
+                                 0.);
```
soheilshahrouz, line 850: named parameter comments

vpr/src/route/lookahead_profiler.h
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * @brief A class which records information used to profile the router lookahead: most importantly,
+ * the actual cost (delay and congestion) from nodes to the sink to which they have been routed, as
+ * well as the lookahead's estimation of this cost.
+ */
+class LookaheadProfiler {
+  public:
+    LookaheadProfiler()
+        : is_empty(true) {}
```
soheilshahrouz, line 19: delete default constuctors and operators, assuming this class is not intented to be copied or moved

vpr/src/route/route_net.tpp
```diff
@@ -351,7 +356,16 @@ inline NetResultFlags pre_route_to_clock_root(ConnectionRouter& router,
      * points. Therefore, we can set the net pin index of the sink node to      *
      * OPEN (meaning illegal) as it is not meaningful for this sink.            */
     vtr::optional<const RouteTreeNode&> new_branch, new_sink;
-    std::tie(new_branch, new_sink) = tree.update_from_heap(&cheapest, OPEN, ((high_fanout) ? &spatial_rt_lookup : nullptr), is_flat);
+    std::tie(new_branch, new_sink) = tree.update_from_heap(&cheapest,
+                                                           OPEN,
```
soheilshahrouz, line 360: named parameter comments

vpr/src/route/route_path_manager.cpp
```diff
@@ -39,7 +39,7 @@ void PathManager::mark_node_visited(RRNodeId node) {
     }
 }
 
-void PathManager::insert_backwards_path_into_traceback(t_heap_path* path_data, float cost, float backward_path_cost, RoutingContext& route_ctx) {
+void PathManager::insert_backwards_path_into_traceback(t_heap_path* path_data, float cost, float backward_path_cost, float backward_path_delay, float backward_path_congestion, RoutingContext& route_ctx) {
```
soheilshahrouz, line 42: this line is too long

vpr/src/route/route_path_manager.h
```diff
@@ -74,7 +74,7 @@ class PathManager {
     void set_enabled(bool enable);
 
     // Insert the partial path data into the main route context traceback
-    void insert_backwards_path_into_traceback(t_heap_path* path_data, float cost, float backward_path_cost, RoutingContext& route_ctx);
+    void insert_backwards_path_into_traceback(t_heap_path* path_data, float cost, float backward_path_cost, float backward_path_delay, float backward_path_congestion, RoutingContext& route_ctx);
```
soheilshahrouz, line 77: too long

vpr/src/route/route_tree.cpp
```diff
@@ -486,13 +486,22 @@ void RouteTree::print(void) const {
...
+                            const ParentNetId& net_id,
+                            const int itry,
+                            bool profile_lookahead) {
     /* Lock the route tree for writing. At least on Linux this shouldn't have an impact on single-threaded code */
     std::unique_lock<std::mutex> write_lock(_write_mutex);
 
     //Create a new subtree from the target in hptr to existing routing
     vtr::optional<RouteTreeNode&> start_of_new_subtree_rt_node, sink_rt_node;
-    std::tie(start_of_new_subtree_rt_node, sink_rt_node) = add_subtree_from_heap(hptr, target_net_pin_index, is_flat);
+    std::tie(start_of_new_subtree_rt_node, sink_rt_node) = add_subtree_from_heap(hptr, target_net_pin_index, is_flat, router_lookahead, cost_params, itry, net_list, net_id, profile_lookahead);
```
soheilshahrouz, line 504: too long

vpr/src/route/route_tree.h
```diff
@@ -357,7 +358,16 @@ class RouteTree {
...
+    update_from_heap(t_heap* hptr,
+                     int target_net_pin_index,
+                     SpatialRouteTreeLookup* spatial_rt_lookup,
+                     bool is_flat,
+                     const RouterLookahead& router_lookahead,
+                     t_conn_cost_params cost_params,
+                     const Netlist<>& net_list,
+                     const ParentNetId& net_id,
+                     int itry = -1,
+                     bool profile_lookahead = false);
```
soheilshahrouz, line 370: i am not a fan of default argument values

itry doesn't seem to need a default value because it is used only by the profiler.
I think we can get rid of `profile_lookahead`. The profiler is a global object, and it can store a flag to determine if it is enabled

vpr/src/route/route_tree.h
```diff
@@ -87,6 +87,7 @@
 #include <unordered_map>
 
 #include "connection_based_routing_fwd.h"
+#include "lookahead_profiler.h"
```
soheilshahrouz, line 90: can't this be replaced by a forward declaration?

vpr/src/route/router_delay_profiling.cpp
```diff
@@ -121,7 +121,14 @@ bool RouterDelayProfiler::calculate_delay(RRNodeId source_node,
         VTR_ASSERT(cheapest.index == sink_node);
 
         vtr::optional<const RouteTreeNode&> rt_node_of_sink;
-        std::tie(std::ignore, rt_node_of_sink) = tree.update_from_heap(&cheapest, OPEN, nullptr, is_flat_);
+        std::tie(std::ignore, rt_node_of_sink) = tree.update_from_heap(&cheapest,
+                                                                       OPEN,
+                                                                       nullptr,
```
soheilshahrouz, line 126: named parameter comment

vpr/src/route/router_delay_profiling.cpp
```diff
@@ -207,7 +218,14 @@ vtr::vector<RRNodeId, float> calculate_all_path_delays_from_rr_node(RRNodeId src
             //Build the routing tree to get the delay
             tree = RouteTree(RRNodeId(src_rr_node));
             vtr::optional<const RouteTreeNode&> rt_node_of_sink;
-            std::tie(std::ignore, rt_node_of_sink) = tree.update_from_heap(&shortest_paths[sink_rr_node], OPEN, nullptr, router_opts.flat_routing);
+            std::tie(std::ignore, rt_node_of_sink) = tree.update_from_heap(&shortest_paths[sink_rr_node],
+                                                                           OPEN,
+                                                                           nullptr,
```
soheilshahrouz, line 223: named parameter comment

vpr/src/route/router_lookahead.cpp
```diff
@@ -61,13 +61,19 @@ std::unique_ptr<RouterLookahead> make_router_lookahead(const t_det_routing_arch&
     return router_lookahead;
 }
 
+void RouterLookahead::scale_delay_and_cong_by_criticality(float& delay, float& cong, const t_conn_cost_params& params) const {
```
soheilshahrouz, line 64: instead of passing t_conn_cost_params, only pass criticality as an argument

vpr/src/route/router_lookahead.cpp
```diff
@@ -61,13 +61,19 @@ std::unique_ptr<RouterLookahead> make_router_lookahead(const t_det_routing_arch&
...
+    delay *= params.criticality;
+    cong *= 1. - params.criticality;
+}
```
soheilshahrouz, line 67: is this member functon is always called after `get_expected_delay_and_cong()`?
In this case, I think it is better to replace this member function with another one that computes the scaled congestion and delay

vpr/src/route/router_lookahead.cpp
```diff
@@ -61,13 +61,19 @@ std::unique_ptr<RouterLookahead> make_router_lookahead(const t_det_routing_arch&
     return router_lookahead;
 }
 
+void RouterLookahead::scale_delay_and_cong_by_criticality(float& delay, float& cong, const t_conn_cost_params& params) const {
+    delay *= params.criticality;
+    cong *= 1. - params.criticality;
```
soheilshahrouz, line 66: 1.f

vpr/src/route/router_lookahead.cpp
```diff
@@ -109,7 +115,7 @@ float NoOpLookahead::get_expected_cost(RRNodeId /*current_node*/, RRNodeId /*tar
     return 0.;
 }
 
-std::pair<float, float> NoOpLookahead::get_expected_delay_and_cong(RRNodeId /*node*/, RRNodeId /*target_node*/, const t_conn_cost_params& /*params*/, float /*R_upstream*/) const {
+std::pair<float, float> NoOpLookahead::get_expected_delay_and_cong(RRNodeId, RRNodeId, const t_conn_cost_params&, float) const {
```
soheilshahrouz, line 118: why did you remove argument name comments?

vpr/src/route/router_lookahead_compressed_map.h
```diff
@@ -27,10 +27,7 @@ class CompressedMapLookahead : public RouterLookahead {
   protected:
     float get_expected_cost(RRNodeId node, RRNodeId target_node, const t_conn_cost_params& params, float R_upstream) const override;
 
-    std::pair<float, float> get_expected_delay_and_cong(RRNodeId from_node,
-                                                        RRNodeId to_node,
-                                                        const t_conn_cost_params& params,
-                                                        float R_upstream) const override;
+    std::pair<float, float> get_expected_delay_and_cong(RRNodeId from_node, RRNodeId to_node, const t_conn_cost_params& params, float R_upstream) const override;
```
soheilshahrouz, line 30: long line

vpr/src/route/router_lookahead_extended_map.cpp
```diff
@@ -181,7 +181,7 @@ float ExtendedMapLookahead::get_chan_ipin_delays(RRNodeId to_node) const {
 //
 //  The from_node can be of one of the following types: CHANX, CHANY, SOURCE, OPIN
 //  The to_node is always a SINK
-std::pair<float, float> ExtendedMapLookahead::get_expected_delay_and_cong(RRNodeId from_node, RRNodeId to_node, const t_conn_cost_params& params, float /*R_upstream*/) const {
+std::pair<float, float> ExtendedMapLookahead::get_expected_delay_and_cong(RRNodeId from_node, RRNodeId to_node, const t_conn_cost_params& params, float) const {
```
soheilshahrouz, line 184: why was the argument name comment removed?

vpr/src/route/router_lookahead_extended_map.cpp
```diff
@@ -237,8 +237,8 @@ std::pair<float, float> ExtendedMapLookahead::get_expected_delay_and_cong(RRNode
     float site_pin_delay = this->get_chan_ipin_delays(to_node);
     expected_delay += site_pin_delay;
 
-    float expected_delay_cost = params.criticality * expected_delay;
-    float expected_cong_cost = (1.0 - params.criticality) * expected_congestion;
+    float expected_delay_cost = expected_delay * params.criticality;
+    float expected_cong_cost = expected_congestion * (1. - params.criticality);
```
soheilshahrouz, line 241: why did you need to change the order?

vpr/test/test_connection_router.cpp
```diff
@@ -87,7 +87,14 @@ static float do_one_route(RRNodeId source_node,
 
         // Get the delay
         vtr::optional<const RouteTreeNode&> rt_node_of_sink;
-        std::tie(std::ignore, rt_node_of_sink) = tree.update_from_heap(&cheapest, OPEN, nullptr, router_opts.flat_routing);
+        std::tie(std::ignore, rt_node_of_sink) = tree.update_from_heap(&cheapest,
+                                                                       OPEN,
+                                                                       nullptr,
```
soheilshahrouz, line 92: named argument comment

vpr/src/route/lookahead_profiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+                               const RouterLookahead& router_lookahead,
+                               const ParentNetId& net_id,
+                               const Netlist<>& net_list,
+                               const std::vector<RRNodeId>& branch_inodes) {
+    auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
+    auto& route_ctx = g_vpr_ctx.routing();
+
+    // If csv file hasn't been opened, open it and write out column headers
+    if (is_empty) {
```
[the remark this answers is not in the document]
nedsels: The only issue is that the profiler object is initialized in ``RoutingContext``, so if opening the file is done in the constructor, it will happen even if the option to use it isn't on

vpr/src/route/lookahead_profiler.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+            auto tile_type = physical_tile_type(cluster_block_id);
+            tile_dimensions[sink_inode] = std::pair(std::to_string(tile_type->width), std::to_string(tile_type->height));
+        } else {
+            atom_block_names[sink_inode] = "--";
+            atom_block_models[sink_inode] = "--";
+            cluster_block_types[sink_inode] = "--";
+            tile_dimensions[sink_inode] = {"--", "--"};
+        }
+    }
```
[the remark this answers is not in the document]
nedsels: Do you mean that I should initialize them as empty maps in the constructor?

vpr/src/route/route_tree.h
nedsels: Where should I store this flag?

vpr/src/route/lookahead_profiler.cpp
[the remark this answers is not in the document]
nedsels: I'm not sure how to do this without creating a constructor for RoutingContext and VprContext as well; would this be okay?

vpr/src/route/route_tree.h
nedsels: Forward declaration here creates compiler errors

vpr/src/route/lookahead_profiler.cpp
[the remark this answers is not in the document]
soheilshahrouz: You're right. RoutingContext has a global instance, meaning that LookaheadProfiler's constructor is called before parsing command line options.

I think we shouldn avoid adding anything to the global context unless it is really necessary. Instead, we could ceate a LookaheadProfiler instance at the start of the routing stage and pass it through a function call hierarchy down to its call site.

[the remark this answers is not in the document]
soheilshahrouz: If you decide to keep LookaheadProfiler instance global, you should free the memory allocated for its member variables after it is no longer needed. This is the global instance doesn't go out of scope

vpr/src/route/lookahead_profiler.cpp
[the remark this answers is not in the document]
soheilshahrouz: What I had in mind was initializing them with all sinks ids as key and proper values. However, this can't be done when LookaheadProfiler has a global instance.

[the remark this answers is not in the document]
soheilshahrouz: I think storing multiple std::string values for each sink id is not memory efficient. For `atom_block_models`, `cluster_block_types`, `and tile_dimensions`, the number of unique values is very low. Even for `atom_block_names` where each unique value appears only a few times, you don't need to store multiple copies of each block name.

std::string_view, vtr::interned_string can be used to avoid unnecessary string copies.
It is also possible to store atom and cluster block ids and use global netlist objects to retrieve references to block/model names and tile sizes.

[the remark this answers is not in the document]
nedsels: I decided to change these variables to store IDs and type pointers. Dereferencing the pointers may be less runtime-efficient, but to use ``std::string_view``, I think I'd have to create a separate ``std::set`` which contains all the ``string_view``s encountered so far to see if we need to fetch the info for the current node, which seems less clean.

nedsels:
@vaughnbetz @AlexandreSinger Python lint is failing because some of the libraries I used can't be imported; is there any way around this?

vpr/src/route/router_lookahead.cpp
nedsels: Always except for the lookahead profiler. I moved the code in ``get_expected_delay_and_cong()`` into a new function ``get_expected_delay_and_cong_ignore_criticality()``, and now ``get_expected_delay_and_cong()`` just calls that new function + the scaling function. So, all the previous calls to ``get_expected_delay_and_cong()`` remain unmodified and the profiler can call ``get_expected_delay_and_cong_ignore_criticality()``.

soheilshahrouz:
> @vaughnbetz @AlexandreSinger Python lint is failing because some of the libraries I used can't be imported; is there any way around this?

It can be probably fixed by updating requirements.txt file

vpr/src/route/lookahead_profiler.cpp
```diff
@@ -0,0 +1,169 @@
...
+        lookahead_verifier_csv_ << cost_params.criticality;     // criticality
+        lookahead_verifier_csv_ << "\n";
+    }
+}
+
+void LookaheadProfiler::clear() {
+    net_pin_blocks_.clear();
+    sink_atom_block_.clear();
+    sink_cluster_block_.clear();
+    tile_types_.clear();
```
soheilshahrouz, line 168: I guess clear() does not deallocate memory. Vtr has a utility function that swaps a given container with an empty one to force memory deallocation. If I remember correctly, it is called release_memory()

nedsels:
@soheilshahrouz @vaughnbetz Some regtests are having some QoR failures. On ``vtr_reg_nightly_test2`` (and ``_odin``), there are a few, but I'm not sure if the first 3 are small enough to just update the golden results:

image

However, on ``vtr_reg_nightly_test7``, there are some failures which seem quite bad, and I'm not sure why or how to approach these. Do you have any advice on how to proceed?

image

vaughnbetz:
I'm not too worried about dsip; it is small.

The long route times on wb_conmax are also probably OK if there are just a few of these.  Adding @amin1377 since those 3D and flat routing ones are probably his tests.

The more I think about it the more I think we should #ifdef adding data members to the heap though; I am really not keen on carrying around extra data in a big and hot data structure for profiling.  I think it would be better to have no impact on the default flow.

vpr/src/route/lookahead_profiler.h
```diff
@@ -0,0 +1,80 @@
...
+     * @param file_name The name of the output csv file.
+     */
+    void set_file_name(const std::string& file_name);
+
+    /**
+     * @brief Record information on nodes on a path from a source to a sink.
+     *
+     * @param iteration The router iteration.
+     * @param target_net_pin_index Target pin of this sink in the net.
+     * @param cost_params
```
vaughnbetz, line 41: Suggest briefly commenting each input's purpose.

vaughnbetz: You need a high-level comment of how the lookahead profiler works somewhere (file level, or maybe on this routine).

vpr/src/route/router_lookahead.h
```diff
@@ -123,7 +169,7 @@ const RouterLookahead* get_cached_router_lookahead(const t_det_routing_arch& det
 class ClassicLookahead : public RouterLookahead {
   public:
     float get_expected_cost(RRNodeId node, RRNodeId target_node, const t_conn_cost_params& params, float R_upstream) const override;
-    std::pair<float, float> get_expected_delay_and_cong(RRNodeId node, RRNodeId target_node, const t_conn_cost_params& params, float R_upstream) const override;
+    std::pair<float, float> get_expected_delay_and_cong_ignore_criticality(RRNodeId node, RRNodeId target_node, const t_conn_cost_params& params, float R_upstream) const override;
```
vaughnbetz, line 172: Long line; maybe break it.

vpr/src/route/router_lookahead.h
```diff
@@ -15,15 +15,61 @@ class RouterLookahead {
...
+     *
      * @return
      */
     virtual float get_expected_cost(RRNodeId node, RRNodeId target_node, const t_conn_cost_params& params, float R_upstream) const = 0;
-    virtual std::pair<float, float> get_expected_delay_and_cong(RRNodeId node, RRNodeId target_node, const t_conn_cost_params& params, float R_upstream) const = 0;
+
+    /**
+     * @brief Get expected (delay, congestion) from node to target_node.
+     *
+     * @attention Either compute or read methods must be invoked before invoking get_expected_delay_and_cong_ignore_criticality.
```
vaughnbetz, line 33: must be invoked --> must be invoked to set up the lookahead data structures

vaughnbetz (changes requested):
Looks good; just have some commenting suggestions. We'll also need a QoR run on VTR & Titan to check cpu time & quality.

nedsels:
Sorry for the delay with the QoR results; for some reason, they have degraded with this change. I'm trying to figure out why; if/when I can, I'll post the results in the description.

vaughnbetz:
Thanks. If you can post what the current QoR is that would be helpful too; even if it is not that good it helps show what the status is.

PR #2686, opened 2024-08-14
[LOGGING] Fixed LUT Sizing Issue

### LINKED ISSUE #2685: [Logging] Incorrect LUT Sizes Circuit Statistic Printing
For circuits such as `clma` from the MCNC benchmarks, the circuit statistics for the number of LUT blocks of a given input size appears to be incorrect:
image

These circuit statistics say that all the LUTs are 6-LUTs; however, looking into the blif file for this circuit, we clearly see that this is not the case:
image

The circuit statistics are printed here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/21d015036969a955298cbc9b94ca74fd35a016be/vpr/src/base/read_circuit.cpp#L147-L217

It uses the port-width of the netlist as the LUT size; but this appears to not be correct. This function should be rewritten to count the number of valid input pins coming into the LUT, not the width of the port.

AlexandreSinger:
Fixed issue where the LUT sizing for printing the circuit statistics was using the port width instead of the number of input pins.

The CLMA example from the issue:
image
Now looks like this:
image

Which makes a lot more sense.

closes #2685 

EDIT:

Based on Vaughn's comments below, changed the format of the statistics to the following:
image

This now includes the total number of LUTs (called .names here), as well as a breakdown of their sizes.


AlexandreSinger:
I am suprised this change failed CI. Looks like the parsing scripts may have been relying on this bug to count the number of LUTs:
image

I think these scripts may need to be checked. Or maybe it is as simple as updating the golden results.

vaughnbetz:
Thanks.  I suggest adding a total LUTs line as well, and then updating the parsing rule. That should fix CI (seems we were parsing the 6-LUT line as the total LUTs! ).

AlexandreSinger:
@vaughnbetz That makes sense to me! Would you like to include 0-LUTs in that total? I am not 100% on the semantics of what a "0-LUT" is beyond being a constant generator.

vaughnbetz:
Sure, I'd count it.  It is a constant generator.

AlexandreSinger:
@vaughnbetz I have updated the circuit statistics to include the total number of LUTs. I have tried to follow the format of the rest of the logs:
image

Let me know what you think. I personally like it; but I can understand if people may be confused by ".output" coming after all the LUT sizes.

vaughnbetz:
Looks good to me!

AlexandreSinger:
@vaughnbetz This PR has passed CI. Please review and merge when you have a moment. I have updated the parsing script, which required me to regenerate the golden results (since some testcases were mis-counting the number of LUTs in designs).

PR #2700, opened 2024-08-21
[Warnings] Cleaned Up Most Warnings in GCC-13

AlexandreSinger:
This PR resolves several small issues which were causing warnings in the GCC13 build. Trying to get GCC13 warning clean before we upgrade the CI.

This resolves the following warnings:
- Fixed incorrect forward declaration
- Fixed ambiguous unique pointer allocation
- Converted an int into a size_t
- Fixed array access out of bounds
- Fixed potential memory leak
- Changed C String to STD String
- Fixed const reference
- Removed -lz flag from compilation
- Added missing headers to a file

See each of the associated commits for more details.

There are still warnings coming from the nDMatrix class (discussed previously) which should be resolved in its own PR. There are also a couple of other small warnings which are slightly more complex to resolve.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,11 +130,6 @@ endif()
 # Build type flags
 #
 
-set(EXTRA_FLAGS "")
-if(VPR_ENABLE_INTERCHANGE)
-    set(EXTRA_FLAGS "-lz")
-endif()
-
 if(NOT MSVC)
     # for GCC and Clang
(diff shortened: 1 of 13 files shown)


AlexandreSinger:
@vaughnbetz This has passed CI. Please review when you have a moment. This PR is cleaning up simple warnings in GCC-13 in preparation for us upgrading the CI. Still more warnings and things to fix, but this fixes most of them.

vaughnbetz:
Thanks!

PR #2701, opened 2024-08-21
fixed Issue When Re-Creating Elements in Atom and Clustered Netlists

### LINKED ISSUE #2690: [Netlist] Potential Issue When Re-Creating Elements in Atom and Clustered Netlists
When implementing another Netlist class based on the Atom and Clustered netlists, I noticed a bug in their create methods.

This bug occurs in a few places, but for example the create block method specifies that it creates a block or returns the ID if it already exists:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/21d015036969a955298cbc9b94ca74fd35a016be/vpr/src/base/atom_netlist.h#L167-L175

However, in the implementation, the `push_back` method is used on the internalized data:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/21d015036969a955298cbc9b94ca74fd35a016be/vpr/src/base/atom_netlist.cpp#L140-L155

`Netlist::create_block` will return the ID of a block if one with that name already exists; but then there will be a mismatch with the internal data. This will cause a crash if a block is re-created into the netlist.

This can be simply fixed by changing `push_back` into `insert`. The performance should not be too different since the `vtr::vector_map` class treats insertions as a resize (similar to a push back).

This should be fixed for many of the `create` methods in both the Atom Netlist and the Clustered Netlist (or the documentation should be updated to reflect that re-creation of IDs is not supported).

ZohairZaidi:
#### Description

I fixed an issue in the AtomNetlist and ClusteredNetlist classes where the create methods (such as create_block, create_port, create_pin) used push_back to add elements to internal data structures. This caused a mismatch when a block, port, or pin with the same name was re-created, leading to potential crashes. The fix involved replacing push_back with insert to ensure that the internal data structures remain consistent even when elements are re-created.

#### Related Issue

closes #2690 
#### Motivation and Context

This change is required to prevent potential crashes and data inconsistencies in the AtomNetlist and ClusteredNetlist classes when blocks, ports, or pins are re-created

The changes were tested by running the VTR project's regression test suites
#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/atom_netlist.cpp
+++ b/vpr/src/base/atom_netlist.cpp
@@ -141,8 +141,9 @@ AtomBlockId AtomNetlist::create_block(const std::string& name, const t_model* mo
     AtomBlockId blk_id = Netlist::create_block(name);
 
     //Initialize the data
-    block_models_.push_back(model);
-    block_truth_tables_.push_back(truth_table);
+    // Use insert instead of push_back
+    block_models_.insert(blk_id, model);
+    block_truth_tables_.insert(blk_id, truth_table);
 
     //Check post-conditions: size
     VTR_ASSERT(validate_block_sizes());
@@ -174,7 +175,8 @@ AtomPortId AtomNetlist::create_port(const AtomBlockId blk_id, const t_model_port
     if (!port_id) {
         port_id = Netlist::create_port(blk_id, model_port->name, model_port->size, type);
 
-        port_models_.push_back(model_port);
+        // Use insert instead of push_back
+        port_models_.insert(port_id, model_port);
         associate_port_with_block(port_id, port_type(port_id), blk_id);
     }
 
--- a/vpr/src/base/atom_netlist.cpp
+++ b/vpr/src/base/atom_netlist.cpp
@@ -141,7 +141,7 @@ AtomBlockId AtomNetlist::create_block(const std::string& name, const t_model* mo
     AtomBlockId blk_id = Netlist::create_block(name);
 
     //Initialize the data
-    // Use insert instead of push_back
+    // Push back new data or overwrite existing data 
     block_models_.insert(blk_id, model);
     block_truth_tables_.insert(blk_id, truth_table);
 
@@ -175,7 +175,7 @@ AtomPortId AtomNetlist::create_port(const AtomBlockId blk_id, const t_model_port
     if (!port_id) {
         port_id = Netlist::create_port(blk_id, model_port->name, model_port->size, type);
 
-        // Use insert instead of push_back
+        // Push back new data or overwrite existing data 
         port_models_.insert(port_id, model_port);
         associate_port_with_block(port_id, port_type(port_id), blk_id);
     }
--- a/vpr/src/base/clustered_netlist.cpp
+++ b/vpr/src/base/clustered_netlist.cpp
@@ -107,6 +107,7 @@ ClusterBlockId ClusteredNetlist::create_block(const char* name, t_pb* pb, t_logi
     if (blk_id == ClusterBlockId::INVALID()) {
         blk_id = Netlist::create_block(name);
 
+        // Push back new data or overwrite existing data
         block_pbs_.insert(blk_id, pb);
         block_types_.insert(blk_id, type);
 
@@ -144,7 +145,7 @@ ClusterPortId ClusteredNetlist::create_port(const ClusterBlockId blk_id, const s
 ClusterPinId ClusteredNetlist::create_pin(const ClusterPortId port_id, BitIndex port_bit, const ClusterNetId net_id, const PinType pin_type_, int pin_index, bool is_const) {
     ClusterPinId pin_id = Netlist::create_pin(port_id, port_bit, net_id, pin_type_, is_const);
 
-    // insert
+    // Push back new data or overwrite existing data
     pin_logical_index_.insert(pin_id, pin_index); 
 
     ClusterBlockId block_id = port_block(port_id);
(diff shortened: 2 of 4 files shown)


vpr/src/base/atom_netlist.cpp
```diff
@@ -141,8 +141,9 @@ AtomBlockId AtomNetlist::create_block(const std::string& name, const t_model* mo
     AtomBlockId blk_id = Netlist::create_block(name);
 
     //Initialize the data
-    block_models_.push_back(model);
-    block_truth_tables_.push_back(truth_table);
+    // Use insert instead of push_back
```
AlexandreSinger, line 144: Can you please make this comment a bit more clear. Something like "push back new data or overwrite data for block". This makes it clear that the next lines are both initializing the data and overwriting it in this line. There is also no need to mention that we are using insert instead of push_back since, after your change, the context of this comment will no longer make sense.

For this, and the other functions you fixed, could you please update the Doxygen comment to reflect this change. It should mention that when an existing block is found, the data is overwritten. Its a good idea to document this stuff well.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/33a82f1c2f0b1a8e95c0efb06691b0d54ecfa629/vpr/src/base/atom_netlist.h#L167-L176

vpr/src/base/atom_netlist.cpp
```diff
@@ -174,7 +175,8 @@ AtomPortId AtomNetlist::create_port(const AtomBlockId blk_id, const t_model_port
     if (!port_id) {
         port_id = Netlist::create_port(blk_id, model_port->name, model_port->size, type);
 
-        port_models_.push_back(model_port);
+        // Use insert instead of push_back
```
AlexandreSinger, line 178: Same comment here. Please use the exact same comment that you use above. It may seem silly to duplicate comments, but it makes it clear that we are using the same "technique" everywhere.

vpr/src/base/clustered_netlist.cpp
```diff
@@ -144,7 +144,8 @@ ClusterPortId ClusteredNetlist::create_port(const ClusterBlockId blk_id, const s
 ClusterPinId ClusteredNetlist::create_pin(const ClusterPortId port_id, BitIndex port_bit, const ClusterNetId net_id, const PinType pin_type_, int pin_index, bool is_const) {
     ClusterPinId pin_id = Netlist::create_pin(port_id, port_bit, net_id, pin_type_, is_const);
 
-    pin_logical_index_.push_back(pin_index);
+    // insert
```
AlexandreSinger, line 147: Please duplicate the comment that you will use from above.

AlexandreSinger (changes requested):
Hi @ZohairZaidi , thank you so much for your swift PR! The changes look good to me. I just left a few code comment changes that I would like to see before we merge it! So far the CI seems to be good.

Also, a super minor comment. Could you please also write "closes #..."; this will close the issue the second this PR gets merged. It just saves one extra button press.

ZohairZaidi:
Thanks for the smooth guide @AlexandreSinger, I have updated the PR addressing the suggestions you made.

AlexandreSinger (approved):
LGTM

AlexandreSinger:
Thank you so much! The changes look good to me. One thing I noticed is that the runtime of `vtr_reg_nightly_test2` seems to have increased slightly. This makes sense since the `insert` method of `vector_map` does an extra check to see if the index is within bounds. @vaughnbetz What do you think? vtr_reg_nightly_test2 increased from 3.5 hours to 4.4 hours. There may be a way to resolve this by 1) using push back and updating the documentation to say that we no longer allow duplicate creations into the netlist; 2) creating an "update" method where the user can update a value (if it exists); 3) just accepting the runtime degradation. Most nightly tests saw around a 30 minute degredation; with nightly test 2 seeing the largest degradation.

I am surprised that this caused such a large increase in runtime. Perhaps it was just a fluke and it should be run locally?

vaughnbetz:
I think we'd better get a controlled QoR experiment -- the developer guide explains how to run one, or @AlexandreSinger could run one.  I am worried about this change as it changes O(1) operations to O(N) so we may have slowed down netlist creation a lot, depending on how these methods are called.

Perhaps we could have a faster method by checking if a netlist item exists and then updating it, and otherwise using push_back() for the common case of adding a new element?

AlexandreSinger:
@vaughnbetz For vtr::vector_map, `push_back` and `insert` should have the same time complexity. The implementation was designed to make insertions fast:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/79a5bc8d163ce15aa195a84459f6350690e1fbf5/libs/libvtrutil/src/vtr_vector_map.h#L149-L157

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/79a5bc8d163ce15aa195a84459f6350690e1fbf5/libs/libvtrutil/src/vtr_vector_map.h#L109-L110

I was suprised how much this may have slowed things down. Perhaps `push_back` is doubling the internal container each time it hits its max size; while `insert` is resizing it to the exact size it needs. Perhaps this can be made more efficient?

But Honestly this may not be worth the time. Perhaps the simplest approach is just changing the documentation for `create_...` for these netlist classes and making a special method for updating if people need it.

vaughnbetz:
Some search seem to suggest resize() has amortized O(1) complexity (i.e. doubles capacity when needed) in which case it would be fast enough.  It seems a controlled QoR test to see if the slowdown is real might be worthwhile.

I'm also fine with changing the documentation if that works and is easier.

ZohairZaidi:
hi @AlexandreSinger, its up to you to decide whether a QoR test should be conducted to confirm the slowdown, or if we should just proceed with updating the documentation. If you choose the latter, I can update the docs immediately.

AlexandreSinger:
@ZohairZaidi I am not confident that a QoR test will demonstrate any difference; we really do not want the CI to increase by an hour on a test like this.

I think the simplest solution would be to update the documentation. Thank you.

vaughnbetz:
Sounds like a good plan to me.  @ZohairZaidi & @AlexandreSinger : you could also consider making new methods (e.g. .update()) that use this new code to safely update existing elements. Then the documentation can be updated to say which method to use when.

If you decide to do that (create new methods) you should create a unit test to make sure they work too.  But if this is too much work I'm also fine with just updating the existing documentation as it seems we don't have a use case for an update() method yet anyway.

PR #2703, opened 2024-08-24
[Prepacking] Re-Designed Prepacker API

AlexandreSinger:
The original Prepacker API created a lot of global variables which were confusing to work with and hard to maintain.

Created a Prepacker class which maintains the state of the Prepacker and provides methods to access prepacking information.

I also tried to remove the accesses to the global scope from within the prepacker to isolate it from the rest of VTR. This is in an attempt to make it easier to work with.

### DIFF
--- a/vpr/src/pack/prepack.h
+++ b/vpr/src/pack/prepack.h
@@ -6,16 +6,175 @@
 
 #ifndef PREPACK_H
 #define PREPACK_H
+
+#include <algorithm>
+#include <map>
 #include <unordered_map>
-#include "atom_netlist_fwd.h"
-#include "arch_types.h"
 #include "vpr_types.h"
+#include "vtr_range.h"
 
-std::vector<t_pack_patterns> alloc_and_load_pack_patterns();
-void free_list_of_pack_patterns(std::vector<t_pack_patterns>& list_of_pack_patterns);
-void free_pack_pattern(t_pack_patterns* pack_pattern);
+class AtomNetlist;
+class AtomBlockId;
+struct t_molecule_stats;
+struct t_logical_block_type;
 
-t_pack_molecule* alloc_and_load_pack_molecules(t_pack_patterns* list_of_pack_patterns,
-                                               std::unordered_map<AtomBlockId, t_pb_graph_node*>& expected_lowest_cost_pb_gnode,
-                                               const int num_packing_patterns);
+/**
+ * @brief Class that performs prepacking.
+ *
+ * This class maintains the prepacking state, allowing the use of molecules
+ * (prepacked atoms) while this object exists.
+ *
+ * To use the prepacker, call the init method with a complete atom netlist.
+ * Then maintain this object (do not reset or destroy it) so long as the molecules
+ * are needed:
+ *
+ * // Initialize device and atom netlist
+ * // ...
+ * Prepacker prepacker;
+ * prepacker.init(atom_ctx.nlist, device_ctx.logical_block_types);
+ * // ...
+ * // Use the prepacked molecules.
+ * // ...
+ * prepacker.reset(); // Or if the prepacker object is destroyed.
+ * // Prepacked molecules can no longer be used beyond this point.
+ *
+ */
+class Prepacker {
+    using atom_molecules_const_iterator = std::multimap<AtomBlockId, t_pack_molecule*>::const_iterator;
+public:
+    // The constructor is default, the init method performs prepacking.
+    Prepacker() = default;
+
+    // This class maintains pointers to internal data structures, and as such
+    // should not be copied or moved (prevents unsafe accesses).
+    Prepacker(const Prepacker&) = delete;
+    Prepacker& operator=(const Prepacker&) = delete;
+
+    /**
+     * @brief Performs prepacking.
+     *
+     * Initializes the prepacker by performing prepacking and allocating the
+     * necessary data strucutres.
+     *
+     *  @param atom_nlist           The atom netlist to prepack.
+     *  @param logical_block_types  A list of the logical block types on the device.
+     */
+    void init(const AtomNetlist& atom_nlist, const std::vector<t_logical_block_type> &logical_block_types);
+
+    /**
+     * @brief Get the cluster molecules containing the given atom block.
+     *
+     *  @param blk_id       The atom block to get the molecules of.
+     */
+    inline vtr::Range<atom_molecules_const_iterator> get_atom_molecules(AtomBlockId blk_id) const {
+        auto rng = atom_molecules.equal_range(blk_id);
+        return vtr::make_range(rng.first, rng.second);
+    }
+
+    /**
+     * @brief Get the expected lowest cost physical block graph node for the
+     *        given atom block.
+     *
+     *  @param blk_id       The atom block to get the pb graph node of.
+     */
+    inline t_pb_graph_node* get_expected_lowest_cost_pb_gnode(AtomBlockId blk_id) const {
+        auto iter = expected_lowest_cost_pb_gnode.find(blk_id);
+        VTR_ASSERT(iter != expected_lowest_cost_pb_gnode.end());
+        t_pb_graph_node* pb_gnode = iter->second;
+        VTR_ASSERT(pb_gnode != nullptr);
+        return pb_gnode;
+    }
+
+    /**
+     * @brief Returns the total number of molecules in the prepacker.
+     */
+    inline size_t get_num_molecules() const {
+        size_t num_molecules = 0;
+        t_pack_molecule* molecule_head = list_of_pack_molecules;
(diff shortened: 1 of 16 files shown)


vpr/src/pack/prepack.h
```diff
@@ -6,16 +6,175 @@
...
+class AtomNetlist;
+class AtomBlockId;
+struct t_molecule_stats;
+struct t_logical_block_type;
+
+/**
+ * @brief Class that performs prepacking.
+ *
+ * This class maintains the prepacking state, allowing the use of molecules
+ * (prepacked atoms) while this object exists.
```
vaughnbetz, line 25: Say molecules currently come from pack patterns in the architecture file, and give an example of a molecule, e.g. a 3-bit carry chain in most architectures would turn into a molecule containing the 3 atoms forming that carry chain.

vaughnbetz: Also comment on whether every atom is part of a molecule or not -- I think they all are, where atoms that are not part of larger pre-pack structures like carry chains still form a single-atom molecule.

vpr/src/pack/prepack.h
```diff
@@ -6,16 +6,175 @@
...
+        std::vector<t_pack_molecule*> molecules;
+        t_pack_molecule* molecule_head = list_of_pack_molecules;
+        for (auto cur_molecule = molecule_head; cur_molecule != nullptr; cur_molecule = cur_molecule->next) {
+            molecules.push_back(cur_molecule);
+        }
+        return molecules;
+    }
+
+    /**
+     * @brief Marks all of the molecules as valid.
```
vaughnbetz, line 113: If you know what valid means in this context you could define it here.

vpr/src/pack/prepack.h
```diff
@@ -6,16 +6,175 @@
...
+        }
+    }
+
+    /**
+     * @brief Calculates maximum molecule statistics accross all molecules,
+     */
+    t_molecule_stats calc_max_molecule_stats(const AtomNetlist& netlist) const;
+
+    /**
+     * @brief Gets the largest number of blocks that any molecule has.
```
vaughnbetz, line 128: blocks that any molecule has -> blocks (atoms) that any molecule contains

vpr/src/pack/prepack.h
```diff
@@ -6,16 +6,175 @@
...
+
+private:
+    /**
+     * @brief A linked list of all the packing molecules that are loaded in
+     *        prepacking stage.
+     *
+     * Is is useful in freeing the pack molecules at the destructor of the Atom
+     * context using free_pack_molecules.
+     *
+     * TODO: Instead of pointers, we should use some form of ID.
```
vaughnbetz, line 158: TODO: should also probably use a vector instead of a linked list for storage (which will give you your ids, so maybe that's already what you mean here too).

vpr/src/timing/PreClusterDelayCalculator.h
```diff
@@ -165,7 +163,7 @@ class PreClusterDelayCalculator : public tatum::DelayCalculator {
     const AtomNetlist& netlist_;
     const AtomLookup& netlist_lookup_;
     const float inter_cluster_net_delay_;
-    const std::unordered_map<AtomBlockId, t_pb_graph_node*> block_to_pb_gnode_;
+    const Prepacker& prepacker_;
```
vaughnbetz, line 166: Tatum is a git submodule so this should be pushed back to its repo too when you're done.

vaughnbetz (approved):

vaughnbetz:
Looks very good; just a few commenting changes. 
I suggest getting a QoR comparison for safety too though (VTR and/or Titan).

vpr/src/pack/prepack.h
AlexandreSinger: Sounds good. I will update the doc. I think you are correct that every atom block will be part of a molecule, that is at least what I am seeing in AP and what the code is suggesting.

There is one confusing aspect I could not document: Based on the API, it is possible for a block to be within multiple molecules (hence `get_atom_molecules` returns a range). I have yet to find out when this happens. In AP I just put an assert to check that the number of molecules per block is 1, and on simple architectures this assert is never hit (I have not tried complicated architectures/circuits yet). Do you happen to know when this case occurs? I can always update this part of the documentation (or force it to always be 1) later.

vpr/src/timing/PreClusterDelayCalculator.h
AlexandreSinger: You are correct, I thought the same exact thing; however, this is actually still part of VTR. Its in src/timing. This code actually inherits from the Tatum DelayCalculator to prevent changing upstream Tatum. This was actually very well done, whoever made this! So I do not think Tatum needs to be updated in this case.

vpr/src/pack/prepack.h
AlexandreSinger: That was the intention; but I will make the comment more clear!

vpr/src/pack/prepack.h
AlexandreSinger: I'll give it a look; honestly I am not sure, and this is the only method that mutates the prepacker (all others are read only, other than init obviously). I am also curious.

vpr/src/pack/prepack.h
AlexandreSinger: The code was very hard to understand, but I think the blocks being part of multiple molecules has something to do with carry chains.

vpr/src/timing/PreClusterDelayCalculator.h
vaughnbetz: Ah, got it!  Kevin Murray created Tatum, and you're right the delay calculators were intentionally left as callbacks into client code.

vpr/src/pack/prepack.h
vaughnbetz: It could be used for things like a carry chain plus LUT/FF.  Say you have a 3-bit carry chain where every bit also drives a FF and the pack pattern say pack LUT/add->FF connections.  Now you could in theory have 4 molecules (partially overlapping) where each carry bit is part of two of them.

LUT_wcarry -> LUT_wcarry -> LUT_wcarry
  |                         |                       |
 FF                      FF                     FF

But I think early code merges those pack patterns into one larger molecule so in the end this is likely a single 6-atom molecule anyway, except perhaps during molecule creation?

AlexandreSinger: AH! That makes a lot of sense! Especially if this is only used internally. The code that makes use of the range was super odd to me. This makes sense now. What I will do is try to add an assert and see if the late-stage of prepacking is cleaning it up; and if so I will make the accessor return a single molecule instead. It will just make the API easier to work with (and make more sense).

AlexandreSinger: If not, I will document this in the API.

AlexandreSinger:
@vaughnbetz I have resolved your comments. I have found that, at least for the strong tests, each block will be contained in only one molecule. This matches what I think I see in the prepacker code. I have updated the API interface to reflect this and made it much more efficient by turning it from a multimap into a vtr::vector. I have added asserts to ensure this assumption. We will see if the CI agrees with this change, I can easily undo it if I am wrong.

I will run the VTR benchmarks soon to get QoR results, I do not expect any slowdown or QoR change.

AlexandreSinger:
@vaughnbetz Here are the QoR results for VTR:

  | vtr_parse_results_new_yosys.txt | vtr_parse_results_prepack.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.995878184
odin_synth_time |   |  
parmys_synth_time | 1 | 0.910618505
abc_depth | 1 | 1
abc_synth_time | 1 | 1.012554285
num_clb | 1 | 1
num_memories | 1 | 1
num_mult | 1 | 1
max_vpr_mem | 1 | 0.994725417
num_pre_packed_blocks | 1 | 1
device_grid_tiles | 1 | 1
pack_time | 1 | 0.999367909
placed_wirelength_est | 1 | 1
place_time | 1 | 1.015163644
placed_CPD_est | 1 | 1
min_chan_width | 1 | 1
routed_wirelength | 1 | 1
min_chan_width_route_time | 1 | 1.005690237
critical_path_delay | 1 | 1
crit_path_route_time | 1 | 0.990201513



The change in parmys_synth_time seems to be caused by smaller circuits (such as `ch_intrinsics.v`, `diffeq1.v`, and `raygentop.v`) being around 2x faster and bringing down the average. Everything else seems to be good. This makes sense. Although I made getting the molecule for a block O(1), the run time of the packing algorithm is dominated by the legalization of packing a molecule into a cluster (not getting the molecule).

When this passes CI, it should be good to merge unless you have any further comments.

vaughnbetz:
Looks good to me!

AlexandreSinger:
@vaughnbetz Please merge when you get the chance! Its a bit of a large PR and I want to prevent it from regressing. My next PR will also depend on this one.

PR #2711, opened 2024-09-03
Update Netlist Creation Function Documentation to Prevent Duplicates

### LINKED ISSUE #2690: [Netlist] Potential Issue When Re-Creating Elements in Atom and Clustered Netlists

ZohairZaidi:
#### Description

Updated the documentation for the AtomNetlist and ClusteredNetlist classes to reflect the new behavior of the create_* functions. The documentation now specifies that these functions will return the existing ID if an element (block, port, pin, or net) with the specified name or ID already exists, instead of creating a duplicate
#### Related Issue

closes #2690 
#### Motivation and Context

The update was necessary to avoid performance issues caused by unnecessary duplicate creations in the netlist
#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/atom_netlist.h
+++ b/vpr/src/base/atom_netlist.h
@@ -165,7 +165,9 @@ class AtomNetlist : public Netlist<AtomBlockId, AtomPortId, AtomPinId, AtomNetId
      */
 
     /**
-     * @brief Create or return an existing block in the netlist
+     * @brief Create a new block in the netlist or returns an existing block.
+              If a block with the specified name already exists, its associated data
+              (model and truth table) will be overwritten with the provided data.
      *
      *   @param name          The unique name of the block
      *   @param model         The primitive type of the block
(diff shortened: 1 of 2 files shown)


ZohairZaidi:
@AlexandreSinger, could you please review the changes to ensure they meet your expectations? Let me know if there's anything else I should adjust.

ZohairZaidi:
My apologies, fixed!

AlexandreSinger:
That looks better to me! Do not worry about the CI checks not passing for the NightlyTests; the CI server for those tests is down. Thank you!

vaughnbetz:
Thanks!

Issue #2712, opened 2024-09-04
VTR throws warning for shorted SB connections


WhiteNinjaZ:
When using switch override to create a short between two wires in a switch block and if the short is the only connection between those two wires (as in the creation of an L shaped wire) VTR places a connection between the wires, but throws a warning that the connection does not exist since the current code only checks for the existence of a switch between two wire types--not a short. An example of the warning thrown can be found bellow:
```
Warning 61: Node: 627691 with RR_type: CHANX  at Location:CHANX:627691 1len2D_x length:2 (2,1,0-> (1,1,0), had no out-going switches
```

#### Expected Behaviour

VTR should recognize when a short is being used between two wires in a switch block and should not throw a warning that a connection does not exist.

#### Possible Solution

A quick look at line 552 of [alloc_and_load_rr_indexed_data.cpp](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/c7b9ce08de4b66221a04a10f9cf00c898fbc477d/libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp#L552), where this warning comes from, shows that this warning is thrown when no switches exist between two wire types. This should be changed to only throw an error when the number of switches and number of shorts between two wires in a SB are 0.
```
if (num_switches == 0) {
            VTR_LOG_WARN("Node: %d with RR_type: %s  at Location:%s, had no out-going switches\n", rr_id,
                         rr_graph.node_type_string(rr_id), node_cords.c_str());
            continue;
}
```

#### Steps to Reproduce/Context
[The 7-series architecture capture](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2301) creates L shaped wires by shorting two different wire types together. In keeping with the true character of an L shaped wires the junction between these two wire types has only 1 shorted connection. The architecture is built around a channel width of 190 and when that width is maintained there should be no dangling connections in the architecture. When the 7-series arch is run with a channel width of 190 (on the stereovision3 benchmark for instance) many warnings occur for shorted connections between diagonal wires (see vpr.out). The RR graph has been checked and connections between the wires that VTR is complaining about do exist, so it is not a problem with the architecture but rather a problem with the current error checking in VTR.

PR #2713, opened 2024-09-04
fixed issue with warning being thrown for short switch types

WhiteNinjaZ:
This PR resolves issue #2712. We have had this fix on our branch of VTR for a long time, but we forgot to merge it into the master branch.

#### Description

This fix simply changes the if statement checking for connections between wires to include short connections. 

We have been running this change on our branch of VTR for a long time and on many designs with the Xilinx architecture.

#### Types of changes

#### Checklist:


vaughnbetz:
@saaramahmoudi : nag, nag ... we should land this since Peter would like to build on it.  Launching CI.

libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp
```diff
@@ -638,20 +641,24 @@ static void load_rr_indexed_data_T_values(const RRGraphView& rr_graph,
...
     for (const auto& edge : fan_in_list[node]) {
         /* want to get C/R/Tdel/Cinternal of switches that connect this track segment to other track segments */
         if (rr_graph.node_type(node) == CHANX || rr_graph.node_type(node) == CHANY) {
             int switch_index = rr_graph.rr_nodes().edge_switch(edge);
 
-            if (rr_graph.rr_switch_inf(RRSwitchId(switch_index)).type() == SwitchType::SHORT) continue;
+            if (rr_graph.rr_switch_inf(RRSwitchId(switch_index)).type() == SwitchType::SHORT) {
+                num_shorts++;
+                continue;
+            }
```
vaughnbetz, line 661: With this code change, it seems I could get divide by 0 errors late in the routine if a wire is only driven by a short (e.g. it might be the second part of an L-shaped wire).  CI seems to have errors, so I think there's a bug here.

vaughnbetz (changes requested):

vaughnbetz:
Relaunching CI as the old logs seem to have been deleted (too old probably and got cleaned up).

vaughnbetz:
@WhiteNinjaZ : Can you check for the divide by zero issue above?

vaughnbetz:
Nag, nag ...

libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp
WhiteNinjaZ: I believe this issue was caused by one of the VTR asserts not accounting for a wire that only has shorts. There shouldn't be a divide by zero error since some architectures have wires that are not driven on the edges. I am re-running the reg-tests with the VTR_asert change.

WhiteNinjaZ:
Thanks for the reminder. This PR fell a little under the radar. Made some changes and am re-running reg-tests locally.

vaughnbetz (approved):
LGTM; thanks!

vaughnbetz (approved):

PR #2721, opened 2024-09-12
Update Wire Cost Map

amin1377:
#### Description
Changing the index order in the router lookahead 6D lookup for hopefully improved cache locality on 3D archs.

Router Lookahead

Change wire delay map (6D array) from [from_layer][from_chan][from_seg][to_layer][dx][dy] → [from_layer][to_layer][from_chan][from_seg][dx][dy] to have a better cache locality.


vaughnbetz:
Should definitely do a QoR test to check this isn't slower (hopefully faster).  Need one 2D and one 3D QoR test.

amin1377:
Titan - 2D
image

PR #2725, opened 2024-09-17
[ClusterLegalizer] Fixed Memory Leak

AlexandreSinger:
A memory leak was able to occur if the external pin utilization string was set to an invalid number. This was caught by the CI, which inexplicably came back to life today.


AlexandreSinger:
@vaughnbetz Against all odds the CI persisted. The CI is back up and running. From one of the successful CI runs, I found that my PR introduced a memory leak into the code in a very specific case (it was actually that weird comment I had the in the constructor). My next PR on the ClusterLegalizer will actually remove all allocations from the constructor which should fix this issue permanently. For now this is a hot fix. Please review and merge as soon as it passes!

AlexandreSinger:
@vaughnbetz This has passed CI. Please review and merge!

PR #2728, opened 2024-09-17
[ClusterLegalizer] Cleaned Up Cluster Placement Stats

AlexandreSinger:
Removed external access to the cluster placement stats.

Made the cluster placement info private for each cluster. The cluster placement info of each cluster was being shared per cluster type. This caused issues when two clusters were being created at the same time.


vpr/src/pack/cluster_legalizer.h
```diff
@@ -377,6 +383,18 @@ class ClusterLegalizer {
...
     void reset();
 
+    /*
+     * @brief Checks if the given molecule is compatible with the given cluster.
+     *
+     * A molecule is compatible with a cluster if there exists "free primitives"
+     * (primitives that are not currently occupied by other atoms) for each atom
+     * in the molecule.
+     *
+     * This is a quick check to see if a molecule can go in the given cluster.
```
vaughnbetz, line 393: I'd expand this a bit ... I'd add:

"This is a necessary but not sufficient test for a molecule to be able to go in a cluster. By calling it you can save runtime for impossible cases vs. calling the full checks."

Also change to "if there exists a free primitive (a primitive that is not currently occupied by other atoms) of the correct type to accomodate each type of atom in the molecule"

vpr/src/pack/cluster_legalizer.h
```diff
@@ -91,6 +92,11 @@ struct LegalizationCluster {
     ///        Contains information about the atoms in the cluster and how they
     ///        can be routed within.
     t_lb_router_data* router_data;
+
+    /// @brief The stats on where the different atoms in the cluster are currently
+    ///        placed in the cluster. This is what contains and decides which
```
vaughnbetz, line 97: "this is what contains ..." -> "This is used when the legalizer decides what sites it should try to put a new molecule into."

vpr/src/pack/cluster_placement.h
```diff
@@ -5,23 +5,142 @@
...
-#include "arch_types.h"
+
+#include <vector>
+#include <unordered_map>
+#include "physical_types.h"
 #include "vpr_types.h"
 
-t_cluster_placement_stats* alloc_and_load_cluster_placement_stats();
+/**
+ * @brief Stats keeper for placement information during packing
```
vaughnbetz, line 15: placement information -> placement within the cluster during packing

vpr/src/pack/cluster_placement.h
```diff
@@ -5,23 +5,142 @@
...
+
+    /**
+     * @brief iterate over elements of a queue and move its elements to valid_primitives
+     *
+     * @param queue the unordered_multimap to work on (e.g. in_flight, tried, or invalid)
+     */
+    void flush_queue(std::unordered_multimap<int, t_cluster_placement_primitive*>& queue);
+};
+
+t_cluster_placement_stats* alloc_and_load_cluster_placement_stats(t_logical_block_type_ptr cluster_type,
```
vaughnbetz, line 128: Move .cpp doxygen to the .h

vaughnbetz (changes requested):

AlexandreSinger:
VTR Results:

  | vtr_parse_results_legalizer.txt | parse_results_new_legalizer.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.021650042
odin_synth_time |   |  
parmys_synth_time | 1 | 0.999246531
abc_depth | 1 | 1
abc_synth_time | 1 | 1.006581805
num_clb | 1 | 0.999327261
num_memories | 1 | 1
num_mult | 1 | 1
max_vpr_mem | 1 | 1.009962447
num_pre_packed_blocks | 1 | 1
num_post_packed_blocks | 1 | 0.999793597
device_grid_tiles | 1 | 0.999267664
pack_time | 1 | 1.071846971
placed_wirelength_est | 1 | 1.000075079
place_time | 1 | 0.978678405
placed_CPD_est | 1 | 0.999502821
min_chan_width | 1 | 0.998666868
routed_wirelength | 1 | 0.992543873
min_chan_width_route_time | 1 | 1.064750414
crit_path_routed_wirelength | 1 | 0.991811581
critical_path_delay | 1 | 1.006753785
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 1.007078544
crit_path_route_time | 1 | 0.967975153

As expected, the runtime of the packer increased slightly due to this change. This is because a data structure is now being created per cluster instead of per logic block type. Will run Titan to ensure the runtime increase is not too large.

Raw results:
[vtr_old_v_new.xlsx](https://github.com/user-attachments/files/17037278/vtr_old_v_new.xlsx)

AlexandreSinger:
Titan results:

  | tan_parse_results_legalizer.txt | parse_results_new_legalizer.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.047841819
num_LAB | 1 | 0.999095457
num_DSP | 1 | 1
num_M9K | 1 | 1
num_M144K | 1 | 1
max_vpr_mem | 1 | 1.015864127
num_pre_packed_blocks | 1 | 1
num_post_packed_blocks | 1 | 0.99914983
device_grid_tiles | 1 | 0.999284855
pack_time | 1 | 1.079054892
placed_wirelength_est | 1 | 0.995888122
place_time | 1 | 1.067726331
placed_CPD_est | 1 | 0.992960886
routed_wirelength | 1 | 0.996480931
critical_path_delay | 1 | 0.981249309
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 0.996111084
crit_path_route_time | 1 | 0.991888639


Pack time seems to have gone up slightly (8%) but quality seems to be a win across the board interestingly enough.

Here are the raw results:
[titan_legalizer_comparison.xlsx](https://github.com/user-attachments/files/17050349/titan_legalizer_comparison.xlsx)

I will update your comments @vaughnbetz and update the golden results.

AlexandreSinger:
@vaughnbetz Actually the Titan result is quite interesting. This should not really have a huge quality difference; yet CPD went down by 2%. I think this may show how much the packing quality affects QoR.

vaughnbetz:
Interesting!  I wonder what is different. In any case I'm fine with merging this.  I suggest rebasing / updating the branch and kicking off CI again. 
Still worth looking at the cluster placement stats data structure, know that you know updates to it consume over 8% of placement time, but it isn't gating for this PR.

AlexandreSinger:
@vaughnbetz This is now passing CI and I think it is ready to merge.

I had to disable one of the figure_8 tasks in the odin nightly test 1, since it had an issue where it would find a low minimum channel width, then use a higher one to route; but the higher one was unroutable (I wish there was some way to detect that):
image

I also increased the minimum channel width for verify rr graph for odin reg strong. I just made it match the regular reg strong.
image

Both of these were in the Odin tests, so I do not think there is much harm in changing them. The other golden results were updated using the standard method.

PR #2733, opened 2024-09-20
[AP] Created The APNetlist Class

AlexandreSinger:
The APNetlist class holds the blocks and nets used during the AP flow. In the AP context, a block is a collection of primitives which would want to be placed in the same location (i.e. they want to move together). Currently, each block represents a pack molecule (created by the Prepacker). The nets are logical connections between blocks. The APNetlist will not contain any nets which are not needed for AP.

The APNetlist is tested using a Unit Test which tests its basic functionality (features that differ from the base Netlist class).

This is the first of many commits which will bring Analytical Placement into VPR.

### DIFF
--- a/vpr/src/analytical_place/ap_netlist.h
+++ b/vpr/src/analytical_place/ap_netlist.h
@@ -1,0 +1,188 @@
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    September 2024
+ * @brief   Defines the APNetlist class used to store the connectivity of
+ *          primitives in the Analytical Placement context.
+ *
+ * In the context of Analytical Placement, a block is a collection of atoms
+ * (primitives) which want to move together. For example, if one atom in a block
+ * should be moved one unit to the left, all atoms in the block want to move one
+ * unit to the left.
+ *
+ * An example of a block is pack molecules, which are atoms which were prepacked
+ * together.
+ *
+ * The nets in the netlist represent the logical connections between atom
+ * blocks, where nets which are unused by Analytical Placement are ignored.
+ */
+
+#pragma once
+
+#include <string>
+#include "netlist.h"
+#include "ap_netlist_fwd.h"
+
+// Forward declarations
+class t_pack_molecule;
+
+/**
+ * @brief Struct to store fixed block location information
+ *
+ * Currently assumes that blocks are fixed to single locations (not ranges).
+ * TODO: This assumption could be relaxed and allow fixing a range of locations.
+ *
+ * -1 implies that the block is not fixed in that dimension.
+ */
+struct APFixedBlockLoc {
+    int x = -1;
+    int y = -1;
+    int layer_num = -1;
+    int sub_tile = -1;
+};
+
+/**
+ * @brief The type of a block in the APNetlist
+ * TODO: It would be nice if the netlist contained lists of moveable and fixed
+ *       block ids.
+ */
+enum class APBlockType : bool {
+    MOVEABLE,   // The block is not constrained in any dimension.
+    FIXED       // The block is fixed.
+};
+
+/**
+ * @brief The netlist used during Analytical Placement
+ *
+ * This class abstracts the placeable blocks and connections between the blocks
+ * away from the atom netlist. An APBlock is assumed to be some collection of
+ * primitive blocks and an APNet is assumed to be some connection between the
+ * APBlocks. These need not have physical meaning.
+ */
+class APNetlist : public Netlist<APBlockId, APPortId, APPinId, APNetId> {
+public:
+    /**
+     * @brief Constructs a netlist
+     *
+     *  @param name The name of the netlist (e.g. top-level module)
+     *  @param id   A unique identifier for the netlist (e.g. a secure digest of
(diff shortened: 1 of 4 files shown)


AlexandreSinger:
@vaughnbetz This PR is the first of many. It brings in the APNetlist into VPR. It also begins the new `analytical_place` folder. Do let me know what you think about the name. I have also added a unit test to lock in this feature. Disregard the failing testcase, this is a bad golden result which I resolved in another PR. Please review!

vpr/src/analytical_place/ap_netlist.h
```diff
@@ -0,0 +1,188 @@
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    September 2024
+ * @brief   Defines the APNetlist class used to store the connectivity of
+ *          primitives in the Analytical Placement context.
+ *
+ * In the context of Analytical Placement, a block is a collection of atoms
```
vaughnbetz, line 8: Will a block ever represent more than one cluster (i.e. grid location) for long carry chains etc.?  If not known we can be silent on this for now.

vpr/src/analytical_place/ap_netlist.h
```diff
@@ -0,0 +1,188 @@
...
+ * (primitives) which want to move together. For example, if one atom in a block
+ * should be moved one unit to the left, all atoms in the block want to move one
+ * unit to the left.
+ *
+ * An example of a block is pack molecules, which are atoms which were prepacked
+ * together.
+ *
+ * The nets in the netlist represent the logical connections between atom
```
vaughnbetz, line 16: between AP blocks (inferred from the atom block connectivity)

vpr/src/analytical_place/ap_netlist.h
```diff
@@ -0,0 +1,188 @@
...
+public: // Public Accessors
+    /*
+     * Blocks
+     */
+
+    /// @brief Returns the molecule that this block represents.
+    const t_pack_molecule* block_molecule(const APBlockId id) const;
+
+    /// @brief Returns the type of this block.
+    APBlockType block_type(const APBlockId id) const;
```
vaughnbetz, line 85: APBlockMobility (and block_mobility) might be better names.

vpr/src/analytical_place/ap_netlist.h
```diff
@@ -0,0 +1,188 @@
...
+     * Sanity checks
+     */
+    // Verify the internal data structure sizes match
+    bool validate_block_sizes_impl(size_t num_blocks) const override;
+    bool validate_port_sizes_impl(size_t num_ports) const override;
+    bool validate_pin_sizes_impl(size_t num_pins) const override;
+    bool validate_net_sizes_impl(size_t num_nets) const override;
+
+private: // Private Data
+    /// @brief Molecule of each block
```
vaughnbetz, line 180: I like adding a notation like this: [0 .. APblocks.size()-1]
Your call if you hate it ...

vpr/src/analytical_place/ap_netlist.h
```diff
@@ -0,0 +1,188 @@
...
+    vtr::vector_map<APBlockId, const t_pack_molecule*> block_molecules_;
+    /// @brief Type of each block
+    vtr::vector_map<APBlockId, APBlockType> block_types_;
```
vaughnbetz, line 183: Could infer the mobility of each block from block_locs_ with a member function.

vaughnbetz: Could also make an enum that is part of the APFixedBlockLoc to represent mobility status.

vaughnbetz (approved):
Some comments -- use your judgement of what to change.
I would change the APBlockType name to APBlockMobility

vpr/src/analytical_place/ap_netlist.h
AlexandreSinger: This is not known. My gut feeling is to make it part of the block; but I would need to do more fiddling around.

vpr/src/analytical_place/ap_netlist.h
AlexandreSinger: Sorry, I really do not like it. I personally find the notation confusing and the comment makes it clear that an entry exists for each block. That notation implies that you can index the array with a `size_t`; but in reality it acts more like a map between the APBlockId and the data members (where evert APBlockId is guaranteed to have an entry).

vpr/src/analytical_place/ap_netlist.h
AlexandreSinger: Looking through my future PRs on AP; the mobility is queried much more often than the fixed block locations. Having that as a quick (cache local) lookup is more valuable to me in my opinion.

vpr/src/analytical_place/ap_netlist.h
vaughnbetz: OK, your call.

AlexandreSinger:
@vaughnbetz This has passed all of the regular CI tests. The NightlyTests seem to be failing again; however, I think this PR can still be merged since it does not touch any of the code in NightlyTests.

PR #2734, opened 2024-09-21
[CI] Fixed Bad Golden Results

AlexandreSinger:
The golden results for one of the tests in vtr_reg_nightly_test3 was probably generated when the machine was more loaded than normal and was significantly slower than the CI. This caused a testcase to fail sometimes. This fixes the golden result to prevent the test from failing.


AlexandreSinger:
@vaughnbetz I took the liberty of merging this in myself. Looks like a couple of testcases were close to the edge and were failing intermittently. I updated the golden solutions to prevent them from failing.

vaughnbetz:
Thanks!

PR #2735, opened 2024-09-21
Trying to fix the failing unit test

soheilshahrouz:
I couldn't reproduce the unit test failure on a local machine, so I created this PR to trigger a CI run and hopefully locate the segault.

### DIFF
--- a/libs/libvtrutil/src/vtr_vector.h
+++ b/libs/libvtrutil/src/vtr_vector.h
@@ -56,9 +56,6 @@ class vector : private std::vector<V, Allocator> {
     class key_iterator;
     typedef vtr::Range<key_iterator> key_range;
 
-    class pair_iterator;
-    typedef vtr::Range<pair_iterator> pair_range;
-
   public:
     //Pass through std::vector's types
     using typename storage::allocator_type;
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
@soheilshahrouz A suggestion which might make things easier is to increase the assert level for the unit test for your test and turning on the sanitizer so you can try and catch the bug in action.

Locally you can try building with:
`CMAKE_PARAMS="-DVTR_ASSERT_LEVEL=3 -DVTR_ENABLE_SANITIZE=ON"`
I set the assert level to 3, but you can increase it to 4 if you really want. I think the sanitizer will be the most help. It will slow things down tremendously, but since this is a unit test it should be pretty quick.

libs/libvtrutil/src/vtr_vector.h
```diff
@@ -157,16 +156,31 @@ class vector : private std::vector<V, Allocator> {
...
      *
-     * This function returns a range object that represents the sequence of key-value
-     * pairs within the vector. The range can be used to iterate over the pairs using
-     * standard range-based loops or algorithms.
+     * This function allows for easy enumeration, yielding a tuple of (index, element)
+     * pairs in each iteration. It is similar in functionality to Python's `enumerate()`.
      *
-     * @return A `pair_range` object representing the range of key-value pairs.
+     * @ return An iterable wrapper that can be used in a range-based for loop to obtain
+     * (index, element) pairs.
```
vaughnbetz, line 165: I think it would be a good idea to give a short example of how to use this iterator (write the outline of a range-based for and extracting keys & values in the loop body).

vaughnbetz (commented):
Suggest adding an example ...

AlexandreSinger (approved):
Hi, @soheilshahrouz , these changes LGTM.

My only comment is that I actually do not like the `pairs()` interface for a vector-like storage class. A vector is basically a contiguous array of values, where the index is not necessarily a part of the structure itself. The issue I see here is that you are generating the `key` of the pair when the iteration is happening; the key is not stored in the data structure directly. This is not going to be returned by reference (I think), or at least it really should not be since this is a reference to a temporary. I am not sure the performance impact of this. In my opinion the best way to iterate over a vector of values when you need the index as well as the value is to use a standard for loop.

Since this interface is optional to use though, I see no issue with adding it. But I do think it should be used sparingly (unless someone wants to dig into the assembly to see how this is being lowered lol).

soheilshahrouz:
@vaughnbetz
Ready to be merged into master.

Issue #1834, opened 2021-08-27
Add Doxygen-style comments to RRGraphView and RRGraphBuilder, and make new VPR API tab


vaughnbetz:
We have nice new rr-graph APIs, so we should auto-generate API documentation on them.  

#### Proposed Behaviour
New RRGraphView and RRGraphBuilder tabs on the vtr api tab in read the docs.  Also should add some high-level comments (if they aren't there already) explaining exactly when each of these should be used.

#### Current Behaviour
Good comments, but I don't think they're doxygen-compatible and we don't build docs on them.

#### Possible Solution
Reformat comments to be doxygen-compatible, and add to the docs page.

#### Context
The comments and use of these classes will be more understandable if they're on doxygen.

vaughnbetz:
Cleaning up my old issues!  @tangxifan : This all looks good, but it would be good if the RRGraphView class had a high level comment on what it is used for that showed up in the documentation.  RRGraphBuilder and RRGraphSpatialLookup already have those high level comments.

tangxifan:
@vaughnbetz No problem. I will work on this.

Issue #2361, opened 2023-08-03
Add some overview to the RRGraphView doxygen documentation


vaughnbetz:
#### Proposed Behaviour
We should give a high level overview of the RRGraphView data structure / API in our doxygen documentation. 
#### Current Behaviour
Currently we dive right into methods.

#### Possible Solution
Explain the RRGraph describes the programmable routing fabric of the FPGA (all wires and switches). RRGraphView provides a read-only interface, which tools like the router, graphics, and statistics need.
Should also explain that the fundamental data structures is a graph (nodes and out-going edges) with extra information like the location within the chip and electrical parameters added to help make algorithms more efficient, draw the chip, and estimate delays.

vaughnbetz:
Re assigning to @tangxifan . This might be a duplicate though!

vaughnbetz:
Is a duplicate of the remaining work in https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/1834 . We can close both together.

tangxifan:
@vaughnbetz Indeed. We may close this one and keep #1834 open. My team will follow up. A PR will happen this summer.

vaughnbetz:
This has been done; closing.

PR #2745, opened 2024-09-24
Add some overview to the RRGraphView doxygen documentation 

treelin611:
#### Description
- Add high level comments on RRGraphView class 
- Reformat comment to be doxygen-compatible

#### Related Issue
 - This Pull request is related to an existing issue #1834


vaughnbetz (approved):
Looks very good, thanks!

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -1,36 +1,42 @@
...
-/* An read-only routing resource graph
- * which is an unified object including pointors to
- * - node storage
- * - TODO: edge_storage
- * - TODO: node_ptc_storage
- * - TODO: node_fan_in_storage
- * - rr_node_indices
+/**
+ * @file
+ * @brief The RRGraphView encapsulates a read-only routing resource graph, providing clients with tailored frame views of the object.
```
vaughnbetz, line 6: I'm not quite sure what "providing clients with tailored frame views of the object." means.  I think this may be something that is hoped for in the future but isn't there now.  I suggest saying:
" ... a read-only routing resource graph as most clients (router, timing analyzer, etc.) only need to read a routing-resource graph."

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -1,36 +1,42 @@
...
- * - node storage
- * - TODO: edge_storage
- * - TODO: node_ptc_storage
- * - TODO: node_fan_in_storage
- * - rr_node_indices
+/**
+ * @file
+ * @brief The RRGraphView encapsulates a read-only routing resource graph, providing clients with tailored frame views of the object. 
+ * 
+ * The RRGraphView represents the full frame view of the routing resource graph, offering several advantages:
```
vaughnbetz, line 8: I think you'll have to explain what this means a bit more, or shorten/delete it.  Is this saying that as the RRGraphView allows all the RRGraph data to be viewed, it can be used by multiple clients, which saves memory?

treelin611: I just rewrote the overview section according to "possible solution" in this link: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2361. 
The final documentation will appear as follows: 
image

vaughnbetz: That looks great ... thanks!

treelin611: No problem! A minor bug has also been fixed. The PR should be ready now.

tangxifan: @treelin611 You may refer to the following text when adding the fundamental description 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/b9760a9665475c8b9211c289a88a8d258145f8f5/libs/librrgraph/src/base/rr_graph_obj.h#L14-L47

vaughnbetz: Lots of good information there n good hat link ... I suggest adding a pointer and/or directly moving some over.

treelin611: Thanks for the link. It is very informative. I have added some information from it into the current RRGraphView. The final documentation will be as follows:
image

vaughnbetz:
Looks great, thanks.  Will build once the basic CI passes (the google runners are down, so the bigger *_nightly* tests won't run but they aren't needed for this change).

PR #2746, opened 2024-09-24
[AP] Added AP Flow to VPR

AlexandreSinger:
This adds the AP flow to VPR, allowing AP to be called from the command line and how the input and output of the flow will interract with the rest of VPR.

Note: I was unable to add a testcase for this since VPR does not have the infrastructure in place to do an "expected fail" (from a full flow call like basic tests and strong tests). Eventually this may be a nice feature to have.


vpr/src/base/CheckSetup.cpp
```diff
@@ -72,6 +73,28 @@ void CheckSetup(const t_packer_opts& PackerOpts,
...
+            VPR_FATAL_ERROR(VPR_ERROR_OTHER,
+                            "Analytical placement should skip packing.\n");
+        }
+
+        // TODO: Should check that read_vpr_constraint_file is non-empty or
+        //       check within analytical placement that the floorplanning has
+        //       some fixed blocks somewhere. Maybe we can live without fixed
+        //       blocks.
+
+        // FIXME: Should we enforce that the size of the device is fixed? Or is
```
vaughnbetz, line 94: Device size isn't fixed in the constraints file; it's specified either in the arch file or can be specified on the command line.

vpr/src/base/SetupVPR.cpp
```diff
@@ -276,6 +279,12 @@ void SetupVPR(const t_options* Options,
             PlacerOpts->doPlacement = STAGE_DO;
         }
 
+        if (Options->do_analytical_placement) {
```
vaughnbetz, line 282: Maybe comment why: analytical placement performs both packing & placement with an integrated algorithm.

vpr/src/base/read_options.cpp
```diff
@@ -1324,6 +1324,11 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .action(argparse::Action::STORE_TRUE)
         .default_value("off");
 
+    stage_grp.add_argument<bool, ParseOnOff>(args.do_analytical_placement, "--analytical_place")
+        .help("Run analytical placement")
```
vaughnbetz, line 1328: I'd expand on this .help.  Should say analytical placement uses an integrated packing & placement algorithm.

(Reason to explain this: otherwise it would make more sense for us to just have another set of placement options, but not a different stage).

vpr/src/base/vpr_types.h
```diff
@@ -1143,6 +1143,26 @@ struct t_placer_opts {
...
+ *******************************************************************/
+
+/**
+ * @brief Various options for the Analytical Placer.
+ *
+ *   @param doAnalyticalPlacement
+ *              True if analytical placement is supposed to be done in the CAD
+ *              flow. False if otherwise.
+ */
+struct t_ap_opts {
```
vaughnbetz, line 1158: I know we've got lots of types in this file already, but I wonder if the t_ap_ops should go in the analytical_place directory (maybe ap_types.h) instead?  Don't worry about it if that would be a big refactor or be more trouble than it is worth due to being inconsistent with other opts.

vaughnbetz (commented):
Looks good.  
You should also make a documentation update to describe this new command line option (easiest to do now right when you're writing the options and .help) and mark it as experimental / under development.

vaughnbetz (approved):
Looks good; feel free to merge whenever CI looks OK to you.

vpr/src/base/vpr_types.h
AlexandreSinger: This struct is actually part of the VPR_Setup struct and is used to pass command-line arguments to the flows. I think it would make more sense if all of these structs were moved to a better place; separating them out to their respective flows seems confusing to me. I think its better to wait on this and do this all at once.

vaughnbetz:
Ok, thanks

PR #2749, opened 2024-09-26
[AP] Added Ability to Read The Atom Netlist

AlexandreSinger:
This method reads the atom netlist and uses the results of the prepacker to generate an APNetlist. It also fixes the block locations based on the user place constraints.

### DIFF
--- a/vpr/src/analytical_place/ap_netlist.cpp
+++ b/vpr/src/analytical_place/ap_netlist.cpp
@@ -58,9 +58,19 @@ APBlockId APNetlist::create_block(const std::string& name, const t_pack_molecule
 void APNetlist::set_block_loc(const APBlockId id, const APFixedBlockLoc& loc) {
     VTR_ASSERT_SAFE(valid_block_id(id));
 
-    // Check that the location is fixed, if all values are -1 then it is not fixed.
-    if (loc.x == -1 && loc.y == -1 && loc.sub_tile == -1 && loc.layer_num == -1)
+    // Check that the location is fixed; if all dims are unfixed then it is not fixed.
+    if (loc.x == APFixedBlockLoc::UNFIXED_DIM &&
+        loc.y == APFixedBlockLoc::UNFIXED_DIM &&
+        loc.sub_tile == APFixedBlockLoc::UNFIXED_DIM &&
+        loc.layer_num == APFixedBlockLoc::UNFIXED_DIM)
         return;
+
+    // Ensure that the block is fixed to a single position on the grid (x, y, layer).
+    // sub-tile is allowed to be unfixed.
+    VTR_ASSERT(loc.x != APFixedBlockLoc::UNFIXED_DIM &&
+               loc.y != APFixedBlockLoc::UNFIXED_DIM &&
+               loc.layer_num != APFixedBlockLoc::UNFIXED_DIM &&
+               "AP: Currently, AP assumes block is locked down to a single position on the device grid.");
 
     block_locs_[id] = loc;
     block_mobilities_[id] = APBlockMobility::FIXED;
--- a/vpr/src/analytical_place/analytical_placement_flow.cpp
+++ b/vpr/src/analytical_place/analytical_placement_flow.cpp
@@ -6,9 +6,12 @@
  */
 
 #include "analytical_placement_flow.h"
+#include "ap_netlist.h"
 #include "atom_netlist.h"
+#include "gen_ap_netlist_from_atoms.h"
 #include "globals.h"
 #include "prepack.h"
+#include "user_place_constraints.h"
 #include "vpr_context.h"
 #include "vpr_error.h"
 #include "vpr_types.h"
@@ -22,10 +25,17 @@ void run_analytical_placement_flow(t_vpr_setup& vpr_setup) {
     // The global state used/modified by this flow.
     const AtomNetlist& atom_nlist = g_vpr_ctx.atom().nlist;
     const DeviceContext& device_ctx = g_vpr_ctx.device();
+    const UserPlaceConstraints& constraints = g_vpr_ctx.floorplanning().constraints;
 
     // Run the prepacker
     Prepacker prepacker;
     prepacker.init(atom_nlist, device_ctx.logical_block_types);
+
+    // Create the ap netlist from the atom netlist using the result from the
+    // prepacker.
+    APNetlist ap_netlist = gen_ap_netlist_from_atoms(atom_nlist,
+                                                     prepacker,
+                                                     constraints);
 
     // AP is currently under-construction. Fail gracefully just in case this
     // is somehow being called.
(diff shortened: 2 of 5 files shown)


vpr/src/analytical_place/ap_netlist.cpp
```diff
@@ -58,8 +58,11 @@ APBlockId APNetlist::create_block(const std::string& name, const t_pack_molecule
 void APNetlist::set_block_loc(const APBlockId id, const APFixedBlockLoc& loc) {
     VTR_ASSERT_SAFE(valid_block_id(id));
 
-    // Check that the location is fixed, if all values are -1 then it is not fixed.
-    if (loc.x == -1 && loc.y == -1 && loc.sub_tile == -1 && loc.layer_num == -1)
+    // Check that the location is fixed, if all dims are unfixed then it is not fixed.
```
vaughnbetz, line 61: nit:  , should be a ;
(sorry for the nit, can't help myself!)

vpr/src/analytical_place/ap_netlist.cpp
```diff
@@ -58,8 +58,11 @@ APBlockId APNetlist::create_block(const std::string& name, const t_pack_molecule
...
+    if (loc.x == APFixedBlockLoc::UNFIXED_DIM &&
+        loc.y == APFixedBlockLoc::UNFIXED_DIM &&
+        loc.sub_tile == APFixedBlockLoc::UNFIXED_DIM &&
+        loc.layer_num == APFixedBlockLoc::UNFIXED_DIM)
         return;
 
     block_locs_[id] = loc;
```
vaughnbetz, line 68: Seems a bit odd to allow fixed locations with only some dimensions fixed (not sure how you're going to handle that; it would seem like a strange/limited form of region constraint that isn't worth supporting since we already have more general region support).  I suggest you assert that all dimensions are fixed if any are fixed; code can then assume FIXED means truly locked down.  You can generalize that later, but I'd do that by going through the partition/region constraints, not by allowing individual dimensions to be -1.

vpr/src/analytical_place/read_atom_netlist.cpp
```diff
@@ -0,0 +1,175 @@
...
+            if (!part_id.is_valid())
+                continue;
+            // We should not fix a block twice. This would imply that a molecule
+            // contains two fixed blocks. This would only make sense if the blocks
+            // were fixed to the same location. I am not sure if that is even
+            // possible.
+            VTR_ASSERT(ap_netlist.block_mobility(ap_blk_id) == APBlockMobility::MOVEABLE);
+            // Get the partition region.
+            const PartitionRegion& partition_pr = constraints.get_partition_pr(part_id);
+            VTR_ASSERT(partition_pr.get_regions().size() == 1 && "Each partition should contain only one region!");
```
vaughnbetz, line 83: Partitions can contain more than one region (we take the union as the legal locations).  If the analytic placer can't handle that you can assert for now (should handle or turn into a proper error later) but the assert message should mention this is an AP limitation.

vpr/src/analytical_place/read_atom_netlist.cpp
```diff
@@ -0,0 +1,175 @@
...
+            const Region& region = partition_pr.get_regions()[0];
+            // Get the x and y.
+            const vtr::Rect<int>& region_rect = region.get_rect();
+            VTR_ASSERT(region_rect.xmin() == region_rect.xmax() && "Expect each region to be a single point in x!");
```
vaughnbetz, line 87: Again, asserts should mention AP I think (can just say "AP: Expect each ..."

vpr/src/analytical_place/read_atom_netlist.cpp
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+        if (is_all_fixed) {
+            remove_net(ap_net_id);
+            continue;
+        }
+    }
+    ap_netlist.remove_and_compress();
+
+    // TODO: Some nets may go to the same block mutliple times. Should that be
```
vaughnbetz, line 157: I'd allow nets to go to the same block multiple times. It can happen (e.g. data and clear of a FF, sign extension, etc.) and the timing won't necessarily be the same (some could go to registers and some to internal combinational paths). So when we optimize timing having both connections is probably cleaner.

vpr/src/analytical_place/read_atom_netlist.cpp
```diff
@@ -0,0 +1,175 @@
...
+    //       fixed? Maybe we want nets to be strongly attracted towards some
+    //       blocks over others.
+
+    // TODO: Should we cleanup the blocks? For example if there is no path
```
vaughnbetz, line 161: Unless it makes the matrix ill-conditioned I'd leave them alone.  And if it makes the matrix ill-conditioned I'd probably find some way to place some of these in a different way and let the rest of the placement evolve around them.

vpr/src/analytical_place/read_atom_netlist.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "prepack.h"
+#include "region.h"
+#include "user_place_constraints.h"
+#include "vpr_types.h"
+#include "vtr_assert.h"
+#include "vtr_geometry.h"
+#include "vtr_time.h"
+#include <unordered_set>
+#include <vector>
+
```
vaughnbetz, line 25: read_atom_netlist is a bit of a confusing name. We normally use read_* for file input, and we do read a blif netlist already.  I think this is gen_APNetlist_from_atoms () or something like that or create_APNetlist_from_atom_netlist ()

vpr/src/analytical_place/ap_netlist.cpp
AlexandreSinger: ...omg you're right. I can't believe I messed that one up; they are two independent clauses! Good catch!

vpr/src/analytical_place/ap_netlist.cpp
AlexandreSinger: It actually is sufficient to just lock it down in the x-dimension, for example, since solvers tend to solve x independent to y. The layer_num is currently always assumed to be 0; but it was easy to add to this struct in this case (being forward looking).

The sub_tile actually need not be locked down (nor should it). The placement constraints file leaves this optional. This was the main reason UNFIXED_DIM exists.

However, since the generation of the APNetlist from the atoms never allows x, y, and layer to be unspecified, I have added an assert to this method to guard against them being unset. I have purposefully left sub_tile out of this assert.

vpr/src/analytical_place/read_atom_netlist.cpp
AlexandreSinger: It will make the matrix ill-conditioned, but Eigen does not have a problem with this. I do not want to lose the note though (since this may be a problem later if we change libraries). I have added a sentence explaining this.

vpr/src/analytical_place/read_atom_netlist.cpp
AlexandreSinger: Very good point. This was historical. Originally it read the atom netlist directly. I have since made it take the prepacked solution. I like the name `gen_ap_netlist_from_atoms`.

AlexandreSinger:
@vaughnbetz I think this is ready to be merged. Should have no effect on NightlyTests.

PR #2751, opened 2024-09-27
Couple fixes from SPEC CPU work

heshpdx:
Couple fixes from SPEC CPU work:
- Provide rr_graph_utils with full definition of lambda function. 
- Remove/replace __attribute since MSVC doesn't know about it.

### DIFF
--- a/libs/librrgraph/src/base/rr_graph_utils.cpp
+++ b/libs/librrgraph/src/base/rr_graph_utils.cpp
@@ -140,6 +140,7 @@ vtr::vector<RRNodeId, std::vector<RREdgeId>> get_fan_in_list(const RRGraphView& 
     //Walk the graph and increment fanin on all dwnstream nodes
     rr_graph.rr_nodes().for_each_edge(
         [&](RREdgeId edge, RRNodeId src, RRNodeId sink) -> void {
+            (void) src;
             node_fan_in_list[sink].push_back(edge);
         });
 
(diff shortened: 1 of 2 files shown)


libs/librrgraph/src/base/rr_graph_utils.cpp
```diff
@@ -139,7 +139,7 @@ vtr::vector<RRNodeId, std::vector<RREdgeId>> get_fan_in_list(const RRGraphView&
 
     //Walk the graph and increment fanin on all dwnstream nodes
     rr_graph.rr_nodes().for_each_edge(
-        [&](RREdgeId edge, __attribute__((unused)) RRNodeId src, RRNodeId sink) {
+        [&](RREdgeId edge, RRNodeId src, RRNodeId sink) -> void {
```
AlexandreSinger, line 142: I agree that `__attribute__((unused))` should not be used since it only works for some compilers; however `src` is still unused. We can just void cast it in the next line.
```
(void)src;
```

AlexandreSinger (commented):
Hi @MahshadJ , thank you for these changes.

Looking through the CI, this causes some of the regular tests to fail due to warnings. This is just from an unused variable which is easily resolved below.

libs/librrgraph/src/base/rr_graph_utils.cpp
heshpdx: Done! thank you

AlexandreSinger (approved):
LGTM

vaughnbetz:
Google part of CI is down, so nightly test failures are expected. There is one unit test failing, but it's a known problem test that @soheilshahrouz is going to turn off for now.

PR #2752, opened 2024-09-27
[AP] Full Legalizer

AlexandreSinger:
The Full Legalizer in the AP flow takes a partial placement (of any legality) and returns a fully legal placement. A legal placement is a set of legal clusterers of atoms and legal places for those clusters to be placed.

To fully test the AP flow, made a temporary Global Placer which will just place the APBlocks at the center of the device. This will be fixed later.

Also includes basic tests to ensure the AP flow does not regress.


AlexandreSinger:
@vaughnbetz Do you know what this error means? This is showing up in my `ch_intrinsics` basic test for the AP flow when the highest level of assertions are on.
Screenshot from 2024-09-27 15-44-59

The error is coming from an ASSERT debug message here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3e53a937f772bcb81998e57ae074e2a929d49420/vpr/src/timing/slack_evaluation.cpp#L286-L288

Which is defined here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3e53a937f772bcb81998e57ae074e2a929d49420/vpr/src/timing/slack_evaluation.cpp#L437-L468

My theory is that the clustering from this test is so bad that its causing weird errors to pop up. I am thinking I can just remove this testcase and add it back once the Global Placer is fully fleshed out.

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -0,0 +1,406 @@
...
+        std::list<t_pack_molecule*> mol_list;
+        for (APBlockId ap_blk_id : blocks_in_tiles[tile_id]) {
+            // FIXME: The netlist stores a const pointer to mol; but the cluster
+            //        legalizer does not accept this. Need to fix one or the other.
+            // For now, using const_cast.
+            t_pack_molecule* mol = const_cast<t_pack_molecule*>(ap_netlist_.block_molecule(ap_blk_id));
+            mol_list.push_back(mol);
+        }
+        // Clustering algorithm: Create clusters one at a time.
+        while (!mol_list.empty()) {
```
vaughnbetz, line 296: Could be another helper function.

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -0,0 +1,406 @@
...
+    const ClusteredNetlist& clb_nlist = g_vpr_ctx.clustering().clb_nlist;
+
+    // Verify the packing and print some info
+    check_netlist(packer_opts_.pack_verbosity);
+    writeClusteredNetlistStats(vpr_setup_.FileNameOpts.write_block_usage);
+    print_pb_type_count(clb_nlist);
+
+    // PLACING:
+    // Create a lookup from the AtomBlockId to the APBlockId
+    vtr::vector<AtomBlockId, APBlockId> atom_to_ap_block(atom_netlist_.blocks().size());
```
vaughnbetz, line 344: I'd make this a helper function

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -0,0 +1,406 @@
...
+    // improve other algorithms.
+    VTR_LOG("Number of clusters which needed to be moved: %zu\n", unplaced_clusters.size());
+
+    // TODO: Print a breakdown per block type. We may find that specific block
+    //       types are always conflicting.
+
+    // FIXME: Allocate and load moveable blocks?
+    //      - This may be needed to perform SA. Not needed right now.
+
+    // FIXME: Check initial placement legality?
```
vaughnbetz, line 404: Yes, should call a legality checker right away.

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -0,0 +1,406 @@
...
+        //        cluster legalizer and the clustered netlist!
+        const std::unordered_set<AtomBlockId>& atoms_in_cluster = g_vpr_ctx.clustering().atoms_lookup[cluster_blk_id];
+        VTR_ASSERT(atoms_in_cluster.size() > 0);
+        AtomBlockId first_atom_blk = *atoms_in_cluster.begin();
+        APBlockId first_ap_blk = atom_to_ap_block[first_atom_blk];
+        size_t blk_x_loc = p_placement.block_x_locs[first_ap_blk];
+        size_t blk_y_loc = p_placement.block_y_locs[first_ap_blk];
+        size_t blk_layer_num = p_placement.block_layer_nums[first_ap_blk];
+        size_t blk_sub_tile = p_placement.block_sub_tiles[first_ap_blk];
+        t_physical_tile_loc tile_loc(blk_x_loc, blk_y_loc,blk_layer_num);
```
vaughnbetz, line 372: I think you should have a helper function that takes a double loc (ap_loc? pl_cont_loc?) and snaps it to the nearest grid tile (taking into account anchor points).  This code will round down, and also may try to place things at locations that aren't the anchor point of a big block.  A helper function would encapsulate the extra checks etc.

vaughnbetz:
The error comes from:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3e53a937f772bcb81998e57ae074e2a929d49420/vpr/src/timing/slack_evaluation.cpp#L463

It appears to be complaining that the incremental timing analysis & full timing analysis do not agree on the worst case slack. To use incremental timing analysis you have to mark the edges of the timing graph / connections in the netlist whose delay has changed as "dirty". If you didn't do that and moved something around, you can't call incremental timing analysis (but can call the full recompute from scratch).

AlexandreSinger:
@vaughnbetz The issue I am running into with this slack evaluation is that I have not even touched the timing graph... Whats worst is that this error comes up on the second routing attempt (since it is doing a min channel width sweep). It looks like this error is actually in routing and that my packing & placement just reveals the issue. I think there may be a small issue with how the incremental timing is being performed in the router. The reason this may not have been caught is that this is a debug check and there are only 4 small testcases which have this check turned on (vtr_reg_basic with timing on). I am not too sure how to proceed on this issue beyond turning off this test for now and raising an issue. My guess is something in the router is not marking a change in the routing graph as dirty, but I am not sure where to look!

vaughnbetz:
OK, that sounds fine.  You can turn off the test for now and raise an issue.  Please tag Fahri on it ... he's the most likely one to figure it out.  Maybe your clustering is so different you show a mistake that usually doesn't show up, as you suggest.  Not sure how else you could impact the router.  (The timing engine does make callbacks to figure out delays, so if you left out some clustering info that would be a possibility -- maybe it is doing invalid look ups? But it could definitely also be a latent bug that you've exposed).

AlexandreSinger:
@vaughnbetz Sounds good. I will raise an issue on it. I do not know why I did not think of this before, but I generated the .net and .place files from the AP flow and then fed those into VPR independently of the AP flow and was able to replicate the issue. This shows that it has nothing to do with the AP flow and this is just a really hidden bug that is exposed by my really crappy clustering and placement.

I'll include all of these generated files and the commands to replicate this in my issue.

AlexandreSinger:
To show how hidden this bug is (and how unlucky I was). This command has the issue:
```
vpr fixed_k6_frac_N8_22nm.xml ch_intrinsics --circuit_file ch_intrinsics.pre-vpr.blif --net_file ch_intrinsics.net --place_file ch_intrinsics.place --route --device unnamed_device --read_vpr_constraints ../../../../ch_intrinsics_fixed_io.xml --min_route_chan_width_hint 40
```
And this command does not:
```
vpr fixed_k6_frac_N8_22nm.xml ch_intrinsics --circuit_file ch_intrinsics.pre-vpr.blif --net_file ch_intrinsics.net --place_file ch_intrinsics.place --route --device unnamed_device --read_vpr_constraints ../../../../ch_intrinsics_fixed_io.xml
```
Just removing the min channel width hint causes the routing problem to change slightly and remove this issue!

vaughnbetz:
Weird.  I don't suppose there's any chance we forgot to re-timing analyze from scratch after restoring the routing or anything like that (which might be more directly related to the hint)?

AlexandreSinger:
@vaughnbetz I have resolved most of your comments:
- I have move count models into the Cluster Legalizer (this just made sense)
- I have split the full legalizer into two methods (Clustering and Cluster Placement)
- Added a method to get the containing physical tile loc from the partial placement position of an APBlock
- Added some code to print the .place file after the AP flow (I just forgot to do that)
- I turned off that failing testcase. Will raise an issue soon.

I have purposfully left large parts of the full legalizer un-outlined since I plan to return to it very soon to further clean it up and completely restructure it. I just wanted to push something that is working asap. For example, the verification function from the initial placer was seg faulting due to some data structures the placer used not being initialized (but I do not need the structures at all). My goal is the properly unify my cluster placement with the initial placer.

Regarding the "constraint_dir=" comment, I completely agree with you; but I am avoiding modifying the scripting infrastructure while the CI is down. I also want to do this in its own PR.

Regarding which tile to place the given APBlock, I disagree that it should be the nearest. At the Full Legalizer level, the APBlocks will be hovering over the tile they want to be in (even if it is at position 0.9,0.9 its still in the 0th tile). The issue with VTR is that we uniquely identify tiles by their top-left corner which is super confusing. In reality, I imagine each APBlock by being placed at the center of the tile. My approach does break-down when the tiles are not 1x1; however, this is the job of the partial legalizer to resolve (at least in the current design). Eventually I would like to make the cluster placer try to place cluster blocks to the tile whose center is closest to the centroid location of all blocks in the cluster; but this is a bit down the road. Right now, I went for simplicity. The partial placement is assumed to be valid (verify passed, which ensures all blocks are on the device) and all blocks want to be in whatever tile they are hovering above.

vpr/src/place/initial_placement.h
```diff
@@ -152,4 +154,14 @@ bool place_one_block(const ClusterBlockId blk_id,
                      std::vector<t_grid_empty_locs_block_type>* blk_types_empty_locs_in_grid,
                      vtr::vector<ClusterBlockId, t_block_score>* block_scores,
                      BlkLocRegistry& blk_loc_registry);
+
+/**
+ * @brief Initializes the grid to empty. It also initialized the location for
+ * all blocks to unplaced.
+ *
+ * @param blk_loc_registry Placement block location information. To be filled with the location
```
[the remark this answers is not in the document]
AlexandreSinger: I just moved this comment; but I think it is correct?

AlexandreSinger:
Issue has been raised: issue #2754

vaughnbetz:
Ok thanks for the suggestion summary. Isn't the anchor point of blocks their bottom left corner though?

AlexandreSinger:
Sorry, yes. My mind is in graphics land, where I consider (0,0) to be the top left corner. The anchor-points of all tiles is the floor of all positions in it. So (0.5, 0.5) is the tile identified by (0,0). TBH mathematically it does not matter if the blocks are placed at the center or corner of tiles, since it would just add a constant offset to all blocks which will have no affect on the HPWL calculation, but it just makes more logical sense; since, as you pointed out, it moves an AP block from (0.9, 0.9) all the way to (0,0) when (1,1) is technically smaller of a displacement.

vaughnbetz:
Thanks ... please merge after you add the comment above and figure out CI failures are spurious etc.

PR #2756, opened 2024-10-02
[AP] Analytical Solver

AlexandreSinger:
This introduces the Analytical Solver class to the AP flow. This is an integral part of the Global Placement stage and what gives Analytical Placement its name.

This PR introduces the QP_HYBRID analytical solver which uses a hybrid Clique and Star net model to optimize the quadratic HPWL objective.

The code is designed to allow for other analytical solvers to be implemented and interchanged without issue. A B2B solver will be coming in a future PR.


AlexandreSinger:
@vaughnbetz The code is functional and ready for review. The tests are failing because Eigen is not currently included in the CI for the regular tests. Do you think it is appropriate to always include Eigen in the CI tests? and to that end, should we make Eigen a required package for VTR? Let me know what you think. One idea is I can make the solve class crash if Eigen is not installed (but everything else can theoretically work) using IFDEFs. It would just be a bit messy.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -0,0 +1,264 @@
...
+            for (APPinId pin_id : netlist_.net_pins(net_id)) {
+                APBlockId blk_id = netlist_.pin_block(pin_id);
+                add_connection_to_system(star_node_id, blk_id, w);
+            }
+            star_node_offset++;
+        } else {
+            // Create a clique connection where every block in a net connects
+            // exactly once to every other block in the net.
+            // Using the weight from FastPlace
+            double w = 1.0 / static_cast<double>(num_pins - 1);
```
vaughnbetz, line 154: This might give too much weight to high fanout nets (n^2 edges * 1/n weight --> n weight for an n-pin net).  ASIC placers may run with more controlled fanout (generally I'd expect some buffer insertion was done before running an ASIC placer) so the scaling of this to the very high fanouts in FPGA nets may not work out well.  I'd add a note / TODO for that.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -0,0 +1,264 @@
...
+ * @brief Helper method to update the linear system with anchors to the current
+ *        partial placement.
+ *
+ * For each moveable block (with row = i) in the netlist:
+ *      A[i][i] = A[i][i] + coeff_pseudo_anchor;
+ *      b[i] = b[i] + pos[block(i)] * coeff_pseudo_anchor;
+ * Where coeff_pseudo_anchor grows with each iteration.
+ *
+ * This is basically a fast way of adding a connection between a moveable block
+ * and a fixed block.
```
vaughnbetz, line 198: Great comment!  

Would like to see something like this for the net coefficient code ...

vaughnbetz: Wouldn't this function be shorter (need fewer inputs) if it was part of the analytical solver class?  
Should comment the inputs (names are very good, so short comments probably OK but I still think it's good to have a full doxygen comment).

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -0,0 +1,264 @@
...
+    // TODO: Use solve with guess to make this faster. Use the previous placement
+    //       as a guess.
+    Eigen::VectorXd x = cg.solve(b_x_diff);
+    VTR_ASSERT(cg.info() == Eigen::Success && "Conjugate Gradient failed at solving b_x!");
+    Eigen::VectorXd y = cg.solve(b_y_diff);
+    VTR_ASSERT(cg.info() == Eigen::Success && "Conjugate Gradient failed at solving b_y!");
+
+    // Write the results back into the partial placement object.
+    for (size_t row_id_idx = 0; row_id_idx < num_moveable_blocks_; row_id_idx++) {
+        APRowId row_id = APRowId(row_id_idx);
```
vaughnbetz, line 256: This code assumes that the first rows are all for movable blocks. I think you commented that elsewhere so it's not the end of the world, but you could rewrite this code to go through all blocks and get their mapping to rows, so it wouldn't have any such assumption.

vpr/src/analytical_place/partial_placement.cpp
```diff
@@ -8,8 +8,29 @@
 #include "partial_placement.h"
 #include <cmath>
 #include <cstddef>
+#include <limits>
 #include "ap_netlist.h"
 
+double PartialPlacement::get_hpwl(const APNetlist& netlist) const {
+    double hpwl = 0.0;
+    for (APNetId net_id : netlist.nets()) {
```
vaughnbetz, line 16: Are the ignored nets removed from the ap netlist so there's no need to check for them here?

vaughnbetz (commented):
Looks good!  A few thoughts and feedbacks.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+}
+
+#ifdef EIGEN_INSTALLED
+
+void QPHybridSolver::init_linear_system() {
+    // Count the number of star nodes that the netlist will have.
+    size_t num_star_nodes = 0;
+    for (APNetId net_id : netlist_.nets()) {
+        if (netlist_.net_pins(net_id).size() > star_num_pins_threshold)
```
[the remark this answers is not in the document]
vaughnbetz: Actually I think you removed all uninteresting nets from the AP netlist already, so ignore that comment ...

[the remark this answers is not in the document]
AlexandreSinger: That is correct. I realized that every single one of these for loops that iterate over the nets in the APNetlist were always ignoring the nets, which I felt was an accident waiting to happen. I completely removed them from the netlist. If we ever need all the nets, we can always go back to the Atom netlist, but I really do not expect to need those nets.

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: This is the star node weight. For a net with P pins, there will be P edges created or the star net model. This weight factor is a correction term to account for these P extra pins. Hence why it is a linear correct. This is from the FastPlace paper. We can explore other weight terms, but I am not sure if they approach the correct solution. I will add a TODO to investigate.

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: Sadly no. It would only save 2 arguments. It also does not make sense to use this method outside of this file's context.

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: This was actually done for performance. I think iterating linearly from 0 to num_moveable_blocks would be more efficient than iterating over all blocks and skipping the fixed blocks. I left a comment explaining that the first rows must represent the moveable blocks and added debug asserts to enforce this.

vpr/src/analytical_place/partial_placement.cpp
AlexandreSinger: That is correct. This HPWL is not meant to be accurate, its meant to track how the Global Placer is converging. If we used a more accurate HPWL which knew more information about the problem than the solver, we may never converge I think.

PR #2691, opened 2024-08-15
Add flat version of sync_netlists_to_routing

fkosar-ql:
Run flat changes the internal cluster routing, and can also rotate truth tables in LUTs. To generate a bitstream or perform post-routing simulation, we need to update the intra-cluster routing data and mapping data of what inputs/outputs have been re-ordered by the router. this PR does that.

#### Description
Made a sync_netlists_to_routing flat routine which rebuilds the clustered netlist and the intra-cluster routing (pb_routes) and updates the atom lookup data structure.

#### Related Issue
Issue from OpenFPGA repo: 

#### Motivation and Context
Flat routing won't be usable by QuickLogic, OpenFPGA etc. without this change.

#### How Has This Been Tested?
- Ran with a QuickLogic architecture and verified the bitstream is correct
- Ran with the flagship architecture and confirmed it doesn't crash.
TODO:
- Run with stratixiv_arch, confirm it doesn't crash (currently it does)
- Run post-routing simulation on both flagship arch and stratixiv_arch to confirm they work.
- Should put at least one run-flat post-routing simulation for flagship arch and stratix iv arch in CI.

#### Types of changes
- [ ] Bug fix (change which fixes an issue)
- [x] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:
- No documentation changes
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

Issue #2754, opened 2024-09-29
[Routing] Incremental Slack Timing Analysis Does Not Match When Computed From Scratch


AlexandreSinger:
While working on the Analytical Placer, I generated a very basic clustering and cluster placement. The quality of both of these was meant to be low as this is mainly just a PoC flow through VPR. These tests go through routing without an issue; however, for one test they fail.

When the assert level is set to 4, there is a debug check which verifies that the incremental slack timing analysis gets the same answer as one computed from scratch:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3e53a937f772bcb81998e57ae074e2a929d49420/vpr/src/timing/slack_evaluation.cpp#L286-L288

I have found on my testcase, this check gets tripped.

VTR was built with the following command:
```
make -j12 CMAKE_PARAMS="-DWITH_ODIN=off -DVTR_ASSERT_LEVEL=4"
```
Note: The assert level being 4 is important.

The testcase circuits and a testing script can be found in this tar file:
[timing_bug.tar.gz](https://github.com/user-attachments/files/17179821/timing_bug.tar.gz)

It contains a bash script with the command I used to run VPR:
```sh
VTR_DIR=~/vtr-verilog-to-routing                                                                                                                                                                            
                                                                                                                                                                                                            
$VTR_DIR/vpr/vpr \                                                                                                                                                                                          
    fixed_k6_frac_N8_22nm.xml \                                                                                                                                                                             
    ch_intrinsics \                                                                                                                                                                                         
    --circuit_file ch_intrinsics.pre-vpr.blif \                                                                                                                                                             
    --net_file ch_intrinsics.net \                                                                                                                                                                          
    --place_file ch_intrinsics.place \                                                                                                                                                                      
    --route \                                                                                                                                                                                               
    --device unnamed_device \                                                                                                                                                                               
    --read_vpr_constraints ch_intrinsics_fixed_io.xml \                                                                                                                                                     
    --min_route_chan_width_hint 40
```
The output looks like as follows:
<img width="1026" alt="image" src="https://github.com/user-attachments/assets/4250d1f9-e6a3-4887-aab0-d04876993d89">

This is the second routing attempt after routing at a channel width of 40.

Interestingly, if you do not provide the min chan width hint of `--min_route_chan_width_hint 40` this does not happen.

AlexandreSinger:
Refer to PR #2752 for more context.

[...]

PR #2758, opened 2024-10-03
Add flat version of sync_netlists_to_routing

duck2:
Re-opened #2691 to see if that fixes the CI.

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1689,6 +1689,9 @@ constexpr std::array<const char*, size_t(SwitchType::NUM_SWITCH_TYPES)> SWITCH_T
  */
 constexpr const char* VPR_DELAYLESS_SWITCH_NAME = "__vpr_delayless_switch__";
 
+/* An intracluster switch automatically added to the RRG by the flat router. */
+constexpr const char* VPR_INTERNAL_SWITCH_NAME = "__vpr_intra_cluster_switch__";
+
 enum class BufferSize {
     AUTO,
     ABSOLUTE
--- a/libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
+++ b/libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
@@ -526,7 +526,7 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         bool found_arch_name = false;
         std::string string_name = std::string(name);
         // The string name has the format of "Internal Switch/delay". So, I have to use compare to specify the portion I want to be compared.
-        bool is_internal_sw = string_name.compare(0, 15, "Internal Switch") == 0;
+        bool is_internal_sw = string_name.compare(0, strlen(VPR_INTERNAL_SWITCH_NAME), VPR_INTERNAL_SWITCH_NAME) == 0;
         for (const auto& arch_sw_inf: arch_switch_inf_) {
             if (string_name == arch_sw_inf.name || is_internal_sw) {
                 found_arch_name = true;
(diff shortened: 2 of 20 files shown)


vaughnbetz:
Spree.v failed again in vtr_reg_strong:

```
regression_tests/vtr_reg_strong/strong_flat_router...[Fail]
[Fail]
k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common vpr_status Task value 'exited with return code 134' does not match golden 'success'
[Fail]
k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common routed_wirelength relative value -9.344921035417251e-05 outside of range [0.6,1.5], above absolute threshold 5.0 and not equal to golden value: 10701.0
[Fail]
k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common logic_block_area_total relative value -1.0868915445272294e-07 outside of range [0.8,1.3] and not equal to golden value: 9200550.0
[Fail]
k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common logic_block_area_used relative value -1.877042456823331e-07 outside of range [0.8,1.3] and not equal to golden value: 5327530.0
[Fail]
k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common min_chan_width_routing_area_total relative value -6.770893283950952e-07 outside of range [0.7,1.3] and not equal to golden value: 1476910.0
[Fail]
k6_frac_N10_frac_chain_mem32K_40nm.xml/spree.v/common min_chan_width_routing_area_per_tile relative value -0.0001327099504593755 outside of range [0.7,1.3] and not equal to golden value: 7535.23
```

There's also a spurious unit test failure (odd_even routing in the NoC tests): @soheilshahrouz can you disable that one?

fkosar-ql:
This is very annoying:
1. Sometimes the valgrind tests cannot find `odin_II` (?)
2. The `NO_GRAPHICS` test acts randomly: it either passes, fails or gets stuck in a loop (??)
3. When I set up ssh access into the runner via tmate I can't reproduce the failure.
3.1. Sometimes I can, but then if I run the test again it passes.
3.1.1. It seems to always pass after the first time. I tried running it in a loop 50 times and there's no failure.
3.2. There's a very interesting case where the test fails with return code 2 but when I print the test log I see return code 0
3.3. I tried rebuilding with ASAN inside the runner and there seems to be no issue.

vaughnbetz:
@AlexandreSinger : maybe 4 cores is running out of memory on the GitHub runners? Perhaps that is why we only used 2 until recently?

AlexandreSinger:
@vaughnbetz I would be very surprised. Since merging the 4-core change the CI was run 7 times on the master branch and did not see any issue (asside from the known unit test randomly failing). However, I do agree that the symptoms Fahri points out would point to this. I agree to trying to set it back to 2-cores and see if that fixes it. But again, I would be very suprised if this is the problem.

AlexandreSinger:
@fkosar-ql The most recent commit you made did not turn off the 4-core thing! You need to set the number of cores explicitly! Or else it will default to the max number of cores on the machine.

AlexandreSinger:
Looking at Fahris most recent commit, he did turn it back to 2 cores successfully; but the Tests still seem to be failing.

fkosar-ql:
🤦🤦🤦 I accidentally included `regression_tests/vtr_reg_basic/hdl_include_yosys` twice in the task list. It probably goes into a race. Let me see if removing it fixes the tests

fkosar-ql:
That seems to fix it... merging

AlexandreSinger:
@fkosar-ql The CI is now failing on Master with the same failures as what was seen in this PR before:

AlexandreSinger:
Actually, reading into these errors its different. The strong tests are just golden test not matching, and the basic test failures are the failures noted in issue #2754 

Something about how this change interacted with other code being merged in may have brought it out.

fkosar-ql:
Yeah, the failures don't look the same. This PR doesn't touch routing or STA so it would be surprising if it was causing the issue.

PR #2759, opened 2024-10-04
[CI] Improved How The CI Handles Num Cores Used

AlexandreSinger:
Found that the CI was always building and running using only 2 cores on the regular tests. However, the CI has 4 cores that it can use. The number of cores can change based on some factors.

Added the ability of the tests to check how many cores they have available, and use all the cores to build and run VTR tests.


AlexandreSinger:
@vaughnbetz This is ready for review. Comparing the runtime of these Tests and Test for a previous run, this runtime is around 20 minutes faster overall. Based on the CI, Test usually takes around 1 hour and 15 minutes; after this change, Test now takes around 55 minutes. Each job in Test appears to be taking around 10-20% less time. I think the overall improvement would continue to improve once we remove the Odin tests.

Although 20 minutes does not seem like much, the CI is being throttled heavily. There appears to be N GitHub runners available, but each run of the CI takes around 2N runners to complete. This creates a bottleneck especially if many people are trying to run the CI at the same time. Optimizing how long each test runs can help alleviate the bottleneck.

PR #2762, opened 2024-10-09
CI test

fkosar-ql:
Just trying to see if the CI is green without any changes.


AlexandreSinger:
@fkosar-ql There is a test that runs every night which runs all of the CI tests. If you click on the Action tab above you should be able to find it. At the moment the NIghtlyTests are all failing, but the other tests should all work. Note: There is a spurriously failing UnitTest which is failing (odd_even_routing) which should be turned off.

AlexandreSinger:
They look like this:
image

PR #2764, opened 2024-10-10
[CI] Sanitizer Update

AlexandreSinger:
Updated the sanitizer to use the standard VTR flow (without Odin) instead of with Odin.

Turned on the Strong tests as well to get better sanitizer coverage. This test was originally turned off since it took too long. Will use a PR to investigate how long this takes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -288,10 +288,14 @@ jobs:
         include: [
           {
             name: 'Basic',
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
@vaughnbetz I found out why the sanitizer was only running the Odin tests. Turns out Yosys itself fails our sanitizers...

AlexandreSinger:
One idea is that I can make the sanitizer on sanitize VPR. What do you think?

vaughnbetz:
Could do but I'd just stick with the Odin-II runs for now. They implicitly sanitize VPR too.

PR #2768, opened 2024-10-12
[CI] Fix Golden Results

AlexandreSinger:
Golden results are incorrect for the strong tests after recent merge conflicts.

There may still be an issue with the basic with highest assertions test. See issue #2754


vaughnbetz:
Thanks Alex. We could temporarily disable the failing test too.

PR #2769, opened 2024-10-13
[Place] Independent Placement Verification

AlexandreSinger:
Created a method that would independently verify the placement in the VPR flow. If a placement passes this verification, it is assumed that it can be used in routing without issue.

By design, this method does not use any global variables (everything needs to be passed in) and recomputes everything that it does not assume. This allows it to be independent of the placement flow being used, so this method can be used in the AP flow as well.

### DIFF
--- a/vpr/src/analytical_place/full_legalizer.cpp
+++ b/vpr/src/analytical_place/full_legalizer.cpp
@@ -7,7 +7,6 @@
 
 #include "full_legalizer.h"
 
-#include <cmath>
 #include <list>
 #include <unordered_set>
 #include <vector>
@@ -27,6 +26,7 @@
 #include "physical_types.h"
 #include "place_constraints.h"
 #include "place_macro.h"
+#include "verify_placement.h"
 #include "vpr_api.h"
 #include "vpr_context.h"
 #include "vpr_error.h"
@@ -396,5 +396,16 @@ void FullLegalizer::legalize(const PartialPlacement& p_placement) {
 
     // Place the clusters based on where the atoms want to be placed.
     place_clusters(clb_nlist, p_placement);
+
+    // Verify that the placement created by the full legalizer is valid.
+    unsigned num_placement_errors = verify_placement(g_vpr_ctx);
+    if (num_placement_errors == 0) {
+        VTR_LOG("Completed placement consistency check successfully.\n");
+    } else {
+        VPR_ERROR(VPR_ERROR_AP,
+                  "Completed placement consistency check, %u errors found.\n"
+                  "Aborting program.\n",
+                  num_placement_errors);
+    }
 }
 
--- a/vpr/src/place/verify_placement.cpp
+++ b/vpr/src/place/verify_placement.cpp
@@ -1,0 +1,254 @@
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    October 2024
+ * @brief   Definitions of the independent placement verifier.
+ *
+ * Most of this code originally came from place.cpp and place_constraints.cpp
+ *
+ * By design, these methods should not use any global variables and anything
+ * that it does not assume should be re-computed from scratch. This ensures that
+ * the placement is actually valid.
+ */
+
+#include "verify_placement.h"
+#include <vector>
+#include "blk_loc_registry.h"
+#include "clustered_netlist.h"
+#include "device_grid.h"
+#include "partition_region.h"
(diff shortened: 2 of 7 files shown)


vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -393,5 +393,16 @@ void FullLegalizer::legalize(const PartialPlacement& p_placement) {
...
     // Place the clusters based on where the atoms want to be placed.
     place_clusters(clb_nlist, p_placement);
+
+    // Verify that the placement created by the full legalizer is valid.
+    unsigned num_placement_errors = verify_placement(g_vpr_ctx);
+    if (num_placement_errors == 0) {
+        VTR_LOG("Completed placement consistency check successfully.\n");
+    } else {
+        VPR_ERROR(VPR_ERROR_AP,
+                  "Completed placement consistency check, %d errors found.\n"
```
AlexandreSinger, line 403: FIXME: This should be %u not %d

vpr/src/place/verify_placement.cpp
```diff
@@ -0,0 +1,250 @@
...
+ *  @param pl_macros            The placement macros to verify over.
+ *
+ *  @return The number of errors in the macro placement.
+ */
+static unsigned check_macro_placement_consistency(const BlkLocRegistry& blk_loc_registry,
+                                                  const std::vector<t_pl_macro>& pl_macros) {
+    const auto& block_locs = blk_loc_registry.block_locs();
+    const auto& grid_blocks = blk_loc_registry.grid_blocks();
+
+    unsigned error = 0;
```
AlexandreSinger, line 149: FIXME: To be consistent with the rest of the file, this should be `num_errors`.

AlexandreSinger:
@vaughnbetz This is the `check_place` changes we discussed on Friday. I split out all of the invariants from the check_place method and used it to create this method. The idea of the `verify_placement` method is to be an independent check to ensure that all the necessary data structures of placement are filled correctly such that the rest of the VPR flow has no issues (nothing more, nothing less). Please review.

I do have a FIXME in this code which I am not sure about. Are the macros required for the rest of the VPR flow? My intention for this code is that this MUST pass in order for the rest of the VPR flow to be used without issue; but I do not see uses of the pl_macros anywhere outside of placement (and drawing). If they are not used then I think this check should happen somewhere else. What do you think?

vaughnbetz:
If the macros aren't respected generally routing will fail, as the main reason macros are created is for fixed routing patterns (no programmability).  So I'd check them.

vpr/src/place/verify_placement.cpp
```diff
@@ -0,0 +1,253 @@
...
+        if (blk_pr.empty())
+            continue;
+        // Check if the block is placed in its constrained partition region.
+        const t_pl_loc& blk_loc = block_locs[blk_id].loc;
+        if (!blk_pr.is_loc_in_part_reg(blk_loc)) {
+            VTR_LOG_ERROR(
+                "Block %zu is not in correct floorplanning region.\n",
+                size_t(blk_id));
+            num_errors++;
+        }
```
soheilshahrouz, line 214: This is kind of my personal preference, so feel free to ignore it.

I think adding a few empty lines make the code easier to read. For example, I think of checking emptiness of `blk_pr` as a logical step. If the if statement and its comment are preceded and followed by empty lines, I can visually recongnize it as a separte step.

soheilshahrouz:
A few checks are done in `sync_grid_to_blocks()` function. I feel those check belong to `verify_placement.cpp`.

vaughnbetz:
This looks good to me. @AlexandreSinger feel free to merge when you are ready. I tend to agree with @soheilshahrouz that a few blank lines to break up concepts are helpful to readability, but will leave it to you to decide if /where to add any blank lines.

vaughnbetz:
As we discussed the sync_grid_to_blocks() function call could/should be moved inside load_placement, which will make the overall control a bit cleaner. You could also move the post_place_sync () function call into *both* the end of placement and the end of load_placement if you like, as that would again simplify the overall control.

Merge whenever you see fit ...

danahow:
If you want to mark each and every grid covered by a tile, fill in non-root grids with a pointer to the associated root grid. These extra pointers never need to change.

vaughnbetz:
That's what happens now; the grid has non-root node locations point back to the root node.

PR #2771, opened 2024-10-14
[Clustering] Independent Clustering Verification

AlexandreSinger:
Created a method that would independently verify the clustering in the VPR flow. If a clustering passes this verification, it is assumed that it can be used in placement and routing without issue.

By design, this method does not use any global variables (everything needs to be passed in) and it recomputes everything that it does not assume. This allows it to be independent of the packing flow, so this method can also be used in the AP flow.

### DIFF
--- a/vpr/src/pack/verify_clustering.cpp
+++ b/vpr/src/pack/verify_clustering.cpp
@@ -222,7 +222,7 @@ static unsigned check_clustering_pb_consistency(const ClusteredNetlist& clb_nlis
  *        with the floorplanning constraints on the atoms.
  *
  * This checks the following invariants:
- *  - If a cluster is unconstrained, all of its atoms are unconrstrained.
+ *  - If a cluster is unconstrained, all of its atoms are unconstrained.
  *  - If a cluster is constrained, at least one of its atoms are constrained.
  *  - If a cluster is constrained, each of its constrained atoms can be placed
  *    within its constrained region.
--- a/vpr/src/pack/verify_clustering.cpp
+++ b/vpr/src/pack/verify_clustering.cpp
@@ -1,0 +1,444 @@
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    October 2024
+ * @brief   Definitions of the independent verifier.
+ *
+ * By design, these methods should not use any global variables and anything
+ * that it does not assume should be re-computed from scratch. This ensures that
+ * the clustering is actually valid.
+ */
+
+#include "verify_clustering.h"
+#include <unordered_set>
+#include "atom_lookup.h"
+#include "atom_netlist.h"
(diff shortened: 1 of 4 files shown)


AlexandreSinger:
@vaughnbetz This is a complementary PR to the Independent Placement Verification PR I raised yesterday (PR #2769 ). I could not find anywhere in the clustering flow that checked these invariants directly so this should improve test coverage. This checks the clustering to ensure that every atom is in a cluster, every cluster has correct floorplanning constraints, and the cluster pbs are correct (for the most part). Please review.

vpr/src/pack/verify_clustering.cpp
```diff
@@ -0,0 +1,444 @@
...
+                                                  atom_nlist,
+                                                  atom_lookup);
+    // Check that all clusters are clustering atoms which are constrained to
+    // overlapping regions and the constraints of the clusters are consistent
+    // with the atoms they contain.
+    num_errors += check_clustering_floorplanning_consistency(clb_nlist,
+                                                             clb_atoms,
+                                                             cluster_constraints,
+                                                             constraints);
+    // TODO: There exists more checks for the clustering in base/check_netlist.cpp
```
vaughnbetz, line 428: I think general netlist features like mapping consistency are good to put in the netlist verify routine if possible. But I guess given that the atom_lookup clb mapping is not in the netlist or tied to it, it's OK to have the checks here instead.  Just suggesting you think about whether we should have a lookup verifier that can be called by this routine for checks that make sense ...

vpr/src/pack/verify_clustering.cpp
```diff
@@ -0,0 +1,444 @@
...
+    }
+    return num_errors;
+}
+
+/**
+ * @brief Check that the floorplanning constraints on the cluster is consistent
+ *        with the floorplanning constraints on the atoms.
+ *
+ * This checks the following invariants:
+ *  - If a cluster is unconstrained, all of its atoms are unconrstrained.
```
vaughnbetz, line 225: unconstrained typo

vpr/src/pack/verify_clustering.h
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * @brief Verifies the clustering of the atom netlist as it appears in the
+ *        given context.
+ *
+ * This performs the verification in the method above, but performs it on the
+ * given VPR context itself. This verifies that the actual clustering being used
+ * through the rest of the flow is valid.
+ *
+ * NOTE: The above method is used when one wishes to verify a temporary
```
vaughnbetz, line 75: I'd say the verify_clustering () function with more parameters is used ...
(above can get out of date with code restructuring)

vaughnbetz (approved):
Looks good; merge when you see fit.

PR #2772, opened 2024-10-14
Increase read buffer to reduce total syscalls

heshpdx:
This patch reduces read syscalls by 96%. `strace` showed that we were doing small incremental reads of large files:

```
openat(AT_FDCWD, "stratixiv_arch.timing.xml", O_RDONLY) = 5
read(5, "<architecture>\n  <models>\n    <!"..., 8191) = 8191
read(5, "sink_ports=\"eccstatus portbdatao"..., 8191) = 8191
read(5, "rtbdataout\" clock=\"clk_portbout\""..., 8191) = 8191
read(5, "ortbdataout\"/>\n        <port nam"..., 8191) = 8191
read(5, "ombinational_sink_ports=\"eccstat"..., 8191) = 8191
read(5, "nal_sink_ports=\"eccstatus portbd"..., 8191) = 8191
...
```
This might have been fine 20 years ago, but we can modernize it. Increasing the buffer size reduces the total number of syscalls, which helps in scale-out scenarios with many copies of vpr running on the same system. With this patch, strace looks like:
```
openat(AT_FDCWD, "stratixiv_arch.timing.xml", O_RDONLY) = 5
read(5, "<architecture>\n  <models>\n    <!"..., 1048576) = 1048576
```

Another win comes with in libblifparse, which is reading in large blif files. I was able to modify my generated yacc output, however I cannot figure out the correct way to capture this upstream in the build:
```
--- a/src/vtr-vpr/libs/EXTERNAL/libblifparse/src/blif_lexer.gen.cpp
+++ b/src/vtr-vpr/libs/EXTERNAL/libblifparse/src/blif_lexer.gen.cpp
@@ -971,12 +971,7 @@ static int input ( yyscan_t yyscanner );
 
 /* Amount of stuff to slurp up with each read. */
 #ifndef YY_READ_BUF_SIZE
-#ifdef __ia64__
-/* On IA-64, the buffer size is 16k, not 8k */
-#define YY_READ_BUF_SIZE 16384
-#else
-#define YY_READ_BUF_SIZE 8192
-#endif /* __ia64__ */
+#define YY_READ_BUF_SIZE 1048576
 #endif
```
Just set all the buffer sizes to 1MB, not sure why ia64 has to be different.

### DIFF
--- a/libs/libpugiutil/src/pugixml_loc.cpp
+++ b/libs/libpugiutil/src/pugixml_loc.cpp
@@ -30,7 +30,7 @@ void loc_data::build_loc_data() {
 
     std::ptrdiff_t offset = 0;
 
-    char buffer[1024];
+    char buffer[1048576];
     std::size_t size;
 
     while ((size = fread(buffer, 1, sizeof(buffer), f)) > 0) {
--- a/libs/libvtrutil/src/vtr_digest.cpp
+++ b/libs/libvtrutil/src/vtr_digest.cpp
@@ -21,7 +21,7 @@ std::string secure_digest_stream(std::istream& is) {
     //Read the stream in chunks and calculate the SHA256 digest
     picosha2::hash256_one_by_one hasher;
 
-    std::array<char, 1024> buf;
+    std::array<char, 1048576> buf;
     while (!is.eof()) {
         //Process a chunk
         is.read(buf.data(), buf.size());


libs/libpugiutil/src/pugixml_loc.cpp
```diff
@@ -30,7 +30,7 @@ void loc_data::build_loc_data() {
 
     std::ptrdiff_t offset = 0;
```
vaughnbetz, line 32: Thanks.  I think we should make a defined constant with a comment for this magic number though, e.g. 
#define VTR_LARGE_BUF 1048576  // Read large files like the arch.xml with large buffers for speed.

vaughnbetz:
@duck2 @AlexandreSinger : any idea how to set the YY_READ_BUF_SIZE in a .lex file so we get the big buffer without editing intermediate files?  If we can't figure out a very clean way to do this, we could have a little sed step that gets called during the build to switch this line.

libs/libpugiutil/src/pugixml_loc.cpp
vaughnbetz: Should have a comment on why too.

libs/libpugiutil/src/pugixml_loc.cpp
```diff
@@ -30,7 +30,7 @@ void loc_data::build_loc_data() {
...
 
-    char buffer[1024];
```
vaughnbetz, line 33: Use same defined constant (put in some header, with a comment, or in as a -D make/cmake variable with a comment)

libs/libvtrutil/src/vtr_digest.cpp
```diff
@@ -21,7 +21,7 @@ std::string secure_digest_stream(std::istream& is) {
     //Read the stream in chunks and calculate the SHA256 digest
     picosha2::hash256_one_by_one hasher;
 
-    std::array<char, 1024> buf;
```
vaughnbetz, line 24: Use defined constant.

vaughnbetz:
@AlexandreSinger @amin1377 : I think Alex has a way to make this work in the lex files too (he sent an email on it).  If one of you wants to update this with the #define/make/cmake variable and a lex fix that would be great.

AlexandreSinger:
@vaughnbetz Here is what I would do since I think CMake would do this the best:

Somewhere in `vpr/CMakeLists.txt` I would add the following line:
```sh
target_compile_definitions(libblifparse PUBLIC YY_READ_BUF_SIZE=1048576)
```
I personally would add it somewhere before this line:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/660ecaba29c224c995d70c2e8c6454e8cbe9b9f3/vpr/CMakeLists.txt#L123-L135

I set it to a size of 1048576, but you can set it to whatever size you want. This will make it so when the VPR executable is linked with libblifparse, the libblifparse library would use the YY_READ_BUF_SIZE that we defined above. This is ideal since libblifparse is actually an external project and its a good idea to avoid modifying it anyways. This is possible since the code in libblifparse only sets the YY_READ_BUF_SIZE to 16k/8k if it has not been set already. Another benefit of this approach is that it should only apply to libblifparse and nothing else; just in case someone else is using the same define.

You would somehow need to verify that this took affect since this is a pre-compiler option. But I am pretty sure this would work!

vaughnbetz:
Thanks Alex.  libblifparse is a separate repo, but part of the VTR project and organization (VTR owns it). So we can also directly modify its code and that's probably easiest if it's just a matter of setting YY_READ_BUF_SIZE in it.

AlexandreSinger:
@vaughnbetz If you think that all users of libblifparse would want this, then you can add the exact same command in this CMake file instead:
`libs/EXTERNAL/libblifparse/CMakeLists.txt`

I would add it right after we set the target properties here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/660ecaba29c224c995d70c2e8c6454e8cbe9b9f3/libs/EXTERNAL/libblifparse/CMakeLists.txt#L39-L46

It will do the same effect, but instead of doing it from the VPR side, we are doing from within libblifparse itself. We can then push these changes upstream when we are done.

The great thing about CMake is that the command does not change, just where you use it.

vaughnbetz:
I'd like to land this PR today if at all possible, so the faster reads go into the VTR 9 release.
@MohamedElgammal FYI, in case you can help land this.

vaughnbetz:
@AlexandreSinger believes he can land this shortly ...!

AlexandreSinger:
I have resolved the comments above and made the changes proposed by Mahesh. Thank you so much @heshpdx for these changes! I could not commit to this PR since it was on Mahesh's fork. I have raised another PR with my changes. See PR #2814

AlexandreSinger:
Closing this pull request since PR #2814 covers this.

PR #2776, opened 2024-10-16
[QuickStart] Yosys/Parmys Update

### LINKED ISSUE #2739: Update VTR quick start guide to use yosys / parmys
The VTR quick start guide uses Odin II as the synthesis tool, but the yosys/parmys flow is now the default VTR flow. We should update the quick start to use yosys/parmys.  The Odin II section should be moved down to a later, optional part of the quick start that just describe how to run Odin II (as the primary example of yosys/parmys would have already covered the whole flow).

#### Proposed Behaviour
Change all mention of Odin II to the equivalent parmys / yosys flow commands.  Move the Odin II command(s) down to the end of the quick start under an Odin II section.

#### Current Behaviour
Odin II is the only synthesis tool described in the quick start.

ZohairZaidi:
#### Description

This pull request updates the VTR Quick Start Guide by replacing Odin II as the primary synthesis tool with the Yosys/Parmys flow. I have also updated the graphic images
#### Related Issue

closes #2739  

#### Checklist:


doc/src/quickstart/index.rst
```diff
@@ -297,9 +269,9 @@ Next, we'll optimize and technology map our circuit using ABC, providing the opt
 
 We'll use the following, simple ABC commands::
```
vaughnbetz, line 271: parmys works a little differently than Odin-II; it shoul have already optimized and technology mapped the design.  I suggest you take a look at the .blif file that comes out of parmys and see if it is mostly LUTs (.names), with some having up to 6 inputs.  If so, it has been tech-mapped. We can still show people how to run abc, but should modify the text to say parmys already called abc internally to optimize the design, but if we'd like to try some further optimization or technology mapping options we can also directly invoke abc on the parmys output.

doc/src/quickstart/index.rst
```diff
@@ -308,17 +280,18 @@ The corresponding command to run is:
...
 
-    ABC command line: "read blink.odin.blif; if -K 6; write_hie blink.odin.blif blink.abc_no_clock.blif".
+    ABC command line: "read blink.parmys.blif; if -K 6; write_hie blink.parmys.blif blink.abc_no_clock.blif".
 
     Hierarchy reader converted 6 instances of blackboxes.
     The network was strashed and balanced before FPGA mapping.
     Hierarchy writer reintroduced 6 instances of blackboxes.
 
-If we now inspect the produced BLIF file (``blink.abc_no_clock.blif``) we see that ABC was able to significantly simplify and optimize the circuit's logic (compared to ``blink.odin.blif``):
+If we now inspect the produced BLIF file (``blink.abc_no_clock.blif``) we see that ABC was able to significantly simplify and optimize the circuit's logic (compared to ``blink.parmys.blif``):
```
vaughnbetz, line 294: Take a look at the two files and see if this is true.  Are the number of .names significantly different?  Confirm they are different a bit at least.  We shouldn't say it is significantly simpler / optimized unless it is.

doc/src/quickstart/index.rst
```diff
@@ -485,6 +458,68 @@ which we can visualize with:
         --route_chan_width 100 \
         --analysis --disp on
 
+Manually Running VTR with ODIN II
+----------------------------------
+Let's start by making a new directory for us to work in:
```
vaughnbetz, line 463: Add (at the very start of this section):

VTR includes a second synthesis tool, ODIN II.  Below we explain how to run this alternative synthesis. flow.

doc/src/quickstart/index.rst
```diff
@@ -332,14 +305,14 @@ However, there is an issue with the above BLIF produced by ABC: the latches (ris
 
 Re-inserting clocks
 ^^^^^^^^^^^^^^^^^^^
-We will restore the clock information by running a script which will transfer that information from the original ODIN BLIF file (writing it to the new file ``blink.pre-vpr.blif``):
+We will restore the clock information by running a script which will transfer that information from the original Parmys BLIF file (writing it to the new file ``blink.pre-vpr.blif``):
```
vaughnbetz, line 308: I think this is only needed because we directly ran ABC on the parmys output, and that should be stated (when run directly, ABC does not preserve clocks so we re-insert them in the re-optimized design with the restore_multiclock_latch.pl script).

doc/src/quickstart/index.rst
```diff
@@ -332,14 +305,14 @@ However, there is an issue with the above BLIF produced by ABC: the latches (ris
...
 
 .. code-block:: bash
 
     > $VTR_ROOT/vtr_flow/scripts/restore_multiclock_latch.pl \
-        blink.odin.blif \
-        blink.abc_no_clock.blif \
-        blink.pre-vpr.blif
+        ~/vtr_work/quickstart/blink_manual/temp/blink.parmys.blif \
```
vaughnbetz, line 313: Does run_vtr_flow directly call abc and the restore_multiclock_latch.pl script when run with parmys?  If not, I think we should leave ABC and this script out of the parmys section, and just refer to them in the Odin II section (which does need them).

vaughnbetz (changes requested):
Thanks @ZohairZaidi !  
I have some comments about the parmys flow (I think we've listed some unnecessary steps for it) that you should take a look at closely. Adding @amirarjmand93 and @KennethKent to review; in case @AlexandreSinger has time to review I've tagged him too

doc/src/quickstart/index.rst
vaughnbetz: Thanks.  I think we should just simplify then to show running parmys in this section, without abc or the clock relabelling (we can leave those in the Odin-II section). If you can confirm the clock relabeling is *not* called for the parmys flow in run_vtr_flow that would be ideal -- then we can confidently remove it from this section.

ZohairZaidi: sure thing. It does not appear to be called in run_vtr_flow. 

My initial confusion stemmed from reading the Parmys paper, which states, "the Parmys pass is called to perform partial mapping according to the target architecture, and then a series of optimization and logic mappings are performed to convert all the remaining Yosys internal types to primitives. Next, the output is passed to ABC and VPR stages."

This led me to believe that running only the Parmys part of the flow would require a separate invocation of ABC since we are executing the command with both -start parmys and -end parmys, it explicitly skips the ABC.

I will update the PR and also wait for final confirmation by the added members

ZohairZaidi:
Thanks @vaughnbetz!, I have addressed your suggestions. Will wait for the added members to review as well.

doc/src/quickstart/index.rst
```diff
@@ -239,124 +231,38 @@ Let's start by making a fresh directory for us to work in:
 
 Next we need to run the three main sets of tools:
 
-* :ref:`odin_ii` performs 'synthesis' which converts our behavioural Verilog (``.v`` file) into a circuit netlist (``.blif`` file) consisting of logic equations and FPGA architecture primitives (Flip-Flops, adders etc.),
+* :ref:`Parmys` performs 'synthesis' which converts our behavioural Verilog (``.v`` file) into a circuit netlist (``.blif`` file) consisting of logic equations and FPGA architecture primitives (Flip-Flops, adders etc.),
 * :ref:`ABC` performs 'logic optimization' which simplifies the circuit logic, and 'technology mapping' which converts logic equations into the Look-Up-Tables (LUTs) available on an FPGA, and
```
AlexandreSinger, line 235: Is this line still correct? Is ABC being used when manually running the VTR flow?

doc/src/quickstart/index.rst
```diff
@@ -440,16 +346,16 @@ which should produce output similar to::
 
     EArch/blink             OK     (took 0.26 seconds)
 
-There are also multiple log files (including for ABC, ODIN and VPR), which by convention the script names with the ``.out`` suffix:
+There are also multiple log files (including for ABC, Parmys and VPR), which by convention the script names with the ``.out`` suffix:
```
AlexandreSinger, line 349: Same comment as above, is ABC log files still generated?

doc/src/quickstart/index.rst
```diff
@@ -462,10 +368,10 @@ You will also see there are several BLIF files produced:
 
     > ls *.blif
 
-    0_blink.abc.blif   0_blink.raw.abc.blif  blink.odin.blif
-    0_blink.odin.blif  blink.abc.blif        blink.pre-vpr.blif
+    0_blink.abc.blif   0_blink.raw.abc.blif  blink.parmys.blif
```
AlexandreSinger, line 371: Are these abc blif files still there?

doc/src/quickstart/index.rst
```diff
@@ -370,7 +276,7 @@ The resulting command is:
 
     > $VTR_ROOT/vpr/vpr \
         $VTR_ROOT/vtr_flow/arch/timing/EArch.xml \
-        blink --circuit_file blink.pre-vpr.blif \
+        ~/vtr_work/quickstart/blink_manual/temp/blink.parmys.blif \
```
AlexandreSinger, line 279: Why are you passing the full file path? If someone is following your steps correctly, should they not already be in the `~/vtr_work/quickstart/blink_manual/` folder? If you do want to make it explicit, I would recommend creating a `$WRK_DIR` variable, similar to `$VTR_ROOT` instead of passing the whole path.

doc/src/quickstart/index.rst
```diff
@@ -379,15 +285,15 @@ and after VPR finishes we should see the resulting implementation files:
...
+    blink.parmys.net  blink.parmys.place  blink.parmys.route
 
 We can then view the implementation as usual by appending ``--analysis --disp on`` to the command:
 
 .. code-block:: bash
 
     > $VTR_ROOT/vpr/vpr \
         $VTR_ROOT/vtr_flow/arch/timing/EArch.xml \
-        blink --circuit_file blink.pre-vpr.blif \
+        ~/vtr_work/quickstart/blink_manual/temp/blink.parmys.blif \
```
AlexandreSinger, line 296: Same path comment as above.

doc/src/quickstart/index.rst
```diff
@@ -485,6 +391,135 @@ which we can visualize with:
         --route_chan_width 100 \
         --analysis --disp on
 
+Manually Running VTR with ODIN II
+----------------------------------
+VTR includes a second synthesis tool, ODIN II. Below we explain how to run this alternative synthesis flow.
```
AlexandreSinger, line 396: I would say "we explain how to manually run". Just to be clear that this is not the recommended way, but its possible if you want to avoid the scripts.

doc/src/quickstart/index.rst
```diff
@@ -485,6 +391,135 @@ which we can visualize with:
...
+.. code-block:: bash
+
+    > mkdir -p ~/vtr_work/quickstart/blink_manual
+    > cd ~/vtr_work/quickstart/blink_manual
+
+To synthesize your Verilog design with ODIN II, you need to build VTR with the following command:
+
+.. code-block:: bash
+
+    > make CMAKE_PARAMS="-DWITH_ODIN=on"
```
AlexandreSinger, line 409: This is confusing since the line above tells the user to change directory to `~/vtr_work/quickstart/blink_manual` and now you are saying to use the make command. This should be in the VTR directory. I would move the build instructions above the prior code block and provide the `cd` command to go to the correct directory.

AlexandreSinger (commented):
@ZohairZaidi Thank you for these changes! I like how you are running Yosys. I think using the vtr_flow script just on the Yosys step is perfectly fine since running Yosys directly is such a pain. The main goal is to know how to make the .blif file.

I have left some comments below, but overall the documentation is very good.

doc/src/quickstart/index.rst
ZohairZaidi: it is used when we used synthesize with ODIN.

AlexandreSinger: Yes, but these bullet points are in the "Manually Running the VTR Flow" section; not the with ODIN section.

vaughnbetz: Agreed. This should be refactored to make clear you either need to run Odin II + ABC, or parmys (which will internally call ABC itself).

vaughnbetz:
Adding @soheilshahrouz in case he can help figure out the required flow ... we need to document this well in the quickstart as Mohamed and I were confused about it.

ZohairZaidi:
I agree @vaughnbetz, I plan to work on it this reading week. Please let me know what you were confused about.

doc/src/quickstart/index.rst
```diff
@@ -485,6 +410,135 @@ which we can visualize with:
...
+
+This step enables ODIN II support in the VTR build process, which is essential for the subsequent synthesis operations.
+
+Lets make a new directory for us to work in:
+
+.. code-block:: bash
+
+    > mkdir -p ~/vtr_work/quickstart/blink_manual
+    > cd ~/vtr_work/quickstart/blink_manual
```
ZohairZaidi, line 431: I was wondering if this is necessary. I would assume that if they opt with ODIN they would skip the synthesizing with Parmys section so I left it here.

doc/src/quickstart/index.rst
vaughnb-cerebras: I think you were right and I was wrong @ZohairZaidi !  After talking to Ken Kent at UNB, it does seem we need to run ABC and the clock add/removal script after Parmys.  So I think your original edits were right.  Can you run the flow to check exactly what happens in run_vtr_flow with parmys?   I think it runs ABC and the clock recovery script after all, so we should document them again.

I also think we should change the order of the text.  After saying how to run vpr, I think we should say how to use run_vtr_flow with parmys to synthesize, place and route a circuit. Then we can have a "running manually" section to dig more into how to run each tool.  Finally we can have the existing Odin II section.

ZohairZaidi:
Hi @vaughnbetz, let me know how this is. I changed the order and made it smoother for the user to follow.

doc/src/quickstart/index.rst
```diff
@@ -73,6 +65,102 @@ from the VTR root directory (hereafter referred to as :term:`$VTR_ROOT`) to buil
...
+.. code-block:: bash
+
+    > $VTR_ROOT/vpr/vpr \
+        $VTR_ROOT/vtr_flow/arch/timing/EArch.xml \
+        blink --circuit_file $VTR_FLOW_DIR/temp/blink.pre-vpr.blif \
+        --route_chan_width 100 \
+        --analysis --disp on
+
 
 Running VPR
```
vaughnbetz, line 165: I think we should rename this to "Running VPR Manually"
and then add a first sentence after the heading that says
"Sometimes you may wish to run only the vpr (placement, routing and timing analysis) parts of the flow rather than the full VTR flow (which includes synthesis). To show how to do this, let's now try taking ..."

vaughnbetz (approved):
This looks good, thanks!  I just have a couple of minor comments / changes to link up the first couple of sections.

vaughnbetz:
Looks good, thanks!

PR #2778, opened 2024-10-16
Add Z Cost to Placement Cost Function [WIP]

amin1377:
When using 3D architectures, two types of bounding boxes are available: per-layer and cube bounding boxes. For the cube bounding box, the cost function previously depended only on the bounding box's dx and dy dimensions, with dz being ignored. In this PR, we added chanz_place_cost_fac_ to the NetCostHandler class, which contains the inverse of the average number of inter-die connections for each [subheight][sublow] pair. This data structure is multiplied by the bounding box’s dz and added to the placement cost.
QoR:
3D SIV without Z cost vs 3D SIV with Z cost:
3d


vaughnbetz:
Please do a Titan check on a 2D arch just to be sure, and mail me the results. But I'll merge in advance.  Also, in the subsequent PR for prefix sum, please make sure there are Doxygen comments on new subroutines.

Issue #2767, opened 2024-10-11
Extend XY_SPECIFIED syntax support to fixed layout XML files


petergrossmann21:
The recent pull request for XY_SPECIFIED custom switch block specification https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2664 was tested with an architecture with an <auto_layout> tag.

In the use case that I have for this feature the requirement will be to use it on layouts using the <fixed_layout> tag instead.  Testing on an architecture of this type yielded the following error

`Error 1: /projects/zeroasic/fpgaa_working/upgrade_custom_switchblock_xml/fpga_architect/test/fpga_gamma_vanilla_baby_falcon/cad/fpga_gamma_vanilla_baby_falcon.xml:489 Specifying SB locations for auto layout devices are not supported yet!`

It would appear that the error message is checking for the wrong layout type (or reporting the wrong layout type).

@saaramahmoudi , is the feature expected to work for both layout types?  If so, we can pick some fixed layout architecture and add an additional test case as part of closing the error message.

In the event that I'm the only user of the feature in this use case I may be able to contribute to the development of a fix, as long as we are lenient with respect to schedule.

saaramahmoudi:
Hi @petergrossmann21, 

The feature is not supported for auto_layout yet, but the process is very easy. When we are parsing the custom switchblock specification in read_xml_arch.cpp, I don't know what it the grid_width and grid_height if we are using auto_layout feature (since that is decided in packer step not in the parsing step). The solution would be to save these information from architecture file as a string in read_xml_arch.cpp and perform the actual parsing in rr_graph.cpp when the packer has already finished, meaning that we just need to move my code from one place to another. 
I was planning to do that with my next PR. However, I am a bit busy right now and can't do that immediately. 
Hope that helps and let me know if you have more questions.

saaramahmoudi:
@petergrossmann21 I used one of the existing architecture to create the task, however I passed --device in task config file if you look into this file: [strong_custom_sb_loc](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_custom_sb_loc/config/config.txt)

petergrossmann21:
Update:  I am able to reproduce the error message described above with the provided vtr_flow/arch/custom_sb_location/stratixiv_arch.timing.xml

petergrossmann21:
[...]

PR #2779, opened 2024-10-17
Clarify corner case behavior of --device option

petergrossmann21:
#### Description
This PR provides an update to the documentation of the --device command line option.

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2767

#### Motivation and Context
The goal for these changes is to help readers get a more complete understanding of when/how to specify --device.  Existing documentation is biased toward the common case 

I performed local testing on Zero ASIC architecture models to confirm that, when no ``<auto_layout>`` tag is present in the architecture file, VPR identifies the smallest of multiple ``<fixed_layout>`` tags for which packing is successful for ``--device auto``.  Architectures of this type appear to be present in the available suite of files under ``vtr_flow/arch``.  I also locally verified with ZA architecture models that VPR can find the ``<auto_layout>`` tag when it is listed after <fixed_layout> tags and ``--device auto`` is selected.  To the best of my ability to determine with manual inspection, an architecture file of this type is not available in the VPR repo.  Unfortunately the ZA test case is not currently available for external use.

If we wish to gate this PR by adding such tests to the CI suite I will need a quick walkthrough of how to do this efficiently.

#### Types of changes
Documentation only

#### Checklist:
- [X] I have updated the documentation accordingly


AlexandreSinger (approved):
LGTM! Thank you so much @petergrossmann21 for looking into this. @vaughnbetz what do you think.

I did verify the claim that "VPR uses the smallest device amongst all `fixed_layout` specifications into which the design can be packed" when no auto_layout is specified and auto is selected. The code doing this can be found here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/dc3e5dc8aa31d6497817e33cd969b08a18542da5/vpr/src/base/SetupGrid.cpp#L141-L256
I remembered seeing this code when I was working on the ClusterLegalizer. When the packer creates a cluster, it checks if it needs more clusters than are currently available, if it needs more space it calls a function to grow the device. The function above is what decides should happen if the "auto" keyword is specified. Just thought people would find it interesting.

doc/src/vpr/command_line_usage.rst
```diff
@@ -200,12 +200,19 @@ General Options
 
 .. option:: --device <string>
 
-    Specifies which device layout/floorplan to use from the architecture file.
+    Specifies which device layout/floorplan to use from the architecture file.  Valid values are:
 
-    ``auto`` uses the smallest device satisfying the circuit's resource requirements.
-    Other values are assumed to be the names of device layouts defined in the :ref:`arch_grid_layout` section of the architecture file.
+    * ``auto`` VPR uses the smallest device satisfying the circuit's resource requirements, with device layout specified using the ``auto_layout`` section of the architecture file.
```
vaughnbetz, line 205: I'd delete the part after a comma and instead say (taken from your comments below):

... resource requirements. This option will use the ``<auto_layout>`` tag if it is present in the architecture file in order to construct the smallest FPGA that has sufficient resources to fit the design. If the ``<auto_layout>`` tag is not present, the ``auto`` option chooses the smallest device amongst all the architecture file's ``fixed_layout`` specifications into which the design can be packed.

doc/src/vpr/command_line_usage.rst
```diff
@@ -200,12 +200,19 @@ General Options
...
+    * Any string matching ``name`` attribute of a device layout defined with a ``fixed_layout`` tag in the :ref:`arch_grid_layout` section of the architecture file.
 
-    .. note:: If the architecture contains both ``<auto_layout>`` and ``<fixed_layout>`` specifications, specifying an ``auto`` device will use the ``<auto_layout>``.
+    If the value specified is neither ``auto`` nor matches the ``name`` attribute value of a ``<fixed_layout>`` tag, VPR issues an error.
+
+    In the event that no ``<auto_layout>`` tag is present in the architecture file, this option has the following behaviour:
```
vaughnbetz, line 210: I would delete thse 3 lines as now they are covered by the introduction of --auto above.

vaughnbetz (changes requested):
Thanks!  Looks good; I just suggest a bit of word tweaking (see comments).

doc/src/vpr/command_line_usage.rst
petergrossmann21: @vaughnbetz I kept the line documenting how VPR does error handling here, but deleted the bullet points below the highlights; I believe this better captures the desired elimination of redundancy.  Let me know if additional changes are desired.

vaughnbetz:
Looks good, thanks!

PR #2780, opened 2024-10-17
[Task] Added Ability to Constrain Circuits in Config

AlexandreSinger:
The VTR task interface uses configuration files to specify what circuits to run for a given task. The current implementation of these config files only allowed the user to run all of their circuits on the same device and using the same flat constraint file (for example fixing the IOs). This was only able to be done through specifying them in the script_params option, but this was limited.

This causes problems for the AP flow since each circuit has different IO constraints and may be constrained to different devices or architectures. The only way to use the VTR tasks is to create one configuration file per circuit, which is very wasteful and challenging to work with. It also makes parsing the QoR more difficult.

Added a new configuration option that can be added to the configuration file which can constrain a circuit to a given option. The syntax is as follows:
  circuit_constraint_list_add=(<circuit>, <constr_key, constr_val>)
This line will tell VTR run task to constrain the circuit on the constr_key option.

The currently supported constraint keys are:
        - arch: Constrain the circuit to only run on the given arch
        - device: Constrain the circuit to only use the given device
        - constraints: Constrain the circuit's atom placement

If a constraint is not specified, the bahaviour is unchanged. So, for example, if an arch constraint is not specified, the circuit will run on all the architectures in the arch_list (as usual).

Future work is to support constraining the route_chan_width so different circuits can be run with different channel widths in the same config file. This can easily be added using this interface.

Example of the interface in use:
Screenshot from 2024-10-17 16-12-47


AlexandreSinger:
@vaughnbetz This is passing CI. The only test that it is failing is Python Lint due to me having to add 5 lines to the create_job method (which was already quite large); and now the linter is complaining that the function is 2 lines too long now:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/dc3e5dc8aa31d6497817e33cd969b08a18542da5/vtr_flow/scripts/python_libs/vtr/task.py#L564-L672

I am thinking of allowing this function to be a bit long by disabling this check for this function. What do you think? If not it may be a bit of work to clean up the method.

vaughnbetz:
I'd rather either factor out something from the function, or, if you think that is a pain, just change the linting parameters to allow a few more statements per function.  I'd rather not pragma off linting for a specific function, which could then grow without bound.

AlexandreSinger:
@vaughnbetz I took your advice and outlined my contributions to the function into its own method. If we add more constraints in the future, this will make it easier so I think that was a good idea. The linter is now passing and the code is ready for review.

vaughnbetz (approved):
Looks good to me!  I second the to-do to add a path to the constraints to clean up the syntax some.

PR #2781, opened 2024-10-18
Chan z prefix sum

amin1377:
#### Description
This PR optimizes the calculation of the `ChanZ` cost factor by precomputing and storing the cumulative number of inter-die connections prior to block placement. During placement, the number of inter-die connections within a bounding box is calculated using the following formula:

```cpp
if (x_low == 0 && y_low == 0) {
    num_inter_dir_conn = acc_tile_num_inter_die_conn_[x_high][y_high];
} else if (x_low == 0) {
    num_inter_dir_conn = acc_tile_num_inter_die_conn_[x_high][y_high] - 
                         acc_tile_num_inter_die_conn_[x_high][y_low - 1];
} else if (y_low == 0) {
    num_inter_dir_conn = acc_tile_num_inter_die_conn_[x_high][y_high] - 
                         acc_tile_num_inter_die_conn_[x_low - 1][y_high];
} else {
    num_inter_dir_conn = acc_tile_num_inter_die_conn_[x_high][y_high] - 
                         acc_tile_num_inter_die_conn_[x_low - 1][y_high] - 
                         acc_tile_num_inter_die_conn_[x_high][y_low - 1] + 
                         acc_tile_num_inter_die_conn_[x_low - 1][y_low - 1];
}
```


vaughnbetz:
1. You should comment how the prefix sum computation works.
2. We can't afford an O(n^2) memory and O(n^2) compute to load this data structure, as N is the size of the circuit.  We shouldn't compute the final matrix, but instead just store the O(n) (2D in x,y) prefix sum. Compute the average z-capacity in a helper function when called.

vaughnbetz:
QoR results are at https://www.eecg.utoronto.ca/~vaughn/vaughnwiki/doku.php?id=amin_weekly_update and look good. On both 3D SW and 2D Titan architectures, pack, place and route time and QoR are essentially unchanged, and on a 3D arch total runtime drops by 2%, which makes sense as the loading of the cost structures should be faster (and may be counted outside place time). @amin1377 : if you can paste the QoR data in for posterity that would be good.

vpr/src/place/net_cost_handler.h
```diff
@@ -196,12 +196,12 @@ class NetCostHandler {
     vtr::NdOffsetMatrix<float, 2> chanx_place_cost_fac_; // [-1...device_ctx.grid.width()-1]
     vtr::NdOffsetMatrix<float, 2> chany_place_cost_fac_; // [-1...device_ctx.grid.height()-1]
     /**
-      @brief This data structure functions similarly to the matrices described above 
-      but is applied to 3D connections linking different FPGA layers. It is used in the 
-      placement cost function calculation, where the height of the bounding box is divided 
-      by the average number of inter-die connections within the bounding box.
+      @brief This data structure stores the cumulative number of inter-die connections from the lower-left corner.
```
vaughnbetz, line 199: I think you can make this more clear to someone who doesn't know what you're doing.  Here's another attempt:

The matrix below is used to calculate a chanz_place_cost_fac based on the average channel width in the cross-die-layer direction over a 2D (x,y) region. We don't assume the inter-die connectivity is the same at all (x,y) locations, so we 
can't compute the full chanz_place_cost_fac for all possible (xlow,ylow)(xhigh,yhigh) without a 4D array, which would 
be too big: O(n^2) in circuit size. Instead we compute a prefix sum that stores the number of inter-die connections per layer from 
(x=0,y=0) to (x,y).  Given this, we can compute the average number of inter-die connections over a (xlow,ylow) to (xhigh,yhigh) region in O(1) (by adding and subtracting 4 entries).
[0..grid_width-1][0..grid_height-1]

vpr/src/place/net_cost_handler.h
```diff
@@ -250,23 +250,17 @@ class NetCostHandler {
...
-     *
-     * @param place_cost_exp It is an exponent to which you take the average inverse channel capacity;
-     * a higher value would favour wider channels more over narrower channels during placement (usually we use 1).
      */
-    void alloc_and_load_chan_w_factors_for_place_cost_(float place_cost_exp);
+    void alloc_and_load_chan_w_factors_for_place_cost_();
 
     /**
     * @brief Allocates and loads the chanz_place_cost_fac array with the inverse of
     * the average number of inter-die connections between [subhigh] and [sublow].
```
vaughnbetz, line 258: This comment looks wrong ... say what it loads.

vpr/src/place/net_cost_handler.h
```diff
@@ -511,4 +505,14 @@ class NetCostHandler {
      */
     double get_net_wirelength_from_layer_bb_(ClusterNetId net_id);
 
+    /**
+     * @brief Calculate the chanz cost factor based on the inverse of the average number of inter-die connections 
+     * in the given bounding box. This cost factor increases the placement cost for blocks that require inter-layer 
+     * connections in areas with, on average, fewer inter-die connections. If inter-die connections are evenly 
+     * distributed across tiles, the cost factor will be the same for all bounding boxes.
```
vaughnbetz, line 512: "... all bounding boxes, but it will still weight z-directed vs. x- and y-directed connections appropriately."

vpr/src/place/net_cost_handler.cpp
```diff
@@ -229,22 +230,32 @@ void NetCostHandler::alloc_and_load_chan_w_factors_for_place_cost_(float place_c
...
+    * of being made from the architecture file, we calculate this number by iterating over RR graph. Once 
+    * tile_num_inter_die_conn is populated, we can start populating acc_tile_num_inter_die_conn_. First,  
+    * we populate the first row and column. Then, we iterate over the rest of blocks and get the number of 
+    * inter-die connections by adding up the number of inter-die block at that location + the accumulative 
+    * for the  block below and  left to it. Then, since the accumulative number of inter-die connection to 
+    * the block on the lower left connection of the block is added twice, that part needs to be removed.
+    */
     for (const auto& src_rr_node : rr_graph.nodes()) {
         for (const auto& rr_edge_idx : rr_graph.configurable_edges(src_rr_node)) {
             const auto& sink_rr_node = rr_graph.edge_sink_node(src_rr_node, rr_edge_idx);
```
soheilshahrouz, line 261: Iterating integers and small types by reference is not necessary. Using explicit type names instead of `auto` makes the code easier to read.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -271,24 +282,24 @@ void NetCostHandler::alloc_and_load_for_fast_vertical_cost_update_(float place_c
         }
```
soheilshahrouz, line 282: streched -> stretched

soheilshahrouz, line 282: both operands of `&&` in lines 279 and 277 are the same.

soheilshahrouz, line 282: Can't two inner loops be replaced with a single loop that iterates over all edges of the node?

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1581,6 +1592,44 @@ double NetCostHandler::get_net_wirelength_from_layer_bb_(ClusterNetId net_id) {
     return ncost;
 }
 
+float NetCostHandler::get_chanz_cost_factor(const t_bb& bounding_box) {
+    float place_cost_exp = placer_opts_.place_cost_exp;
+    int x_high = bounding_box.xmax;
+    int x_low = bounding_box.xmin;
+    int y_high = bounding_box.ymax;
+    int y_low = bounding_box.ymin;
```
soheilshahrouz, line 1600: I think adding these new variable that separate member variables of bounding_box is not necessary. Without them, the code will be shorter and easier to read.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1581,6 +1592,44 @@ double NetCostHandler::get_net_wirelength_from_layer_bb_(ClusterNetId net_id) {
...
+                             acc_tile_num_inter_die_conn_[x_high][y_low-1];
+    } else if (y_low == 0) {
+        num_inter_dir_conn = acc_tile_num_inter_die_conn_[x_high][y_high] - \
+                             acc_tile_num_inter_die_conn_[x_low-1][y_high];
+    } else {
+        num_inter_dir_conn = acc_tile_num_inter_die_conn_[x_high][y_high] - \
+                             acc_tile_num_inter_die_conn_[x_low-1][y_high] - \
+                             acc_tile_num_inter_die_conn_[x_high][y_low-1] + \
+                             acc_tile_num_inter_die_conn_[x_low-1][y_low-1];
+    }
```
soheilshahrouz, line 1617: using `vtr::NdOffsetMatrix` will be consistent with thest of file and will get rid of this if statement.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1581,6 +1592,44 @@ double NetCostHandler::get_net_wirelength_from_layer_bb_(ClusterNetId net_id) {
...
+    
+    int bb_num_tiles = (x_high - x_low + 1) * (y_high - y_low + 1);
```
soheilshahrouz, line 1619: move the declartion to where it is used (inside else statement)

vpr/src/place/net_cost_handler.h
```diff
@@ -511,4 +505,14 @@ class NetCostHandler {
...
+     * @param bounding_box Bounding box of the net which chanz cost factor is to be calculated
+     * @return ChanZ cost factor
+     */
+    float get_chanz_cost_factor(const t_bb& bounding_box);
```
soheilshahrouz, line 516: Add an underline to the end of function name to be consistent with the rest of file.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1478,7 +1489,7 @@ double NetCostHandler::get_net_cube_bb_cost_(ClusterNetId net_id, bool use_ts) {
     ncost = (bb.xmax - bb.xmin + 1) * crossing * chanx_place_cost_fac_[bb.ymax][bb.ymin - 1];
```
soheilshahrouz, line 1489: `is_multi_layer` has the same value in all calls. It can be a member varaible initialized in constructor (or one of the member functions it calls). This would save us an integer comparison for each call to this function.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1478,7 +1489,7 @@ double NetCostHandler::get_net_cube_bb_cost_(ClusterNetId net_id, bool use_ts) {
...
     ncost += (bb.ymax - bb.ymin + 1) * crossing * chany_place_cost_fac_[bb.xmax][bb.xmin - 1];
     if (is_multi_layer) {
-        ncost += (bb.layer_max - bb.layer_min) * crossing * chanz_place_cost_fac_[bb.xmax][bb.ymax][bb.xmin][bb.ymin];
+        ncost += (bb.layer_max - bb.layer_min) * crossing * get_chanz_cost_factor(bb);
```
soheilshahrouz, line 1492: Can't we multiply `crossing` once at the end? It would save us two doulbe precision multiplications

vpr/src/place/net_cost_handler.cpp
vaughnbetz: Yes, that would be cleaner.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -229,22 +230,32 @@ void NetCostHandler::alloc_and_load_chan_w_factors_for_place_cost_(float place_c
...
     const auto& rr_graph = device_ctx.rr_graph;
     
     const size_t grid_height = device_ctx.grid.height();
     const size_t grid_width = device_ctx.grid.width();
 
 
-    chanz_place_cost_fac_ = vtr::NdMatrix<float, 4>({grid_width, grid_height, grid_width, grid_height}, 0.);
+    acc_tile_num_inter_die_conn_ = vtr::NdMatrix<int, 2>({grid_width, grid_height}, 0.); 
 
     vtr::NdMatrix<float, 2> tile_num_inter_die_conn({grid_width, grid_height}, 0.);
```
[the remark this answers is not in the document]
vaughnbetz: It's a count, so int is probably better.  Float could have some round-off issues with big chips, as this matrix is filled in by adding small numbers to a running count, which can get problematic if the big number ever becomes more than 16 million times or so bigger than the small number (the small number then gets thrown away by round off).

vaughnbetz:
This is separate from this PR, but it's possible that using the same technique in the x- and y- directions that you're using for the z-direction to calculate the average channel width by adding and subtracting the proper values would be faster for x- and y- too.  

It saves memory (would take storage for x and y average channel capacities down from O(device size) to O(sqrt(device size)), which would help the cache and memory bandwidth. If would mean two matrix accesses instead of 1, and an extra reciprocal and pow() call, but it might still be overall just as fast or faster.  After this PR is landed, if you have th energy for it you could do a quick experiment to check by recoding the x- and y- dimensions to match the z-dimension approach.

soheilshahrouz:
> This is separate from this PR, but it's possible that using the same technique in the x- and y- directions that you're using for the z-direction to calculate the average channel width by adding and subtracting the proper values would be faster for x- and y- too.

@amin1377 If you're busy with other tasks, I am interested in working on this idea for chan_x and chan_y. Let me know if I can take it over.

amin1377:
> > This is separate from this PR, but it's possible that using the same technique in the x- and y- directions that you're using for the z-direction to calculate the average channel width by adding and subtracting the proper values would be faster for x- and y- too.
> 
> @amin1377 If you're busy with other tasks, I am interested in working on this idea for chan_x and chan_y. Let me know if I can take it over.

How could I possibly say no to such a generous offer, Soheil? Thanks a lot! :) Let me know if I can be of any help.

vpr/src/place/net_cost_handler.cpp
amin1377: change the loops to the following format:

```
for (const auto& src_rr_node : rr_graph.nodes()) {
        for (auto edge_range: {rr_graph.configurable_edges(src_rr_node), rr_graph.non_configurable_edges(src_rr_node)}) {
            for (const auto& rr_edge_idx : edge_range) {
                const auto& sink_rr_node = rr_graph.edge_sink_node(src_rr_node, rr_edge_idx);
                if (rr_graph.node_layer(src_rr_node) != rr_graph.node_layer(sink_rr_node)) {
                    // We assume that the nodes driving the inter-layer connection or being driven by it
                    // are not stretched across multiple tiles
                    int src_x = rr_graph.node_xhigh(src_rr_node);
                    int src_y = rr_graph.node_yhigh(src_rr_node);
                    VTR_ASSERT(rr_graph.node_xlow(src_rr_node) == src_x && rr_graph.node_ylow(src_rr_node) == src_y);

                    tile_num_inter_die_conn[src_x][src_y]++;
                }
            }
        }
    }
```

vpr/src/place/net_cost_handler.cpp
amin1377: I think the code would be easier to read with these variables. As you can see in the following lines, I use these variables in various places. Without them, I'd have to repeatedly use the `bounding_box.` prefix, which would make the lines unnecessarily long. However, I don’t have a strong opinion on this. If you still think I should remove the variables, please let me know.

vpr/src/place/net_cost_handler.cpp
amin1377: @soheilshahrouz, if you remember, my initial attempt was to use this data structure, but I encountered some memory issues with it. I’ll create an issue for this and, in another PR, attempt to replace this data structure with `vtr::NdOffsetMatrix`.

soheilshahrouz: I think I figured out what the problem was.
Check if 68ecb55 sovles the memory issue.

vpr/src/place/net_cost_handler.cpp
soheilshahrouz: What do think about renaming `bounding_box` to `bb`? There are a few other methods in this calss that use the same name.

vpr/src/place/net_cost_handler.cpp
soheilshahrouz: I am not sure about this, but isn't `edges()` the union of `configurable_edges` and `non_configurable_edges`?

vpr/src/place/net_cost_handler.cpp
amin1377: Ok.

vpr/src/place/net_cost_handler.cpp
amin1377: You are right. For some reason, I missed this function. Thanks!

vpr/src/place/net_cost_handler.cpp
amin1377: Sounds good. I'll try it once the PR is merged.

amin1377:
QoR:
titan other 3d CB cube bb
3d_titan_other_full_opin_cube_bb
titan_quick_qor
titan_quick_qor

vaughnbetz:
QoR looks good.

amin1377:
@vaughnbetz, I’ve addressed both your comments and Soheil’s. After making the changes, I reran the benchmarks to ensure everything is still working correctly, and the tests passed without issues. I think the PR is ready to be merged.

vaughnbetz (approved):

Issue #2677, opened 2024-08-06
VTR format checking / auto-format not checked by CI?


vaughnbetz:
Various PRs have shown formatting changes to break long multi-lines.  I think make format should do this automatically; this issue is to see if we have a missing rule in autoformat, or even worse, if auto-format is breaking manually made format changes for readability.

We should also have CI checking that our formatting is consistent.
Assigning to @soheilshahrouz since he has been cleaning up some code formatting; I think it would be better to figure out if there is a shortcoming in make format.  I think @jgoeders knows the autoformatting, so he could potentially help if you're stuck on how to improve it.

#### Expected Behaviour
make format should format long subroutine calls or subroutine variable declarations properly.  CI should test that we are format compliant.

#### Current Behaviour
I'm not 100% sure, but it seems long lines aren't caught by CI, and perhaps are not fixed by make format.

soheilshahrouz:
@vaughnbetz
Currently, CI doesn't check for long lines and `make format` does not break them into multiple shorter lines. We can set a line length limit in `.clang-format` file so that lines longer than the limit cause a CI failure. I think 80 is too short for widescreen displays. What do you thing about setting the line length limit to 120 characters?

vaughnbetz:
I'd be fine with 120 characters.

soheilshahrouz:
Closing this issue as it was resolved by merging #2935

PR #2790, opened 2024-10-28
Fix clang format not working in CI

soheilshahrouz:
Solves [issue 2667](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2677)


libs/libarchfpga/src/device_grid.cpp
```diff
@@ -18,19 +18,15 @@ DeviceGrid::DeviceGrid(std::string grid_name,
 size_t DeviceGrid::num_instances(t_physical_tile_type_ptr type, int layer_num) const {
     size_t count = 0;
     //instance_counts_ is not initialized
-    if (instance_counts_.empty()) {
-        return 0;
-    }
+    if (instance_counts_.empty()) { return 0; }
```
soheilshahrouz, line 21: AllowShortIfStatementsOnASingleLine: true

libs/libarchfpga/src/histogram.cpp
```diff
@@ -9,17 +9,16 @@
...
     if (values.empty()) return histogram;
 
-    if (std::isnan(min_value)) {
-        min_value = *std::min_element(values.begin(), values.end());
-    }
-    if (std::isnan(max_value)) {
-        max_value = *std::max_element(values.begin(), values.end());
-    }
+    if (std::isnan(min_value)) { min_value = *std::min_element(values.begin(), values.end()); }
+    if (std::isnan(max_value)) { max_value = *std::max_element(values.begin(), values.end()); }
```
soheilshahrouz, line 21: AllowShortIfStatementsOnASingleLine: true

vpr/src/base/check_netlist.cpp
```diff
@@ -48,20 +48,19 @@ void check_netlist(int verbosity) {
     /* Check that nets fanout and have a driver. */
     int global_to_non_global_connection_count = 0;
     for (auto net_id : cluster_ctx.clb_nlist.nets()) {
-        h_net_ptr = insert_in_hash_table(net_hash_table, cluster_ctx.clb_nlist.net_name(net_id).c_str(), size_t(net_id));
+        h_net_ptr
+            = insert_in_hash_table(net_hash_table, cluster_ctx.clb_nlist.net_name(net_id).c_str(), size_t(net_id));
```
soheilshahrouz, line 52: BreakBeforeBinaryOperators: All

soheilshahrouz:
I changed ColumnLimit from 0 (unlimited) to 120. I suggest a few other changes in `.clang-format` file. Let me know whay your thought are. For each suggestion, I added a few comments to show how it changes the code.

-`AllowShortIfStatementsOnASingleLine` should be set to false. 
-`BreakBeforeBinaryOperators` should be set to NonAssignment.
- Add `AllowShortEnumsOnASingleLine: false`

libs/libarchfpga/src/device_grid.cpp
vaughnbetz: I don't like single line ifs -- turning it off makes sense to me.

libs/libarchfpga/src/histogram.cpp
vaughnbetz: Turn off ! (Ugh.)

vpr/src/base/check_netlist.cpp
vaughnbetz: Ugly. Prefer the BreakBeforeBinaryOperators: NonAssignment to keep the = on the first line.

vpr/src/pack/cluster_util.h
```diff
@@ -20,30 +20,17 @@ struct AtomContext;
...
-};
+constexpr int AAPACK_MAX_HIGH_FANOUT_EXPLORE
+    = 10; /* For high-fanout nets that are ignored, consider a maximum of this many sinks, must be less than packer_opts.feasible_block_array_size */
+constexpr int AAPACK_MAX_TRANSITIVE_EXPLORE
+    = 40; /* When investigating transitive fanout connections in packing, consider a maximum of this many molecules, must be less than packer_opts.feasible_block_array_size */
+
+enum e_gain_update { GAIN, NO_GAIN };
+enum e_feasibility { FEASIBLE, INFEASIBLE };
+enum e_gain_type { HILL_CLIMBING, NOT_HILL_CLIMBING };
+enum e_removal_policy { REMOVE_CLUSTERED, LEAVE_CLUSTERED };
```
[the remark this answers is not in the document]
vaughnbetz: No strong preference from me on this one.

vaughnbetz:
Adding @AlexandreSinger and @duck2 in case they want to weigh in.

libs/libarchfpga/src/histogram.cpp
AlexandreSinger: I agree with Vaughn. I think it should not be required. Its nice if it was optional, for certain places, but I do not want to make this required. The way Clang Format is set up it will automatically set it to this version of the line is small enough.

libs/libarchfpga/src/device_grid.cpp
AlexandreSinger: I agree. Please turn this off. I would prefer if the above were turned into:
```cpp
if (instance_counts_.empty())
    return 0;
```
The curly braces are not necessary.

vpr/src/base/check_netlist.cpp
AlexandreSinger: I am indifferent on this one. My personal preference for this line would be:
```cpp
h_net_ptr = insert_in_hash_table(net_hash_table, 
                                 cluster_ctx.clb_nlist.net_name(...),
                                 ...)
```
Or the line above should use more temporary variables to make this look nicer (which is good practice on the programmer's part).

vpr/src/pack/cluster_util.h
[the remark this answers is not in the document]
AlexandreSinger: I am also indifferent on this one. I actually think its nice; but I do understand if we want consistency.

AlexandreSinger:
Hi @soheilshahrouz , thank you so much for fixing this! I cannot believe the problem was just a submodule issue. I worry that this PR will trample all PRs currently in flight (it will cause conflicts with all code); but I think it is so worth it.

I was looking at how it reformatted my AP code (because my coding style is perfect 😁) and beyond the if statement to one line thing (which I mentioned above as bad), I agree with most of the changes.

I do have a couple of comments:
- **May I request that we use a line width of 80 instead of 120?** I know that its a bit of a meme that I like to follow the LLVM style guide, but I agree with their reasoning as to why to choose 80: https://llvm.org/docs/CodingStandards.html#source-code-width . This is also in agreement with the Google C++ style-guide as well: https://google.github.io/styleguide/cppguide.html#Line_Length . However, I will not kill for this. If the people want 120, so be it. I am just from the class of people that believe that well-written code should never be longer than 80 columns (if it must be, then you can refactor your code better; i.e. more functions and local variables for readability). The issue for me is that I prefer 80 characters, and when I write for 80 characters, this clang format will "fix" it to 120... 
- **May I request that we add "InsertNewlineAtEOF: true"?** This adds new lines to the end of files. The reason you would want this is that git sometimes gets confused when you modify files in a commit and then undo your modifications. It then "pretends" to delete the last line that does not exist and causes random git conflicts. I strongly believe that all files should end with a new line character to prevent this.
- I also noticed that Odin-II is being tracked by the clang format, was this intentional? Why are we ensuring that Odin-II follows this style guide? If we do not care about it, we should prevent changing it.

Sorry for the long response, I am very passionate about this!

AlexandreSinger:
Don't let my message above discourage you Soheil though, I really really really like this change and am excited to get this merged in! I like it when code is consistent throughout the entire repo!

vaughnbetz:
I think 80 characters is too wrenching; we have a lot of long variable names (which I generally like), and made the decision long ago to abandon 80 characters (partially pushed by Google actually, so I don't think 80 characters is a universal rule there).  I think if we went to 80 characters it would be a big make work project.

I think anything controversial should be run by the Thursday meeting too, and an 80 character limit would be in that category in my opinion.

AlexandreSinger:
@vaughnbetz Understood! I do agree this is something to bring up in the weekly meeting; however, this **should not block this PR from getting merged in**. As you say, the code is already quite wide, so reducing it now will be just as complicated as reducing it later.

Push come to shove you can change small portions of the clang style in sub-projects (but I do not think the fact that I "want it to be 80" is enough argument to not follow the VTR style guide).

vaughnbetz:
@soheilshahrouz nag, nag,

soheilshahrouz:
Superceded by #2931

PR #2792, opened 2024-10-30
[CI] Turned Off Nightly Tests on PR and Merge

AlexandreSinger:
The Nightly Tests have not worked for weeks. There is no reason to keep trying to run them everytime we raise a PR.

I still left the cron job on so we can see if it ever comes back.

This just makes merging PRs easier.


AlexandreSinger:
@vaughnbetz This has passed CI. Please review and merge when you have a moment. I am turning off the nightly test to make it easier to tell if a real test is actually failing or not.

PR #2793, opened 2024-10-30
Random number generation classes

soheilshahrouz:
This PR replaces random number generation function in `vtr` namespace with clasess.

### DIFF
--- a/libs/libvtrutil/src/vtr_random.cpp
+++ b/libs/libvtrutil/src/vtr_random.cpp
@@ -1,91 +1,56 @@
-#include <cstddef>
 
 #include "vtr_random.h"
 #include "vtr_util.h"
 #include "vtr_error.h"
-#include "specrand.h"
-
-#define CHECK_RAND
 
 namespace vtr {
-/* Portable random number generator defined below.  Taken from ANSI C by  *
- * K & R.  Not a great generator, but fast, and good enough for my needs. */
 
-constexpr size_t IA = 1103515245u;
-constexpr size_t IC = 12345u;
-constexpr size_t IM = 2147483648u;
-
-static RandState random_state = 0;
-
-/**
- * @brief The pseudo-random number generator is initialized using the argument passed as seed.
- */
-void srandom(int seed) {
-    random_state = (unsigned int)seed;
-#ifdef SPEC_CPU
-    /* SPEC CPU requires a different random number generator */
-    spec_init_genrand((unsigned long)seed);
-#endif
+RandomNumberGenerator::RandomNumberGenerator(int seed) {
+    random_state_ = (unsigned int)seed;
 }
 
-/* returns the random_state value */
-RandState get_random_state() {
-    return random_state;
+RandomNumberGenerator::RandomNumberGenerator()
+    : RandomNumberGenerator(0) {}
+
+void RandomNumberGenerator::srandom(int seed) {
+    random_state_ = (unsigned int)seed;
 }
 
-int irand(int imax, RandState& state) {
-#ifdef SPEC_CPU
-    /* SPEC CPU requires a different random number generator */
-    return (int)(spec_genrand_int31() % (imax + 1));
-#else
-    /* Creates a random integer between 0 and imax, inclusive.  i.e. [0..imax] */
+int RandomNumberGenerator::irand(int imax) {
+    // Creates a random integer between 0 and imax, inclusive.  i.e. [0..imax]
     int ival;
 
-    /* state = (state * IA + IC) % IM; */
-    state = state * IA + IC; /* Use overflow to wrap */
-    ival = state & (IM - 1); /* Modulus */
+    // state = (state * IA + IC) % IM;
+    random_state_ = random_state_ * IA + IC; // Use overflow to wrap
+    ival = random_state_ & (IM - 1); // Modulus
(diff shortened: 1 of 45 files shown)


soheilshahrouz:
| titan_quick_qor | pack-time   | place time   | place WL    | place cpd     | place_mem    | n_swaps       |
|-----------------|-------------|--------------|-------------|---------------|--------------|---------------|
| master          | 648.6044623 | 1228.921211  | 2497566.214 | 16.78232328   | 5506.560339  | 18118945.13   |
| pr              | 651.6227905 | 1232.363505  | 2497566.214 | 16.78232328   | 5506.772081  | 18118945.13   |
| ratio           | 1.004653573 | 1.002801069  | 1           | 1             | 1.000038453  | 1             |

vpr/src/base/CheckSetup.cpp
```diff
@@ -111,8 +111,8 @@ void CheckSetup(const t_packer_opts& PackerOpts,
         }
     }
 
-    for (i = 0; i < (int)Segments.size(); ++i) {
-        Tmp = Segments[i].arch_opin_switch;
+    for (int i = 0; i < (int)Segments.size(); ++i) {
+        int Tmp = Segments[i].arch_opin_switch;
```
vaughnbetz, line 115: Tmp is a really bad name here ... can we use a better name like opin_switch?

vpr/src/base/CheckSetup.h
```diff
@@ -13,6 +10,6 @@ void CheckSetup(const t_packer_opts& PackerOpts,
                 const t_det_routing_arch& RoutingArch,
                 const std::vector<t_segment_inf>& Segments,
                 const t_timing_inf& Timing,
-                const t_chan_width_dist Chans);
+                const t_chan_width_dist& Chans);
```
vaughnbetz, line 13: If you have the energy for it, these variables that start with a capital are not in line with typical vtr style -- better to start with a lower case variable. But they're not your code so if you want to leave them it's OK.

libs/libvtrutil/src/vtr_random.cpp
```diff
@@ -1,91 +1,67 @@
...
+int RandomNumberGenerator::irand(int imax) {
+    // Creates a random integer between 0 and imax, inclusive.  i.e. [0..imax]
     int ival;
 
-    /* state = (state * IA + IC) % IM; */
-    state = state * IA + IC; /* Use overflow to wrap */
-    ival = state & (IM - 1); /* Modulus */
+    // state = (state * IA + IC) % IM;
+    random_state_ = random_state_ * IA + IC; // Use overflow to wrap
+    ival = random_state_ & (IM - 1); // Modulus
```
vaughnbetz, line 26: Could/should we just use the SPEC random number generator all the time?  I put in the original random number generator, but if the SPEC one has better randomness and isn't noticeably slower we could just make these routines call it.  

I don't have a really strong opinion about this, but maybe we can simplify a code path.

vaughnbetz (commented):
Looks good.
We're passing references to the random class all over the place, but I don't have a great alternative.  Are there any optimization etc. classes that are already passed around that we could add a rng reference to instead?  There may not be a good alternative though, in which case passing this around is fine.

As noted in another comment I made above, maybe we can use the SPEC random utility all the time and delete the body of the vtr one?

Should do a final QoR comparison for safety on Titan anyway before trying to merge this.

libs/libvtrutil/src/vtr_random.cpp
soheilshahrouz: I am afraid that if we use SPEC random generator, we need to update golden results for a large number of tests. We can explore the performance/quality trade-off of these two random generators later.

vaughnbetz:
Looks good.  As time permits we should try moving to one (spec) random number generator.

vaughnbetz:
@amin1377 @heshpdx 

I think spec is using spec_genrand_int31 () to generate random numbers.  This PR ( https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2793/files ) is refactoring things a bit.

From my reading of the code, it looks like it is around an O(1000) computation to generate a random number with spec_genrand_int1, vs. a roughly O(1) computation with the simpler random number generator VPR usually uses.  The implementation by VPR is embedded in the code so it will be the same across platforms, and VPR doesn't need a very good random number generator (fast but not great is OK).

So I'm not sure it is a good idea to switch to the spec one ... it could throw the program execution statistics off some by making random number generation too slow.

This has been merged as we need a class around the random number generators anyway, but I'd like to see if we really should have the SPEC version in the code base.

heshpdx:
If there is a simpler RNG that spec can use, I am all for it. I don't think the spec RNG shows up as taking time in our function profiles of the benchmark, but I will recheck and share what I find.

vaughnbetz:
Thanks @heshpdx .  I think the default vpr random number generator would be fine ... it is fast, and has enough randomness for our needs, and the code is all in vpr (no dependency on a library that could change results on different platforms etc.).

Also, the latest master branch has parallel routing (and parallel timing analysis has been in for a while).  I think you don't like TBB though (which both use); just mentioning in case this is potentially useful for any multi-thread spec benchmarks.  Both algorithms are deterministic, so they should produce results that are as reproducible on different platforms as the serial versions.

Issue #2791, opened 2024-10-30
[Prepacker] Pack Molecule Data Structure Clarity


AlexandreSinger:
The molecule data structure is defined here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/f520c1765987675c9401832f8a0dc64797219a23/vpr/src/base/vpr_types.h#L369-L410

I have noticed that there are some parts of the data structure that are not clear and a bit wasteful. I think this struct should be reworked in the context of the Prepacker to make it easier to work with.

- [ ] The `atom_blk_ids` member is confusing. This is what caused me to look into this. Some of the elements of this vector may be invalid. The reason is that this is **not** a list of atoms in the molecule; it is a mapping between the IDs in the pack pattern and the molecule atoms. The invalid entries are where the pack pattern is not completely full. This is very confusing but also not very good for performance. Instead we should have an ordered list of all atom blocks in the molecule, and then another list which stores the pattern ID for each of those atoms. That or the documentation should be made more clear.
- [ ] The `valid` member is hardly used anymore and should be removed. This variable just says if this molecule has been packed already or not, which is now being stored in the Cluster Legalizer class. This should be removed / phased out.
- [ ] The `num_blocks` member is poorly named. It is not the number of blocks currently in the molecule, it is the maximum number of blocks that can be stored in the molecule. I bet you this was used to allocate a C-style array, but in recent years we moved to a vector. This variable can probably be removed.
- [ ] The `root` member is very strange. Why store the ID in the array instead of the AtomBlockId itself (they are both basically integers). This should be investigated and maybe cleaned up.
- [ ] The `base_gain` member is specific to the clustering algorithm being used. This should not be in the molecule struct in my opinion.
- [ ] The pack pattern part of this struct is a bit confusing. It may be a good idea to encapsulate that into another struct to make it clear.
- [ ] The molecules should not be a linked list. Instead we should have MoleculeIDs which the Prepacker manages. Having unique IDs for each molecule is a good idea and can make traversing the molecules much more efficient. Since we do not care about the order of the molecules list (and we do not expect to be adding or removing molecules often within the list), storing in a standard vector would actually be more performant!
- [ ] The declaration of this type should be moved to the Prepacker. It only makes sense in the context of the Prepacker.
- [ ] We should add an independent verifier class which verifies that the Prepacker created all the molecules correctly and everything works as expected. This also creates a clear interface for what user of the Prepacker should expect to see.

PR #2796, opened 2024-11-02
[Prepacker] Removed Valid Flag From Molecules

AlexandreSinger:
Pack molecules originally contained a flag called "valid" which signified that the molecule has not been clustered yet. This flag is not necessary since the Cluster Legalizer maintains this information internally. Removed the valid flag from the pack molecule struct.

See issue #2791

### DIFF
--- a/vpr/src/pack/cluster_legalizer.cpp
+++ b/vpr/src/pack/cluster_legalizer.cpp
@@ -997,61 +997,28 @@ static void update_molecule_chain_info(t_pack_molecule* chain_molecule, const t_
     VTR_ASSERT(false);
 }
 
-/**
- * @brief Revalidate the molecules associated with this pb. The mol_validated
- *        flag is used to check if the mol has already been validated.
+/*
+ * @brief Reset molecule information created while trying to cluster it.
+ *
+ * This code only resets information that has to do with long chains.
+ *
+ * TODO: This information should not be stored in the molecule, but should be
+ *       stored in the ClusterLegalizer class instead.
+ *
+ * TODO: This code may be removable. Tried turning it off and found no test
+ *       failures or QoR degredations. Should be investigated in more detail.
  */
-static void revalid_molecules(const t_pb* pb,
-                              bool &mol_validated,
-                              const Prepacker& prepacker) {
-    const t_pb_type* pb_type = pb->pb_graph_node->pb_type;
-
-    if (pb_type->blif_model == nullptr) {
-        int mode = pb->mode;
-        for (int i = 0; i < pb_type->modes[mode].num_pb_type_children && pb->child_pbs != nullptr; i++) {
-            for (int j = 0; j < pb_type->modes[mode].pb_type_children[i].num_pb && pb->child_pbs[i] != nullptr; j++) {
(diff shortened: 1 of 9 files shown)


AlexandreSinger:
@vaughnbetz This has passed CI. Please review and merge when you have a moment. This removed the "valid" flag within the pack molecules which really should be part of the Cluster Legalizer object.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -997,11 +997,70 @@ static void update_molecule_chain_info(t_pack_molecule* chain_molecule, const t_
     VTR_ASSERT(false);
 }
 
+/**
+ * @brief Revalidate the molecules associated with this pb. The mol_validated
```
vaughnbetz, line 1001: If you know anything more about what this routine does (how does it record that molecules are valid? what does it do if mol_validated is true?) it would be good to expand this comment.  If you don't know (you're just moving the code), then we can leave the comment as is.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -997,11 +997,70 @@ static void update_molecule_chain_info(t_pack_molecule* chain_molecule, const t_
...
+                    }
+                }
+            }
+        }
+    }
+}
+
 /*
  * @brief Revert trial atom block iblock and free up memory space accordingly.
  */
```
vaughnbetz, line 1060: Good to expand this comment with as much as you know; you should at least know what mol_validated does so it could be commented.

vpr/src/pack/cluster_legalizer.cpp
AlexandreSinger: I actually knew very little about this routine. In all honesty it made no sense. I added the "mol_validated" flag as a hot fix to keep the behaviour exactly the same. I agree that it was a rough fix. I decided to dig into this method in detail and found that it was completely not necessary anymore (given my recent changes). I greatly simplified the method and turned it into a "reset_molecule_info" method, which does basically the same thing.

vpr/src/pack/cluster_legalizer.cpp
AlexandreSinger: See previous response. I agree that the mol_validated variable was confusing and reworked the function to make it better.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -997,12 +997,33 @@ static void update_molecule_chain_info(t_pack_molecule* chain_molecule, const t_
...
+ *
+ * This code only resets information that has to do with long chains.
+ *
+ * TODO: This information should not be stored in the molecule, but should be
+ *       stored in the ClusterLegalizer class instead.
+ */
+static void reset_molecule_info(t_pack_molecule* mol) {
+    // when invalidating a molecule check if it's a chain molecule
+    // that is part of a long chain. If so, check if this molecule
+    // have modified the chain_id value based on the stale packing
```
vaughnbetz, line 1011: have -> has

vaughnbetz: Want to add a TODO to say this code may be removable?

vaughnbetz (approved):
Looks good; two minor commenting suggestions. Merge whenever you are ready.

PR #2802, opened 2024-11-07
[Placement] Resolved Inconsistency with Placing at Root Tile Locs

### LINKED ISSUE #2801: [Placement] Inconsistency in Grid Blocks Data Structure
For context. I have been creating an independent placement verifier which will ensure that all placement data strcutures are valid before going through the rest of the VPR flow (i.e. routing). The idea is that, if you pass this verification, then your placement can be used for the rest of the VPR flow without issue (regardless of how you created it). This checks invariants that the placement must pass (used as assumptions later in the flow). See PR #2769 

There is a method that is called after placement which synchronizes the grid to the block placement:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/eb8c1862823c4047a188bf76dd57f36e7fcaa3d7/vpr/src/base/vpr_api.cpp#L785-L797

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/eb8c1862823c4047a188bf76dd57f36e7fcaa3d7/vpr/src/util/vpr_utils.cpp#L91-L165

All of the checks found in this method are already done in the independent verifier; however this synchronization is doing something different than what the verifier expects.

The verifier is made to match what the placer generates. The placer will only place blocks at the root tile locations (width_offset = 0; height_offset = 0). For large blocks, the other locations in the tiles are left empty. The verifier ensure that this is what happens. The issue is that the synchronization method above does not follow this. Instead it places the block at all tile locations in the large tile (not just the root). I tried to remove this functionality, but some code (specifically in the flat router) is relying on this for some reason. This causes the `vtr_reg_strong/strong_flat_router` regression test to fail.

I think that it is wasteful and redundant to duplicate the placed block in all tile locations in the large tile. I think that just placing the cluster block in the root tile is sufficient. The `sync_grid_to_blocks` method should be modified to follow this, and the code in the flat router should be fixed to reflect this change (it should always look at the root tile location for the placed block). This method is also only needed when loading the placement, the placement flow should already maintain this.

Eventually it would be a good idea to fix the grid_blocks data structure so that it is impossible to place blocks at non-root tile locations.

AlexandreSinger:
Found an inconsistency with how the placement of cluster blocks into large tiles was handled in VPR. During placement, clusters are only placed at the rool tile location of the grid; however, specific parts of the later stages require that clusters are placed at all tile locations of large tiles. This is inconsistent and makes it confusing to know where clusters should be placed.

Enforced that clusters must be placed at the root tile location and never placed in other tile locations in the large tiles. This required fixing an oversite in the flat router which was relying on the fact that clusters will not be placed at the root tile location.

Resolves #2801


AlexandreSinger:
@amin1377 This was the PR I mentioned at the VTR meeting where I fixed an oversite in the Flat Router.

AlexandreSinger:
@vaughnbetz This has passed CI. Please review. This resolves the issue with the placement of clusters in large tiles that we discussed.

vaughnbetz (approved):

vaughnbetz:
LGTM!

PR #2804, opened 2024-11-08
Fix segmentation fault when moving blocks manually

soheilshahrouz:
This bug was introduced when `BlkLocRegistry` was added to the project. In `manual_move_generator`, the global instance of `BlkLocRegistry` was used instead the one stored in `placer_state_`. This caused a segfault as the global object is initialized at the end of placement.


vaughnbetz:
Thanks @soheilshahrouz !

MohamedElgammal:
Still facing segmentation fault whenever you accept or reject the move. Fixed in PR #2805

PR #2806, opened 2024-11-12
[AP] Global Placer

AlexandreSinger:
Created the Global Placer base class which will create a partial placement based on the netlist and the architecture. This attempts to find a "globally" good placement, without considering all of the complex constraints of the FPGA architecture (without clustering).

Implemented a SimPL-based Global Placer which maintains an upper and lower bound solution which slowly approach each other over several iterations. The lower-bound is the analytically solved solution which tries to optimize the placement (hinted by the upper bound solution). The upper-bound solution is the lower-bound solution which has been partially legalized.


AlexandreSinger:
@vaughnbetz This has passed CI. Please review. This is a much smaller PR than the last one. It adds the Global Placer class into the AP flow which handles how the Analytical Solver interacts with the Partial Legalizer.

vpr/src/analytical_place/global_placer.h
```diff
@@ -0,0 +1,135 @@
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    October 2024
+ * @brief   The declarations of the Global Placer base class which is used to
+ *          define the functionality of all global placers in the AP flow.
+ *
+ * Global Placers create a Partial Placement given onlythe netlist and the
```
vaughnbetz, line 8: Global placers create -> A global placer creates
onlythe -> only the
uses analytical -> uses analytical (i.e. efficient numerical minimization of an objective function of a placement)

vpr/src/analytical_place/global_placer.h
```diff
@@ -0,0 +1,135 @@
...
+ * provides a standard interface for the global placers so they can be used
+ * interchangably. This makes it very easy to test and compare different global
+ * placers.
+ */
+class GlobalPlacer {
+public:
+    virtual ~GlobalPlacer() {}
+
+    /**
+     * @brief Constructor of the base GlobalPlacer class
```
vaughnbetz, line 44: Might want to briefly say what the parameters are. For example:
@param netlist:  netlist of the design at some abstraction level; typically this would have atoms and groups of atoms locked together by carry chains etc. as its blocks.

vpr/src/analytical_place/global_placer.h
```diff
@@ -0,0 +1,135 @@
...
+class SimPLGlobalPlacer : public GlobalPlacer {
+private:
+
+    /// @brief The maximum number of iterations the global placer can perform.
+    static constexpr size_t max_num_iterations_ = 100;
+
+    /// @brief The target gap between the HPWL of the upper-bound and lower-
+    ///        bound placements. The placer will stop if the gap is smaller
+    ///        than this number.
+    static constexpr double target_hpwl_gap_ = 100.0;
```
vaughnbetz, line 110: This should probably be a percentage, as an absolute number probably won't scale well to very big circuits as they show up over time.

vpr/src/analytical_place/global_placer.cpp
```diff
@@ -0,0 +1,133 @@
...
+                                                netlist);
+}
+
+/**
+ * @brief Helper method to print the header of the per-iteration status updates
+ *        of the global placer.
+ */
+static void print_SimPL_status_header() {
+    VTR_LOG("---- ---------------- ---------------- ----------- -------------- ----------\n");
+    VTR_LOG("Iter Lower Bound HPWL Upper Bound HPWL Solver Time Legalizer Time Total Time\n");
```
vaughnbetz, line 51: Might want to space out the header more, as it will be harder to see how to read the spacing within a phrase and across phrases is the same (Iter Lower Bound    HPWL vs. Iter    Lower Bound HPWL). although the dashes below will help.  You can shorten some wording if that makes it too long, e.g.
Iter   L. Bound HPWL ...

vpr/src/analytical_place/global_placer.cpp
```diff
@@ -0,0 +1,133 @@
...
+            float iter_end_time = runtime_timer.elapsed_sec();
+            print_SimPL_status(i, lb_hpwl, ub_hpwl,
+                               solver_end_time - solver_start_time,
+                               legalizer_end_time - legalizer_start_time,
+                               iter_end_time - iter_start_time);
+        }
+
+        // Exit condition: If the upper-bound and lower-bound HPWLs are
+        // sufficiently close together then stop.
+        if (ub_hpwl - lb_hpwl < target_hpwl_gap_)
```
vaughnbetz, line 125: I think a fractional test will work better here.

AlexandreSinger:
@vaughnbetz Thank you for the review. I have resolved your comments. I agree that using an absolute gap was not a sustainable idea. I changed it to a percent gap, as you suggested, where the percent gap is equal to (UB - LB) / UB [since UB should be larger than LB). I set the target percent gap to 10%, which seems to work relatively well for some of the smaller circuits I test on. In the future I will tune this to a better value. Let me know if you have any further comments or if we can merge this.

vaughnb-cerebras:
Great, thanks.

vpr/src/analytical_place/global_placer.h
```diff
@@ -0,0 +1,143 @@
...
+ * Over several iterations the upper-bound and lower-bound solutions will
+ * approach each other until a good quality, mostly-legal solution is found.
+ */
+class SimPLGlobalPlacer : public GlobalPlacer {
+private:
+
+    /// @brief The maximum number of iterations the global placer can perform.
+    static constexpr size_t max_num_iterations_ = 100;
+
+    /// @brief The target percent gap between the HPWL of the upper-bound and
```
vaughnb-cerebras, line 114: we should really call this target_hpwl_relative_gap_  (percent would be 100x greater).
And call it relative gap in the comment, not percent gap.

vaughnb-cerebras (commented):
Looks good, but one last change: we should call it relative gap instead of percent gap.

AlexandreSinger:
@vaughnbetz Completely agree. I like that way better than percent! I have updated the code and documentation. Shall we merge?

vaughnbetz:
Done!

PR #2812, opened 2024-11-16
[CMake] Upgrade Minimum Version to 3.10

AlexandreSinger:
The CI currently always runs on the latest version of CMake. I am not sure why this is, its not the best practice, but CMake is a fairly stable application so it should be fine.

CMake is deprecating version 3.9, and a few CMake files require the minimum version to be 3.9 or larger. Changed these files to be 3.10 or later.

This is currently causing the CI to fail for basically all of the tests with the following errors:
image


AlexandreSinger:
@vaughnbetz The CI is currently failing for all tests due to it tracking the most recent version of CMake and CMake deprecating version 3.9. I have raised our requirements to version 3.10 of CMake (this only affected 3 files, 2 of which were EZGL). This fixed the issues. Please review and merge when you get a moment.

vaughnbetz:
Thanks @AlexandreSinger !

PR #2813, opened 2024-11-18
Placer class

soheilshahrouz:
Add `Placer` class that encapsulates all objects and variable that are placement instance specific. Shared objects like delay model are still constructed outside this class and shared across all `Placer` instances.

Don't look at the code yet; it's a mess! :)


vpr/src/place/place.cpp
```diff
@@ -192,454 +89,26 @@ void try_place(const Netlist<>& net_list,
...
-        blk_loc_registry.place_sync_external_block_connections(block_id);
-    }
-
     const int width_fac = placer_opts.place_chan_width;
-    init_draw_coords((float)width_fac, blk_loc_registry);
-
-    /* Allocated here because it goes into timing critical code where each memory allocation is expensive */
-    IntraLbPbPinLookup pb_gpin_lookup(device_ctx.logical_block_types);
-    //Enables fast look-up of atom pins connect to CLB pins
-    ClusteredPinAtomPinsLookup netlist_pin_lookup(cluster_ctx.clb_nlist, atom_ctx.nlist, pb_gpin_lookup);
```
soheilshahrouz, line 255: @vaughnbetz 
`pb_gpin_lookup` and `netlist_pin_lookup` are initialized after the placement timer starts in line 215. Is their initialization times intentionally included in the total placement time?

vpr/src/place/place_log_util.cpp
```diff
@@ -0,0 +1,324 @@
...
+            blk_loc_registry.place_macros().macros().size(), num_macro_members,
+            float(num_macro_members) / blk_loc_registry.place_macros().macros().size());
+    VTR_LOG("\n");
+
+    sprintf(msg_.data(),
+            "Initial Placement.  Cost: %g  BB Cost: %g  TD Cost %g \t Channel Factor: %d",
+            costs.cost, costs.bb_cost, costs.timing_cost, placer_opts.place_chan_width);
+
+    // Draw the initial placement
+    update_screen(ScreenUpdatePriority::MAJOR, msg_.data(), PLACEMENT, timing_info);
```
AlexandreSinger, line 216: Why is drawing code in the logging utility? Was this intentional? Or is this just printing the initial placement stats to the screen? If it is the latter, I recommend updating the comment above this line.

vpr/src/place/place_log_util.h
```diff
@@ -0,0 +1,44 @@
+
```
AlexandreSinger, line 1: Same comment as the definition file. A doxygen comment for the file would be useful!

vpr/src/place/placer.h
```diff
@@ -0,0 +1,129 @@
+
```
AlexandreSinger, line 1: A doxygen comment for the file here would be nice!

vpr/src/place/placer.h
```diff
@@ -0,0 +1,129 @@
...
+
+#ifndef VTR_PLACER_H
+#define VTR_PLACER_H
```
AlexandreSinger, line 4: Why not use `#pragma once`, we are using C++ we might as well. I worry VTR_PLACER_H may not be very unique. What we do in LLVM would be `VPR_SRC_PLACE_PLACER_H`, which is the file path which necessarily must be unique!

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,435 @@
...
+   } else {
+       VPR_ERROR(VPR_ERROR_PLACE,
+                 "\nCompleted placement consistency check, %d errors found.\n"
+                 "Aborting program.\n",
+                 error);
+   }
+}
+
+int Placer::check_placement_costs_() {
+   int error = 0;
```
AlexandreSinger, line 246: Small nitpick. I would make this an unsigned! I doubt we can ever have negative errors.

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,435 @@
...
+    for (const ClusterBlockId block_id : cluster_ctx.clb_nlist.blocks()) {
+       placer_state_.mutable_blk_loc_registry().place_sync_external_block_connections(block_id);
+    }
+
+    check_place_();
+
+    log_printer_.print_post_placement_stats();
+}
+
+void Placer::copy_locs_to_global_state() {
```
AlexandreSinger, line 389: This is up to you, but I recommend having the user pass in the mutable `place_ctx` object. This makes it clear that we are updating global state (even though it is obvious from the name). Its just nice since when people call this class, they know at what point the global state gets modified (so the rest of the code they can assume it is constant).

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,435 @@
...
+    auto& global_blk_loc_registry = place_ctx.mutable_blk_loc_registry();
+    global_blk_loc_registry = placer_state_.blk_loc_registry();
+
+#ifndef NO_GRAPHICS
+    // update the graphics' reference to placement location variables
+    get_draw_state_vars()->set_graphics_blk_loc_registry_ref(global_blk_loc_registry);
+#endif
+}
+
+const PlacementAnnealer& Placer::annealer() const {
```
AlexandreSinger, line 405: For simple getter methods like these, I recommend just declaring them inline in the header file. Or even consider making them public variables (since most of these are const anyways we do not have to worry about people screwing with variables they are not supposed to).

AlexandreSinger (commented):
Hi @soheilshahrouz these changes look great to me! I left some minor comments. I am excited to see this PR merged!

vpr/src/place/place.cpp
vaughnbetz: Probably just what was convenient. They can be moved outside placement if it cleans up the code, but please measure how long they take and if it can be significant they should get their own timers (either on their own or together).

vpr/src/place/place_log_util.cpp
vaughnbetz: I'm OK with leaving it here (@soheilshahrouz tells me it does keep the code shorter), but I agree the class should have a comment saying it is controlling drawing and logging.

vpr/src/place/placer.h
```diff
@@ -0,0 +1,123 @@
+/**
+ * @file placer.h
+ * @brief Declares the Placer class, which encapsulates the functionality, data structures,
+ * and algorithms required for the placement stage.
```
vaughnbetz, line 4: Could be more clear to say "for the (annealing-based) placement stage."

vpr/src/place/placer.h
```diff
@@ -0,0 +1,123 @@
...
+           const t_placer_opts& placer_opts,
+           const t_analysis_opts& analysis_opts,
+           const t_noc_opts& noc_opts,
+           const std::vector<t_direct_inf>& directs,
+           std::shared_ptr<PlaceDelayModel> place_delay_model,
+           bool cube_bb,
+           bool is_flat,
+           bool quiet);
+
+    void place();
```
vaughnbetz, line 48: Comment what place () does.

vpr/src/place/placer.h
```diff
@@ -0,0 +1,123 @@
...
+    std::shared_ptr<PlaceDelayModel> place_delay_model_;
+    /// Prints logs during placement
+    const PlacementLogPrinter log_printer_;
+    /// Indicates if flat routing resource graph and delay model is used. It should be false.
+    const bool is_flat_;
+
+    /// Stores a placement state as a retrievable checkpoint in case the placement quality deteriorates later.
+    t_placement_checkpoint placement_checkpoint_;
+
+    std::shared_ptr<SetupTimingInfo> timing_info_;
```
vaughnbetz, line 83: Doxygen comments. Should explain the shared ones vs. unique ones too.

vpr/src/place/placer.h
```diff
@@ -0,0 +1,123 @@
...
+    std::shared_ptr<PlacementDelayCalculator> placement_delay_calc_;
+    std::unique_ptr<PlacerSetupSlacks> placer_setup_slacks_;
+    std::unique_ptr<PlacerCriticalities> placer_criticalities_;
+    std::unique_ptr<NetPinTimingInvalidator> pin_timing_invalidator_;
+    tatum::TimingPathInfo critical_path_;
+
+    std::unique_ptr<vtr::ScopedStartFinishTimer> timer_;
+
+    IntraLbPbPinLookup pb_gpin_lookup_;
```
vaughnbetz, line 92: Comment all data members

vpr/src/place/placer.h
```diff
@@ -0,0 +1,123 @@
...
+    ClusteredPinAtomPinsLookup netlist_pin_lookup_;
+
+    std::unique_ptr<PlacementAnnealer> annealer_;
+
+    t_timing_analysis_profile_info pre_place_timing_stats_;
```
vaughnbetz, line 97: Comment, along with why we store these.

vpr/src/place/placer.h
```diff
@@ -0,0 +1,123 @@
...
+    t_timing_analysis_profile_info pre_quench_timing_stats_;
+    t_timing_analysis_profile_info post_quench_timing_stats_;
+
+    friend class PlacementLogPrinter;
+
+  private:
+    void alloc_and_init_timing_objects_(const Netlist<>& net_list,
```
vaughnbetz, line 104: Doxygen comment.

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+    init_placement_context(placer_state_.mutable_blk_loc_registry(), directs);
+
+    // create a NoC cost handler if NoC optimization is enabled
+    if (noc_opts.noc) {
+        noc_cost_handler_.emplace(placer_state_.block_locs());
+    }
+
+    // Start measuring placement time
+    timer_ = std::make_unique<vtr::ScopedStartFinishTimer>("Placement");
```
vaughnbetz, line 53: Shouldn't we move this to the start of the constructor?

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,403 @@
...
+    //create the move generator based on the chosen placement strategy
+    auto [move_generator, move_generator2] = create_move_generators(placer_state_, placer_opts, move_lim, noc_opts.noc_centroid_weight, rng_);
+
+    if (!placer_opts.write_initial_place_file.empty()) {
+        print_place(nullptr, nullptr, placer_opts.write_initial_place_file.c_str(), placer_state_.block_locs());
+    }
+
+#ifdef ENABLE_ANALYTIC_PLACE
+    /*
+     * Analytic Placer:
```
vaughnbetz, line 82: Cluster-level Analytic Placer:

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,403 @@
...
+     * the relative importance of NoC cost terms with respect to each other, while
+     * the importance of total NoC cost to conventional placement cost is determined
+     * by NoC placement weighting factor.
+     */
+    if (noc_opts.noc) {
+        normalize_noc_cost_weighting_factor(const_cast<t_noc_opts&>(noc_opts));
+    }
+
+    BlkLocRegistry& blk_loc_registry = placer_state_.mutable_blk_loc_registry();
+    initial_placement(placer_opts, placer_opts.constraints_file.c_str(),
```
vaughnbetz, line 69: Could pass a block location registry to a separate update_locations (or some such) function instead if you wished. But I'm OK with this if it serves the needs of everybody.

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,403 @@
...
+}
+
+void Placer::place() {
+   const auto& timing_ctx = g_vpr_ctx.timing();
+   const auto& cluster_ctx = g_vpr_ctx.clustering();
+
+
+   bool skip_anneal = false;
+#ifdef ENABLE_ANALYTIC_PLACE
+   // When enabled, skip most of the annealing and go straight to quench
```
vaughnbetz, line 280: Cluster-level analytic placer: when enabled, skip most of ...

vpr/src/place/placer.cpp
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    if (placer_opts_.placement_saves_per_temperature >= 1) {
+       std::string filename = vtr::string_fmt("placement_%03d_%03d.place",
+                                              annealing_state.num_temps + 1, 0);
+       VTR_LOG("Saving final placement to file: %s\n", filename.c_str());
+       print_place(nullptr, nullptr, filename.c_str(), placer_state_.mutable_block_locs());
+    }
+
+    //#ifdef VERBOSE
```
vaughnbetz, line 375: Can delete.

vpr/src/place/timing_place_lookup.h
```diff
@@ -2,7 +2,7 @@
 #define TIMING_PLACE_LOOKUP_H
 #include "place_delay_model.h"
 
-std::unique_ptr<PlaceDelayModel> compute_place_delay_model(const t_placer_opts& placer_opts,
+std::shared_ptr<PlaceDelayModel> compute_place_delay_model(const t_placer_opts& placer_opts,
```
vaughnbetz, line 5: Comment this function and use of shared_ptr

vaughnbetz (commented):
Looks good; some suggested commenting and a few minor tweaks.

soheilshahrouz:
@vaughnbetz 
This is ready to merge into master

PR #2814, opened 2024-11-18
[Sys] Increased Read Buffers to Reduce Total Syscalls

AlexandreSinger:
This change was originally proposed by @heshpdx

See PR #2772

By increasing the read buffer, the number of syscalls can be reduced since more of the large files are read per call.


libs/libpugiutil/src/pugixml_loc.cpp
```diff
@@ -1,8 +1,14 @@
 #include <cstdio>
 #include <algorithm>
+#include <vector>
 #include "pugixml_util.hpp"
 #include "pugixml_loc.hpp"
 
+// The size of the read buffer when reading from a file.
+#ifndef PUGI_UTIL_READ_BUF_SIZE
+#define PUGI_UTIL_READ_BUF_SIZE 1024
```
vaughnbetz, line 9: Why not make this big too?  Nobody needs small buffers that I can think of ...

libs/libvtrutil/src/vtr_digest.cpp
```diff
@@ -3,10 +3,15 @@
...
 #include <iostream>
 #include <fstream>
-#include <array>
+#include <vector>
 
 #include "picosha2.h"
 
+// The size of the read buffer when reading from a file.
+#ifndef VTR_UTIL_READ_BUF_SIZE
+#define VTR_UTIL_READ_BUF_SIZE 1024
```
vaughnbetz, line 12: Why not make this big in the #define?

vaughnbetz (commented):
Couldn't we just #define big buffer sizes? That seems less complex and equally useful.

vaughnbetz:
LGTM.  Will merge as soon as CI passes.

heshpdx:
1 MB will be good for the next 20 years. 😄

vaughnbetz:
And in 21 years, it will be Alex's problem :).

PR #2815, opened 2024-11-19
7-Series Pin-Equivalent Ports

amin1377:
#### Description  
This PR updates the 7-series architecture to mark the pins connected to LUTs as pin-equivalent.


vaughnbetz:
We shouldn't merge this in due to the discussion about these pins not really being equivalent when used in fracturable LUT mode, and not being equivalent when we consider the internal routing etc.

PR #2816, opened 2024-11-20
[Packer] Created GreedyClusterer Class

AlexandreSinger:
Began encapsulating the different parts of the packer into classes. This will help organize the packer better, which will make it easier to modify in the future.

My plan is to clean up the different parts of the packer so I can add flat placement information into the gain calculation so it can be integrated into the AP flow.


AlexandreSinger:
@vaughnbetz After looking into the greedy clusterer more, I realized that creating my own AP clusterer would cause me to reinvent the wheel a lot. It will be much easier (and probably more useful for the VTR ecosystem) if I clean up the Packer to make it easier to add the ability to optionally pass a flat placement into the clusterer. This PR starts the process of cleaning things up by creating a GreedyClusterer class. I plan to move some of the functions from cluster_utils.cpp into this class and continue to clean up dead / confusing code. I wanted to merge this in pieces to make it easier to review. Let me know what you think!

vaughnbetz:
That seems reasonable to me. I think the current clusterer has 3 limitations / trrade-offs:
- doesn't have any physical info. You are solving that by creating a flat placement that you can use to guide it.
- Creates one cluster at a time (greedy).  This is a separate problem, but probably requires some significant refactoring (or a new clusterer that shares some code) to overcome by making some kind of LSC/simultaneous clustering approach
- Full legality checking is slow. You may need an even faster heuristic legality checker. This is mostly a separate problem though that ideally wouldn't have to be tackled in your work.

vpr/src/pack/cluster_util.cpp
```diff
@@ -1879,6 +1879,35 @@ void print_pb_type_count_recurr(t_pb_type* pb_type, size_t max_name_chars, size_
     }
 }
 
+/**
```
vaughnbetz, line 1882: Put this comment in the header file.

vaughnbetz (commented):
Looks good; some commenting suggestions.

PR #2819, opened 2024-11-21
Router Lookahead Sampling 

amin1377:
#### Description
The number of samples taken for each router lookahead is reduced from 8 to 1 to reduce the time to build router lookahead.


vaughnbetz:
Looks good.  You'll need to either update the golden results on a few QoR failures (I checked and they all look fine -- small circuits and more got better than got worse), or we'll need to combine this with Mohamed's cluster retune PR so you piggyback on his golden results. @MohamedElgammal

vaughnbetz:
Updating branch so it has Mohamed's cluster parameter change.  Then @amin1377 will update the various nightly golden results, as we're likely to have some re-sync needed there.

PR #2820, opened 2024-11-21
Change the default output pin utilization of logic blocks to 0.6

MohamedElgammal:
#### Description

As shown by the attached results run by @amin1377, light packing is better for both runtime and QoR. Hence, changing the default value to 0.6

#### Checklist:


vaughnbetz:
Thanks Mohamed.  We'll also have to update the various nightly test golden results.  If this is time critical I can merge in advance of that, but we'll need those updates very soon or no one will be able to run the nightly test, so even if we merge without them we'd need to get an update of them ASAP.

MohamedElgammal:
I just talked to Amin and I think we can land this right away. Then, he will rebase it in his branch and update the golden results for nightly tests so that we don't run it twice.

vaughnbetz:
OK, done!

PR #2822, opened 2024-11-22
[Packer] Created GreedySeedSelector Class

AlexandreSinger:
Cleaned up the way that new seed molecules were picked in the greedy
clusterer by abstracting it into a class. This will make it simpler to
build upon this or change how seeds are selected in the future.


vpr/src/pack/greedy_seed_selector.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include "echo_files.h"
+#include "prepack.h"
+#include "vpr_error.h"
+#include "vpr_types.h"
+#include "vtr_assert.h"
+#include "vtr_math.h"
+#include "vtr_vector.h"
+
+/**
+ * @brief Helper method that computes the seed gain of the give atom block.
```
vaughnbetz, line 23: give -> given
A higher gain means it is a more desirable seed (starting point) for a new cluster.

vpr/src/pack/greedy_seed_selector.cpp
```diff
@@ -0,0 +1,237 @@
...
+ * gain.
+ */
+static inline float get_seed_gain(AtomBlockId blk_id,
+                                  const AtomNetlist& atom_netlist,
+                                  const Prepacker& prepacker,
+                                  const e_cluster_seed seed_type,
+                                  const t_molecule_stats& max_molecule_stats,
+                                  const vtr::vector<AtomBlockId, float>& atom_criticality) {
+    switch (seed_type) {
+        // By criticality
```
vaughnbetz, line 35: . Intuition: starting a cluster with primitives that have timing-critical connections may help timing.

vpr/src/pack/greedy_seed_selector.cpp
```diff
@@ -0,0 +1,237 @@
...
+        case e_cluster_seed::TIMING:
+            return atom_criticality[blk_id];
+        // By number of used molecule input pins
```
vaughnbetz, line 38: Intuition: molecules that use more inputs can be difficult to legally pack into partially full clusters. Use them as seeds instead.

vpr/src/pack/greedy_seed_selector.cpp
```diff
@@ -0,0 +1,237 @@
...
+            const t_molecule_stats molecule_stats = calc_molecule_stats(blk_mol, atom_netlist);
+            VTR_ASSERT(max_molecule_stats.num_used_ext_inputs > 0);
+
+            float used_ext_input_pin_ratio = vtr::safe_ratio<float>(molecule_stats.num_used_ext_inputs, max_molecule_stats.num_used_ext_inputs);
+            float blend_gain = (seed_blend_fac * atom_criticality[blk_id]
+                                + (1 - seed_blend_fac) * used_ext_input_pin_ratio);
+            blend_gain *= (1 + 0.2 * (molecule_stats.num_blocks - 1));
+            return blend_gain;
+        }
+        // By pins per molecule (i.e. available pins on primitives, not pins in use)
```
vaughnbetz, line 63: Intuition (a weak one): primitive types with more pins or inputs might be harder to pack.

vaughnbetz (commented):
LGTM!

PR #2830, opened 2024-11-30
[Warnings] Resolved Basic Warnings for GCC13

AlexandreSinger:
Resolved a warning in search_bar.cpp which warned about a reference to a temporary. Simply fixed by splitting the statement into two lines.

Resolved a warning in draw_rr.cpp where a print statement was trying to access an invalid location in an array. The print statement did not make sense, it was trying to print an undefined side. This was removed.

Resolved a warning in the automatically generated vpr_constraints file. This was an interesting warning due to the decltype method returning the attributes of the FILE pointer, but this was passed into a template which discards the attributes (this generates a warning). Resolved this manually locally and raised a PR on the uxsdcxx repo with this fix: https://github.com/SymbiFlow/uxsdcxx/pull/53


AlexandreSinger:
@vaughnbetz This is a very small PR which resolves most of the last warnings in GCC13. After issue #2831 is resolved, we should be able to raise the CI to 24.04 while remaining warning clean.

vaughnbetz (approved):
Thanks!

PR #2832, opened 2024-11-30
[Warnings] Fixed Minor Warnings

### LINKED ISSUE #2831: [SPECRand] Uninitialized Induction Variable in spec_rand.cpp
In the GCC13 build of VTR, we are getting the following warning from `spec_rand.cpp`:
image

This code can be found in VTR here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/ec23fea1f56d52a0dcd94787f3ee81517636671c/libs/libvtrutil/src/specrand.cpp#L122-L125

Leaving the induction variable uninitialized is undefined behavior, I believe; however, the compiler is likely initializing it to zero for us. The fix for this is to simply update it to `size_t kk = 0` in the preamble of the for loop; however, since this file is from outside of the VTR repo I am not sure if we can change this file.

AlexandreSinger:
Fixed a warning with a loop induction variable not being initialized.

Fixed a warning where a method overrides the base class, but was not explicitly declared as override.

Resolves #2831

### DIFF
--- a/libs/libvtrutil/src/specrand.cpp
+++ b/libs/libvtrutil/src/specrand.cpp
@@ -119,7 +119,7 @@ unsigned long SpecRandomNumberGenerator::spec_genrand_int32_() {
             y = (mt[kk] & UPPER_MASK) | (mt[kk + 1] & LOWER_MASK);
             mt[kk] = mt[kk + M] ^ (y >> 1) ^ mag01[y & 0x1UL];
         }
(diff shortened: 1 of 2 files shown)


vaughnbetz:
@heshpdx @amin1377 : worth checking if this uninitialized variable also exists in the spec version of the code.

`for (size_t kk; kk < N - 1; kk++) {`

heshpdx:
Yup, got it, thank you!

PR #2838, opened 2024-12-04
Update koios tasks and add a koios_task_list that runs all koios benchmarks

MohamedElgammal:
With this update PR, running the `koios_task_list.txt` will run 37 Koios benchmarks


vtr_flow/tasks/regression_tests/vtr_reg_weekly/koios_bwave_large/config/config.txt
```diff
@@ -35,4 +36,4 @@ qor_parse_file=qor_standard.txt
 pass_requirements_file=pass_requirements.txt
 
 #Script parameters
-script_params=-track_memory_usage -crit_path_router_iterations 100 --route_chan_width 300 --target_utilization 0.28
\ No newline at end of file
+script_params=-track_memory_usage -crit_path_router_iterations 100 --route_chan_width 400 --target_utilization 0.12
```
vaughnbetz, line 39: What's the reason for the .12 target_utilization?  We should at least comment it (and it is quite a bit lower than the old .28).

vaughnbetz (commented):
Thanks @MohamedElgammal .  Does the documentation need and update with this?  Currently I believe it has instructions on running various Koios sub-suites -- can we now just say they should run and parse this one (and give the commands for that)?

vaughnbetz:
See https://docs.verilogtorouting.org/en/latest/README.developers/#example-koios-benchmarks-qor-measurement for the current Koios run documentation.

vaughnbetz:
Nag, nag ...

MohamedElgammal:
@vaughnbetz I think this PR is no longer needed. I have opened a new PR that includes all the changes from v9.0 branch (PR #2866 )

vaughnbetz:
@MohamedElgammal : What runs do we recommend people run / extract when checking QoR on the Koios designs?  The first four configs (the ones with hard blocks)?  I think we should mention that somewhere in the documentation.

PR #2843, opened 2024-12-13
IPIN Cost Multiplier

amin1377:
The IPIN cost multiplier was set to a value below one to prioritize pulling IPINs faster during wave expansion from the heap, enabling quicker access to the SINK if the IPIN is connected to the target SINK. Previously, this multiplier was set to 0.95, which is not perfectly representable in the IEEE 754 floating-point format. To ensure consistent results across all platforms, a condition was added to use 0.875 (1/2 + 1/4 + 1/8) when the SPEC flag was enabled. However, since 0.95 and 0.875 are very close in value, we have now removed the conditional logic and only used 0.875.


vaughnbetz:
Thanks @amin1377 .  I don't expect this will lead to any meaningful change, but could you run a QoR check on Titan to be sure?

amin1377:
Titan quick qor:
image

amin1377:
@vaughnbetz: The QoR remains nearly the same, but placement time has increased by 6%, likely contributing to a 3% rise in overall runtime. I ran the benchmark on SAVI to minimize machine load fluctuations, but it's possible the load changed during the test with the new code. I don’t think the changes made in this PR would significantly impact runtime.

vaughnbetz:
Thanks. Yes it's hard to imagine how this could change place time

Issue #2840, opened 2024-12-05
Placement delay lookup can not be readback 


tangxifan:
The placement delay lookup is a useful binary file which saves computing timing for a fixed device.

#### Expected Behaviour
Similar to the routing resource graph (RRGraph), its usage is expected to be simple.
Users can enable the option, e.g.,  ``--write_placement_delay_lookup pdlylkup.bin``, to output the data into a binary file.
When in later VPR runs, users can enable the option, e.g., ``--read_placement_delay_lookup pdlylkup.bin``, to readback the data from the binary file.

#### Current Behaviour
The binary file written out cannot be read back. A segmentation fault is shown. 

#### Possible Solution
Should fix this bug if verified. Readback is the simplest case. 
Expect to have a dedicated testcase in strong regression tests, to avoid the bug to occur later

#### Steps to Reproduce

1. Download the latest VPR
2. Compile with ``make all -j32``
3. ``python3 run_quick_test.py``
4. ``cd vtr_flow/tasks/regression_tests/vtr_reg_basic/basic_timing/run001/k6_N10_mem32K_40nm.xml/ch_intrinsics.v/common``
5. ``<YOUR_VTR_ROOT>/vtr-verilog-to-routing/vpr/vpr k6_N10_mem32K_40nm.xml ch_intrinsics --circuit_file ch_intrinsics.pre-vpr.blif --route_chan_width 60 --write_placement_delay_lookup pdlylkup.bin``
6. ``<YOUR_VTR_ROOT>/vtr-verilog-to-routing/vpr/vpr k6_N10_mem32K_40nm.xml ch_intrinsics --circuit_file ch_intrinsics.pre-vpr.blif --route_chan_width 60 --read_placement_delay_lookup pdlylkup.bin``
7. You should see the same error as I do

![image](https://github.com/user-attachments/assets/7693ba07-a1a1-49d1-a85a-14c604cbcd30)

#### Your Environment

* VTR revision used: [ce706d5](https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/ce706d579de8d370994fb3d3eb1a469d98707866)

* Operating System and version: Ubuntu 22.04
* Compiler version: GCC-12

tangxifan:
@vaughnbetz I believe that this is a critical bug, as a show stopper to anyone who is following latest VTR.

vaughnbetz:
[...]

PR #2844, opened 2024-12-13
Simple Place Delay Model Read/Write 

amin1377:
#### Description
This PR implements the read/write functionality for the simple placement delay model. The feature was tested on the Titan `murax_stratixiv_arch_timing` circuit, and the resulting placement was consistent with the placement delay model read from the previously written file.

#### Related Issue
This PR addresses Issue  #2840


vpr/src/place/place_delay_model.cpp
```diff
@@ -198,6 +206,64 @@ static void FromFloat(VprFloatEntry::Builder* out, const float& in) {
...
+    //
+    // Note that capnproto does not validate that the incoming data matches the
+    // schema.  If this property is required, some form of check would be
+    // required.
+    model = reader.getRoot<VprDeltaDelayModel>();
+
+    // ToNdMatrix is a generic function for converting a Matrix capnproto
+    // to a vtr::NdMatrix.
+    //
+    // The use must supply the matrix dimension (2 in this case), the source
```
vaughnbetz, line 238: Use --> user
2 dimensions in comment --> should be 5

vpr/src/place/place_delay_model.cpp
```diff
@@ -198,6 +206,64 @@ static void FromFloat(VprFloatEntry::Builder* out, const float& in) {
...
+    // target C++ type (flat), and a function to convert from the source capnproto
+    // type to the target C++ type (ToFloat).
+    //
+    // The second argument should be of type Matrix<X>::Reader where X is the
+    // capnproto element type.
+    ToNdMatrix<5, VprFloatEntry, float>(&delays_, model.getDelays(), ToFloat);
+}
+
+void SimpleDelayModel::write(const std::string& file) const {
+    // MallocMessageBuilder object is the generate capnproto message builder,
```
vaughnbetz, line 249: Generate --> generic? Generated?

vpr/src/place/place_delay_model.h
```diff
@@ -243,8 +243,11 @@ class SimpleDelayModel : public PlaceDelayModel {
     float delay(const t_physical_tile_loc& from_loc, int /*from_pin*/, const t_physical_tile_loc& to_loc, int /*to_pin*/) const override;
     void dump_echo(std::string /*filepath*/) const override {}
 
-    void read(const std::string& /*file*/) override {}
-    void write(const std::string& /*file*/) const override {}
+    void read(const std::string& /*file*/) override;
+    void write(const std::string& /*file*/) const override;
+    const vtr::NdMatrix<float, 5>& delays() const {
```
vaughnbetz, line 248: Any reasonable n why this is always a template parameter of 5 rather than a passed one n template parameter?

vaughnbetz: Doxygen comment would be good.

vpr/src/place/place_delay_model.cpp
```diff
@@ -170,6 +170,14 @@ float SimpleDelayModel::delay(const t_physical_tile_loc& from_loc, int /*from_pi
         "is disable because VTR_ENABLE_CAPNPROTO=OFF." \
         "Re-compile with CMake option VTR_ENABLE_CAPNPROTO=ON to enable."
 
+void SimpleDelayModel::read(const std::string& /*file*/) {
```
vaughnbetz, line 173: Why are these defined and then redefined?

vaughnbetz (changes requested):

vpr/src/place/place_delay_model.cpp
amin1377: The definitions are guarded by a macro that checks whether Cap’n Proto is enabled. If it is not enabled, an error is thrown if these functions are called. Otherwise, Cap’n Proto functions are used to read and write the array.

vaughnbetz: OK, thanks.

vaughnbetz (approved):
Thanks for the updates.

PR #2848, opened 2024-12-21
[CI] Upgraded Runners to Ubuntu 24.04

AlexandreSinger:
Upgraded the GitHub-hosted runners from Ubuntu 22.04 to Ubuntu 24.04. This ensures that all changes to VTR will work on the most recent version of Ubuntu (as well as being warning clean).

Removed the prior tests for Ubuntu 24.04.

Added a tests to ensure Ubuntu 22.04 remains warning clean as well.

Change Summary:
- Changed the default GCC version (the one most tests are built with) from GCC-11 to GCC-13 (the default on Ubuntu 24.04).
- Changed the runners for all tests from `ubuntu-22.04` to `ubuntu-24.04`.
- Changed the python version for all tests from `3.10.10` to `3.12.3` (the default on Ubuntu 24.04). However, the Python Lint testcase needed to remain on `3.10.10` due to new linting errors that showed up on `3.12.3`; this should be resolved in a different PR.
- Turned off IPO on the `Regression` tests, since they are used to check for warning cleanliness (IPO can cause some warnings) and do  not need to be performant; so they can compile faster without IPO.
- Updated the `Compatibility` tests:
  - Removed GCC-9 and GCC-10
  - Removed Clang-11, Clang-12, Clang-13, and Clang-14
  - Added GCC-11 and GCC-12
  - Added Clang-15, Clang-16, Clang-17, Clang-18
- Added a compatibility test for Ubuntu 22.04 (with GCC-11) to ensure that it is also warning clean.

The only code that needed to be modified was around some of the Eigen include statements, since Eigen has some small warnings in GCC-13.


AlexandreSinger:
@vaughnbetz This has passed CI. The required tests which say "expected" above are the old tests that I turned off in the CI. Since they were labelled as required, GitHub is expecting them to eventually run. What do you think of this PR? Also, are you ok with me making the "expected" tests above no longer required?

vaughnbetz:
Looks great, thanks!  Yes, please remove the tests we're no longer running from the required list.

AlexandreSinger:
The tests have been removed from the merge rules. We can verify that these changes took affect the next time a PR tries to run the tests.

PR #2854, opened 2025-01-05
Get VTR working on macOS again

Liamolucko:
VTR's support for macOS has bitrotted over time; this gets it working again.

#### Description
The only problems that were actually related to macOS were the missing includes and Parmys using `.dylib` instead of `.so`; the rest were just a result of the tests being run in a different order and can be reproduced on Linux with `test_vpr --order rand`.

`vtr_reg_strong` still has two failing tests: `fix_clusters_test_arch/apex2` and `k6_frac_N10_frac_chain_mem32K_40nm/or1200`. `or1200` fails on Linux too for me, so I didn't bother investigating further (it's saying that the circuit is unroutable).

After some cursory investigation, it seems like `apex2` might be failing because two blocks which the `.place` file specifies should be in different locations are getting packed into the same cluster before the `.place` file is read. I don't know enough about how VPR works to figure out what's preventing that from happening on Linux as well (or anything about it at all really, and I could have completely misunderstood the issue).

I ran the `ctest` unit tests, `vtr_reg_basic`, and `vtr_reg_strong` (which had the failures mentioned above. On Linux, `k6_frac_N10_40nm/ex1010.pre-vpr` also failed with an error saying it was unroutable, but only when using GCC; it passed when using Clang).

`ctest` and `vtr_reg_basic` were run via. [this Nix package](https://github.com/Liamolucko/nix-config/blob/9b9af3b1b43928d492d12bee916a748382265234/pkgs/vtr/default.nix), and `vtr_reg_strong` was run in a `nix develop` shell for the same package.

#### Types of changes

#### Checklist:


hzeller:
(off topic: I think it would be good if you contributed the nix package to https://github.com/NixOS/nixpkgs  ... just this morning I was looking if `vtr` needs to be verion-updated in nixpkgs ... and realized that it was actually not even there. So having it in the regular nixpkgs would actually be very nice)

vaughnbetz:
Thanks!  The changes look OK; launching CI.

vaughnbetz:
CI passed; I'm guessing the apex2 failure is due to some different round off making us create a different clustering which doesn't match the .place file (indicating a flaw in the test really).  The routability failure was probably a random one due to imperfections in the algorithms and different round off leading to a different decision.

PR #2855, opened 2025-01-07
convert switch id check with architecture to warning instead of error

saaramahmoudi:
This PR converts the switch error check with the architecture to a warning. 

@vaughnbetz

### DIFF
--- a/libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
+++ b/libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
@@ -534,7 +534,7 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
             }
         }
         if (!found_arch_name) {
-            VTR_LOG("Switch name '%s' found in RR graph input from file but not in the architecture file creating it.\n", string_name.c_str());
+            VTR_LOG("Switch name '%s' found in RR graph input from file but not in the architecture file; creating it.\n", string_name.c_str());
         }
         sw->intra_tile = is_internal_sw;
         sw->name = string_name;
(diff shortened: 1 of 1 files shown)


libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -534,7 +534,7 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
             }
         }
         if (!found_arch_name) {
-            report_error("Switch name '%s' not found in architecture\n", string_name.c_str());
+            VTR_LOG_WARN("Switch name '%s' not found in architecture\n", string_name.c_str());
```
vaughnbetz, line 537: Probably we should not make this a warning, and instead just make it an info message saying something like:
"Switch name '%s' found in RR graph input from file but not in architecture file; creating it.\n", string_name.c_str());

Otherwise you are likely to get many warnings with the RR graphs you are bringing in, and they will be spurious.

vaughnbetz (changes requested):
Can you summarize the testing for this case? The nightly tests aren't running automatically right now, so if you're not sure if this case is tested in the github runners you should run them manually.

Do we need any new tests?

saaramahmoudi:
@vaughnbetz I don't think there is a need to add new tests. However, I have ran the titan benchmarks locally and made sure that nothing is going to break with this change. 
We also treat internal switches for flat router with the same logic, we don't look for them in the architecture file and we just don't print any error with those. So, I think it should be working fine

vaughnbetz (commented):
Thanks. One nit: there should be a semicolon (or a period and new sentence if you prefer) before "creating it." i.e. "; creating it."

PR #2861, opened 2025-01-16
[APPack] Added read_flat_place Option and Verifier

AlexandreSinger:
Added the read_flat_place option to VPR's CLI. This allows the user to pass a flat placement file into the packer.

Added code to parse this file into a FlatPlacementInfo object which only verifies that the atom given in the file exists in the atom netlist.

Added code in the packer to read in the flat placement and verify that the flat placement is valid for use in the packer.

### DIFF
--- a/doc/src/vpr/command_line_usage.rst
+++ b/doc/src/vpr/command_line_usage.rst
@@ -408,6 +408,31 @@ Use the options below to override this default naming behaviour.
 
     Prefix for output files
 
+.. option:: --read_flat_place <file>
+
+    Reads a file containing the locations of each atom on the FPGA.
+
+    This is used by the packer to better cluster atoms together.
+
+    The flat placement file (which often ends in ``.fplace``) is a text file
+    where each line describes the location of an atom. It has the following
+    expected syntax:
+
+    .. code-block:: none
+
+        <atom_name : str> <atom_x_pos : float> <atom_y_pos : float> <atom_sub_tile? : int> <atom_site_idx? : int>
+        n523  6 8 0 3
+        n522  6 8 0 5
+        n520  6 8 0 2
+        n518  6 8 0 16
+
+    The ``sub_tile`` and ``site_idx`` are optional parameters which may be used
+    as a hint to reconstruct clusters.
+
+    .. warning::
+
+        This interface is currently experimental and under active development.
+
 .. option:: --write_flat_place <file>
 
     Writes the post-placement locations of each atom into a flat placement file.
--- a/doc/src/vpr/command_line_usage.rst
+++ b/doc/src/vpr/command_line_usage.rst
@@ -411,23 +411,42 @@ Use the options below to override this default naming behaviour.
 .. option:: --read_flat_place <file>
 
     Reads a file containing the locations of each atom on the FPGA.
-
     This is used by the packer to better cluster atoms together.
 
     The flat placement file (which often ends in ``.fplace``) is a text file
-    where each line describes the location of an atom. It has the following
-    expected syntax:
+    where each line describes the location of an atom. Each line in the flat
+    placement file should have the following syntax:
 
     .. code-block:: none
 
-        <atom_name : str> <atom_x_pos : float> <atom_y_pos : float> <atom_sub_tile? : int> <atom_site_idx? : int>
-        n523  6 8 0 3
-        n522  6 8 0 5
-        n520  6 8 0 2
-        n518  6 8 0 16
+        <atom_name : str> <x : float> <y : float> <layer : float> <atom_sub_tile : int> <atom_site_idx? : int>
 
-    The ``sub_tile`` and ``site_idx`` are optional parameters which may be used
-    as a hint to reconstruct clusters.
+    For example:
+
+    .. code-block:: none
+
+        n523  6 8 0 0 3
+        n522  6 8 0 0 5
+        n520  6 8 0 0 2
+        n518  6 8 0 0 16
+
+    The position of the atom on the FPGA is given by 3 floating point values
+    (``x``, ``y``, ``layer``). We allow for the positions of atom to be not
+    quite legal (ok to be off-grid) since this flat placement will be fed into
+    the packer and placer, which will snap the positions to grid locations. By
+    allowing for off-grid positions, the packer can better trade-off where to
+    move atom blocks if they cannot be placed at the given position.
+    For 2D FPGA architectures, the ``layer`` should be 0.
+
+    The ``sub_tile`` is a clustered placement construct: which cluster-level
+    location at a given (x, y, layer) should these atoms go at (relevant when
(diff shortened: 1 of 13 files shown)


doc/src/vpr/command_line_usage.rst
```diff
@@ -408,6 +408,31 @@ Use the options below to override this default naming behaviour.
...
+
+    This is used by the packer to better cluster atoms together.
+
+    The flat placement file (which often ends in ``.fplace``) is a text file
+    where each line describes the location of an atom. It has the following
+    expected syntax:
+
+    .. code-block:: none
+
+        <atom_name : str> <atom_x_pos : float> <atom_y_pos : float> <atom_sub_tile? : int> <atom_site_idx? : int>
```
vaughnbetz, line 423: We should add layer so this is 3D-ready.
Do all atoms have to be specified or can a partial placement be specified (either is OK, but it should be documented).

doc/src/vpr/command_line_usage.rst
```diff
@@ -408,6 +408,31 @@ Use the options below to override this default naming behaviour.
...
+        n523  6 8 0 3
+        n522  6 8 0 5
+        n520  6 8 0 2
+        n518  6 8 0 16
+
+    The ``sub_tile`` and ``site_idx`` are optional parameters which may be used
```
vaughnbetz, line 429: sub_tile is a clustered placement construct: which cluster-level location at a given (x,y,layer) should these atoms go at (relevant when multiple clusters can be stacked there).  The location of a clustered block is specified by (x, y, layer, sub_tile).  site_idx is an index within a cluster.  
(Documentation text should be updated as currently it makes both sub_tile and site_idx sound similar).

I guess making sub_tile optional is in case an analytic placer doesn't understand block stacking?  I guess that is reasonable ...

vpr/src/base/FlatPlacementInfo.h
```diff
@@ -0,0 +1,71 @@
...
+    static constexpr int UNDEFINED_SITE_IDX = -1;
+
+    /// @brief The x-positions of each atom block. Is UNDEFINED_POS if undefined.
+    vtr::vector<AtomBlockId, float> blk_x_pos;
+    /// @brief The y-positions of each atom block. Is UNDEFINED_POS if undefined.
+    vtr::vector<AtomBlockId, float> blk_y_pos;
+    /// @brief The sub tile location of each atom block. Is UNDEFINED_SUB_TILE
+    ///        if undefined.
+    vtr::vector<AtomBlockId, int> blk_sub_tile;
+    /// @brief The flat site idx of each atom block. Is UNDEFINED_SITE_IDX if
```
vaughnbetz, line 36: Explain what a flat_site_idx is -- an optional index into a linearized list of primitive locations within a cluster-level block.

Also, should add layer.

vpr/src/base/FlatPlacementInfo.h
```diff
@@ -0,0 +1,71 @@
...
+    /// @brief Identifier for an undefined position.
+    static constexpr float UNDEFINED_POS = -1.f;
+    /// @brief Identifier for an undefined sub tile.
+    static constexpr int UNDEFINED_SUB_TILE = -1;
+    /// @brief Identifier for an undefined site idx.
+    static constexpr int UNDEFINED_SITE_IDX = -1;
+
+    /// @brief The x-positions of each atom block. Is UNDEFINED_POS if undefined.
+    vtr::vector<AtomBlockId, float> blk_x_pos;
+    /// @brief The y-positions of each atom block. Is UNDEFINED_POS if undefined.
```
vaughnbetz, line 31: Explain why these are floats (OK to be off grid / not quite legal)

vpr/src/base/load_flat_place.cpp
```diff
@@ -77,6 +82,84 @@ void write_flat_placement(const char* flat_place_file_path,
...
+        }
+
+        // Get the x and y position of the atom. These functions have error
+        // checking built in. We parse x and y as floats to allow for reading
+        // in more global atom positions.
+        flat_placement_info.blk_x_pos[atom_blk_id] = vtr::atof(tokens[1]);
+        flat_placement_info.blk_y_pos[atom_blk_id] = vtr::atof(tokens[2]);
+
+        // If a sub-tile is given, parse the sub-tile as an integer.
+        if (tokens.size() >= 4 && tokens[3][0] != '#')
```
vaughnbetz, line 147: I think you should add layer and make sub_tile mandatory.  You could choose to default layer to 0 if you wanted, but maybe it makes the format too complicated.

vpr/src/pack/pack.h
```diff
@@ -17,7 +18,8 @@ bool try_pack(t_packer_opts* packer_opts,
               const t_model* user_models,
```
vaughnbetz, line 18: It would be good to write a doxygen comment for this routine, documenting at least the flat_placement_info parameter (and more if you know what they are for).

vpr/src/pack/verify_flat_placement.cpp
```diff
@@ -0,0 +1,94 @@
...
+            if (flat_placement_info.blk_x_pos[mol_blk_id] != root_pos_x ||
+                flat_placement_info.blk_y_pos[mol_blk_id] != root_pos_y ||
+                flat_placement_info.blk_sub_tile[mol_blk_id] != root_sub_tile) {
+                VTR_LOG_ERROR(
+                    "Molecule with root atom block %s contains atom block %s "
+                    "which is not at the same position as the root atom "
+                    "block.\n",
+                    atom_netlist.block_name(root_blk_id).c_str(),
+                    atom_netlist.block_name(mol_blk_id).c_str());
+                num_errors++;
```
vaughnbetz, line 87: I'd test with large carry chains to see what molecule is created -- does it span multiple logic blocks?  If so, the x and y of all blocks in the molecule wouldn't have to be the same (and in fact shouldn't be).

vpr/src/pack/verify_flat_placement.cpp
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+    }
+
+    return num_errors;
```
vaughnbetz, line 92: Good to add layer (or at least a TODO that it should be added soon).

vaughnbetz (commented):
Some suggested updates to commenting, and ideally we'd support 3D layer too.

Should there be a test made that reading in a file works at this point?

vaughnbetz (approved):
Looks good.  I suggest trying a bunch of broken files right now (2 things on a line, 8 things on a line, some atoms missing, some atoms duplicated, some atom names that don't exist, -ve x, -ve y, etc.).  Doesn't have to be automated but worth trying right now.

Possibly you can add max val checks on x,y,layer and subtile when the device is not auto-sized, and could wrap them in a get_max_coord or some such function that keeps the code simple (returns +max_int or something if the device size isn't known yet).

PR #2862, opened 2025-01-16
Increase C++ version from 17 to 20

soheilshahrouz:
To see if CI passes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -439,7 +439,6 @@ jobs:
         - { name: 'GCC 11 (Ubuntu Noble - 24.04)',    eval: 'CC=gcc-11 && CXX=g++-11',           }
         - { name: 'GCC 12 (Ubuntu Noble - 24.04)',    eval: 'CC=gcc-12 && CXX=g++-12',         }
         - { name: 'GCC 14 (Ubuntu Noble - 24.04)',    eval: 'CC=gcc-14 && CXX=g++-14',         }
-        - { name: 'Clang 15 (Ubuntu Noble - 24.04)',  eval: 'CC=clang-15 && CXX=clang++-15',   }
         - { name: 'Clang 16 (Ubuntu Noble - 24.04)',  eval: 'CC=clang-16 && CXX=clang++-16',   }
         - { name: 'Clang 17 (Ubuntu Noble - 24.04)',  eval: 'CC=clang-17 && CXX=clang++-17',   }
         - { name: 'Clang 18 (Ubuntu Noble - 24.04)',  eval: 'CC=clang-18 && CXX=clang++-18',   }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,7 +160,7 @@ else()
         "-Wcast-align"                  #Warn if a cast causes memory alignment changes
         "-Wshadow"                      #Warn if local variable shadows another variable
         "-Wformat=2"                    #Sanity checks for printf-like formatting
-        "-Wno-format-nonliteral"        # But don't worry about non-literal formtting (i.e. run-time printf format strings)
+        "-Wno-format-nonliteral"        # But don't worry about non-literal formatting (i.e. run-time printf format strings)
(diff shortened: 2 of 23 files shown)


AmirhosseinPoolad:
There's been a breaking change in C++20 regarding templates and constructors, a lot of the warnings and errors are because of that.

https://godbolt.org/z/7n6saxzee
https://stackoverflow.com/questions/63513984/can-class-template-constructors-have-a-redundant-template-parameter-list-in-c2

AlexandreSinger:
@AmirhosseinPoolad That is correct, I recall seeing warnings of that when I was upgrading to newer versions of GCC. Those actually should be fixed anyways since they are strange. The issue comes from:

```cpp
template<T>
class MyClass {
  MyClass<T>(...); // constructor
};
```

The issue with the code above is that it looks like you are specifying a specific constructor for a specific type. But that type is the template argument. I think this causes duplicate constructors in the compiler which may increase compile time (or at worst cause the constructor to be all wonky). This error in c++20 may be actually fixing some of these issues.

AmirhosseinPoolad:
Talked to Soheil about it and there's also a whole lot of errors/warnings because C++20 requires more functions to be declared noexcept and we have -Wnoexcept on. Also trivially fixable (AFAIK) but it's a lot of busy work.

AlexandreSinger:
@soheilshahrouz FYI, I do not think this is an error. I think this is a warning. But for the regression test and treat all warnings as errors to keep the code warning clean.
Screenshot from 2025-01-17 10-36-45

We explicitly turn on these warnings in the CMake file here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/b8c1ef15737b560d5d456383cc947821be44ef3a/CMakeLists.txt#L173-L175

It could be possible that our code is causing this issue, but based on what I found online it may be the STL that has some issues:
https://stackoverflow.com/a/59381014

I think we can find ways to add `noexcept` to some classes (for example, strong IDs seem to be the main culprit!).

Worst case scenario, we can turn this warning off; but I do not like that idea.

soheilshahrouz:
Compilation with clang 15 fails. It is probably related to [this](https://stackoverflow.com/questions/64300832/why-does-clang-think-gccs-subrange-does-not-satisfy-gccs-ranges-begin-functi) and [this](https://github.com/llvm/llvm-project/issues/44178).

Is it ok to remove clang 15 from CI tests?

AlexandreSinger:
> Compilation with clang 15 fails. It is probably related to [this](https://stackoverflow.com/questions/64300832/why-does-clang-think-gccs-subrange-does-not-satisfy-gccs-ranges-begin-functi) and [this](https://github.com/llvm/llvm-project/issues/44178).
> 
> Is it ok to remove clang 15 from CI tests?

I agree with removing Clang 15 for the reason that Soheil gives. The culture behind Clang releases is slightly different from GCC, where older versions do not get as much long term support as GCC in my opinion. Based on the Clang C++ standard support page ([here](https://clang.llvm.org/cxx_status.html)), Clang 16 and 17 saw a lot of the c++-20 support. Based on the GCC c++ standard support page ([here](https://gcc.gnu.org/projects/cxx-status.html)), GCC-11 should support basically all of c++-20 features.

To replace clang 15, we may be want to add Clang 19 to the CI (which is the most recent version of Clang I could find; but I am not sure if the GitHub runners support this version of Clang or not yet). However, that should be done in a different PR.

@vaughnbetz What do you think? If you agree, I can make Clang 15 no longer be a required test and we can turn off that test in this PR.

AlexandreSinger (approved):
These changes all LGTM. Even without changing c++ version, all of these changes are good to have and may make the code slightly faster (will allow the compiler to make things a bit faster).

vaughnbetz:
I'm OK with dropping clang 15 to enable C++ 20

AlexandreSinger:
I have removed Clang-15 from the branch protection rules on the Master branch (it is no longer required). @soheilshahrouz feel free to remove the Clang-15 test from the CI tests.

vaughnbetz (approved):
LGTM

PR #2799, opened 2024-11-05
Chan x/y placement cost factors using prefix sum

soheilshahrouz:
Implements the prefix sum idea used for chanz in #2781 for chanx/y.

PR #2863, opened 2025-01-16
Remove place_cost_exp from documentation

soheilshahrouz:
place_cost_exp was set to 1 by default. This option was removed in #2799. This PR removes this command line option from documentation.


AlexandreSinger (approved):

vaughnbetz:
Thanks. The changes look good. Just to confirm: the dusty_sa annealing schedule has already been removed from the annealing code, correct? (I did a quick search and didn't find it, so I think it's gone, in which case removing from the documentation in this PR makes sense).

Issue #2373, opened 2023-08-21
Toggle Routing Cong Cost seg faulting


alanbu1:
Clicking an option under the "toggle routing cong cost" under the routing menu results in a segmentation fault.

vaughnbetz:
@nedsels When you move on to graphics please test this (with and without flat routing) and see if there is still a problem. I think it is gone but am not sure.

andrecavalcante:
This issue persists.

vaughnbetz:
@AmirhosseinPoolad : since you're playing with the graphics, can you try this option and see if there is a problem?  You can try it with a successful routing (shouldn't show any congestion) and with a failed one; you can run with 
--route_chan_width 100 --max_router_iterations 10  (or some other related numbers) to get the router to fail on a not-too-big circuit.

AmirhosseinPoolad:
I think the PR I just sent fixes this, but maybe we should use at() instead of operator[] to access vectors to avoid this type of issue in the future. At least in places where performance is not as important.

vaughnbetz:
Thanks; fixed by #2865.
We could use at() more in non-performance critical code, but running the sanitized build (or valgrind) provides equivalent checking with no runtime penalty in release build.

PR #2865, opened 2025-01-16
Fixed segfault in draw_routing_costs

AmirhosseinPoolad:
We didn't allocate any space for the 'rr_node_costs' vector and accessed it directly using operator[], resulting in segfaults. This commit fixes the crash by preallocating all the needed space before any accesses.

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2373


vaughnbetz:
Thanks!

PR #2867, opened 2025-01-21
More fixes to sync_netlists_to_routing_flat

fkosar-ql:
This PR contains a number of fixes to enable OpenFPGA bitstream generation using flat router results:

* "Unset" port equivalence in the architecture (in memory) before flat routing starts. This is for compatibility with OpenFPGA (also may simplify the sync code)
* Rename new cluster nets created while rebuilding the clustered netlist more consistently (avoid name clashes)
* In `annotate_rr_node_nets`, don't fail right away if two cluster nets point to the same RR nodes, instead check if they are mapped to the same atom net. (This lets OpenFPGA have net -> rr node annotations without having to split the flat route trees if an atom net produced two or more cluster nets)


vaughnbetz:
Crash in flat routing so some bug still.
Please add an output message (that is easy to interpret) when you turn off pin equivalence in memory (should say you're doing this and why).
Should run the nightly tests manually once basic CI passes.

vpr/src/base/netlist_writer.cpp
```diff
@@ -1245,7 +1245,9 @@ class NetlistWriterVisitor : public NetlistVisitor {
 
         //Add the single output connection
         {
-            auto atom_net_id = top_pb_route[sink_cluster_pin_idx].atom_net_id; //Connected net in atom netlist
+            AtomNetId atom_net_id = AtomNetId::INVALID();
+            if (top_pb_route.count(sink_cluster_pin_idx))
```
vaughnbetz, line 1249: Suggest putting a comment here saying what this is doing (as best you understand it anyway).

vpr/src/base/vpr_api.cpp
```diff
@@ -115,6 +116,9 @@ static void get_intercluster_switch_fanin_estimates(const t_vpr_setup& vpr_setup
                                                     int* opin_switch_fanin,
                                                     int* wire_switch_fanin,
                                                     int* ipin_switch_fanin);
+
+static void unset_port_equivalences(DeviceContext& device_ctx);
```
vaughnbetz, line 120: Doxygen comment.

vaughnbetz: (Explain what this does and why it exists)

vpr/src/base/vpr_api.cpp
```diff
@@ -369,6 +373,25 @@ void vpr_init_with_options(const t_options* options, t_vpr_setup* vpr_setup, t_a
     device_ctx.pad_loc_type = vpr_setup->PlacerOpts.pad_loc_type;
 }
 
+/** Port equivalence does not make sense during flat routing.
+ * Remove port equivalence from all ports in the architecture */
+static void unset_port_equivalences(DeviceContext& device_ctx){
+    for(auto& physical_type: device_ctx.physical_tile_types){
```
vaughnbetz, line 379: Add an output message saying you're doing this.
Have a somewhat more expansive comment (why are you doing this?) at the declaration.

vpr/src/pack/sync_netlists_to_routing_flat.cpp
```diff
@@ -297,26 +275,26 @@ static void sync_clustered_netlist_to_routing(void){
 
     for(auto net_id: clb_netlist.nets()){
         auto atom_net_id = atom_lookup.atom_net(net_id);
-        if(!clock_net_is_routed && route_ctx.is_clock_net[atom_net_id])
+        if(!route_ctx.route_trees[ParentNetId(int(atom_net_id))])
```
vaughnbetz, line 278: Maybe comment why you're doing this?

vpr/src/pack/sync_netlists_to_routing_flat.cpp
```diff
@@ -297,26 +275,26 @@ static void sync_clustered_netlist_to_routing(void){
...
-            continue;
-
-        pins_to_remove.push_back(pin_id);
-    }
     for(auto port_id: clb_netlist.ports()){
-        ClusterNetId clb_net_id = clb_netlist.port_net(port_id, 0);
-        auto atom_net_id = atom_lookup.atom_net(clb_net_id);
-        if(!clock_net_is_routed && atom_net_id && route_ctx.is_clock_net[atom_net_id])
-            continue;
+        size_t skipped_pins = 0;
```
vaughnbetz, line 284: Maybe comment what this for loop is doing / goal?

vpr/src/route/annotate_routing.cpp
```diff
@@ -47,11 +49,18 @@ vtr::vector<RRNodeId, ClusterNetId> annotate_rr_node_nets(const ClusteringContex
                  * In some routing architectures, node capacity is more than 1
                  * which allows a node to be mapped by multiple nets
                  * Therefore, the sanity check should focus on the nodes
-                 * whose capacity is 1
+                 * whose capacity is 1.
+                 * Flat routing may create two clustered nets from a single
```
vaughnbetz, line 53: Might want to say why / what makes two clustered nets (split a single atom net onto two cluster block outputs)

vaughnbetz (commented):
Looks mostly good -- some commenting requested.

fkosar-ql:
No idea how it's passing with the current version of sync_netlists_to_routing_flat, but `spree.v` seems to connect some global nets to CLBs and routing fails when I try to run with `--constant_net_method route`. This means the synchronized results will be invalid since the flat router does not see which pins the globals are using.

For now I'm just adding a warning for that case to make the CI work, but it should be handled at some point.

amin1377:
Adding @AmirhosseinPoolad as a reviewer for the code I modified in the packer.

vpr/src/pack/sync_netlists_to_routing_flat.cpp
```diff
@@ -185,18 +168,16 @@ static void sync_pb_routes_to_routing(void) {
...
+        /* Don't erase entries for nets without routing in place (clocks, globals...) */
         std::vector<int> pins_to_erase;
         auto& pb_routes = cluster_ctx.clb_nlist.block_pb(clb_blk_id)->pb_route;
         for (auto& [pin, pb_route] : pb_routes) {
-            if (clock_net_is_routed || !route_ctx.is_clock_net[pb_route.atom_net_id])
-                pins_to_erase.push_back(pin);
+            /* No route tree: no routing in place, it is global or clock */
+            if (!route_ctx.route_trees[ParentNetId(int(pb_route.atom_net_id))])
+                continue;
+            pins_to_erase.push_back(pin);
```
AmirhosseinPoolad, line 180: I'm not sure if it's safe to delete elements of a vector with this method. From what I understand, you're adding indexes/keys to a flat_map2 to a list and then erasing those indexes. But it says this in the documentation for flat_map2:
>  - Iterators may be invalidated on insertion/erase (i.e. if the vector is reallocated)

vpr/src/pack/sync_netlists_to_routing_flat.cpp
```diff
@@ -276,37 +257,37 @@ static void sync_clustered_netlist_to_routing(void) {
...
+     * Do not remove entries for nets without an existing route tree,
+     * since we don't have the information to rebuild those parts. */
     std::vector<ClusterNetId> nets_to_remove;
     std::vector<ClusterPinId> pins_to_remove;
     std::vector<ClusterPortId> ports_to_remove;
 
     for (auto net_id : clb_netlist.nets()) {
         auto atom_net_id = atom_lookup.atom_net(net_id);
-        if (!clock_net_is_routed && route_ctx.is_clock_net[atom_net_id])
+        if (!route_ctx.route_trees[ParentNetId(int(atom_net_id))])
```
AmirhosseinPoolad, line 269: Same as above regarding the cast.

vpr/src/pack/sync_netlists_to_routing_flat.cpp
```diff
@@ -276,37 +257,37 @@ static void sync_clustered_netlist_to_routing(void) {
...
-        ClusterNetId clb_net_id = clb_netlist.port_net(port_id, 0);
-        auto atom_net_id = atom_lookup.atom_net(clb_net_id);
-        if (!clock_net_is_routed && atom_net_id && route_ctx.is_clock_net[atom_net_id])
-            continue;
+        size_t skipped_pins = 0;
+
+        for (auto pin_id : clb_netlist.port_pins(port_id)) {
+            ClusterNetId clb_net_id = clb_netlist.pin_net(pin_id);
+            auto atom_net_id = atom_lookup.atom_net(clb_net_id);
+            if (atom_net_id && !route_ctx.route_trees[ParentNetId(int(atom_net_id))]) {
```
AmirhosseinPoolad, line 282: Ditto

vpr/src/pack/cluster_router.cpp
```diff
@@ -545,13 +551,15 @@ bool try_intra_lb_route(t_lb_router_data* router_data,
...
-t_pb_routes alloc_and_load_pb_route(const std::vector<t_intra_lb_net>* intra_lb_nets, t_pb_graph_node* pb_graph_head) {
+t_pb_routes alloc_and_load_pb_route(const std::vector<t_intra_lb_net>* intra_lb_nets,
+                                    t_logical_block_type_ptr logic_block_type,
+                                    const IntraLbPbPinLookup& intra_lb_pb_pin_lookup) {
     const std::vector<t_intra_lb_net>& lb_nets = *intra_lb_nets;
-    int total_pins = pb_graph_head->total_pb_pins;
+    int total_pins = logic_block_type->pb_graph_head->total_pb_pins;
     t_pb_routes pb_route;
 
     for (int inet = 0; inet < (int)lb_nets.size(); inet++) {
```
AmirhosseinPoolad, line 561: Since we're here, can you change inet to be size_t to get rid of the ugly cast? Thanks!

AmirhosseinPoolad (changes requested):
Thank you for the changes @amin1377! I just had a bunch of nitpicks.

vpr/src/pack/cluster_router.cpp
AmirhosseinPoolad: Or even better, use a range-based for loop instead.

vpr/src/pack/cluster_router.cpp
```diff
@@ -545,13 +551,15 @@ bool try_intra_lb_route(t_lb_router_data* router_data,
 
 /* Creates an array [0..num_pb_graph_pins-1] lookup for intra-logic block routing.  Given pb_graph_pin id for clb, lookup atom net that uses that pin.
  * If pin is not used, stores OPEN at that pin location */
```
vaughnbetz, line 553: This comment should be made Doxygen compatible and moved to the .h file

vaughnbetz: It would be good to beef it up to include documentation of the parameters too if you understand them all.  If not, do what you can :).
Is the type t_pb_routes the specified array?  If so, good to mention it in the comment as right now the return type you document isn't an obvious match for the function signature.

vpr/src/pack/cluster_router.cpp
```diff
@@ -582,7 +590,13 @@ void free_intra_lb_nets(std::vector<t_intra_lb_net>* intra_lb_nets) {
  ****************************************************************************/
 
 /* Recurse through route tree trace to populate pb pin to atom net lookup array */
-static void load_trace_to_pb_route(t_pb_routes& pb_route, const int total_pins, const AtomNetId net_id, const int prev_pin_id, const t_lb_trace* trace) {
+static void load_trace_to_pb_route(t_pb_routes& pb_route,
```
vaughnbetz, line 593: Where is the "pb pin to atom net lookup array" stored?  Inside pb_route?  This should be documented in a bigger Doxygen comment for this routine.

vpr/src/pack/cluster_router.h
```diff
@@ -23,7 +23,9 @@ bool try_intra_lb_route(t_lb_router_data* router_data, int verbosity, t_mode_sel
 void reset_intra_lb_route(t_lb_router_data* router_data);
 
 /* Accessor Functions */
```
vaughnbetz, line 25: Make a proper Doxygen comment.  "Accessor functions" seems out of date -- suggest deleting that line.

vpr/src/pack/sync_netlists_to_routing_flat.cpp
```diff
@@ -378,6 +359,15 @@ static void sync_clustered_netlist_to_routing(void) {
             }
             PinType pin_type = node_type == OPIN ? PinType::DRIVER : PinType::SINK;
 
+            /* Pin already exists. This means a global was connected to here. */
```
vaughnbetz, line 362: I think you should say it was an "a global net that was not routed (i.e. 'ideal' mode)"

vaughnbetz (commented):
Thanks! Some commenting enhancements suggested.

vpr/src/pack/cluster_router.cpp
```diff
@@ -593,12 +607,14 @@ static void load_trace_to_pb_route(t_pb_routes& pb_route, const int total_pins,
             pb_route.insert(std::make_pair(cur_pin_id, t_pb_route()));
             pb_route[cur_pin_id].atom_net_id = net_id;
             pb_route[cur_pin_id].driver_pb_pin_id = driver_pb_pin_id;
+            auto pb_graph_pin = intra_lb_pb_pin_lookup.pb_gpin(logic_block_type->index, cur_pin_id);
```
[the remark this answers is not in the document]
amin1377: t_pb_graph_pin is the name of the class itself. I'm not sure if it's good practice to name a variable exactly the same as the class.

vpr/src/pack/sync_netlists_to_routing_flat.cpp
amin1377: We store the physical pin numbers in pins_to_erase and later remove the pb_routes corresponding to those pin numbers.

vpr/src/pack/cluster_router.cpp
amin1377: Added the following comment:

```
/**
 * @brief Creates an array [0..num_pb_graph_pins-1] for intra-logic block routing lookup. 
 * Given a pb_graph_pin ID for a CLB, this lookup returns t_pb_route corresponding to that
 * pin.
 *
 * @param intra_lb_nets Vector of intra-logic block nets.
 * @param logic_block_type Logic block type of the current cluster.
 * @param intra_lb_pb_pin_lookup Intra-logic block pin lookup to get t_pb_graph_pin from a pin ID.
 * @return t_pb_routes An array [0..num_pb_graph_pins-1] for intra-logic block routing lookup.
 */
```

amin1377:
Thanks @vaughnbetz  and @AmirhosseinPoolad  for reviewing the code. I've addressed your comments. I also ran the nightly tests a few days ago to ensure they are passing ([[Link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14682373939)]). Aside from some AP tests (which are also failing on the master branch, [[Link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14688741280)]), all other tests passed. 

@AlexandreSinger  FYI.

vpr/src/pack/sync_netlists_to_routing_flat.cpp
AmirhosseinPoolad: I see now, you're not working directly with the vector indices. Marking as resolved.

AmirhosseinPoolad (approved):

AlexandreSinger:
> Thanks @vaughnbetz and @AmirhosseinPoolad for reviewing the code. I've addressed your comments. I also ran the nightly tests a few days ago to ensure they are passing ([[Link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14682373939)]). Aside from some AP tests (which are also failing on the master branch, [[Link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14688741280)]), all other tests passed.
> 
> @AlexandreSinger FYI.

@amin1377 Thanks for pointing this out! I got the notification this morning. I have a fix that I am testing now. Will raise a PR today fixing the AP failure.

amin1377:
@AmirhosseinPoolad All tests are passing, and the merge conflicts have been resolved. Feel free to merge the PR if you have no further comments.

Issue #2858, opened 2025-01-14
[Pack] Random Number Generator in Packer is Non-Deterministic


AlexandreSinger:
Within the GreedyCandidateSelector used in the packer, there is code which adds cluster molecule candidates by attraction groups. In this code, when the number of available atoms in the attraction group is larger than some threshold (currently 500), it randomly selects candidates from this group:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/e46e9c22feeb5ec7921bb702206d5bd5c1ed8834/vpr/src/pack/greedy_candidate_selector.cpp#L785-L807

The issue is that the random number generator used here is way overkill for this application and is non-deterministic. We should be using `vtr::irand` instead which would be faster.

PR #2871, opened 2025-01-24
Changed the random number generator in packer to deterministic vtr::irand

haydar-c:
#### Description

Changed the random number generator in add_cluster_molecule_candidates_by_attraction_group in packer to be vtr::irand. Initiated the vtr::RngContainer in do_clustering before each reference of try_glow_cluster.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2858

[Pack] Random Number Generator in Packer is Non-Deterministic #2858
#### Motivation and Context

Aim is to make the adding cluster molecule candidate by attraction groups in packing deterministic.  

Local tests: vtr_reg_basic, vtr_reg_strong
CI Test Workflow

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/pack/greedy_candidate_selector.cpp
+++ b/vpr/src/pack/greedy_candidate_selector.cpp
@@ -77,7 +77,8 @@ GreedyCandidateSelector::GreedyCandidateSelector(
                           is_clock_(is_clock),
                           is_global_(is_global),
                           net_output_feeds_driving_block_input_(net_output_feeds_driving_block_input),
-                          timing_info_(timing_info) {
+                          timing_info_(timing_info),
+                          rng_(0) {
     // Initialize the list of molecules to pack, the clustering data, and the
     // net info.
 
@@ -779,17 +780,9 @@ void GreedyCandidateSelector::add_cluster_molecule_candidates_by_attraction_grou
         return;
     }
 
-    int min = 0;
-    int max = num_available_atoms - 1;
-
     for (int j = 0; j < attraction_group_num_atoms_threshold_; j++) {
-        // FIXME: This is a non-deterministic random number generator and it is
-        //        overkill to what this needs to be. Should use vtr::irand which
-        //        would be faster.
-        std::random_device rd;
-        std::mt19937 gen(rd());
-        std::uniform_int_distribution<> distr(min, max);
-        int selected_atom = distr(gen);
+        //Get a random atom between 0 and the number of available atoms - 1
+        int selected_atom = rng_.irand(num_available_atoms - 1);
 
         AtomBlockId blk_id = available_atoms[selected_atom];
 
--- a/vpr/src/pack/greedy_candidate_selector.cpp
+++ b/vpr/src/pack/greedy_candidate_selector.cpp
@@ -17,7 +17,6 @@
 #include "vpr_context.h"
 #include "vpr_types.h"
 #include "vtr_assert.h"
-#include "vtr_random.h"
 
 /*
  * @brief Get gain of packing molecule into current cluster.
@@ -519,8 +518,7 @@ t_pack_molecule* GreedyCandidateSelector::get_next_candidate_for_cluster(
                                 LegalizationClusterId cluster_id,
                                 const ClusterLegalizer& cluster_legalizer,
                                 const Prepacker& prepacker,
-                                AttractionInfo& attraction_groups,
-                                vtr::RngContainer& rng) {
+                                AttractionInfo& attraction_groups) {
     /* Finds the block with the greatest gain that satisfies the
      * input, clock and capacity constraints of a cluster that are
      * passed in.  If no suitable block is found it returns nullptr.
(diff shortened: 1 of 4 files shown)


vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -779,17 +783,10 @@ void GreedyCandidateSelector::add_cluster_molecule_candidates_by_attraction_grou
...
 
     for (int j = 0; j < attraction_group_num_atoms_threshold_; j++) {
-        // FIXME: This is a non-deterministic random number generator and it is
-        //        overkill to what this needs to be. Should use vtr::irand which
-        //        would be faster.
-        std::random_device rd;
-        std::mt19937 gen(rd());
-        std::uniform_int_distribution<> distr(min, max);
-        int selected_atom = distr(gen);
+        int selected_atom = rng.irand(max);
```
AlexandreSinger, line 789: Small nit-pick here. I would inline the max term into the irand:
```
int select_atom = rng.irand(num_available_atoms - 1)
```
And then I would add a comment above this line detailing what it does. "Get a random atom between 0 and the number of available atoms - 1" or something. This will help future-proof this function.

vpr/src/pack/greedy_candidate_selector.h
```diff
@@ -16,6 +16,7 @@
 #include "cluster_legalizer.h"
 #include "physical_types.h"
 #include "vtr_vector.h"
+#include "vtr_random.h"
```
AlexandreSinger, line 19: Since you are not using the RngContainer directly, you should forward declare the RngContainer instead of including it. See forward declarations below.

We do this to help keep the compilation time low. The way it is written now, if someone changes vtr_random, ALL files that include the greedy candidate selector will have to be completely rebuilt. If you forward declare it, it will not be the case.

vpr/src/pack/greedy_candidate_selector.h
```diff
@@ -301,7 +302,8 @@ class GreedyCandidateSelector {
                                     LegalizationClusterId cluster_id,
                                     const ClusterLegalizer& cluster_legalizer,
                                     const Prepacker& prepacker,
-                                    AttractionInfo& attraction_groups);
+                                    AttractionInfo& attraction_groups,
+                                    vtr::RngContainer& rng);
```
AlexandreSinger, line 306: Please update the Doxygen comment above with the new parameter follow the style of the other parameters.

vpr/src/pack/greedy_clusterer.cpp
```diff
@@ -160,6 +161,8 @@ GreedyClusterer::do_clustering(ClusterLegalizer& cluster_legalizer,
 
     print_pack_status_header();
 
+    vtr::RngContainer rng(0);
```
AlexandreSinger, line 164: I consider this more initialization than clustering, I would move this line higher to at most line 157. Also you should add a comment above this line specifying what you are doing and why you are doing it.

vpr/src/pack/greedy_clusterer.h
```diff
@@ -13,6 +13,7 @@
 #include <vector>
 #include "cluster_legalizer.h"
 #include "physical_types.h"
+#include "vtr_random.h"
```
AlexandreSinger, line 16: Same with the comment I made in the candidate selector header file. We do not want to include header files in other header files if we do not have to. Please forward declare the RngContainer class.

AlexandreSinger (changes requested):
@haydar-c Thank you for this change! I have some style comments that I think would improve your change and improve the documentation.

vpr/src/pack/greedy_clusterer.cpp
soheilshahrouz, line 164: I guess that the only user of this RngContainer object is GreedyCandidateSelector. If this is the case, I think it makes more sense to add it as a member variable to GreedyCandidateSelector and initilize it in its constructor instead of passing it down a deep function call hierarchy.

AlexandreSinger: I agree! That would make things much cleaner! Especially since it is an independent object!

AlexandreSinger (approved):
@haydar-c This is a beautiful PR now! I really like this implementation! Great job. LGTM

haydar-c:
> @haydar-c This is a beautiful PR now! I really like this implementation! Great job. LGTM

Thanks for the review! Merging this now.

AlexandreSinger:
@haydar-c Perhaps wait to merge this until Vaughn has a quick look!

haydar-c:
> @haydar-c Perhaps wait to merge this until Vaughn has a quick look!
> 
> Edit: He may have other comments than mine.

Sure, still learning the conventions. Thanks for pointing that out.

vpr/src/pack/greedy_candidate_selector.h
```diff
@@ -516,5 +517,9 @@ class GreedyCandidateSelector {
     /// @brief A count on the number of unrelated clustering attempts which
     ///        have been performed.
     int num_unrelated_clustering_attempts_ = 0;
+
+    /// @brief Random number generator to get a random atom between 0 and
```
vaughnbetz, line 521: I'd make this slightly more general to start.
I'd change this to:
@brief Random number generator used by the clusterer. Currently this is used only when selecting atoms from
attraction groups, but could be used for other purposes in the future.

vaughnbetz (commented):
Looks good; I suggest one commenting tweak.

AlexandreSinger:
LGTM. Merging.

PR #2873, opened 2025-01-30
Add initial nightly test workflow

AmirhosseinPoolad:
This commit adds the new nightly test workflow to the list of workflows. Unfortunately to test GitHub actions workflows there has to be a version of it in the master branch, but after this PR is merged we can debug and resolve any issues in the "savi_nightly_test" branch so thankfully we don't have to test on master. This workflow can only be manually dispatched so there shouldn't be any automatic runs that could overwhelm the SAVI machines while we're working on the workflow and fixing any potential issues.

### DIFF
--- a/.github/workflows/nightly_test_manual.yml
+++ b/.github/workflows/nightly_test_manual.yml
@@ -1,0 +1,71 @@
+name: NightlyTestManual
+
+# This workflow can only be dispatched.
+on:
+  workflow_dispatch:
+
+# We want to cancel previous runs for a given PR or branch / ref if another CI
+# run is requested.
+# See: https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  # default compiler for all non-compatibility tests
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly_test_manual.yml
```diff
@@ -0,0 +1,71 @@
...
+# We want to cancel previous runs for a given PR or branch / ref if another CI
+# run is requested.
+# See: https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  # default compiler for all non-compatibility tests
+  MATRIX_EVAL: "CC=gcc-11 && CXX=g++-11"
```
AlexandreSinger, line 16: I think this needs to be GCC 13. This was a leftover before I upgraded the CI to Ubuntu 24.04.

AlexandreSinger (commented):
Hi @AmirhosseinPoolad Thank you for raising this PR! I had a small comment on the GCC version. Its best just to fix this now, so we are testing on the correct version!

Other than that, LGTM

AmirhosseinPoolad:
Good catch!

PR #2874, opened 2025-01-31
Update VTR 9 branch with faster check_non_configurable_edges

soheilshahrouz:
We previously checked every node in a route tree against all non-configurable RR sets, leading to O(n²) complexity. This PR limits checks to RR sets within the route tree, reducing complexity to O(n). As a result, the check_all_non_configurable_edges() function's execution time dropped from 2045s to 1.1s on the bgm benchmark in a 7-series architecture.


AlexandreSinger:
The checks that are not passing will never complete since they are only on the Master branch. @vaughnbetz Who should merge this?

PR #2880, opened 2025-02-03
[WIP] Per-edge delays

soheilshahrouz:
This PR enables VPR to ingest a file that adds edge-specific delays to switch delays. This way, RR edges that represent the same switch type with the same fanin, can have different delays.


libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -354,6 +360,7 @@ void t_rr_graph_storage::assign_first_edges() {
     size_t num_edges = edge_src_node_.size();
     VTR_ASSERT(edge_dest_node_.size() == num_edges);
     VTR_ASSERT(edge_switch_.size() == num_edges);
+    VTR_ASSERT(edge_switch_offset_inf_.size() == num_edges);
```
vaughnbetz, line 363: I think one flyweight for edge data would be best, rather than two (switch_inf and edge_switch_offset_inf).  Try to store one flyweight id per edge; we can set it up the same way as today during rr-graph building and then update it if the user requests that with an option saying they want a detailed delay file.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -637,16 +649,16 @@ class t_rr_graph_storage {
...
+    //     Note: Either arch_switch_inf indices or rr_switch_inf should be
     //     used with emplace_back_edge and alloc_and_load_edges.  Do not mix
-    //     indicies, otherwise things will be break.
+    //     indices, otherwise things will be break.
     //
     //     The rr_switch_inf switches are remapped versions of the
     //     arch_switch_inf switch indices that are used when we have
     //     different delays and hence different indices based on the fanout
     //     of a switch.  Because fanout of the switch can only be computed
     //     after the graph is built, the graph is initially built using
```
vaughnbetz, line 660: We're already perturbing the switch index flyweight id, so I think this new feature is OK to just perturb the switch_inf_id some more to make a more detailed flyweight.

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -413,6 +419,20 @@ class RRGraphView {
         return node_storage_.edge_switch(id, iedge);
     }
 
+    inline short edge_switch(RREdgeId edge_id) const {
```
vaughnbetz, line 422: I like this style of wrapping all the flyweight data in accessor functions.

vaughnbetz: Check performance though, as we might do multiple flyweight lookups in sequence to get the components of delay ... if slower we might have to expose the flyweight instead (at least in hot loops).

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +116,58 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+
+        if (!(iss >> edge_id >> rr_switch_detailed_inf.Tdel)) {
+            throw std::runtime_error("Invalid line format: " + line);
+        }
+
+        if (edge_id < 0) {
+            throw std::runtime_error("Negative integer found: " + std::to_string(edge_id));
+        }
+
+        auto switch_idx = rr_graph.rr_nodes().edge_switch((RREdgeId)edge_id);
```
vaughnbetz, line 152: Good to get some feedback on the file format.  Simple table format?  xml?  binary or xml (capnproto)?  Should store delays, not offsets I think.

libs/librrgraph/src/io/rr_graph_reader.h
```diff
@@ -35,4 +35,8 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
                   const char* echo_file_name,
                   bool is_flat);
 
+void load_rr_edge_attribute_offset_file(RRGraphBuilder& rr_graph_builder,
```
vaughnbetz, line 38: Comment when you really put this back.

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -2160,6 +2169,7 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
     RRGraphBuilder* rr_graph_builder_;
     RRGraphView* rr_graph_;
     vtr::vector<RRSwitchId, t_rr_switch_inf>* rr_switch_inf_;
+    vtr::vector<RRSwitchOffsetInfoId, t_rr_switch_offset_inf>* rr_switch_offset_inf_;
```
vaughnbetz, line 2172: write out the new flyweight I think ... hard not to do that if we merge the flyweight.

vpr/src/base/SetupVPR.cpp
```diff
@@ -232,6 +232,7 @@ void SetupVPR(const t_options* options,
     SetupPackerOpts(*options, packerOpts);
     routingArch->write_rr_graph_filename = options->write_rr_graph_file;
     routingArch->read_rr_graph_filename = options->read_rr_graph_file;
+    routingArch->read_rr_edge_delay_offset_filename = options->read_rr_edge_delay_offset_file;
```
vaughnbetz, line 235: probably shouldn't call it offset -- detailed rr_delay file I think

vpr/src/base/read_options.cpp
```diff
@@ -1623,6 +1623,13 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .metavar("RR_GRAPH_FILE")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    file_grp.add_argument(args.read_rr_edge_delay_offset_file, "--read_rr_edge_delay_offset_file")
```
vaughnbetz, line 1626: Offset -> detailed delay

vpr/src/base/vpr_types.h
```diff
@@ -1416,6 +1416,7 @@ struct t_det_routing_arch {
     float R_minW_pmos;
 
     std::string read_rr_graph_filename;
+    std::string read_rr_edge_delay_offset_filename;
```
vaughnbetz, line 1419: delay_offset --> rename

vpr/src/route/connection_router.cpp
```diff
@@ -702,7 +702,10 @@ void ConnectionRouter<Heap>::evaluate_timing_driven_node_costs(RTExploredNode* t
     bool switch_buffered = rr_switch_inf_[iswitch].buffered();
     bool reached_configurably = rr_switch_inf_[iswitch].configurable();
     float switch_R = rr_switch_inf_[iswitch].R;
-    float switch_Tdel = rr_switch_inf_[iswitch].Tdel;
+    // Instead of looking up the delay from the RR switch type info,
+    // we call edge_delay() method that returns the exact delay calculated
+    // by adding edge-specific delays to the nominal delay of switch type
```
vaughnbetz, line 707: Test what the speed impact is of wrapping all the delay calculation functions in accessors, vs. getting an iswitch.
Might also be an opportunity to make the delay calculators more abstract.
Could make a method that returns a tuple of frequently used values in a hot loop if needed.

vaughnbetz: Also think about whether we should take the Elmore delay model out of the router, and just have a linear delay model that is abstracted (ask for an edge delay, get a delay and you don't know how)

vaughnbetz (commented):
Some thoughts on how to do this .... good to make a presentation on this for a Thursday meeting.

soheilshahrouz:
Superseded by #2930.

PR #2881, opened 2025-02-04
[Pack] Cleaned Up try_pack API

AlexandreSinger:
The try_pack API required an approximation for the inter-cluster delay; however, this approximation was always being made the same way and did not need to be brought in externally. Inlined this into the packer itself to make the interface easier to work with.

This cleanup will help with integrating the APPack code into the AP flow.

### DIFF
--- a/vpr/src/pack/pack.cpp
+++ b/vpr/src/pack/pack.cpp
@@ -480,9 +480,7 @@ static void get_intercluster_switch_fanin_estimates(const t_arch& arch,
      */
 
     /* Fan-in to opin/ipin/wire switches depends on whether the architecture is unidirectional/bidirectional */
-    // FIXME This line may have precision issues. Should 2 and 4 be floats?
-    //       This should also use parenthesis to make it more clear what each term means.
-    (*opin_switch_fanin) = 2 * type->num_drivers / 4 * Fc_out;
+    (*opin_switch_fanin) = 2.f * type->num_drivers / 4.f * Fc_out;
     (*wire_switch_fanin) = routing_arch.Fs;
     (*ipin_switch_fanin) = Fc_in;
     if (routing_arch.directionality == UNI_DIRECTIONAL) {
--- a/vpr/src/pack/pack.cpp
+++ b/vpr/src/pack/pack.cpp
@@ -23,12 +23,38 @@ static bool try_size_device_grid(const t_arch& arch,
                                  float target_device_utilization,
                                  const std::string& device_layout_name);
 
+/**
+ * Since the parameters of a switch may change as a function of its fanin,
+ * to get an estimation of inter-cluster delays we need a reasonable estimation
+ * of the fan-ins of switches that connect clusters together. These switches are
+ * 1) opin to wire switch
+ * 2) wire to wire switch
+ * 3) wire to ipin switch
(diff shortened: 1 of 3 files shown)


vpr/src/pack/pack.cpp
```diff
@@ -382,3 +414,134 @@ static bool try_size_device_grid(const t_arch& arch,
...
+     * An input pin of a CLB simply receives Fc_in connections.
+     *
+     * Unidirectional: Fc_in
+     * Bidirectional:  Fc_in
+     */
+
+    /* Fan-in to opin/ipin/wire switches depends on whether the architecture is unidirectional/bidirectional */
+    // FIXME This line may have precision issues. Should 2 and 4 be floats?
+    //       This should also use parenthesis to make it more clear what each term means.
+    (*opin_switch_fanin) = 2 * type->num_drivers / 4 * Fc_out;
```
AlexandreSinger, line 485: @vaughnbetz My IDE is complaining about potential precision issues with this line. It is concerned with the division by 4 using integer division when it should not (since fc_out is a float). I also did check order of operations and the multiplication by 2 happens first, division by 4 second, and multiplications by fc_out third; but I think we should use parenthesis to make this more clear. What do you think?

My thought is to turn the 2 and 4 into floats and put parenthesis around the division.

vaughnbetz:
I'm fine with putting 2. or 4. etc. in the equation to remove the IDE warning and make the fact that it is floating point clear.
But this code should not be moved to pack.cpp/pack.h.  It is code to figure out the delay for general (inter-cluster) routing muxes when a user specifies he/she wants the mux delay to be a function of its size.  So the code should either be left where it is or (probably even better) moved to an rr_graph building-related file.

AlexandreSinger:
> I'm fine with putting 2. or 4. etc. in the equation to remove the IDE warning and make the fact that it is floating point clear. But this code should not be moved to pack.cpp/pack.h. It is code to figure out the delay for general (inter-cluster) routing muxes when a user specifies he/she wants the mux delay to be a function of its size. So the code should either be left where it is or (probably even better) moved to an rr_graph building-related file.

I do not disagree! I am just trying to make the `try_pack` API more obvious. I found the `inter_cluster_delay` variable to be confusing since it was always being computed by that chunk of code. I do not mind moving it to another RRGraph related file; I will try to find one. Do you think that the call to this function occur outside of `try_pack` or inside? I personally think inside makes more sense.

vaughnbetz:
I misread the code -- it is doing all sorts of rr-graph related computations to try to get an estimate of inter-cluster delay so it is pack related. I think it could be simplified as we can't get a good estimate no matter what, and the clusterer is pretty insensitive to its value anyway (it just affects timing analysis a little, but at this early stage we can't get very accurate relative slacks/criticalities anyway).  Merging, and if you can create an issue to track simplifying this code later that would be great.

AlexandreSinger:
> I misread the code -- it is doing all sorts of rr-graph related computations to try to get an estimate of inter-cluster delay so it is pack related. I think it could be simplified as we can't get a good estimate no matter what, and the clusterer is pretty insensitive to its value anyway (it just affects timing analysis a little, but at this early stage we can't get very accurate relative slacks/criticalities anyway). Merging, and if you can create an issue to track simplifying this code later that would be great.

Issue has been raised: Issue #2890

PR #2882, opened 2025-02-06
Add documentation for NightlyTestManual

AmirhosseinPoolad:
This adds documentation on how to run the new NightlyTestManual CI workflow. As there's a chicken and egg problem with the image links in the MD files, I couldn't find a way to preview exactly how it would look like, but the preview of only the MD file (not as part of the docs' HTML files) looked like this:
image


README.developers.md
```diff
@@ -301,6 +301,18 @@ For the very large runs, you can submit your runs on a large cluster. A template
 a Slurm-managed cluster can be found under vtr_flow/tasks/slurm/
 
 ## Continuous integration (CI)
+
+### Manual Nightly Tests
+
+You can use remote servers to run the [vtr_reg_nightly_test1-7](#vtr_reg_nightly_test1-N) tests. These tests are triggered manually by going to the GitHub Actions menu, selecting the NightlyTestManual workflow and selecting run workflow. Once you do that, the [Nightly Test Manual Workflow](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.github/workflows/nightly_test_manual.yml) will be triggered. This run will take approximately 15 hours to complete and will cancel any currently running workflows.
```
AlexandreSinger, line 307: Small update to the last sentence. I think the workflow will only cancel any currently running workflows on the same branch. For example, if someone runs the workflow on Master and another person then runs one on another branch, the run on Master will not be cancelled. I think the other run will just get queued.

AlexandreSinger (commented):
LGTM, one small comment.

README.developers.md
AmirhosseinPoolad: So I double checked this and you're correct it will only cancel runs that are triggered by the same PR (which doesn't happen right now since we're not running this per PR) or same branch, as they are part of the same concurrency group. However I don't see any documentation mentioning that jobs running on different concurrency group are queued. I think it actually just runs the job for multiple branches at the same time, but I might be wrong. If correct, is that intended behavior? I don't think we should run multiple jobs at the same time.

Here's the definition of the concurrency groups for this workflow:
'${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}'
This gets expanded to something like nightly_test_manual.yml-master.

And here's the relevant part in GitHub's documentation:
'When you define a concurrency key, GitHub Actions ensures that only one workflow or job with that key runs at any given time.'
I think this implies that jobs with different keys can run at the same time.

AlexandreSinger: Hi Amir, you are correct; for this workflow, the same workflow will not run at the same time on the same branch; however, it will allow multiple workflows to run at the same time from different branches! But we only have one SAVI self-hosted runner, so multiple runs cannot happen at the same time (as far as I am aware).

The GitHub documentation is assuming you have more than one runner who can pick up the task. I think if all runners are busy, the workflow is queued until a runner opens up.

I think if we wanted to allow multiple runs on the same machine to occur, we would need to set up another (independent) runner on the SAVI machine; however, Vaughn and I discussed this in the VTR meeting today and he is worried about running out of RAM and CPU resources. So lets not do that!

AmirhosseinPoolad: Oh I didn't know that a runner can only run one job at a time, but it makes total sense. Everything is fine then!

AlexandreSinger: No worries, I am pretty sure this is the case from my observations of the Google Runners as they were on their last life. You can verify this by checking the status of the SAVI self-hosted runner during the next run!
Screenshot from 2025-02-06 15-44-27

AlexandreSinger:
I am making use of this feature! Excited to see if this works on the field!
Screen Shot 2025-02-06 at 6 23 52 PM

README.developers.md
```diff
@@ -301,6 +301,18 @@ For the very large runs, you can submit your runs on a large cluster. A template
...
+### Manual Nightly Tests
+
+You can use remote servers to run the [vtr_reg_nightly_test1-7](#vtr_reg_nightly_test1-N) tests. These tests are triggered manually by going to the GitHub Actions menu, selecting the NightlyTestManual workflow and selecting run workflow on the branch you want to test. Once you do that, the [Nightly Test Manual Workflow](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.github/workflows/nightly_test_manual.yml) will be triggered. This run will take approximately 15 hours to complete and will cancel all other workflow runs for the same branch.
+
+<img src="https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/master/doc/src/dev/run_ci_manual/select_actions.png" alt="Select GitHub Actions menu" width="60%"/>
+<br/>
+<img src="https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/master/doc/src/dev/run_ci_manual/select_workflow.png" alt="Select the NightlyTestManual workflow" width="30%"/>
+<img src="https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/master/doc/src/dev/run_ci_manual/run_workflow.png" alt="Run the Workflow" width="30%"/>
+
+### Other CI Tests
```
vaughnbetz, line 314: I think we should change this from Other CI Tests to "Automatic (Github runner) CI tests"

vaughnbetz: I also think we should move this above the "manual nightly tests" as it is the first thing people should see for CI.

README.developers.md
```diff
@@ -301,6 +301,18 @@ For the very large runs, you can submit your runs on a large cluster. A template
 a Slurm-managed cluster can be found under vtr_flow/tasks/slurm/
 
 ## Continuous integration (CI)
+
+### Manual Nightly Tests
+
+You can use remote servers to run the [vtr_reg_nightly_test1-7](#vtr_reg_nightly_test1-N) tests. These tests are triggered manually by going to the GitHub Actions menu, selecting the NightlyTestManual workflow and selecting run workflow on the branch you want to test. Once you do that, the [Nightly Test Manual Workflow](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.github/workflows/nightly_test_manual.yml) will be triggered. This run will take approximately 15 hours to complete and will cancel all other workflow runs for the same branch.
```
vaughnbetz, line 307: The vtr_reg_nightly_test1-N link seems to be broken.

README.developers.md
```diff
@@ -301,6 +301,18 @@ For the very large runs, you can submit your runs on a large cluster. A template
...
+
+<img src="https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/master/doc/src/dev/run_ci_manual/select_actions.png" alt="Select GitHub Actions menu" width="60%"/>
+<br/>
+<img src="https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/master/doc/src/dev/run_ci_manual/select_workflow.png" alt="Select the NightlyTestManual workflow" width="30%"/>
+<img src="https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/master/doc/src/dev/run_ci_manual/run_workflow.png" alt="Run the Workflow" width="30%"/>
+
+### Other CI Tests
+
```
vaughnbetz, line 315: The link a few lines below this (vtr_reg_nightly_test1-N) also seems to be broken.

README.developers.md
```diff
@@ -301,6 +301,18 @@ For the very large runs, you can submit your runs on a large cluster. A template
 a Slurm-managed cluster can be found under vtr_flow/tasks/slurm/
 
 ## Continuous integration (CI)
+
+### Manual Nightly Tests
+
+You can use remote servers to run the [vtr_reg_nightly_test1-7](#vtr_reg_nightly_test1-N) tests. These tests are triggered manually by going to the GitHub Actions menu, selecting the NightlyTestManual workflow and selecting run workflow on the branch you want to test. Once you do that, the [Nightly Test Manual Workflow](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.github/workflows/nightly_test_manual.yml) will be triggered. This run will take approximately 15 hours to complete and will cancel all other workflow runs for the same branch.
+
+<img src="https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/master/doc/src/dev/run_ci_manual/select_actions.png" alt="Select GitHub Actions menu" width="60%"/>
+<br/>
```
vaughnbetz, line 310: These pictures aren't showing up.  Not sure what the problem is; the file paths etc. are persnickety.  You can see the rendered documentation built in the last check in CI (click on the doc link that is built in a CI test in this PR; it is always built twice for some reason, so either test/link works).

vaughnbetz (changes requested):

vaughnbetz:
@AmirhosseinPoolad : one more question about the new CI: did you update the golden results for nightly_test7?  If not, can you update those as well so we're up to date?

README.developers.md
AmirhosseinPoolad: Fixed this in the latest commit

PR #2884, opened 2025-02-06
[Prepack] Pack Molecule Data Structure Cleanup

AlexandreSinger:
Since many people are starting to look into the Packer for different reasons (refactoring / paralellism / etc.) I wanted to clean up the prepacker. The packed molecules being a linked list has been a painful thing to work with; so changing it to a vector and giving each molecule a unique ID really makes it easier to work with the molecules. I also resolved a data dependency in the packer regarding these molecules. The packer was modifying data in the molecules for carry chains. I moved this data out, into the cluster legalizer, which means that the prepacked molecules are never modified after they are created. This should make parallelizing the packer a bit easier.

Moved the t_pack_molecule data structure to the prepacker.

Updated the t_pack_molecule data structure from a linked list to a vector, where each molecule is identified by a unique strong ID.

Updated the part of the t_pack_molecule which stores chain information from a shared pointer to a unique ID. Moved clustering related data out of this and into the cluster legalizer. This made the prepacked molecules 100% immutable after they are constructed which makes them much safer to work with.

Removed the old "num_blocks" member from t_pack_molecule since it is already available from the atom_blk_ids

General code quality cleanups while I was working on this. One header file removed from another header file sent me down a rabbit hole of header files to update. Eventually we need to clean up how we handle header files...

The resolves some of the parts of Issue #2791


AlexandreSinger:
VTR Chain Results vs Master (Baseline):

  | baseline_parse_results.txt | k_mol_cleanup_parse_results.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.98978
odin_synth_time |   |  
parmys_synth_time | 1 | 0.992269
abc_depth | 1 | 1
abc_synth_time | 1 | 1.00504
num_clb | 1 | 1
num_memories | 1 | 1
num_mult | 1 | 1
max_vpr_mem | 1 | 1.000618
num_pre_packed_blocks | 1 | 1
device_grid_tiles | 1 | 1
pack_time | 1 | 0.991515
placed_wirelength_est | 1 | 1
place_time | 1 | 0.960937
placed_CPD_est | 1 | 1
min_chan_width | 1 | 1
routed_wirelength | 1 | 1
min_chan_width_route_time | 1 | 0.963625
critical_path_delay | 1 | 1
crit_path_route_time | 1 | 1.006236

QoR is exactly the same. Runtime is a wash (differences likely due to load). Memory is a very slight increase; I ripped the following data regarding the pack memory:

baseline | cleanup
-- | --
pack_mem | pack_mem
122.9 MiB | 123.7 MiB
299.4 MiB | 299.8 MiB
89.3 MiB | 89.3 MiB
30.9 MiB | 31.1 MiB
26.6 MiB | 26.6 MiB
29.5 MiB | 29.6 MiB
28.3 MiB | 28.2 MiB
51.2 MiB | 52.2 MiB
31.3 MiB | 31.7 MiB
40.8 MiB | 41.1 MiB
52.9 MiB | 52.9 MiB
42.5 MiB | 42.2 MiB
40.3 MiB | 40.4 MiB
33.0 MiB | 33.0 MiB
146.1 MiB | 145.1 MiB
142.9 MiB | 143.0 MiB
289.6 MiB | 292.4 MiB
25.2 MiB | 26.0 MiB
279.7 MiB | 280.1 MiB
884.5 MiB | 887.0 MiB
1056.6 MiB | 1057.3 MiB

Each of these increases are under 1% and are likely just due to adding new data structures to maintain information on the molecules (to separate out the clustering information from the molecules).

I have launched the nightly tests manually. Currently NightlyTest1 completed with the same QoR results as Amir's run yesterday (1 expected QoR failure).

AlexandreSinger:
@vaughnbetz This is ready for review. It looks big, but most of this PR are one-line changes to convert `t_pack_molecule*`s to `PackMoleculeId`s. I do not think any pointers to the pack molecules exist in VTR anymore outside of the prepacker.

The only files that had big changes were prepack.h / .cpp and cluster_legalizer.h / .cpp where I canged some of the API interfaces.

AlexandreSinger:
All nightly tests finished successfully. It had the same exepcted 18 QoR failures (and the values in these failed testcases are the same):
image

vpr/src/pack/prepack.h
```diff
@@ -165,45 +243,84 @@ class Prepacker {
...
+    }
 
 private:
     /**
-     * @brief A linked list of all the packing molecules that are loaded in
-     *        prepacking stage.
+     * Pre-pack atoms in netlist to molecules
+     * 1.  Single atoms are by definition a molecule.
+     * 2.  Forced pack molecules are groupings of atoms that matches a t_pack_pattern definition.
+     * 3.  Chained molecules are molecules that follow a carry-chain style pattern,
```
vaughnbetz, line 280: If you know when such chained molecules are split into smaller molecules at logic block boundaries (I assume they are split during packing?) then that would be good to document here.

vaughnbetz (approved):
Looks good, thanks!  I'll merge this to get it out of the way -- it is a nice clean up!  I have one suggested comment to add (if you know the answer) but that can be done in another PR.

vpr/src/pack/prepack.h
AlexandreSinger: Raised the PR here: PR #2891 

The molecules are not split during clustering. The molecules are never modified after the prepacking stage. The prepacker will create segment chains into molecules which each fit inside of a cluster. It will also maintain IDs so molecules know if they belong to the same chain. The reason I think we do this is because the cluster legalizer only works on molecules and I think it just makes more sense to put the WHOLE molecule into a cluster during legalization than potentially leave a part of it unclustered.

PR #2841, opened 2024-12-08
Synlig, new system-verilog tool integration

amirarjmand93:
#### Types of changes

- [ ] Bug fix (change which fixes an issue)
- [ ] New feature (change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

#### Checklist:

- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

PR #2885, opened 2025-02-07
[CI] Added System Verilog Regression Tests to GitHub Runners

AlexandreSinger:
The system verilog regression tests were put into the currently not-working nightly tests. Since they are so small, putting them in the GitHub runners tests instead so they do not regress.

The strong parmys test also has system verilog tests (I think); however, that test may be too large for the GitHub runners. We may consider moving that test to the SAVI workflows in the future.


AlexandreSinger:
@amirarjmand93 The System Verilog regression tests do not appear to be working. There seems to be a build error with the `-DYOSYS_F4PGA_PLUGINS=ON` CMAKE option:
Screenshot from 2025-02-07 15-12-16

Have you seen this error before and know how to fix it?

amirarjmand93:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2841#issuecomment-2544658083

I ran this test locally and the result was okay. 

have you tried "-DYOSYS_F4PGA_PLUGINS=ON" ? I had changed to  "-DSYNLIG_SYSTEMVERILOG=ON" in [.github/workflows/nightly_test.yml](url)

AlexandreSinger:
> [#2841 (comment)](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2841#issuecomment-2544658083)
> 
> I ran this test locally and the result was okay.
> 
> have you tried "-DYOSYS_F4PGA_PLUGINS=ON" ? I had changed to "-DSYNLIG_SYSTEMVERILOG=ON" in [nightly_testl](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/.github/workflows/nightly_test.yml)

What is the difference between `-DYOSYS_F4PGA_PLUGINS=ON` and `-DSYNLIG_SYSTEMVERILOG=ON`? Currently I am matching what is written in the NightlyTests:
Screenshot from 2025-02-07 16-02-27

But the NightlyTests are _never_ being run; so I am moving it to our GitHub runners. Which of the two arguments should I give to this test? Both?

AlexandreSinger:
> Before Synlig , we had used F4PGA pluggin and `-DYOSYS_F4PGA_PLUGINS=ON` was the command to make vtr with system verilog. in Synlig, i changed `-DYOSYS_F4PGA_PLUGINS=ON` to `-DSYNLIG_SYSTEMVERILOG=ON`. so just this new cmake command should work and make vtr + synlig.

The test I added is using  `-DSYNLIG_SYSTEMVERILOG=ON`, everything else is using the default libraries and build commands for Ubuntu 24.04. Is it possible that you have something locally that is not in this test?

amirarjmand93:
git submodule init
    git submodule update --recursive
    

I think adding this to the YML might solve the problem. I usually have tried them before making and testing a VTR with Synlig.

AlexandreSinger:
> ```
> git submodule init
> git submodule update --recursive
> ```
> 
> I think adding this to the YML might solve the problem. I usually have tried them before making and testing a VTR with Synlig.

The `--recursive` is interesting. We do not do that by default in the CI, we normally just do a regular submodule update. I have pushed a revision which does the submodule update recursively. Hopefully that fixes it!

amirarjmand93:
image

let's add commands under **run**:

 ```
run: |
        git submodule sync --recursive  # Sync submodules
        git submodule update --init --recursive  # Ensure they are initialized & updated'
.
.
.
```

do you think it might help?

AlexandreSinger:
> git submodule update --init --recursive  # Ensure they are initialized & updated'

Lets give it a shot, but why is this necessary?

amirarjmand93:
I think this Building discrepancy might arise from some library dependency, a path conflicting with CI or a race condition between Cloning and Updating modules. but I checked and it seems no new system or Python packages are required for Synlig. (you can also check https://github.com/chipsalliance/synlig) 
Also, I rebuilt on local env and no errors appeared. 

- Can I manipulate this branch and run some tests?

AlexandreSinger:
@amirarjmand93 Feel free to modify the branch anyway you like! I recommend just deleting the other tests so that just the SystemVerilog test is active if you will use the CI to test (just to save yourself time waiting for the other known working tests).

If you think it is a race condition in the build script (which honestly could be the problem!), you can make the build script single threaded by setting the NUM_PROC to 1 here:
image

amirarjmand93:
Thank you, 
Yes usually I disable nightly tests but in this test, I think working tests should be removed from this list, right?
image

AlexandreSinger:
> Thank you, Yes usually I disable nightly tests but in this test, I think working tests should be removed from this list, right? image

No, thats a different list just for the coverity scan test. To turn off the other jobs you should be able to just delete the jobs from the workflow:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/f41ea8ceaafe9aa0089a49402edb948efb88de05/.github/workflows/test.yml#L36-L581

So basically just delete everything under "Build:", "Format:", "UnitTests:", etc. Leaving behind just the "SystemVerilog:" job.

The regular tests work slightly different from the Nightly Tests.

amirarjmand93:
- The previous error during Build time was resolved. 

- Currently, we have error on the Test. `<ERROR: TCL interpreter returned an error: couldn't execute "/home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/build/bin/synlig_install/usr/local/bin/synlig": no such file or directory`
probably Synlig has not been installed!

- How can I get access to working files for this test(working VTR folder)? I tried Action pathway but I couldn't find it.

- Also there is no problem with Building, Test and QOR on the local server for this test(SystemVerilog):
image_2025-02-09_03-30-19

vaughnbetz:
The path looks a bit funny ... verilog-to-routing is in it twice.  Is that the issue (Cmake issue where the path isn't set properly?):

/home/runner/work/vtr-verilog-to-routing/**vtr-verilog-to-routing**/build/bin/synlig_install/usr/local/bin/synlig

amirarjmand93:
I believe there is a problem related to CMake in the Synlig project. 

in CMake in EXTERNAL:

-When I use the explicit build and install command, it can create an execution file on the local PC (works fine) but cannot do it in CI ( I don't know why?). 
`BUILD_COMMAND ${MAKE_PROGRAM} -C ${SYNLIG_SOURCE_DIR} install DESTDIR=${CMAKE_BINARY_DIR}/synlig_install`

-When I don't use any command and let CMake automatically build and install, it can't create executable files for synlig in PC or CI. (due to some issue in synlig's CMake file)

amirarjmand93:
@AlexandreSinger, is there a time limit on the GitHub runner? I receive a termination signal (due to the Synlig compilation workload), and if possible, could you please increase it ?

[100%] Built target test_vpr
-- synlig build command succeeded.  See also /home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/libs/EXTERNAL/synlig/src/synlig-stamp/synlig-build-*.log
[100%] Performing install step for 'synlig'
**make[3]: *** [libs/EXTERNAL/CMakeFiles/synlig.dir/build.make:104: /home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/libs/EXTERNAL/synlig/src/synlig-stamp/synlig-install] Terminated
make[2]: *** [CMakeFiles/Makefile2:1515: libs/EXTERNAL/CMakeFiles/synlig.dir/all] Terminated
make[1]: *** [Makefile:146: all] Terminated
make: *** [Makefile:90: all] Terminated
Error: Process completed with exit code 143.**

AlexandreSinger:
> @AlexandreSinger, is there a time limit on the GitHub runner? sometimes, I receive a termination signal (due to the Synlig compilation workload), and if possible, could you please increase it?
> 
> [100%] Built target test_vpr -- synlig build command succeeded. See also /home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/libs/EXTERNAL/synlig/src/synlig-stamp/synlig-build-*.log [100%] Performing install step for 'synlig' **make[3]: *** [libs/EXTERNAL/CMakeFiles/synlig.dir/build.make:104: /home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/libs/EXTERNAL/synlig/src/synlig-stamp/synlig-install] Terminated make[2]: *** [CMakeFiles/Makefile2:1515: libs/EXTERNAL/CMakeFiles/synlig.dir/all] Terminated make[1]: *** [Makefile:146: all] Terminated make: *** [Makefile:90: all] Terminated Error: Process completed with exit code 143.**

There is not a time limit on these tests as far as I am aware. What you are seeing can happen when you are using too many resources. For example too many threads, too much memory, or too much storage.

I have noticed that turning on System Verilog causes a lot of extra libraries to be downloaded and installed on the machine (which appears to cause the compile time to increase by 3x to 4x). Why are all of these libraries necessary? I am wondering if you are installing too much on the GitHub runners and it is causing issues.

amirarjmand93:
Hello All, 

[Synlig ](https://github.com/chipsalliance/synlig)has not been maintained for 6 months. Also, it has very dependencies(ext libs) and tests that need deep reformation to match our server building and resource usage restrictions. 

We can switch to some alternatives : 

1. [sv2v ](https://github.com/zachjs/sv2v?tab=readme-ov-file) is a tool that converts SystemVerilog (SV) → Verilog-2005.It parses full synthesizable SV, simplifies it, and rewrites it as plain Verilog. it seems sv2v used internally with yosys [sv2v-yosys](https://blog.yosyshq.com/p/community-spotlight-sv2v/) !?
2. [Surelog ](https://github.com/chipsalliance/Surelog?tab=readme-ov-file)  is a SystemVerilog parser and elaborator that converts SV source files into a structured UHDM (Surelog had been used in Synlig). Yosys does not natively understand UHDM, but it can use it through the [yosys-uhdm](https://github.com/alainmarcel/yosys-uhdm-plugin-integration) plugin( UHDM → RTLIL) , which allows SystemVerilog support via UHDM. The plugin has not been maintained for 2 years!
3.  **[Slang ](https://github.com/MikePopoloski/slang) and [ yosys-slang](https://github.com/povik/yosys-slang?utm_source=chatgpt.com)** : Slang is a fast SystemVerilog parser supporting IEEE 1800-2017/2023. The Yosys-Slang plugin integrates it into Yosys via the `read_slang `command for direct SystemVerilog loading. [Yosys read slang officialy](https://yosyshq.readthedocs.io/projects/yosys/en/docs-preview-bugpoint/using_yosys/more_scripting/load_design.html#externally-maintained-plugins)
4. [Siliconcompiler ](https://github.com/siliconcompiler/siliconcompiler/tree/main): in SiliconCompiler:

- sv2v is used as a conversion step to make SystemVerilog designs compatible with Verilog-only tools.
- Slang and SureLog are used as full-featured frontends for parsing and elaborating SystemVerilog in flows that require advanced language support.


I would appreciate it if @vaughnbetz, @hzeller and @AlexandreSinger could take a look and guide us to start new SV integration. I think number 3, Yosys-Slang,  is more rational due to its integration in Yosys.
 

> Also yosys-slang can parse a number of open-source IPs, including:
> 
> [Black Parrot](https://github.com/black-parrot/black-parrot/)
> 
> [BSC Core Tile](https://github.com/bsc-loca/core_tile/)
> 
> [OpenHW Group's CV32E40P](https://github.com/openhwgroup/cv32e40p)
> 
> [Ibex RISC-V Core](https://github.com/lowRISC/ibex)
> 
> Supporters
> The following organizations have supported the project and contributed to yosys-slang development:
> 
> [Microelectronics Design Center of ETH Zürich](https://dz.ethz.ch/)
> 
> [PULP platform](https://pulp-platform.org/)
> 
> [Silimate](https://www.silimate.com/)
> 
> [YosysHQ](https://www.yosyshq.com/)
> 
> [OpenTitan: Open source silicon root of trust](https://github.com/lowRISC/opentitan)
> 
> [RSD RISC-V Out-of-Order Superscalar Processor](https://github.com/rsd-devel/rsd/)
>

AlexandreSinger:
@amirarjmand93 Thank you for the summary of alternatives! I am sure many people in the VTR community will have opinions about this choice! I recommend that we discuss this at the next VTR Industry Sync (Thursday Meeting).

petergrossmann21:
Some additional feedback:

* Within Silicon Compiler, slang is becoming preferred over Surelog as slang appears to be in a better place with respect to its maintenance.  
* Silicon Compiler is a full-on flow manager, so adopting it just for its front end support doesn't make a lot of sense to me.  Adopting it as an alternative to bespoke Python scripting for managing end-to-end execution of the VTR flow makes more sense if that's something people are interested in.  Within that paradigm, solving the yosys/slang integration would be free(ish) since there are pre-defined flows for using them in concert.
* The lowest-overhead solution is probably yosys + slang if the read_slang command support is well-maintained, and sv2v if it isn't.  sv2v is expressly designed to produce Verilog as its output, so as long as it a) has good enough SystemVerilog syntax coverage and b) produces Verilog yosys can ingest without modification.  At that point it's just one more step to execute at the beginning of the flow before running Yosys.  I've tried this on a couple of non-VTR benchmarks in the past with success.
* Due to high complexity, I would not recommend putting effort into a UHDM to RTLIL interface unless there is an interesting research problem there.

vaughnbetz:
@loglav03 : I think you own this one now, and yosys-slang is the chosen technology.

Issue #2886, opened 2025-02-07
Doxygen not building the VPR Grid doc properly


vaughnbetz:
https://docs.verilogtorouting.org/en/latest/api/vpr/grid/ currently gives an error:
doxygenclass: Cannot find class “DeviceGrid” in doxygen xml output for project “vpr” from directory: ../_build/doxygen/vpr/xml

I think either this class was moved or some part of its doxygen comment has become corrupted.  @soheilshahrouz : did you make any changes to this file? Can you take a look?

soheilshahrouz:
Fixed in #2887

PR #2887, opened 2025-02-07
Fix DeviceGrid doxygen

soheilshahrouz:
Fixes [this isuue](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2886).


vaughnbetz:
Thanks!

PR #2891, opened 2025-02-08
[Prepack] Improved Documentation on Molecule Chains

AlexandreSinger:
Improved the documentation on how chains are handled in the prepacker. The prepacker segments chains into multiple molecules (where each molecule can fit into a single cluster). The prepacker maintains MoleculeChainIds to the chain these molecules are a part of. This allows the cluster legalizer to easily put the molecules into clusters and understand when a molecule is part of a special chain (for gain calculation).


vpr/src/pack/prepack.h
```diff
@@ -56,9 +56,12 @@ enum class e_pack_pattern_molecule_type : bool {
  * @brief Represents a grouping of atom blocks that match a pack_pattern,
  *        these groups are intended to be placed as a single unit during packing
  *
- * A chain is a special type of pack pattern.  A chain can extend across multiple logic blocks.
- * Must segment the chain to fit in a logic block by identifying the actual atom that forms the root of the new chain.
- * Assumes that the root of a chain is the primitive that starts the chain or is driven from outside the logic block
+ * A chain is a special type of pack pattern since it can extend across multiple
+ * logic blocks. The prepacker segments the chain into molecules that fit in a
+ * logic block by identifying the actual atom that forms the root of the new
+ * chain. This assumes that the root of a chain is the primitive that starts the
```
vaughnbetz, line 62: <Change to>
A chain is a special type of pack pattern sice it can extend across multiple logic blocks. The prepacker segments the chain into molecules that each fit in a logic block by identifying the atom that forms the root of the chain, and starting the first molecule from it. Long chains can lead to multiple molecules; a new molecule is created as the chain is traversed every time we exceed the maximum number of bits a single logic block can implement.  The MoleculeChainId can be used to identify the molecules that are part of the same chain; it is set to Invalid if a given molecule did not come from a chain.

PR #2892, opened 2025-02-10
[AP] Added APPack to the AP Flow as a Full Legalizer

AlexandreSinger:
APPack is a Full Legalizer which modifies the Packer and Placer to try and improve their solutions based on a given Flat Placement. Integrated this into the AP Flow by converting the Partial Placement generated by the Global Placer into a Flat Placement and calling the Packer and Placer.

Kept the original, Naive Full Legalizer as something to compare to and since it may be useful to keep this legalizer around.

Added tests for both the Naive and APPack Full Legalizers, leaving the other tests to always run on the default AP Flow.


vpr/src/analytical_place/ap_flow_enums.h
```diff
@@ -0,0 +1,20 @@
...
+#pragma once
+
+/**
+ * @brief The type of a Full Legalizer.
+ *
+ * The Analytical Placement flow may implement different Full Legalizers. This
+ * enum can select between these different Full Legalizers.
+ */
+enum class e_ap_full_legalizer {
+    Naive,  ///< The Naive Full Legalizer, which builds and places cluster in the first possible place.
```
vaughnbetz, line 17: Expanding on this would be useful ... what does 'first possible place' mean?  Does it pay attention to a flat placement at all?

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -449,3 +486,65 @@ void FullLegalizer::legalize(const PartialPlacement& p_placement) {
...
+              vpr_setup_.PlacerOpts,
+              vpr_setup_.RouterOpts,
+              vpr_setup_.AnalysisOpts,
+              vpr_setup_.NocOpts,
+              arch_.Chans,
+              &vpr_setup_.RoutingArch,
+              vpr_setup_.Segments,
+              arch_.directs,
+              flat_placement_info,
+              false /* is_flat */);
```
vaughnbetz, line 543: Any reason why is_flat is always false?  If you're going to move this around as part of the initial placement / detailed placement separation you can ignore this comment, but eventually we should make sure we work with flat routing (which will just change the rr-graph for the placer delay lookup/matrix calculation I think).

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -449,3 +486,65 @@ void FullLegalizer::legalize(const PartialPlacement& p_placement) {
...
+
+    // TODO: This was taken from vpr_api. Not sure why it is needed. Should be
+    //       made part of the placement and verify placement should check for
+    //       it.
+    post_place_sync();
```
vaughnbetz, line 548: I think this makes sure the pin locations on blocks match the pin locations on the tile where they were placed (for equivalent_sites).  I agree it probably makes more sense to put it at the end of placement.

vaughnbetz (commented):
A few suggestions.
Also, you should add the new command line options to the VPR command line documentation (.rst file)

vpr/src/analytical_place/full_legalizer.cpp
AlexandreSinger: Its strange. The vpr_api.cpp file forces it to be false explicitly. Also in the try_place function it actually asserts that it is false:
Screenshot from 2025-02-11 14-04-19

I am not sure why it does this...

vpr/src/analytical_place/full_legalizer.cpp
AlexandreSinger: I looked into how this is being used in the vpr_api and it is used after both try_place and reading a placement from a file. Its not super simple to remove, so I will leave this as a TODO.

vaughnbetz (approved):
The docs look good, thanks.

PR #2894, opened 2025-02-11
[Pack] Removed Global Overfilled Partition Regions

AlexandreSinger:
After each pack, the partition regions are checked to see if any of the partition regions are overfilled. During this process, the overfilled partition regions were stored in the Floorplanning context (likely for legacy reasons). This was then read from a few lines later in another function.

Since this is only used in such a localized scenario, turned this into a local variable instead of a global variable.

### DIFF
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -6,18 +6,17 @@
 #include <mutex>
 
 #include "FlatPlacementInfo.h"
+#include "user_place_constraints.h"
+#include "user_route_constraints.h"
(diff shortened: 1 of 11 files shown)


AlexandreSinger:
@vaughnbetz This is a small PR where I cleaned up a global variable in the FloorPlanning Context which should not have been there. This was just something I noticed while I was separating out the Initial Placer from the Placer class. Please review when you have a moment!

vaughnbetz (approved):
LGTM, thanks.

PR #2896, opened 2025-02-13
[Place] Moved PlaceMacros Out of BlkLocRegistry

AlexandreSinger:
The PlaceMacros only require the Clustering and atom/arch info to be constructed; after which they are never modified. Since they are unchanged by the placement, they were out of place in the BlkLocRegistry.

Moved them to the placement context.

This will make it easier to separate out the initial placement from the SA placer.

### DIFF
--- a/vpr/src/base/vpr_api.cpp
+++ b/vpr/src/base/vpr_api.cpp
@@ -14,9 +14,11 @@
 #include <cstdio>
 #include <cstring>
 #include <cmath>
+#include <memory>
 
 #include "FlatPlacementInfo.h"
 #include "cluster_util.h"
+#include "place_macro.h"
 #include "verify_placement.h"
 #include "vpr_context.h"
 #include "vtr_assert.h"
@@ -392,7 +394,8 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
 
     { //Place
         const auto& placement_net_list = (const Netlist<>&)g_vpr_ctx.clustering().clb_nlist;
-        bool place_success = vpr_place_flow(placement_net_list, vpr_setup, arch);
+        const PlaceMacros& place_macros = *g_vpr_ctx.clustering().place_macros;
+        bool place_success = vpr_place_flow(placement_net_list, vpr_setup, arch, place_macros);
 
         if (!place_success) {
             return false; //Unimplementable
@@ -417,9 +420,10 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
 
     bool is_flat = vpr_setup.RouterOpts.flat_routing;
     const Netlist<>& router_net_list = is_flat ? (const Netlist<>&)g_vpr_ctx.atom().nlist : (const Netlist<>&)g_vpr_ctx.clustering().clb_nlist;
+    const PlaceMacros& place_macros = *g_vpr_ctx.clustering().place_macros;
     RouteStatus route_status;
     { //Route
-        route_status = vpr_route_flow(router_net_list, vpr_setup, arch, is_flat);
+        route_status = vpr_route_flow(router_net_list, vpr_setup, arch, place_macros, is_flat);
     }
     { //Analysis
         vpr_analysis_flow(router_net_list, vpr_setup, arch, route_status, is_flat);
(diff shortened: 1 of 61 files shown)


vpr/src/base/vpr_api.cpp
```diff
@@ -392,7 +394,8 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
 
     { //Place
         const auto& placement_net_list = (const Netlist<>&)g_vpr_ctx.clustering().clb_nlist;
-        bool place_success = vpr_place_flow(placement_net_list, vpr_setup, arch);
+        const PlaceMacros& place_macros = *g_vpr_ctx.clustering().place_macros;
```
vaughnbetz, line 397: Not sure it makes sense to make the place_macros part of the clustering context. 
We could have place_macros in the future that come from user constraints (this is one reason we don't directly look at carry chains etc. in the placer).  Basically place_macros are used by the placer, even though our currently creation of them only uses data from the clustered netlist and architecture.

Is there a better owner?  Placement context?

vaughnbetz:
I'm OK with passing around place_macros more if you want, but not sure why this would be a member of the clustering context.  That seems like a step backward.

vpr/src/base/vpr_api.cpp
AlexandreSinger: I basically flipped a coin and settled on Clustering context. Original I thought of putting it in the Placement context and having the placement context create and maintain it; however, I then realized that the Placement Macros only depend on the clustering and never change (at least currently). Since they only made sense once the clusters were created, I thought it made sense to put them in the clustering context.

Originally I wanted to take them out of the global context all together; but then found out that they are used in the drawer and Router (for min W search); so they ended up going all the way to the vtr_flow which was strange.

I think you are correct though that the macros only make sense in the context of placement (especially given their name PlaceMacos). I will move them to the placement context and put their initializer code somewhere that makes more sense (in place.cpp).

vaughnbetz:
Sounds good, thanks.

AlexandreSinger:
I have updated the location of the PlaceMacros to the placement context. @soheilshahrouz do you have any comments?

vpr/src/place/annealer.cpp
```diff
@@ -388,16 +391,16 @@ e_move_result PlacementAnnealer::try_swap_(MoveGenerator& move_generator,
...
 #endif //NO_GRAPHICS
     } else if (router_block_move) {
         // generate a move where two random router blocks are swapped
-        create_move_outcome = propose_router_swap(blocks_affected_, rlim, blk_loc_registry, rng_);
+        create_move_outcome = propose_router_swap(blocks_affected_, rlim, blk_loc_registry, place_macros_, rng_);
         proposed_action.move_type = e_move_type::UNIFORM;
     } else {
         //Generate a new move (perturbation) used to explore the space of possible placements
-        create_move_outcome = move_generator.propose_move(blocks_affected_, proposed_action, rlim, placer_opts_, criticalities_);
+        create_move_outcome = move_generator.propose_move(blocks_affected_, proposed_action, rlim, place_macros_, placer_opts_, criticalities_);
```
soheilshahrouz, line 403: I suggest that the MoveGenerator class hold a const reference to the placement macros to avoid repeatedly passing an unchanging argument to propose_move().

vpr/src/place/place_macro.h
```diff
@@ -247,7 +260,8 @@ class PlaceMacros {
      * chain connection.
      * @param directs Contains information about all direct connections in the architecture.
      */
-    void alloc_and_load_idirect_from_blk_pin_(const std::vector<t_direct_inf>& directs);
+    void alloc_and_load_idirect_from_blk_pin_(const std::vector<t_direct_inf>& directs,
+                                              const std::vector<t_physical_tile_type>& physical_tile_types);
```
soheilshahrouz, line 264: Update the doxygen comment.

vpr/src/place/verify_placement.h
```diff
@@ -61,6 +62,7 @@ class VprContext;
  *          log messages for each error found.
  */
 unsigned verify_placement(const BlkLocRegistry& blk_loc_registry,
+                          const PlaceMacros& place_macros,
```
soheilshahrouz, line 65: Update the doxygen comment.

AlexandreSinger:
Thanks for the review @soheilshahrouz ! @vaughnbetz Do you have any further comments?

vaughnbetz:
LGTM, thanks.

PR #2876, opened 2025-02-01
[APPack] Added Max Displacement Metric

AlexandreSinger:
Added the ability to print the max displacement from the placement reconstruction to the original flat placement.

This metric will help show another dimension to the flat placement reconstruction. Generally, if the average displacement is low, that may not be good if that means that one atom is placed way farther away than where it would like. This can really hurt quality.

I also turned the distance calculation from the Euclidean distance to the Manhattan distance. This should better represent the objective in terms of FPGAs (where the fastest path between two points is not necessarily a straight line).

I reran the results of the flat placement reconstruction on MCNC and VTR-Chain (same testing procedure as the last PR).

MCNC:
iter_num | pack_time(s) | device_util | total_wl | CPD(ns) | percent_cluster_errors | percent_atoms_displaced | average_atom_displacement | max_atom_displacement
-- | -- | -- | -- | -- | -- | -- | -- | --
0 | 1.66 | 0.55 | 15915.75 | 5.74 | -1.00 | -1.00 | -1.00 | -1.00
1 | 1.38 | 0.55 | 15845.75 | 5.72 | 0.06 | 0.11 | 1.03 | 24.90
2 | 1.36 | 0.55 | 15760.75 | 5.78 | 0.02 | 0.05 | 0.62 | 23.35
3 | 1.35 | 0.55 | 15741.55 | 5.75 | 0.01 | 0.03 | 0.50 | 24.20
4 | 1.36 | 0.55 | 15731.45 | 5.77 | 0.00 | 0.03 | 0.47 | 22.10
5 | 1.37 | 0.55 | 15744.25 | 5.75 | 0.00 | 0.03 | 0.47 | 22.60
6 | 1.36 | 0.55 | 15738.00 | 5.70 | 0.00 | 0.03 | 0.46 | 23.70
7 | 1.36 | 0.55 | 15718.60 | 5.75 | 0.00 | 0.03 | 0.47 | 22.40
8 | 1.37 | 0.55 | 15771.60 | 5.72 | 0.00 | 0.03 | 0.44 | 22.05
9 | 1.36 | 0.55 | 15771.50 | 5.72 | 0.00 | 0.03 | 0.46 | 22.60
10 | 1.36 | 0.55 | 15737.35 | 5.73 | 0.00 | 0.03 | 0.46 | 20.75

VTR-Chain:
[...]

PR #2870, opened 2025-01-23
[APPack] Added Flat Placement Reconstruction

AlexandreSinger:
Added logic to the packer to reconstruct clusters from the flat placement, if one is provided. The packer will fall-back on the original packing algorithm if it runs out of options for candidates to cluster.

Added logic to the initial placer to reconstruct the placement of each cluster from the flat placement, if one is provided. It will try to place the cluster at the centroid location of all atoms in a cluster, if it cannot place the cluster there it will fall-back on the original initial placement algorithm.

Added a simple testcase to ensure the reading and writing of the flat placement file is working.

This resolves one of the problems discussed in issue #2868 

This has been tested on all of the MCNC and VTR_chain benchmarks. The results below were collected by running default VTR (iteration 0), which generated a flat placement file. This flat placement file was passed back into VTR to reconstruct the clustering and initial placement (the SA annealer is then run), and another flat placement file is generated. This process is then iteratively repeated. This was run on fixed-size architectures (large enough so the limiting device utilization was not too high) with fixed channel widths equal to 1.3 times the minimum channel width.

MCNC (average over all benchmarks):

iter_num | pack_time(s) | device_util | total_wl | CPD(ns) | percent_cluster_errors | percent_atoms_displaced | average_atom_displacement
-- | -- | -- | -- | -- | -- | -- | --
0 | 1.7125 | 0.551 | 15915.75 | 5.7354425 | -1 | -1 | -1
1 | 1.444 | 0.5525 | 15845.75 | 5.717265 | 0.0556921 | 0.11208215 | 0.8230784
2 | 1.4045 | 0.551 | 15760.75 | 5.781261 | 0.0169662 | 0.0540315 | 0.497987
3 | 1.427 | 0.5505 | 15741.55 | 5.7544645 | 0.00690835 | 0.0347828 | 0.4006579
4 | 1.426 | 0.5505 | 15731.45 | 5.77473 | 0.00038335 | 0.03156595 | 0.3836701
5 | 1.3945 | 0.5505 | 15744.25 | 5.7533945 | 0 | 0.0308358 | 0.3788053
6 | 1.509 | 0.5505 | 15738 | 5.704249 | 0 | 0.03185395 | 0.376404
7 | 1.413 | 0.5505 | 15718.6 | 5.748659 | 0 | 0.03113375 | 0.37807795
8 | 1.397 | 0.5505 | 15771.6 | 5.721962 | 0 | 0.0296978 | 0.35781635
[...]

PR #2897, opened 2025-02-14
Changing subtile selection in the try_centroid_placement of initial_placement

haydar-c:
Summary: changed the subtile selection in the initial_placement to prevent falling into random macro placement while we have available subtiles in the selected location.
#### Description

There was discrepancy in the MCNC benchmark results while reconstructing from flat placement. The initial results can be seen in the "[APPack] Added Max Displacement Metric #2876". The discrepancy was that even though we were able to reconstruct the clusters with zero error, there were displacements in atoms. 

I found that the displacement occurs when we fall into try_place_macro_randomly even though we have available subtiles in the desired location. This seems to be caused by trying to place the block into a randomly selected compatible subtile and if it is occupied falling into try_place_macro_randomly. My change suggests to select the subtile in the desired location from the unused compatible subtitles, falling back into try_place_macro_randomly if we do not have unused compatible subtile in the desired location.

An example of the results indicating discrepancy also given for the ALU4 circuit in the tables below before and after the change.

Before change:

| iter_num | pack_time(s) | device_util | total_wl | CPD(ns) | percent_cluster_errors | percent_atoms_displaced | average_atom_displacement | max_atom_displacement |
|----------|--------------|-------------|----------|---------|------------------------|-------------------------|--------------------------|-----------------------|
| 0        | 0.70         | 0.67        | 6398     | 4.92578 | -1                     | -1                      | -1                       | -1                    |
| 1        | 0.56         | 0.68        | 6413     | 5.06346 | 0.227723               | 0.261242                | 1.406852                 | 12.000000             |
| 2        | 0.62         | 0.68        | 6276     | 5.0679  | 0.039604               | 0.061028                | 0.265525                 | 10.000000             |
| 3        | 0.58         | 0.68        | 6256     | 5.00301 | 0.000000               | 0.004283                | 0.033191                 | 11.000000             |
| 4        | 0.57         | 0.68        | 6284     | 5.12653 | 0.000000               | 0.001071                | 0.002141                 | 2.000000              |
| 5        | 0.59         | 0.68        | 6277     | 5.0527  | 0.000000               | 0.002141                | 0.020343                 | 10.000000             |
| 6        | 0.57         | 0.68        | 6345     | 5.00962 | 0.000000               | 0.002141                | 0.010707                 | 7.000000              |
| 7        | 0.56         | 0.68        | 6318     | 4.96222 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 8        | 0.55         | 0.68        | 6245     | 5.06431 | 0.000000               | 0.001071                | 0.011777                 | 11.000000             |
| 9        | 0.55         | 0.68        | 6311     | 4.95586 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 10       | 0.58         | 0.68        | 6339     | 4.96665 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |

After change:

| iter_num | pack_time(s) | device_util | total_wl | CPD(ns) | percent_cluster_errors | percent_atoms_displaced | average_atom_displacement | max_atom_displacement |
|----------|--------------|-------------|----------|---------|------------------------|-------------------------|--------------------------|-----------------------|
| 0        | 0.64         | 0.67        | 6271     | 4.93428 | -1                     | -1                      | -1                       | -1                    |
| 1        | 0.57         | 0.68        | 6362     | 4.96405 | 0.227723               | 0.225910                | 1.207709                 | 13.000000             |
| 2        | 0.57         | 0.68        | 6371     | 5.08332 | 0.039604               | 0.130621                | 0.509636                 | 11.000000             |
| 3        | 0.54         | 0.68        | 6491     | 4.98076 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 4        | 0.54         | 0.68        | 6446     | 4.98408 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 5        | 0.54         | 0.68        | 6391     | 4.97744 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 6        | 0.59         | 0.68        | 6357     | 4.986   | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 7        | 0.55         | 0.68        | 6344     | 5.04787 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 8        | 0.57         | 0.68        | 6282     | 5.04263 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 9        | 0.55         | 0.68        | 6287     | 5.05193 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              |
| 10       | 0.54         | 0.68        | 6317     | 4.99501 | 0.000000               | 0.000000                | 0.000000                 | 0.000000              | 

#### Related Issue

This is not an issue but related to the results mentioned in "[APPack] Added Flat Placement Reconstruction #2870".

#### Motivation and Context

To investigate the discrepancy between cluster reconstruction error and atom displacement, especially in the MCNC benchmarks.

Basic and strong tests.
#### Types of changes

#### Checklist:


AlexandreSinger (changes requested):
@haydar-c Thank you so much for looking into this! This is clearly a problem with the original centroid placement code where it may suggest a sub_tile at a tile-site which is currently taken!

However, I am not sure if this is the root of our problem. The code that you modified should only be tripped if the location given did not have a sub-tile; but the flat placement does, in fact, provide a sub-tile! I think the real problem here is the `neighbor_legal_loc` is always set to false in our case (when reconstructing). It really should only be set to false if the sub-tile we produce is legal and the sub-tile is taken. I would suggest adding an else statement at line 585 which checks if there is a block at the location, and setting `neighbor_legal_loc` to true if not.

The reason to do this is that it appears as though our sub-tile suggestion from flat placement is always being ignored, and just being set to random. We just happen to be lucky and get approximately the one we wanted! Your change is still super important and I recommend keeping it!

vpr/src/place/initial_placement.cpp
```diff
@@ -597,7 +597,23 @@ static bool try_centroid_placement(const t_pl_macro& pl_macro,
         const auto& compressed_block_grid = g_vpr_ctx.placement().compressed_block_grids[block_type->index];
         const auto& type = device_ctx.grid.get_physical_type({centroid_loc.x, centroid_loc.y, centroid_loc.layer});
         const auto& compatible_sub_tiles = compressed_block_grid.compatible_sub_tile_num(type->index);
-        centroid_loc.sub_tile = compatible_sub_tiles[rng.irand((int)compatible_sub_tiles.size() - 1)];
+
```
vaughnbetz, line 600: It seems like this code should be hoisted earlier; maybe as a helper function called in both find_centroid_loc and find_centroid_neighbor.  It seems strange that we find a compatible subtile only if we didn't have to do a search for a neighbour; I think we should find a compatible and empty subtile whenever we return a centroid location, or else we should return an invalid one.  Or, if we can't do that because some users of the find_centroid_loc don't care if the location is occupied or not we should add a call to a helper function to figure out what the unoccupied sub_tile would be right after calling the find_centroid* routines.

vpr/src/place/initial_placement.cpp
```diff
@@ -339,6 +359,37 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
 
+void find_subtile_in_location(t_pl_loc& centroid, 
+                             t_logical_block_type_ptr block_type,
+                             const BlkLocRegistry& blk_loc_registry,
+                             const PartitionRegion& pr,
+                             vtr::RngContainer& rng) {
+    //check if the location is on chip and legal, if yes try to update subtile
+    if (is_loc_on_chip({centroid.x, centroid.y, centroid.layer}) &&  is_loc_legal(centroid, pr, block_type)) {
+        //finding the subtile location
+        auto& device_ctx = g_vpr_ctx.device();
```
AlexandreSinger, line 370: This should also be const. It is technically fine right now since the type of device_ctx is inferred by the compiler as being a const, but its good to be explicity.

vpr/src/place/initial_placement.cpp
```diff
@@ -339,6 +359,37 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
+        const auto& compressed_block_grid = g_vpr_ctx.placement().compressed_block_grids[block_type->index];
+        const auto& type = device_ctx.grid.get_physical_type({centroid.x, centroid.y, centroid.layer});
+        const auto& compatible_sub_tiles = compressed_block_grid.compatible_sub_tile_num(type->index);
+        
+        //filter out occupied subtiles
+        const GridBlock& grid_blocks = blk_loc_registry.grid_blocks();
+        std::vector<int> available_sub_tiles;
```
AlexandreSinger, line 377: Small optimization: We know that the size of `available_sub_tiles` is bounded by the size of `compatible_sub_tiles`. I would recommend reserving space for available_sub_tiles so the push_back becomes practically free!

Something like:
```cpp
available_sub_tiles.reserve(compatible_sub_tiles.size());
```

vpr/src/place/initial_placement.cpp
```diff
@@ -339,6 +359,37 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
+        for (int sub_tile : compatible_sub_tiles) {
+            t_pl_loc pos = {centroid.x, centroid.y, sub_tile, centroid.layer};
+            if (!grid_blocks.block_at_location(pos)) {
+                available_sub_tiles.push_back(sub_tile);
+            }
+        }
+
+        //if there is at least one available subtile, update the centroid and do not change otherwise
```
AlexandreSinger, line 385: You may want to do more than just leaving it alone. Since the location is on the chip and is legal, it wont do the neighbor exploration! This will fall directly into random selection. Is it possible to make this method return a boolean if a sub-tile was found or not? That way we can bubble this information up?

vpr/src/place/initial_placement.cpp
```diff
@@ -565,7 +621,7 @@ static bool try_centroid_placement(const t_pl_macro& pl_macro,
         //       location near the flat placement centroid.
         if (!is_loc_on_chip({centroid_loc.x, centroid_loc.y, centroid_loc.layer}) ||
             !is_loc_legal(centroid_loc, pr, block_type)) {
-            unplaced_blocks_to_update_their_score = find_centroid_loc(pl_macro, centroid_loc, blk_loc_registry);
+            unplaced_blocks_to_update_their_score = find_centroid_loc(pl_macro, centroid_loc, blk_loc_registry, block_type, pr, rng);
         }
     }
```
AlexandreSinger, line 627: For example, down below, we want `find_centroid_neighbor` to be called if the loc is on the chip, and it is legal, but there are no available sub-tiles to place it. This can happen if many blocks want to be placed in the same tile, but there are plenty of tiles around it.

vpr/src/place/initial_placement.cpp
```diff
@@ -590,15 +646,6 @@ static bool try_centroid_placement(const t_pl_macro& pl_macro,
...
 
     auto& device_ctx = g_vpr_ctx.device();
-    //choose the location's subtile if the centroid location is legal.
-    //if the location is found within the "find_centroid_neighbor", it already has a subtile
-    //we don't need to find one again
-    if (!neighbor_legal_loc) {
```
AlexandreSinger, line 596: Minor nitpick: The `neighbor_legal_loc` variable is now in a weird scope above. Could you move it into the if statement instead of declaring the variable outside.

AlexandreSinger (changes requested):
@haydar-c This change looks great! Just some small comments. Also there seems to be some tests failing.

AlexandreSinger (approved):
This LGTM, can you please update the golden results for the one failing test whenever you get the chance! Thank you for looking into this!

AlexandreSinger:
Will merge after CI passes.

haydar-c:
It seems like I need to revise the last test again.

vpr/src/place/initial_placement.cpp
```diff
@@ -147,6 +147,26 @@ static bool is_loc_legal(const t_pl_loc& loc,
                          const PartitionRegion& pr,
                          t_logical_block_type_ptr block_type);
 
+/**
+ * @brief  Helper function to choose a subtile in specified location if compatible and available one exits.
```
vaughnbetz, line 151: the type is compatible and an available one exists.

vpr/src/place/initial_placement.cpp
```diff
@@ -147,6 +147,26 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
+ *  
+ *   @param centroid The centroid location at which the subtile will be selected using its x,y, and layer.
+ *   @param block_type Logical block type of the macro head member.
```
vaughnbetz, line 154: Logical block type we would like to place here

vpr/src/place/initial_placement.cpp
```diff
@@ -147,6 +147,26 @@ static bool is_loc_legal(const t_pl_loc& loc,
                          const PartitionRegion& pr,
                          t_logical_block_type_ptr block_type);
 
+/**
+ * @brief  Helper function to choose a subtile in specified location if compatible and available one exits.
+ *  
+ *   @param centroid The centroid location at which the subtile will be selected using its x,y, and layer.
+ *   @param block_type Logical block type of the macro head member.  
+ *   @param block_loc_registry Placement block location information. To be filled with the location
+ *   where pl_macro is placed.
```
vaughnbetz, line 156: I think you should remove the "To be filled with the location where pl_macro is placed."
I'd just say:
"Information on where other blocks have been placed."

vpr/src/place/initial_placement.cpp
```diff
@@ -147,6 +147,26 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
+ *   @param pr The PartitionRegion of the macro head member - represents its floorplanning constraints, is the size of
```
vaughnbetz, line 157: I'd just say:
The PartitionRegion of the block we are trying to place - represents its floorplanning constraints; it is the size of the whole chip if the block is not constrained.

vpr/src/place/initial_placement.cpp
```diff
@@ -147,6 +147,26 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
+ *   the whole chip if the macro is not constrained.
+ *   @param rng A random number generator to select subtile from available and compatible ones.
```
vaughnbetz, line 159: (grammatical nit): a subtile from the available and compatible ones

vpr/src/place/initial_placement.cpp
```diff
@@ -147,6 +147,26 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
+ *
+ * @return False if location on chip, legal, but no available subtile found. True otherwise. False leads us to
```
vaughnbetz, line 161: if the location is on the chip and legal but no available subtile is found at that location. 

I'd delete the "False leads us to neighbour placement currently"

vpr/src/place/initial_placement.cpp
```diff
@@ -340,6 +360,42 @@ static bool is_loc_legal(const t_pl_loc& loc,
...
+        //If there is at least one available subtile, update the centroid. Otherwise, sincel location
+        //is legal and on chip but no subtile found, return false for trying neighbour placement.
+        if (!available_sub_tiles.empty()) {
+            centroid.sub_tile = available_sub_tiles[rng.irand((int)available_sub_tiles.size() - 1)];
+        } else {
+            return false;
+        }
+    }
+
+    return true;
```
vaughnbetz, line 396: A bit strange -- we return true if we find a subtile, or if we don't even try.  Could we return true if we found a legal subtile and false otherwise (and change the calling code).  It's a more clear interface.

vpr/src/place/initial_placement.cpp
```diff
@@ -577,9 +641,8 @@ static bool try_centroid_placement(const t_pl_macro& pl_macro,
 
     //centroid suggestion was either occupied or does not match block type
     //try to find a near location that meet these requirements
-    bool neighbor_legal_loc = false;
-    if (!is_loc_legal(centroid_loc, pr, block_type)) {
-        neighbor_legal_loc = find_centroid_neighbor(centroid_loc, block_type, false, blk_loc_registry, rng);
+    if (!is_loc_legal(centroid_loc, pr, block_type) || try_neighbour_due_to_subtile) {
```
vaughnbetz, line 644: I think you can simplify this to if (!found_legal_subtile)    // pick whatever variable name makes sense
if you change the find_subtile_in_location routine to return false if it fails to find a valid subtile at the given location for *any* reason.

It makes that routine more clear, and the logic here more straightforward.

vaughnbetz (changes requested):
Some simplifications proposed.

AlexandreSinger:
I have launched the Nigthly Tests for this branch:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/13467086924

vpr/src/place/initial_placement.cpp
```diff
@@ -551,10 +603,15 @@ static bool try_centroid_placement(const t_pl_macro& pl_macro,
     t_pl_loc centroid_loc(OPEN, OPEN, OPEN, OPEN);
     std::vector<ClusterBlockId> unplaced_blocks_to_update_their_score;
 
+    bool found_legal_subtile = false;
+
     if (!flat_placement_info.valid) {
         // If a flat placement is not provided, use the centroid of connected
         // blocks which have already been placed.
         unplaced_blocks_to_update_their_score = find_centroid_loc(pl_macro, centroid_loc, blk_loc_registry);
+        if(find_subtile_in_location(centroid_loc, block_type, blk_loc_registry, pr, rng)) {
```
vaughnbetz, line 612: Simplify to found_legal_subtile = (function return 
￼

vpr/src/place/initial_placement.cpp
```diff
@@ -567,6 +624,9 @@ static bool try_centroid_placement(const t_pl_macro& pl_macro,
         if (!is_loc_on_chip({centroid_loc.x, centroid_loc.y, centroid_loc.layer}) ||
             !is_loc_legal(centroid_loc, pr, block_type)) {
             unplaced_blocks_to_update_their_score = find_centroid_loc(pl_macro, centroid_loc, blk_loc_registry);
+            if(find_subtile_in_location(centroid_loc, block_type, blk_loc_registry, pr, rng)) {
```
vaughnbetz, line 627: Simplify to found_legal_subtile = (function return value)

vaughnbetz (changes requested):
Almost there! Just one suggested stylistic simplification

vaughnbetz (approved):
LGTM! Thanks.

PR #2899, opened 2025-02-18
[AP][PartialLegalizer] Density Abstraction

AlexandreSinger:
Abstracted how density is managed in the AP flow into a manager class. This manager class contains two sub-classes: one for the bins and one for the mass calculation.

Density is a function of mass and volume.

The volume is related to how we break the FPGA chip into discrete bins. These bins contain atoms and they take up a rectangular amount of space on the FPGA.

The mass is related to how "heavy" each APBlock is inside of the bin. This is computed using an M-dimensional vector where M is the number of models in the architecture. Each logical / physical tile has a capacity that it may hold.

The density manager class manages how the bins are constructed, how much mass each bin can contain, and how much mass each bin currently contains. This current version of the density manager class creates a bin for each tile (with a unique root tile location). The capacity of each bin is related to the capacity of the physical tile at that location.

Partial Legalizers can ask questions to the density manager class. They may ask for all of the overfilled bins, what bin is located at a given location, what bin contains a given atom, how overfilled / underfilled a bin is, etc.

This abstraction will make it much easier to optimize the density of the flat placement, since it separates algorithm from density calculation. This will make writing future Partial Legalizers considerably easier.


vpr/src/analytical_place/flat_placement_density_manager.cpp
```diff
@@ -0,0 +1,302 @@
...
+                                  bin_region.top_right().x() - epsilon);
+    double y = std::clamp<double>(p_placement.block_y_locs[blk_id],
+                                  bin_region.bottom_left().y() + epsilon,
+                                  bin_region.top_right().y() - epsilon);
+    return vtr::Point<double>(x, y);
+}
+
+void FlatPlacementDensityManager::export_placement_from_bins(PartialPlacement& p_placement) const {
+    // Updates the partial placement with the location of the blocks in the bin
+    // by moiving the blocks to the point with the bin closest to where they
```
vaughnbetz, line 196: moiving -> moving

vpr/src/analytical_place/flat_placement_density_manager.cpp
```diff
@@ -0,0 +1,302 @@
...
+            return false;
+        }
+    }
+    // Make sure that every bin has the correct utilization, supply, and demand.
+    for (FlatPlacementBinId bin_id : bins_.bins()) {
+        PrimitiveVector calc_utilization;
+        for (APBlockId blk_id : bins_.bin_contained_blocks(bin_id)) {
+            calc_utilization += mass_calculator_.get_block_mass(blk_id);
+        }
+        if (bin_utilization_[bin_id] != calc_utilization) {
```
vaughnbetz, line 243: Could add a verification routine to check that what blocks a bin thinks are placed in it matches what the blocks think in terms of what bins they are in.  That should go in a verify routine in the bin class itself.

vaughnbetz (commented):
LGTM; a few suggestions.

PR #2900, opened 2025-02-18
Change NightlyTest workflow to not run automatically

AmirhosseinPoolad:
Since NightlyTest is currently not working, we should stop it from running automatically everyday. We still have the option to manually run the workflow if we need to.


AlexandreSinger (approved):
LGTM

PR #2902, opened 2025-02-19
[Place] Cleaned Up Placement Context Initialization

AlexandreSinger:
Cleaned up how the variables in the placement context were initialized to make it easier to initialize the placement context outside of the placement stage (Analytical Placement).

Added an optional argument to the Placer class to allow an initial placement to be passed in. This allows a different initial placement to be used without having to pull the initial placer out of the Placer class.

With this change, it should be easier to create the Detailed Placer in Analytical Placement.

### DIFF
--- a/vpr/src/base/blk_loc_registry.cpp
+++ b/vpr/src/base/blk_loc_registry.cpp
@@ -1,12 +1,60 @@
 
 #include "blk_loc_registry.h"
 
+#include "device_grid.h"
 #include "move_transactions.h"
 #include "globals.h"
 #include "physical_types_util.h"
+#include "vpr_context.h"
+
+/**
+ * @brief Initialize `grid_blocks`, the inverse structure of `block_locs`.
+ *
+ * The container at each grid block location should have a length equal to the
+ * subtile capacity of that block. Unused subtile would be marked ClusterBlockId::INVALID().
+ */
+static GridBlock init_grid_blocks(const DeviceGrid& device_grid);
 
 BlkLocRegistry::BlkLocRegistry()
     : expected_transaction_(e_expected_transaction::APPLY) {}
+
+void BlkLocRegistry::init(const ClusteredNetlist& clb_nlist, const DeviceGrid& device_grid) {
+    auto& block_locs = mutable_block_locs();
+    auto& grid_blocks = mutable_grid_blocks();
+
+    /* Initialize the lookup of CLB block positions */
+    block_locs.clear();
+    block_locs.resize(clb_nlist.blocks().size());
+
+    /* Initialize the reverse lookup of CLB block positions */
+    grid_blocks = init_grid_blocks(device_grid);
+
+    /* Initialize the grid blocks to empty.
+     * Initialize all the blocks to unplaced.
+     */
+    clear_all_grid_locs();
+}
+
+static GridBlock init_grid_blocks(const DeviceGrid& device_grid) {
+    size_t grid_width = device_grid.width();
+    size_t grid_height = device_grid.height();
+    int num_layers = device_grid.get_num_layers();
+
+    /* Structure should have the same dimensions as the grid. */
+    auto grid_blocks = GridBlock(grid_width, grid_height, num_layers);
+
+    for (int layer_num = 0; layer_num < num_layers; layer_num++) {
+        for (size_t x = 0; x < grid_width; x++) {
+            for (size_t y = 0; y < grid_height; y++) {
+                const t_physical_tile_loc tile_loc({(int)x, (int)y, layer_num});
+                auto type = device_grid.get_physical_type(tile_loc);
+                grid_blocks.initialized_grid_block_at_location(tile_loc, type->capacity);
+            }
+        }
+    }
+
+    return grid_blocks;
+}
 
 const vtr::vector_map<ClusterBlockId, t_block_loc>& BlkLocRegistry::block_locs() const {
     return  block_locs_;
--- a/vpr/src/base/blk_loc_registry.cpp
+++ b/vpr/src/base/blk_loc_registry.cpp
@@ -7,18 +7,12 @@
 #include "physical_types_util.h"
 #include "vpr_context.h"
 
-/**
- * @brief Initialize `grid_blocks`, the inverse structure of `block_locs`.
- *
- * The container at each grid block location should have a length equal to the
- * subtile capacity of that block. Unused subtile would be marked ClusterBlockId::INVALID().
- */
-static GridBlock init_grid_blocks(const DeviceGrid& device_grid);
-
 BlkLocRegistry::BlkLocRegistry()
     : expected_transaction_(e_expected_transaction::APPLY) {}
 
-void BlkLocRegistry::init(const ClusteredNetlist& clb_nlist, const DeviceGrid& device_grid) {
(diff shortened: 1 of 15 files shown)


vpr/src/place/place.cpp
```diff
@@ -107,18 +110,74 @@ void try_place(const Netlist<>& net_list,
...
-    placer.copy_locs_to_global_state(place_ctx);
+    placer.copy_locs_to_global_state(g_vpr_ctx.mutable_placement());
+
+    // Clean the variables in the placement context. This will deallocate memory
+    // used by variables which were allocated in the placement context and are
+    // never used outside of placement.
+    clean_placement_context(g_vpr_ctx.mutable_placement());
+}
+
+void init_placement_context(PlacementContext& place_ctx,
```
soheilshahrouz, line 130: I suggest making this function a member function of PlacementContext. It also modifies FloorplanningContext, which is not implied by the function name. It might not be a bad idead to add an init member function to FloorplanningContext.

vpr/src/place/place.cpp
```diff
@@ -107,18 +110,74 @@ void try_place(const Netlist<>& net_list,
...
+    alloc_and_load_compressed_cluster_constraints();
+
+    /* To make sure the importance of NoC-related cost terms compared to
+     * BB and timing cost is determine only through NoC placement weighting factor,
+     * we normalize NoC-related cost weighting factors so that they add up to 1.
+     * With this normalization, NoC-related cost weighting factors only determine
+     * the relative importance of NoC cost terms with respect to each other, while
+     * the importance of total NoC cost to conventional placement cost is determined
+     * by NoC placement weighting factor.
+     * FIXME: Why is this modifying the noc_opts?
```
soheilshahrouz, line 164: The documentation of the function and the help message for command line options explains why. I think this like should be removed.

vpr/src/place/place.cpp
```diff
@@ -107,18 +110,74 @@ void try_place(const Netlist<>& net_list,
...
+     */
+    if (noc_opts.noc) {
```
soheilshahrouz, line 166: This line does not belong here. `noc_opts` is not part of the placement context.

vpr/src/place/place.cpp
soheilshahrouz: It might be a good idea to print a warning message when NoC weighting factors are renormalized. I'll submit a PR when this one is merged.

vpr/src/place/place.cpp
```diff
@@ -107,18 +110,74 @@ void try_place(const Netlist<>& net_list,
...
+        normalize_noc_cost_weighting_factor(const_cast<t_noc_opts&>(noc_opts));
+    }
+}
+
+void clean_placement_context(PlacementContext& place_ctx) {
```
soheilshahrouz, line 171: I think init and clean functions should be member functions of the class they are initializing and cleaning. This function modifies the floorplanning context, which is not implied by them.

vpr/src/place/placer.cpp
```diff
@@ -46,41 +50,34 @@ Placer::Placer(const Netlist<>& net_list,
...
-    if (flat_placement_info.valid) {
-        log_flat_placement_reconstruction_info(flat_placement_info,
-                                               blk_loc_registry.block_locs(),
-                                               g_vpr_ctx.clustering().atoms_lookup,
-                                               g_vpr_ctx.atom().lookup,
-                                               g_vpr_ctx.atom().nlist,
-                                               g_vpr_ctx.clustering().clb_nlist);
+    if (init_place.has_value()) {
+        // If an initial placement has been provided, use that.
+        blk_loc_registry = *init_place;
```
soheilshahrouz, line 64: We pass the initial placement by reference and then copy it. Why not pass it by value and move it?

If we do this, I guess we can move the ownership of the BlkLocRegistry object to this class.

soheilshahrouz: Even copy_locs_to_global_state can be renamed to move_locs_to_global_state to avoid a copy.

If you agree, I can do this in another PR.

vpr/src/base/blk_loc_registry.cpp
```diff
@@ -1,13 +1,61 @@
...
+ *
+ * The container at each grid block location should have a length equal to the
+ * subtile capacity of that block. Unused subtile would be marked ClusterBlockId::INVALID().
+ */
+static GridBlock init_grid_blocks(const DeviceGrid& device_grid);
 
 BlkLocRegistry::BlkLocRegistry()
     : expected_transaction_(e_expected_transaction::APPLY) {}
 
+void BlkLocRegistry::init(const ClusteredNetlist& clb_nlist, const DeviceGrid& device_grid) {
```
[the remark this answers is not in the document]
AlexandreSinger: Although I do not like it, you are 100% right. Instead of making the entire file like this (whcih would be a process), I made the initialization method take no arguments and use the global context instead. Thanks for flagging this!

vpr/src/place/place.cpp
AlexandreSinger: I agree. I have moved this out of this method and moved it into place.cpp with a FIXME saying that the NoC weighting factors should not be modified there. I personally think that this should be checked when the factors are loaded from the user and fixed if necessary there. I find it strange that the NoC Opts object is being modified (especially since its using const_cast which is bad practice).

AlexandreSinger:
@soheilshahrouz Thank you very much for your comments. It was a good idea to move the init / clean methods into the context classes and separating them based on which contexts they update.

vpr/src/base/grid_block.h
```diff
@@ -38,6 +38,14 @@ class GridBlock {
         grid_blocks_.resize({layers, width, height});
     }
 
+    /**
+     * @brief Initialize `grid_blocks`, the inverse structure of `block_locs`.
+     *
+     * The container at each grid block location should have a length equal to the
+     * subtile capacity of that block. Unused subtile would be marked ClusterBlockId::INVALID().
```
vaughnbetz, line 45: nit: subtiles

vpr/src/base/vpr_context.h
```diff
@@ -480,6 +505,42 @@ struct RoutingContext : public Context {
...
+     * the constraints of clusters.
+     *
+     * This must be called before using the cluster_constraints.
+     */
+    void update_floorplanning_context_post_pack();
+
+    /**
+     * @brief Update the floorplanning constraints before placement.
+     *
+     * Placement groups clusters together into molecules which must be placed
```
vaughnbetz, line 522: molecules -> macros

vpr/src/base/vpr_context.h
```diff
@@ -337,6 +338,30 @@ struct PlacementContext : public Context {
...
 
+    /**
+     * @brief Initialize the variables stored within the placement context. This
+     *        must be called before performing placement, but must be called
+     *        after the clusters are loaded.
+     *
+     *  @param placer_opts
+     *      The options passed into the placer.
+     *  @param directs
+     *      A lost of the direct connections in the architecture.
```
vaughnbetz, line 349: A list of the direct

vaughnbetz (changes requested):
Small changes.

vaughnbetz:
Alex tells me the results are identical before and after this change.

Issue #2903, opened 2025-02-21
remove :  capacity attribute in tile


markram1729:
I see in vpr code ProcessTileProp doesn't have attribute capacity which is handled in sub_tile

in [documentation](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/4c81b1d168f64974b33ab7bb5fc45671e88f4f79/doc/src/arch/reference.rst?plain=1#L852) tiles have attribute capacity

PR #2904, opened 2025-02-21
remove : capacity attribute in tile

markram1729:
#### Related Issue

#2903 
#### Types of changes

#### Checklist:


doc/src/arch/reference.rst
```diff
@@ -849,7 +849,7 @@ Each tile type is specified with the ``<tile>`` tag withing the ``<tiles>`` tag.
 
 Tile
 ~~~~
-.. arch:tag:: <tile name="string" capacity="int" width="int" height="int" area="float"/>
+.. arch:tag:: <tile name="string"  width="int" height="int" area="float"/>
```
AlexandreSinger, line 852: You left behind two spaces between "string" and "width", could you please remove that?

AlexandreSinger (commented):
This LGTM, thanks for noticing this! I just had one minor comment.

@vaughnbetz I have checked the code and ensured that this property is not being parsed within VTR:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3ef570f852d9c430fb1f3d6ed267f5c0120d60ee/libs/libarchfpga/src/read_xml_arch_file.cpp#L2995-L3013

The `t_physical_tile_type` struct does have a member called "capacity"; however, according to the parser this is defined as the sum of the capacities of all sub-tiles in this physical tile type. So I think the change in this PR is correct.

doc/src/arch/reference.rst
markram1729: I have removed the extra space now

PR #2905, opened 2025-02-22
[CI] Made Nightly Tests Continue on Error

AlexandreSinger:
Currently, the Nightly Tests will stop whenever any of the Nightly Tests fail (and will not run the following tests). So for example, if Nightly Test 1 has a QoR failure, all other Nightly tests will not be run.

Made each step in the Nightly Test manual job continue on error. This will run the following steps even if one of them fail.

I found this while running the Nightly Tests on Haydar's recent PR (PR #2897 ):
image

I do not think this is the intended behavior we want based on previous discussions.

### DIFF
--- a/.github/workflows/nightly_test_manual.yml
+++ b/.github/workflows/nightly_test_manual.yml
@@ -63,37 +63,45 @@ jobs:
     # TODO: We could expose more parallelism if we had one task list which ran
     #       all of these.
     - name: 'Run Nightly Test 1'
+      continue-on-error: true
       run: |
         source .venv/bin/activate
         ./run_reg_test.py -j12 vtr_reg_nightly_test1
 
     - name: 'Run Nightly Test 2'
+      continue-on-error: true
       run: |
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
I ran the current Nightly Tests (which I know have QoR failures) and found the following problem with my solution:
image

According to the annotations, Nightly Tests 1 and 3 failed; however the overall workflow did not fail! This is because we marked them continue on failure. This is not desired behaviour. We want a big red X on the worflow saying something went wrong.

@AmirhosseinPoolad Can you look into this? I think as a work-around we can add a step at the end which will produce a "summary" of the nightly tests run and that one can fail. I think its possible to have each nightly test step produce some variable (pass / fail) which the summary step can consume at the end. What do you think?

AmirhosseinPoolad:
Decided to make a new PR since this one had a merge commit and I don't feel confident enough with my gitfu to change stuff from 2 commits ago. Closing this.

PR #2906, opened 2025-02-23
Clarified target_ext_pin_utilization command line option documentation

vaughnbetz:
#### Description
Clarified that setting any target_ext_pin_utilization values turns off auto mode and the target pin utilization for any unspecified pin types is set to 100% in that case.

#### Related Issue
This previously confused researchers at Cornell.


AlexandreSinger (approved):
LGTM

PR #2910, opened 2025-02-24
Make Nightly Tests Continue on Error

AmirhosseinPoolad:
Basically the same thing as #2905 but uses a different way to do it that hopefully avoids the issue of the run being considered successful even if some tests have failed. Did it in a different PR and branch because I didn't want to mess with the origin too much.


AmirhosseinPoolad:
@AlexandreSinger Do you mind checking to see if this fixes the problem?

AlexandreSinger (approved):
This worked perfectly!
image

Very well done!

PR #2912, opened 2025-02-26
[AP][DetailedPlacer] Added Detailed Placer Class

AlexandreSinger:
The Detailed Placer is a pass in the AP flow which takes a legal placement and optimizes it, while still remaining legal.

Previously in the AP flow, the annealer from the Placement stage was always used to optimize the placement. Now the user can choose to not use the annealer, by selecting the "none" Detailed Placer. Other Detailed Placers can now also be added to explore other Detailed Placement algorithms in VTR.


AlexandreSinger:
@vaughnbetz I have finally separated out the Detailed Placer into its own stage in the AP flow and it can be selected from the command-line:
Screen Shot 2025-02-25 at 8 35 48 PM

Please review when you have a moment.

AlexandreSinger:
@amin1377 FYI

vpr/src/base/read_options.cpp
```diff
@@ -1791,6 +1826,14 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("appack")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    ap_grp.add_argument<e_ap_detailed_placer, ParseAPDetailedPlacer>(args.ap_detailed_placer, "--ap_detailed_placer")
+        .help(
+            "Controls which Detailed Placer to use in the AP Flow.\n"
+            " * none: Do not perform any detailed placement. Directly routes the result of the Full Legalizer.\n"
```
vaughnbetz, line 1832: Delete "Directly routes the result ..." (Implies --route is on, which may not be the case."
Instead could say "i.e. the output of the analytic placer will be used without modification."

vaughnbetz:
Looks like there is a bug; I assume a debug assert is firing, as many circuits failed in the highest assertion level test.  I don't see the assert failure in the output though; here are the last lines (which imply we vpr crashes right at the start of placement):

### Computing src/opin lookahead took 0.01 seconds (max_rss 74.8 MiB, delta_rss +0.0 MiB)
## Computing router lookahead map took 0.21 seconds (max_rss 74.8 MiB, delta_rss +0.0 MiB)
Serial Connection Router is being destroyed. Time spent on path search: 0.000 seconds.
# Computing placement delta delay look-up took 1.63 seconds (max_rss 74.8 MiB, delta_rss +0.0 MiB)
# Placement
## Initial Placement

AlexandreSinger:
> Looks like there is a bug; I assume a debug assert is firing, as many circuits failed in the highest assertion level test. I don't see the assert failure in the output though; here are the last lines (which imply we vpr crashes right at the start of placement):
> 
> ```
> 
> ### Computing src/opin lookahead took 0.01 seconds (max_rss 74.8 MiB, delta_rss +0.0 MiB)
> ## Computing router lookahead map took 0.21 seconds (max_rss 74.8 MiB, delta_rss +0.0 MiB)
> Serial Connection Router is being destroyed. Time spent on path search: 0.000 seconds.
> # Computing placement delta delay look-up took 1.63 seconds (max_rss 74.8 MiB, delta_rss +0.0 MiB)
> # Placement
> ## Initial Placement
> ```

@vaughnbetz Is this comment for another PR? This branch is passing all of the tests and after merging into master, master also passes all of the tests.

PR #2913, opened 2025-02-26
Clean up AtomContext

AmirhosseinPoolad:
Changes the fields of AtomContext from public to private and provides two methods to get each field: One that returns an immutable const reference and another that returns a mutable reference. The main advantage here is that it is now simple to see where the global context is written to and where it is only read from.

This is similar to the work done on PlacementContext and would make refactoring the packing code a bit simpler knowing where exactly the global context is changed (anywhere a mutable_X function is called). There are a lot of changed lines but they are mostly trivial and should not affect functionality in any way.

Note that a lot of changes seem to be my text editor automatically removing extra spaces and things like that when I saved. Sorry for making reviewing this already lengthy PR even harder, just turned that option off!


vpr/src/route/route_budgets.cpp
```diff
@@ -950,4 +950,4 @@ void route_budgets::set_should_reroute(ParentNetId net_id, bool value) {
     if (set) {
         should_reroute_for_hold[net_id] = value;
     }
-}
\ No newline at end of file
+}
```
AlexandreSinger, line 953: This is actually a perfect example of when Git can get confused! Notice that you did not modify line 953 here, but Git got confused and said that you modified it. This is the exact reason why adding spaces to the end of files can be so usefull.

vpr/src/base/vpr_context.h
```diff
@@ -72,15 +72,27 @@ struct AtomContext : public Context {
...
+    public:
+
+    const AtomNetlist& netlist() const {return nlist_;}
+    AtomNetlist& mutable_netlist() {return nlist_;}
+
+    const AtomLookup& lookup() const {return lookup_;}
+    AtomLookup& mutable_lookup() {return lookup_;}
+
+    const FlatPlacementInfo& flat_placement_info() const {return flat_placement_info_;}
+    FlatPlacementInfo& mutable_flat_placement_info() {return flat_placement_info_;}
```
AlexandreSinger, line 94: I would declare all of these accessor methods `inline`:
```cpp
inline AtomNetlist& mutable_netlist() { ... }
```
The inline keyword tells the compiler to directly inline this line of code whenever it is used in the code. It should be obvious to the compiler that it should inline this method, but it can be good to be explicit in this case since we access these variables so often!

AlexandreSinger (commented):
@AmirhosseinPoolad Thank you so much for this cleanup. I agree with adding mutable and immutable accessors to the members of the Atom Context (I also welcome the change of `nlist` to `netlist`). I left some comments.

I welcome the removal of some of the white spaces in the codebase; you would be suprised how much space this can save! However, I made a comment about this in the review; I recommend leaving new line characters at the end of files.

vpr/src/base/vpr_context.h
AmirhosseinPoolad: Agree with this. Just a thought: inline keywords just suggests that it should be inlined but the compiler isn't forced to do so. Do we have -Winline turned on to give warnings if the compiler *doesn't* inline an inline function? Could be a useful compiler enforced "please don't make this function complicated" label.

vpr/src/analytical_place/analytical_placement_flow.cpp
```diff
@@ -107,4 +107,3 @@ void run_analytical_placement_flow(t_vpr_setup& vpr_setup) {
                                                                         device_ctx.grid);
     full_legalizer->legalize(p_placement);
 }
-
```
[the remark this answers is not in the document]
AmirhosseinPoolad: Yeah that makes sense, will revert these.

vpr/src/base/vpr_context.h
AlexandreSinger: You are correct. The compiler is not forced to inline even if we say so. I do not think we have the -Winline flag turned on. We should discuss with Vaughn / the VTR meeting if we should! I do not think its extremely important, but declaring your intent on making a function inline in a header file is usually just good practice!

PR #2914, opened 2025-02-26
[VTR][Util] Added Prefix Sum Class

AlexandreSinger:
A prefix sum is a classic data structure for quickly finding the sum of any sub-region of fixed N-Dimensional array of values. It pre-computes the partial sums of regions within an array and uses that information later to save time.

This is currently used in the net_cost_handler to pre-compute number of inter-die connections.

This class can be useful in other algorithms. For example, in a Partial Legalizer in an AP flow, it can allow for quick lookups into the capacity and utilization of regions on the FPGA.

Moved the Prefix Sum class from the net_cost_handler into the VTR Util.

### DIFF
--- a/libs/libvtrutil/src/vtr_prefix_sum.h
+++ b/libs/libvtrutil/src/vtr_prefix_sum.h
@@ -1,0 +1,283 @@
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    February 2025
+ * @brief   Definition of the Prefix Sum class which enables O(1) time-complexity
+ *          sums over regions of an unchanging grid of values.
+ */
+
+#pragma once
+
+#include <functional>
+#include <vector>
+#include "vtr_assert.h"
+#include "vtr_ndmatrix.h"
+
+namespace vtr {
+
+/**
+ * @brief 1D Prefix Sum manager class.
+ *
+ * Given an array of values, it may be necessary to find the sum of values
+ * within a continuous sub-section of the array. If this operation needs to be
+ * performed many times, this may be expensive in runtime to calculate.
+ *
+ * If the array of values does not change, we can create a prefix sum which will
+ * allow us to get the sum of values in some continuous sub-section of the array
+ * in O(1) time, instead of O(k) time where k is the number of values in the
+ * sub-section.
+ *
+ * This class has a space complexity of O(l) where l is the length of the array
+ * of values.
+ *
+ *
+ * Static Array of Values Example (values stored in a vector):
+ *
+ * std::vector<float> vals = {...};
+ *
+ * // Build the Prefix Sum
+ * vtr::PrefixSum1D<float> prefix_sum(vals);
+ *
+ * // Compute the sum of the values between index 3 and 7 of the array (inclusive)
+ * float sum = prefix_sum.get_sum(3, 7);
+ *
+ *
+ * Dynamic Vector of Values Example (values derived at runtime):
+ *
+ * // Build the Prefix Sum using a lambda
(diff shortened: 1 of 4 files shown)


libs/libvtrutil/src/vtr_prefix_sum.h
```diff
@@ -0,0 +1,135 @@
...
+            return prefix_sum_[upper_x][upper_y];
+        }
+        if (lower_x == 0) {
+            return prefix_sum_[upper_x][upper_y] - prefix_sum_[upper_x][lower_y - 1];
+        }
+        if (lower_y == 0) {
+            return prefix_sum_[upper_x][upper_y] - prefix_sum_[lower_x - 1][upper_y];
+        }
+
+        // General formula without bounds checking.
```
vaughnb-cerebras, line 116: Could comment why this works -- first term is the sum from (0,0) to (upper_x, upper_y), the next two terms subtract off the parts we should not include due to lower_x and lower_y, and the last term adds back in the part from (0,0) to (lower_x-1,lower_y-1) as it has been double-subtracted.

libs/libvtrutil/src/vtr_prefix_sum.h
```diff
@@ -0,0 +1,135 @@
...
+        // Some safety asserts.
+        VTR_ASSERT_SAFE_MSG(lower_x <= upper_x, "lower_x is larger than upper_x");
+        VTR_ASSERT_SAFE_MSG(lower_y <= upper_y, "lower_y is larger than upper_y");
+        VTR_ASSERT_SAFE_MSG(lower_x < prefix_sum_.dim_size(0), "lower_x out of range");
+        VTR_ASSERT_SAFE_MSG(upper_x < prefix_sum_.dim_size(0), "upper_x out of range");
+        VTR_ASSERT_SAFE_MSG(lower_y < prefix_sum_.dim_size(1), "lower_y out of range");
+        VTR_ASSERT_SAFE_MSG(upper_y < prefix_sum_.dim_size(1), "upper_y out of range");
+
+        // Get sum when on the bounds.
+        if (lower_x == 0 && lower_y == 0) {
```
vaughnb-cerebras, line 106: I don't have a strong preference on this, but you could get rid of the if-statements by storing the appropriate (0) values in a larger matrix that starts at (-1,-1)

vaughnb-cerebras (commented):
Nice change!  A couple of thoughts embedded.
I believe also use 1D prefix sums in the placer  in case you want to also pull those out. See https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/5a2231318067922e3a704f958beba4d85789360e/vpr/src/place/net_cost_handler.cpp#L154

libs/libvtrutil/src/vtr_prefix_sum.h
AlexandreSinger: This was a good idea. Originally I wanted to save space by not doing this; but then I realized that the default case would need to do 3 comparisons before doing useful work. However, I did want to avoid using NdOffsetMatrix since I think it is overkill for this application. Its much easier just to offset the accesses.

AlexandreSinger:
@vaughnbetz Thank you for your feedback on this! I did not realize that those were 1D Prefix Sums; I have created a class for PrefixSum1D and have converted them into using this class. I found that the original code was a little confusing with how it handled 0 channel widths, so I rewrote it slightly to make this more obvious. This does not change the functionality at all, just makes it more explicit that we assume the channel widths are at least 1.

vaughnb-cerebras (approved):
Thanks, LGTM!

AlexandreSinger:
Cool! Ill merge when it passes CI.

PR #2920, opened 2025-03-03
[Router] Upstream Fine-Grained Parallel Router (FPT'24)

ueqri:
Upstreamed the fine-grained parallel router implementation into the VTR master. The original branch is https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/mq-parallel-router.

Modified the MultiQueue (SPAA'24) implementation and integrated it into the VTR codebase.

VTR strong regression tests passed.

### TODOs

- [x] add more comments and descriptions
- [x] eliminate more duplicated codes
- [ ] collect performance results


doc/src/vpr/command_line_usage.rst
```diff
@@ -1488,13 +1488,84 @@ The following options are only valid when the router is in timing-driven mode (t
...
+    This parameter has no effect if :option:`--enable_parallel_connection_router` is not set.
+
+    **Default:** ``2``
+
+.. option:: --multi_queue_direct_draining {on | off}
+
+    Controls whether to enable queue draining optimization for MultiQueue-based parallel connection router.
+
+    When enabled, queues can be emptied quickly by draining all elements if no further solutions need to be explored in the
+    path search to guarantee optimality or determinism after reaching the target.
```
AlexandreSinger, line 1563: I think this may be confusing and needs more explanation. The queue draining optimization requires the ordering heuriststic (astar fac/offset) to be admissible in order to be deterministic.

The wording above also makes it sound like you can only use this if optimality and determinism is required; however this optimization can be turned on even if you do not care about either of these.

utils/route_diag/src/main.cpp
```diff
@@ -97,7 +97,8 @@ static void do_one_route(const Netlist<>& net_list,
                                                   segment_inf,
                                                   is_flat);
 
-    ConnectionRouter<FourAryHeap> router(
+    // TODO: adding tests for parallel connection router
```
AlexandreSinger, line 100: Is this TODO still valid?

vpr/src/route/NestedNetlistRouter.h
```diff
@@ -67,19 +73,38 @@ class NestedNetlistRouter : public NetlistRouter {
     /** Route all nets in a PartitionTree node and add its children to the task queue. */
     void route_partition_tree_node(PartitionTreeNode& node);
 
-    ConnectionRouter<HeapType> _make_router(const RouterLookahead* router_lookahead, bool is_flat) {
+    ConnectionRouter<HeapType>* _make_router(const RouterLookahead* router_lookahead,
```
AlexandreSinger, line 76: Instead of returning a bare pointer, I recommend returning a unique_ptr and using the `make_unique` method to make them. This is safer than a bare pointer and will clean itself up once the object is destroyed so the destructor is unnecessary.

See the Analytical Placement code for an example:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/36ba8dab8444893817823eef76975848a11ba2ef/vpr/src/analytical_place/full_legalizer.cpp#L57-L88

vpr/src/route/NestedNetlistRouter.h
```diff
@@ -109,17 +134,17 @@ class NestedNetlistRouter : public NetlistRouter {
...
      * These are maps because thread::id is a random integer instead of 1, 2, ... */
-    std::unordered_map<std::thread::id, ConnectionRouter<HeapType>> _routers_th;
+    std::unordered_map<std::thread::id, ConnectionRouter<HeapType>*> _routers_th;
     std::unordered_map<std::thread::id, RouteIterResults> _results_th;
     std::mutex _storage_mutex;
 
     /** Get a thread-local ConnectionRouter. We lock the id->router lookup, but this is
      * accessed once per partition so the overhead should be small */
-    ConnectionRouter<HeapType>& get_thread_router() {
+    ConnectionRouter<HeapType>* get_thread_router() {
```
AlexandreSinger, line 143: This is up to you, but I think you can still return by reference here by dereferencing the pointer to the router being returned. We know the router objects will not be destroyed until later so we know the reference should be valid.

vpr/src/route/NestedNetlistRouter.tpp
```diff
@@ -66,10 +66,9 @@ void NestedNetlistRouter<HeapType>::route_partition_tree_node(PartitionTreeNode&
     /* Route all nets in this node serially */
     for (auto net_id : nets) {
         auto& results = get_thread_results();
-        auto& router = get_thread_router();
 
         auto flags = route_net(
-            router,
+            *get_thread_router(),
```
AlexandreSinger, line 71: Especially since down here we are going to dereference it anyways.

vpr/src/route/SerialNetlistRouter.h
```diff
@@ -32,30 +34,51 @@ class SerialNetlistRouter : public NetlistRouter {
...
+    }
 
     RouteIterResults route_netlist(int itry, float pres_fac, float worst_neg_slack);
     void handle_bb_updated_nets(const std::vector<ParentNetId>& nets);
     void set_rcv_enabled(bool x);
     void set_timing_info(std::shared_ptr<SetupHoldTimingInfo> timing_info);
 
   private:
-    ConnectionRouter<HeapType> _make_router(const RouterLookahead* router_lookahead, bool is_flat) {
+    ConnectionRouter<HeapType>* _make_router(const RouterLookahead* router_lookahead,
```
AlexandreSinger, line 47: Same comment down here. This should be a unique pointer for improved safety.

vpr/src/route/multi_queue_d_ary_heap.h
```diff
@@ -0,0 +1,125 @@
...
+
+#ifndef _MULTI_QUEUE_D_ARY_HEAP_H
+#define _MULTI_QUEUE_D_ARY_HEAP_H
+
+#include <tuple>
+#include "device_grid.h"
+#include "heap_type.h"
+#include "multi_queue_d_ary_heap.tpp"
+
+using MQHeapNode = std::tuple<HeapNodePriority, uint32_t /*FIXME*/>;
```
AlexandreSinger, line 28: What is this FIXME for? Is this something we should look at? If you want to handle it in the future I recommend making the comment more clear.

If this is because we are assuming the node is never larger that the largest int32, I recommend adding an assert to ensure this! i.e. the number of Nodes never exceeds the max size of uint32_t.

AlexandreSinger: There also appear to be a lot of TODOs and FIXMEs in this file. Just a heads up.

AlexandreSinger (commented):
Hi @ueqri 

These changes are looking great! I have left some comments.

You mentioned to me that the parallel router code was not 100% commented yet, so I will review that after you give me the go ahead.

vpr/src/route/multi_queue_d_ary_heap.h
ueqri: Enhanced comments.

Most of the FIXMEs are related to MQ_IO refactoring (e.g., making MQ_IO compatible with unified heap node struct and comparator defined in heap_type.h), which is left for another PR.

vpr/src/route/NestedNetlistRouter.h
ueqri: Good idea! Thanks, Alex.

doc/src/vpr/command_line_usage.rst
ueqri: Agree. Has added more explanation.

ueqri:
Hi Vaughn, Alex, and Fahri, this PR is ready to review eventually! @vaughnbetz @AlexandreSinger @duck2

The weird bug (causing the VTR router to get stuck) has been fixed, which is detailed in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2920/commits/46d6c2c4663da400f956c0b19f98076c301c19da. I have run `vtr_reg_qor_chain` for three different mode: (1) 'serial mode' 1T+2Q (1 thread, 2 queues), (2) 2T+4Q, and (3) 4T+2Q. All of them have passed and determinism has been maintained between different runs using the same mode.

- [ ] I am still running tests on wintermute to collect QoR results from `vtr_reg_qor_chain`, `koios_medium`, `koios_large`, and `titan_quick_qor`. The result table (and comparison with baseline at master) will be posted in the PR once the tests are done.
- [ ] CI tests for parallel connection router are taking quite a long time (as we tested three/six different combinations of CLI options for five existing tasks, roughly adding 18 new test tasks to the CI tests). I will fix that by examining all newly-added tests and removing those very time-consuming test tasks. The golden results will be pushed once the CI tests have been updated.

Hi Fahri @duck2, please switch to the [branch in this PR](https://github.com/ueqri/vtr-verilog-to-routing/tree/upstream-fpt24-fine-grained-parallel-router) if needed since this one has been more extensively tested and works well.

AlexandreSinger:
@ueqri That's great news! I am very happy that the bug was something simple. It happens whenever an API has two arguments of the same type. Thanks for fixing my comments, I really like using smart pointers whenever possible in VTR. I will give it another once over sometime today.

AlexandreSinger (approved):
LGTM! Thank you so much for all your hard work @ueqri 

I want to merge this soon so it does not regress. Do you have anything more you want to change about this PR before it is merged?

In another PR we should change the spin barrier to be a mutex barrier instead. But that should be explored in more detail and in a future PR.

vaughnbetz:
For safety, it would be good to run a QoR test on Titan for the default flow before and after this change to confirm no regression (cpu time, QoR) in the default flow.  I'll merge in advanced of that though, but please add a QoR link and tag me when you have it.

ueqri:
@AlexandreSinger Nothing more involving code changes in this PR. The barrier changes and multi-queue implementation refactoring (as mentioned in the `FIXME`) will be in a future PR.

@vaughnbetz Will do! I am running the Titan QoR tests and will keep you updated with the results. Thanks, Vaughn!

PR #2921, opened 2025-03-03
[AP][FullLegalizer] Basic Min. Disturbance Full Legalizer

haydar-c:
Added the "basic_min_disturbance" command line options into the AP Flow. Ideally, this will be another Full Legalizer method when implemented on top of the ones here (https://docs.verilogtorouting.org/en/latest/vpr/command_line_usage/#cmdoption-vpr-ap_full_legalizer).

I also added the reading from flat placement and writing to flat placement into AP Flow. If the read option is provided, Global Placement is skipped and Partial Placement is constructed with provided flat placement information. If the write option is provided, flat placement is written at the end of AP Flow.


vpr/src/analytical_place/analytical_placement_flow.cpp
```diff
@@ -77,15 +95,25 @@ void run_analytical_placement_flow(t_vpr_setup& vpr_setup) {
...
+    } else {
+        // Run the Global Placer
+        std::unique_ptr<GlobalPlacer> global_placer = make_global_placer(e_global_placer::SimPL,
+                                                                        ap_netlist,
+                                                                        prepacker,
+                                                                        atom_nlist,
+                                                                        device_ctx.grid,
+                                                                        device_ctx.logical_block_types,
+                                                                        device_ctx.physical_tile_types);
+        p_placement = global_placer->place();
```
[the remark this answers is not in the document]
AlexandreSinger: Another idea is that you can use std::move; but in order to use it you would need to define a move constructor and move assignment operator for the partial placement class:
https://www.geeksforgeeks.org/move-constructors-in-c-with-examples/

Note: The reason this did not work in my example I was showing you is because I forgot to add the move constructor and move assignment operator.

vpr/src/base/read_options.cpp
```diff
@@ -1787,7 +1791,8 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .help(
             "Controls which Full Legalizer to use in the AP Flow.\n"
             " * naive: Use a Naive Full Legalizer which will try to create clusters exactly where their atoms are placed.\n"
-            " * appack: Use APPack, which takes the Packer in VPR and uses the flat atom placement to create better clusters.")
+            " * appack: Use APPack, which takes the Packer in VPR and uses the flat atom placement to create better clusters.\n"
+            " * basic_min_disturbance: Use The Basic Min. Disturbance Full Legalizer which uses flat placement to reconstruct clusters identically, but the orphan cluster handling method is yet to be determined.")
```
AlexandreSinger, line 1795: I would turn this into `basic-min-disturbance` to match other CLI options in VTR.

AlexandreSinger: Also I would reword this so it does not say "identically". Instead, I would say something more indicative that we are trying to get as close as we can. See my other comment above, your description should be what it will do, not necessarily what it does.

AlexandreSinger (changes requested):
Hi @haydar-c , this PR is looking good. Just some comments.

vpr/src/analytical_place/analytical_placement_flow.cpp
```diff
@@ -58,6 +58,75 @@ static void print_ap_netlist_stats(const APNetlist& netlist) {
     VTR_LOG("\n");
 }
 
+/**
+ * @brief Passes the flat placement information to a provided partial placement.
+ *
+ *  @param flat_placement_info    The flat placement information to be read.
+ *  @param ap_netlist             The APNetlist that used to iterate over its blocks.
+ *  @param prepacker              The Prepacker to get molecule of blocks in the ap_netlist.
+ *  @param p_placement            The partial placement to be updated which is assumend
```
vaughnbetz, line 67: same blocks -> the same blocks

vpr/src/analytical_place/analytical_placement_flow.cpp
```diff
@@ -58,6 +58,75 @@ static void print_ap_netlist_stats(const APNetlist& netlist) {
...
+        bool found_valid_atom = false;
+        for (AtomBlockId atom_blk_id: mol.atom_block_ids) {
+            if (!atom_blk_id.is_valid())
+                continue;
+            const t_pl_loc current_loc(flat_placement_info.blk_x_pos[atom_blk_id],
+                                       flat_placement_info.blk_y_pos[atom_blk_id],
+                                       flat_placement_info.blk_sub_tile[atom_blk_id],
+                                       flat_placement_info.blk_layer[atom_blk_id]);
+            if (found_valid_atom) {
+                VTR_ASSERT_MSG(current_loc == atom_loc, "Verify all atoms of a molecule provided in the flat placement share same location.\n");
```
vaughnbetz, line 87: same location -> the same location

vaughnbetz: I'd give more information, and say what the error is.
"Molecule " mol_id "contains atom %s (%d) with a location of " atom_loc "that conflicts with the location of other atoms in this molecule of " current_loc "."

Probably we should make this a VTR_ERROR (more user-friendly).

vpr/src/analytical_place/ap_flow_enums.h
```diff
@@ -15,6 +15,7 @@
  */
 enum class e_ap_full_legalizer {
     Naive,  ///< The Naive Full Legalizer, which clusters atoms placed in the same tile and tries to place them in that tile according to the flat placement.
-    APPack  ///< The APPack Full Legalizer, which uses the flat placement to improve the Packer and Placer.
+    APPack,  ///< The APPack Full Legalizer, which uses the flat placement to improve the Packer and Placer.
+    Basic_Min_Disturbance ///< The Basic Min. Disturbance Full Legalizer, which tries to reconstruct a clustered placement as close to the incoming flat placement as it can.
```
vaughnbetz, line 19: as close -> that is as close

vpr/src/base/read_options.cpp
```diff
@@ -1787,7 +1791,8 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .help(
             "Controls which Full Legalizer to use in the AP Flow.\n"
             " * naive: Use a Naive Full Legalizer which will try to create clusters exactly where their atoms are placed.\n"
-            " * appack: Use APPack, which takes the Packer in VPR and uses the flat atom placement to create better clusters.")
+            " * appack: Use APPack, which takes the Packer in VPR and uses the flat atom placement to create better clusters.\n"
+            " * basic-min-disturbance: Use the Basic Min. Disturbance Full Legalizer which tries to reconstruct a clustered placement as close to the incoming flat placement as it can.\n")
```
vaughnbetz, line 1795: as close -> that is as close to the incoming flat placement as possible.

vaughnbetz (changes requested):
A few nits -- otherwise good!

vpr/src/base/read_options.cpp
```diff
@@ -141,6 +141,8 @@ struct ParseAPFullLegalizer {
             conv_value.set_value(e_ap_full_legalizer::Naive);
         else if (str == "appack")
             conv_value.set_value(e_ap_full_legalizer::APPack);
+        else if (str == "basic_min_disturbance")
```
AlexandreSinger, line 144: You changed this to dashes in the help comment. I think you also need to change it here.

`basic-min-disturbance`

vpr/src/base/read_options.cpp
```diff
@@ -158,14 +160,16 @@ struct ParseAPFullLegalizer {
             case e_ap_full_legalizer::APPack:
                 conv_value.set_value("appack");
                 break;
+            case e_ap_full_legalizer::Basic_Min_Disturbance:
+                conv_value.set_value("basic_min_disturbance");
```
AlexandreSinger, line 164: Same comment as above. These should use dashes.

vpr/src/analytical_place/analytical_placement_flow.cpp
```diff
@@ -58,6 +58,81 @@ static void print_ap_netlist_stats(const APNetlist& netlist) {
...
+        PackMoleculeId mol_id = ap_netlist.block_molecule(ap_blk_id);
+        const t_pack_molecule& mol = prepacker.get_molecule(mol_id);
+        // Get location of a valid atom in the molecule and verify that
+        // all atoms of the molecule share same placement information.
+        t_pl_loc atom_loc;
+        bool found_valid_atom = false;
+        for (AtomBlockId atom_blk_id: mol.atom_block_ids) {
+            if (!atom_blk_id.is_valid())
+                continue;
+            const t_pl_loc current_loc(flat_placement_info.blk_x_pos[atom_blk_id],
```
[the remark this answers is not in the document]
AlexandreSinger: t_pl_loc is used for the clustered placement, which is why it uses integers (clusters can only be placed on grid coordinates).

AlexandreSinger (changes requested):
Hi @haydar-c , last few comments.

The most major one is the use of t_pl_loc in the AP context. We have to be careful when using structs designed for clustered placement. This struct, for example, stores position as integers; however, in AP we store positions as floats.

AlexandreSinger (approved):
LGTM

PR #2924, opened 2025-03-06
Add a NestedNetlistRouter to enable integration with fine grained parallel router

duck2:
TBB's thread pool doesn't let us block in tasks, so we have to implement a custom thread pool if we want to integrate a parallel netlist router with a parallel connection router.

### DIFF
--- a/libs/libvtrutil/src/vtr_thread_pool.h
+++ b/libs/libvtrutil/src/vtr_thread_pool.h
@@ -1,0 +1,159 @@
+#pragma once
+
+/** 
+ * @file vtr_thread_pool.h
+ * @brief A generic thread pool for parallel task execution
+ */
+
+#include <thread>
+#include <queue>
+#include <mutex>
+#include <condition_variable>
+#include <memory>
+#include <atomic>
+#include <functional>
+#include <cstddef>
+#include <vector>
+#include "vtr_log.h"
+#include "vtr_time.h"
+
+namespace vtr {
+
+/**
+ * A thread pool for parallel task execution. It is a naive
+ * implementation which uses a queue for each thread and assigns
+ * tasks in a round robin fashion.
+ *
+ * Example usage:
+ *
+ * vtr::thread_pool pool(4);
+ * pool.schedule_work([]{
+ *     // Task body
+ * });
+ * pool.wait_for_all(); // There's no API to wait for a single task
+ */
+class thread_pool {
+  private:
+    /* Thread-local data */
+    struct ThreadData {
+        std::thread thread;
+        /* Per-thread task queue */
+        std::queue<std::function<void()>> task_queue;
+
+        /* Threads wait on cv for a stop signal or a new task
+         * queue_mutex is required for condition variable */
+        std::mutex queue_mutex;
+        std::condition_variable cv;
+        bool stop = false;
+    };
+
+    /* Container for thread-local data */
+    std::vector<std::unique_ptr<ThreadData>> threads;
+    /* Used for round-robin scheduling */
+    std::atomic<size_t> next_thread{0};
+    /* Used for wait_for_all */
+    std::atomic<size_t> active_tasks{0};
+
+    /* Condition variable for wait_for_all */
+    std::mutex completion_mutex;
+    std::condition_variable completion_cv;
+
+  public:
+    thread_pool(size_t thread_count) {
+        threads.reserve(thread_count);
+
+        for (size_t i = 0; i < thread_count; i++) {
+            auto thread_data = std::make_unique<ThreadData>();
+
+            thread_data->thread = std::thread([&]() {
+                ThreadData* td = thread_data.get();
+
+                while (true) {
+                    std::function<void()> task;
+
+                    { /* Wait until a task is available or stop signal is received */
+                        std::unique_lock<std::mutex> lock(td->queue_mutex);
+
+                        td->cv.wait(lock, [td]() {
+                            return td->stop || !td->task_queue.empty();
+                        });
+
+                        if (td->stop && td->task_queue.empty()) {
+                            return;
+                        }
+
+                        /* Fetch a task from the queue */
+                        task = std::move(td->task_queue.front());
+                        td->task_queue.pop();
+                    }
+
+                    vtr::Timer task_timer;
+                    task();
+                }
+            });
+
+            threads.push_back(std::move(thread_data));
+        }
+    }
+
+    template<typename F>
+    void schedule_work(F&& f) {
+        active_tasks++;
+
+        /* Round-robin thread assignment */
+        size_t thread_idx = (next_thread++) % threads.size();
+        auto thread_data = threads[thread_idx].get();
+
+        auto task = [this, f = std::forward<F>(f)]() {
+            vtr::Timer task_timer;
+
+            try {
+                f();
+            } catch (const std::exception& e) {
+                VTR_LOG_ERROR("Thread %zu failed task with error: %s\n",
+                              std::this_thread::get_id(), e.what());
+                throw;
+            } catch (...) {
+                VTR_LOG_ERROR("Thread %zu failed task with unknown error\n",
+                              std::this_thread::get_id());
+                throw;
+            }
+
+            size_t remaining = --active_tasks;
+            if (remaining == 0) {
+                completion_cv.notify_all();
+            }
+        };
+
+        /* Queue new task */
+        {
+            std::lock_guard<std::mutex> lock(thread_data->queue_mutex);
+            thread_data->task_queue.push(std::move(task));
+        }
+        thread_data->cv.notify_one();
+    }
+
+    void wait_for_all() {
+        std::unique_lock<std::mutex> lock(completion_mutex);
+        completion_cv.wait(lock, [this]() { return active_tasks == 0; });
+    }
+
+    ~thread_pool() {
+        /* Stop all threads */
+        for (auto& thread_data : threads) {
+            {
+                std::lock_guard<std::mutex> lock(thread_data->queue_mutex);
+                thread_data->stop = true;
+            }
+            thread_data->cv.notify_one();
+        }
+
+        for (auto& thread_data : threads) {
+            if (thread_data->thread.joinable()) {
+                thread_data->thread.join();
+            }
+        }
+    }
+};
+
+} // namespace vtr
--- a/vpr/src/base/ShowSetup.cpp
+++ b/vpr/src/base/ShowSetup.cpp
@@ -252,6 +252,9 @@ static void ShowRouterOpts(const t_router_opts& RouterOpts) {
 
     VTR_LOG("RouterOpts.router_algorithm: ");
     switch (RouterOpts.router_algorithm) {
+        case NESTED:
+            VTR_LOG("NESTED\n");
+            break;
         case PARALLEL:
             VTR_LOG("PARALLEL\n");
             break;
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -282,7 +282,9 @@ struct ParseRoutePredictor {
 struct ParseRouterAlgorithm {
     ConvertedValue<e_router_algorithm> from_str(const std::string& str) {
         ConvertedValue<e_router_algorithm> conv_value;
-        if (str == "parallel")
+        if (str == "nested")
+            conv_value.set_value(NESTED);
+        else if (str == "parallel")
             conv_value.set_value(PARALLEL);
         else if (str == "parallel_decomp")
             conv_value.set_value(PARALLEL_DECOMP);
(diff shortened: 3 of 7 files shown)


libs/libvtrutil/src/vtr_thread_pool.h
```diff
@@ -0,0 +1,171 @@
...
+#include <functional>
+#include <cstddef>
+#include <vector>
+#include "vtr_log.h"
+#include "vtr_time.h"
+
+namespace vtr {
+
+class thread_pool {
+private:
```
vaughnbetz, line 23: Comment every member variable with doxygen and the overall class.  I'd give an example of how to use the class in the overall comment.
This should also go into the vtr utility documentation as a new tab eventually.

libs/libvtrutil/src/vtr_thread_pool.h
```diff
@@ -0,0 +1,171 @@
...
+    struct ThreadData {
+        std::thread thread;
+        std::queue<std::function<void()>> task_queue;
+        std::mutex queue_mutex;
+        std::condition_variable cv;
+        bool stop = false;
+        size_t thread_id;  // For debugging
+        size_t tasks_completed{0};  // Track number of completed tasks
```
vaughnbetz: For logging, or something else?

libs/libvtrutil/src/vtr_thread_pool.h
```diff
@@ -0,0 +1,171 @@
...
+
+namespace vtr {
+
+class thread_pool {
+private:
+    struct ThreadData {
+        std::thread thread;
+        std::queue<std::function<void()>> task_queue;
+        std::mutex queue_mutex;
+        std::condition_variable cv;
```
vaughnbetz, line 28: Shouldn't the name give an idea of what the condition variable controls/is used for, rather than just cv?

libs/libvtrutil/src/vtr_thread_pool.h
```diff
@@ -0,0 +1,171 @@
...
+
+        ThreadData(size_t id) : thread_id(id) {}
+    };
+
+    std::vector<std::unique_ptr<ThreadData>> threads;
+    std::atomic<size_t> next_thread{0};  // For round-robin assignment
+    std::atomic<size_t> total_tasks_queued{0};
+    vtr::Timer pool_timer;  // Track pool lifetime
+    std::atomic<size_t> active_tasks{0};
+    std::mutex completion_mutex;
```
vaughnbetz, line 41: Comment these.  (all the variables).  
From our discussion: completion_mutex is to wait for all the tasks in the queue to finish.

libs/libvtrutil/src/vtr_thread_pool.h
```diff
@@ -0,0 +1,171 @@
...
+    std::condition_variable completion_cv;
+
+public:
+    thread_pool(size_t thread_count) {
```
vaughnbetz, line 45: Between the class overview and here, you need to comment how this thread pool works.

libs/libvtrutil/src/vtr_thread_pool.h
```diff
@@ -0,0 +1,171 @@
...
+            thread_data->task_queue.push(std::move(task));
+            // VTR_LOG("Task %zu queued to thread %zu (queue size: %zu)\n", 
+            //        task_id, thread_data->thread_id, thread_data->task_queue.size());
+        }
+        thread_data->cv.notify_one();
+    }
+
+    void wait_for_all() {
+        std::unique_lock<std::mutex> lock(completion_mutex);
+        completion_cv.wait(lock, [this]() { return active_tasks == 0; });
```
vaughnbetz, line 120: why capture [this] ?  Seems it's not needed.

libs/libvtrutil/src/vtr_thread_pool.h
```diff
@@ -0,0 +1,171 @@
...
+            if (thread_data->thread.joinable()) {
+                thread_data->thread.join();
+                // VTR_LOG("Thread %zu joined after completing %zu tasks\n", 
+                //        thread_data->thread_id, thread_data->tasks_completed);
+            }
+        }
+    }
+};
+
+}  // namespace vtr
```
vaughnbetz, line 141: Also you tell me you have some dead variables / experimental code to delete here. If you no longer need it, it would be good to delete before the PR merge.

vpr/src/base/ShowSetup.cpp
```diff
@@ -254,6 +254,9 @@ static void ShowRouterOpts(const t_router_opts& RouterOpts) {
 
     VTR_LOG("RouterOpts.router_algorithm: ");
     switch (RouterOpts.router_algorithm) {
+        case MIXED:
```
vaughnbetz, line 257: NESTED_PARALLEL would be a better name.

vpr/src/base/read_options.cpp
```diff
@@ -171,7 +171,9 @@ struct ParseRoutePredictor {
 struct ParseRouterAlgorithm {
     ConvertedValue<e_router_algorithm> from_str(const std::string& str) {
         ConvertedValue<e_router_algorithm> conv_value;
-        if (str == "parallel")
+        if (str == "mixed")
```
vaughnbetz, line 174: nested_parallel better.

vpr/src/base/read_options.cpp
```diff
@@ -2405,7 +2411,7 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
             " * parallel: timing_driven with nets in different regions of the chip routed in parallel\n"
             " * parallel_decomp: timing_driven with additional parallelism obtained by decomposing high-fanout nets, possibly reducing quality\n")
         .default_value("timing_driven")
-        .choices({"parallel", "parallel_decomp", "timing_driven"})
+        .choices({"mixed","parallel", "parallel_decomp", "timing_driven"})
```
vaughnbetz, line 2414: nested_parallel

vpr/src/base/vpr_types.h
```diff
@@ -1186,6 +1186,7 @@ struct t_ap_opts {
  * read_rr_graph_name:  stores the file name of the rr graph to be read by vpr */
 
 enum e_router_algorithm {
+    MIXED,
```
vaughnbetz, line 1189: NESTED_PARALLEL

vpr/src/route/MixedNetlistRouter.h
```diff
@@ -0,0 +1,121 @@
+#pragma once
+
+/** @file Mixed case for NetlistRouter */
```
vaughnbetz, line 3: Explain more, say nested parallelism, explain what that is.

vaughnbetz: Explain that it avoids using TBB so it can devote some threads to coarse-grained (net- or part of net) parallelism and some to fine-grained (wave expansion) paralellism.

vpr/src/route/MixedNetlistRouter.h
```diff
@@ -0,0 +1,121 @@
...
+/* Add cmd line option for this later */
+constexpr int MAX_THREADS = 4;
+
+/** Mixed impl for NetlistRouter.
+ * Holds enough context members to glue together ConnectionRouter and net routing functions,
+ * such as \ref route_net. Keeps the members in thread-local storage where needed,
+ * i.e. ConnectionRouters and RouteIterResults-es.
+ * See \ref route_net. */
+template<typename HeapType>
+class MixedNetlistRouter : public NetlistRouter {
```
vaughnbetz, line 18: MixedNetlistRouter -> NestedParallelNetlistRouter (or NestedPNetlistRouter)

vpr/src/route/MixedNetlistRouter.h
```diff
@@ -0,0 +1,121 @@
...
+
+    /** Cached routing parameters for current iteration (inputs to \see route_netlist()) */
+    int _itry;
+    float _pres_fac;
+    float _worst_neg_slack;
+
+    /** The partition tree. Holds the groups of nets for each partition */
+    vtr::optional<PartitionTree> _tree;
+
+    vtr::thread_pool _thread_pool;
```
vaughnbetz, line 98: Comment the thread_pool (can be brief, but say it's not TBB so we can split across two levels of parallelism).

vpr/src/route/MixedNetlistRouter.h
```diff
@@ -0,0 +1,121 @@
...
+
+    /* Thread-local storage */
```
vaughnbetz, line 100: Explain this a bit more --> a connection router and the results of each routing iteration can be looked up by thread id.  storage_mutex protects both these data structures in case a thread is inserting data, which could reallocate the unordered_map.  This could be made more efficient for the case where everyone is only reading, but that probably isn't worthwhile as these lookups aren't used often.

vpr/src/route/MixedNetlistRouter.tpp
```diff
@@ -0,0 +1,141 @@
+#pragma once
+
+/** @file Impls for ParallelNetlistRouter */
```
vaughnbetz, line 3: NestedPNetlistRouter

vpr/src/route/MixedNetlistRouter.h
```diff
@@ -0,0 +1,121 @@
...
+        , _thread_pool(MAX_THREADS) {}
+    ~MixedNetlistRouter() {}
+
+    /** Run a single iteration of netlist routing for this->_net_list. This usually means calling
+     * \ref route_net for each net, which will handle other global updates.
+     * \return RouteIterResults for this iteration. */
+    RouteIterResults route_netlist(int itry, float pres_fac, float worst_neg_slack);
+    /** Inform the PartitionTree of the nets with updated bounding boxes */
+    void handle_bb_updated_nets(const std::vector<ParentNetId>& nets);
+    void set_rcv_enabled(bool x);
```
vaughnbetz, line 54: Put a comment on set_rcv_enabled that it sets this control variable for each thread (in its local storage).

vpr/src/route/MixedNetlistRouter.h
```diff
@@ -0,0 +1,121 @@
...
+    void set_timing_info(std::shared_ptr<SetupHoldTimingInfo> timing_info);
+
```
vaughnbetz, line 56: Same for set_timing_info --> explain it does this for all threads.

vpr/src/route/netlist_routers.h
```diff
@@ -71,6 +71,7 @@ class NetlistRouter {
 
 /* Include the derived classes here to get the HeapType-templated impls */
 #include "SerialNetlistRouter.h"
+#include "MixedNetlistRouter.h"
```
vaughnbetz, line 74: NestedPNetlistRouter

vpr/src/route/netlist_routers.h
```diff
@@ -104,6 +105,20 @@ inline std::unique_ptr<NetlistRouter> make_netlist_router_with_heap(
             routing_predictor,
             choking_spots,
             is_flat);
+    } else if (router_opts.router_algorithm == e_router_algorithm::MIXED) {
```
vaughnbetz, line 108: NESTED_PARALLEL

vaughnbetz (changes requested):
Some commenting and renaming suggestions.
Also add QoR data to the PR.

libs/libvtrutil/src/vtr_thread_pool.h
CimboAkinci: looks like we do need to capture this since active_tasks is a member

duck2: Looks like we do need to capture this since `active_tasks` is a class member

vpr/src/base/ShowSetup.cpp
duck2: changed to NESTED

libs/libvtrutil/src/vtr_thread_pool.h
vaughnb-cerebras: OK.

vaughnb-cerebras:
Looks good. We can merge once the commenting and dead code feedback is addressed / resolved.

vpr/src/route/netlist_routers.h
duck2: changed

vpr/src/route/netlist_routers.h
duck2: changed

vpr/src/base/read_options.cpp
duck2: changed

vpr/src/base/read_options.cpp
duck2: changed

vaughnbetz (approved):
Thanks for the additional commenting.  I think you should add per function doxygen comments to the thread pool though (there are still some member functions with no comment).

PR #2926, opened 2025-03-06
Overused Node Report

amin1377:
Fix a bug in the overused node report where an invalid node ID was accessed.


vpr/src/route/overuse_report.cpp
```diff
@@ -470,7 +478,12 @@ void print_block_pins_nets(std::ostream& os,
     for (int pin = pin_num_range.low; pin <= pin_num_range.high; pin++) {
         t_rr_type rr_type = (get_pin_type_from_pin_physical_num(physical_type, pin) == DRIVER) ? t_rr_type::OPIN : t_rr_type::IPIN;
         RRNodeId node_id = get_pin_rr_node_id(rr_graph.node_lookup(), physical_type, layer, root_x, root_y, pin);
-        VTR_ASSERT(node_id != RRNodeId::INVALID());
+        // When flat router is enabled, RR Node chains collapse into a single node. Thus, when 
+        // looking up the RR Node ID, it may return an invalid node ID. In this case, we skip
+        // this pin.
+        if (!pin_on_tile && node_id == RRNodeId::INVALID()) {
```
AlexandreSinger, line 484: Because you are checking if the pin is not on the tile and the node ID is invalid, it _may_ be possible that the pin is on the tile and the node ID is invalid. In that case, do we want an assert to ensure that never happens?

In other words, shouldn't it be this:
```cpp
if (node_id == RRNodeId::INVALID()) {
    VTR_ASSERT(!pin_on_tile);
    continue;
}
```

My thinking is that we may be leaving the door open for an invalid node to be used below, which may be tricky to debug later (unless this is something you expect; in which case, ignore this comment).

AlexandreSinger (commented):
LGTM. Just had one minor comment.

AlexandreSinger (approved):
LGTM

PR #2929, opened 2025-03-12
Fixed ignored y channels in computing routing channel utilization histogram.

soheilshahrouz:
This PR fixes a bug that ignored y channels when computing channel utilization histogram and maximum channel utilization.


amin1377 (approved):
Good catch!

PR #2930, opened 2025-03-12
Override edge attributes in RR graph

soheilshahrouz:
Adds a new command line option to override attributes of specific edges in the RR graph.

#### Description
The user can pass a text file using --read_rr_edge_override, which modifies the delay of selected edges. The expected file format:
```
# edge    Tdel   
64812     5.9e-11 
9981      4.2e-11   
1234      7.1e-11   
4321      9.4e-11
(42, 64)  7.3e-11
```
The first line must start with `#` and is always ignored by the parser.
Each following line must include an RR edge ID and delay (Tdel). Other attributes (resistance, capacitances) are optional. Edges can also be specified using source/sink node pairs.

By manual inspection of generated RR graph files to see if the edge attributes are overridden correctly.

#### Types of changes

#### Checklist:


soheilshahrouz:
@petergrossmann21 I'd apprecite any feedback on this

libs/librrgraph/src/base/rr_graph_builder.h
```diff
@@ -264,6 +264,10 @@ class RRGraphBuilder {
         node_storage_.alloc_and_load_edges(rr_edges_to_create);
     }
 
+    inline void override_edge_switch(RREdgeId edge_id, RRSwitchId switch_id) {
```
amin1377, line 267: To maintain consistency with other functions, I think it would be better to add a Doxygen comment. (just add a brief description)

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -731,6 +730,8 @@ class t_rr_graph_storage {
      */
     void partition_edges(const vtr::vector<RRSwitchId, t_rr_switch_inf>& rr_switches);
 
+    void override_edge_switch(RREdgeId edge_id, RRSwitchId switch_id);
```
amin1377, line 733: DoxyGen comment (@brief only) similar to other functions

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -413,6 +413,10 @@ class RRGraphView {
         return node_storage_.edge_switch(id, iedge);
     }
 
+    inline short edge_switch(RREdgeId id) const {
```
amin1377, line 416: DoxyGen

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +134,140 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+        RRNodeId sink_node_id = RRNodeId(second);
+
+        for (RREdgeId outgoing_edge_id : rr_graph.rr_nodes().edge_range(src_node_id)) {
+            if (rr_graph.rr_nodes().edge_sink_node(outgoing_edge_id) == sink_node_id) {
+                edge_id = outgoing_edge_id;
+                break;
+            }
+        }
+
+        if (!edge_id.is_valid()) {
```
amin1377, line 165: To make it more concise, I suggest using VTR_LOGV_ERROR(expr, ...).

libs/librrgraph/src/io/rr_graph_reader.h
```diff
@@ -28,11 +28,15 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
                   const char* read_rr_graph_name,
-                  std::string* read_rr_graph_filename,
+                  std::string* loaded_rr_graph_filename,
                   bool read_edge_metadata,
                   bool do_check_rr_graph,
                   bool echo_enabled,
                   const char* echo_file_name,
                   bool is_flat);
 
+void load_rr_edge_overrides(std::string_view filename,
```
amin1377, line 38: Doxygen comment

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +134,140 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+               lhs.intra_tile == rhs.intra_tile &&
+               lhs.type() == rhs.type();
+    }
+};
+
+void load_rr_edge_overrides(std::string_view filename,
+                            RRGraphBuilder& rr_graph_builder,
+                            const RRGraphView& rr_graph) {
+    std::ifstream file(filename.data());
+    if (!file) {
```
amin1377, line 227: VTR_LOGV_ERROR

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +134,140 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+
+    float value;
+    while (iss >> value) {
+        overridden_values.push_back(value);
+    }
+
+    return edge_id;
+}
+
+struct t_rr_switch_inf_hash {
```
amin1377, line 181: I think a better way to implement a hash function for `t_rr_switch_inf` is to define the hash function within the struct itself. This ensures that if someone adds a new field to the data structure, they can update the hash function directly in the same place.  

Additionally, you can add the following specialization in the same file:  
```cpp
namespace std {
    template <>
    struct hash<t_rr_switch_inf> {
        size_t operator()(const t_rr_switch_inf& s) const {
            return s.hash();
        }
    };
}
```
An added benefit of this approach is that you wouldn't need to explicitly pass the hash function when using data structures like `std::unordered_map` or `std::unordered_set`, as it will be automatically recognized.

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +134,140 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+        hash_combine(static_cast<int>(s.power_buffer_type));
+        hash_combine(s.power_buffer_size);
+        hash_combine(s.intra_tile);
+        hash_combine(static_cast<int>(s.type()));
+
+        return seed;
+    }
+};
+
+struct t_rr_switch_inf_equal {
```
amin1377, line 207: To ensure that any additional fields added to this data structure are properly handled, I suggest overloading the operator within the struct itself.

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +134,140 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+
+void load_rr_edge_overrides(std::string_view filename,
+                            RRGraphBuilder& rr_graph_builder,
+                            const RRGraphView& rr_graph) {
+    std::ifstream file(filename.data());
+    if (!file) {
+        VTR_LOG_ERROR("Failed to open the RR edge override file: %s\n", filename.data());
+    }
+
+    std::unordered_map<t_rr_switch_inf, RRSwitchId, t_rr_switch_inf_hash, t_rr_switch_inf_equal> unique_switch_info;
```
amin1377, line 231: I also needed to do something similar when creating edges while collapsing RR Node chains in the flat router. If you could create an issue to remind me to implement a function for this, that would be helpful.

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +134,140 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+            switch_override_info.Tdel = (overridden_values.size() >= 1) ? overridden_values[0] : switch_override_info.Tdel;
+            switch_override_info.R = (overridden_values.size() >= 2) ? overridden_values[1] : switch_override_info.R;
+            switch_override_info.Cin = (overridden_values.size() >= 3) ? overridden_values[2] : switch_override_info.Cin;
+            switch_override_info.Cout = (overridden_values.size() >= 4) ? overridden_values[3] : switch_override_info.Cout;
+            switch_override_info.Cinternal = (overridden_values.size() >= 5) ? overridden_values[4] : switch_override_info.Cinternal;
+
+            RRSwitchId new_switch_id;
+            auto it = unique_switch_info.find(switch_override_info);
+            if (it == unique_switch_info.end()) {
+                new_switch_id = rr_graph_builder.add_rr_switch(switch_override_info);
```
amin1377, line 265: Now that I think about it, I'm pretty sure I wrote a function that retrieves switch info, searches the RR Node switches for one with similar characteristics, and, if none is found, creates a new one and returns the corresponding RR switch ID (I think the name was something like get_or_create_rr_switch_id). In the issue I mentioned, if you could reference this function, I'll update this part of the code myself.

amin1377 (changes requested):
Overall, this looks good to me. I’ve added several small comments, mainly related to adding comments. :)

One more thing—I think it would be good to add a test for this PR. As we discussed, testing this requires additional functionality beyond what the current infrastructure provides. I suggest creating an issue to track it, and we can implement it in a separate PR.

petergrossmann21:
@soheilshahrouz Some initial thoughts:

* I think it might be beneficial to separate RC parameters from Tdel parameters.  While it's nice to lay out all the parameters together, I don't think per-edge RC annotation makes sense.  R and C are node properties rather than edge properties (see https://docs.verilogtorouting.org/en/latest/vpr/file_formats/#routing-resource-graph-file-format-xml) .  So even if both parameter types are specified in a single file, I would recommend capturing them on separate lines of text, each with their own syntax.  Separate files would also be fine, if slightly less elegant (see aside below)
* The (src_node #, sink_node #) notation is definitely useful, as it addresses scenarios where the generator of the file does not have integer IDs for edges available during delay annotation (this happens to be the case for me, but is more generically useful because the node IDs can be reliably associated with the nets in the FPGA circuit that are used in delay capture, whereas I think edge ID associations are less reliable).
* A quick sanity check on file format size:  in the general case, there will be some number of bytes per edge to specify the edge ID and delay parameters.  I currently develop relatively small, routing-rich architectures and am typically working with node counts at O(10K) and edge counts at O(100K).  If 6 digits of precision are supplied per delay parameter, plus five characters to format in scientific notation ('.', 'e-XX'), then each delay parameter formatted in (src_node #, sink_node #) notation uses up to 27 bytes (5 for each node ID, 5 more to format them and space them apart from the delay, and then 11 for the delay constant and one for the carriage return).  For a 500K-edge graph that works out to 13.5 MB -- not small, but not unworkable.  This supports an architecture with O(1K) LUTs.  for something with O(100K) LUTs, we would expect the file to be 100X larger, i.e. O(1GB).

I think the file size analysis above is sufficient to motivate implementing some method of shorthand in the file format so that the architect can express when multiple edges have the same delay parameter.  My first instinct is that some kind of relationship between the physical tile definitions and the graph edges can help with this.  Currently, the way that I do that is by relating switches in the `<switchlist>` section of the architecture XML to different physical tile types in my XML generator.

An aside:  Note that in automatic place and route flows for ASICs, the R and C values are either not generated at all (because the wire delay is lumped with the "intrinsic" delay of the logic gates and so the VPR model will set all the RC values to 0) OR the intrinsic delay is captured independently of the RC values and thus it's convenient to annotate them separately anyway.

petergrossmann21:
@soheilshahrouz one additional thought:  if the delay annotation is to be performed as a standalone step, having a dedicated executable utility to run it will be helpful--this will avoid having to call vpr for this step and having to pass in an arbitrary blif file.

libs/librrgraph/src/io/rr_graph_reader.cpp
vaughnbetz: I kind of like @amin1377's suggestion, although @soheilshahrouz tells me he's less keen.  I like the simplicity of having a hash function that gets used automatically where appropriate.

libs/libarchfpga/src/physical_types.cpp
```diff
@@ -71,6 +71,42 @@ bool t_rr_switch_inf::configurable() const {
...
+           && mux_trans_size == other.mux_trans_size
+           && buf_size == other.buf_size
+           && power_buffer_type == other.power_buffer_type
+           && power_buffer_size == other.power_buffer_size
+           && intra_tile == other.intra_tile
+           && type() == other.type();
+}
+
+std::size_t t_rr_switch_inf::Hasher::operator()(const t_rr_switch_inf& s) const {
+    std::size_t seed = 0;
```
vaughnbetz, line 89: hash_val seems like a better name

libs/libarchfpga/src/physical_types.h
```diff
@@ -1902,16 +1917,22 @@ struct t_rr_switch_inf {
...
+    /// separate DC-connected subcircuits
     bool buffered() const;
 
-    //Returns true if this switch type is configurable
+    /// Returns true if this switch type is configurable
     bool configurable() const;
 
+    bool operator==(const t_rr_switch_inf& other) const;
+
+    struct Hasher {
```
vaughnbetz, line 1932: Add Doxygen comment.

libs/librrgraph/src/base/rr_graph_builder.h
```diff
@@ -264,6 +264,11 @@ class RRGraphBuilder {
         node_storage_.alloc_and_load_edges(rr_edges_to_create);
     }
 
+    /** @brief Overrides the associated switch for a given edge. */
```
vaughnbetz, line 267: add (update the edge to use the passed in switch)

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -393,13 +393,27 @@ class t_rr_graph_storage {
...
-    RREdgeId edge_id(const RRNodeId& id, t_edge_size iedge) const {
+    RREdgeId edge_id(RRNodeId id, t_edge_size iedge) const {
         RREdgeId first_edge = this->first_edge(id);
         RREdgeId ret(size_t(first_edge) + iedge);
         VTR_ASSERT_SAFE(ret < last_edge(id));
         return ret;
     }
 
+    /**
+     * @brief Retrieve the RREdgeId that connect the given source and sink nodes.
```
vaughnbetz, line 404: connect -> connects

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -731,6 +744,9 @@ class t_rr_graph_storage {
      */
     void partition_edges(const vtr::vector<RRSwitchId, t_rr_switch_inf>& rr_switches);
 
+    /** @brief Overrides the associated switch for a given edge. */
```
vaughnbetz, line 747: Add a few words saying what this does (sets the switch index)

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -29,6 +30,22 @@
...
 
+/**
+ * @brief Parses a line from the RR edge delay override file.
+ *
+ * @details Expected formats:
+ *          edge_id Tdel
+ *          (source_node_id, sink_node_id) Tdel
+ *
+ * @param line The line to parse.
+ * @param overridden_Tdel Parsed override delay.
```
vaughnbetz, line 41: Comment that you return overridden_Tdel through this reference parameter.  I'd also probably make it the last argument in that case.

Or just return a tuple.

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +133,85 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+                                         const RRGraphView& rr_graph) {
+    std::istringstream iss(line);
+    char ch;
+    RREdgeId edge_id;
+
+    if (std::isdigit(line[0])) {
+        // Line starts with an integer
+        int first;
+        iss >> first;
+        edge_id = (RREdgeId)first;
```
vaughnbetz, line 148: Try all sorts of broken input, make sure you get an error message.
Ideally, you'd accept comment characters of # to the end of the line, as our other file formats do.

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -115,3 +133,85 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
+        unique_switch_info.insert({sw, rr_sw_idx});
+    }
+
+    std::string line;
+    bool firstLine = true;
+
+    while (std::getline(file, line)) {
+        if (firstLine) {
+            if (line.empty() || line[0] != '#') {
+                VTR_LOG_ERROR("Error: First line must start with #\n");
```
vaughnbetz, line 192: I'd remove this and just throw anything after a # away (make it a comment character).

libs/librrgraph/src/io/rr_graph_reader.h
```diff
@@ -28,11 +28,26 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
...
                   bool echo_enabled,
                   const char* echo_file_name,
                   bool is_flat);
 
+/**
+ * @brief Reads a text file where the intrinsic delay of edges are overridden.
+ *
+ * @details This function tries to find a switch with the overridden delay. If such a
+ *          switch exists, the edge will point to it as its corresponding switch.
+ *          Otherwise, a new switch is created so that the edge can point to a valid switch.
```
vaughnbetz, line 43: I'd add a short description of why you would use this file.

vpr/src/base/read_options.cpp
```diff
@@ -1769,19 +1769,23 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
...
-            "The routing resource graph file to load."
-            " The loaded routing resource graph overrides any routing architecture specified in the architecture file.")
+        .help("The routing resource graph file to load. "
+              "The loaded routing resource graph overrides any routing architecture specified in the architecture file.")
         .metavar("RR_GRAPH_FILE")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    file_grp.add_argument(args.read_rr_edge_override_file, "--read_rr_edge_override")
+        .help(
+            "The routing resource edge attributes override file to load.")
```
vaughnbetz, line 1779: Explain why you can use this.

vpr/src/base/vpr_context.h
```diff
@@ -269,12 +269,13 @@ struct DeviceContext : public Context {
...
     t_clock_arch* clock_arch;
 
-    /**
-     * @brief Name of rrgraph file read (if any).
-     *
-     * Used to determine when reading rrgraph if file is already loaded.
-     */
-    std::string read_rr_graph_filename;
+    /// @brief Name of rrgraph file read (if any).
+    /// Used to determine if the file is already loaded when reading rrgraph.
```
vaughnbetz, line 273: "if the specified rr-graph file is already loaded, so we can avoid redundant reading of the rr-graph"

vpr/src/base/vpr_types.h
```diff
@@ -1308,59 +1308,52 @@ struct t_noc_opts {
...
- *   @param R_minW_nmos  Resistance (in Ohms) of a minimum width nmos transistor.
- *             Used only in the FPGA area model.
- *   @param R_minW_pmos  Resistance (in Ohms) of a minimum width pmos transistor.
- *   @param read_rr_graph_filename  File to read the RR graph from (overrides
- *             architecture)
- *   @param write_rr_graph_filename  File to write the RR graph to after generation
  */
 struct t_det_routing_arch {
-    enum e_directionality directionality; /* UDSD by AY */
+    /// Should the tracks be uni-directional or bi-directional? (UDSD by AY)
```
vaughnbetz, line 1313: Can remove UDSD by AY

vaughnbetz (changes requested):
LGTM, thanks!
I'd like to beef up the error-handling on the file input though. Try to make sure any invalid file you can detect gives an error message.

Also, you should update the documentation.
- File format (give the format, but also the motivation / use cases)
- Command line option

soheilshahrouz:
@vaughnbetz @amin1377
I applied your comments. Let me know if more changes are needed.

doc/src/vpr/file_formats.rst
```diff
@@ -1100,6 +1100,27 @@ To aid in handling large graphs, rr_graph files can also be :ref:`saved in <file
...
+    4321      9.4e-11
+    (42, 64)  7.3e-11
+
+.. _end:
+
+Lines starting with # are comments and ignored. Each other line should specify either: an edge ID and its new delay, or
+a source/sink node pair and its delay.
+
+This allows more accurate modeling of switch delays in the RR graph without creating many switch types
+in the architecture file and limiting them to small regions.
```
vaughnbetz, line 1123: I'd add: "This can be useful for more detailed modeling of a fabricated FPGA where layout differences lead to small delay differences in the same type of routing switch."

vaughnbetz (approved):
Looks good to me!  Just have one suggestion for an extra line in the file format documentation.

PR #2931, opened 2025-03-12
Partial update of the code base to be consistent with rules in .clang-format file

soheilshahrouz:
This PR patially updates the code base according to the `.clagn-format` file.

The current formatting rules remove empty lines at the end of file. I'll update `.clagn-format` after the format check CI test is brought back to fix this.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -68,10 +68,10 @@ IncludeIsMainRegex: '([-_](test|unittest))?$'
 IndentCaseLabels: true
 IndentWidth:     4
 IndentWrappedFunctionNames: false
-IndentPPDirectives: AfterHash
+IndentPPDirectives: None
 JavaScriptQuotes: Leave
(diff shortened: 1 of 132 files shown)


amin1377 (approved):
Reviewed up to cluster_placement.h and didn’t find any changes that affect functionality. The only issue is line width, which, as discussed, will be fixed later.

AmirhosseinPoolad (approved):
Reviewed from bottom of the changed files list to vpr/src/pack/cluster_util.cpp.

AlexandreSinger (approved):
LGTM, thanks @soheilshahrouz !

You read my mind about EOF lines. I am ok with removing them for this commit and adding them back after we add the CI test back in!

All of the other changes seem reasonable to me (obviously, I still think there should be a line limit; but I know I am not winning that fight lol).

PR #2932, opened 2025-03-12
Remove usage of atom to pb lookup from packing

AmirhosseinPoolad:
Adds new AtomPBLookUp class that holds the atom to/from pb lookup data in the cluster legalizer class. The packing stage now uses that instead of the global context. At the end of packing, AtomPBLookUp gets copied to the global context and everything goes on as normal.

To keep this invariant from being changed in the future, a lock was added to prevent any access to the atom pb data.


AmirhosseinPoolad:
On a kind of unrelated note: Why are there three different find_pb_graph_pin functions? In TimingGraphBuilder, vpr_utils.cpp and PreClusterDelayCalculator.

AmirhosseinPoolad:
@AlexandreSinger I want to run the nightly tests on this branch before merging, but meanwhile can you take a look at this? I'm done doing the obvious clean ups and fixes.

vpr/src/pack/atom_pb_bimap.cpp
```diff
@@ -0,0 +1,48 @@
+#include "atom_pb_bimap.h"
```
AlexandreSinger, line 1: Add file documentation for this file. See https://docs.verilogtorouting.org/en/latest/dev/code_documentation/#documenting-files

vpr/src/pack/atom_pb_bimap.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+AtomPBBimap::AtomPBBimap(const vtr::bimap<AtomBlockId, const t_pb*, vtr::linear_map, std::unordered_map>& atom_to_pb) {
+    atom_to_pb_ = atom_to_pb;
```
AlexandreSinger, line 4: This is another expensive copy. It may be necessary, but just something to keep an eye out for.

vpr/src/pack/atom_pb_bimap.cpp
```diff
@@ -0,0 +1,48 @@
...
+        //Found
+        return pb->pb_graph_node;
+    }
+    return nullptr;
+}
+
+void AtomPBBimap::set_atom_pb(const AtomBlockId blk_id, const t_pb* pb) {
+    //If either of blk_id or pb are not valid,
+    //remove any mapping
+
```
AlexandreSinger, line 37: I personally would remove this space. I think the comment would make more sense closer to the if.

vpr/src/pack/atom_pb_bimap.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
```
AlexandreSinger, line 1: Document the file. See: https://docs.verilogtorouting.org/en/latest/dev/code_documentation/#documenting-files

vpr/src/pack/atom_pb_bimap.h
```diff
@@ -0,0 +1,42 @@
...
+ * @brief Class that holds a bimap between atoms and pb types.
+ *        This means that you can get a pb from an atom and the
+ *        other way around.
+ *        
+ *        Used in the global AtomLookup context and in ClusterLegalizer
+ */
+class AtomPBBimap {
+    public:
+    AtomPBBimap() = default;
+    AtomPBBimap(const vtr::bimap<AtomBlockId, const t_pb*, vtr::linear_map, std::unordered_map> &atom_to_pb);
```
AlexandreSinger, line 18: I recommend creating an alias for this bimap. We use it in several places and it can really clean up that code!

vpr/src/pack/atom_pb_bimap.h
```diff
@@ -0,0 +1,42 @@
...
+
+
+    /**
+     * @brief Returns the leaf pb associated with the atom blk_id
+     * @note  this is the lowest level pb which corresponds directly to the atom block
+     */
+    const t_pb* atom_pb(const AtomBlockId blk_id) const;
+
+    ///@brief Returns the atom block id associated with pb
```
AlexandreSinger, line 27: I recommend updating this comment to say something about this needing to be a leaf pb.

vpr/src/pack/atom_pb_bimap.h
AlexandreSinger: You can even make the alias part of the class. So it is access as AtomPBBimap::bimap_impl or something

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -234,16 +236,16 @@ static bool check_cluster_noc_group(AtomBlockId atom_blk_id,
...
         return true;
     }
 
     // If the cluster belongs to a different NoC group than the atom's group,
     // they are incompatible.
     VTR_LOGV(log_verbosity > 3,
-             "\t\t\t NoC Group: Atom block %d failed NoC group check for cluster. Cluster's NoC group: %d, atom's NoC group: %d\n",
-             atom_blk_id, (size_t)cluster_noc_grp_id, (size_t)atom_noc_grp_id);
+            "\t\t\t NoC Group: Atom block %d failed NoC group check for cluster. Cluster's NoC group: %d, atom's NoC group: %d\n",
+            atom_blk_id, (size_t)cluster_noc_grp_id, (size_t)atom_noc_grp_id);
```
AlexandreSinger, line 248: I think what happened here is that you are conflicting with Soheil's changes. I think your rebase accidently undid soheils formatting cleanups. I recommend going through this file carefully and fixing these.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -258,7 +260,7 @@ static enum e_block_pack_status check_chain_root_placement_feasibility(const t_p
                                                                        const t_clustering_chain_info& clustering_chain_info,
                                                                        t_pack_patterns* mol_pack_patterns,
                                                                        const AtomBlockId blk_id) {
-    const AtomContext& atom_ctx = g_vpr_ctx.atom();
+    const AtomNetlist& netlist_context = g_vpr_ctx.atom().netlist();
```
AlexandreSinger, line 263: I would not call this netlist_context. I think atom_netlist would be better.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1624,13 +1627,14 @@ ClusterLegalizer::ClusterLegalizer(const AtomNetlist& atom_netlist,
...
-                                       high_fanout_thresholds,
-                                       atom_noc_grp_id_);
+                                    atom_netlist,
+                                    high_fanout_thresholds,
+                                    atom_noc_grp_id_);
     // Copy the options passed by the user
     cluster_legalization_strategy_ = cluster_legalization_strategy;
     enable_pin_feasibility_filter_ = enable_pin_feasibility_filter;
     log_verbosity_ = log_verbosity;
+    atom_pb_lookup_ = AtomPBBimap(g_vpr_ctx.atom().lookup().atom_pb_bimap());
```
AlexandreSinger, line 1637: I do not think this lookup should have anything in it at this point... If it does, that would be very strange. I think this line shouldn't be here and assert should be added to ensure the bimap is unset.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1624,13 +1627,14 @@ ClusterLegalizer::ClusterLegalizer(const AtomNetlist& atom_netlist,
...
 }
 
 void ClusterLegalizer::reset() {
```
AlexandreSinger, line 1640: For safety, I recommend adding an assert to ensure that after reset that the bimap is now "empty". Destroy cluster should be resetting this lookup, but it would be helpful for you to check this just in case.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1770,6 +1773,18 @@ void ClusterLegalizer::finalize() {
     }
 }
 
+bool ClusterLegalizer::is_atom_blk_in_cluster_block(const AtomBlockId blk_id, const AtomBlockId clustered_blk_id) const {
```
AlexandreSinger, line 1776: Why is this method needed? The cluster legalizer has a lookup from atom blocks to cluster blocks:
```
vtr::vector_map<AtomBlockId, LegalizationClusterId> atom_cluster_;
```

Also why does this take two atom blocks? Shouldn't this really be "are atoms in the same cluster?". If so, this can be made much much simpler...

vpr/src/pack/cluster_legalizer.h
```diff
@@ -22,12 +22,14 @@
...
 #include "vtr_vector.h"
 #include "vtr_vector_map.h"
+#include "atom_pb_bimap.h"
 
 // Forward declarations
 class Prepacker;
 class t_intra_cluster_placement_stats;
 class t_pb_graph_node;
 struct t_lb_router_data;
+class UserPlaceConstraints;
```
AlexandreSinger, line 32: Why do we need to forward declare this?

vpr/src/pack/cluster_legalizer.h
```diff
@@ -190,6 +192,7 @@ struct LegalizationCluster {
  *
  * // new_cluster_id now contains a fully legalized cluster.
  */
+
```
AlexandreSinger, line 195: Empty space should be removed.

vpr/src/pack/cluster_legalizer.h
```diff
@@ -513,6 +516,16 @@ class ClusterLegalizer {
         log_verbosity_ = verbosity;
     }
 
+    /*
+    * @brief Determine if atom block is in cluster block.
```
AlexandreSinger, line 520: Same comment as above. Why is this needed? Also AtomBlockId is not a clusterBlockId...

vpr/src/pack/cluster_legalizer.h
```diff
@@ -513,6 +516,16 @@ class ClusterLegalizer {
...
+    *
+    */
+    bool is_atom_blk_in_cluster_block(const AtomBlockId blk_id, const AtomBlockId clustered_blk_id) const;
+
+    inline const AtomPBBimap &atom_pb_lookup() const {return atom_pb_lookup_;}
```
AlexandreSinger, line 525: Please add documentation for what this method is and why it is here.

vpr/src/pack/cluster_legalizer.h
```diff
@@ -580,4 +593,6 @@ class ClusterLegalizer {
     /// @brief The prepacker object that stores the molecules which will be
     ///        legalized into clusters.
     const Prepacker& prepacker_;
+    AtomPBBimap atom_pb_lookup_;
```
AlexandreSinger, line 596: Need documenting comment as above.

vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -415,8 +392,10 @@ void GreedyCandidateSelector::update_connection_gain_values(
     /* may wish to speed things up by ignoring clock nets since they are high fanout */
     for (AtomPinId pin_id : atom_netlist_.net_pins(net_id)) {
         AtomBlockId blk_id = atom_netlist_.pin_block(pin_id);
+        // TODO: Should investigate this. Using the atom pb bimap through is_atom_blk_in_cluster_block
+        // in this class is very strange
         if (cluster_legalizer.get_atom_cluster(blk_id) == legalization_cluster_id
-            && is_atom_blk_in_pb(blk_id, atom_ctx.lookup().atom_pb(clustered_blk_id))) {
+            && cluster_legalizer.is_atom_blk_in_cluster_block(blk_id, clustered_blk_id)) {
```
AlexandreSinger, line 398: Yeah I think the intention of this is to check that these two atoms are in the same cluster. That would be a better method to have in the legalizer class.

vpr/src/pack/pack.cpp
```diff
@@ -178,6 +175,8 @@ bool try_pack(t_packer_opts* packer_opts,
                               is_global,
                               flat_placement_info);
 
+    g_vpr_ctx.mutable_atom().mutable_lookup().lock_atom_pb_bimap = true;
```
AlexandreSinger, line 178: Yeah, I think it would be safer to have setter here.

Like `mutable_lookup().lock_bimap()` or something

vpr/src/pack/pack.cpp
```diff
@@ -289,9 +288,10 @@ bool try_pack(t_packer_opts* packer_opts,
         }
 
         //Reset clustering for re-packing
+        // g_vpr_ctx.mutable_atom().mutable_lookup().lock_atom_pb = false;
```
AlexandreSinger, line 291: Commented out code.

vpr/src/pack/pack.cpp
```diff
@@ -289,9 +288,10 @@ bool try_pack(t_packer_opts* packer_opts,
...
         for (auto blk : g_vpr_ctx.atom().netlist().blocks()) {
             g_vpr_ctx.mutable_atom().mutable_lookup().set_atom_clb(blk, ClusterBlockId::INVALID());
```
AlexandreSinger, line 293: Why does this need to be reset? Did we ever set the atom clb at this point? This maybe something I left behind. It may be able to be removed.

vpr/src/pack/pack.cpp
```diff
@@ -289,9 +288,10 @@ bool try_pack(t_packer_opts* packer_opts,
...
-            g_vpr_ctx.mutable_atom().mutable_lookup().set_atom_pb(blk, nullptr);
+            cluster_legalizer.mutable_atom_pb_lookup().set_atom_pb(blk, nullptr);
```
AlexandreSinger, line 294: I think this would be better as a clear method. Instead of setting every block to nullptr, I think calling this as a single clear method would be cleaner and safer!

vpr/src/pack/pack.cpp
```diff
@@ -301,7 +301,7 @@ bool try_pack(t_packer_opts* packer_opts,
 
         // Reset the cluster legalizer for re-clustering.
         cluster_legalizer.reset();
-
+        // g_vpr_ctx.mutable_atom().mutable_lookup().lock_atom_pb = true;
```
AlexandreSinger, line 304: Commented out code.

vpr/src/pack/pack.cpp
```diff
@@ -319,7 +319,8 @@ bool try_pack(t_packer_opts* packer_opts,
      * }
      */
     /******************** End **************************/
-
+    g_vpr_ctx.mutable_atom().mutable_lookup().lock_atom_pb_bimap = false;
+    g_vpr_ctx.mutable_atom().mutable_lookup().set_atom_to_pb_bimap(cluster_legalizer.atom_pb_lookup());
```
AlexandreSinger, line 323: This copy may be expensive. It may make sense to make this a move if we are certain that the lookup in the cluster legalizer is never used again. But I am not sure. Maybe its not worth it! In the future this interface should be made more clear once we stop writing and reading from the .net file.

AlexandreSinger (changes requested):
@AmirhosseinPoolad Thank you so much for this cleanup. It was desperately needed. I think we are getting close to removing this lookup entirely in the packer. I had some comments throughout. Be careful when you rebase onto a very large change like Soheil's formatting change. It can really cause havoc on PRs like this and it will undo Soheil's changes.

I recommend letting Vaughn review this PR as well (when he gets back).

vpr/src/pack/atom_pb_bimap.h
AmirhosseinPoolad: I kind of disagree with this. This class itself is the alias for the bimap, any users should be using the class and not the underlying vtr::bimap. For the non-default constructor, which shouldn't really be used used anywhere other than in pack.cpp, I feel like having it be obvious what the type is is better than obfuscating it with a typedef.

vpr/src/pack/greedy_candidate_selector.cpp
AmirhosseinPoolad: Yeah there was actually a TODO to investigate why this method is doing this lookup, but I didn't want to do the investigation and change the code and left the code as is.

```
// TODO: Should investigate this. Using the atom pb bimap through is_atom_blk_in_cluster_block
// in this class is very strange
```

AmirhosseinPoolad:
Hi @AlexandreSinger, thanks for the review! I did some more changes based on your comments, do you mind taking another look? Thanks.

vpr/src/pack/cluster_legalizer.cpp
AlexandreSinger: Instead of an assert, you should move the resetting of the bimap into this method.

AlexandreSinger (changes requested):
Hi @AmirhosseinPoolad 

This is coming along great! I agree with your changes so far. I do have some more comments. There are also some comments I made before which I think still need to be addressed.

vpr/src/pack/atom_pb_bimap.cpp
```diff
@@ -0,0 +1,63 @@
...
+        atom_to_pb_.erase(blk_id);
+    } else if (blk_id && pb) {
+        //If both are valid store the mapping
+        atom_to_pb_.update(blk_id, pb);
+    }
+}
+
+void AtomPBBimap::reset_bimap(const AtomNetlist &netlist) {
+    for (auto blk : netlist.blocks()) {
+        set_atom_pb(blk, nullptr);
```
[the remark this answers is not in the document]
AmirhosseinPoolad: It does have one. The reason I didn't use it is at first I thought this was a "just set to nullptr but don't actually erase the data structure" which wouldn't be the same thing as calling clear, but the code for set_atom_pb does actually erase the data structure if one of the arguments is nullptr so it *should* be equivalent.

vpr/src/pack/greedy_candidate_selector.cpp
AmirhosseinPoolad: So, this code is trying to see if blk_id's t_pb is a child t_pb of clusterered_blk_id's t_pb (both blk_ids are mapped to a t_pb using the bimap). Before these changes, update_connection_gain_values would use atom_pb lookup from global context to get the t_pb for clustered_blk_id and that would be the second argument for is_atom_blk_in_pb. The first argument would be blk_id that was passed as is and would be mapped/converted to a t_pb in the is_atom_blk_in_pb function.

Personally I think it should be consistent, either we do all of the mapping from AtomBlockId to t_pb in the is_atom_blk_in_* method or all of it outside of it. i.e. the arguments should be either both AtomBlockIds or both t_pbs. I do get your point however, this isn't a good fit for ClusterLegalizer. Currently changing it to a static function in greedy_candidate_selector.cpp that takes two t_pb* arguments since otherwise you'd have to pass it the AtomPBBimap which I don't think is very nice.

AmirhosseinPoolad:
Oh btw I tried to remove the is_blom_blk function completely and somehow the tests weren't passing. Not sure why.

AlexandreSinger (approved):
LGTM! Very well done on this PR Amir; sorry for all the comments. This is a tricky bit of code.

Please resolve the conflicts and ensure it passes CI and then I think this can be merged.

PR #2933, opened 2025-03-13
Another update of the code base to be consistent with code formatting rules

soheilshahrouz:
Read #2931.

### DIFF
--- a/vpr/test/test_vpr_constraints.cpp
+++ b/vpr/test/test_vpr_constraints.cpp
@@ -14,18 +14,16 @@
  * UserPlaceConstraints, Region, PartitionRegions, and Partition.
  */
 
-namespace Catch
-{
-template <typename T>
-struct StringMaker<vtr::Rect<T>>
-{
-    static std::string convert(vtr::Rect<T> const &value) {
+namespace Catch {
+template<typename T>
+struct StringMaker<vtr::Rect<T>> {
+    static std::string convert(vtr::Rect<T> const& value) {
         std::ostringstream oss;
         oss << "Rectangle: (" << value.xmin() << ", " << value.ymin() << ") to (" << value.xmax() << ", " << value.ymax() << ")";
         return oss.str();
     }
 };
-}
+} // namespace Catch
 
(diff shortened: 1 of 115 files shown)


vpr/test/test_vpr_constraints.cpp
```diff
@@ -458,13 +448,12 @@ TEST_CASE("PartRegionIntersect4MultiLayer", "[vpr]") {
...
     pr2.add_to_part_region(r4);
 
-
     PartitionRegion int_pr = intersection(pr1, pr2);
     const std::vector<Region>& regions = int_pr.get_regions();
 
     REQUIRE(regions.size() == 1);
     REQUIRE(regions[0].get_layer_range() == std::make_pair(1, 1));
-    REQUIRE(regions[0].get_rect() == vtr::Rect<int> (1, 2, 3, 4));
+    REQUIRE(regions[0].get_rect() == vtr::Rect<int>(1, 2, 3, 4));
```
amin1377, line 456: I usually add a space before passing the constructor parentheses myself, but I'm not sure what the common approach is.

amin1377 (changes requested):
Thanks, Soheil! I reviewed all the files and left some comments here and there to draw your attention to certain spots, just to ensure everything is formatted as expected.  

I think that if you add the format check now, it should pass. So, I strongly recommend adding it to the CI test runnig on GitHub servers in THIS PR —otherwise, you may wind up doing this every once in a while! :)

AlexandreSinger (approved):
LGTM

Basically the same comments as the last PR. I really want to stress that we should set the column limit to something. I am seeing clang format convert shorter lines that are using return carriages to make code more readible into one single massive line which is very hard to read on any editor. Something we can add in the future I guess.

vpr/test/test_vpr_constraints.cpp
AlexandreSinger: I normally do not add a space before the constructor myself. I personally agree with Clang Format here.

amin1377 (approved):

PR #2937, opened 2025-03-17
Initial Placement Wirelength Est.

amin1377:
When printing information about the initial placement, the wirelength cost, td cost, and CPD estimate are shown, but the wirelength estimate is not printed. I think it would be useful to include this information as well, and this PR addresses that.

### DIFF
--- a/vpr/src/place/net_cost_handler.cpp
+++ b/vpr/src/place/net_cost_handler.cpp
@@ -1625,6 +1625,20 @@ void NetCostHandler::recompute_costs_from_scratch(const PlaceDelayModel* delay_m
     }
 }
 
+double NetCostHandler::get_total_wirelength_estimate() const {
+    const auto& cluster_ctx = g_vpr_ctx.clustering();
+    const auto& place_move_ctx = placer_state_.move();
+
+    double estimated_wirelength = 0.0;
+    for (ClusterNetId net_id : cluster_ctx.clb_nlist.nets()) { /* for each net ... */
+        if (!cluster_ctx.clb_nlist.net_is_ignored(net_id)) {   /* Do only if not ignored. */
+                estimated_wirelength += get_net_wirelength_estimate(net_id, place_move_ctx.bb_coords[net_id]);
+        }
+    }
+
+    return estimated_wirelength;
+}
+
 void NetCostHandler::set_ts_bb_coord_(const ClusterNetId net_id) {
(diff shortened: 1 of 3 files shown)


vpr/src/place/net_cost_handler.cpp
```diff
@@ -1625,6 +1625,20 @@ void NetCostHandler::recompute_costs_from_scratch(const PlaceDelayModel* delay_m
     }
 }
 
+double NetCostHandler::get_total_wirelength_estimate() const {
+    const auto& cluster_ctx = g_vpr_ctx.clustering();
```
AlexandreSinger, line 1629: Nit: You are bringing in the cluster_ctx, but you are only using the clustered netlist variable. I recommend creating a variable just for the netlist. It also makes cleaning up these global accesses easier.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1625,6 +1625,20 @@ void NetCostHandler::recompute_costs_from_scratch(const PlaceDelayModel* delay_m
...
+    const auto& place_move_ctx = placer_state_.move();
```
AlexandreSinger, line 1630: Similarly here, we only need the bb_coords. I think it will just make the later lines easier to read.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1625,6 +1625,20 @@ void NetCostHandler::recompute_costs_from_scratch(const PlaceDelayModel* delay_m
...
+
+    double estimated_wirelength = 0.0;
+    for (ClusterNetId net_id : cluster_ctx.clb_nlist.nets()) { /* for each net ... */
+        if (!cluster_ctx.clb_nlist.net_is_ignored(net_id)) {   /* Do only if not ignored. */
+                estimated_wirelength += get_net_wirelength_estimate(net_id, place_move_ctx.bb_coords[net_id]);
```
AlexandreSinger, line 1635: This is double indented.

AlexandreSinger (commented):
Hi @amin1377 , thank you so much for looking into this! This will make it much easier to debug APPack. I left some minor comments. Otherwise LGTM.

AlexandreSinger (approved):
LGTM

soheilshahrouz (commented):
I suggest files under `vtr_flow/parse` are also updated so that we can parse the initial placement wirelength estimate.

PR #2939, opened 2025-03-18
Remove Redundant print_pb

amin1377:
The print_pb_type_count function is currently called twice: once in vpr_oad_packing and again in AAPack::legalize after loading the packing from a file. In this PR, the redundant second call is removed.

### DIFF
--- a/vpr/src/analytical_place/full_legalizer.cpp
+++ b/vpr/src/analytical_place/full_legalizer.cpp
@@ -523,11 +523,6 @@ void APPack::legalize(const PartialPlacement& p_placement) {
     // The Packer stores the clusters into a .net file. Load the packing file.
     // FIXME: This should be removed. Reading from a file is strange.
     vpr_load_packing(vpr_setup_, arch_);
-    const ClusteredNetlist& clb_nlist = g_vpr_ctx.clustering().clb_nlist;
-
-    // Verify the packing and print some info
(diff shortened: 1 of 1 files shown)


vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -528,7 +528,6 @@ void APPack::legalize(const PartialPlacement& p_placement) {
     // Verify the packing and print some info
     check_netlist(vpr_setup_.PackerOpts.pack_verbosity);
     writeClusteredNetlistStats(vpr_setup_.FileNameOpts.write_block_usage);
-    print_pb_type_count(clb_nlist);
```
AlexandreSinger, line 531: Thank you so much for noticing this Amin. Looking back at this code, I actually realize that all three of these lines are unnecessary since they are done in load_packing anyways:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/7905df17ec554abaa2a4116f9141635d6e8083ad/vpr/src/base/vpr_api.cpp#L702-L722

Would you mind removing `check_netlist` and `writeClusteredNetlistStats` as well?

AlexandreSinger (commented):
LGTM! Thank you so much for fixing this! It was a leftover from an older version of the Full Legalizer. I just had one minor comment.

AlexandreSinger (approved):
LGTM

PR #2941, opened 2025-03-19
[RR Graph] RR Node Indices Value Type

amin1377:
This PR updates rr_node_indices_ to store RRNodeId instead of int. This change addresses issues that arise when the RR graph contains more than 2 million nodes. Additionally, it simplifies handling of invalid RRNodeId values and eliminates the need to cast unsigned int to int. This issue was flagged by @heshpdx.


amin1377:
@heshpdx FYI

vaughnb-cerebras:
Looks good to me!  This is a nice cleanup, and it looks like you found and eliminated some unused routines. Thanks.

vaughnbetz (approved):
LGTM

PR #2944, opened 2025-03-20
[AP][Solver] Supporting Unfixed Blocks

AlexandreSinger:
When no fixed blocks are provided by the user, the AP flow can still work. Currently, in the first iteration, the solver will put all blocks at 0,0 and use the legalized solution in the next iteration as fixed points. Instead of (0,0), it makes more sense to put the blocks in the center of the device.

Also added a guess to the solver to help CG converge faster each iteration.

Added a regression test to ensure that not describing the fixed blocks is supported.


vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -296,8 +360,23 @@ void QPHybridSolver::solve(unsigned iteration, PartialPlacement& p_placement) {
...
-        p_placement.block_x_locs[blk_id] = x[row_id_idx];
-        p_placement.block_y_locs[blk_id] = y[row_id_idx];
+        // Due to the iterative nature of CG, it is possible for the solver to
+        // overstep 0 and return a negative number by an incredibly small margin.
+        // Clamp the number to 0 in this case.
+        // TODO: Should investigate good bounds on this, the bounds below were
+        //       chosen since any difference higher than 1e-9 would concern me.
+        double x_pos = x_soln[row_id_idx];
+        if (x_pos < 0.0) {
+            VTR_ASSERT_SAFE(std::abs(x_pos) < 1e-9);
```
amin1377, line 370: We encountered a similar issue in the timing analyzer, where a global variable was defined for the file so it could be modified later. I’d suggest following the same approach here.

vpr/src/analytical_place/analytical_solver.h
```diff
@@ -180,6 +185,14 @@ class QPHybridSolver : public AnalyticalSolver {
      */
     void init_linear_system();
 
+    /**
+     * @brief Intializes the guesses which will be used in the solver.
+     *
+     * The guesses will be used as starting points for the CG solver. The better
```
amin1377, line 191: I think based on our discussion, this was what we initially thought. But since it doesn’t seem to work for the first iteration—when the hint is positioned in the middle and there’s no fixed block—do you still think this is how the initial guess is supposed to work?

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -280,14 +334,24 @@ void QPHybridSolver::solve(unsigned iteration, PartialPlacement& p_placement) {
...
-    Eigen::VectorXd y = cg.solve(b_y_diff);
+    Eigen::VectorXd y = cg.solveWithGuess(b_y_diff, guess_y);
     VTR_ASSERT(cg.info() == Eigen::Success && "Conjugate Gradient failed at solving b_y!");
 
     // Write the results back into the partial placement object.
+    store_solution_into_placement(x, y, p_placement);
+
+    // Update the guess. The guess for the next iteration is the solution in
+    // this iteration.
+    guess_x = x;
```
amin1377, line 347: By the way, did this update affect the results? Given how fast the solver already is, I wouldn’t expect to see any significant improvement in runtime, though.

amin1377 (changes requested):
Thanks, Alex!

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: It improved the runtime. The QoR does change, but I would not expect it to change by much. I will run Titan to see!

amin1377 (approved):

PR #2945, opened 2025-03-21
Update clang-format version to 18

AmirhosseinPoolad:
Ubuntu 24.04 installs clang-format version 18 by default and this PR updates the version used by CI and 'make format' from 14 to 18. Since clang-format 18 is a bit smarter than 14, there were some parts of the code where it fixed some formatting issues that version 14 did not find problematic. The second commit of this PR fixes those issues.


.github/scripts/install_jammy_dependencies.sh
```diff
@@ -51,7 +51,7 @@ sudo apt install -y \
   gcc-11 \
   g++-12 \
   gcc-12 \
-  clang-format-14 \
+  clang-format-18 \
```
AlexandreSinger, line 54: I would not change this one. This is what is causing the error in the Ubuntu 22.04 CI test.

clang-format-18 may not be provided for Ubuntu 22.04. Since we do not run clang-format on the Ubuntu 22.04, I think you can just remove this line.

AlexandreSinger: The 'tcl' error you are seeing in the failing CI test is something I saw a lot when I was upgrading to Ubuntu 24.04. It happens when the dependencies do not get installed correctly. If one package does not exist in that long list, none of them get installed I think.

soheilshahrouz (approved):

PR #2946, opened 2025-03-21
3D SW Architecture

amin1377:
Fixes the architecture with 3D switch blocks by adding delays on connections to L4 and L16. The code has also been updated to account for delays originating either from an OPIN on another layer or from another routing wire.


vpr/src/base/vpr_types.h
```diff
@@ -1377,11 +1377,10 @@ struct t_det_routing_arch {
...
  *                     at the end of build_rr_graph
- *   @param arch_opin_between_dice_switch Index of the switch type that connects output
- *                     pins (OPINs) *to* this segment from *another dice*.
- *                     Note that this index is in relation to the switches from
- *                     the architecture file, not the expanded list of switches that is built
- *                     at the end of build_rr_graph
+ *   @param arch_inter_die_switch Index of the switch type that connects a driver
+ *                     to this segment from another die. Note that this index is in relation
+ *                     to the switches from the architecture file, not the expanded list of
+ *                     switches that is built at the end of build_rr_graph
```
soheilshahrouz, line 1383: Don't use `@param` for a data member. Use `@var` instead

amin1377: To be consistent with your earlier comment, I moved the comments for the struct fields inside the struct.

amin1377:
@soheilshahrouz Thanks for reviewing the code. I’ve applied your comments—please take a look and feel free to merge if everything looks good to you.

libs/libarchfpga/src/physical_types.h
```diff
@@ -1599,88 +1599,138 @@ enum e_Fc_type {
...
- *           be GENERAL.                                                     *
- * meta: Table storing extra arbitrary metadata attributes.                  */
+/**
+ * @brief Lists all the important information about a certain segment type.  Only
+ * used if the route_type is DETAILED.  [0 .. det_routing_arch.num_segment]
+ */
 struct t_segment_inf {
+    /** @var name
+     *  @brief The name of the segment type 
+     */
```
soheilshahrouz, line 1609: I think using `@var` is redundant when comments are moved inside the struct.

amin1377: I think if someone wants to later generate documentation based on Doxygen comments, using @var might make the output look cleaner. But I agree with you—it adds unnecessary clutter to the comments, so I’ve removed it.

PR #2947, opened 2025-03-21
[APPack] Flat-Placement Informed Unrelated Clustering

AlexandreSinger:
Used flat placement information provided by APPack to try and select better unrelated candidates. This searches for candidates as close to the flat placement position of the cluster.

There are two parameters that control how this is performed:

1) max_unrelated_tile_distance decides how far the algorithm will search
   for unrelated candidates. The algorithm will check for candidates in
   the same tile as the cluster, and then will search farther and
   farther out

2) max_unrelated_clustering_attempts decides how many failing attempts
   the cluster will try unrelated clustering. This matches the option of
   the same name in the candidate selector class; but this was made
   separate since likely it will be different for APPack.


vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -1154,6 +1212,112 @@ PackMoleculeId GreedyCandidateSelector::get_unrelated_candidate_for_cluster(
...
+    // Push the position of the cluster to the queue.
+    search_queue.push(cluster_gain_stats.flat_cluster_position);
+
+    while (!search_queue.empty()) {
+        // Pop a position to search from the queue.
+        const t_flat_pl_loc& node_loc = search_queue.front();
+        VTR_ASSERT_SAFE(node_loc.layer == 0);
+
+        // If this position is too far from the source, skip it.
+        float dist = get_manhattan_distance(node_loc, cluster_gain_stats.flat_cluster_position);
```
amin1377, line 1267: I think given that this function only accepts parameters of type `t_flat_pl_loc`, it would be better to define a method in this struct to get L1 norm.

vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -1154,6 +1212,112 @@ PackMoleculeId GreedyCandidateSelector::get_unrelated_candidate_for_cluster(
...
+                // If this molecule is not compatible with the current cluster
+                // skip it.
+                if (!cluster_legalizer.is_molecule_compatible(mol_id, cluster_id))
+                    continue;
+                // Return this molecule as the unrelated candidate.
+                return mol_id;
+            }
+        }
+
+        // Push the neighbors of the position to the queue.
```
amin1377, line 1303: Why not include the diagonal neighbors of the tile as well?

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/unrelated_clustering/config/config.txt
```diff
@@ -0,0 +1,54 @@
+###############################################################################
```
amin1377, line 1: Do we have a test when IO blocks are not fixed?

vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -105,44 +105,93 @@ GreedyCandidateSelector::GreedyCandidateSelector(
...
+            vtr::NdMatrix<std::vector<std::vector<PackMoleculeId>>, 2>({flat_grid_width,
+                                                                        flat_grid_height});
+        for (size_t x = 0; x < flat_grid_width; x++) {
+            for (size_t y = 0; y < flat_grid_height; y++) {
+                appack_unrelated_clustering_data_[x][y].resize(max_molecule_stats.num_used_ext_pins + 1);
+            }
+        }
+
+        // Fill the grid with molecule information.
+        for (PackMoleculeId mol_id : molecules_vector) {
```
amin1377, line 168: I think it would be helpful to add a comment mentioning that iterating over the sorted molecule list is important to ensure that the elements in tile_uc_data remain sorted.

amin1377 (changes requested):

vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -105,44 +105,93 @@ GreedyCandidateSelector::GreedyCandidateSelector(
...
+        /**
+         * For APPack, we build a spatial data structure where for each 1x1 grid
+         * position on the FPGA, we maintain lists of molecule candidates.
+         * The lists are in order of number of used external pins by the molecule.
+         * Within each list, the molecules are sorted by their base gain.
+         */
+        // Get the max x, y, and layer from the flat placement.
+        t_flat_pl_loc max_loc({0.0f, 0.0f, 0.0f});
+        for (PackMoleculeId mol_id : molecules_vector) {
+            t_flat_pl_loc mol_pos = get_molecule_pos(mol_id, prepacker_, appack_ctx_);
```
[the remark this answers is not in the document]
AlexandreSinger: get_mol_pos does not return a reference to a t_flat_pl_loc but an rvalue created within. No copy is created in this case, regardless of if it is declared as a const reference or not I am pretty sure. This is also a very small object.

vpr/src/pack/greedy_candidate_selector.cpp
AlexandreSinger: I personally feel that having a method in this form makes more sense. This is similar to how things are done in the STL. For example std::max, std::min, std::distance; I think it makes more sense for these binary operators to not belong to the class.

vpr/src/pack/greedy_candidate_selector.cpp
AlexandreSinger: We are doing the manhattan / L1 distance here, not the euclidean / L2 distance; technically the diagonals are 2 away from the current tile, not one. Even in euclidean space, the diagonals are 1.4 distance away, so we want to check the direct neighbors before attempting the diagonals anyways.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/unrelated_clustering/config/config.txt
AlexandreSinger: This is just the case I want to support the most right now. Its also simple to add since I already have the fixed IOs for these testcases. I still consider not providing IO blocks an experimental feature.

AlexandreSinger:
@amin1377 Do you have any further comments on this PR (or any push-back on my explanations above)?

vpr/src/pack/greedy_candidate_selector.cpp
amin1377: The functions you’re mentioning are template functions, which is why it makes sense to have them as separate functions. In contrast, get_manhattan_distance only accepts one type, so I don’t see a strong need to separate it out. That said, I don’t have a strong preference—feel free to ignore this comment.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/unrelated_clustering/config/config.txt
amin1377: Got it. When we were discussing the changes, you mentioned that you might want to claim the feature that IOs don’t need to be fixed. I just brought this up to check whether this is a feature you’d like AP to support or not. If it is supported, we should have a corresponding test for it.

amin1377 (approved):
Thanks, Alex. The PR looks good to me. Feel free to merge it if you've made all the changes you wanted.

vpr/src/pack/greedy_candidate_selector.cpp
AlexandreSinger: I also do not have a strong preference; but I do not think its worth changing lol. Thanks!

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/unrelated_clustering/config/config.txt
AlexandreSinger: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/e7f3a02dfb3f6954efce7146e5564026c2c0d0c4/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/no_fixed_blocks/config/config.txt

:) Added in the last PR!

vpr/src/pack/greedy_candidate_selector.cpp
AlexandreSinger: That is true; however throughout the AP flow I use the L1 norm. In theory it should mean that diagonals are worst than up and down! Perhaps its something to investigate in the future!

PR #2951, opened 2025-03-24
[Pack] Device Util Output 

amin1377:
Fix the placement of device utilization and netlist info printing to prevent them from being printed when the clustered netlist is not built (i.e., when AP is enabled).

### DIFF
--- a/vpr/src/base/SetupGrid.cpp
+++ b/vpr/src/base/SetupGrid.cpp
@@ -15,6 +15,7 @@
 #include "vtr_assert.h"
 #include "vtr_math.h"
 #include "vtr_log.h"
+#include "stats.h"
 
 #include "vpr_types.h"
 #include "vpr_error.h"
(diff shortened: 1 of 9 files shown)


vpr/src/base/SetupGrid.cpp
```diff
@@ -15,6 +15,7 @@
 #include "vtr_assert.h"
 #include "vtr_math.h"
 #include "vtr_log.h"
+#include "stats.h"
```
AlexandreSinger, line 18: Is this header file needed to be included here?

vpr/src/base/stats.h
```diff
@@ -3,6 +3,7 @@
 #include <limits>
 #include <algorithm>
 #include "vpr_types.h"
+#include "netlist.h"
```
AlexandreSinger, line 6: I think this include should be `netlist_fwd.h` instead of `netlist.h`

vpr/src/base/stats.h
```diff
@@ -23,6 +24,25 @@ void get_num_bends_and_length(ParentNetId inet, int* bends, int* length, int* se
 
 int count_netlist_clocks();
 
+/**
+ * @brief Calculate the device utilization
+ *
+ * Calculate the device utilization (i.e. fraction of used grid tiles)
+ * foor the specified grid and resource requirements
```
AlexandreSinger, line 31: foor -> for

vpr/src/base/vpr_api.cpp
```diff
@@ -384,9 +384,14 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
         }
     }
 
-    // For the time being, we decided to create the flat graph after placement is done. Thus, the is_flat parameter for this function
-    //, since it is called before routing, should be false.
-    vpr_create_device(vpr_setup, arch, false);
+    vpr_create_device(vpr_setup, arch);
+    if (vpr_setup.PackerOpts.doPacking != STAGE_SKIP) {
```
AlexandreSinger, line 388: I wonder if this should be if the AP flow is not active. My thought process is what if people skip the pack flow for another reason, would they not see this info? If not, I think this needs a comment to explain what it means.

vpr/src/pack/pack.cpp
```diff
@@ -18,6 +18,7 @@
 #include "vpr_context.h"
 #include "vpr_error.h"
 #include "vpr_types.h"
+#include "stats.h"
```
AlexandreSinger, line 21: Is this include file needed?

AlexandreSinger (approved):
LGTM. Thank you for doing this Amin, it was annoying to see the stats for an empty device being printed in the AP flow. This should print more interesting information regarding device utilization in the AP flow.

AlexandreSinger:
Heads up, there may be a small issue in the printing code:

Screenshot from 2025-03-24 16-42-42

vpr/src/base/SetupGrid.cpp
amin1377: I am using the functions defined in this header file (e.g., calculate_device_utilization)

vpr/src/base/stats.h
amin1377: I initially included netlist_fwd.h, but since the default values for the template parameters aren't defined in the forward declaration, it complained about using Netlist<> as a function parameter. I did a bit of research, and it seems it's not good practice to specify default template values in a forward declaration, so I had to include the full header file instead.

vpr/src/pack/pack.cpp
amin1377: `calculate_device_utilization` is called in this file which is defined in `pack.cpp`.

vpr/src/base/stats.h
AlexandreSinger: That makes sense to me! Thank you! It is interesting that some of these files are doing Netlist<>, but not a big deal!

amin1377: It was a workaround I used to enable inheritance from a templated class. Ideally, the base class wouldn’t be templated, but since the return types of the virtual functions are StrongId, I had to implement this (probably "hacky") solution. It also caused some issues when UBSan is enabled, and we had to disable it. @heshpdx has encountered some problems with it in SPEC as well. At some point, we should revisit this and think about a cleaner solution.

PR #2952, opened 2025-03-25
[Place] Add CLI Option to Skip Annealing Phase in VPR

amin1377:
This PR adds a CLI parameter to VPR that allows users to skip the annealing phase and run only the quench phase. By default, this parameter is disabled, meaning the annealing phase is performed as usual.

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -2309,6 +2309,12 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("0")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    place_grp.add_argument<bool, ParseOnOff>(args.place_skip_anneal, "--place_skip_anneal")
+        .help(
+            "Skip the placement annealing phase and go straight to the placement quench.")
+        .default_value("off")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
     place_grp.add_argument<e_agent_algorithm, ParsePlaceAgentAlgorithm>(args.place_agent_algorithm, "--place_agent_algorithm")
         .help("Controls which placement RL agent is used")
         .default_value("softmax")
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -2315,7 +2315,7 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("0")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
-    place_grp.add_argument<bool, ParseOnOff>(args.place_skip_anneal, "--place_skip_anneal")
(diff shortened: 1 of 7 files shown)


vpr/src/base/read_options.cpp
```diff
@@ -2309,6 +2309,12 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("0")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    place_grp.add_argument<bool, ParseOnOff>(args.place_skip_anneal, "--place_skip_anneal")
+        .help(
+            "Skip the placement annealing phase and go straight to the placement quench.")
```
AlexandreSinger, line 2314: Should we add this CLI option to the documentation as well?

vpr/src/place/placer.cpp
```diff
@@ -285,15 +286,14 @@ void Placer::place() {
     const auto& timing_ctx = g_vpr_ctx.timing();
     const auto& cluster_ctx = g_vpr_ctx.clustering();
 
-    bool skip_anneal = false;
 #ifdef ENABLE_ANALYTIC_PLACE
     // Cluster-level analytic placer: when enabled, skip most of the annealing and go straight to quench
     if (placer_opts_.enable_analytic_placer) {
-        skip_anneal = true;
+        skip_anneal_ = true;
```
AlexandreSinger, line 292: skip_anneal_ was declared as a const member in the class. This assignment will not work when the cluster-level AP flow is active. You could create a temporary variable in this method which gets set to true when ENABLE_ANALYTIC_PLACE is active. I agree that the member variable should be const.

vpr/src/base/read_options.cpp
```diff
@@ -2309,6 +2309,12 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("0")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    place_grp.add_argument<bool, ParseOnOff>(args.place_skip_anneal, "--place_skip_anneal")
```
AlexandreSinger, line 2312: I feel like "skip_anneal" is a strange name for this option. People who are unfamiliar with VTR may think that means skip the entire place flow (since the whole flow is called anneal). I wonder if this should be "quench_only" or something?

AlexandreSinger (commented):
LGTM. Some minor comments.

vpr/src/place/placer.cpp
amin1377: Good catch!

vpr/src/base/read_options.cpp
amin1377: Good suggestion. I changed the name to --place_quench_only

amin1377:
@AlexandreSinger: This PR is ready for merge. Please feel free to merge if there are no more suggestions.

AlexandreSinger (approved):
LGTM

PR #2953, opened 2025-03-26
[Place] Fix regex pattern note in constraints doc

amin1377:
Updated the placement constraint documentation to clarify that the `name_pattern` specified in the `<add_atom>` tag is interpreted as a regular expression. Also fixed a minor issue in the examples: the documentation previously suggested using `*` after a name to match any suffix (e.g., `"alu*"`), but in regex, `*` applies only to the preceding character. To correctly match any suffix, the pattern should use `.*` (e.g., `"alu.*"`).


vaughnb-cerebras:
LGTM!

vaughnbetz (approved):

PR #2954, opened 2025-03-30
[VTR Script] Add Run Number to Parse Script

amin1377:
This PR introduces a run option to the parse script, allowing users to specify a specific run directory instead of always parsing the latest. If a run directory is provided, it will be parsed; otherwise, the script defaults to the latest run directory.

### DIFF
--- a/vtr_flow/scripts/python_libs/vtr/parse_vtr_task.py
+++ b/vtr_flow/scripts/python_libs/vtr/parse_vtr_task.py
@@ -32,6 +32,7 @@ from vtr import (
     VtrError,
     create_jobs,
     paths,
+    set_global_run_dir_number,
 )
 
 # pylint: enable=wrong-import-position
@@ -130,7 +131,7 @@ def vtr_command_argparser(prog=None):
         help="QoR geomeans are not computed by default",
     )
 
-    parser.add_argument("-run", default=None, type=str, help="")
+    parser.add_argument("-run", default=None, type=int, help="Run number to parse. If not provided, the latest run will be parsed.")
 
     parser.add_argument("-revision", default="", help="Revision number")
 
@@ -144,6 +145,8 @@ def vtr_command_main(arg_list, prog=None):
     """
     # Load the arguments
     args = vtr_command_argparser(prog).parse_args(arg_list)
+    if args.run is not None:
+        set_global_run_dir_number(args.run)
     try:
         task_names = args.task
 
--- a/vtr_flow/scripts/python_libs/vtr/util.py
+++ b/vtr_flow/scripts/python_libs/vtr/util.py
@@ -22,6 +22,10 @@ from vtr.error import CommandError
 from vtr import paths
 
 
+# The run number passed by get_latest_run_number is the latest run number for the task if this number is None. Otherwise
+# it is the run number to parse.
+global_run_dir_number = None
+
 class RawDefaultHelpFormatter(
     argparse.ArgumentDefaultsHelpFormatter, argparse.RawDescriptionHelpFormatter
 ):
@@ -223,6 +227,13 @@ class CommandRunner:
         return cmd_output, cmd_returncode
 
     # pylint: enable=too-many-arguments, too-many-instance-attributes, too-few-public-methods, too-many-locals
+
+def set_global_run_dir_number(run_dir_number):
+    """
+    Set the global run directory number.
+    """
+    global global_run_dir_number
+    global_run_dir_number = run_dir_number
 
 
 def check_cmd(command):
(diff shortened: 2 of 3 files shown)


vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -563,20 +574,26 @@ def get_latest_run_number(base_dir):
     """
     Returns the highest run number of all run directories with in base_dir
     """
-    run_number = 1
-    run_dir = Path(base_dir) / run_dir_name(run_number)
+    global global_run_dir_number
+    run_number = None
+    if global_run_dir_number is None:
```
soheilshahrouz, line 579: I suggest you return `global_run_dir_number` if it is not `None` at the beginning of function.

```
if global_run_dir_number is not None:
    return global_run_dir_number
```

This way, the rest of function can be left untouched, and I think it would be easier to read.

vtr_flow/scripts/python_libs/vtr/parse_vtr_task.py
```diff
@@ -130,7 +131,7 @@ def vtr_command_argparser(prog=None):
         help="QoR geomeans are not computed by default",
     )
 
-    parser.add_argument("-run", default=None, type=str, help="")
+    parser.add_argument("-run", default=None, type=int, help="Run number to parse. If not provided, the latest run will be parsed.")
```
soheilshahrouz, line 134: If you accept a string argument, the user can rename runxxx directories to more meaningful names and pass the run directory name directly.

amin1377: Good suggestion!

amin1377:
@soheilshahrouz: I've addressed your comments. Please review the code again, as your feedback led to some significant changes. Let me know if you have any further suggestions.

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -549,34 +582,37 @@ def get_next_run_number(base_dir):
...
+    next_run_number = 1
+    if match:
+        latest_run_number = int(match.group(1))
         next_run_number = latest_run_number + 1
 
     return next_run_number
 
 
-def get_latest_run_number(base_dir):
+def get_latest_run_dir_name(base_dir):
```
soheilshahrouz, line 595: The function name is confusing because the user provided directory name may not be the latest one.

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -549,34 +582,37 @@ def get_next_run_number(base_dir):
...
     """
     Returns the highest run number of all run directories with in base_dir
     """
-    run_number = 1
-    run_dir = Path(base_dir) / run_dir_name(run_number)
+    latest_run_dir_name = ""
```
soheilshahrouz, line 599: The variable name can be confusing for the same reason.

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -225,6 +250,13 @@ def run_system_command(
     # pylint: enable=too-many-arguments, too-many-instance-attributes, too-few-public-methods, too-many-locals
 
 
+def set_global_run_dir(current_run_dir_name):
+    """
+    Set the global run directory name.
+    """
+    RunDir.set_run_dir_name(current_run_dir_name)
+
+
```
soheilshahrouz, line 259: Is this setter needed? 
`RunDir.set_run_dir_name()` is class method and does the same thing.

vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -525,12 +557,13 @@ def get_latest_run_dir(base_dir):
     """
     Returns the run directory with the highest run number in base_dir
     """
-    latest_run_number = get_latest_run_number(base_dir)
+    latest_run_dir_name = get_latest_run_dir_name(base_dir)
```
soheilshahrouz, line 560: The function name, the comment and the variable name can be confusing becuase the user might specify another run directory.

amin1377: I changed the function name to `get_active_run_dir_name` and changed the comment to:

```
Returns the active run directory name. If the user has specified
a run directory name, it will be returned. Otherwise, the
highest run number of all run directories within base_dir
will be returned.
```

vtr_flow/scripts/python_libs/vtr/util.py
amin1377: Good point. I removed the function call and am calling the class method directly.

PR #2955, opened 2025-03-31
[Place] rename get_initial_move_lim to get_place_inner_loop_num_move

amin1377:
Updated function name from `get_initial_move_lim` to `get_place_inner_loop_num_move`. This clarifies that the number of placement moves within the inner loop is a constant limit, not a variable that changes during annealing.


vpr/src/place/place_util.cpp
```diff
@@ -50,7 +50,7 @@ t_placer_costs& t_placer_costs::operator+=(const NocCostTerms& noc_delta_cost) {
     return *this;
 }
 
-int get_initial_move_lim(const t_placer_opts& placer_opts, const t_annealing_sched& annealing_sched) {
+int get_place_inner_loop_num_move(const t_placer_opts& placer_opts, const t_annealing_sched& annealing_sched) {
```
soheilshahrouz, line 53: Probably renaming it to `comp_place_inner_loop_num_move()` makes it clear that it actually computes the value.

vpr/src/place/place_util.h
```diff
@@ -197,7 +197,7 @@ class t_placer_statistics {
 };
 
 /**
- * @brief Get the initial limit for inner loop block move attempt limit.
+ * @brief Get the number of moves tried in the inner loop of the annealer.
```
soheilshahrouz, line 200: This comment can also be made more clear:

Compute the number of moves attempted by the annealer's inner loop in each outer loop iteration.

vpr/src/place/place_util.cpp
amin1377: I honestly don't see that much difference between these two names :)

vpr/src/place/place_util.h
amin1377: The rest of the comment explains how the inner num is calculated.

vpr/src/place/place_util.cpp
amin1377: That’s a good way to decide whether to name a function calc_* or get_*. Thanks, Alex!

I think we shouldn’t print the log message inside the helper function. I’ve removed it from there and instead added the log in the annealer constructor where the function is called.

amin1377:
@soheilshahrouz: Thanks for reviewing the PR. Feel free to merge it if you don’t have any further suggestions.

Issue #2959, opened 2025-04-01
Expand y_range Search for ymax


amin1377:
In the `find_compatible_compressed_loc_in_range` function ([Link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/place/move_utils.cpp#L961-L1063)), after an x location is chosen (in the compressed grid), the function retrieves the y locations of the corresponding x location (where blocks of the given type exist in that column) and checks if search_range.ymin exists in the column. If it doesn't, the search range is expanded to include the entire column ([Link](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/place/move_utils.cpp#L1012-L1025)). I was wondering if expanding the search range to the entire column when ymax doesn't exist has been tried. The current implementation effectively moves IO blocks from the top edge to the bottom edge, but not vice versa.
@vaughnbetz: I can implement it if you think it is worth trying, and let you know the results.

vaughnbetz:
That sounds reasonable to me.  Not being able to move I/Os from the top to the bottom edge will not be good -- easy to get stuck. Seems worth trying.

PR #2960, opened 2025-04-01
[Place] Expand search range for sparse blocks

amin1377:
This PR addresses Issue #2959. The solution to fix the problem is a bit different from the one stated there, though. To ensure moving sparse blocks (e.g., IO blocks), we expand the search range to include the whole column if the number of compatible blocks in the given column is less than a certain threshold (currently, this number is set to 3)

The above update changed the placement of the top picture to the placement of the bottom one (where there is only one IO block left on the top).
Screenshot 2025-04-01 151404
*****
Screenshot 2025-04-01 150305


vpr/src/place/move_utils.cpp
```diff
@@ -1002,26 +1010,36 @@ bool find_compatible_compressed_loc_in_range(t_logical_block_type_ptr type,
...
+                    search_range.ymin = y_lower_iter->first;
+                    search_range.ymax = (y_upper_iter - 1)->first;
+                }
+            }
+        } else { // search_range is not fixed and there are less than G_MIN_NUM_BLOCKS_IN_COLUMN blocks at this x location
+            y_lower_iter = block_rows.begin();
+            y_upper_iter = block_rows.end();
+
+            search_range.ymin = y_lower_iter->first;
+            search_range.ymax = (y_upper_iter - 1)->first;
```
soheilshahrouz, line 1042: I think it is better to update the search range outside this function and leave this function as is. For example, `intersect_range_limit_with_floorplan_constraints()` updates the search range before this function is called. I suggest you add another function to modify the search range for this case as well.

vpr/src/base/vpr_types.h
```diff
@@ -442,6 +455,7 @@ struct t_bb {
     int ymax = OPEN;
     int layer_min = OPEN;
     int layer_max = OPEN;
+    bool is_fixed = false;
```
soheilshahrouz, line 458: I don't think this extra member variable is necessary. Instead of expanding the search range in `find_compatible_compressed_loc_in_range()`, you can modify it in another function before trying to find a compatible location.

vpr/src/place/move_utils.cpp
```diff
@@ -14,6 +14,14 @@
 #include "placer_state.h"
 #include "PlacerCriticalities.h"
 
+/**
+ * @brief If the number of blocks compatible with the moving block is less than this value,
+ * the search reagion is expanded to include all blocks in the column. It is specially useful
+ * for IO blocks which are on the perimeter of the device. This would allow the IO blocks to
+ * moved between top and bottom edges even when the rlim is small.
+ */
+size_t G_MIN_NUM_BLOCKS_IN_COLUMN = 3;
```
soheilshahrouz, line 23: Why is this global? If a single function uses it, it would be better to move it inside the function where it is used.

soheilshahrouz (commented):
I think the implementation can be made more modular by adding another function to modify the search limit instead of modifying when you're searching for a compatibale location.

soheilshahrouz:
Does the described problem happen only when IO blocks are located at the top/bottom rows? What if they are initially placed at left/right sides of the device?

amin1377:
> > Does the described problem happen only when IO blocks are located at the top/bottom rows? What if they are initially placed at left/right sides of the device?
> 
> I'm not entirely sure, but given that the x-axis of the compressed grid is fully dense, I don't think this happens for the x-axis.

Upon further discussion with Soheil, it seems likely that the same issue occurs along the x-axis. So, it would make sense to expand the x-axis range limit if the number of compatible blocks in a given row falls below a certain threshold. Otherwise, there’s a bias toward placing IO blocks on the top and bottom edges rather than on the left and right.

I’ll look into this in a separate PR.

@tpagarani FYI

amin1377:
@soheilshahrouz: I’ve addressed all your comments and implemented the changes you requested. Since the code has changed significantly since your last review, I’d appreciate it if you could take another look. Thanks!

vaughnbetz:
You should add QoR data on a couple of big benchmark suites.

vpr/src/place/move_utils.h
```diff
@@ -328,11 +328,12 @@ int find_empty_compatible_subtile(t_logical_block_type_ptr type,
  * is_median: true if this is called from find_to_loc_median
  * to_layer_num: the layer number of the new location (set by the caller)
  * search_for_empty: indicates that the returned location must be empty
+ * fixed_search_range: indicates that the search range is fixed and should not be adjusted
```
soheilshahrouz, line 331: remove this

vpr/src/place/move_utils.cpp
```diff
@@ -28,6 +28,51 @@ void set_placer_breakpoint_reached(bool flag) {
...
+ * 
+ * @return true if the search range was adjusted, false otherwise
+ */
+static bool adjust_search_range(t_logical_block_type_ptr block_type,
+                                ClusterBlockId block_id,
+                                t_bb& search_range,
+                                int& delta_cx,
+                                int to_layer_num) {
+
+    auto block_constrained = is_cluster_constrained(block_id);
```
soheilshahrouz, line 48: replacing auto with bool makes the code more readable

vaughnbetz:
Results on 3 seeds show the cpd degradation isn't consistent (the other 2 seeds were fine).  I think this is good to merge. Can you also add a link to the 3 seed data here for posterity @amin1377 ?

vaughnbetz:
It looks like some golden results need to be updated.  I looked at the parmys basic failure and it is a single small design that improved too much, so it's not a problem (actually it's good news):
[Fail]
k6_frac_N10_mem32K_40nm.xml/multiclock_separate_and_latch.v/common routed_wirelength relative value 0.4 outside of range [0.6,1.5], above absolute threshold 5.0 and not equal to golden value: 10.0

vpr/src/place/move_utils.cpp
```diff
@@ -28,6 +28,51 @@ void set_placer_breakpoint_reached(bool flag) {
...
+                                                                          to_layer_num);
+        if (!intersect) {
+            return false;
+        }
+    }
+
+    // TODO: Currently this is how we determine whether
+    // the moving block is of type IO. We need to have a function
+    // to infer IO type index (similar to what's done for CLBs)
+    if (block_type->index == 1 && !block_constrained) {
```
vaughnbetz, line 63: Shouldn't do this.
1. We have a function to check if a physical tile type is an IO type in physical_type_util.h (is_io_type).  You could use that to build a check of whether a logical block type is an IO type (by checking what blocks are compatible with that physical tile). Perhaps we already have such a function; if not we should make one as it is a useful utility. See https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/444b9ac72476768be95b4ebd9f7ec82b6b111054/vpr/src/place/initial_placement.cpp#L982 for an example of is_io_type.

2. Some devices have IOs in columns (area IO). You should check what happens on stratix 10, which has column IO, and I think you should test this code on it and make sure it makes sense (I suspect it should be modified or turned off in that case).  Possibly you should just change the whole test for whether you do this or not to check a condition on whether the current block type has a perimeter style, and you could precompute a table of which IO types have perimeter style locations with a preprocessing step that looks at the grid.

vpr/src/place/move_utils.cpp
```diff
@@ -28,6 +28,51 @@ void set_placer_breakpoint_reached(bool flag) {
     f_placer_breakpoint_reached = flag;
 }
 
+/**
+ * @brief Adjust the search range based on the block type and constraints
+ *
```
vaughnbetz, line 33: Should explain a bit about why and how you adjust the search range.

vaughnbetz (changes requested):
Some fragile code in there that I think should be changed.

vaughnbetz:
Here's a Stratix 10 architecture with two columns of IO: https://raw.githubusercontent.com/verilog-to-routing/vtr-verilog-to-routing/refs/heads/master/vtr_flow/arch/titan/stratix10_arch.timing.xml

vpr/src/place/move_utils.cpp
amin1377: Good point, thanks! I added a function to check for that and am now testing it to make sure it works as expected. I’ll post the results for both SIV and S10 once I have them.

vpr/src/place/move_utils.cpp
amin1377: Added the following:

```
* If the block is an IO block, we expand the search range cover the whole column
 * We found empirically that this is a good strategy for IO blocks given they are located in
 * the periphery for most FPGA architectures
```

vpr/src/place/move_utils.cpp
```diff
@@ -28,6 +28,52 @@ void set_placer_breakpoint_reached(bool flag) {
...
+            return false;
+        }
+    }
+
+    if (is_io_type(block_type) && !block_constrained) {
+        /* We empirically found that for the IO blocks,
+         * Given their sparsity, we expand the y-axis search range 
+         * to include all blocks in the column
+         */
+        const t_compressed_block_grid& compressed_block_grid = g_vpr_ctx.placement().compressed_block_grids[block_type->index];
```
vaughnbetz, line 69: You could make this code work well for area-IO (e.g. stratix 10 where IOs are in columns) and for other sparse blocks by just checking if the number of blocks in the compressed columns is less than some threshold.  I suspect you don't actually need to check if it is an IO block -- IOs are just an example of this kind of pattern.

(I still like the utility to check if a logical block type is an IO though).

vaughnbetz (approved):
I'm OK with this if S10 works well, but see my comment on removing the IO check and instead checking for < SPARSE_BLOCK_THRESHOLD (which you could set to 3 or so). Should be a little simpler and more general.

vpr/src/place/move_utils.cpp
amin1377: I actually tried that a while ago by not specifying the block type and simply expanding the search range if the number of blocks in the chosen column was less than 3. However, after making that change, in the example I mentioned earlier, we encountered the same issue where the IO blocks became scattered between the top and bottom edges.

vaughnbetz:
That’s strange. If you only tried one seed I suggest trying a few to see if this is a consistent effect. I’d also print out a message to check the code is activating.VaughnOn Apr 28, 2025, at 9:23 AM, Amin Mohaghegh ***@***.***> wrote:﻿
@amin1377 commented on this pull request.



In vpr/src/place/move_utils.cpp:
> +    if (block_constrained) {
+        bool intersect = intersect_range_limit_with_floorplan_constraints(block_id,
+                                                                          search_range,
+                                                                          delta_cx,
+                                                                          to_layer_num);
+        if (!intersect) {
+            return false;
+        }
+    }
+
+    if (is_io_type(block_type) && !block_constrained) {
+        /* We empirically found that for the IO blocks,
+         * Given their sparsity, we expand the y-axis search range 
+         * to include all blocks in the column
+         */
+        const t_compressed_block_grid& compressed_block_grid = g_vpr_ctx.placement().compressed_block_grids[block_type->index];

I actually tried that a while ago by not specifying the block type and simply expanding the search range if the number of blocks in the chosen column was less than 3. However, after making that change, in the example I mentioned earlier, we encountered the same issue where the IO blocks became scattered between the top and bottom edges.

amin1377:
@vaughnbetz
I tested five different seeds for both cases. With the changes in this branch, the IO blocks were placed close to each other in 3 out of 5 runs. After modifying the code to expand the search range only when the number of blocks in a column is less than 3, the IO blocks were placed close together in 1 out of 5 runs.

Unfortunately, I lost the S10 results due to some issues on Wintermute. I’m rerunning the benchmark now and will post the results as soon as they’re available.

vaughnbetz:
Thanks @amin1377 . Intriguing ... I would have thought the algorithms would be equivalent in that case.

amin1377:
> Thanks @amin1377 . Intriguing ... I would have thought the algorithms would be equivalent in that case.

Yes, I’m also a bit surprised that the results aren’t consistent across different seeds with the current change. I’ll need to dig into it a bit more, but I think this highlights that adding those weak links you mentioned earlier is likely a more reliable approach than relying on simulated annealing to place IO blocks close to each other.

amin1377:
Results on S10: [[Link](https://docs.google.com/spreadsheets/d/1YU-Ue48XifP83SG1zTvzBxASunV95XU2bXbZZdvg5oU/edit?usp=sharing)] Timing is basically tied (I also included the number of swaps and heap operations). Packing time increased by 10%, which I believe is due to machine load variation, as the packing code hasn’t changed in this PR. In terms of QoR, CPD increased by 1% while WL decreased by 1%. 

Overall, the QoR is pretty much a tie, though I have to admit, I like the current code better since it’s cleaner... maybe I'm just a little biased :)

amin1377:
Titan results: [Link ](https://docs.google.com/spreadsheets/d/1xZHS5p1z8le7uYsLZusYc16Qbe6mbGiZJm7dUAzR0X4/edit?usp=sharing)

VTR results: [Link](https://docs.google.com/spreadsheets/d/1XfmLUb8rBjboc1TIt23NotEf1WxTPx_0YNY5B1EqI60/edit?usp=sharing)

vpr/src/place/move_utils.cpp
```diff
@@ -19,6 +19,8 @@
 //Note: The flag is only effective if compiled with VTR_ENABLE_DEBUG_LOGGING
 bool f_placer_breakpoint_reached = false;
 
+constexpr int MIN_NUMBER_OF_BLOCK_PER_COLUMN = 3;
```
vaughnb-cerebras, line 22: This needs a comment.  Better name would be MIN_BLK_PER_COLUMN_EXPAND (or something like that). Current name makes it sound like we can't have fewer blocks than that in a column.
I also think this should be pushed down to the routine that uses it, if it is only used in one place.

vaughnb-cerebras (commented):
I have one minor feedback I think you can improve.
Other than that this is OK to merge.

amin1377:
@vaughnbetz: I made a few changes since there were some outliers in Titan results (CPD for two circuits increased by about 20%). I am gathering QoR now after these changes.

vpr/src/place/move_utils.cpp
amin1377: Moved the variable to the routine and added the following comment:
```
// If the number of blocks in a column is less than this number, we 
// will expand the search range to the whole column
```

vaughnbetz:
LGTM, thanks!

vaughnbetz:
There are a few CI failures. I looked at the strong failures and it is 4 small circuits with larger than bounds QoR changes, all of which look harmless. So you can check in new golden results, or widen the ranges for those smaller tests if you prefer. Then it's good to merge.

amin1377:
> There are a few CI failures. I looked at the strong failures and it is 4 small circuits with larger than bounds QoR changes, all of which look harmless. So you can check in new golden results, or widen the ranges for those smaller tests if you prefer. Then it's good to merge.

Thanks, Vaughn! I'm running the nightly tests to check if there are any results that need updating. I'll merge the PR afterward.

amin1377:
@vaughnbetz: This PR is ready to merge.

amin1377:
@vaughnbetz: Should I go ahead and merge this PR?

PR #2965, opened 2025-04-06
[Pack][Timing] Abstracted How Timing is Used in the Packer

AlexandreSinger:
Timing was intermixed into the packer. It appears as though the code originally was designed to recalculate the timing information every so often in the packer, but the idea was abandoned. This left timing code in disperse locations around the Packer and the timing was being recomputed every time clustering was restarted which was unecessary.

Collecting all of the timing information from the Packer into a single object called PreClusterTimingManager which abstracts all of the timing info in the Packer.

The ultimate goal is to bring this Manager class into the AP flow to be used together with the Global Placer. By sharing this manager class, the AP flow may be able to update the timing info with flat placement information to make the timing more accurate.

### DIFF
--- a/vpr/src/timing/PreClusterTimingManager.cpp
+++ b/vpr/src/timing/PreClusterTimingManager.cpp
@@ -1,0 +1,276 @@
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    April 2025
+ * @brief   Implementation of the pre-cluster timing manager class.
+ */
+
+#include "PreClusterTimingManager.h"
+#include <algorithm>
+#include <memory>
+#include "PreClusterDelayCalculator.h"
+#include "PreClusterTimingGraphResolver.h"
+#include "SetupGrid.h"
+#include "atom_lookup.h"
+#include "atom_netlist.h"
+#include "atom_netlist_fwd.h"
+#include "concrete_timing_info.h"
+#include "physical_types_util.h"
+#include "prepack.h"
+#include "tatum/TimingReporter.hpp"
+#include "tatum/echo_writer.hpp"
+#include "vpr_types.h"
+#include "vtr_assert.h"
+
+/**
+ * Since the parameters of a switch may change as a function of its fanin,
+ * to get an estimation of inter-cluster delays we need a reasonable estimation
+ * of the fan-ins of switches that connect clusters together. These switches are
+ * 1) opin to wire switch
+ * 2) wire to wire switch
+ * 3) wire to ipin switch
+ * We can estimate the fan-in of these switches based on the Fc_in/Fc_out of
+ * a logic block, and the switch block Fs value
+ */
+static void get_intercluster_switch_fanin_estimates(const t_arch& arch,
+                                                    const t_det_routing_arch& routing_arch,
+                                                    const std::string& device_layout,
+                                                    const int wire_segment_length,
+                                                    int* opin_switch_fanin,
+                                                    int* wire_switch_fanin,
+                                                    int* ipin_switch_fanin);
+
+static float get_arch_switch_info(short switch_index, int switch_fanin, float& Tdel_switch, float& R_switch, float& Cout_switch);
+
+static float approximate_inter_cluster_delay(const t_arch& arch,
+                                             const t_det_routing_arch& routing_arch,
+                                             const std::string& device_layout);
+
+PreClusterTimingManager::PreClusterTimingManager(bool timing_driven,
+                                                 const AtomNetlist& atom_netlist,
+                                                 const AtomLookup& atom_lookup,
+                                                 const Prepacker& prepacker,
+                                                 e_timing_update_type timing_update_type,
+                                                 const t_arch& arch,
+                                                 const t_det_routing_arch& routing_arch,
+                                                 const std::string& device_layout,
+                                                 const t_analysis_opts& analysis_opts) {
+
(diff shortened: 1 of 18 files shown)


vaughnbetz:
Should do a QoR run to check it is all good (run Titan and/or VTR and confirm equivalent QoR or no change).  Linking a spreadsheet would be best.

vpr/src/timing/PreClusterTimingManager.h
```diff
@@ -0,0 +1,99 @@
...
+                            const AtomLookup& atom_lookup,
+                            const Prepacker& prepacker,
+                            e_timing_update_type timing_update_type,
+                            const t_arch& arch,
+                            const t_det_routing_arch& routing_arch,
+                            const std::string& device_layout,
+                            const t_analysis_opts& analysis_opts);
+
+    /**
+     * @brief Calculates the setup criticality of the given primitive block.
```
vaughnbetz, line 64: Define what the setup criticality of a block is.  Currently defined as the maximum criticality over the block inputs.

vpr/src/timing/PreClusterTimingManager.h
```diff
@@ -0,0 +1,99 @@
...
+     */
+    float calc_atom_setup_criticality(AtomBlockId blk_id,
+                                      const AtomNetlist& atom_netlist) const;
+
+    /**
+     * @brief Returns true of the pre-cluster timing manager was initialized
```
vaughnbetz, line 70: Returns whether or not the pre-cluster timing manager was initialized

vpr/src/timing/PreClusterTimingManager.cpp
```diff
@@ -0,0 +1,278 @@
...
+        tatum::write_echo(getEchoFileName(E_ECHO_PRE_PACKING_TIMING_GRAPH),
+                          *timing_ctx.graph, *timing_ctx.constraints, *clustering_delay_calc_, timing_info_->analyzer());
+
+        tatum::NodeId debug_tnode = id_or_pin_name_to_tnode(analysis_opts.echo_dot_timing_graph_node);
+        write_setup_timing_graph_dot(getEchoFileName(E_ECHO_PRE_PACKING_TIMING_GRAPH) + std::string(".dot"),
+                                     *timing_info_, debug_tnode);
+    }
+
+    // Write a timing report.
+    // TODO: Is this a good place for this? Or should it be moved to just after
```
vaughnbetz, line 96: I'd leave this ... this gives an early timing report with a very simple 'net-type' delay model.

vpr/src/timing/PreClusterTimingManager.h
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * @brief Pre-cluster timing manager class.
+ *
+ * This class encapsulates the timing computations used prior to clustering.
+ * This maintains all of the state necessary to perform these timing computations.
+ */
+class PreClusterTimingManager {
+  public:
+    /**
+     * @brief Constructor for the manager class. Builds the timing graph.
```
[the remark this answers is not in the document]
vaughnbetz: Some of the comment above could go in the delay calculator -- probably best to split it.

vaughnbetz (approved):

vpr/src/timing/PreClusterTimingManager.h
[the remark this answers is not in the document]
AlexandreSinger: @vaughnbetz After reviewing the code, I am still not sure if the carry-chains are being used in the timing analyzer directly. When computing the max edge delay, the code performs the following:
image
If the edge between two atoms in a carry chain are not labelled at combinational, then the inter_cluster_net_delay will be used instead.

The code that labels these edges is (I think) created by the TimingGraphBuilder, which builds the timing graph as part of initializing VPR (prior to packing). Thus, it could not know that the edge between two molecules within the same molecules are combinational (also, based on the name, it wouldn't make sense for them to be labelled combinational anyways). This implies to me that the delays within molecules are not being observed when they should. Someone should check the timing reports to ensure that this is correct.

If this is true, a simple fix to this would be to check if the edge exists within a molecule, and if so use the delay of the edge within the molecule instead of the interconnect delay (or at least the worst case of the molecule. This is because, if an atom was packed into a molecule, it must be packed together (as far as I am aware).

Should I open an issue on this? It's not directly in my path, and could be a good thing for someone else to look into.

AlexandreSinger:
QoR results on Titan:

  | baseline | pk_timing_mngr
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.997409372882633
num_LAB | 1 | 1
num_DSP | 1 | 1
num_M9K | 1 | 1
num_M144K | 1 | 1
max_vpr_mem | 1 | 1.00241342168808
num_pre_packed_blocks | 1 | 1
device_grid_tiles | 1 | 1
pack_time | 1 | 1.00468192080581
placed_wirelength_est | 1 | 1
place_time | 1 | 1.00447348452275
placed_CPD_est | 1 | 1
routed_wirelength | 1 | 1
critical_path_delay | 1 | 1
crit_path_route_time | 1 | 1.00336920535485

There was absolutely no change in quality, and the change in memory / runtime appear to be just noise from the machine.

Full script:
[comparison_output.xlsx](https://github.com/user-attachments/files/19690374/comparison_output.xlsx)

Looks good to merge.

vaughnb-cerebras:
Thanks!  The hard data makes me feel more at ease :).

PR #2970, opened 2025-04-10
[OpenFPGA] Update

amin1377:
Update OpenFPGA branch with the master branch


tangxifan:
@amin1377 can you avoid rebase and use merge instead? The commit history is now complicated. Thanks!

PR #2971, opened 2025-04-11
[AP][Test] Added Titan Nightly Test of WL-Driven AP Flow

AlexandreSinger:
Created a nightly test for the WL-driven AP flow on Titan. This will help prevent regression on these benchmarks.

For this NightlyTest I chose to fix the channel width at 300 (matching the other Titan tests) and I left the IO unfixed.

For the parsing and requirement files, I used the same files as Titan. This will catch the major changes (like WL and CPD), but wont get tripped if the Global Placement HPWL got worst all of a sudden. This should prevent us from having to update this test too often as we are updating the AP flow.


AlexandreSinger:
@amin1377 Do you mind reviewing this today if you have time? I would like to ensure the test works on the NightlyTestManual run tomorrow. Its a very short PR that just adds a WL-driven AP flow regression test on the Titan circuits.

amin1377 (approved):
Looks good, Alex! Thanks!

AlexandreSinger:
Thank you!

PR #2975, opened 2025-04-13
[AP][InitialPlacement] Improved Initial Placement

AlexandreSinger:
Found that the Initial Placer stage of the AP flow (after APPack, but before Detailed Placement) was not working as expected. The intention was that clusters would be placed at their centroid location accordin to the flat placement, and if that site was illegal or taken it would take a nearby point instead (falling back on the original initial placer if nothing can be found).

To achieve this, I was using a method called find_centroid_neighbor which I thought would return the nearest legal location to the given location. This was not correct. This method just creates a bounding-box and tries to find a random point in that box around the given point. This was causing our AP flow to move clusters WAY farther than they wanted, which moved them into places other clusters wanted to go. This was also not exhaustive, so it was often falling back on the original approach which was putting clusters in practically random locations. All of this was causing the post-FL placement from the AP flow to actually have worse quality than the default AP flow!

To resolve this, I wrote the actual method I was intending. It performs a BFS-style search from the src location to all legal locations and returns the closest one. By doing this BFS on the compressed grid, I found that this is actually quite efficient. With these changes, I found that the quality of the post-FL placement more than doubled and the average atom displacement from the GP solution decrease dramatically.

Results on the 9 largest VTR benchmarks (with fixed IO):

  | Post-FL WL | Post-Route WL | Total Runtime
-- | -- | -- | --
No AP | 1 | 1 | 1
AP Baseline | 1.151589662 | 0.9310077442 | 1.477490535
AP Improved IP | 0.5489232291 | 0.9195574479 | 1.352550022

As we can see, the quality of the post-FL placement was actually 15% worse on the VTR benchmark that the non-AP flow. This did not make sense since we are giving it more detailed information on where blocks should be placed relative to the objectives. After this change, things make a lot more sense. The quality of the post-FL placement is now 2x better than the non-AP flow. This translated to another 1% improvement in WL. The run time improved slightly; however, I think this may just be machine load. When observing Titan (which I am running now), I find that the FL stage increases in run time (due to IP taking a bit longer) and the DP decreases very slightly.

Some more information comparing the old AP implementation to the new:
  | AP Baseline | AP Improved IP | Change Over Baseline
-- | -- | -- | --
Post-FL WL | 739521.555 | 352504.519 | 0.477
Post-Route WL | 197028.628 | 194605.409 | 0.988
Post-FL Avg. Atom Displ. | 29.324 | 10.492 | 0.358
Post-FL Max Atom Displ. | 107.180 | 69.581 | 0.649
FL Run Time | 42.818 | 40.380 | 0.943
DP Run Time | 18.556 | 13.450 | 0.725
Total Run Time | 106.627 | 97.611 | 0.915

We can see that the average atom displacement is 3x better now (being 10.5 tiles on average which is much more reasonable). The max atom displacement also greatly decreased. For the Titan results the FL and DP run times appear to have decreased; however, I would trust the Titan results once I get them!

I think we can see even more gains if we tune the annealer for the AP flow. Even though the quality of the placement is 2x better, the runtime of the annealer did not change very much and the quality did not change as much as you would expect. This should be investigated.


vpr/src/place/initial_placement.cpp
```diff
@@ -549,47 +547,188 @@ static std::vector<ClusterBlockId> find_centroid_loc(const t_pl_macro& pl_macro,
...
+    VTR_ASSERT(num_layers == 1);
+
+    // Get the closest (approximately) compressed location to the src location.
+    // This does not need to be perfect (in fact I do not think it is), but the
+    // closer it is, the faster the BFS will find the best solution.
+    t_physical_tile_loc src_grid_loc(src_flat_loc.x, src_flat_loc.y, src_flat_loc.layer);
+    const t_physical_tile_loc compressed_src_loc = compressed_block_grid.grid_loc_to_compressed_loc_approx(src_grid_loc);
+
+    // Weighted-BFS search the compressed grid for an empty compatible subtile.
+    size_t num_rows = compressed_block_grid.get_num_rows(0);
```
vaughnbetz, line 621: I suggest setting layer = 0 and then passing layer to the get_num_rows and get_num_columns calls to make it more clear what you're doing, and to make it a bit easier to update to multi-layer later.

vpr/src/place/initial_placement.cpp
```diff
@@ -549,47 +547,188 @@ static std::vector<ClusterBlockId> find_centroid_loc(const t_pl_macro& pl_macro,
...
+        if (visited[loc.x][loc.y])
+            continue;
+        visited[loc.x][loc.y] = true;
+
+        // Get the distance from this loc to the src_loc in grid space
+        // Note: In compressed space, distances are not what they appear. We are
+        //       using the true grid positions to get the truly closest loc.
+        //       Here, we use the flat centroid of the macro and measure its
+        //       distance from the center of this grid loc (hence the 0.5 offset).
+        auto grid_loc = compressed_block_grid.compressed_loc_to_grid_loc(loc);
```
vaughnbetz, line 644: I wonder if anything special needs to be done for larger tiles (e.g. a DSP block that spans multiple rows and columns). The compressed loc will point at the anchor point for that block (lower left corner) rather than the middle of the physical block.  I suggest at least commenting that you're not worrying about this for now, if you're not quite sure if there's anything better to do.  Possibly you should consider the true grid location to be in the middle of the physical tile, which would mean adding width/2 + 0.5, height/2 + 0.5 to the grid_loc.x and grid_loc.y.

vpr/src/place/initial_placement.cpp
```diff
@@ -549,47 +547,188 @@ static std::vector<ClusterBlockId> find_centroid_loc(const t_pl_macro& pl_macro,
...
+        //       have a higher distance than the current node.
+        if (grid_dist >= best_dist)
+            continue;
+
+        // In order to ensure our BFS finds the closest compatible location, we
+        // traverse compressed grid locations which may not actually be valid
+        // (i.e. no tile exists there). This is fine, we just need to check for
+        // them to ensure we never try to put a cluster there.
+        bool is_valid_compressed_loc = false;
+        const auto& compressed_col_blk_map = compressed_block_grid.get_column_block_map(loc.x, 0);
```
vaughnbetz, line 662: Again, layer would be better than 0 here.

vpr/src/place/initial_placement.cpp
```diff
@@ -549,47 +547,188 @@ static std::vector<ClusterBlockId> find_centroid_loc(const t_pl_macro& pl_macro,
...
+            for (int sub_tile : compatible_sub_tiles) {
+                if (blk_loc_registry.grid_blocks().is_sub_tile_empty(grid_loc, sub_tile)) {
+                    new_sub_tile = sub_tile;
+                    break;
+                }
+            }
+            if (new_sub_tile != -1) {
+                // If a sub-tile is available, set this to be the first sub-tile
+                // available and check if this site is legal for this macro.
+                // TODO: Should we pick a random sub-tile instead?
```
vaughnbetz, line 682: First should be fine ... there's no spatial meaning of difference between the subtiles at a certain (x,y,layer).

vpr/src/place/initial_placement.cpp
```diff
@@ -549,47 +547,188 @@ static std::vector<ClusterBlockId> find_centroid_loc(const t_pl_macro& pl_macro,
...
+                    //       may present a location which is closer in compressed
+                    //       space earlier than a location which is closer in
+                    //       grid space.
+                    best_dist = grid_dist;
+                    best_loc = new_loc;
+                }
+            }
+        }
+
+        // Push the neighbors (in the compressed grid) onto the queue.
```
vaughnbetz, line 704: I think this is worth a bigger comment.  I think you're searching in the compressed grid space, and it's OK if you ask for a compressed location that doesn't exist (which I think the code below can do), as the code that finds the closest grid tile will find a legal spot, and if you already explored it the code at the top of the loop will skip it.

vpr/src/place/initial_placement.cpp
```diff
@@ -549,47 +547,188 @@ static std::vector<ClusterBlockId> find_centroid_loc(const t_pl_macro& pl_macro,
...
+        if (loc.x > 0) {
+            t_physical_tile_loc new_loc = t_physical_tile_loc(loc.x - 1,
```
vaughnbetz, line 706: I'd work comp_ or compressed_ into the name of this variable (e.g. new_comp_loc or new_compressed_loc) as it is easy to get confused as to what grid we're looking things up in.

vpr/src/place/initial_placement.cpp
```diff
@@ -549,47 +547,188 @@ static std::vector<ClusterBlockId> find_centroid_loc(const t_pl_macro& pl_macro,
...
+        // (i.e. no tile exists there). This is fine, we just need to check for
+        // them to ensure we never try to put a cluster there.
+        bool is_valid_compressed_loc = false;
+        const auto& compressed_col_blk_map = compressed_block_grid.get_column_block_map(loc.x, 0);
+        if (compressed_col_blk_map.count(loc.y) != 0)
+            is_valid_compressed_loc = true;
+
+        // If this distance is better than the best we have seen so far, try
+        // to see if this is a better solution.
+        if (is_valid_compressed_loc) {
```
vaughnbetz, line 668: This is a fairly long routine. The logic is pretty clean, but I wonder if there's a helper function you can extract to shorten it? The code below for example could be a small helper function to check if there is a sub_tile at a location.

vaughnbetz (commented):
Great results!  A few minor comments on the code for you to consider.

AlexandreSinger:
Titan Results (No fixed IO, channel width = 300):

  | Post-FL WL | Post-Route WL | Total Runtime
-- | -- | -- | --
No AP | 1.000 | 1.000 | 1.000
AP Baseline | 1.176 | 0.923 | 1.495
AP Improved IP | 0.356 | 0.905 | 1.485

More information on the AP improvements:
  | AP Baseline | AP Improved IP | Change Over Baseline
-- | -- | -- | --
Post-FL WL | 1.89E+07 | 5.73E+06 | 0.303
Post-Route WL | 2.64E+06 | 2.59E+06 | 0.980
Post-FL Avg. Atom Displ. | 63.735 | 14.084 | 0.221
Post-FL Max Atom Displ. | 345.646 | 199.538 | 0.577
FL Run Time | 659.955 | 667.575 | 1.012
DP Run Time | 631.548 | 598.406 | 0.948
Total Run Time | 2534.326 | 2517.297 | 0.993

On Titan, these changes made the quality of the post-legalized solution 3x better; with the AP flow now outperforming the baseline no AP glow by around 3x in quality before placement. This further reduced the post-routed WL by around 2%, making the AP flow almost 10% better in WL over non-AP.

Similar to the VTR results, the average atom displacement has been reduced by over 4x and the max atom displacement was reduced by 40%. The FL runtime increased slightly due to the increased logic in the initial placer, but looking through the results, directrf appears to have taken the longest for initial placement at ~160 seconds. I am not concerned about this since other large circuits like bitcoinminer took only 4 seconds. I think that directrf may just have a very illegal global placement solution.

The detailed placement run time reduced by 5%, which I find a bit suprising since I would expect the DP runtime to reduce a lot due to the improved initial placement quality. For many of the circuits, I am noticing that the annealer tends to make the placement solution worse, and not returning to the original quality after around 10 iterations (some circuits are less). For example, bitcoin_miner looks like this:
image
I think the starting temperature may be too high which is costing run time and perhaps even quality.

Overall this change is a win across the board, which is good news!

vaughnbetz:
Looks good.  I agree; the wirelength bump is big at the start of annealing. You could try quenching (for a long time) instead, or (as you suggest) a lower initial T.
Looks like these are non-timing driven, so the cost increase can't be due to the annealer trying to fix up timing.

vpr/src/place/initial_placement.cpp
AlexandreSinger: This was a really really good point Vaughn, thank you so much for pointing this out. You are correct that this 0.5 offset only works if all tiles are 1x1. I completely forgot about big tiles!

I have changed the calculation of distance to be as follows: distance is the amount the src_flat_loc would need to move in order to be within the given tile. So, if the src_flat_loc is within the bounds of the tile, its distance would be zero. If src_flat_loc is 1 tile away in the x-dimension from the edge of the tile, it would have a distance of 1. Since we are doing math with L1 distances, this is actually really easy and cheap to compute.

I am testing it locally on VTR and am finding that it does improve things, albeit very slightly. The post-routed WL went up on some circuits and down on others, but in general the post-legalized placement seems to always decrease. I do not think it is worth re-collecting the results on Titan, but this was a good catch and makes this method much more general!

PR #2981, opened 2025-04-15
Add Github action to close stale issues

AmirhosseinPoolad:
We currently have too many old and abandoned issues, which makes the issues tab not very important for maintainers. 

The added workflow will close up to 30 old issues every day, starting from the oldest ones. Issues that have been inactive for more than a year will be first marked as stale, and if they remain stale after 15 days they will be automatically closed. If for any reason you want an issue to be exempt from being close by this workflow, you can mark it as 'no-stale'. The workflow runs every day at 1PM UTC or 9AM Toronto time.

I have commented out some code that does the same thing for PRs but I think it might be too much currently and we only agreed to automatically close old issues, not PRs. I left it in there in case the situation is different in the future.

Since there's a limit of 30 issues per day because of the GitHub API restrictions, this should take many weeks to go through all of the issues and would dampen the change by a good amount, which is a good thing in my opinion.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,31 @@
+name: 'Close Stale Issues'
+on:
+  schedule:
+    # Run everyday at 1 PM UTC
+    - cron: '0 13 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          # The message to be shown for stale issues
+          stale-issue-message: 'This issue has been inactive for a year and has been marked as stale. It will be closed in 15 days if it continues to be stale. If you believe this is still an issue, please add a comment.'
+          close-issue-message: 'This issue has been marked stale for 15 days and has been automatically closed.'
+          # If you want to exempt an issue from being marked stale/deleted, label it as 'no-stale'
+          exempt-issue-labels: 'no-stale'
+          days-before-issue-stale: 365
+          days-before-issue-close: 15
+          # Start from the oldest issues
+          ascending: true
+          
+          # The configuration below can be used to allow the same behaviour with PRs.
+          # Since we currently don't want to close old PRs, it is commented out but 
+          # left here in case we change our mind.
+
+          # stale-pr-message: 'This PR has been inactive for a year and has been marked as stale. It will be closed in 15 days if it continues to be stale. If you are still working on this PR, please add a comment.'
+          # close-pr-message: 'This PR has been marked stale for 15 days and has been automatically closed.'
+          # exempt-pr-labels: 'no-stale'
+          # days-before-pr-stale: 365
+          # days-before-pr-close: 15
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,7 +77,7 @@ This information helps us to quickly reproduce (and hopefully fix) the issue:
 
     Tell us what version of VTR you are using (e.g. the output of `vpr --version`), which Operating System and compiler you are using, or any other relevant information about where or how you are building/running VTR.
 
-Once you've gathered all the information [open an Issue](https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/new?template=bug_report.md) on our issue tracker.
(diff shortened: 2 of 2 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          # The message to be shown for stale issues
+          stale-issue-message: 'This issue has been inactive for a year and has been deemed stale. It will be closed in 15 days if it continues to be stale. If you believe this is still an issue, please add a comment.'
+          close-issue-message: 'This issue has been marked stale for 15 days and is automatically closed.'
```
AlexandreSinger, line 15: Nit: tense. I think it should be "and has been automatically closed". Or it should be "and is automatically being closed".

I wonder if you should add steps to this message about how people can reopen the issue if they think it is still relevant.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+    - cron: '0 13 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          # The message to be shown for stale issues
+          stale-issue-message: 'This issue has been inactive for a year and has been deemed stale. It will be closed in 15 days if it continues to be stale. If you believe this is still an issue, please add a comment.'
```
AlexandreSinger, line 14: I personally do not like the word "deemed" here. If the person who raised this issue is not from VTR, it may seem like we developers deemed it stale. I recommend just saying "and has been marked as stale".

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
+name: 'Close stale issues'
```
AlexandreSinger, line 1: Nitpick: Our capitalization convention for these actions is title case. For example:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/661debee9102027a8f813e43db133a7c7e554dd7/.github/workflows/labeler.yml#L1

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+          stale-issue-message: 'This issue has been inactive for a year and has been deemed stale. It will be closed in 15 days if it continues to be stale. If you believe this is still an issue, please add a comment.'
+          close-issue-message: 'This issue has been marked stale for 15 days and is automatically closed.'
+          # If you want to exempt an issue from being marked stale/deleted, label it as 'no-stale'
+          exempt-issue-labels: 'no-stale'
+          days-before-issue-stale: 365
+          days-before-issue-close: 15
+          # Start from the oldest issues
+          ascending: true
+          
+          # The configuration below can be used to have the allow the same behaviour with PRs.
```
AlexandreSinger, line 23: "used to have the allow"

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+          # Since we currently don't want to close old PRs, it is commented out but left here
+          # in case we change our mind.
+
+          # stale-pr-message: 'This PR has been inactive for a year and has been deemed stale. It will be closed in 15 days if it continues to be stale. If you believe are still working on this PR, please add a comment.'
```
AlexandreSinger, line 27: Same comment as above about "deemed". Even though this has been commented out, we should still update it as if it is being used.

AlexandreSinger (commented):
Hi Amir, this is a great change! I just left some minor comments on wording.

AlexandreSinger:
@AmirhosseinPoolad Lets bring this up in the weekly industry sync meeting to see what people think!

.github/workflows/stale.yml
AmirhosseinPoolad: Good catch!

.github/workflows/stale.yml
AmirhosseinPoolad: Agreed, fixed.

.github/workflows/stale.yml
AmirhosseinPoolad: Haha oops. I guess I forgot to delete the old text when I reworded the message.

PR #2984, opened 2025-04-16
Remove atom_net global context mutation from packer

AmirhosseinPoolad:
While making the packer parallel, we removed these writes to the global context and nothing went wrong. This WIP PR does the same to VTR master. If these mutations are not needed, removing them will make the code cleaner (less "why is it doing this?") and could potentially make packing a bit faster.

The reason I think things should be mostly fine (other than passing the strong tests) is that atom_net lookup data is only ever changed after packing (add_atom_clb_net is only called after packing is done) and atom_clb is the same (set_atom_clb is only called after packing).

### DIFF
--- a/vpr/src/pack/pack.cpp
+++ b/vpr/src/pack/pack.cpp
@@ -250,12 +250,8 @@ bool try_pack(const t_packer_opts& packer_opts,
             VPR_FATAL_ERROR(VPR_ERROR_OTHER, "Failed to find device which satisfies resource requirements required: %s (available %s)", resource_reqs.c_str(), resource_avail.c_str());
         }
 
-        //Reset clustering for re-packing
-        for (auto net : g_vpr_ctx.atom().netlist().nets()) {
-            g_vpr_ctx.mutable_atom().mutable_lookup().remove_atom_net(net);
-        }
+        //Reset floorplanning constraints for re-packing
         g_vpr_ctx.mutable_floorplanning().cluster_constraints.clear();
-        //attraction_groups.reset_attraction_groups();
 
         // Reset the cluster legalizer for re-clustering.
         cluster_legalizer.reset();


vpr/src/base/clustered_netlist.cpp
```diff
@@ -170,8 +170,15 @@ ClusterNetId ClusteredNetlist::create_net(const std::string& name) {
 }
 
 void ClusteredNetlist::remove_block_impl(const ClusterBlockId blk_id) {
+    AtomPBBimap& atom_pb_bimap = g_vpr_ctx.mutable_atom().mutable_lookup().mutable_atom_pb_bimap();
+    AtomBlockId atom_blk_id = atom_pb_bimap.pb_atom(block_pbs_[blk_id]);
```
AlexandreSinger, line 174: This is not going to work how you would expect. pbs are stored as a hierarchy. The pb found in block_pbs_ is the root pb for the entire cluster. pb_atom will only return the atom_blk_id associated with a leaf pb (i.e. a primitive). Since you are passing in a root pb, this will never return an atom.

Clusters actually contain many atoms, not just one.

I would recommend iterating over the lookup for this cluster and collect all of the atom_blk_ids, then set all of the atoms to an invalid cluster. Or even better, since this is a bimap I think, you may be able to just tell the bi-map to invalidate all atoms which point to a given cluster blk id.

vpr/src/base/clustered_netlist.cpp
```diff
@@ -170,8 +170,15 @@ ClusterNetId ClusteredNetlist::create_net(const std::string& name) {
...
+
+    // Remove the corresponding atom_clb lookup entry from global context (if exsits)
```
AlexandreSinger, line 176: exists*

AlexandreSinger (changes requested):

vpr/src/base/clustered_netlist.cpp
AmirhosseinPoolad: Ah yes I forgot that free_pb is recursive. Yeah I think in that case it's worth rethinking this from a code organization stand point, since we'll be basically duplicating the tree traversal logic which I don't really feel good about. I think I'll just remove the atom_net mutation in pack.cpp for this PR and leave the free_pb changes for another PR since it'll need more extensive reworks to do properly.

AmirhosseinPoolad:
Hey @AlexandreSinger, I removed the free_pb changes and I think this can be merged, can you do another review? Thanks a lot!

AlexandreSinger (approved):
LGTM. Can you please raise an issue regarding the other code so we can handle it in the future?

PR #2987, opened 2025-04-17
Change some internal packer APIs to not use C-style arrays

AmirhosseinPoolad:
This commit changes some functions in the packer and prepacker that used C-style arrays to use std::vector instead. Previously we used the .data() method of std::vector to pass a pointer to these functions.

### DIFF
--- a/vpr/src/pack/prepack.cpp
+++ b/vpr/src/pack/prepack.cpp
@@ -53,13 +53,13 @@ static t_pb_graph_edge* find_expansion_edge_of_pattern(const int pattern_index,
                                                        const t_pb_graph_node* pb_graph_node);
 
 static void forward_expand_pack_pattern_from_edge(const t_pb_graph_edge* expansion_edge,
-                                                  t_pack_patterns* list_of_packing_patterns,
+                                                  std::vector<t_pack_patterns>& list_of_packing_patterns,
                                                   const int curr_pattern_index,
                                                   int* L_num_blocks,
                                                   const bool make_root_of_chain);
 
 static void backward_expand_pack_pattern_from_edge(const t_pb_graph_edge* expansion_edge,
-                                                   t_pack_patterns* list_of_packing_patterns,
+                                                   std::vector<t_pack_patterns>& list_of_packing_patterns,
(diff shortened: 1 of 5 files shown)


vpr/src/pack/prepack.cpp
```diff
@@ -610,7 +610,7 @@ static void forward_expand_pack_pattern_from_edge(const t_pb_graph_edge* expansi
  *               destination blocks
  */
 static void backward_expand_pack_pattern_from_edge(const t_pb_graph_edge* expansion_edge,
-                                                   t_pack_patterns* list_of_packing_patterns,
+                                                   std::vector<t_pack_patterns>& list_of_packing_patterns,
```
soheilshahrouz, line 613: It seems that only the element specified with `curr_pattern_index` is accessed in `backward_expand_pack_pattern_from_edge()` and `forward_expand_pack_pattern_from_edge()`. I suggest you only pass the element modified by these functions instead the entire vector. If the index is needed, you can use `t_pack_patterns::index`.

AmirhosseinPoolad: Makes sense. Fixed.

vpr/src/pack/prepack.cpp
```diff
@@ -136,7 +134,7 @@ static void print_chain_starting_points(t_pack_patterns* chain_pattern);
  */
 static std::vector<t_pack_patterns> alloc_and_load_pack_patterns(const std::vector<t_logical_block_type>& logical_block_types) {
     int L_num_blocks;
-    std::vector<t_pack_patterns> list_of_packing_patterns;
+    std::vector<t_pack_patterns> packing_patterns;
```
AlexandreSinger, line 137: Minor nitpick: This is an old C-style where they declared all variables at the top of the function before using them. Could you move the declaration for packing patterns to where the alloc happens.

AlexandreSinger (commented):
LGTM. Just one minor comment.

vpr/src/pack/prepack.cpp
AmirhosseinPoolad: Thanks for the review, fixed this.

PR #2988, opened 2025-04-17
[AP][InitialPlacement] Created Isolated AP Flow

AlexandreSinger:
The old Initial Placer used in the AP flow was constructed within the initial placer of the non-AP flow. This forced the AP flow to try to place blocks one at a time with minimum displacement. This is non-ideal since blocks that were placed earlier were being getting first picks at locations, which may displace a future cluster which may be a better fit for that location.

Separated out the AP initial placement code. For AP, initial placement is done in passes.

The first pass will try to place clusters exactly at the tile that the centroid of all atoms within the cluster want to be placed (according to the global placement). Any clusters that could not be placed are reserved for the next pass.

The second pass will allow clusters to be placed within 1 tile of their centroid.

All subsequent passes will allow cluster to be placed exponentially farther from their centroid.

The initial placement terminates when all clusters have been placed or if the max displacement is the size of the entire device.

The clusters are sorted based on the size of the macro that contains them and the variance of the placement of the atoms within the macro. This allows large macro blocks with low variance to be placed first.

Results on the largest VTR circuits (fixed IOs):
Metric|Change
-- | --
Normalized Post FL WL | 0.947
Normalized Post-Route WL | 1.003
Normalized Atom Errors | 0.922
Normalized Atom Displacement | 0.950
Normalized Max Atom Displacement | 1.047

This improved the initial placement quality by around 5%, the amount of atom errors (where atoms are placed in a tile they do not want to be placed in according to the global placement) went down by around 8%. Atom displacement improved by 5%. The max atom displacement got worst by around 5%. I think this increase in max displacement is ok. I will collect Titan results to verify this.

### DIFF
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -1,7 +1,9 @@
+#include "clustered_netlist.h"
 #include "flat_placement_types.h"
 #include "atom_netlist_fwd.h"
 #include "physical_types_util.h"
 #include "place_macro.h"
+#include "vtr_assert.h"
 #include "vtr_geometry.h"
 #include "vtr_ndmatrix.h"
 #include "vtr_random.h"
@@ -18,6 +20,7 @@
 #include "move_utils.h"
 #include "region.h"
 #include "noc_place_utils.h"
+#include "vtr_vector.h"
 
 #include <algorithm>
 #include <cmath>
@@ -590,10 +593,10 @@ static inline float get_dist_to_tile(const t_flat_pl_loc& src_flat_loc,
     // Note: The get_tile_bb function will not work in this case since it
     //       subtracts 1 from the width and height.
     auto tile_type = device_grid.get_physical_type(tile_loc);
-    float tile_xmin = tile_loc.x;
-    float tile_xmax = tile_loc.x + tile_type->width;
-    float tile_ymin = tile_loc.y;
-    float tile_ymax = tile_loc.y + tile_type->height;
+    float tile_xmin = tile_loc.x - device_grid.get_width_offset(tile_loc);
+    float tile_xmax = tile_xmin + tile_type->width;
+    float tile_ymin = tile_loc.y - device_grid.get_height_offset(tile_loc);
+    float tile_ymax = tile_ymin + tile_type->height;
 
     // Get the closest point in the bounding box (including the edges) to
     // the src_flat_loc. To do this, we project the point in L1 space.
@@ -656,6 +659,7 @@ static inline int get_first_available_sub_tile_at_grid_loc(const t_physical_tile
  *          be OPEN if a locations could not be found.
  */
 static inline t_pl_loc find_nearest_compatible_loc(const t_flat_pl_loc& src_flat_loc,
+                                                   float max_displacement_threshold,
                                                    t_logical_block_type_ptr block_type,
                                                    const t_pl_macro& pl_macro,
                                                    const BlkLocRegistry& blk_loc_registry) {
@@ -707,6 +711,11 @@ static inline t_pl_loc find_nearest_compatible_loc(const t_flat_pl_loc& src_flat
         //       positional distance. The un-visitied neighbors of a node should
         //       have a higher distance than the current node.
         if (grid_dist >= best_dist)
+            continue;
+
+        // If this distance is beyond the max_displacement_threshold, drop this
+        // location.
+        if (grid_dist > max_displacement_threshold)
             continue;
 
         // In order to ensure our BFS finds the closest compatible location, we
@@ -814,6 +823,7 @@ static bool try_centroid_placement(const t_pl_macro& pl_macro,
         t_flat_pl_loc centroid_flat_loc = find_centroid_loc_from_flat_placement(pl_macro, flat_placement_info);
         // Then find the nearest legal location to this centroid for this macro.
         centroid_loc = find_nearest_compatible_loc(centroid_flat_loc,
+                                                   static_cast<float>(rlim),
                                                    block_type,
                                                    pl_macro,
                                                    blk_loc_registry);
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
The output status messages look like this:
Screenshot from 2025-04-17 18-33-58

AlexandreSinger:
@amin1377 FYI

AlexandreSinger:
Titan results are in:

Metric | Improvement over baseline
-- | --
Post FL HPWL | 1.0033
Post Route HPWL | 0.9968
Atom Errors | 0.9470
Average Atom Displacement | 1.0527
Max Atom Displacement | 1.0657

This change had very little change on the titan results; however more atoms appear to be placed where they wanted to go according to the global placement. Looking through the raw data, I do notice a lot of outliers which are bringing up a lot of these numbers. I think this may be related to the mass legalizer not exactly knowing how much can be put into different clusters.

vpr/src/place/initial_placement.cpp
```diff
@@ -1447,27 +1487,283 @@ bool place_one_block(const ClusterBlockId blk_id,
...
+    vtr::vector<ClusterBlockId, float> cluster_score(cluster_netlist.blocks().size(), 0.0f);
+    for (ClusterBlockId blk_id : cluster_netlist.blocks()) {
+        // Compute the standard deviation of the positions of all atoms in the
+        // given macro. This is a measure of how much the atoms "want" to be
+        // at the centroid location.
+        t_pl_macro pl_macro = get_or_create_macro(blk_id, place_macros);
+        float variance = get_flat_variance(pl_macro, flat_placement_info);
+        float std_dev = std::sqrt(variance);
+        // Normalize the standard deviation to be a number between 0 and 1.
+        float normalized_std_dev = std_dev / (std_dev + 1.0f);
```
vaughnbetz, line 1626: This function will have relatively little difference for most std_dev values; it has a reasonable difference for small standard deviations (e.g. 1 maps to .5 and 2 to .67, but not much for larger ones (50 maps to .98 and 100 to .99).  Is that what you want?

vpr/src/place/initial_placement.cpp
```diff
@@ -1447,27 +1487,283 @@ bool place_one_block(const ClusterBlockId blk_id,
...
+            // Place the macro
+            for (const t_pl_macro_member& pl_macro_member : pl_macro.members) {
+                t_pl_loc member_pos = centroid_loc + pl_macro_member.offset;
+                ClusterBlockId iblk = pl_macro_member.blk_index;
+                blk_loc_registry.set_block_location(iblk, member_pos);
+            }
+
+            // Finally, fix the IO blocks if the user specified the option to do
+            // so.
+            fix_IO_block_types(pl_macro, centroid_loc, pad_loc_type, blk_loc_registry.mutable_block_locs());
```
vaughnbetz, line 1730: I'd beef up this comment a bit more ... if the user asked for a random pad location and this is an IO block, lock down the macro at this location so the placer can't move it.

I also think this code isn't quite right -- locking the IOs is done to test how well you work with randomly locked IOs, or IOs locked to specific locations by board level constraints and expressed with a pad_file that locks them down.  This code seems to put the IOs where it wants, and then lock them down (unless the flat placement already put them in the right spot).  If the latter, that should be commented.  If the former, there should be a TODO to fix this eventually; it is a form of placement constraint, so supporting placement constraints should eventually fix it.

vaughnbetz (approved):
Nice clean code!  One comment on fixed IOs / placement constraints embedded.

amin1377:
Hi @AlexandreSinger,

I have a quick question. In this part of the code, and elsewhere in AP, when block ordering is needed, you always sort based on the size of the macro. That makes perfect sense for the ASIC flow. However, for FPGA placement, I wonder if it might be better to sort based on the number of pins.

I realize there's likely a high correlation between macro size and pin count, but making the sorting criteria explicitly based on pins might result in a better QoR.

AlexandreSinger:
> Hi @AlexandreSinger,
> 
> I have a quick question. In this part of the code, and elsewhere in AP, when block ordering is needed, you always sort based on the size of the macro. That makes perfect sense for the ASIC flow. However, for FPGA placement, I wonder if it might be better to sort based on the number of pins.
> 
> I realize there's likely a high correlation between macro size and pin count, but making the sorting criteria explicitly based on pins might result in a better QoR.

Hi @amin1377 , I am not sure what you mean by "elsewhere in AP", this is the only place in the AP flow that I sort by the size of the macro as far as I am aware. Where else do I do this?

Regarding my use of it here; it has to do with finding a legal placement. Suppose we had a macro with 2 clusters in it. This macro will be harder to place than just a single cluster since it would need to find two open, legal clusters right next to each other (which gets harder and harder to find as the macro size increases). If we place these large macros too late, they may never find a legal site to be placed into. That was the intuition of sorting based on macro size; it was not necessarily for finding better quality solutions.

I am currently working on the mass abstraction; I think sorting based on the mass of all blocks within each macro would achieve what you are describing. We could add this to the cost function, but that would have to come in later though.

vpr/src/place/initial_placement.cpp
AlexandreSinger: This was somewhat intentional. I tried a few different normalization functions and found this one to work well. My intuition behind it is that we want variances close to 0 to be placed first; anything beyond around 5 have such a high variance that it implies that the atoms within do not know where they want to go.

The goal of this cost term is to ensure that clusters with variance 0 get placed first. Eventually I would like to add some mass information so we try to place more "massive" clusters first (clusters with the most pins / number of RAM bits).

vpr/src/place/initial_placement.cpp
AlexandreSinger: Hi Vaughn, I have updated the comment to beef it up.

The current implementation of the non-AP initial placer does this; whenever it find a legal site to place a macro it tries to lock them down using this method (the method checks if the option was passed in). This matches what the fix-pins option expects:
Screenshot from 2025-04-21 13-34-34

The IO pads are fixed to "arbitrary" locations (when the IOs are not fixed, AP GP will place them basically anywhere it wants). However, I see what you mean since this is not truly "random" since AP is choosing good sites for these blocks. I added a TODO to investigate this.

PR #2989, opened 2025-04-19
Remove PlacerMoveContext

soheilshahrouz:
`PlacerMoveContext` class contained bounding-box information for clustered nets, some scratchpad vectors, and `first_rlim`. This PR moves all bounding-box related info to `NetCostHandler`. Scratchpad vectors are moved to classes that use them, and `first_rlim` is now a static data member of `MoveGenerator`.


vpr/src/place/move_generators/move_generator.h
```diff
@@ -157,9 +159,13 @@ class MoveGenerator {
                                               double delta_c,
                                               float timing_bb_factor);
 
+  public:
+    static float first_rlim;
```
amin1377, line 163: I'd suggest adding some comments above it to explain the purpose of this variable and why it is defined as a static variable.

vpr/src/place/move_generators/feasible_region_move_generator.h
```diff
@@ -34,6 +34,8 @@ class FeasibleRegionMoveGenerator : public MoveGenerator {
                                float rlim,
                                const t_placer_opts& placer_opts,
                                const PlacerCriticalities* criticalities) override;
-};
 
-#endif
+  private:
+    std::vector<int> X_coord;
```
amin1377, line 39: I'd suggest adding a brief comment to clarify that these variables are used as a scratchpad to store the locations of relevant blocks.

amin1377: Also, one more question: before your changes, didn’t we used to reserve the capacity of all blocks in these scratchpad variables?

vpr/src/place/move_generators/median_move_generator.h
```diff
@@ -64,6 +64,9 @@ class MedianMoveGenerator : public MoveGenerator {
                                              t_bb& bb_coord_new,
                                              ClusterBlockId moving_block_id,
                                              bool& skip_net);
-};
 
-#endif
+  private:
+    std::vector<int> X_coord;
```
amin1377, line 69: Instead of redefining these variables in all child classes, can we define them as protected members in the parent class? Or would that cause any issues?

vpr/src/place/net_cost_handler.h
```diff
@@ -169,6 +169,21 @@ class NetCostHandler {
     /* [0...num_affected_nets] -> net_id of the affected nets */
     std::vector<ClusterNetId> ts_nets_to_update_;
 
+    // [0..cluster_ctx.clb_nlist.nets().size()-1]. Store the number of blocks on each of a net's bounding box (to allow efficient updates)
```
amin1377, line 172: I think it would be useful to add a top-level comment here explaining that the following variables store the bounding box coordinates, etc., after a move is committed (to clarify how they differ from the ts_ variables).

amin1377 (changes requested):
Thanks, Soheil! Overall, the PR looks good. I just left a few minor comments. Also, since you're modifying some performance-critical parts of the code, I'd suggest running the Titan benchmark on the SAVI machine and comparing its runtime against the master branch to ensure there are no significant regressions.

vpr/src/place/move_generators/median_move_generator.h
soheilshahrouz: Only 3 move generators use `X_coord` and `Y_coord`, while `layer_coord` is only used by two of them.

vpr/src/place/move_generators/feasible_region_move_generator.h
soheilshahrouz: No, we don't reserve memory space for these vectors.

soheilshahrouz:
Results for VTR benchmarks

| Metric                                              | master | feature       |
|-----------------------------------------------------|--------|---------------|
| vtr_flow_elapsed_time                               | 1      | 0.9956233242  |
| parmys_synth_time                                   | 1      | 0.9984187005  |
| abc_depth                                           | 1      | 1             |
| abc_synth_time                                      | 1      | 0.9992354382  |
| num_clb                                             | 1      | 1             |
| num_memories                                        | 1      | 1             |
| num_mult                                            | 1      | 1             |
| max_vpr_mem                                         | 1      | 1.001330976   |
| num_pre_packed_blocks                               | 1      | 1             |
| num_post_packed_blocks                              | 1      | 1             |
| device_grid_tiles                                   | 1      | 1             |
| pack_time                                           | 1      | 0.9930076     |
| placed_wirelength_est                               | 1      | 1             |
| place_time                                          | 1      | 1.010553331   |
| placed_CPD_est                                      | 1      | 1             |
| min_chan_width                                      | 1      | 1             |
| routed_wirelength                                   | 1      | 1             |
| min_chan_width_route_time                           | 1      | 0.9870603223  |
| crit_path_routed_wirelength                         | 1      | 1             |
| critical_path_delay                                 | 1      | 1             |
| geomean_nonvirtual_intradomain_critical_path_delay  | 1      | 1             |
| crit_path_route_time                                | 1      | 1.008222132   |

PR #2991, opened 2025-04-21
[Packer] Pack Pattern Get Sink

amin1377:
WIP

### DIFF
--- a/vpr/src/pack/prepack.cpp
+++ b/vpr/src/pack/prepack.cpp
@@ -107,8 +107,7 @@ static void init_molecule_chain_info(const AtomBlockId blk_id,
                                      const AtomNetlist& atom_nlist);
 
 static AtomBlockId get_sink_block(const AtomBlockId block_id,
-                                  const t_model_ports* model_port,
-                                  const BitIndex pin_number,
+                                  const t_pack_pattern_connections& connections,
                                   const AtomNetlist& atom_nlist);
 
 static AtomBlockId get_driving_block(const AtomBlockId block_id,
@@ -1047,9 +1046,7 @@ static bool try_expand_molecule(t_pack_molecule& molecule,
             // this block is the driver of this connection
             if (block_connection->from_block == pattern_block) {
                 // find the block this connection is driving and add it to the queue
-                auto port_model = block_connection->from_pin->port->model_port;
-                auto ipin = block_connection->from_pin->pin_number;
-                auto sink_blk_id = get_sink_block(block_id, port_model, ipin, atom_nlist);
+                auto sink_blk_id = get_sink_block(block_id, *block_connection, atom_nlist);
                 // add this sink block id with its corresponding pattern block to the queue
                 pattern_block_queue.push(std::make_pair(block_connection->to_block, sink_blk_id));
                 // this block is being driven by this connection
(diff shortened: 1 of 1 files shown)


vpr/src/pack/prepack.cpp
```diff
@@ -1076,23 +1071,50 @@ static bool try_expand_molecule(t_pack_molecule& molecule,
 /**
  * Find the atom block in the netlist driven by this pin of the input atom block
  * If doesn't exist return AtomBlockId::INVALID()
- * Limitation: The block should be driving only one sink block
```
AlexandreSinger, line 1079: Isn't this limitation still there? Based on your code changes, it appears as though you have just handled the special case where this is a chain pattern.

Is it possible for these changes to be generalized? If its complicated I would leave as a TODO at least so people are aware of this limitation.

vpr/src/pack/prepack.cpp
```diff
@@ -1076,23 +1071,50 @@ static bool try_expand_molecule(t_pack_molecule& molecule,
...
  *      block_id   : id of the atom block that is driving the net connected to the sink block
- *      model_port : the model of the port driving the net
- *      pin_number : the pin_number of the pin driving the net (pin index within the port)
+ *      connections : pack pattern connections from the given block
```
AlexandreSinger, line 1075: Missing atom_nlist and is_chain_pattern.

I am not worried about documenting atom_nlist; but I am more worried about is_chain_pattern.

vpr/src/pack/prepack.cpp
```diff
@@ -1076,23 +1071,50 @@ static bool try_expand_molecule(t_pack_molecule& molecule,
...
+                                  bool is_chain_pattern) {
+    const t_model_ports* from_port_model = connections.from_pin->port->model_port;
+    const int from_pin_number = connections.from_pin->pin_number;
+    auto from_port_id = atom_nlist.find_atom_port(block_id, from_port_model);
+
+    const t_model_ports* to_port_model = connections.to_pin->port->model_port;
+    const int to_pin_number = connections.to_pin->pin_number;
+    const auto& to_pb_type = connections.to_block->pb_type;
+
+    if (from_port_id) {
```
AlexandreSinger, line 1089: This can be made more readible and cleaner if you bail from the function early if this is false.

For example:
```cpp
// If a from port id could not be found, ...
if (!from_port_id)
  return AtomBlockId::INVALID();
```

This reduces the indentation level and makes it clear what happens in this case.

vpr/src/pack/prepack.cpp
```diff
@@ -1076,23 +1071,50 @@ static bool try_expand_molecule(t_pack_molecule& molecule,
...
+        auto net_id = atom_nlist.port_net(from_port_id, from_pin_number);
+        if (net_id.is_valid()) {
```
AlexandreSinger, line 1091: Same comment as `from_port_id`. This can be made cleaner by doing:

```cpp
// Comment explaining why we return invalid in this case
if (!net_id.is_valid())
  return AtomBlockId::INVALID();
```

vpr/src/pack/prepack.cpp
```diff
@@ -1076,23 +1071,50 @@ static bool try_expand_molecule(t_pack_molecule& molecule,
...
+            const auto& net_sinks = atom_nlist.net_sinks(net_id);
+            if (is_chain_pattern) {
+                // If the pattern is a chain, allow nets with multiple sinks.
+                // This enables forming chains where the COUT is connected both to
+                // the next element in the chain and to the block's output pin.
```
AlexandreSinger, line 1096: Is this always valid? Is there ever a case where we could get a packing failure as you detail below?

Is it somehow possible to check for this?

vpr/src/pack/prepack.cpp
```diff
@@ -1076,23 +1071,50 @@ static bool try_expand_molecule(t_pack_molecule& molecule,
...
+                            return sink_block_id;
+                        }
+                    }
+                }
+            } else {
+                // For non-chain patterns, we conservatively only consider the sink block
+                // if the net fanout is 1. To clarify, consider a case where the output of a LUT
+                // is connected to both a register and an unregistered output that feeds another block.
+                // If the intra-cluster architecture doesn't support having both registered and
+                // unregistered outputs simultaneously, this could lead to a packing failure.
```
AlexandreSinger, line 1112: Same comment as above. Is it possible to check the architecture and see if we can ignore this assumption. I am wondering if this can be made more general. If it is complicated, perhaps we can raise this as an issue to investigate further later.

Prepacking is super powerful, and the more intelligent we can make it the better in my opinion.

vpr/src/pack/prepack.cpp
```diff
@@ -1104,33 +1126,32 @@ static AtomBlockId get_sink_block(const AtomBlockId block_id,
...
+                                     const t_pack_pattern_connections& connections,
                                      const AtomNetlist& atom_nlist) {
-    auto port_id = atom_nlist.find_atom_port(block_id, model_port);
+    auto to_port_model = connections.to_pin->port->model_port;
+    auto to_pin_number = connections.to_pin->pin_number;
+    auto to_port_id = atom_nlist.find_atom_port(block_id, to_port_model);
 
-    if (port_id) {
-        auto net_id = atom_nlist.port_net(port_id, pin_number);
+    if (to_port_id) {
```
AlexandreSinger, line 1138: I recognize that this is not your code here, but I think the code can be made more readible if we reduce the nesting level of these if statements. A lot of the code in this file can be made more readible; so it would be nice to chip away at it!

AlexandreSinger (commented):
Hi @amin1377 , great changes! I left some minor comments, but other than that it looks good to me.

vpr/src/pack/prepack.cpp
amin1377: Here, I’m making the assumption that if a pin in the pack pattern has a fan-out greater than one, the architecture explicitly allows it. If you have a case where, for example, two adders are connected together and the carry connection between them is also connected to another block, but the architecture does not support this behavior, that would result in a problem.

However, without this update, even if the architecture does support such a mechanism, the design would still fail to implement correctly due to the current restriction.

vpr/src/pack/prepack.cpp
amin1377: Good point! Analyzing the architecture file to determine whether to consider multi-fanout nets in pack patterns is a good idea. I’ve created Issue #2996 to track this.

vpr/src/pack/prepack.cpp
amin1377: It’s always a good idea to leave the code a little better than you found it :) I’ll clean up the nesting here.

amin1377:
@AlexandreSinger The PR has now passed Nightly Test 1 (tested on Wintermute). Feel free to merge it if you don’t have any further comments.

amin1377:
Nightly tests results: https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14625001418

AlexandreSinger (approved):
LGTM

vaughnb-cerebras:
@amin1377 @AlexandreSinger : I'm late commenting on this, but I agree with both of you that analyzing the architecture file to figure out if we can make a molecule without triggering an unroute is the right long term change. The comment about LUT + FF and one output only is really documenting a very specific (but real) case; there are lots of different cases one could come up with that should or should not be turned into a molecule based on routability.

Issue #2859, opened 2025-01-14
[Pack] List of Feasible Candidates Should be a Priority Queue


AlexandreSinger:
In the GreedyCandidateSelector class, used within the packer, a list of feasible blocks is maintained which is used to select the next candidate to propose:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/e46e9c22feeb5ec7921bb702206d5bd5c1ed8834/vpr/src/pack/greedy_candidate_selector.h#L86-L91

Currently this is stored as a vector where it is ordered based on the gain of the block. To maintain the order, an insertion sort is used every time a block is added to the list. Not only is this code confusing to read, it may also be slow. This should instead be an `std::priority_queue`.

RonZ13:
I ll take this.

PR #2994, opened 2025-04-22
[packer] Changing List of Feasible Candidates to Priority Queue

RonZ13:
#### Description

[packer] Changing the vector of candidate molecules into LazyPopUniquePriorityQueue.

    The class LazyPopUniquePriorityQueue is a priority queue that allows for lazy deletion of elements.
    The elements are pair of key and sort-value. The key is a unique value to identify the item, and the sort-value is used to sort the item.
    It is implemented using a vector and 2 sets, one set keeps track of the elements in the queue, and the other set keeps track of the elements that are pending deletion, so that they can be removed from the queue when they are popped.
    The class definiation can be found in vpr/src/util/lazy_pop_unique_priority_queue.h

    Currently, the class supports the following functions:
        LazyPopUniquePriorityQueue::push(): Pushes a key-sort-value (K-SV) pair into the priority queue and adds the key to the tracking set.
        LazyPopUniquePriorityQueue::pop(): Returns the K-SV pair with the highest SV whose key is not pending deletion.
        LazyPopUniquePriorityQueue::remove(): Removes an element from the priority queue immediately.
        LazyPopUniquePriorityQueue::remove_at_pop_time(): Removes an element from the priority queue when it is popped.
        LazyPopUniquePriorityQueue::empty(): Returns whether the queue is empty.
        LazyPopUniquePriorityQueue::clear(): Clears the priority queue vector and the tracking sets.
        LazyPopUniquePriorityQueue::size(): Returns the number of elements in the queue.
        LazyPopUniquePriorityQueue::contains(): Returns true if the key is in the queue, false otherwise.

#### Related Issue

This PR tries to resolve this issue: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2859

#### Motivation and Context

The original pack uses a fixed size vector to store the proposed candiates, and the candiates are sorted using insertion sort. If we want to increase the size, this data structure be less efficient than a priority queue.

It has been tested using Titan_quick_qor tests. It shows slight worse CPD, 1% increase, and better WL, 1% decrease. 3% reduction in pack time, and 11% reduction in placement time.
image

#### Types of changes

#### Checklist:


vpr/src/util/lazy_pop_unique_priority_queue.h
```diff
@@ -0,0 +1,174 @@
...
+            } 
+        }
+
+        /**
+         * @brief Check if the priority queue is empty.
+         *
+         *  @return True if the priority queue is empty, false otherwise.
+         */
+        bool empty(){
+            return heap.empty();
```
AlexandreSinger, line 139: This is not technically correct. What if the heap is filled with elements which were "removed_at_pop_time"? If the heap is not empty, I think it will have to check if all elements in the list have been marked for deletion.

This relates to my comment in the candidate selector. It is not sufficient to check if the size of the heap is zero. You will have to check if all elements in the heap have not been marked for deletion.

vpr/src/util/lazy_pop_unique_priority_queue.h
```diff
@@ -0,0 +1,174 @@
...
+            delete_pending_set.clear();
+        }
+
+        /**
+         * @brief Get the size of the priority queue.
+         *
+         *  @return The size of the priority queue.
+         */
+        size_t size(){
+            return heap.size();
```
AlexandreSinger, line 159: This size includes the elements which have been marked for deletion. Is this intentional? You should add to the comment above if so.

vpr/src/util/lazy_pop_unique_priority_queue.h
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        /**
+         * @brief Check if the item refered by the key is in the priority queue.
```
AlexandreSinger, line 163: refered -> referred to

AlexandreSinger (changes requested):
Overall a very welcome change, thank you!

I left some comments. I am a little concerned about how you tell if the heap is empty or not. Just getting the size of the heap may not include any elements which have been marked for deletion.

AlexandreSinger:
@RonZ13 These changes look good to me. We need to run the Nightly Tests on this branch before merging. I cannot run this PR on the NightlyTestManual runner from github due to this branch being on your private fork (we can only run it on branches on Master). I have a feeling that the Nightly Tests will also have some failures...

RonZ13:
> @RonZ13 These changes look good to me. We need to run the Nightly Tests on this branch before merging. I cannot run this PR on the NightlyTestManual runner from github due to this branch being on your private fork (we can only run it on branches on Master). I have a feeling that the Nightly Tests will also have some failures...

I am a bit confused. Do I run the Nightly test locally? or do I merge this packer-feasible-candidate-queue branch into master branch in my forked repo, then the CI will automatically run on this PR?

AlexandreSinger:
> > @RonZ13 These changes look good to me. We need to run the Nightly Tests on this branch before merging. I cannot run this PR on the NightlyTestManual runner from github due to this branch being on your private fork (we can only run it on branches on Master). I have a feeling that the Nightly Tests will also have some failures...
> 
> @AlexandreSinger I am a bit confused. Do I run the Nightly test locally? or do I merge this packer-feasible-candidate-queue branch into master branch in my forked repo, then the CI will automatically run on this PR?

The CI is run weekly on Sundays automatically only on Sundays. I do not recommend merging your code in and then waiting for that event to happen. You can run the Nightly Tests locally on Wintermute (you will need many cores to run any of the Nightly Tests in a reasonable amount of time).

Another option is to use the interface that @AmirhosseinPoolad made: https://docs.verilogtorouting.org/en/latest/README.developers/#manual-nightly-tests
However, this approach requires the branch you want to run to be in the VTR repository (not on a fork). You can raise another branch on origin with the same code and run it using the instructions above. Regardless we need the NightlyTests ran (and most likely the golden results generated) before merging this PR. If we do not do this, the NightlyTests will fail on Master which is very nasty to fix after the PR has been merged.

RonZ13:
@AlexandreSinger Can you please do the final round of review? It passed the Nightly test.

AlexandreSinger (approved):
LGTM, will merge once CI passes.

PR #2998, opened 2025-04-23
Invert use of macro_can_be_placed argument in check_all_legality 

gigeresk:
The use of the check_all_legality function parameter was the inverse of what was documented.  

#### Description
The check_all_legality parameter performs additional legality checks regarding macro placement.
Invert the use of the value and the value itself in most cases. Some uses are unchanged because they were initially incorrect.

#### Related Issue
None

#### Motivation and Context
Improves code clarity and reverts to expected behaviour.

TODO: run qor and test suite

#### Types of changes

#### Checklist:


vpr/src/place/initial_placement.cpp
```diff
@@ -1210,10 +1208,9 @@ bool try_place_macro(const t_pl_macro& pl_macro,
         return macro_placed;
     }
 
-    bool mac_can_be_placed = macro_can_be_placed(pl_macro, head_pos, /*check_all_legality=*/false, blk_loc_registry);
-
-    if (mac_can_be_placed) {
-        // Place down the macro
+    // called from initial placement
```
AlexandreSinger, line 1211: What does this comment mean? Was this a temporary comment? Perhaps it needs to be expanded more.

vpr/src/place/initial_placement.cpp
```diff
@@ -1210,10 +1208,9 @@ bool try_place_macro(const t_pl_macro& pl_macro,
...
+    if (macro_can_be_placed(pl_macro, head_pos, /*check_all_legality=*/true, blk_loc_registry)) {
+        // Place down the macromacro_can_be_placed
```
AlexandreSinger, line 1213: Was this a typo? The added macro_can_be_placed seems out of place.

vpr/src/place/place_util.cpp
```diff
@@ -191,7 +191,7 @@ bool macro_can_be_placed(const t_pl_macro& pl_macro,
          * floorplan constraint is not supported by analytical placement yet, 
          * hence, if macro_can_be_placed is called from analytical placer, no further actions are required. 
          */
-        if (check_all_legality) {
+        if (not check_all_legality) {
```
AlexandreSinger, line 194: nitpick: People in VTR tend to avoid alternative operators like "not", preferring to instead just explicitly write the not operator "!". I recommend using the not operator instead.

AlexandreSinger (commented):
Left some minor comments. Overall this change looks good to me! Thank you so much for fixing this!

PR #3000, opened 2025-04-24
[STA] Added Option to Remove Parameters from Post-Implementation Netlist

AlexandreSinger:
When performing post-implementation timing analysis using OpenSTA, the generated netlist cannot use parameters since each module needs to correspond with a cell in a liberty file.

Added a command-line option which tells the netlist writer to not use parameters when generating the netlist. If a primitive cannot be generated without using parameters, it will error out.

### DIFF
--- a/doc/src/vpr/command_line_usage.rst
+++ b/doc/src/vpr/command_line_usage.rst
@@ -1730,6 +1730,16 @@ Analysis Options
 
     **Default:** ``unconnected``
 
+.. option:: --post_synth_netlist_module_parameters { on | off }
+
+    Controls whether the post-synthesis netlist output by VTR can use Verilog parameters
+    or not. When using the post-synthesis netlist for external timing analysis,
+    some tools cannot accept the netlist if it contains parameters. By setting
+    this option to ``off``, VPR will try to represent the netlist using non-parameterized
+    modules.
+
+    **Default:** ``on``
+
 .. option:: --timing_report_npaths <int>
 
     Controls how many timing paths are reported.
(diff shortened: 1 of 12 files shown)


ganeshgore (approved):

doc/src/vpr/command_line_usage.rst
```diff
@@ -1730,6 +1730,16 @@ Analysis Options
 
     **Default:** ``unconnected``
 
+.. option:: --post_synth_netlist_module_parameters { on | off }
+
+    Controls whether the post-synthesis netlist modules should use Verilog parameters
```
vaughnbetz, line 1735: Maybe make clear this is output, e.g. "... any post-synthesis netlist output by VTR can use Verilog parameters in its modules."

vpr/src/base/read_options.cpp
```diff
@@ -2997,6 +2997,16 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("unconnected")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    analysis_grp.add_argument<bool, ParseOnOff>(args.post_synth_netlist_module_parameters, "--post_synth_netlist_module_parameters")
+        .help(
+            "Controls whether the post-synthesis netlist modules should use Verilog parameters "
```
vaughnbetz, line 3002: See suggested change to the .rst

vaughnbetz (approved):
LGTM; just one documentation tweak.

PR #3001, opened 2025-04-24
enum class rr_type

soheilshahrouz:
Opened to trigger CI.


libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -592,7 +592,7 @@ static bool has_adjacent_channel(const RRGraphView& rr_graph, const DeviceGrid&
     /* TODO: this function should be reworked later to adapt RRGraphView interface 
      *       once xlow(), ylow(), side() APIs are implemented
      */
-    VTR_ASSERT(rr_graph.node_type(node.id()) == IPIN || rr_graph.node_type(node.id()) == OPIN);
+    VTR_ASSERT(rr_graph.node_type(node.id()) == t_rr_type::IPIN || rr_graph.node_type(node.id()) == e_rr_type::OPIN);
```
amin1377, line 595: Shouldn't both of them be t_rr_type?

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -781,9 +781,9 @@ class t_rr_graph_storage {
         const RRNodeId& id,
         const e_side& side) {
         auto& node_data = node_storage[id];
-        if (node_data.type_ != IPIN && node_data.type_ != OPIN) {
+        if (node_data.type_ != e_rr_type::IPIN && node_data.type_ != e_rr_type::OPIN) {
             VTR_LOG_ERROR("Attempted to access RR node 'side' for non-IPIN/OPIN type '%s'",
-                          rr_node_typename[node_data.type_]);
+                          rr_node_typename[(size_t)node_data.type_]);
```
amin1377, line 786: Can't we change it to a vtr::vector so that it only allows the node data type as an index? Feel free to ignore this suggestion if it's not worth the effort.

libs/librrgraph/src/base/rr_spatial_lookup.cpp
```diff
@@ -55,27 +55,27 @@ RRNodeId RRSpatialLookup::find_node(int layer,
...
         return RRNodeId::INVALID();
     }
 
-    if (size_t(ptc) >= rr_node_indices_[type][layer][node_x][node_y][node_side].size()) {
+    if (size_t(ptc) >= rr_node_indices_[(size_t)type][layer][node_x][node_y][node_side].size()) {
         return RRNodeId::INVALID();
     }
 
-    return rr_node_indices_[type][layer][node_x][node_y][node_side][ptc];
+    return rr_node_indices_[(size_t)type][layer][node_x][node_y][node_side][ptc];
```
amin1377, line 78: Same comment as above. Can we use vtr vector/matrix to be able use node type directly to index it?

vpr/src/place/place_macro.cpp
```diff
@@ -211,7 +211,7 @@ int PlaceMacros::find_all_the_macro_(const ClusteredNetlist& clb_nlist,
             // Note that the restriction that constant nets are not driven from another direct ensures that
             // blocks in the middle of a chain with internal constant signals are not detected as potential
             // head blocks.
-            if (to_src_or_sink == SINK && to_idirect != OPEN
+            if (to_src_or_sink == (int)e_rr_type::SINK && to_idirect != OPEN
```
amin1377, line 214: Can we change the return type of direct_type_from_blk_pin_ to e_rr_type?

vpr/src/route/route.cpp
```diff
@@ -618,7 +618,7 @@ bool route(const Netlist<>& net_list,
             "total_internal_heap_pushes: %zu total_internal_heap_pops: %zu total_external_heap_pushes: %zu total_external_heap_pops: %zu ",
             router_stats.intra_cluster_node_pushes, router_stats.intra_cluster_node_pops,
             router_stats.inter_cluster_node_pushes, router_stats.inter_cluster_node_pops);
-        for (int node_type_idx = 0; node_type_idx < t_rr_type::NUM_RR_TYPES; node_type_idx++) {
+        for (int node_type_idx = 0; node_type_idx < (int)t_rr_type::NUM_RR_TYPES; node_type_idx++) {
```
amin1377, line 621: If we change all t_rr_type to e_rr_type, wouldn't lines like this cause issues?

amin1377 (changes requested):
Looks good, Soheil! Thanks!

I talked to Soehil and he mentioned that he is going to remove t_rr_type completely and replace it with e_rr_type.

vpr/src/place/place_macro.cpp
soheilshahrouz: `direct_type_from_blk_pin_` stores `OPEN` for for pins that could not be part of a direct chain connection. OPEN is not part of `e_rr_type`.

vpr/src/route/route.cpp
soheilshahrouz: How?

libs/librrgraph/src/base/rr_graph_storage.h
soheilshahrouz: I'll add a `vtr::array` class to address this.

amin1377 (approved):
Thanks, Soheil! The PR looks good.

AmirhosseinPoolad:
Missing documentation for the new vtr::array container in the API reference.
https://docs.verilogtorouting.org/en/latest/api/vtrutil/containers/

AmirhosseinPoolad (changes requested):
Missing documentation for the new vtr::array container in the API reference.
https://docs.verilogtorouting.org/en/latest/api/vtrutil/containers/

AmirhosseinPoolad (approved):

PR #3003, opened 2025-04-26
3D Strong Test

amin1377:
Remove the redundant 3D test and only keep 2 tests to cover per-layer and cube bounding boxes.


AlexandreSinger (approved):
LGTM. Will merge when CI passes.

PR #3005, opened 2025-04-28
[AP][HotFix] Fixed Bug With Solver Putting Blocks Off-Device

AlexandreSinger:
After moving fixed blocks to the center of tiles, there is a very small chance that blocks go off the device due to rounding. This is such a small effect that it does not show up locally on my machine, but it shows up on CI. Clamping the positions of blocks after solving to be just within the device region.

This should fix the failure in the nightly tests.


amin1377 (approved):
Thanks for the quick fix, Alex!

PR #3007, opened 2025-04-28
[Pack] Pack Pattern Root IPIN

amin1377:
WIP

### DIFF
--- a/vpr/src/pack/prepack.cpp
+++ b/vpr/src/pack/prepack.cpp
@@ -114,6 +114,8 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
                                      const t_pack_pattern_connections& connections,
                                      const AtomNetlist& atom_nlist);
 
+static t_pb_graph_pin* get_compatible_chain_root_pin(const t_pack_patterns* chain_pattern, const AtomBlockId blk_id);
+
 static void print_chain_starting_points(t_pack_patterns* chain_pattern);
 
 /*****************************************/
@@ -1163,6 +1165,17 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
     return AtomBlockId::INVALID();
 }
 
+static t_pb_graph_pin* get_compatible_chain_root_pin(const t_pack_patterns* chain_pattern, const AtomBlockId blk_id) {
+    for (const auto& chain : chain_pattern->chain_root_pins) {
+        for (const auto& tie_off : chain) {
+            if (primitive_type_feasible(blk_id, tie_off->parent_node->pb_type)) {
+                return tie_off;
+            }
+        }
+    }
+    return nullptr;
+}
+
 static void print_pack_molecules(const char* fname,
                                  const std::vector<t_pack_patterns>& list_of_pack_patterns,
                                  const int num_pack_patterns,
@@ -1324,10 +1337,9 @@ static AtomBlockId find_new_root_atom_for_chain(const AtomBlockId blk_id,
 
     VTR_ASSERT(list_of_pack_patterns->is_chain == true);
     VTR_ASSERT(list_of_pack_patterns->chain_root_pins.size());
-    root_ipin = list_of_pack_patterns->chain_root_pins[0][0];
-    root_pb_graph_node = root_ipin->parent_node;
+    root_ipin = get_compatible_chain_root_pin(list_of_pack_patterns, blk_id);
 
-    if (primitive_type_feasible(blk_id, root_pb_graph_node->pb_type) == false) {
+    if (root_ipin == nullptr) {
         return AtomBlockId::INVALID();
     }
 
@@ -1622,7 +1634,7 @@ static void init_molecule_chain_info(const AtomBlockId blk_id,
     // pattern assigned to it and the input block should be valid
     VTR_ASSERT(molecule.pack_pattern && blk_id);
 
-    auto root_ipin = molecule.pack_pattern->chain_root_pins[0][0];
+    auto root_ipin = get_compatible_chain_root_pin(molecule.pack_pattern, blk_id);
     auto model_pin = root_ipin->port->model_port;
     auto pin_bit = root_ipin->pin_number;
 


vpr/src/pack/prepack.cpp
```diff
@@ -1601,9 +1655,11 @@ static void get_all_connected_primitive_pins(const t_pb_graph_pin* cluster_input
         const auto& output_edge = cluster_input_pin->output_edges[iedge];
         for (int ipin = 0; ipin < output_edge->num_output_pins; ipin++) {
             if (output_edge->output_pins[ipin]->is_primitive_pin()) {
-                connected_primitive_pins.push_back(output_edge->output_pins[ipin]);
+                if (pattern_blocks.find(output_edge->output_pins[ipin]->parent_node->pb_type) != pattern_blocks.end()) {
```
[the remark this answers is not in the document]
AmirhosseinPoolad: It's also not obvious at a glance what it would mean for pattern_blocks to contain the pb_type of output_pins[ipin]. You could add a comment here to help with that

vpr/src/pack/prepack.cpp
```diff
@@ -1172,6 +1182,47 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
     return AtomBlockId::INVALID();
 }
 
+static std::unordered_set<t_pb_type*> get_pattern_blocks(const t_pack_patterns* pack_pattern) {
+    std::unordered_set<t_pb_type*> pattern_blocks;
+
+    auto connections = pack_pattern->root_block->connections;
```
AmirhosseinPoolad, line 1188: Change auto to actual type name

vpr/src/pack/prepack.cpp
```diff
@@ -1172,6 +1182,47 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
...
+    if (connections == nullptr) {
+        return pattern_blocks;
+    }
+    std::unordered_set<t_pb_graph_pin*> visited_from_pins;
+    std::unordered_set<t_pb_graph_pin*> visited_to_pins;
+    std::queue<t_pack_pattern_block*> pack_pattern_blocks;
+    pack_pattern_blocks.push(connections->from_block);
+    /** Start from the root block of the pack pattern and add the connected block to the queue */
+    while (!pack_pattern_blocks.empty()) {
+        auto current_pattern_block = pack_pattern_blocks.front();
```
AmirhosseinPoolad, line 1198: Same as above

vpr/src/pack/prepack.cpp
```diff
@@ -1172,6 +1182,47 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
...
+        pack_pattern_blocks.pop();
+        auto current_connenction = current_pattern_block->connections;
```
AmirhosseinPoolad, line 1200: Auto

vpr/src/pack/prepack.cpp
```diff
@@ -1172,6 +1182,47 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
...
+        /** Iterate through all the connections of the current pattern block to
+         * add the connected block to the queue
+         */
+        while (current_connenction != nullptr) {
+            if (visited_from_pins.count(current_connenction->from_pin)) {
+                if (visited_to_pins.count(current_connenction->to_pin)) {
```
AmirhosseinPoolad, line 1206: These two can use std::unordered_set::contains instead of count.

vpr/src/pack/prepack.cpp
```diff
@@ -117,6 +119,14 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
                                      const t_pack_pattern_connections& connections,
                                      const AtomNetlist& atom_nlist);
 
+/**
+ * @brief Get an unordered set of all pb_types in the given pack pattern
+ * 
+ * @param pack_pattern Pack pattern to get pb_types from
+ * @return std::unordered_set<t_pb_type*> Set of pb_types in the pack pattern
+ */
+static std::unordered_set<t_pb_type*> get_pattern_blocks(const t_pack_patterns* pack_pattern);
```
AmirhosseinPoolad, line 128: Nitpick: why not use a reference to pass pack_pattern? This part of the code unfortunately uses a ton of C-style arrays so using the reference syntax makes it super clear that it's not an array.

AmirhosseinPoolad (changes requested):
Thanks for the changes Amin! I'll let Vaughn comment on the actual functionality changes, but meanwhile I had a couple of style related comments and nitpicks.

vpr/src/pack/prepack.cpp
```diff
@@ -1590,7 +1642,9 @@ static void update_chain_root_pins(t_pack_patterns* chain_pattern,
  *  the Cin pin of all the adder primitives connected to this pin. Which is for typical architectures
  *  will be only one pin connected to the very first adder in the cluster.
  */
-static void get_all_connected_primitive_pins(const t_pb_graph_pin* cluster_input_pin, std::vector<t_pb_graph_pin*>& connected_primitive_pins) {
+static void get_all_connected_primitive_pins(const t_pb_graph_pin* cluster_input_pin,
+                                             const std::unordered_set<t_pb_type*>& pattern_blocks,
+                                             std::vector<t_pb_graph_pin*>& connected_primitive_pins) {
```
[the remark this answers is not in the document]
amin1377: Added the following comment:

```
/**
 * @brief Get all primitive pins connected to the given cluster input pin
 * 
 * @param cluster_input_pin Cluster input pin to get connected primitive pins from
 * @param pattern_blocks Set of pb_types in the pack pattern. Pins on the blocks in this set will 
 * be added to the connected_primitive_pins vector
 * @param connected_primitive_pins Vector to store connected primitive pins
 */
```

vpr/src/pack/prepack.cpp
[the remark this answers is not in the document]
amin1377: I actually considered using contains instead of find, but since I want this code to be compatible with OpenFPGA — which isn’t compiled with C++20 — I decided against it. Also, I think std::unordered_set::contains is just more expressive than find and in terms of performance, it is not different from std::find.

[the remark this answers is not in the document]
amin1377: added the followign comment:

```
/** Add the output pin to the vector only if it belongs to a pb_type registered in the pattern_blocks set */
```

vpr/src/pack/prepack.cpp
amin1377: You’re really not a fan of auto :)

vpr/src/pack/prepack.cpp
amin1377: Same as I mentioned above, but let me know if that’s not convincing, and I’m happy to switch to contains instead of find.

vpr/src/pack/prepack.cpp
amin1377: You’re right, it makes more sense to pass it as a const reference. I was trying to stay consistent with other parts of the code, but I agree that it’s better to follow modern C++ style for new code. We can always refactor the older parts later for consistency.

amin1377:
@AmirhosseinPoolad Thanks for reviewing the code! I’ve addressed your comments. Please resolve the conversations if you think they’ve been properly handled. Let me know if you have any further suggestions.

vpr/src/pack/prepack.cpp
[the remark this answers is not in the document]
AmirhosseinPoolad: Makes sense. I guess the only thing remaining is that you used find here and count in the other place to do this thing. I think it's better to just one method for consistency. Don't have a particular preference, use whichever method you like more.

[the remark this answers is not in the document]
amin1377: Good point. I changed the counts to find.

AmirhosseinPoolad (approved):

vpr/src/pack/prepack.cpp
```diff
@@ -1172,6 +1190,47 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
...
+    std::queue<t_pack_pattern_block*> pack_pattern_blocks;
+    pack_pattern_blocks.push(connections->from_block);
+    /** Start from the root block of the pack pattern and add the connected block to the queue */
+    while (!pack_pattern_blocks.empty()) {
+        t_pack_pattern_block* current_pattern_block = pack_pattern_blocks.front();
+        pack_pattern_blocks.pop();
+        t_pack_pattern_connections* current_connenction = current_pattern_block->connections;
+        /** Iterate through all the connections of the current pattern block to
+         * add the connected block to the queue
+         */
```
soheilshahrouz, line 1211: Comment blocks starting with `/**` are parsed as doxygen comments by some IDEs.

vpr/src/pack/prepack.cpp
```diff
@@ -1172,6 +1190,47 @@ static AtomBlockId get_driving_block(const AtomBlockId block_id,
...
+    std::unordered_set<t_pb_graph_pin*> visited_from_pins;
+    std::unordered_set<t_pb_graph_pin*> visited_to_pins;
+    std::queue<t_pack_pattern_block*> pack_pattern_blocks;
+    pack_pattern_blocks.push(connections->from_block);
+    /** Start from the root block of the pack pattern and add the connected block to the queue */
+    while (!pack_pattern_blocks.empty()) {
+        t_pack_pattern_block* current_pattern_block = pack_pattern_blocks.front();
+        pack_pattern_blocks.pop();
+        t_pack_pattern_connections* current_connenction = current_pattern_block->connections;
+        /** Iterate through all the connections of the current pattern block to
```
soheilshahrouz, line 1209: Adding some empty lines can make the code more readable, especially when added between different logical steps that are taken in the code.

AlexandreSinger (approved):
LGTM

PR #3008, opened 2025-04-29
[Hotfix][CI] Remove PR staling

AmirhosseinPoolad:
This commit sets the number of days before marking issues or PRs as stale to 100 years. This number is overridden for issues to be 1 years but stays 100 years for PRs. This means that PRs effectively do not get marked as stale.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,6 +10,11 @@ jobs:
     steps:
       - uses: actions/stale@v9
         with:
+          # Set default number of days before being marked stale to 100 years
+          # This will be overriden by "days-before-issue-stale" and "days-before-pr-stale"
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -10,6 +10,9 @@ jobs:
     steps:
       - uses: actions/stale@v9
         with:
+          # Set default number of days before being marked stale to 100 years
+          # This will be overriden by "days-before-issue-stale" and "days-before-pr-stale"
```
AlexandreSinger, line 14: You should add a comment for why here. I doubt that the intention is to have our great-great grand-children deal with our old PRs lol.

I would just add a comment saying that, by default, we do not want this running on PRs.

AlexandreSinger (approved):
Minor comment on adding more information to the comment.

PR #3009, opened 2025-04-29
[AP][HotFix] Placed Fixed Blocks First During IP

### LINKED ISSUE #3002: [AP] Fixed IO block handling in the initial placement
While sweeping the high fan-out net thresholds for the issue #2974, the AP was failing especially for 2 circuits in Titan even with high threshold of 128. It is also observed that the issue still exist without any net thresholding (i.e., when using the current master) on the Titan benchmarks with fixed IO blocks. 

The below assertion catches that state in the current initial placement.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d5dc5f7b89d83f2f163e79e189182cf30fbea866/vpr/src/place/initial_placement.cpp#L1771

After discussing with @AlexandreSinger , it is thought that the issue might be about not placing the fixed IO blocks before any other cluster placement.

The circuits facing that issue are sparcT1_chip2_stratixiv_arch_timing and LU_Network_stratixiv_arch_timing, two of which having stratixiv_io blocks in common and these blocks are not passed as fixed blocks to the AP netlist. Therefore, these stratixiv_io blocks might be blocking the other fixed IO blocks from being placed to their desired locations.

AlexandreSinger:
The cost terms in the AP initial placer were not placing fixed blocks early enough, causing other blocks to take their place and causing the initial placer to not return a solution.

Blocks which have region constraints are now placed first based on how constrained they are. More constrained blocks (can only be placed in a smaller region) will be placed first.

Also found that macros that contained fixed blocks were not observing these constraints when calculating the centroid position of the macro. For constrained macros, projected the centroid position onto the partition region to get the closest point in the partition region to the calculated centroid. This new centroid is used to then perform the placement.

Resolves #3002

### DIFF
--- a/vpr/src/place/initial_placement.cpp
+++ b/vpr/src/place/initial_placement.cpp
@@ -24,6 +24,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <cstdlib>
 #include <iterator>
 #include <limits>
 #include <optional>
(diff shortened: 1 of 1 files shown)


vpr/src/place/initial_placement.cpp
```diff
@@ -579,6 +580,43 @@ static t_flat_pl_loc find_centroid_loc_from_flat_placement(const t_pl_macro& pl_
...
+                best_projected_pos.y = proj_y;
+                best_distance = dist;
+            }
+        }
+        VTR_ASSERT_SAFE(best_distance != std::numeric_limits<float>::max());
+        // Return the point within the partition region that is closest to the
+        // original centroid.
+        return best_projected_pos;
+    }
+
```
haydar-c, line 619: This is a clever solution I guess.

haydar-c (approved):
This looks good to me. Thank you for fixing that @AlexandreSinger!

PR #3011, opened 2025-04-29
[Tatum][Parse] Fixed Extraneous Warning With get_clocks

AlexandreSinger:
The get_clocks command is used in an SDC file to reference a set of clocks by name using a regex string. The code to do this tries to produce a warning if get_clocks is used on a regex string and no clocks could be found. The issue is that the code to do this was mistakenly producing this warning for each clock in the circuit. For example, if we had {clk1, clk2, clk3} and we wanted to do "get_clocks {clk3}", we will get two warnings since clk1 and clk2 did not match.

Fixed this by moving the warning out of one loop nest.

I noticed this when I was looking at the output of the directrf circuit and saw weird warnings:
Screenshot from 2025-04-29 17-15-54

This was strange since we clearly create the clocks with those names:
Screenshot from 2025-04-29 17-16-10

I think this was just a mistake and, since it was a warning, no one noticed.

### DIFF
--- a/vpr/src/timing/read_sdc.cpp
+++ b/vpr/src/timing/read_sdc.cpp
@@ -946,12 +946,11 @@ class SdcParseCallback : public sdcparse::Callback {
                         }
                     }
                 }
-
-                if (!found) {
-                    VTR_LOGF_WARN(fname_.c_str(), lineno_,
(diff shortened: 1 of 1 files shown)


vaughnbetz (approved):
Nice catch!

PR #3013, opened 2025-04-30
Add functions to p_tb_type for checking if it's root or primitive

AmirhosseinPoolad:
This PR adds two new functions to the p_tb_type struct, is_root and is_primitive and changes the users to use these new functions instead of writing the logic every time. There's now less code duplication and the code a bit cleaner and more understandable.

I did have two questions about this part of code:

1.  What is the meaning of a t_pb without a name? There's a couple of places that call is_primitive and see if the associated t_pb doesn't have a name which could be cleaned up further
2. Regarding the added TODO comment in cluster_legalizer.cpp:836 of this PR: Should it be an assert?

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1586,7 +1586,7 @@ enum e_directionality {
 };
 
 /* X_AXIS: Data that describes an x-directed wire segment (CHANX)                     *
- * Y_AXIS: Data that describes an y-directed wire segment (CHANY)                     *     
+ * Y_AXIS: Data that describes an y-directed wire segment (CHANY)                     *
  * BOTH_AXIS: Data that can be applied to both x-directed and y-directed wire segment */
 enum e_parallel_axis {
     X_AXIS,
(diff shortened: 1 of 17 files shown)


libs/libarchfpga/src/physical_types.h
```diff
@@ -1068,6 +1068,14 @@ struct t_pb_type {
     t_pb_type_power* pb_type_power = nullptr;
 
     t_metadata_dict meta;
+
+    inline bool is_root() const {
```
AlexandreSinger, line 1072: Please add some doxygen comments for this. It does not have to be long, just to help explain what this means.

libs/libarchfpga/src/physical_types.h
```diff
@@ -1068,6 +1068,14 @@ struct t_pb_type {
...
+        return parent_mode == nullptr;
+    }
+
+    inline bool is_primitive() const {
```
AlexandreSinger, line 1076: Same comment here. A small doxygen comment would be helpful.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -835,6 +833,7 @@ static void try_update_lookahead_pins_used(t_pb* cur_pb,
         // find if this child (primitive) pb block has an atom mapped to it,
         // if yes compute and mark lookahead pins used for that pb block
         AtomBlockId blk_id = atom_to_pb.pb_atom(cur_pb);
+        // TODO: Primitive pb_types should have non-null blif_model. Shoud this be an assertion?
```
AlexandreSinger, line 836: I think you are correct. A primitive without a pb_type does not really make sense. memories can be a bit tricky since they may have blif models at a higher level of the heirarchy; but the leaves should have a blif model as far as I am aware.

AlexandreSinger (commented):
LGTM, some minor comments on documentation.

Regarding t_pb's being unnamed, t_pb's are a heirarchical structure based on the pb_type. I am not exactly sure, but I think there may be pb_types in the heirarchy which are unnamed. You can try to investigate by seeing how t_pbs are constructed if you have the energy for it.

PR #3016, opened 2025-05-01
[STA] Generating SDC Commands Post-Implementation

AlexandreSinger:
Added an option to have VPR generate an SDC file containing the timing commands required for an external timing analysis of the post- implementation netlist to match VPR's timing analysis.

### DIFF
--- a/vpr/src/base/netlist_writer.cpp
+++ b/vpr/src/base/netlist_writer.cpp
@@ -72,6 +72,7 @@
 #include <vector>
 #include "atom_netlist.h"
 #include "atom_netlist_utils.h"
+#include "clock_modeling.h"
 #include "globals.h"
 #include "logic_vec.h"
 #include "netlist_walker.h"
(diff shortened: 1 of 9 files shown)


AlexandreSinger:
@vaughnbetz This is the PR which adds the ability to generate a post-implementation SDC file, as I presented at the meeting this afternoon.

Based on the feedback, I have modified it to always set (non-virtual) clocks to propagated; even when the clock model is set to ideal.

vpr/src/base/netlist_writer.cpp
```diff
@@ -2644,22 +2645,145 @@ std::string join_identifier(std::string lhs, std::string rhs) {
     return lhs + '_' + rhs;
 }
 
+/**
+ * @brief Add the original SDC constraints that VPR used during its flow to the
+ *        given SDC file.
+ */
+void add_original_sdc_to_post_implemented_sdc_file(std::ofstream& sdc_os,
```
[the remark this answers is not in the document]
vaughnbetz: Comment if you append to the sdc_os stream, or expect it to be empty.

vpr/src/base/netlist_writer.cpp
```diff
@@ -2644,22 +2645,145 @@ std::string join_identifier(std::string lhs, std::string rhs) {
...
+    if (non_virtual_clocks.empty()) {
+        return;
+    }
+
+    // Append a header to explain why these commands are added.
+    sdc_os << "\n";
+    sdc_os << "#******************************************************************************#\n";
+    sdc_os << "# The following are clock domains in VPR which have delays on their edges.\n";
+    sdc_os << "#\n";
+    sdc_os << "# Clocks that have been routed or whose input signal have been placed to an\n";
```
vaughnbetz, line 2718: I think you can make the comment more general. "Any non-virtual clock has its delay determined and written out as part of a propagated clock command. If VPR was instructed not to route the clock, this delay will be an underestimate."
"propageted" -> "propagated"

vpr/src/base/netlist_writer.cpp
```diff
@@ -2644,22 +2645,145 @@ std::string join_identifier(std::string lhs, std::string rhs) {
...
+    sdc_os << "# input pad have delays on their signals. Thus, they should be treated as\n";
+    sdc_os << "# propageted (non-ideal) clocks.\n";
+    sdc_os << "#\n";
+    sdc_os << "# Note: Virtual clocks do not get routed and are treated as ideal.\n";
+    sdc_os << "#******************************************************************************#\n";
+
+    // Add the SDC commands to set the non-virtual clocks as propagated (non-ideal);
+    // Note: It was decided that "ideal" clock modeling in VPR should still
```
vaughnbetz, line 2726: "ideal (don't route)"

vpr/src/base/netlist_writer.cpp
```diff
@@ -2644,22 +2645,145 @@ std::string join_identifier(std::string lhs, std::string rhs) {
...
+        sdc_os << "set_propagated_clock ";
+        sdc_os << timing_constraints->clock_domain_name(clock_domain_id);
+        sdc_os << "\n";
+    }
+}
+
+/**
+ * @brief Generates a post-implementation SDC file with the given file name
+ *        based on the timing info and clock modeling set for VPR.
+ */
```
vaughnbetz, line 2740: Make methods static or move comments to the header file.

vpr/src/base/netlist_writer.h
```diff
@@ -16,7 +16,12 @@ class LogicalModels;
  * All written filenames end in {basename}_post_synthesis.{fmt} where {basename} is the
  * basename argument and {fmt} is the file format (e.g. v, blif, sdf)
  */
```
vaughnbetz, line 18: If you have the energy for it, you could document the parameters.

vpr/src/base/read_options.cpp
```diff
@@ -2945,6 +2945,16 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("off")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    analysis_grp.add_argument<bool, ParseOnOff>(args.generate_post_implementation_sdc, "--gen_post_implementation_sdc")
+        .help(
+            "Generates an SDC file including a list of constraints that would "
+            "replicate the timing constraints that the timing analysis within "
+            "VPR followed during the flow. This file is expected to be used "
```
vaughnbetz, line 2952: Match the .rst text comment.

vaughnbetz (approved):
A few nits, but looks good.

vpr/src/base/netlist_writer.cpp
```diff
@@ -2644,22 +2645,145 @@ std::string join_identifier(std::string lhs, std::string rhs) {
...
+ *
+ * This is necessary since VPR decides if clocks are routed or not, which has
+ * affects on how timing analysis is performed on the clocks.
+ */
+void add_propagated_clocks_to_sdc_file(std::ofstream& sdc_os,
+                                       e_clock_modeling clock_modeling) {
+
+    // Ideal and routed clocks are handled by the code below. Other clock models
+    // like dedicated routing are not supported yet.
+    if (clock_modeling != e_clock_modeling::ROUTED_CLOCK && clock_modeling != e_clock_modeling::IDEAL_CLOCK) {
```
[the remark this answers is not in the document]
AlexandreSinger: FYI, I tried this and it mostly worked by the CPD was off by 0.003 ns for the test circuit I used (the CLMA example with an H-tree clock network).

I am not sure where this discrepancy is coming from, the timing graph is open STA seems sounds; I think it may be an issue with the API calls used to parse the timing data from Tatum into the SDF file. Left a TODO to look into this so that this does not block this PR from getting merged.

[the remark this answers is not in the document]
vaughnbetz: Sounds good. I'd enable it then. For the small difference: I agree it's not a blocking issue. I'd look closely at the clock path delays. Perhaps we have a min/max spread and openSTA is doing some clock pessimism removal or some such? If you get listings of the critical paths (including clock delays) in both flows I can go over it with you.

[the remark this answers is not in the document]
AlexandreSinger: Understood. I will raise an issue today to track this. The issue seemed to be caused by the edge delays of clocks reported by Tatum not matching what was written in the SDF file. All paths reported by OpenSTA were off by 0.003 ns.

PR #3018, opened 2025-05-01
[Place] Add a link between IO Blocks

amin1377:
WIP

### DIFF
--- a/libs/libarchfpga/src/physical_types_util.cpp
+++ b/libs/libarchfpga/src/physical_types_util.cpp
@@ -652,6 +652,10 @@ bool is_io_type(t_physical_tile_type_ptr type) {
            || is_output_type(type);
 }
 
+bool is_io_type(t_logical_block_type_ptr logical_block) {
+    return is_io_type(pick_physical_type(logical_block));
+}
+
 std::string block_type_pin_index_to_name(t_physical_tile_type_ptr type, int pin_physical_num, bool is_flat) {
     int max_ptc = get_tile_pin_max_ptc(type, is_flat);
     VTR_ASSERT(pin_physical_num < max_ptc);
--- a/libs/libarchfpga/src/physical_types_util.h
+++ b/libs/libarchfpga/src/physical_types_util.h
@@ -126,6 +126,7 @@ bool is_input_type(t_physical_tile_type_ptr type);
 bool is_output_type(t_physical_tile_type_ptr type);
 ///@brief Returns true if the given physical tile type can implement either a .input or .output block type
 bool is_io_type(t_physical_tile_type_ptr type);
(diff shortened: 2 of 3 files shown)


amin1377:
@vaughnbetz:

In this PR, I updated the initial placement logic as we discussed. Instead of choosing a random location within a fixed range of 15 around the centroid, the new approach gradually increases the search range from 1 up to half the device size. The function will now exhaust all available locations within the current range before expanding it.

[Here](https://docs.google.com/spreadsheets/d/1j6t3k_e-fHOn443mhq6XpB4Sqp2hcE9yiqRBXyORVgs/edit?usp=sharing) are the results for Titan (including initial placement WL and CPD).

* **Initial placement WL** and **CPD** improved significantly — reduced by 48% and 27%, respectively.
* **Final placement WL** remained about the same, while **placed CPD** increased by 1%.
* **Routed WL** was roughly a tie, and **routed CPD** increased by \~2%.

For the large VTR benchmarks ([results](https://docs.google.com/spreadsheets/d/1ckKr_Mz8Ut2yrdR9BSuUQi6PBay-f5hZVLxZgbOfiys/edit?usp=sharing)):

* **Min channel width** increased by 5%.
* **Routed WL** and **CPD** increased by 1% and 5%, respectively.
* However, **initial placement WL** and **CPD** still improved by 46% and 18%.

I was expecting the QoR to at least stay the same, if not improve overall. Do you think it would be worth digging into the circuits with the most degradation to better understand what's going on?

@tpagarani, FYI.

vaughnbetz:
I wonder if the final result quality drop off is due to making a very dense placement with a deep local minimum?  You fill in around a seed block completely.
It's a bit more complicated, but you could choose a target porosity based on how full the design is and search in a ring only a certain percentage (matching the target porosity) of the locations around it.  Or you could directly try to hit a certain porosity by rejecting locations that had an overly dense local region, compared to the overall average density that will be needed.

vaughnbetz:
@soheil also suggests checking if the initial temperature is appropriate, or if it should be lowered.

vaughnbetz:
Looking at the circuits with the most degradation makes sense to me too, to see if you can see what happened with them.

amin1377:
> @soheil also suggests checking if the initial temperature is appropriate, or if it should be lowered.

I considered doing that, but since the initial temperature is determined by performing some moves and assigning a value based on the resulting cost changes, I'm not sure how effective it would be.

vaughnbetz:
For Titan, LU_Network is a clear outlier, and is responsible for most or
all of the degradation I think.  Worth looking at it.
For VTR, the trend is a more consistent, moderate degradation, but there
are also fewer circuits.

Vaughn

PR #3021, opened 2025-05-04
[APPack] Updated Max Candidate Distance Interface

AlexandreSinger:
The max candidate distance is used by APPack to decide which molecules to ignore when packing, based on their distance from the cluster being formed.

Cleaned up the interface of this by pre-computing the max candidate distance of all logical blocks ahead of time and reading from these pre-computed values during packing.

Added a command-line option to allow the user to override some or all of these max distance thresholds. By default, VPR will select values based on the type of logical block and the primitives it contains.

Fixed issue with APPack creating too many IO blocks for some circuits due to the max candidate distance thresholds for IO blocks being too low.

More tuning should be done on these values once the mass legalizer has been cleaned up a bit more.


AlexandreSinger:
New command line argument:
image

@amin1377 I think you will find this option useful. I have not tuned the "auto" values yet; but I plan to use this interface to find good values. This also resolves an issue that @haydar-c resolves. Please review when you have a moment.

vpr/src/pack/appack_max_dist_th_manager.cpp
```diff
@@ -0,0 +1,276 @@
...
+ * TODO: This should be a graph traversal instead of a recursive function.
+ */
+static bool has_memory_pbs(const t_pb_type* pb_type);
+
+APPackMaxDistThManager::APPackMaxDistThManager(bool should_initialize,
+                                               const std::vector<std::string>& max_dist_ths,
+                                               const std::vector<t_logical_block_type>& logical_block_types,
+                                               const DeviceGrid& device_grid) {
+    // Check if we should initialize or not. If not, just exit.
+    if (!should_initialize) {
```
amin1377, line 48: I'm not sure if it's common practice to pass a flag to a constructor to determine whether a class should be initialized or not. I think it's a bit unusual. There are other classes in VPR that follow a different pattern, where an initialize() method is called explicitly after construction. I tend to prefer that approach, but I’ll leave it to your judgment.

amin1377 (changes requested):
Looks good Alex, thanks! Regarding the default scaling values for different block types, I was wondering if you ran an extensive parameter sweep or just tried a bunch of values and found the current ones to work best.

AlexandreSinger:
@amin1377 Regarding where the default values came from, I did perform a parameter sweep around 1 month ago when I was tuning the flow for a paper. However, a lot has changed in the last month which will have caused these values to change. I plan to do another sweep later. This interface will make this sweep significantly easier.

AlexandreSinger:
@amin1377 Thank you so much for the review. I agreed with your comment and updated how the class is initialized. Please let me know if you have any further comments.

amin1377 (approved):
Looks good, Alex! Thanks!

PR #3022, opened 2025-05-04
[LibArchFPGA] Added Library Model IDs as constexpr

AlexandreSinger:
Created static constexpr references to the library model IDs (such as input, output, names, and latch). It is very common to reference these models during execution of VPR. Often we would search for these models in the LogicalModels structure; however, these model IDs will always be the same.

By storing static references to these IDs, the compiler can easily optimize parts of the code which need to check if a model ID is an input/output for example.


AlexandreSinger:
@vaughnbetz Poke. Please review.

vaughnbetz (commented):
Nice cleanup -- thanks!

PR #3025, opened 2025-05-06
Fix the regex pattern for initial placement WL

amin1377:
The wirelength from the initial placement is printed in scientific notation (e.g., `Initial placement BB estimate of wirelength: 2.5437e+07`), but the pattern was expecting an integer. I updated the pattern to accept any numeric format.


AlexandreSinger (approved):
LGTM

Issue #2678, opened 2024-08-06
Make a short coding style guide in the VTR developer guide


vaughnbetz:
Initial ideas below. This shouldn't be too long, and should be reviewed by others to make sure we agree.

1. Write code that is easier for others to read.
2. Add big picture comments.
 - New functions: Doxygen comment in .h file or on static declaration.
Data structures / classes: comment overall use and every member.
Files: comment overall purpose of file.
3. Meaningful variable names. Types usually start with a Capital (older code may not follow this).
4. Avoid long functions whenever possible (refactor into shorter routines).
Details:
Avoid overuse of auto (give rule)
camel or snake case OK.

#### Proposed Behaviour
We have some basic coding guidelines written down.

#### Current Behaviour
We don't have a style guide.  We also are gradually evolving some code style (e.g. types moving from t_ to start with a capital, and using references more and pointers less, using Doxygen comments).

#### Possible Solution
A short style guide under the VTR developer guide.

PR #3026, opened 2025-05-07
[WIP] Coding Style Guide

soheilshahrouz:
Addresses #2678

### DIFF
--- a/doc/src/dev/coding_style.rst
+++ b/doc/src/dev/coding_style.rst
@@ -1,0 +1,46 @@
+
+Use of `auto`
+=============
+
+Use `auto` only when the type is long, complex, or hard to write explicitly.
+
+Examples where `auto` is appropriate:
+
+.. code-block:: cpp
+
+    auto it = container.begin();  // Iterator type is long and not helpful to spell out
+
+    // The return type is std::vector<std::vector<std::pair<int, float>>>
+    auto matrix = generate_adjacency_matrix();
+
+    // Lambdas have unreadable and compiler-generated types — use auto for them
+    auto add = [](int x, int y) { return x + y; };
+    
+
+Avoid `auto` when the type is simple and clear:
+
+.. code-block:: cpp
+
+    // Use type names when they are short and readable.
+    for (RRNodeId node_id : device_ctx.rr_graph.nodes()) {
+        t_rr_node_route_inf& node_inf = route_ctx.rr_node_route_inf[rr_id];
+    }
+
+    int count = 0;
+    std::string name = "example";
+    std::vector<int> numbers = {1, 2, 3};
+
+Avoid:
+
+.. code-block:: cpp
+
+    auto count = 0;                // Simple and obvious type
+    auto name = std::string("x");  // Hides a short, clear type
+
+    for (auto node_id : device_ctx.rr_graph.nodes()) {
+        // node_id is RRNodeId. Write it out for clarity.
+        auto& node_inf = route_ctx.rr_node_route_inf[rr_id];
+        // node_inf is t_rr_node_route_inf. Use the explicit type since it's simple and meaningful.
+    }
+
+Rationale: clear, explicit types help with readability and understanding. Avoid hiding simple types behind `auto`.
--- a/doc/src/dev/coding_style.rst
+++ b/doc/src/dev/coding_style.rst
@@ -15,7 +15,7 @@ Examples where `auto` is appropriate:
 
     // Lambdas have unreadable and compiler-generated types — use auto for them
     auto add = [](int x, int y) { return x + y; };
-    
+
 
 Avoid `auto` when the type is simple and clear:
 
@@ -44,3 +44,53 @@ Avoid:
     }
 
 Rationale: clear, explicit types help with readability and understanding. Avoid hiding simple types behind `auto`.
+
+
+Commenting Style
+================
+Comment types and rules:
+
+- Use `/* ... */` **only** for Doxygen documentation comments.
+  Do **not** use block comments (`/* */`) to describe code logic or individual lines in function bodies.
+
+- Use `//` for all regular comments within function bodies or implementation code.
+
+Formatting rules for `//` comments:
+
+- Always include a space between `//` and the comment text.
+- Use full sentences when appropriate.
+- For multi-line comments, prefix each line with `// `.
+
+Examples (correct usage):
+
+.. code-block:: cpp
+
+    // Check if the node has already been visited
+    if (visited[node_id]) {
+        return;
+    }
+
+    // Enable debug output for route estimation
+    bool debug_enabled = true;
+
+    /**
+     * @brief Estimates the wirelength of a net using bounding box.
+     *
+     * @param net_id ID of the net to analyze.
+     * @return Estimated wirelength in units.
+     */
+    float estimate_wirelength(ClusterNetId net_id);
+
+Incorrect usage:
+
+.. code-block:: cpp
+
+    /* Check if visited */  // Block comments are not allowed here
+    if (visited[node_id]) {
+        return;
+    }
+
+    //Missing space
+    //inconsistent formatting
+
+    /* Non-Doxygen block comment */  // Not permitted
--- a/doc/src/dev/index.rst
+++ b/doc/src/dev/index.rst
@@ -11,5 +11,6 @@ Developer Guide
     c_api_doc
     code_documentation
     tutorials/index
+    coding_style
     ../SUPPORT
     ../LICENSE


soheilshahrouz:
@amin1377 
@AlexandreSinger 
@AmirhosseinPoolad 
@vaughnbetz 

Any comments on what should be added to this guide is appreciated.

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,146 @@
+
+Use of `auto`
+=============
```
AlexandreSinger, line 3: I think that these "======" are for the title of the page only. I think these should be "~~~~~" or something and you should create a title like "Coding Style". This will fix the following issue in the documentation:
image

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,146 @@
...
+
+**1. File-Level Comments**
+- Every source/header file should begin with a brief comment explaining the overall purpose of the file.
+
+**2. Classes and Structs**
+- Add a comment describing what the class or struct is for.
+- Comment every member variable to explain its role.
+
+Example:
+.. code-block:: cpp
```
AlexandreSinger, line 102: I think you need a return carriage before this line for this to show properly. This should also be fixed below. The generated documentation looks weird.

AmirhosseinPoolad:
Fully agree with the usage of auto as documented here. I guess we also should talk about the naming schemes we use. What I personally extrapolated from the code base and remember currently is that Classes are PascalCase, everything else is snake_case, structs are trivial and C-like and start with t_ but it should all be written down.

One more thing is saying that people should prioritize (but not blindly use in all circumstances) using vtr:: containers over STL and use the various LOG/ERROR macros over directly using printf/cout.

AlexandreSinger (commented):
Hi Soheil, thank you so much for starting to add documentation on coding style! This is extremely important for such a large project such as VTR.

I think the coding style that we tell the students in ECE297 will be a good reference for a checklist of things to add to this documentation:
[tutorial5_code_style.pdf](https://github.com/user-attachments/files/20091463/tutorial5_code_style.pdf)

Specifically this slide:
image

I think the major things that are missing are:

- The variable, function, and classes naming conventions that we use in VTR
- Keeping function short (i.e. pulling out code into helper methods)
- Self-checking code (i.e. proper use of VTR_ASSERT)

I think the three above are very important to document since we can point people to this page if they raise a PR that does not have proper style; since I believe these three things (as well as the things you have documented) would block a PR from being merged in my opinion!

I know our goal is to keep this style guide brief; but if you would like more things to add to the style guide, I can point you to some of the different style guides that I am familiar with!

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,217 @@
...
+  .. code-block:: cpp
+
+      int net_count;
+      void estimate_wirelength();
+
+- **Private member variables and private methods** should end with an underscore `_`.
+
+  .. code-block:: cpp
+
+      class Example {
```
vaughnbetz, line 39: Note: this convention is not followed very consistently today but we should move to it.

vaughnbetz: Should we say that?

vaughnbetz: Or maybe we just say that at the top of the style guide.

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,217 @@
...
+~~~~~~~~~~~~~~~~
+
+Comments help readers understand the purpose, structure, and reasoning behind the code.
+This section outlines when and how to use comments in a consistent and helpful way.
+
+General Rules
+-------------
+
+- Focus on explaining *why* the code exists or behaves in a certain way.
+- Do not explain *what* the code is doing if it's already clear from the code itself.
```
vaughnbetz, line 128: .h file should say what the code does; if not obvious explain why the code behaves that way as well.
.cpp file should explain how the code works if it is not obvious.

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,217 @@
...
+- Keep comments up to date. Outdated comments are worse than no comments.
+- Use Doxygen-style `/** ... */` or `///` for documenting APIs, classes, structs, members, and files.
+
+Comment types and rules:
+
+- Use `/* ... */` **only** for Doxygen documentation comments.
```
vaughnbetz, line 134: I'd weaken this a bit / explain the reasoning.  Seem a bit strong on the /*  vs. //

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,217 @@
...
+.. code-block:: cpp
+
+    // Skip ignored nets
+    if (net.is_ignored()) {
+        continue;
+    }
+
+
+
+
```
vaughnbetz, line 216: Other guidelines:
- Prefer short functions, so they can be reused and their purpose can be more easily commented.
- Avoid unnecessary use of complex language features; prefer easier-to-read code
- Check the code and documentation at <link> to see if a utility or routine already exists before making a new one

vaughnbetz: assertion guidelines:
- use assertions and data structure validators wherever possible.
- In CPU-critical code, use VTR_ASSERT_SAFE for potentially expensive checks `<link>`

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,217 @@
...
+This section outlines when and how to use comments in a consistent and helpful way.
+
+General Rules
+-------------
+
+- Focus on explaining *why* the code exists or behaves in a certain way.
+- Do not explain *what* the code is doing if it's already clear from the code itself.
+- Keep comments up to date. Outdated comments are worse than no comments.
+- Use Doxygen-style `/** ... */` or `///` for documenting APIs, classes, structs, members, and files.
+
```
vaughnbetz, line 131: Follow the instructions at `<link>` to add new utilities, APIs or internal functionality doxygen documentation to the html pages at `<link>`.

vaughnbetz (commented):
Thanks!

AlexandreSinger:
@soheilshahrouz We should also add a section on the use of header files. Specifically that "#pragma once" should be the first line in any header file (even before file comments).

Through looking around online, for large projects it is recommended for "#pragma once" to be declared before the file comment as well as code. The idea is that the file comment will be duplicated in files that may include it multiple times, which for large projects can slow down the compilation speed due to IO.

I am currently working on a PR that moves all include files to this style.

doc/src/dev/coding_style.rst
soheilshahrouz: I think having a uniform commenting style is helpful. This doesn’t forbid multiline comments. It just means implementation-related comments should start with `//`, including each line of a multiline comment. This makes it easy to distinguish implementation comments from documentation comments

doc/src/dev/coding_style.rst
soheilshahrouz: I am not sure which page in the docs I should point to. Can you send the link here?

vaughnbetz: first link: https://docs.verilogtorouting.org/en/latest/dev/code_documentation/

and 

for second link, could do:

https://docs.verilogtorouting.org/en/latest/api/vpr/
(or just a generic link to the docs, which would end at /latest)

vaughnbetz:
1. Should move the 'auto' discussion to the end.  It's not the most important point.
2. We should tell them to use the vtr logging utilities, and error functions and have a link to the docs for them.
3. We should tell them to familiarize themselves with the vtr container utilities (e.g. vectors with type-safe indices, matrices) and use them -- give a link.

AmirhosseinPoolad:
image
This section's formatting seems to be a little weird with some bold items and some random newlines.

doc/src/dev/coding_style.rst
```diff
@@ -0,0 +1,349 @@
...
+      }
+
+
+
+
+Avoid Unnecessary Complexity
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Prefer easier-to-read language features and idioms.
+- Avoid overusing advanced C++ features (e.g., template metaprogramming, operator overloading, SFINAE) unless they are essential.
```
duck2, line 222: My (unsolicited) opinion on this: I agree with the sentiment, but maybe the examples aren't very accurate. Operator overloading has been in C++ since 1998 and should be OK unless people want to `+` rr graphs together. Also the `std::vector<int>` example below relies on knowing the compiler would optimize out that copy. It could return a const reference, but if the underlying storage isn't exactly a vector, it makes sense to return a `view` or `range` with an `auto` return type. For instance, Tatum's interfaces are written assuming `std::vector` and I once gave up trying to make a parallel incremental analyzer based on `tbb::concurrent_vector` because of this. All in all, I think advanced and new features should be OK as long as they make things simpler.

I think it's good to advise against deep abstract class hierarchies here as it's really hard for tools to jump to the definition of a virtual function (another tatum thing but it may have examples in other places). The combination of templates and virtual functions is particularly hard to navigate. Maybe a piece of code trying to be generic should use one or the other, and not both (so don't do what uxsdcxx-generated code does. or NetlistRouter haha)

soheilshahrouz:
> @soheilshahrouz : I think I should merge this and you can walk people through it on Thursday. Agreed? Let me know if I should mere now, or if you want to make any final tweaks in response to the feedback above first.

I applied the remaining comments. This is ready to merge.

PR #3031, opened 2025-05-10
Congestion-Aware Initial Accumulated Cost

soheilshahrouz:
This PR adds a new option to bias routing away from congested channels during early iterations by adjusting the initial accumulated cost based on estimated CHANX/CHANY utilization.


soheilshahrouz:
TODO: compute channel widths by iterating over RR nodes.

soheilshahrouz:
titan_quick_qor for channel width of 300

|       | place_time | wl         | cpd        | route_time | pack_time | heap_pops   | total_swaps |
|-------------|------------|------------|------------|------------|-----------|-------------|--------------|
| ratio  | 1.0007     | 0.9991     | 0.9965     | 0.9993     | 0.9950    | 1.0067      | 1           |

soheilshahrouz:
The impact of increasing `acc_fac` to 1.1 on Titan benchmarks

| W       | place_time   | wl          | cpd         | route_time  | pack_time   | heap_pops    | total_swaps  |
|---------------|--------------|-------------|-------------|-------------|-------------|--------------|---------------|
| 250 | 1.017395588  | 1.001856283 | 1.00593143  | 1.043113712 | 1.000524706 | 1.054397302  | 1             |
| 270  | 1.008064369  | 1.000474773 | 1.004645215 | 0.9784847728 | 1.003675494 | 1.000754474  | 1             |

libs/librrgraph/src/base/rr_graph_type.h
```diff
@@ -14,7 +14,7 @@ struct t_chan_width {
     std::vector<int> y_list;
 };
 
-enum e_route_type {
+enum class e_route_type {
```
vaughnbetz, line 17: Doxygen comment (you touched it, you document it!)

libs/librrgraph/src/base/rr_node_types.h
```diff
@@ -128,3 +128,45 @@ struct t_rr_rc_data {
 // This is the data type of fast lookups of an rr-node given an (rr_type, layer, x, y, and the side)
 //[0..num_rr_types-1][0..num_layer-1][0..grid_width-1][0..grid_height-1][0..NUM_2D_SIDES-1][0..max_ptc-1]
 typedef vtr::array<e_rr_type, vtr::NdMatrix<std::vector<RRNodeId>, 4>, (size_t)e_rr_type::NUM_RR_TYPES> t_rr_node_indices;
+
+/**
```
vaughnbetz, line 132: This is better than vpr_types.h.  But should it be in a route_common.h or router.h file instead of here?

vpr/src/base/read_options.cpp
```diff
@@ -2881,6 +2881,20 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("map")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    route_timing_grp.add_argument<double>(args.router_initial_acc_cost_chan_congestion_threshold, "--router_initial_acc_cost_chan_congestion_threshold")
```
vaughnbetz, line 2884: To turn this on by default you should get data on VTR and Koios designs too to make sure it is a good idea in general (at least for the default flows on each).
Can run Koios just at the default 300 channel width, but make sure it isn't hurt.
VTR runs Wmin*1.3; care mostly about the final routing, but you can also check what happens to the binary search.
Also need to check 3D chips (talk to Amin for best test).

vpr/src/base/stats.cpp
```diff
@@ -129,6 +129,41 @@ void routing_stats(const Netlist<>& net_list,
...
+std::pair<vtr::NdMatrix<int, 3>, vtr::NdMatrix<int, 3>> calculate_channel_width() {
+    const auto& device_ctx = g_vpr_ctx.device();
+    const auto& rr_graph = device_ctx.rr_graph;
+
+    auto chanx_width = vtr::NdMatrix<int, 3>({{(size_t)device_ctx.grid.get_num_layers(),
+                                               device_ctx.grid.width(),
+                                               device_ctx.grid.height()}},
+                                             0);
+
+    auto chany_width = vtr::NdMatrix<int, 3>({{(size_t)device_ctx.grid.get_num_layers(),
```
vaughnbetz, line 141: We might be able to use these to calculate more robust channel widths for drawing ... not sure if we already did that and not needed in this PR, but good to check to see.  Also could use for the linear congestion costs in the placer.
If you use more widely, we might want to move them to a new file under rr_graph/

vpr/src/base/stats.cpp
```diff
@@ -129,6 +129,41 @@ void routing_stats(const Netlist<>& net_list,
...
+                                             0);
+
+    for (RRNodeId node_id : rr_graph.nodes()) {
+        e_rr_type rr_type = rr_graph.node_type(node_id);
+
+        if (rr_type == e_rr_type::CHANX) {
+            int y = rr_graph.node_ylow(node_id);
+            int layer = rr_graph.node_layer(node_id);
+            for (int x = rr_graph.node_xlow(node_id); x <= rr_graph.node_xhigh(node_id); x++) {
+                chanx_width[layer][x][y]++;
```
vaughnbetz, line 153: I suspect you should increment by rr_node.capacity.  It should make this code work for global routing, and if someone makes a regular pattern that can be represented as a bus.

vpr/src/base/stats.cpp
```diff
@@ -129,6 +129,41 @@ void routing_stats(const Netlist<>& net_list,
...
+            }
+        } else if (rr_type == e_rr_type::CHANY) {
+            int x = rr_graph.node_xlow(node_id);
+            int layer = rr_graph.node_layer(node_id);
+            for (int y = rr_graph.node_ylow(node_id); y <= rr_graph.node_yhigh(node_id); y++) {
+                chany_width[layer][x][y]++;
```
vaughnbetz, line 159: Incremental by capacity.

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1757,3 +1755,108 @@ std::pair<t_bb, t_bb> NetCostHandler::union_2d_bb_incr(ClusterNetId net_id) cons
 
     return std::make_pair(merged_num_edge, merged_bb);
 }
+
+std::pair<vtr::NdMatrix<double, 3>, vtr::NdMatrix<double, 3>> NetCostHandler::estimate_routing_chann_util() const {
```
vaughnbetz, line 1759: chann vs. chan?

vpr/src/place/net_cost_handler.h
```diff
@@ -126,6 +126,16 @@ class NetCostHandler {
      */
     double get_total_wirelength_estimate() const;
 
+    /**
+     * @brief Estimates routing channel utilization.
+     *
+     * For each net, distributes estimated wirelength across its bounding box
+     * and accumulates demand for different routing channels. Normalizes by channel widths.
+     *
+     * @return Pair of matrices with relative CHANX and CHANY utilization.
```
vaughnbetz, line 135: List the dimension order here.

vaughnbetz: Normalizes by channel width (e.g. a value of 0.5 means 50% of the wiring in a channel is expected to be used).

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1757,3 +1755,108 @@ std::pair<t_bb, t_bb> NetCostHandler::union_2d_bb_incr(ClusterNetId net_id) cons
...
+    auto chanx_util = vtr::NdMatrix<double, 3>({{(size_t)device_ctx.grid.get_num_layers(),
+                                                 device_ctx.grid.width(),
+                                                 device_ctx.grid.height()}},
+                                               0);
+
+    auto chany_util = vtr::NdMatrix<double, 3>({{(size_t)device_ctx.grid.get_num_layers(),
+                                                 device_ctx.grid.width(),
+                                                 device_ctx.grid.height()}},
+                                               0);
+
```
vaughnbetz, line 1772: Give some idea of what the algorithm is.  You are assuming the expected wiring in each dimension is evenly spread over the net's bounding box. You can also comment on how / why the expected wirelength is allocated to different dimensions.

vpr/src/place/placer.cpp
```diff
@@ -371,14 +371,22 @@ void Placer::place() {
...
     // the placement location variables should be unlocked before being accessed
-    place_ctx.unlock_loc_vars();
+    mutable_palce_ctx.unlock_loc_vars();
 
     // copy the local location variables into the global state
-    auto& global_blk_loc_registry = place_ctx.mutable_blk_loc_registry();
+    auto& global_blk_loc_registry = mutable_palce_ctx.mutable_blk_loc_registry();
     global_blk_loc_registry = placer_state_.blk_loc_registry();
 
+    auto [chanx_util, chany_util] = net_cost_handler_.estimate_routing_chann_util();
```
vaughnbetz, line 385: Not sure we should have a side effect like this in a routine ...suggest instead directly calling a routine to do this at the end of placement.  Or you could rename it but still not sure that's as good.

vaughnbetz: At least rename and update the comment of this routine.

vpr/src/route/route_common.cpp
```diff
@@ -401,38 +411,73 @@ void free_route_structs() {
...
 
     for (auto i : device_ctx.rr_node_to_non_config_node_set) {
         route_ctx.non_configurable_bitset.set(i.first, true);
     }
 }
 
-void reset_rr_node_route_structs() {
-    /* Resets some extra information about each rr_node that is used only   *
-     * during routing.                                                         */
+static float comp_initial_acc_cost(RRNodeId node_id,
```
vaughnbetz, line 431: Explain the reasoning / intuition behind this.  "Usually acc_cost starts at 1 for all rr_nodes, but we ... to push the router to ..."

vpr/src/route/router_lookahead/router_lookahead.h
```diff
@@ -82,12 +82,6 @@ class RouterLookahead {
 /**
  * @brief Force creation of lookahead object.
  * @attention This may involve recomputing the lookahead, so only use if lookahead cache cannot be used.
```
vaughnbetz, line 84: Not sure it is worth deleting @param just because it's not fully filled in (better to fill in the ones you know ...)

vaughnbetz (changes requested):
Needs lots of testing to be on by default:
Koios QoR
VTR QoR on binary search and final routing at 1.3Wmin
3D FPGAs work / have reasonable QoR.

Follow on possible work: check if drawing depends on arch file channels -> update if needed.

vaughnbetz:
Could use this as an opportunity to make 
1. Drawing use channel widths calculated from the rr-graph (not the arch file) so it draws properly with a read-in rr-graph.  See https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/ea896e238ca18316dcc1f4c228bb940e57afb357/vpr/src/draw/draw.cpp#L581
2. Placement cost (linear congestion / scaling of wiring by average channel width over bb) also respond to the rr-graph, not the arch file.

vaughnbetz:
Probably moving this code out of stats.cpp would also be good -- it'll become more fundamental.

soheilshahrouz:
VTR benchmarks minimum channel width search

| Metric                                                      | feature.txt      |
|-------------------------------------------------------------|------------------|
| vtr_flow_elapsed_time                                       | 1.109599868      |
| odin_synth_time                                             |                  |
| parmys_synth_time                                           | 0.9942209611     |
| abc_depth                                                   | 1                |
| abc_synth_time                                              | 1.001840561      |
| num_clb                                                     | 1                |
| num_memories                                                | 1                |
| num_mult                                                    | 1                |
| max_vpr_mem                                                 | 0.9914816969     |
| num_pre_packed_blocks                                       | 1                |
| num_post_packed_blocks                                      | 1                |
| device_grid_tiles                                           | 1                |
| pack_time                                                   | 1.000728945      |
| placed_wirelength_est                                       | 1                |
| place_time                                                  | 1.025983213      |
| placed_CPD_est                                              | 1                |
| min_chan_width                                              | 0.9906351292     |
| routed_wirelength                                           | 1.002237975      |
| min_chan_width_route_time                                   | 1.352100907      |
| crit_path_routed_wirelength                                 | 1.001354999      |
| critical_path_delay                                         | 1.000111178      |
| geomean_nonvirtual_intradomain_critical_path_delay          | 1.000111178      |
| crit_path_route_time                                        | 0.9794380194     |

Minimum channel width was reduced in two circuits.

vaughnbetz:
Looks good, thanks.  We should get koios as well, since this is a big change.

vpr/src/base/stats.cpp
soheilshahrouz: Opened issue #3100

soheilshahrouz:
vtr_reg_nightly_test7/vtr_reg_qor_large_run_flat

| Metric                                              | feature.txt |
|-----------------------------------------------------|-------------|
| vtr_flow_elapsed_time                              | 1.043       |
| odin_synth_time                                    |             |
| parmys_synth_time                                  | 1.038       |
| abc_depth                                          | 1           |
| abc_synth_time                                     | 1.025       |
| num_clb                                            | 1           |
| num_memories                                       | 1           |
| num_mult                                           | 1           |
| max_vpr_mem                                        | 1.003       |
| num_pre_packed_blocks                              | 1           |
| num_post_packed_blocks                             | 1           |
| device_grid_tiles                                  | 1           |
| pack_time                                          | 1.005       |
| placed_wirelength_est                              | 1           |
| place_time                                         | 1.024       |
| placed_CPD_est                                     | 1           |
| routed_wirelength                                  | 1.011       |
| critical_path_delay                                | 0.993       |
| geomean_nonvirtual_intradomain_critical_path_delay | 0.993       |
| crit_path_route_time                               | 1.049       |

soheilshahrouz:
@vaughnbetz 
I added QoR for flat routing and 3D. This PR is ready to merge

soheilshahrouz:
> Please summarize where we see benefits, and take a look at the minimum channel width results (cpu time is up --> just the two circuits with better results?)

The minimum channel width is reduced for both `mcml` and `LU32PEEng`. For `LU32PEEng`, this leads to an 8.6x increase in routing time at minimum channel width. `mcml` experiences only a minimum channel width route time increase of 1.07x.

When the channel width is low enough to cause congestion, routing converges faster if the `acc_cost` of RR nodes in congested channels starts with a value larger than 1.

vpr/src/base/vpr_context.h
```diff
@@ -559,6 +561,11 @@ struct RoutingContext : public Context {
 
     /** Is flat routing enabled? */
     bool is_flat;
+
+    /// @brief Post-placement estimate of CHANX routing utilization per (layer, x, y) location.
```
vaughnbetz: We could minimize the lifetime of these variables; we can load them, update the accumulated costs and then destroy them.

vaughnbetz: Test with --route (only) to see this works.  We might have a flow issue where you have to go through placement to run routing.  We could have the router call the placement bounding box etc. code instead in order to localize the creation of these matrices.

vaughnbetz: If you leave the lifetime of these as "placer creates, router consumes" I think that should be commented here.  But I'm not crazy about it if there is an alternative ...

vaughnbetz (approved):
Some testing for --route, and think about whether the state transfer from place->route would be cleaner if the router loaded everything.

PR #3037, opened 2025-05-14
[AP][Solver] Enabled Parallel Eigen

### LINKED ISSUE #3030: [AP] Eigen automatically and without asking the user uses multiple threads when VPR is compiled with -fopenmp
When you compile VPR with OpenMP, Eigen automatically spawns the maximum number of threads it can. While VPR is not currently compiled with OpenMP, the number of threads used by Eigen should be ideally explicitly set to either a command-line option provided by user or a fixed number to be safe. I ran into this issue while working on parallel packing and I kind of fixed it by adding "Eigen::setNbThreads(1);" in AnalyticalSolver::AnalyticalSolver but I really don't think that's the best place for it.

Relevant document:
https://eigen.tuxfamily.org/dox/TopicMultiThreading.html

AlexandreSinger:
The Eigen solver has the ability to use OpenMP to run the solver computations in parallel. Made the AP flow use the num_workers option to set the number of threads that Eigen can use.

VPR did not have the ability to build with OpenMP in its CMAKE. Added an option to the CMAKE to allow the user to enable OpenMP.

Resolves #3030 

Currently running the VTR circuits using 4 threads, will show the results. As a preview, MCML's Global Placement went from 141 seconds to 92 seconds on 4 threads.

I made OpenMP disabled by default for now since only the AP flow uses OpenMP. Perhaps we can enable this by default only when AP is enabled or just always enable it.

### DIFF
--- a/vpr/src/analytical_place/analytical_solver.cpp
+++ b/vpr/src/analytical_place/analytical_solver.cpp
@@ -11,6 +11,7 @@
 #include <cstdio>
 #include <limits>
 #include <memory>
+#include <thread>
 #include <utility>
 #include <vector>
 #include "PreClusterTimingManager.h"
@@ -32,6 +33,7 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wnull-dereference"
 
+#include <Eigen/src/Core/products/Parallelizer.h>
 #include <Eigen/src/SparseCore/SparseMatrix.h>
 #include <Eigen/SVD>
 #include <Eigen/Sparse>
@@ -48,7 +50,22 @@ std::unique_ptr<AnalyticalSolver> make_analytical_solver(e_ap_analytical_solver 
                                                          const AtomNetlist& atom_netlist,
                                                          const PreClusterTimingManager& pre_cluster_timing_manager,
                                                          float ap_timing_tradeoff,
+                                                         unsigned num_threads,
                                                          int log_verbosity) {
+#ifdef EIGEN_INSTALLED
+    // Set the number of threads that Eigen can use.
+    unsigned eigen_num_threads = num_threads;
+    if (num_threads == 0) {
+        eigen_num_threads = std::thread::hardware_concurrency();
+    }
+    // Set the number of threads globally used by Eigen (if OpenMP is enabled).
+    // NOTE: Since this is a global update, all solvers will have this number
+    //       of threads.
+    Eigen::setNbThreads(eigen_num_threads);
+#else
+    (void)num_threads;
+#endif // EIGEN_INSTALLED
+
     // Based on the solver type passed in, build the solver.
     switch (solver_type) {
         case e_ap_analytical_solver::QP_Hybrid:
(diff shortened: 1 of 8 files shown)


AlexandreSinger:
@AmirhosseinPoolad You may find this update interesting since I added a CMAKE option which enables OpenMP.

AlexandreSinger:
Results on largest VTR circuits:

  | OpenMP Disabled | OpenMP Enabled 1T | OpenMP Enabled 4T
-- | -- | -- | --
GP run time | 1 | 0.9956934236 | 0.6646688318
Total run time | 1 | 0.9909095635 | 0.8437623927

Turning on parallelism for 4 threads reduced the overall run time by 15%. I think this number can be even larger with bigger circuits such as the Titan circuits.

AlexandreSinger:
@vaughnbetz This is ready for review. It looks like this does not have perfect scaling, but it may scale better to larger circuits. Perhaps this can be explored more in a future paper.

What do you think about having OpenMP disabled by default? I could have it enabled by default and then disable it if it cannot be found on the system; or only turn it on if Eigen is installed. I do not think having OpenMP enabled and not used will have any affect; but I wanted to see what you thought.

AmirhosseinPoolad:
> disable it if it cannot be found on the system

About this, I don't think VTR supports any system/compiler that does not support OpenMP. It's part of GCC/Clang/MSVC so we're not really changing anything with regards to supported platforms.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -48,7 +49,17 @@ std::unique_ptr<AnalyticalSolver> make_analytical_solver(e_ap_analytical_solver
                                                          const AtomNetlist& atom_netlist,
                                                          const PreClusterTimingManager& pre_cluster_timing_manager,
                                                          float ap_timing_tradeoff,
+                                                         unsigned num_threads,
                                                          int log_verbosity) {
+#ifdef EIGEN_INSTALLED
+    // Set the number of threads globally used by Eigen (if OpenMP is enabled).
+    // NOTE: Since this is a global update, all solvers will have this number
+    //       of threads.
+    Eigen::setNbThreads(num_threads);
```
AmirhosseinPoolad, line 58: In the docs for -j it says:

> 0 implies VPR may execute with up to the maximum concurrency supported by the host machine

I don't think this is compliant with that.

Also, not for this PR but in general, there should be better documentation on what exactly VPR can do in parallel in that section and maybe link to the relevant options for parallel routing etc.

AmirhosseinPoolad (commented):
Thanks for the PR! I personally think OpenMP should be on by default to make --num_workers work as expected but it's okay either way. I only had one super small problem with the implementation.

vpr/src/analytical_place/analytical_solver.cpp
vaughnb-cerebras: Agreed, we should follow that behaviour.
If you have the energy to add the parallel algorithms to the help Alex, they are:
Current parallel algorithms include:
- Timing analysis
- Routing (if --routing_algorithm parallel or parallel_decomp is selected)
- Portions of the analytic placement flow (if --analytical_place is specified)

vaughnb-cerebras (commented):
Excellent result!  I agree with @AmirhosseinPoolad 's  --j 0 comment.
We should enable open mp builds by default in my opinion, and disable it only if it is not found (as you suggest).

vpr/src/analytical_place/analytical_solver.cpp
AmirhosseinPoolad: I'll open another PR to add this information to the docs.

PR #3040, opened 2025-05-14
Eliminate vtr_free and vtr_malloc/vtr_calloc/vtr_realloc

SamuelHo10:
Replaced most of the free/malloc/calloc with new/delete in ArchFPGA library. I left a few unchanged because they either conflict with vtr::strdup or are used by other libraries such as VPR or Odin. This is my first pull request, so please let me know if I am forgetting anything.


libs/libarchfpga/src/arch_util.cpp
```diff
@@ -163,7 +163,7 @@ void free_arch(t_arch* arch) {
     vtr::free(arch->architecture_id);
 
     if (arch->clocks) {
-        vtr::free(arch->clocks->clock_inf);
+        delete[] arch->clocks->clock_inf;
```
soheilshahrouz, line 166: `t_clock_arch` holds an array of `t_clock_network` elements. It can be replaced with `std::vector<t_clock_network>`.

soheilshahrouz: I think we should get rid of `t_clock_arch` struct

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -335,17 +335,17 @@ static void free_pb_type(t_pb_type* pb_type) {
                 for (int m = 0; m < pb_type->modes[i].interconnect[j].annotations[k].num_value_prop_pairs; ++m) {
                     vtr::free(pb_type->modes[i].interconnect[j].annotations[k].value[m]);
                 }
-                vtr::free(pb_type->modes[i].interconnect[j].annotations[k].prop);
-                vtr::free(pb_type->modes[i].interconnect[j].annotations[k].value);
+                delete[] pb_type->modes[i].interconnect[j].annotations[k].prop;
+                delete[] pb_type->modes[i].interconnect[j].annotations[k].value;
```
soheilshahrouz, line 339: Instead of using `int*` and `char**` in `t_pin_to_pin_annotation`, we can use a `std::vector<std::pair<e_pin_to_pin_delay_annotations, std::string>>`. This way, when a `t_pin_to_pin_annotation` object goes out of scope, `prop`-`value` pairs are deleted automatically.

libs/libarchfpga/src/echo_arch.cpp
```diff
@@ -383,7 +383,7 @@ static void print_model(FILE* echo, const t_model& model) {
 static void PrintPb_types_rec(FILE* Echo, const t_pb_type* pb_type, int level, const LogicalModels& models) {
     char* tabs;
 
-    tabs = (char*)vtr::malloc((level + 1) * sizeof(char));
+    tabs = new char[level + 1];
```
soheilshahrouz, line 386: We should replace this with `std::string` to make the lifetime management automatic.
You can use `c_str()` method to pass the underlying `char*` pointer to `fprintf()`

soheilshahrouz (commented):
Thanks @SamuelHo10 

I left a few comments suggesting how the code can be improved in future PRs.

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -495,13 +495,12 @@ void alloc_and_load_default_child_for_pb_type(t_pb_type* pb_type,
     copy->num_pb = 1;
 
     /* Power */
-    copy->pb_type_power = (t_pb_type_power*)vtr::calloc(1,
-                                                        sizeof(t_pb_type_power));
+    copy->pb_type_power = new t_pb_type_power();
```
vaughnbetz, line 498: calloc 0-initializes the data it allocates.  Looking at the code in physical_types.h, is no default constructor (i.e. the constructor with no arguments) for t_pb_type_power defined. Hence the compiler will make one. That constructor does not initialize built in types like integers and floats. 

Hence this code change is dangerous: we have gone from a block of 0-initialized memory to a struct that has unknown state.  Probably @jgoeders used calloc on purpose to initialize the power elements to 0.

To convert calloc uses like this to new, you have to make a default constructor for the struct/class in question and initialize all its data members to 0.

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -495,13 +495,12 @@ void alloc_and_load_default_child_for_pb_type(t_pb_type* pb_type,
...
     copy->pb_type_power->estimation_method = power_method_inherited(pb_type->pb_type_power->estimation_method);
 
     /* Ports */
     copy->num_ports = pb_type->num_ports;
-    copy->ports = (t_port*)vtr::calloc(pb_type->num_ports, sizeof(t_port));
+    copy->ports = new t_port[pb_type->num_ports]();
```
vaughnbetz, line 503: Unsafe unless you make a default constructor for t_port, or you confirm we don't rely on one.  Easier to make a default constructor, and useful.

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -514,8 +513,7 @@ void alloc_and_load_default_child_for_pb_type(t_pb_type* pb_type,
         copy->ports[i].index = pb_type->ports[i].index;
         copy->ports[i].absolute_first_pin_index = pb_type->ports[i].absolute_first_pin_index;
 
-        copy->ports[i].port_power = (t_port_power*)vtr::calloc(1,
-                                                               sizeof(t_port_power));
+        copy->ports[i].port_power = new t_port_power();
```
vaughnbetz, line 516: Unsafe without a default constructor for t_port_power that sets everything to 0.

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -549,8 +547,8 @@ void alloc_and_load_default_child_for_pb_type(t_pb_type* pb_type,
         copy->annotations[i].format = pb_type->annotations[i].format;
         copy->annotations[i].type = pb_type->annotations[i].type;
         copy->annotations[i].num_value_prop_pairs = pb_type->annotations[i].num_value_prop_pairs;
-        copy->annotations[i].prop = (int*)vtr::malloc(sizeof(int) * pb_type->annotations[i].num_value_prop_pairs);
-        copy->annotations[i].value = (char**)vtr::malloc(sizeof(char*) * pb_type->annotations[i].num_value_prop_pairs);
+        copy->annotations[i].prop = new int[pb_type->annotations[i].num_value_prop_pairs];
+        copy->annotations[i].value = new char*[pb_type->annotations[i].num_value_prop_pairs];
```
vaughnbetz, line 551: If you follow @soheilshahrouz 's suggestion on a vector of pairs of ints/strings above this copy will become easier -- should turn into something like copy->annotations = pb_type->annotations since the vector/pair/string classes know how to properly do a deep-copy assignment.

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -613,10 +610,9 @@ void ProcessLutClass(t_pb_type* lut_pb_type) {
 
     lut_pb_type->modes[0].interconnect[0].parent_mode_index = 0;
     lut_pb_type->modes[0].interconnect[0].parent_mode = &lut_pb_type->modes[0];
-    lut_pb_type->modes[0].interconnect[0].interconnect_power = (t_interconnect_power*)vtr::calloc(1, sizeof(t_interconnect_power));
+    lut_pb_type->modes[0].interconnect[0].interconnect_power = new t_interconnect_power();
```
vaughnbetz, line 613: Unsafe without a default constructor for t_interconnect_power

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -613,10 +610,9 @@ void ProcessLutClass(t_pb_type* lut_pb_type) {
...
 
-    lut_pb_type->modes[0].interconnect[0].annotations = (t_pin_to_pin_annotation*)vtr::calloc(lut_pb_type->num_annotations,
-                                                                                              sizeof(t_pin_to_pin_annotation));
+    lut_pb_type->modes[0].interconnect[0].annotations = new t_pin_to_pin_annotation[lut_pb_type->num_annotations]();
```
vaughnbetz, line 615: calloc -> new: unsafe unless you have a default constructor for annotations (if you use the vector of pair<int,string> suggested by @soheilshahrouz you will have a default constructor since all standard library classes have one.

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -643,30 +637,29 @@ void ProcessLutClass(t_pb_type* lut_pb_type) {
...
-            free(lut_pb_type->annotations[i].value[j]);
+            vtr::free(lut_pb_type->annotations[i].value[j]);
         }
-        free(lut_pb_type->annotations[i].value);
-        free(lut_pb_type->annotations[i].prop);
+        delete[] lut_pb_type->annotations[i].value;
+        delete[] lut_pb_type->annotations[i].prop;
         if (lut_pb_type->annotations[i].input_pins) {
-            free(lut_pb_type->annotations[i].input_pins);
+            vtr::free(lut_pb_type->annotations[i].input_pins);
```
vaughnbetz, line 652: Possibly we could convert these C-strings to std::string so we don't need free calls for them, but it depends on how many places they are used (how much code you have to go change to use C++ strings)

vaughnbetz: Same for .output_pins and .clock

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -944,7 +934,7 @@ void SyncModelsPbTypes_rec(t_arch* arch,
 
         pb_type->model_id = model_match_prim_id;
         vtr::t_linked_vptr* old = model_match_prim.pb_types;
-        model_match_prim.pb_types = (vtr::t_linked_vptr*)vtr::malloc(sizeof(vtr::t_linked_vptr));
+        model_match_prim.pb_types = new vtr::t_linked_vptr;
```
vaughnbetz, line 937: We could probably change this linked list (which currently stores pb_type* pointers) to a vector of pb_type* (or maybe even just to direct pb_type elements).  Probably don't do that yet though ... we should refactor in stages.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -219,7 +219,7 @@ static t_port get_generic_port(t_arch* arch,
     port.is_non_clock_global = false;
     port.model_port = nullptr;
     port.port_class = vtr::strdup(nullptr);
-    port.port_power = (t_port_power*)vtr::calloc(1, sizeof(t_port_power));
+    port.port_power = new t_port_power();
```
vaughnbetz, line 222: calloc, need default constructor

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -1293,7 +1293,7 @@ struct ArchReader {
         lut->model_id = get_model(arch_, LogicalModels::MODEL_NAMES);
 
         lut->num_ports = 2;
-        lut->ports = (t_port*)vtr::calloc(lut->num_ports, sizeof(t_port));
+        lut->ports = new t_port[lut->num_ports]();
```
vaughnbetz, line 1296: calloc -> need t_port default constructor

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -517,10 +517,9 @@ void XmlReadArch(const char* ArchFile,
                 /* This information still needs to be read, even if it is just
                  * thrown away.
                  */
-                t_power_arch* power_arch_fake = (t_power_arch*)vtr::calloc(1,
-                                                                           sizeof(t_power_arch));
+                t_power_arch* power_arch_fake = new t_power_arch();
```
vaughnbetz, line 520: calloc -> should make a default constructor for t_power_arch

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -533,11 +532,10 @@ void XmlReadArch(const char* ArchFile,
                 /* This information still needs to be read, even if it is just
                  * thrown away.
                  */
-                t_clock_arch* clocks_fake = (t_clock_arch*)vtr::calloc(1,
-                                                                       sizeof(t_clock_arch));
+                t_clock_arch* clocks_fake = new t_clock_arch();
```
vaughnbetz, line 535: calloc -> need default constructor of t_clock_arch

vaughnbetz (changes requested):
Thanks for starting the cleanup on this @SamuelHo10 .
I think some of these changes are unsafe in isolation though ... see my comments on calloc and default constructors.  I think we need to write default constructors for the structs in question to be able to safely use new instead of calloc.
We should also take this as an opportunity to get rid of some complexity in the code and simplify memory management by moving to std:: container classes like vectors as @soheilshahrouz suggested.  I think you should take this changes a little at a time though, running tests (or pushing to CI to have it run tests) frequently, and committing intermediate cleanups at each working stage.

We could discuss in person next week what good stages are.

vaughnbetz:
It might be good enough to use = default to explicitly ask for default constructors, as I believe in that case all built-in type data members are initialized to 0 (e.g. int, float, etc.).  That would only be a few lines of code, but you should test to be sure I'm right about the behaviour of that.  
C++ has messed around a bit with the behaviour of constructors over the years, which is why the =default seems to change the behaviour of compiler-built default constructors.

vpr/src/power/power.cpp
```diff
@@ -645,7 +645,7 @@ static void power_usage_blocks(t_power_usage* power_usage) {
  * Calculates the total power usage from the clock network
  */
 static void power_usage_clock(t_power_usage* power_usage,
-                              t_clock_arch* clock_arch) {
+                              std::vector<t_clock_network> clock_arch) {
```
soheilshahrouz, line 648: You should probably pass this by reference as it is modifed by this function.

libs/libarchfpga/src/read_fpga_interchange_arch.cpp
```diff
@@ -243,16 +243,12 @@ static bool block_port_exists(t_pb_type* pb_type, std::string port_name) {
...
-
     pp.type = E_ANNOT_PIN_TO_PIN_PACK_PATTERN;
     pp.format = E_ANNOT_PIN_TO_PIN_CONSTANT;
-    pp.prop[0] = (int)E_ANNOT_PIN_TO_PIN_PACK_PATTERN_NAME;
-    pp.value[0] = vtr::strdup(pp_name.c_str());
+    pp.pairs.push_back(std::make_pair(E_ANNOT_PIN_TO_PIN_PACK_PATTERN_NAME, pp_name));
     pp.input_pins = vtr::strdup(input.c_str());
     pp.output_pins = vtr::strdup(output.c_str());
-    pp.num_value_prop_pairs = 1;
+
```
soheilshahrouz, line 251: remove this empty line

SamuelHo10:
I've just realized that during the build process, I am getting warnings of the form:

warning: argument 1 value ‘18446744073709551615’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  737 |     mem_pb_type->modes[0].interconnect = new t_interconnect[mem_pb_type->modes[0].num_interconnect];
  
I'm pretty sure this is most likely caused by the fact that I haven't casted int to size_t, even though it still pasts all of the tests. Should I properly cast everything?

vaughnbetz (approved):
Looks good to me -- thanks!

vaughnbetz:
I think this is ready to merge.   Future clean ups: at least one num,pointer combo can be changed to a vector.
Can we get rid of pointers like the pointer to clocks?
Can we get rid of some C-strings (and hence won't need the vtr::free() calls anymore)?

PR #3041, opened 2025-05-14
Add documentation to explain which parts of VPR are parellel

AmirhosseinPoolad:
This PR adds information to the --num_workers command line option docs to explain which parts of VPR will be parallelized when using this option.


AlexandreSinger (approved):
LGTM

Issue #2978, opened 2025-04-14
Clean up confusing rr-node lookup convention for x/y location of wires.


vaughnbetz:
The code at https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/50901246919105b42a534cb58e622cada82be1bf/vpr/src/route/rr_graph2.cpp#L1225C79-L1225C90 swaps the x and y coordinates of CHANX wires when putting them in the rr_node_lookup data structure, and the find function similarly swaps them. This is baffling. 

#### Proposed Behaviour
We should just put them in the lookup with their proper x,y values and the lookup function shouldn't swap them either.

#### Current Behaviour
It works, but the convention is bizarre, making the code less readable.

soheilshahrouz:
Closing the issue as #3042 was merged.

Issue #355, opened 2018-06-07
Clean up get_rr_node_index a bit


vaughnbetz:
get_rr_node_index  has some odd code that should be cleaned up.

1. For rr_node index lookups of CHANX types, the x and y index is swapped when rr_node_indices is loaded:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph2.cpp#L1255,L1264

Then when the lookup is done in get_rr_node_index, the indices are swapped again. 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph2.cpp#L1354,L1356

We should stop doing that, and delete the comment about it.

2. We allow lookups by side for CHANX and CHANY, even though side makes no sense for them.  We should at least always go to side 0, rather than the enumerated type at sides[0] as that will break if sides[0] is not equal to 0.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/route/rr_graph2.cpp#L1338,L1341

#### Proposed Behaviour
Same as current, but less confusing code under the hood.

#### Current Behaviour
None.

#### Possible Solution
See above.

#### Context
Code cleanup.

vaughnbetz:
@kmurray

vaughnbetz:
@soheilshahrouz : I think you just fixed this -- is that right?

soheilshahrouz:
Closing the issue as #3042 was merged.

Issue #3029, opened 2025-05-09
[Parallel Router] Random strong test failures


AmirhosseinPoolad:
Strong tests for parallel routing can randomly fail. Here's the failed CI run for a PR that did not touch VTR's code in any way and only changed an unrelated workflow file:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14929980706/job/41943464845#step:8:3197

There was a run failure and some QoR failures in some of the parallel router tests. When I ran the strong tests for the branch locally it successfully worked without any run or QoR failures, and VTR master also doesn't have any strong test failures.

AmirhosseinPoolad:
@AlexandreSinger @ueqri Any ideas why this might have happened? Thank a lot!

AmirhosseinPoolad:
Update: Just re-ran the test for the PR and it was successful.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14929980706/job/41947736930

ueqri:
@AmirhosseinPoolad Hi Amir, I am investigating the test failures you mentioned. I have checked the stdout log files of the failed tests in the two action runs:

- [Log file](https://pastebin.ubuntu.com/p/MRkMqMqxtV/) of the test that failed at `strong_multiclock` when using two threads + four queues in the [action run](https://github.com/AmirhosseinPoolad/vtr-verilog-to-routing/actions/runs/14933274445/job/41954656119)
- [Log file](https://pastebin.ubuntu.com/p/gXYWXwTzsq/) of the test that failed at `strong_multiclock` when using two threads + eight queues + queue draining (deterministic for Dijkstra mode) in the [action run](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/14929980706/job/41943464845#step:8:3197)

The log files seems to be incomplete (only partial routing logs were printed and no post-routing log is there), which might be because the corresponding VPR executions (specifically the routing stage) were interrupted. I was wondering if the GitHub runner could have killed the program due to some restrictions (?) on parallel execution.

Additionally, I ran these specific tests locally on wintermute 70 times with no errors or failures occurring. I highly doubt that the issue might be related to the GitHub runner environment.

If there are any GitHub runner docs I should refer to (for the limitation/restrictions) or more context other than the test log files, please let me know and I will try to fix those tests based on that info. Also, could you please pin me the next time this happens? Hopefully that would provide more helpful context for debugging.

[...]

PR #3042, opened 2025-05-14
Clean up confusing rr-node lookup convention for x/y location of CHANX nodes

soheilshahrouz:
`RRSpatialLookup` swaps the x and y coordinates of CHANX wires when putting them in the `rr_node_lookup` data structure, and `find_node()` function swaps them as well. This PR gets rid of these swaps.

Addresses #355 and #2978


soheilshahrouz:
Radom failure due to a known issue #3029 
Re-ran the test

vaughnbetz (approved):
Thanks @soheilshahrouz .  A nice clean up -- the code looks good!

PR #3048, opened 2025-05-16
Add documentation for include sanitization

### LINKED ISSUE #624: Support include_what_you_use running on CI
#### Proposed Behaviour

[`include-what-you-use` is a tool](https://include-what-you-use.org/) which makes sure that your C and C++ files include the correct set of headers. This makes it easier to refactor code as it doesn't stop working randomly because a include is missing.

#### Current Behaviour

You can accidentally forget to add an include but the code still works because something else includes it.

#### Context

This is another code quality improvement tool.

AmirhosseinPoolad:
Adds documentation explaining how you can use IWYU or clangd diagnostics to clean up includes. This is the first step in trying to slowly fix #624. The number of include errors we currently have is very high and I don't think it's feasible to fix the entire thing in one go, so I'm proposing to start with making anyone that touches a file in a PR responsible for fixing the includes for that file (This should probably and ideally be enforced by CI). Doing this should slowly make the code cleaner and maybe in the future we can add an IWYU stage to CI.

Note that IWYU can automatically fix includes but I don't think it's a good idea to do something like this automatically.


AlexandreSinger (approved):
LGTM

PR #3054, opened 2025-05-17
[AP] Removed Old Cluster-Level AP Flow

AlexandreSinger:
Prior to the flat AP flow, a cluster-level AP flow existed in VPR which performed a SimPL-style algorithm on the clusters created during packing before performing a placement quench.

Although well-written, this flow was not shown to outperform the SA placer in VPR. It has also been becoming confusing to keep in VPR since the new flat AP flow supercedes it. It is unclear if a cluster-level AP flow will work well with the flat AP flow; however in that case the cluster-level AP flow would be made using the new AP APIs written.

Removed the old cluster-level AP flow to reduce confusion.


AlexandreSinger:
@vaughnbetz What is your opinion on this? I recall we discussed this a month or two ago. I think it is tricky to keep this code around in Master while the flat AP flow is becoming stronger. We can always find this code again by going to the previous releases of VTR (however, this AP flow was never added to the documentation as far as I could tell; it only appeared in the code).

vaughnbetz:
Looks good. One question: you're removing the cmakelists check for analytic placement, which brings in eigen. Do you have a separate such check for your new analytic placer or do you just always pull in eigen?

AlexandreSinger:
> Looks good. One question: you're removing the cmakelists check for analytic placement, which brings in eigen. Do you have a separate such check for your new analytic placer or do you just always pull in eigen?

The CMake check that I removed does not bring in Eigen. It only checks if Eigen is installed and turns on a compiler directive for the cluster-level Analytical Placement flow (which I purposefully did not use). In the CMakeLists for VPR there is another compiler directive called "EIGEN_INSTALLED" which will change the compile for certain portions of the atom-level AP flow (i.e. the solver).

The CI also tests both Eigen not being installed and Eigen being installed.

PR #3065, opened 2025-05-20
[Infra] Cleaned Up Header Files in Route Folder

AlexandreSinger:
Continued the header file cleanup effort in the route folder.

Some of these files may need to be revisited in more detail, but got some of the major header include issues.

Found that some definitions were in the wrong place, so moved them to the correct implementation file.


vpr/src/route/router_delay_profiling.h
```diff
@@ -1,7 +1,10 @@
 #ifndef ROUTER_DELAY_PROFILING_H_
 #define ROUTER_DELAY_PROFILING_H_
```
soheilshahrouz, line 2: `#pragma once`

vpr/src/route/route_utilization.h
```diff
@@ -1,8 +1,8 @@
 #ifndef VPR_ROUTE_UTIL_H
 #define VPR_ROUTE_UTIL_H
```
soheilshahrouz, line 2: `#pragma once`

soheilshahrouz (commented):
Thanks for cleaning up header file includes!

I left a few comments to gradually move the code base towards a more uniform style. These comments are not related to the code you wrote, but I though it would be a good opportunity to make these changes.

soheilshahrouz:
Thanks @AlexandreSinger

PR #3067, opened 2025-05-20
[FASM] Fixed Bug With Wire Creation

AlexandreSinger:
Found a bug within FASM's wire generation where it uses the index of the output pin to create the wire instead of the index of the input pin.

This stemmed from some confusing code which both verified that the wire was being used and returning the first valid pin. It just so happens that it checked the outputs first and returned the output pin instead.

Cleaned up the code and added more error checking to prevent issues like this in the future.

### DIFF
--- a/utils/fasm/src/fasm.cpp
+++ b/utils/fasm/src/fasm.cpp
@@ -234,6 +234,12 @@ static bool is_pin_used(const t_pb_graph_pin* pin, const t_pb_routes &top_pb_rou
  *
  * Wires in VPR are a special primitive which is a LUT which acts like a wire
  * pass-through. Only one input of this LUT should be used.
+ *
+ *  @param top_pb_route
+ *      The top pb route for the cluster that contains the wire.
+ *  @param pb_graph_node
+ *      The pb_graph_node of the wire primitive that we are getting the input
+ *      pin for.
  */
 static const t_pb_graph_pin* get_wire_input_pin(const t_pb_routes &top_pb_route, const t_pb_graph_node* pb_graph_node) {
     const t_pb_graph_pin* wire_input_pin = nullptr;
@@ -254,6 +260,11 @@ static const t_pb_graph_pin* get_wire_input_pin(const t_pb_routes &top_pb_route,
  * @brief Returns true if the given wire is used.
  *
  * A wire is used if it has a used output pin.
+ *
+ *  @param top_pb_route
+ *      The top pb route for the cluster that contains the wire.
+ *  @param pb_graph_node
+ *      The pb_graph_node of the wire primitive that we are checking is used.
  */
 static bool is_wire_used(const t_pb_routes &top_pb_route, const t_pb_graph_node* pb_graph_node) {
     // A wire is used if it has a used output pin.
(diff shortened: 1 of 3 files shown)


AlexandreSinger:
@vaughnbetz , when upgrading to the most recent version of VTR, our testing suite at ZeroASIC found this issue. We simulated the FASM generated by genfasm and found that the lookup table for the wire implementation was incorrect. Based on this issue, we believe that the first inputs to wires were always being used; however, due to recent changes to the packer, we may have allowed different inputs of "wire" mode LUTs to be used which exposed this issue.

This PR fixes the issue and adds some self-checking code to try and prevent headache in the future. Please review when you have a moment.

utils/fasm/src/fasm.cpp
```diff
@@ -217,26 +217,62 @@ std::string FasmWriterVisitor::build_clb_prefix(const t_pb *pb, const t_pb_graph
...
+ * A wire is used if it has a used output pin.
+ */
+static bool is_wire_used(const t_pb_routes &top_pb_route, const t_pb_graph_node* pb_graph_node) {
+    // A wire is used if it has a used output pin.
+    const t_pb_graph_pin* wire_output_pin = nullptr;
+    for(int port_index = 0; port_index < pb_graph_node->num_output_ports; ++port_index) {
+        for(int pin_index = 0; pin_index < pb_graph_node->num_output_pins[port_index]; ++pin_index) {
+            const t_pb_graph_pin* pin = &pb_graph_node->output_pins[port_index][pin_index];
+            if (is_pin_used(pin, top_pb_route)) {
+                VTR_ASSERT_MSG(wire_output_pin == nullptr,
```
amin1377, line 265: I think I’m not fully understanding this part of the code. As far as I know, t_pb_routes contains the intra-cluster routing for a cluster, which can include multiple atom nets. Here, you're iterating over all output pins inside of the cluster and generating a warning if more than one output pin is used.

Why do we need to warn the user about that?

utils/fasm/src/fasm.cpp
```diff
@@ -217,26 +217,62 @@ std::string FasmWriterVisitor::build_clb_prefix(const t_pb *pb, const t_pb_graph
...
-    for(int port_index = 0; port_index < pb_graph_node->num_input_ports; ++port_index) {
-        for(int pin_index = 0; pin_index < pb_graph_node->num_input_pins[port_index]; ++pin_index) {
-            pin = &pb_graph_node->input_pins[port_index][pin_index];
-            if (top_pb_route.count(pin->pin_count_in_cluster) > 0 && top_pb_route[pin->pin_count_in_cluster].atom_net_id != AtomNetId::INVALID()) {
-                return pin;
+    return wire_input_pin;
+}
+
+/**
+ * @brief Returns true if the given wire is used.
```
amin1377, line 254: The parameters passed to this function include the data structure containing all the pins within a cluster and a pb_graph_node. What does the "given wire" refer to in this context?

amin1377 (changes requested):
Thanks, Alex! I know you didn’t write the code, but I thought you might have some insight into my questions. For both get_wire_input_pin and is_wire_used, based on the input parameters, I’m not quite sure what the term "wire" refers to. If you could help clarify that, I’d really appreciate it.

utils/fasm/src/fasm.cpp
AlexandreSinger: This particular piece of the code is not iterating over all output pins of the cluster; it is looping over all output pins of the primitive which in this case must be a "wire". I do not think this is a warning since if a wire has 2 output pins something is screwed up internally to VTR.

utils/fasm/src/fasm.cpp
AlexandreSinger: Good point, let me update this!

utils/fasm/src/fasm.cpp
AlexandreSinger: I added more comments to make this more clear; thanks for pointing out this confusion!

amin1377: You’re right, if a net has two output pins, something unusual is definitely going on. I think what I’m trying to understand is this: suppose I have a pb_graph_node with 5 output pins, and there’s a net connected to each of them. In that case, this code would trigger the assertion.

My guess is that this function is making certain assumptions about the pb_graph_node—for example, that only one of its output pins should be used, but those assumptions aren’t explicitly stated.

amin1377 (approved):

utils/fasm/src/fasm.cpp
AlexandreSinger: For a wire specifically, I think the assumption is there (but not explicitly stated). The documentation for the "wire" mode needs to be better documented in my opinion; but that is independent of this PR.

Issue #3064, opened 2025-05-20
make vpr viewer work properly with flat routing enabled


w0lek:
#### Current Behaviour
if we run vpr viewer (--disp on) with flat routing (--flat_routing on) 

full command is here
```
./vpr test_post_verilog_arch.xml unconnected.eblif --device auto --timing_analysis on --constant_net_method route --clock_modeling ideal --exit_before_pack off --circuit_format eblif --absorb_buffer_luts off --route_chan_width 180 --flat_routing on --gen_post_synthesis_netlist on --timing_report_npaths 100 --timing_report_detail netlist --allow_dangling_combinational_nodes on --analysis --disp on
```

attempt to use SearchBar and Search NetID or Net Name, will draw the route path, which doesn't correspond data in placement file

for instance, if we rely on

```
Net 1 (reset)

Node:	12214	SOURCE (2,0,0)  Pad: 709  Switch: 0
Node:	12263	  OPIN (2,0,0)  Pad: 1194  Switch: 13
Node:	3210	  OPIN (2,0,0)  Pad: 154  Switch: 2
Node:	8426	 CHANX (2,0,0)  Track: 22  Switch: 3
Node:	11154	 CHANY (2,1,0)  Track: 158  Switch: 1
Node:	3319	  IPIN (2,1,0)  Pin: 24   clb.I2[4] Switch: 15
Node:	12586	  IPIN (2,1,0)  Pin: 4156   lut5inter.in[1] Switch: 4
Node:	12799	  IPIN (2,1,0)  Pin: 4312   lut5.in[1] Switch: 4
Node:	13140	  IPIN (2,1,0)  Pin: 4318   lut.in[1] Switch: 0
Node:	12272	  SINK (2,1,0)  Class: 1848  Switch: -1 Net_pin_index: 1
```
and try to search Net ID = 1, we got routing which uses other Nodes, and not one which are listed in placement file.

#### Proposed Behaviour

The VPR viewer in flat routing should show Net ID or NetName corresponds to data in placement file (.route)

PR #3070, opened 2025-05-21
Vpr viewer and flat routing on

w0lek:
#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3064

manual test:

- run vpr viewer with following command
./vpr test_post_verilog_arch.xml unconnected.eblif --device auto --timing_analysis on --constant_net_method route --clock_modeling ideal --exit_before_pack off --circuit_format eblif --absorb_buffer_luts off --route_chan_width 180 --flat_routing on --gen_post_synthesis_netlist on --timing_report_npaths 100 --timing_report_detail netlist --allow_dangling_combinational_nodes on --analysis --disp on
- select different Node Id or Node name based on unconnected.route in search bar, observe visual.
check that visual nodes representation corresponds to nodes listed in unconnected.route file.

#### Types of changes

#### Checklist:


vpr/src/draw/search_bar.h
```diff
@@ -17,6 +17,9 @@
 
 #include "ezgl/application.hpp"
 
+bool is_net_unrouted(AtomNetId atomic_net_id);
+bool is_net_fully_absorbed(AtomNetId atomic_net_id);
```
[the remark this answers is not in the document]
amin1377: The functionality of is_net_fully_absorbed isn't specific to drawing, so I think it would be better to move it to the route_utils file rather than defining it here.

vpr/src/draw/search_bar.cpp
```diff
@@ -41,6 +42,34 @@
...
+    return !route_ctx.route_trees[atomic_net_id].has_value();
+}
+
+bool is_net_fully_absorbed(AtomNetId atomic_net_id) {
+    const RRGraphView& rr_graph = g_vpr_ctx.device().rr_graph;
+    RoutingContext& route_ctx = g_vpr_ctx.mutable_routing();
+
+    bool is_absorbed = true;
+
+    t_trace* head = TracebackCompat::traceback_from_route_tree(route_ctx.route_trees[atomic_net_id].value());
```
amin1377, line 56: The more common approach in VTR for iterating over the routing path of a net is to use the route tree (in routing context), which can also be used to check whether a net is routed. I’d suggest using that instead of constructing the net trace manually.

amin1377: Also, add a comment and explain how you determine whether a net is completely absorbed.

amin1377 (changes requested):
Thanks, Alex, for fixing the issues with the GUI when flat router is used. In the PR description, it would be helpful if you could explain the specific case you tested that was previously incorrect and is now working properly. That would be useful for future reference.

vpr/src/draw/search_bar.h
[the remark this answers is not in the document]
w0lek: > I did a bit of digging through the codebase and found that the routing context includes a data structure called net_status, which has a method called is_routed. I think it would be better to use that instead of writing a custom function to determine whether a net is routed.

I just tested this approach, and it's not robust. It seems the net_status structure is filled only on --route stage, and never is filled with, let's say if we run vpr viewer in --analysis only stage. Due to this, if we run vpr viewer in analysis mode, the each net id will be reported as not routed if we attempt to use net_status.
With respect of this information, could i use

'''
route_ctx.route_trees[net_id].has_value()
''' 
instead? or we should think how to fix filling net_stats structure for all possible cases? About fixing net_status filling case, i think it more looks like separate issue.

vpr/src/draw/search_bar.cpp
w0lek: > The more common approach in VTR for iterating over the routing path of a net is to use the route tree (in routing context), which can also be used to check whether a net is routed. I’d suggest using that instead of constructing the net trace manually.

resolved here 80675e37397bbb6ad2ef410ede68d3d7f3b2bb0b

w0lek: > Also, add a comment and explain how you determine whether a net is completely absorbed.
doxgen doc is added 2e9a2d62c1fdf578fef9fcfbf48ee92a352c7d57

w0lek: resolved https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/80675e37397bbb6ad2ef410ede68d3d7f3b2bb0b and https://github.com/verilog-to-routing/vtr-verilog-to-routing/commit/2e9a2d62c1fdf578fef9fcfbf48ee92a352c7d57

vpr/src/draw/search_bar.h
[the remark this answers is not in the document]
w0lek: > I think adding Doxygen comments for these two functions would be better.

added 2e9a2d62c1fdf578fef9fcfbf48ee92a352c7d57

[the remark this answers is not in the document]
amin1377: > It seems the net_status structure is filled only on --route stage

If you can file an issue on it, that would be good. Thanks!

amin1377:
Looks good. Thanks!

PR #3071, opened 2025-05-21
[Infra] Cleaned Up Includes in Draw Dir

AlexandreSinger:
Cleaned up the includes in the draw files. These ones were much messier than I originally thought. Many of the header files in the draw directory included way more than they needed which was causing false dependencies anywhere in VPR which included any draw files.


soheilshahrouz (approved):
Thanks @AlexandreSinger

Issue #3074, opened 2025-05-22
win32 and VTR enums collision


w0lek:
#### Current Behaviour
whenever gateio.h is included and if such transaction unit includes VTR enum headers, compile errors with enum names collisions arise

#### Possible Solution
include sockpp headers directly in gateio transaction unit, this will avoid win32 and VTR enums collision.

PR #3075, opened 2025-05-22
Hotfix win32 sockpp and vtr enum collisions

w0lek:
include sockpp headers directly in gateio.cpp transaction unit, this will avoid win32 and VTR enums collision.

#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3074

#### Motivation and Context
allow use app on windows

it was tested as a part of Aurora windows package


AlexandreSinger (approved):
LGTM

PR #3077, opened 2025-05-22
[Infra] The Great VPR Pragma PR

AlexandreSinger:
VPR is moving to a style that uses "#pragma once" instead of header gaurds. These are less error prone and may be slightly more performant.

Converted all of the header gaurds in VPR into pragma once's.

Also moved all pragma onces to the top of all header files to maintain a consistent style. It is a good idea to have them as the very first line in all header files.

While going through all header files, cleaned up any extra header includes which were including things they did not need.

### DIFF
--- a/vpr/src/analytical_place/analytical_placement_flow.h
+++ b/vpr/src/analytical_place/analytical_placement_flow.h
@@ -1,11 +1,10 @@
+#pragma once
 /**
  * @file
  * @author  Alex Singer
  * @date    September 2024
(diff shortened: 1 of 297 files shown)


AlexandreSinger:
@soheilshahrouz Ah! I see, I just looked at the code and what you are saying is that it already used double quotation marks and I added angle brackets justs on the ones I added. I understand why thats not good. I turned them into double quotation marks so this does not block the PR; however, we should eventually revisit this topic. Not urgent at all though.

PR #3078, opened 2025-05-22
[CI] Consolidated Build Variation Tests

AlexandreSinger:
Different build variations of VTR were being run on different CI runners which was wasteful.

Consolidated these build variations into a single job which will run on a single runner.


AlexandreSinger:
Marking as WIP since the build variations test is taking a bit longer than I would like. Although it may save time overall, need to investigate how to mitigate the long run time.

vaughnbetz:
Looks good to me.

PR #3079, opened 2025-05-22
[CI] Removed Redundant Warning Test

AlexandreSinger:
Prior to the updates to the CI to make all regression tests warning clean, there was another warning test which was not as comprehensive as the tests we have now.

Since this test was superceded, removed it from the CI. The CMAKE param that enabled it was also used and replaced with a more comprehensive CMAKE option.

### DIFF
--- a/vpr/CMakeLists.txt
+++ b/vpr/CMakeLists.txt
@@ -227,24 +227,6 @@ else()
     message(ERROR "Unsupported VPR_PGO_CONFIG '${VPR_PGO_CONFIG}'")
 endif()
 
-if (VTR_ENABLE_STRICT_COMPILE)
-    message(STATUS "VPR: building with strict flags")
-
-    set(VPR_STRICT_COMPILE_FLAGS_TO_CHECK
-        #GCC-like
-        "-Werror"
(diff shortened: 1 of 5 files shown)


vpr/CMakeLists.txt
```diff
@@ -227,24 +227,6 @@ else()
...
 endif()
 
-if (VTR_ENABLE_STRICT_COMPILE)
-    message(STATUS "VPR: building with strict flags")
-
-    set(VPR_STRICT_COMPILE_FLAGS_TO_CHECK
-        #GCC-like
-        "-Werror"
-        # due to the pointer hackery in timing_driven_route_structs and BinaryHeap.heap_
-        "-Wno-error=free-nonheap-object"
```
AmirhosseinPoolad, line 237: I don't see this flag set anywhere else. I assume it's not needed anymore and was here just because of legacy reasons.

AmirhosseinPoolad (commented):
LGTM, just had a small question.

vpr/CMakeLists.txt
AlexandreSinger: Correct. The fact that we are no longer getting this warning anymore indicates that this was fixed at some point.

PR #3081, opened 2025-05-24
[Infra] Cleaned Up Includes in Non-External Libs

AlexandreSinger:
Updated the header files in the non-external libraries of VTR such that they use pragma once instead of ifdefs and removed false include files.


libs/libvtrcapnproto/serdes_utils.cpp
```diff
@@ -1,5 +1,6 @@
 #include "serdes_utils.h"
 
+#include <capnp/serialize.h>
```
soheilshahrouz, line 3: repalce `<>` with `""`

libs/libvtrcapnproto/serdes_utils.h
```diff
@@ -1,9 +1,9 @@
-#ifndef SERDES_UTILS_H_
-#define SERDES_UTILS_H_
+#pragma once
 
+#include <capnp/message.h>
```
soheilshahrouz, line 3: repalce `<>` with `""`

libs/libvtrutil/src/vtr_geometry.tpp
```diff
@@ -1,3 +1,5 @@
+#include <vtr_assert.h>
```
soheilshahrouz, line 1: repalce `<>` with `""`

PR #3095, opened 2025-05-29
[AP][Legalizer] Added Ability to Generate a Mass Report

AlexandreSinger:
While working on the mass abstraction in the partial legalizer of the globlal placer, found that I needed a lot more information on the device to be able to debug the mass calculations.

Added a command-line option which will generate a mass report if requested. This mass report contains useful information on the device, the netlist, and the mass / capacities computed in the mass calculator.

Example reports:
- [arm_core_ap_mass.txt](https://github.com/user-attachments/files/20512153/arm_core_ap_mass.txt)
- [stereo_vision_ap_mass.txt](https://github.com/user-attachments/files/20512154/stereo_vision_ap_mass.txt)

Notably, I added the ability to print the complex block graph; which is extremely useful for debugging the mass calculator:

image


AlexandreSinger:
@amin1377 I am not sure how much use you will get out of this report, but I think it wouldn't hurt! Can you please review when you have a moment?

doc/src/vpr/command_line_usage.rst
```diff
@@ -1336,6 +1336,15 @@ Analytical Placement is generally split into three stages:
 
     **Default:** ``1``
 
+.. option:: --ap_generate_mass_report {on | off}
```
amin1377, line 1339: I looked at the report and noticed that there is no AP-specific section in it. So, I think it might be better to change the name of the option to avoid implying that it's AP-specific, and instead treat it as one of the general options for generating a report.

vpr/src/base/ShowSetup.cpp
```diff
@@ -657,6 +657,7 @@ static void ShowAnalyticalPlacerOpts(const t_ap_opts& APOpts) {
 
     VTR_LOG("AnalyticalPlacerOpts.ap_timing_tradeoff: %f\n", APOpts.ap_timing_tradeoff);
     VTR_LOG("AnalyticalPlacerOpts.log_verbosity: %d\n", APOpts.log_verbosity);
+    VTR_LOG("AnalyticalPlacerOpts.generate_mass_report: %s\n", APOpts.generate_mass_report ? "true" : "false");
```
amin1377, line 660: I'm not sure what the convention is for displaying parameter values in the vpr.out report. I think it would be better to avoid printing all parameter values and instead only include those that are necessary for analyzing the results. If you agree, I suggest we omit the value of the generate_mass_report parameter (or any similar parameter that simply controls whether a report should be generated).

vpr/src/analytical_place/ap_mass_report.cpp
```diff
@@ -0,0 +1,542 @@
...
+#include "logic_types.h"
+#include "physical_types.h"
+#include "physical_types_util.h"
+#include "primitive_vector.h"
+#include "vpr_context.h"
+#include "vpr_error.h"
+#include "vtr_time.h"
+#include "vtr_vector.h"
+
+namespace {
```
amin1377, line 27: Just out of curiosity: I think the main reason for using an anonymous namespace here is because of the structs. But if the structs weren't present, is there any real difference between using static functions and defining functions in an anonymous namespace?

vpr/src/analytical_place/ap_mass_report.cpp
```diff
@@ -0,0 +1,542 @@
...
+ * @brief Print the given tree node.
+ *
+ *  @param node
+ *      The node to print.
+ *  @param os
+ *      The output file stream to print the node to.
+ *  @param prefix
+ *      The prefix that all children of this node will print before their names.
+ */
+void print_tree_node(const PrintingTreeNode& node,
```
amin1377, line 58: I'm not sure whether it's a common convention or not, but in other parts of VTR, the "recur" suffix is added to the names of recursive functions.

amin1377 (changes requested):
Thanks, Alex! As we discussed before, I really liked the idea of generating this report. I’ve added a bunch of small comments.

The most important one is that generating this report; at least the parts based solely on the architecture file shouldn’t be made AP-specific. For the other parts (like generating the mass report), it would probably be more challenging, since that would require adding the capability to calculate mass for the placed atom netlist, which is likely beyond the scope of this PR.

doc/src/vpr/command_line_usage.rst
AlexandreSinger: The mass calculation is AP specific, the report cannot be generated outside of the AP flow. The drawing of the complex graph is useful for understanding the mass report, hence why it is part of this report.

I do not want to gate this PR on this, but in a future PR I can move this to an external report which reports on the structure of the architecture file. I think that is beyond the scope of this PR and would include more features. What do you think?

amin1377: I agree. Let's do it in another PR.

amin1377 (approved):

PR #3097, opened 2025-05-30
Improve comments in Makefile

AmirhosseinPoolad:
This PR better documents the possible options users have when using the makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-#This is a simple wrapper which hides cmake (for convenience, and from non-expert end users).
+# This is a simple wrapper which hides cmake (for convenience, and from non-expert end users).
 #
 # It supports the targets:
 #   'make'           - builds everything (all libaries/executables)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,13 +15,15 @@
...
-# Possible values:
-#    release_pgo	#Perform a 2-stage build with profile-guided compiler optimization
-#    release		#Build with compiler optimization
-#    debug			#Build with debug info and no compiler optimization
-#    strict			#Build VPR with warnings treated as errors
-BUILD_TYPE ?= release
+# Build type
+# Possible values (not case sensitive):
+#    Release		#Build with compiler optimization
+#    RelWithDebInfo #Build with debug info and compiler optimizations (Default)
```
AlexandreSinger: Update this comment. "release" is the default, not "RelWithDebInfo".

Makefile
```diff
@@ -15,13 +15,15 @@
...
+#    Debug			#Build with debug info and no compiler optimization
+# Possible suffixes:
+#    _pgo           #Perform a 2-stage build with profile-guided compiler optimization
+#    _strict        #Build VPR with warnings treated as errors
```
AlexandreSinger: Can you align the "#" comments so they all start on the same column. I think it would make this more readible.

Makefile
```diff
@@ -15,13 +15,15 @@
...
+BUILD_TYPE ?= Release
```
AlexandreSinger: This is personal preference of mine, but I much prefer "release" over "Release" and I think its more standard. I recommend making "Release" and "Debug" lower case and fix it in the comment above.

AlexandreSinger: I am cool with saying "this is not case sensitive", but I think if we make the default uppercase it may get a bit confusing.

AlexandreSinger (commented):
Hi @AmirhosseinPoolad , thanks for this update! Some comments below.

Makefile
AmirhosseinPoolad: The issue with having it be "release" and "debug" is that then you would need to say "relwithdebinfo" which just looks terrible. I also don't see how it's confusing.

Makefile
AmirhosseinPoolad: This looked completely okay in my editor so I was very confused at first:
image
Then I saw that the old comments used tabs and my comments used spaces. Fixed by changing everything to spaces.

Makefile
AlexandreSinger: I honestly do not mind "relwithdebinfo" being that way, but thats personal preference. I do not really care that much about this, I just find "release" to be more common. TBH you could make RelWithDebInfo capitalized like that and leave the others alone.

AlexandreSinger (approved):
LGTM!

PR #3098, opened 2025-05-30
[Infra] Fixed False Forward Declarations

AlexandreSinger:
The Clang builds were warning that there were several forward declarations of structs which were supposed to be classes and vice-versa.

This is not necessarily a problem since in C++ classes and structs end up being basically the same from the compiler's perspective, but its still incorrect. Fixed the cases I could see in the Clang builds.


AmirhosseinPoolad (approved):
LGTM

PR #3101, opened 2025-06-03
[Infra] Converted Pin to Pin Annotations into Vector

AlexandreSinger:
Pin to Pin annotations were stored as C-style arrays which creates confusing pointers around VTR.

Converted to a standard vector.


libs/libarchfpga/src/physical_types.h
```diff
@@ -1160,6 +1159,43 @@ struct t_mode {
...
+struct t_pin_to_pin_annotation {
+
+    std::vector<std::pair<int, std::string>> annotation_entries;
+
+    enum e_pin_to_pin_annotation_type type;
+    enum e_pin_to_pin_annotation_format format;
+
+    char* input_pins;
+    char* output_pins;
+    char* clock;
```
AmirhosseinPoolad, line 1183: If you have the energy, consider converting these to std::string or std::string_view. Otherwise feel free to open an issue to document its existence.

libs/libarchfpga/src/physical_types.h
```diff
@@ -1201,9 +1235,6 @@ struct t_interconnect {
         type = (e_interconnect)0;
         name = nullptr;
         input_string = nullptr;
-        output_string = nullptr;
```
AmirhosseinPoolad, line 1204: output_string still exists in the struct, why did you remove it from the constructor?

vpr/src/pack/pb_type_graph_annotations.cpp
```diff
@@ -70,33 +66,32 @@ void load_pb_graph_pin_to_pin_annotations(t_pb_graph_node* pb_graph_node) {
...
-                        VTR_ASSERT(annotations[k].annotation_entries.size() == 1);
-                        load_pack_pattern_annotations(annotations[k].line_num, pb_graph_node, i,
-                                                      annotations[k].input_pins,
-                                                      annotations[k].output_pins,
-                                                      annotations[k].annotation_entries[0].second);
+                    } else if (annot.type == E_ANNOT_PIN_TO_PIN_PACK_PATTERN) {
+                        VTR_ASSERT(annot.annotation_entries.size() == 1);
+                        load_pack_pattern_annotations(annot.line_num, pb_graph_node, i,
+                                                      annot.input_pins,
+                                                      annot.output_pins,
```
AmirhosseinPoolad: I think "annot" is too terse, "annotation" isn't very long anyway and should be fine to use.

AmirhosseinPoolad (commented):
Thanks for the PR @AlexandreSinger. Very welcome changes overall. Had a couple of thoughts.

libs/libarchfpga/src/physical_types.h
AlexandreSinger: Oh! Great catch! Thanks!

vpr/src/pack/pb_type_graph_annotations.cpp
AlexandreSinger: Yep. I had the same struggle. I changed to annot since other places use it, but I agree "annotation" is better!

libs/libarchfpga/src/physical_types.h
AlexandreSinger: I agree with this; however, I think this is disjoint from this PR. I will raise an issue.

AlexandreSinger: Issue has been raised: #3102

PR #3104, opened 2025-06-04
[CI] Added Quick Titanium S10 Tests

AlexandreSinger:
The titanium benchmarks were not being tested by the CI. Added the Titanium benchmarks which could be run in under around 2 hours to NightlyTest7.

5 circuits in this benchmark set currently fail through VTR. The failures are mainly in the initial placer, which is struggling to create an initial placement when logical blocks can be placed into different physical block types which are constrained resources.


vaughnbetz:
Thanks @AlexandreSinger ! Is this with an auto sized device? If so, it is strange. @amin1377 : I think you landed some initial placement changes a while ago. Do blocks eventually do an exhaustive search?

AlexandreSinger:
> Thanks @AlexandreSinger ! Is this with an auto sized device? If so, it is strange. @amin1377 : I think you landed some initial placement changes a while ago. Do blocks eventually do an exhaustive search?

Hi Vaughn, yes, this is with an auto-sized device. I raised an issue on this topic which provides a bit more information: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3105

I think a "simple" solution would be to place the logical blocks which can only be placed in one and only one physical tile first (since they are more constrained), and then placing the more free logical blocks second. We could add a third iteration of the initial placer which does this, which would prevent this update from affecting other benchmarks. Initial Placement is quite quick, so running 3 iterations of it is not crazy in my opinion.

vaughnbetz:
@AlexandreSinger : your proposed fix makes sense to me. Here's a tweak that could also help: detect what kind of block failed to place and move it to the front in later iterations (we already detect blocks that failed, but we could generalize that to put blocks that failed to place first in the second iteration, and block types that failed right after the blocks that failed).  We should also check that those terms swamp any "connected to placed blocks" terms so they dominate in later iterations.  Anybody want to try coding a fix?

PR #3106, opened 2025-06-04
[AP] Optimized Primitive Vector Class

AlexandreSinger:
The primitive vector class was assumed to be quite sparse, and as such used an unordered map as its internal data structure.

Found through experimentation that most of the time in the partial legalizer was being spent in the operations of the primitive vector class. Also, while improving the mass abstraction, I found a need to separate logical models from the dimensions of the primitive vector to allow multiple models to point to the same dimension in the primitive vector.

This PR kills two birds with one stone by turning the unordered map into a VTR vector map and creating a new PrimitiveVectorDim which can allow the models to be separate from the dimensions.

Future PRs will make use of this feature to improve the mass abstraction.

Running Titan now, but early results show a 5x runtime improvement to Partial Legalization with no change in max memory usage.


AlexandreSinger:
Results on Titan (timing analysis off, no fixed IOs):

Metric | Normalized to AP Master
-- | --
Total Run Time | 0.913285647
AP Run Time | 0.8823572013
GP Run Time | 0.6868591075
Post-Route WL | 1.005144097
Post-GP WL | 0.9904709347

All values are normalized to the AP flow on Master right now.

Overall, this upgrade has improved the GP run time by over 30%, which resulted in an overall run time improvement of 8.7%. The wirelength did not change much (basically just noise) and the memory overhead is negligable.

Looking at directrf (the slowest circuit), the solver time is around 20 seconds per iteration and the legalizer is around 5 seconds. The solver should not be this slow relative to the legalizer. This may indicate that there are more optimizations available in the solver.

@vaughnbetz FYI

@amin1377 You are familiar with my Primitive Vector data structure. Could you please review this if you have a moment. Sorry for the longer PR here!

AlexandreSinger:
> Nice! The solver is parallel so we should get a speed up just by running with that enabled.

I agree! Before I do that, I do want to look at some profiling results to see if I can make it even faster serially.

vpr/src/analytical_place/primitive_dim_manager.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+/**
+ * @file
+ * @author  Alex Singer
+ * @date    June 2025
```
vaughnbetz, line 5: I think in this file, or somewhere if you think there is a better location, you should give an overview of how the various mappings fit together.  From reading the code I don't have a great sense of what the overall  look up / tracking strategy is, so I think some high level commenting on how it all works would be good.

AlexandreSinger: Thats a good point! I will do this. I also think that some comments should be made in the Primitive vector header file and the mass calculator file to give a more global picture and to reference each other.

I am not the biggest fan of how I separated things; but in order to get good performance without making crazy global variables, I think this is the best I can do.

AlexandreSinger: Updated the documentation:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3106/commits/7b73a478f998271323a644258570953340a3b7a4

vpr/src/analytical_place/primitive_vector.h
```diff
@@ -35,58 +45,50 @@ class PrimitiveVector {
...
-    inline float get_dim_val(size_t dim) const {
-        const auto it = data_.find(dim);
-        // If there is no data in the dim, return 0. By default the vector is
-        // empty.
-        if (it == data_.end())
-            return 0.f;
-        // If there is data at this dimension, return it.
-        return it->second;
+    inline float get_dim_val(PrimitiveVectorDim dim) const {
+        if ((size_t)dim >= data_.size())
```
amin1377, line 81: In other data structures in VPR, similar methods (like get, add, subtract) typically have two versions: one that raises an error if the dimension doesn't exist, and another, clearly implied by its name, that adds the dimension if it doesn't exist. It’s your call, but I personally prefer this approach, as it’s safer and helps catch bugs earlier.

vpr/src/analytical_place/flat_placement_mass_calculator.h
```diff
@@ -4,10 +4,22 @@
  * @author  Alex Singer
  * @date    February 2024
  * @brief   Mass calculation for AP blocks and logical/physical block/tile types
+ *
+ * This class is used to abstract away the concept of mass in the partial legalizer.
```
amin1377, line 8: I think this comment could be expanded to clarify what exactly is meant by "mass." I suggest adding an example, for instance, explaining that a molecule might consist of one LUT and one FF, and could be mapped to a cluster containing n LUTs and FFs. That would help solidify the meaning of "mass" in this context.

vpr/src/analytical_place/primitive_vector.h
AlexandreSinger: That would be a good optimization to add in the future; however I am avoiding adding an interface like that for now. The flow is too new and I would prefer to always check for now! I will keep this in mind for future upgrades to this interface.

vpr/src/analytical_place/flat_placement_mass_calculator.h
AlexandreSinger: Good point! I have added the following:

image

AlexandreSinger:
@amin1377 Thank you for the feedback! Let me know if you have any further comments!

amin1377 (approved):

PR #3113, opened 2025-06-05
[VPR][Utils] Add helper function to get RRNodeId from AtomPinId

amin1377:
#### Description
A helper function is added to return RRNodeId from the given AtomPinId.

#### Motivation and Context
@SamuelHo10 requested this helper function to implement flat-router graphics.

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -709,6 +709,7 @@ struct t_physical_tile_type {
 
     std::unordered_map<int, std::unordered_map<t_logical_block_type_ptr, t_pb_graph_pin*>> on_tile_pin_num_to_pb_pin; // [root_pin_physical_num][logical_block] -> t_pb_graph_pin*
     std::unordered_map<int, t_pb_graph_pin*> pin_num_to_pb_pin;                                                       // [intra_tile_pin_physical_num] -> t_pb_graph_pin
+    std::unordered_map<t_pb_graph_pin*, int> pb_pin_to_pin_num;                                                       // [t_pb_graph_pin*] -> intra_tile_pin_physical_num
 
     std::vector<t_fc_specification> fc_specs;
 
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -709,7 +709,7 @@ struct t_physical_tile_type {
 
     std::unordered_map<int, std::unordered_map<t_logical_block_type_ptr, t_pb_graph_pin*>> on_tile_pin_num_to_pb_pin; // [root_pin_physical_num][logical_block] -> t_pb_graph_pin*
     std::unordered_map<int, t_pb_graph_pin*> pin_num_to_pb_pin;                                                       // [intra_tile_pin_physical_num] -> t_pb_graph_pin
-    std::unordered_map<t_pb_graph_pin*, int> pb_pin_to_pin_num;                                                       // [t_pb_graph_pin*] -> intra_tile_pin_physical_num
+    std::unordered_map<const t_pb_graph_pin*, int> pb_pin_to_pin_num;                                                       // [t_pb_graph_pin*] -> intra_tile_pin_physical_num
 
     std::vector<t_fc_specification> fc_specs;
 
--- a/vpr/src/base/SetupVPR.cpp
+++ b/vpr/src/base/SetupVPR.cpp
@@ -901,7 +901,9 @@ static void add_logical_pin_to_physical_tile(int physical_pin_offset,
     for (auto logical_pin_pair : logical_block_ptr->pin_logical_num_to_pb_pin_mapping) {
         auto pin_logical_num = logical_pin_pair.first;
         auto pb_pin = logical_pin_pair.second;
-        physical_type->pin_num_to_pb_pin.insert(std::make_pair(pin_logical_num + physical_pin_offset, pb_pin));
+        int pin_physical_num = pin_logical_num + physical_pin_offset;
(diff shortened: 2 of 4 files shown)


vpr/src/util/vpr_utils.h
```diff
@@ -248,6 +248,14 @@ RRNodeId get_pin_rr_node_id(const RRSpatialLookup& rr_spatial_lookup,
                             const int root_j,
                             int pin_physical_num);
 
+/**
+ * @brief Returns the RR node ID for the given atom pin ID.
+ * **Warning**: This function should be called only if flat-router is enabled,
+ * since, otherwise, the routing resources inside clusters are not added to the RR graph.
+ * @param atom_pin_id The atom pin ID.
+ */
+RRNodeId get_atom_pin_rr_node_id(const AtomPinId& atom_pin_id);
```
soheilshahrouz: AtomPinId is an integer. Don't pass it by reference

vpr/src/util/vpr_utils.cpp
```diff
@@ -1720,6 +1720,41 @@ RRNodeId get_class_rr_node_id(const RRSpatialLookup& rr_spatial_lookup,
...
+    ClusterBlockId clb_blk_id = atom_lookup.atom_clb(atom_blk_id);
+
+    t_pl_loc clb_blk_loc = place_ctx.block_locs()[clb_blk_id].loc;
+
+    t_physical_tile_type_ptr physical_tile = device_ctx.grid.get_physical_type({clb_blk_loc.x, clb_blk_loc.y, clb_blk_loc.layer});
+
+    const t_pb_graph_pin* atom_pb_pin = atom_lookup.atom_pin_pb_graph_pin(atom_pin_id);
+    int pin_physical_num = physical_tile->pb_pin_to_pin_num.at(atom_pb_pin);
+
+    auto rr_node_id = get_pin_rr_node_id(device_ctx.rr_graph.node_lookup(),
```
soheilshahrouz: auto --> explicit type name

vpr/src/base/SetupVPR.cpp
```diff
@@ -901,7 +901,9 @@ static void add_logical_pin_to_physical_tile(int physical_pin_offset,
     for (auto logical_pin_pair : logical_block_ptr->pin_logical_num_to_pb_pin_mapping) {
         auto pin_logical_num = logical_pin_pair.first;
         auto pb_pin = logical_pin_pair.second;
-        physical_type->pin_num_to_pb_pin.insert(std::make_pair(pin_logical_num + physical_pin_offset, pb_pin));
+        int pin_physical_num = pin_logical_num + physical_pin_offset;
+        physical_type->pin_num_to_pb_pin.insert(std::make_pair(pin_physical_num, pb_pin));
+        physical_type->pb_pin_to_pin_num.insert(std::make_pair(pb_pin, pin_physical_num));
```
soheilshahrouz: `physical_type->pb_pin_to_pin_num.insert({pb_pin, pin_physical_num});` is more concise

vpr/src/base/SetupVPR.cpp
```diff
@@ -901,7 +901,9 @@ static void add_logical_pin_to_physical_tile(int physical_pin_offset,
     for (auto logical_pin_pair : logical_block_ptr->pin_logical_num_to_pb_pin_mapping) {
         auto pin_logical_num = logical_pin_pair.first;
         auto pb_pin = logical_pin_pair.second;
-        physical_type->pin_num_to_pb_pin.insert(std::make_pair(pin_logical_num + physical_pin_offset, pb_pin));
+        int pin_physical_num = pin_logical_num + physical_pin_offset;
```
soheilshahrouz, line 904: Can you please replace these auto keywords with explicit type names.

libs/libarchfpga/src/physical_types.h
```diff
@@ -709,6 +709,7 @@ struct t_physical_tile_type {
 
     std::unordered_map<int, std::unordered_map<t_logical_block_type_ptr, t_pb_graph_pin*>> on_tile_pin_num_to_pb_pin; // [root_pin_physical_num][logical_block] -> t_pb_graph_pin*
     std::unordered_map<int, t_pb_graph_pin*> pin_num_to_pb_pin;                                                       // [intra_tile_pin_physical_num] -> t_pb_graph_pin
+    std::unordered_map<const t_pb_graph_pin*, int> pb_pin_to_pin_num;                                                 // [t_pb_graph_pin*] -> intra_tile_pin_physical_num
```
AlexandreSinger, line 712: @amin1377 If this is planned to be used in drawing code (and performance is not super critical), why store this information?

The pin_num_to_pb_pin has this information. Why not iterate over that map at draw time and reverse it when you need the reverse lookup?

Just curious on your thoughts on this. We are increasing the amount of memory and performance for every tile for a feature unused in the main flow (as of now). Do you think this has use outside of drawing?

amin1377: The number of pins in a cluster can be quite large, since we cover all modes for each pb. As a result, iterating over pin_num_to_pb_pin each time, especially considering it's an unordered_map (which can be slower to iterate than a vector), could be quite costly.

If the flat router is not enabled, these two data structures won't be initialized. Also, t_physical_tile_type is instantiated once per tile type, not per tile, so the increase in memory usage should be minimal.

Additionally, while adding this function, I realized that other parts of the code could be updated to use this data structure when retrieving the physical pin number. That refactoring is beyond the scope of this PR, but having this field in place will be necessary for that future improvement.

AlexandreSinger: Ah! This is per physical tile type, not per tile! In that case this makes sense to me.

PR #3115, opened 2025-06-05
Update golden results

AmirhosseinPoolad:
Fix for broken golden results because of recent changes to the packer.


AlexandreSinger (approved):
LGTM

Issue #3117, opened 2025-06-06
[TestCoverage] Failing VTR Benchmarks on the 7Series Architecture


AlexandreSinger:
Within VTR, we have 3 tests which are testing the 7series architecture (one nightly test and two strong tests):

1. https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/regression_tests/vtr_reg_nightly_test2/vtr_xilinx_qor/config/config.txt
2. https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_xilinx_support/config/config.txt
3. https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/regression_tests/vtr_reg_strong_odin/strong_xilinx_support/config/config.txt

The issue is that none of these tests are actually running on CI since none of them were added to the appropriate task lists. As such, they appear to have regressed.

The two strong tests are actually ill-formed. It looks like they got mangled in an upgrade and lost their testcases. I believe they were just quick tests that tested on stereovision3, but I am not sure.

The NightlyTest2 test is the most important one. It runs the full (I think) 7series architecture on 6 large VTR benchmarks. I tried to revive this testcase on the CI, however all 6 of these circuits are failing.

All 6 circuits are failing at the same point in the Prepacking stage:

<img width="1434" alt="Image" src="https://github.com/user-attachments/assets/433975e8-2b71-4878-99fe-eede09257973" />

The issue appears to be within the following function:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/45945b8f48d8f0978766f986d20574f918c366c9/vpr/src/pack/prepack.cpp#L1620-L1642

This seems to be related to a recent PR which changed the `get_all_connected_pins` function:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3007

To debug this, we may need to view the version just before this PR was merged and try to figure out what changed and why it is not finding any connected primitive pins for this architecture.

amin1377:
Thanks for pointing this out, Alex. I’ll look into the issue.

AlexandreSinger:
I am planning on writing a script that will check all of our task lists and ensure that all tasks in the folders are found in the correct task list. I will add a feature to ignore some tests just to get this in, but we really need to ensure that tasks that are added to VTR are actually being run.

[...]

PR #3120, opened 2025-06-06
Add Xilinx arch tests to CI

WhiteNinjaZ:
#### Description

This PR adds Xilinx tests to CI. 

#### Related Issue

Part of the fix for #3117

#### Types of changes


AmirhosseinPoolad:
The test failures appear to be because the results are "too good." I am a little bit suspicious about the ~40% decrease in the number of pre-packed blocks. @AlexandreSinger Did we upgrade our synthesis tools since October last year?

Regardless, if everything is fine make sure you also update the golden results in regression_tests/vtr_reg_nightly_test2/vtr_xilinx_qor in addition to updating the strong tests golden results.

WhiteNinjaZ:
@AmirhosseinPoolad and @vaughnbetz Unfortunately I don't totally remember exactly what happened between now and the last time the golden was updated for the strong tests but here is what I know:

The strong tests work on a very dumbed down version of the xilinx arch. So its not the most accurate. From what I can tell, the last time the QOR for the strong tests where updated was in 2022. So, there have been significant changes to the code base since the last time the QOR was updated. It might be worth removing the simple-7series.xml checks altogether since that was mostly a placeholder check for making sure that the diagonal wiring connections didn't break while we worked on refining the actual architecture.  I could have the strong tests use the 

Unfortunately, I can't really compare QOR right now for the flagship Xilinx arch ([7series_BRAM_DSP_carry.xml](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/xilinx/7series_BRAM_DSP_carry.xml)) because of issue #3117 but looking at a rolled back version of VTR from about a month ago (May 1st) that I have locally, my QOR does end up being within spec when I run it again.

vaughnb-cerebras:
Sounds good.  I am neutral on the simple arch, but we should definitely have at least one (fast) test on the full 7-series arch in the strong tests so it gets run by CI every time.

AlexandreSinger (approved):
LGTM, thanks @WhiteNinjaZ for looking into this! I am happy that you found the issue!

PR #3121, opened 2025-06-07
[CI] Added Test Suite Verification to CI

AlexandreSinger:
Found that we were regressing on many features in VTR due to tasks being
added to the appropriate test suite directory, but not being included in
the necessary task list. As such, it appeared as though the tests were
being run, but in reality they were not.

Added a script which will be run by the CI which will verify that all of
the test suites that we care about have all their tasks in the
appropriate task list.

From this tool, found many tasks which were not in the task lists.
Marked these tasks as "ignored" for now. These should be handled in a
separate PR.

### DIFF
--- a/dev/vtr_test_suite_verifier/verify_test_suites.py
+++ b/dev/vtr_test_suite_verifier/verify_test_suites.py
@@ -1,0 +1,240 @@
+#!/usr/bin/env python3
+"""
+    Module to verify VTR test suites run by the GitHub CI.
+
+    Test suites in VTR are verified by ensuring that all tasks in the test suite
+    appear in the task list and that there are no tasks in the task list which
+    are not in the test suite.
+
+    A JSON file is used to tell this module which test suites to verify.
+
+    This module is designed to be used within the CI of VTR to ensure that tasks
+    within test suites are running all the tasks they are intended to.
+"""
+import os
+import argparse
+import json
+import sys
+from dataclasses import dataclass, field
+from typing import List, Set
+from pathlib import Path
+
+
+@dataclass(order=True, frozen=True)
+class TestSuite:
+    """
+    Data class used to store information about a test suite.
+    """
+
+    name: str
+    ignored_tasks: List[str] = field(default_factory=list)
+
+
+def parse_test_suite_info(test_suite_info_file: str) -> List[TestSuite]:
+    """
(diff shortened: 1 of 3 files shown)


AmirhosseinPoolad:
Thanks for the PR @AlexandreSinger. I disagree with the current iteration of this CI task. I see the value in this and fully agree with having something like this but I feel like this is putting the cart before the horse a little bit. This CI task should come after the task lists are updated to include all tasks and shouldn't have any exceptions or even the option to have exceptions. Right now it's basically adding a CI that always passes and needs to get updated accordingly which is not really different from the current situation. I don't like the correctness of the CI results being tied to maintainers keeping a config file updated.

dev/vtr_test_suite_verifier/verify_test_suites.py
```diff
@@ -0,0 +1,212 @@
...
+        # a mistake.
+        if task in ignored_tasks:
+            print(f"\tError: Task '{task}' found in task list but was marked as ignored")
+
+    return num_failures
+
+
+def verify_test_suites():
+    """
+    Verify the VTR test suites.
```
vaughnbetz, line 195: Good to say what the verification is.

dev/vtr_test_suite_verifier/verify_test_suites.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env python3
+"""
+    Module to verify VTR test suites run by the GitHub CI.
```
vaughnbetz, line 3: Good to say what the verification is.

dev/vtr_test_suite_verifier/verify_test_suites.py
```diff
@@ -0,0 +1,212 @@
...
+    Data class used to store information about a test suite.
+    """
+
+    name: str
+    ignored_tasks: List[str] = field(default_factory=list)
+
+
+def parse_test_suite_info(test_suite_info_file: str) -> List[TestSuite]:
+    """
+    Parses the given test_suite_info file. The test suite info file is expected
```
vaughnbetz, line 35: Might want to define the json format here, and what it is used for.

vaughnbetz (commented):
LGTM but I suggest a couple of commenting upgrades.

PR #3123, opened 2025-06-08
[AP][Legalization] Updated the Mass Abstraction

AlexandreSinger:
Within the partial legalizer of the AP flow, a mass abstraction was used to abstract the complex logical block constraints of the device. Originally, this abstraction created a primitive vector which kept a count of each model used by the atom and the capacity of each model in a given tile. This worked well, but it was very innacurate due to not capturing differences between models with different numbers of inputs/outputs.

Updated the mass abstraction in three ways:

1) Used the number of bits stored in a memory as the mass for memory
   primitives.

2) Used the number of pins used for all other types of primitives
   (clamped by the capacity of pins for the given logical block).

3) Created the concept of "one-hot" primitives to assign multiple models
   to the same dim if they are part of a primitive which can only every
   implement one of them.

These changes improved the ability of the partial legalizer to understand the details of the complex blocks in the architecture such that it can better legalize during analytical placement.

From experimentation, found that this greatly improved the post-FL quality of the placement with practically no cost in run time (since this does not change the complexity of the legalizer, just spends slightly more time deciding the mass of the atoms / tiles).


AlexandreSinger:
Results on the largest VTR circuits with no fixed IOs (not timing driven):

| Metric | Value normalized to AP baseline |
| ------ | ------------------------------- |
| post_fl_hpwl | 0.795 |
| post_dp_hpwl | 1.015 |
| post-route hpwl | 1.010 |
| ap_gp_runtime | 0.991 |
| ap_fl_runtime | 1.000 |
| ap_dp_runtime | 0.861 |
| total_runtime | 0.980 |

The important metric to look at here is the post-Full Legalization HPWL which improved by 20%. This allowed the detailed placement to run for less time (24% faster); but it got a worst final solution. The quality of the final solution is hard to control, but it does show that there is something wrong with the initial temperature.

AlexandreSinger:
Titan results are in. These results have no fixed IOs and is not timing driven:

| Metric | Value Normalized to AP Baseline |
| ------- | ------------------------------- |
| post_fl_hpwl | 0.761 |
| post_dp_hpwl | 0.978 |
| Post-route WL | 0.983 |
| ap_gp_runtime | 1.027 |
| ap_fl_runtime | 1.024 |
| ap_dp_runtime | 0.955 |
| total_runtime | 1.013 |

The post Full Legalizer HPWL improved by 24%, which is fantastic. This yielded a 2% decrease in post-detailed placement HPWL, which ultimately led to a 2% improved post-route HPWL. This also reduced the run time of the detailed placer since it started and a better starting placement. However, global placement runtime increased slightly which washed any run time gains. Overall this improvement is nothing but good.

I would like to further investigate target density in the partial legalizer; currently it just aims to fill all bins to 100% capacity; but I think if I set the target density based on properties of the physical tiles (i.e. make CLB tiles less filled) we can see even further gains based on some small scale experiments I have been working on.

vpr/src/analytical_place/flat_placement_mass_calculator.cpp
```diff
@@ -6,38 +6,182 @@
...
+    // This will give us a better estimate for how many resources this atom will
+    // use. For example, a LUT atom may only use 2 input pins; however, the
+    // architecture may only have 4-LUTs and 5-LUTs available. This method should
+    // return the smallest compatible resource for this atom (i.e. the 4-LUT).
+    const t_pb_graph_node* primitive = prepacker.get_expected_lowest_cost_pb_gnode(blk_id);
+
+    // Get the mass of this atom (i.e. how much resources on the device this atom
+    // will likely use).
+    float mass = 0.0f;
+    if (!is_primitive_memory_pb_type(primitive->pb_type)) {
```
vaughnbetz, line 145: You might be able to shorten the code by calling your physical routine, and then just blending in two different proportions based on whether it is a memory or not.

vpr/src/analytical_place/flat_placement_mass_calculator.cpp
```diff
@@ -76,19 +219,39 @@ static PrimitiveVector calc_pb_type_capacity(const t_pb_type* pb_type,
...
+    float total_curr_score = pb_type->num_input_pins + pb_type->num_output_pins + pb_type->num_clock_pins;
+    for (PrimitiveVectorDim dim : capacity.get_non_zero_dims()) {
+        // If this dimension corresponds to a memory, do not clamp the capacity.
+        // Memories just count bits.
+        if (memory_model_dims.count(dim) != 0)
+            continue;
+        capacity.set_dim_val(dim, std::min(capacity.get_dim_val(dim), total_curr_score));
+    }
+
+    // A pb_type may contain multiple copies of the same PB. Multiply the capcity
```
vaughnbetz: capacity

vpr/src/analytical_place/flat_placement_mass_calculator.cpp
```diff
@@ -76,19 +219,39 @@ static PrimitiveVector calc_pb_type_capacity(const t_pb_type* pb_type,
...
     // For now, we simply mix the capacities of modes by taking the max of each
     // dimension of the capcities. This provides an upper-bound on the amount of
     // primitives this pb can contain.
     for (int mode = 0; mode < pb_type->num_modes; mode++) {
-        PrimitiveVector mode_capacity = calc_mode_capacity(pb_type->modes[mode], dim_manager);
+        PrimitiveVector mode_capacity = calc_mode_capacity(pb_type->modes[mode], memory_model_dims, dim_manager);
         capacity = PrimitiveVector::max(capacity, mode_capacity);
     }
+
+    // The current pb only has a set number of pins. Therefore, each dimension
```
vaughnbetz: Not super clear to me ... not sure if the comment can be clarified.

vpr/src/analytical_place/flat_placement_mass_calculator.cpp
```diff
@@ -176,28 +345,217 @@ static PrimitiveVector calc_block_mass(APBlockId blk_id,
...
 }
 
+namespace {
+
+/**
+ * @brief A struct to hold information on pb types which act like one-hot primitves.
+ */
+struct OneHotPbType {
+    /// @brief The root pb type which contains the modes which act in a one-hot
+    ///        fasion.
```
vaughnbetz: fasion -> fashion

vpr/src/analytical_place/flat_placement_mass_calculator.cpp
```diff
@@ -176,28 +345,217 @@ static PrimitiveVector calc_block_mass(APBlockId blk_id,
...
+                                   int log_verbosity) {
     // Set the mapping between model IDs and Primitive Vector IDs
 
+    // Find all of the pb types that act like a one-hot primitive.
+    std::vector<OneHotPbType> one_hot_pb_types = find_one_hot_pb_types(logical_block_types,
+                                                                       models,
+                                                                       log_verbosity);
+
+    // For each model, label them with their shared model ID if they are part
+    // of one, -1 otherwise.
```
vaughnbetz: of one -> of a one_hot_pb_type

vpr/src/analytical_place/primitive_dim_manager.h
```diff
@@ -47,17 +47,37 @@ class PrimitiveDimManager {
...
         dims_.push_back(new_dim);
         dim_name_.push_back(name);
-        model_dim_.insert(model_id, new_dim);
+        return new_dim;
+    }
+
+    /**
+     * @brief Add the given model ID to the given primitive vector dim.
+     *
+     * It is assumed that the given model is not part of any other dimensions.
```
vaughnbetz: dimensions -> dimension

vaughnbetz (approved):

vpr/src/analytical_place/flat_placement_mass_calculator.cpp
```diff
@@ -176,28 +345,217 @@ static PrimitiveVector calc_block_mass(APBlockId blk_id,
...
+            std::set<LogicalModelId> contained_models;
+            // Check if every pb_type in each mode is single and unique.
+            for (int mode_idx = 0; mode_idx < pb_type->num_modes; mode_idx++) {
+                const t_mode& mode = pb_type->modes[mode_idx];
+                // Check if the mode is singular.
+                if (mode.num_pb_type_children != 1) {
+                    is_one_hot = false;
+                    break;
+                }
+                // Check if the mode is unique.
```
AlexandreSinger: Note to self: This is missing a condition to check. What is being checked is that it is singular and is a primitive. It is not checking uniqueness.

PR #3124, opened 2025-06-09
Update ezgl to use submodules

AmirhosseinPoolad:
This PR is the first step in trying to move our internally owned external libraries such as ezgl and tatum to use submodules instead of subtrees since we want these libraries to stay in sync and apply all changes to their own repository and not have many/any downstream patches in the VTR repository.

The code currently compiles but is not tested.


SamuelHo10:
I've tested the GUI and verified that the documentation is building correctly. I think this PR is ready to merge.

AlexandreSinger:
Whats great about this is that now it should be tracked by Dependabot, so we should always have the most up-to-date EZGL documentation.

@SamuelHo10 Do be sure to write down the steps you took to verify the GUI. We will need to perform those steps every time we update EZGL versions on VTR. EZGL does not change much, but its good to be safe.

PR #3125, opened 2025-06-09
Read/Write Encrypted XML Files

amin1377:
In this PR, libraries related to encrypting and decrypting files from OpenFPGA have been added. The PugiXML utility has been updated to support reading and writing encrypted files.


libs/libdecrypt/src/obfuscate.h
```diff
@@ -0,0 +1,182 @@
+#pragma once
```
soheilshahrouz, line 1: This file seems to have come from [this repo](https://github.com/adamyaxley/Obfuscate). I think it should be added as a submodule

libs/libencrypt/src/obfuscate.h
```diff
@@ -0,0 +1,182 @@
+#pragma once
```
soheilshahrouz, line 1: This is exactly the same as another file with the same name. I think we should remove both files and add them as submodules.

soheilshahrouz (changes requested):
There is duplicated header file that should be added as a submodule. Are libdecrypt and libencrypt part of OpenFPGA or they are external libraries?

Which xml files are expected to be encrypted? Is reading encrypted xml files an essential feature of OpenFPGA?

amin1377:
@soheilshahrouz:

I'm not sure it's an "essential" part of OpenFPGA, but it is certainly a useful feature, and we do use it, mainly to read the architecture file.

Within OpenFPGA, it's used directly in the codebase (not as a submodule or external dependency). Given that, I'm inclined to follow the same approach and avoid adding it as a submodule.

.github/scripts/hostsetup.sh
```diff
@@ -70,6 +70,8 @@ apt install -y \
   g++-9 \
   gcc-9 \
   wget \
+  openssl \
+  libssl-dev \
```
AlexandreSinger: I do not really like the idea of including more requirements for VTR for an optional feature which is not used by most users. We should only include SSL if the XML features are used. There are CMake ways of doing that.

AlexandreSinger: Also, this script is only used for the CI, and I do not think this library is used by any CI tests.

CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set_property(CACHE VTR_IPO_BUILD PROPERTY STRINGS auto on off)
 set(VTR_ASSERT_LEVEL "2" CACHE STRING "VTR assertion checking level. 0: no assertions, 1: fast assertions, 2: regular assertions, 3: additional assertions with noticeable run-time overhead, 4: all assertions (including those with significant run-time cost)")
 set_property(CACHE VTR_ASSERT_LEVEL PROPERTY STRINGS 0 1 2 3 4)
 
+option(VTR_ENABLE_STRICT_COMPILE "Specifies whether compiler warnings should be treated as errors (e.g. -Werror)" OFF)
```
AlexandreSinger: This was an option that I removed a few weeks ago. Why add it back?

install_apt_packages.sh
```diff
@@ -9,7 +9,9 @@ sudo apt-get install -y \
     bison \
     flex \
     python3-dev \
-    python3-venv
+    python3-venv \
+    openssl \
+    libssl-dev
```
AlexandreSinger: Same comment as above. By adding openSSL to this script, you are saying that every user of VTR needs OpenSSL to run VTR.

libs/libdecrypt/ README.md
```diff
@@ -0,0 +1,98 @@
+ # libdecrypt
```
AlexandreSinger, line 1: Is this a fully custom library, or something from the internet? I feel like encrypting XML files should not be complicated enough to warent its own library? Are we sure a library for this does not already exist?

libs/libdecrypt/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+file(READ "config.txt" PASSPHRASE_CONTENTS)
+
+string(REPLACE "\n" "\\n" PRIVATE_KEY_CONTENTS "${PRIVATE_KEY_CONTENTS}")
+string(REPLACE "\n" "\\n" PASSPHRASE_CONTENTS "${PASSPHRASE_CONTENTS}")
+set(PRIVATE_KEY "${PRIVATE_KEY_CONTENTS}")
+set(PASS_PHRASE "${PASSPHRASE_CONTENTS}")
+add_compile_definitions(PASS_PHRASE="${PASS_PHRASE}")
+add_compile_definitions(PRIVATE_KEY="${PRIVATE_KEY}")
+find_package(PkgConfig REQUIRED)
+pkg_search_module(OPENSSL REQUIRED openssl)
```
AlexandreSinger, line 14: This line makes OpenSSL a required library for all of VTR. Be aware. I would recommend creating a top-level CMake variable to allow the user to build with this feature or not. I think this is a large dependency for this feature.

libs/libdecrypt/config.txt
```diff
@@ -0,0 +1 @@
+abcd
```
AlexandreSinger: What is this file?

libs/libdecrypt/private_key.pem
```diff
@@ -0,0 +1,30 @@
+-----BEGIN ENCRYPTED PRIVATE KEY-----
```
AlexandreSinger, line 1: Is it common to include a private key in a repository like this? Shouldnt it be public?

libs/libpugiutil/src/pugixml_loc.hpp
```diff
@@ -6,7 +6,8 @@
 
 #include <vector>
 #include "pugixml.hpp"
-
+#include "decryption.h"
```
AlexandreSinger: Why add the decryption header file to the hpp file and not the CPP file?

libs/libpugiutil/src/pugixml_util.cpp
```diff
@@ -8,21 +8,46 @@ namespace pugiutil {
...
-                           + " (line: " + std::to_string(line) + " col: " + std::to_string(col) + ")",
-                       filename.c_str(), line);
+    //store the position of last '.' in the file name
+    size_t position = filename.find_last_of(".");
+    std::string result = "";
+    if (position != std::string::npos) {
+        //store the characters after the '.' from the file_name string
+        result = filename.substr(position);
+    }
+    if (result == ".xmle") {
```
AlexandreSinger, line 20: This should really be an optional feature in my opinion. I think IFDEFing this out would be a good idea.

AlexandreSinger (commented):
Hi @amin1377 ,

I have some concerns with bringing these libraries in. What is the context of this library? Was this copied from somewhere else? I find it a bit strange that this is separated into two separate libraries. I also am surprised a library like this does not already exist. I think rolling our own encryption / decryption flow can come with unwanted baggage.

Also, I think that this should be an optional feature (similar to CapNProto where it will only turn on if CapNProto is installed). I think we should avoid adding dependencies unless it is a feature that we know everyone wants.

CMakeLists.txt
amin1377: I wasn’t aware it had been in VTR before. I came across it on the OpenFPGA side and thought it was a useful feature to have. I've removed it now. Just out of curiosity, do you know why it was removed?

libs/libdecrypt/ README.md
amin1377: I agree with you and have noted both your and Soheil’s concerns. If you’re okay with it, I suggest we prioritize completing the full integration with OpenFPGA first, and we can revisit and roll back any changes we don’t want afterward.

libs/libdecrypt/CMakeLists.txt
amin1377: I believe that now, with the addition of the new variable VTR_ENABLE_ENCRYPTION to indicate whether the user wants encryption, this CMake will only be triggered if the user wants.

libs/libdecrypt/config.txt
amin1377: According to the README, this file is used if the user chooses to use a passphrase. I added the following line from the README to this file:

```
Write your passphrase in the file
```

CMakeLists.txt
AlexandreSinger: There is a much better option provided by CMake which can be used which works globally on VTR which does everything this option does and a bit more:
```
-DCMAKE_COMPILE_WARNING_AS_ERROR=on
```

The other option did some strange things because it actively allowed specific warnings through and it only worked on the VPR executable and not the other executables that are built for VTR.

Edit: If I recall correctly, it did also work on some libraries, but it did not work on everything which was a bit of a problem when we were trying to make the build warning clean.

libs/libdecrypt/private_key.pem
amin1377: It’s just a placeholder for the user to insert the private key they intend to use.

libs/libdecrypt/ README.md
AlexandreSinger: I am not sure I agree. I thought I understood from the meeting last Friday that the goal was to bring in the necessary data structures first, then the necessary features, then the bonus features last. This feels more like a bonus feature which requires a bit more discussion.

My worry about this is that this is another thing that we will need to support in VTR; and encryption can come with a good amount of baggage. We need to really think about how much we want this feature.

libs/libpugiutil/src/pugixml_util.cpp
```diff
@@ -8,8 +8,37 @@ namespace pugiutil {
 //Returns loc_data look-up for xml node line numbers
 loc_data load_xml(pugi::xml_document& doc,      //Document object to be loaded with file contents
                   const std::string filename) { //Filename to load from
+#ifdef VTR_ENABLE_ENCRYPTION
```
AlexandreSinger, line 11: You are missing something in your CMake file. CPP code cannot read variables from CMake like this. In CMake you are going to need to set this as a define for the PUGI target. See how this is done for Eigen in the AP flow.

libs/libpugiutil/src/pugixml_util.cpp
```diff
@@ -8,8 +8,37 @@ namespace pugiutil {
...
+#include "decryption.h"
```
AlexandreSinger, line 12: Normally to do this we still put the header file at the top of the file and wrap ifdefs around it. Its a bit strange to put the header all the way down here.

libs/libpugiutil/src/pugixml_util.cpp
```diff
@@ -8,8 +8,37 @@ namespace pugiutil {
...
+            auto line = location_data.line(load_result.offset);
+            auto col = location_data.col(load_result.offset);
+            throw XmlError("Unable to load XML file '" + filename + "', " + msg
+                               + " (line: " + std::to_string(line) + " col: " + std::to_string(col) + ")",
+                           filename.c_str(), line);
+        }
+        delete[] final;
+        return location_data;
+    }
+#endif
```
AlexandreSinger, line 39: Minor nit: I would comment what this is ending.

libs/libpugiutil/src/pugixml_util.cpp
```diff
@@ -8,8 +8,37 @@ namespace pugiutil {
...
+    //auto location_data = loc_data(end_result_fname);
```
AlexandreSinger, line 40: Commented code.

PR #3128, opened 2025-06-10
[CI] Revived Simple Missing Strong Tests

AlexandreSinger:
The new test suite verifying tool showed that a few tasks from the strong tests were not actually being tests. Some of these tests worked without issue.

Revived the tests which worked without issue.


AmirhosseinPoolad:
LGTM.

PR #3129, opened 2025-06-10
Subtile Pin Location

amin1377:
In the architecture file, when a sub-tile has a capacity bigger than one, it is assumed that the pin location of all sub-tiles should be assigned together:

```
<tile name="io">
      <sub_tile name="io" capacity="2">
       ....
                      <loc side="left">  io.core_in[0:24]  io.core_out[0:19]  io.clk[0:2]</loc>
       ....
```
In the above case, the first half of the IO pins for both sub-tiles will be placed on the left side of the device.

In this PR, a feature from OpenFPGA is added where one can define different locations for each sub-tile capacity if necessary:

```
<tile name="io">
      <sub_tile name="io" capacity="2">
       ....
                      <loc side="left">  io[0:0].core_in[0:24]  io[0:0].core_out[0:19]  io[1:1].clk[0:2]</loc>
       ....
```


libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -123,6 +123,11 @@ static std::pair<int, int> ProcessPinString(pugi::xml_node Locations,
                                             T type,
                                             const char* pin_loc_string,
                                             const pugiutil::loc_data& loc_data);
+template<typename T>
+static std::pair<int, int> ProcessInstanceString(pugi::xml_node Locations,
```
soheilshahrouz: Most function names in this file are camel case, but the rest of code base uses snake case. We should gradually change this. NoC-realted function names are snake case. I think new functions should use snake case too.

Why is this a template function? It is only called once with `T=t_sub_tile*`. I also suggest we pass it by const reference instead of pointer.

Add doxygen comment for this function

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -703,10 +708,15 @@ static void LoadPinLoc(pugi::xml_node Locations,
                                                                                &sub_tile,
                                                                                token.c_str(),
                                                                                loc_data);
-
+                                /* Get the offset in the capacity range */
+                                auto capacity_range = ProcessInstanceString<t_sub_tile*>(Locations,
```
soheilshahrouz: You can use `auto` to unpack the pair and give each of its members a name. For example:
` auto [capacity_range_low, capacity_range_high] = ...`

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -703,10 +708,15 @@ static void LoadPinLoc(pugi::xml_node Locations,
                                                                                &sub_tile,
                                                                                token.c_str(),
                                                                                loc_data);
-
+                                /* Get the offset in the capacity range */
```
soheilshahrouz: Use `//` to be consistent with the coding style guide.

libs/libarchfpga/src/read_xml_arch_file.cpp
AlexandreSinger: This needs a doxygen comment. Also I do not think this should be templated. From glossing over the code, I think that the template must be of type t_sub_tile* to make sense.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -703,10 +708,15 @@ static void LoadPinLoc(pugi::xml_node Locations,
...
+                                auto capacity_range = ProcessInstanceString<t_sub_tile*>(Locations,
+                                                                                         &sub_tile,
+                                                                                         token.c_str(),
+                                                                                         loc_data);
+                                VTR_ASSERT(0 <= capacity_range.first && capacity_range.second < sub_tile_capacity);
```
soheilshahrouz: Might be a good idea to assert that `second` is greater than or equal to `first`

AlexandreSinger: This can be a confusing assert to hit for people unfamiliar. I recommend using VTR_ASSERT_MSG and adding a message to explain what failed when this assert is hit.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -736,6 +746,99 @@ static void LoadPinLoc(pugi::xml_node Locations,
     }
 }
 
+/* Parse the string to extract instance range, e.g., io[4:7] -> (4, 7)
+ * If no instance range is explicitly defined, we assume the range of type capacity, i.e., (0, capacity - 1) */
```
soheilshahrouz: This comment can be moved to the declaration and made a doxygen comment.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -736,6 +746,99 @@ static void LoadPinLoc(pugi::xml_node Locations,
...
+template<typename T>
+static std::pair<int, int> ProcessInstanceString(pugi::xml_node Locations,
+                                                 T type,
+                                                 const char* pin_loc_string,
+                                                 const pugiutil::loc_data& loc_data) {
+    int num_tokens;
+    auto tokens = GetTokensFromString(pin_loc_string, &num_tokens);
```
soheilshahrouz: replace `auto` with explicit type

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -736,6 +746,99 @@ static void LoadPinLoc(pugi::xml_node Locations,
...
+
+    int token_index = 0;
+    auto token = tokens[token_index];
```
soheilshahrouz: `auto`

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -736,6 +746,99 @@ static void LoadPinLoc(pugi::xml_node Locations,
...
+
+    if (token.type != TOKEN_STRING || 0 != strcmp(token.data, type->name.c_str())) {
+        archfpga_throw(loc_data.filename_c_str(), loc_data.line(Locations),
```
AlexandreSinger, line 772: Although it is not "required", I think `freeTokens` should be called before all of the throw statements in the function for cleanliness. I have run into issues in the past where throws fail due to memory not being deallocated properly.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -736,6 +746,99 @@ static void LoadPinLoc(pugi::xml_node Locations,
...
+                                                 T type,
+                                                 const char* pin_loc_string,
+                                                 const pugiutil::loc_data& loc_data) {
+    int num_tokens;
+    auto tokens = GetTokensFromString(pin_loc_string, &num_tokens);
+
+    int token_index = 0;
+    auto token = tokens[token_index];
+
+    if (token.type != TOKEN_STRING || 0 != strcmp(token.data, type->name.c_str())) {
```
soheilshahrouz: `token.data != type->name` is easier to read

AlexandreSinger (changes requested):
This looks good to me other than some minor comments. I am not as familiar with this code, but I gave some general comments.

Can you please add some documentation for this new feature to the architecture docs? I think you can just directly copy the docs from OpenFPGA.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -736,6 +746,99 @@ static void LoadPinLoc(pugi::xml_node Locations,
...
+                       "Wrong physical type name of the port: %s\n", pin_loc_string);
+    }
+
+    token_index++;
+    token = tokens[token_index];
+
+    int first_inst = 0;
+    int last_inst = type->capacity.total() - 1;
+
+    /* If there is a dot, such as io.input[0:3], it indicates the full range of the capacity, the default value should be returned */
```
soheilshahrouz: `//` to be consistent with coding style guide

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -736,6 +746,99 @@ static void LoadPinLoc(pugi::xml_node Locations,
...
+    if (token.type == TOKEN_DOT) {
+        freeTokens(tokens, num_tokens);
+        return std::make_pair(first_inst, last_inst);
+    }
+
+    /* If the string contains index for capacity range, e.g., io[3:3].in[0:5], we skip the capacity range here. */
```
soheilshahrouz: `//` to be consistent

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -755,6 +858,20 @@ static std::pair<int, int> ProcessPinString(pugi::xml_node Locations,
     token_index++;
     token = tokens[token_index];
 
+    /* If the string contains index for capacity range, e.g., io[3:3].in[0:5], we skip the capacity range here. */
```
soheilshahrouz: `//` to be consitent with style guide

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3423,13 +3554,15 @@ static void ProcessPinLocations(pugi::xml_node Locations,
...
         //Check for any pins missing location specs
-        for (const auto& port : SubTile->ports) {
-            for (int ipin = 0; ipin < port.num_pins; ++ipin) {
-                if (!port_pins_with_specified_locations[port.name].count(ipin)) {
-                    //Missing
-                    archfpga_throw(loc_data.filename_c_str(), loc_data.line(Locations),
-                                   "Pin '%s.%s[%d]' has no pin location specified (a location is required for pattern=\"custom\")",
-                                   SubTile->name.c_str(), port.name, ipin);
+        for (int iinst = SubTile->capacity.low; iinst < SubTile->capacity.high; ++iinst) {
+            for (const auto& port : SubTile->ports) {
```
soheilshahrouz: `auto`

soheilshahrouz (changes requested):

soheilshahrouz (changes requested):
Thanks for this new feature.
I think it should be also documented in the architecture reference.

libs/libarchfpga/src/read_xml_arch_file.cpp
amin1377: Yes, you’re right. For some reason, which isn’t clear to me either, the functions that take t_sub_tile* are templated. I didn’t change it in order to stay consistent with the other functions, but I agree it would be better to use the known type if templating isn’t necessary.

libs/libarchfpga/src/read_xml_arch_file.cpp
amin1377: Modify the assert to:

```
VTR_ASSERT_MSG(0 <= capacity_range.first && capacity_range.second < sub_tile_capacity,
                                               vtr::string_fmt("Capacity range is out of bounds: capacity_range.first: %d, "
                                                "capacity_range.second: %d, sub_tile_capacity: %d",
                                                capacity_range.first,
                                                capacity_range.second,
                                                sub_tile_capacity)
                                               .c_str());
```

libs/libarchfpga/src/read_xml_arch_file.cpp
amin1377: Interesting! I added a helper function throw_xml_arch_error that takes the tokens and frees them before throwing an error. I replaced all archfpga_throw calls with this helper function (it took me a while :) )

```
static void throw_xml_arch_error(const char* filename, 
                                  int line, 
                                  const std::string& err_msg, 
                                  t_token* tokens, 
                                  int num_tokens);
```

libs/libarchfpga/src/read_xml_arch_file.cpp
amin1377: I changed the name of all functions in this file from camel case to snake case.

libs/libarchfpga/src/read_xml_arch_file.cpp
amin1377: Look who’s suddenly in favor of using auto :)) (just kidding)

amin1377:
@soheilshahrouz @AlexandreSinger: Thanks for reviewing this PR. I’ve applied your comments. Please let me know if you have any further suggestions.

AlexandreSinger (commented):
LGTM. The throw function freeing memory makes sense, but its a bit unclear to someone using the code that they need to do that; however I know that Soheil is working on cleaning up the token code to prevent this so it does not matter.

I still think it would be good to document this change!

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -113,73 +113,109 @@ struct t_pin_locs {
...
+ * @param line Line number associated with the error
+ * @param err_msg Error message to be displayed
+ * @param tokens If tokens are allocated before calling this function,
+ * they will be freed here. If not, nullptr should be passed.
+ * @param num_tokens Number of tokens to be freed. If not, 0 should be passed.
+ */
+static void throw_xml_arch_error(const char* filename,
+                                 int line,
+                                 const std::string& err_msg,
+                                 t_token* tokens,
```
soheilshahrouz: In another PR, I created a PR to make memory management for tokens automatic. I think this is no longer necessary.

PR #3132, opened 2025-06-11
Skip syncing netlist to routing resutls

amin1377:
There may be cases where the user doesn’t want to sync the netlist to the routing results. This situation may especially arise when the flat router is enabled and global signals are not routed. Since the pins reserved for global nets are not visible to the flat router, it may use them, which can lead to crashes.

For this case, the current workaround is to enable skipping netlist syncing. However, the long-term solution would be to make these pins visible to the flat router.

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -3098,7 +3098,9 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
     analysis_grp.add_argument<bool, ParseOnOff>(args.skip_sync_clustering_and_routing_results, "--skip_sync_clustering_and_routing_results")
         .help(
             "Select to skip the synchronization on clustering results based on routing optimization results."
-            "Note that when this sync-up is disabled, clustering results may be wrong (leading to incorrect bitstreams)!");
+            "Note that when this sync-up is disabled, clustering results may be wrong (leading to incorrect bitstreams)!")
+        .default_value("off")
+        .show_in(argparse::ShowIn::HELP_ONLY);
 
     analysis_grp.add_argument<bool, ParseOnOff>(args.generate_net_timing_report, "--generate_net_timing_report")
         .help(
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -3095,6 +3095,11 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .help("Writes implemented design final timing summary to the specified JSON, XML or TXT file.")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    analysis_grp.add_argument<bool, ParseOnOff>(args.skip_sync_clustering_and_routing_results, "--skip_sync_clustering_and_routing_results")
+        .help(
+            "Select to skip the synchronization on clustering results based on routing optimization results."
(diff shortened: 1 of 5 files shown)


vpr/src/base/read_options.cpp
```diff
@@ -3095,6 +3095,13 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .help("Writes implemented design final timing summary to the specified JSON, XML or TXT file.")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    analysis_grp.add_argument<bool, ParseOnOff>(args.skip_sync_clustering_and_routing_results, "--skip_sync_clustering_and_routing_results")
```
AlexandreSinger, line 3098: I will leave this up to you if you want to add this to the documentation or not. Since this is a band-aid fix to get around some crashes I understand leaving it only in the help and not in the docs.

vpr/src/base/read_options.cpp
```diff
@@ -3095,6 +3095,13 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
...
+        .help(
+            "Select to skip the synchronization on clustering results based on routing optimization results."
+            "Note that when this sync-up is disabled, clustering results may be wrong (leading to incorrect bitstreams)!")
```
AlexandreSinger, line 3101: Is this something that we want to allow at all? This seems kinda dangerous to sometimes allow clustering to be wrong...

vpr/src/base/vpr_api.cpp
```diff
@@ -1426,23 +1426,28 @@ bool vpr_analysis_flow(const Netlist<>& net_list,
...
+                                        vpr_setup.PackerOpts.pack_verbosity > 2);
+                }
         } else {
-            sync_netlists_to_routing(net_list,
-                                     g_vpr_ctx.device(),
-                                     g_vpr_ctx.mutable_atom(),
-                                     g_vpr_ctx.mutable_clustering(),
-                                     g_vpr_ctx.placement(),
-                                     vpr_setup.PackerOpts.pack_verbosity > 2);
+            VTR_LOG_WARN("Sychronization between packing and routing results is not applied due to users select to skip it\n");
```
AlexandreSinger: Sychronization -> Synchronization

AlexandreSinger: Also I think this should be past-tense. "is" -> "was".

AlexandreSinger: Also, I would just name the option:

`skip_sync_clustering_and_routing_results`

I think saying "users select to skip it" very confusing. You may also want to warn that the bitstream may not be correct in this warning since that I think is the important reason why this is a warning.

AlexandreSinger (commented):
LGTM, some comments below.

vpr/src/base/read_options.cpp
amin1377: Yes, that's exactly why I didn't put it in the docs.

vpr/src/base/read_options.cpp
amin1377: The code for syncing the netlist to the routing results (especially when the flat router is enabled) proved to be tricky. In fact, before Fahri added support for it, we had disabled netlist syncing when the flat router was used. However, since this support has been added recently, we now need to have a way to disable it if necessary.

amin1377:
@AlexandreSinger: Thanks for reviewing the code. Feel free to merge the PR if there are no further comments.

PR #3136, opened 2025-06-12
[AP] Updated Timing Logging

AlexandreSinger:
The global placer printed the post-GP estimated wirelength, but it did not print the post-GP CPD or TNS. Added these values to the log file.

Also updated the QoR parse script that I use internally to grab this information throughout the flow to track timing.

### DIFF
--- a/vpr/src/analytical_place/global_placer.cpp
+++ b/vpr/src/analytical_place/global_placer.cpp
@@ -124,9 +124,18 @@ SimPLGlobalPlacer::SimPLGlobalPlacer(e_ap_analytical_solver analytical_solver_ty
  */
 static void print_placement_stats(const PartialPlacement& p_placement,
                                   const APNetlist& ap_netlist,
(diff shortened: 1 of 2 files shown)


amin1377 (approved):
Looks good, Alex! Thanks!

Two questions:

1- Don’t you need to update the timing information before calling the API for the CPD info?

2- Is there anywhere in the documentation where we explain the meaning of these metrics? Thanks to your presentations, we’re familiar with gp, fl, dp, etc., but I’m not sure it would be obvious for someone new to the project. The same comment applies to other metrics as well. I don’t think we have documentation that explains what each metric means. While I understand that most of them are probably self-explanatory, it can be a bit confusing, especially when searching for minimum channel width, since it’s not always clear which metric refers to minimum channel width and which one refers to relaxed channel width.

I realize this is beyond the scope of this PR, but if you agree, I think we should open an issue to track it.

Issue #2974, opened 2025-04-13
[AP][GP] High-Fanout Nets Handling


AlexandreSinger:
I am noticing in the VTR and Titan runs that many of the designs have some very high fanout nets which are not labelled as clock or global.

In the AP flow, I explicitly remove nets from the AP netlist which are clock / global since they do not contribute much to the solution (clocks / global signals go all over the chip, so their HPWL will not change much during GP). This was also done to improve performance, since high-fanout connections add many terms to the connectivity matrix.

Below is a part of the GP report for MCML from the VTR benchmarks (in this case with fixed IOs, but it should not matter):
<img width="666" alt="Image" src="https://github.com/user-attachments/assets/499c499a-f8e5-41a8-ba9d-f5173e193096" />

As you can see, the max fanout of any net in the AP netlist is 51,878, while the average is only around 5. Based on my intuition, the solver should be faster than the legalizer; however, I believe that these high-fanout nets are causing the solver to slow-down dramatically with very little gain.

We should investigate adding a threshold to the solver to have it ignore nets that are "high-fanout". We should be able to just ignore the nets which are above this threshold when we are building the connectivity matrices.

AlexandreSinger:
This has been completed. We also found that we were not actually ignoring global / clock nets correctly. This has also been fixed.

PR #3137, opened 2025-06-12
High Fanout Net Thresholding in AP to Speed Up Solver

haydar-c:
#### Description
This PR filters out the nets with higher fanout than a specified threshold from the APNetlist. The aim is to make the solver faster without hurting the quality or results. This PR also adds a command line option to set this threshold. The default value is set to 256 by trying on Titan, Koios, and VTR Chain benchmarks. This is a safe value. The command-line option also allows tuning if needed, depending on the solver configuration.

#### Related Issue
This PR solves the issue mentioned in https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2974.

#### Motivation and Context
The motivation is to make the solver faster by ignoring the high fanout nets without degrading the quality of results.

After selecting the default value of 256, all three benchmarks of Titan, Koios, and VTR Chain is run with the default value and with master (without the changes introduced in this PR). 

In each run, the following build command was used:

```bash
make CMAKE_PARAMS="-DVTR_IPO_BUILD=on -DVTR_ASSERT_LEVEL=1"
```

All metrics below are normalized to the corresponding master run.

| Benchmark Suite | Post GP HPWL | Post FL HPWL | Routed Wirelength | GP Solver Runtime | GP Runtime | AP Runtime | Total Runtime | Num LABs/CLBs | Total CG Iters |
|------------------|---------------|----------------|--------------------|--------------------|-------------|-------------|----------------|------------|------------------|
| Titan            | 0.9837        | 1.0108         | 1.0006             | 0.7706             | 0.7957      | 0.9249      | 0.9384         | 0.9993     | 0.9881           |
| Koios            | 0.9476        | 0.9868         | 0.9967             | 0.6993             | 0.7389      | 0.9334      | 0.9520         | 0.9990     | 0.9956           |
| VTR Chain        | 0.9958        | 0.9749         | 0.9903             | 0.7650             | 0.7868      | 0.9300      | 0.9396         | 1.0018     | 0.9775           |

All in all, applying the high fanout net thresholding reduces GP solver runtime by **23–30%**, translating to an overall total runtime improvement of **5–6%** across **Titan, Koios, and VTR Chain** benchmarks. Quality metrics such as routed wirelength and LAB count are preserved, with observed differences assumed to be noise.

### DIFF
--- a/doc/src/vtr/run_vtr_flow.rst
+++ b/doc/src/vtr/run_vtr_flow.rst
@@ -90,6 +90,13 @@ Enable Synlig tool with the ``-DSYNLIG_SYSTEMVERILOG=ON`` compile flag for the P
 
 Will run the VTR flow (default configuration) with Yosys frontend using Parmys plugin as partial mapper. To utilize the Parmys plugin, the ``-DYOSYS_PARMYS_PLUGIN=ON`` compile flag should be passed while building the VTR project with Yosys as a frontend.
 
+.. code-block:: bash
+
+    # Using the Parmys (Partial Mapper for Yosys) plugin as partial mapper with include files
+    ./run_vtr_flow <path/to/Verilog/File> <path/to/arch/file> -include <path/to/include/directory>/*.v*
+
+Will run the VTR flow (default configuration) with Yosys frontend using Parmys plugin as partial mapper. In addition to the main circuit passed in with the architecture, it will also pass in every HDL file with the specified file type within the include directory.
+
 Detailed Command-line Options
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -123,6 +130,15 @@ Detailed Command-line Options
       * ``vpr``
 
     **Default:** ``vpr``
+    
+.. option:: -include <path_to_file(s)>/*.<file_type(s)>
+
+    List of include files to a benchmark circuit 
+    (pass to VTR frontends as a benchmark design set).
+    
+    Include files can be any file supported by yosys+parmys (normally .v or .vh files).
+    
+    The include directory should not contain the circuit passed in with the architecture.
 
 .. option:: -power
 
--- a/libs/libvtrutil/src/vtr_util.cpp
+++ b/libs/libvtrutil/src/vtr_util.cpp
@@ -375,67 +375,6 @@ char* fgets(char* buf, int max_size, FILE* fp) {
     return nullptr;
 }
 
-/**
- * @brief to get an arbitrary long input line and cut off any
- * comment part 
- * 
- * the getline function is exaly like the __get_delim function 
- * in GNU with '\n' delimiter. As a result, to make the function 
- * behaviour identical for Windows (\r\n) and Linux (\n) compiler 
- * macros for checking operating systems have been used.
- * 
- * @note user need to take care of the given pointer,
- * which will be dynamically allocated by getdelim
- */
-char* getline(char*& _lineptr, FILE* _stream) {
-    int i;
-    int ch;
-    size_t _n = 0;
-    ssize_t nread;
-
-#if defined(__unix__)
-    nread = getdelim(&_lineptr, &_n, '\n', _stream);
-#elif defined(_WIN32)
-#define __WIN_NLTK "\r\n"
-    nread = getdelim(&_lineptr, &_n, __WIN_NLTK, _stream);
-#endif
-
-    if (nread == -1) {
-        int errsv = errno;
-        std::string error_msg;
-
-        if (errsv == EINVAL)
-            error_msg = string_fmt("[%s] Bad arguments (_lineptr is NULL, or _stream is not valid).", strerror(errsv));
-        else if (errsv == ENOMEM)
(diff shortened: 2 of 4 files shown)


vpr/src/analytical_place/gen_ap_netlist_from_atoms.cpp
```diff
@@ -153,6 +155,13 @@ APNetlist gen_ap_netlist_from_atoms(const AtomNetlist& atom_netlist,
             ap_netlist.set_net_is_ignored(ap_net_id, true);
             continue;
         }
+        // If fanout number of the net is higher than the threshold, mark as ignored for AP.
+        size_t num_pins = ap_netlist.net_pins(ap_net_id).size();
+        VTR_ASSERT_DEBUG(num_pins > 1);
+        if (num_pins - 1 > static_cast<size_t>(high_fanout_threshold)) {
+            ap_netlist.set_net_is_ignored(ap_net_id, true);
```
AlexandreSinger, line 162: Heads up about setting the net as ignored. This does ignore the net during global placement; however, it also ignores the net when calculating the HPWL during Global Placement!

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/844665353ecaf34206fda732c5044dc2a6291fde/vpr/src/analytical_place/partial_placement.cpp#L14-L34

This only affects calculating the HPWL during GP, the other HPWL calculations (such as post-FL and post-DP) will not be affected. I originally did this for debugging (since we do not really care if nets that we are ignoring are getting longer); however this may be a bit confusing now that this is becoming more mature. I honestly have no idea how to resolve this issue in practice. Should we even be ignoring nets when computing HPWL?

@vaughnbetz I guess this is more of a question for you. Do you see any issue with ignoring nets during the HPWL estimation. We do use this for debugging as well as part of the algorithm to estimate the quality of the placement. My gut feeling is to ignore the nets when computing HPWL since its just an estimate anyways. What do you think?

vaughnbetz: I think we should have a separate flag (or just inline code) to control what goes in the solver. The rest of the flow only ignores nets that are assumed to be perfectly routed on a global network. Ignoring some algorithmically selected nets in howl calculations is going to be confusing as it doesn't match the rest of the flow.

AlexandreSinger: Thats a great idea Vaughn! I like that idea. In the solver we can ignore the nets (since it would speed up computing HPWL each iteration and it would make the value more accurate to what we are optimizing), and then when we report the final HPWL we can only ignore nets marked as global!

@haydar-c Lets not gate your change! This is something that I can add after your PR is merged.

haydar-c: Thanks, sounds great!

vpr/src/analytical_place/gen_ap_netlist_from_atoms.h
```diff
@@ -16,13 +16,16 @@ class UserPlaceConstraints;
...
+ *  @param high_fanout_threshold The threshold above which nets with higher fanout will
+ *                               be ignored.
  *
  *  @return             An APNetlist object, generated from the prepacker results.
  */
 APNetlist gen_ap_netlist_from_atoms(const AtomNetlist& atom_netlist,
                                     const Prepacker& prepacker,
-                                    const UserPlaceConstraints& constraints);
+                                    const UserPlaceConstraints& constraints,
+                                    const int& high_fanout_threshold);
```
AlexandreSinger: Nit: int do not need to be passed by reference. In fact, it may actually be slower to pass an int by reference. Turn into pass by value.

AlexandreSinger: For simple types, like integers, you can just pass them directly:

```
int high_fanout_threshold);
```

We pass the other arguments to this function by const reference to prevent deep copies which are very very expensive for these types.

AlexandreSinger (changes requested):
Looks fantastic to me! Very well done @haydar-c 

One minor comment and I think this is good to go!

vpr/src/analytical_place/gen_ap_netlist_from_atoms.h
vaughnb-cerebras: @haydar-c General rule: if it is smaller than or equal to the size of a pointer (64-bits) and you don't need to update it in the callee, pass by value. It avoids a pointer access to get it and hence will be faster (unless the compiler is clever enough to optimize out the pointer access, but I wouldn't count on that).

If it is bigger than 64-bit then pass by const ref if you don't need to modify it. Faster than a large copy.

If you need to modify it, no option but to pass by (non-const) reference.

AlexandreSinger (approved):
LGTM

vpr/src/analytical_place/gen_ap_netlist_from_atoms.h
haydar-c: Thanks Vaughn!

PR #3141, opened 2025-06-15
[AP][GP] Added Post-Global Placement Wirelength Estimation

AlexandreSinger:
After a recent change where we are ignoring high-fanout nets in the global placer, this caused the high-fanout nets to also be ignored during wirlength estimation in the global placement.

This wirelength is used in the global placer to decide when to stop and to print logs to the user. I would like to continue to ignore these nets when computing the objective of the global placer since it is the wirelength of the nets which are not ignored; however, for the user, I would like them to see a more accurate estimation of the wirelength.

Created a function which tries to estimate the wirelength of the global placement solution assuming all blocks can be placed exactly where their flat placement wants them to be placed.

See https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3137 for context.


haydar-c:
Thank you @AlexandreSinger for adding this! I like the distinction between the objective HPWL and estimated wire length. LGTM!

haydar-c (approved):

PR #3148, opened 2025-06-17
[AP] Generalized Argument Parsing and Added Target Density

AlexandreSinger:
The `appack_max_dist_th` cli option has a useful regex based interface which can be used in other parts of VPR, especially in AP. Generalized the interface so it can be used outside of the context of max distance thresholding.

Used this general interface to pass in a target density per physical tile type. This spoofs the capacity into being smaller than it appears which tells the partial legalizer to not fill the tiles beyond the capacity.

Setting the target density to interesting values revealed a bug in the partial legalizer regarding numerical precision when computing the prefix sum. Changed the underlying data type within the prefix sum class used into fixed-point which resolved the issue.


AlexandreSinger:
This did not change quality of results majorly. Some circuits minorly changed, but this resolved some hidden issues within the partial legalizer.

I did a quick sweep on VTR and found that setting the CLB target density to 0.8 improved post-FL HPWL by 9.8% and post-FL CPD by 2.7%; however basically a tie for post-routed HPWL adn CPD

Trying a 0.8 target density for LABs on Titan achieved similar results. I think more tuning is needed before we pull the trigger on a specific target density. For now, this option will make performing sweeps in the future much easier.

doc/src/vpr/command_line_usage.rst
```diff
@@ -1281,6 +1281,33 @@ Analytical Placement is generally split into three stages:
 
     **Default:** ``0.5``
 
+.. option:: --ap_partial_legalizer_target_density { auto | <regex>:<float>,<float> }
```
amin1377, line 1284: When you say "target density," do you mean the number of blocks packed into the cluster relative to its capacity, or are you referring to external pin density? My guess is that, since there are two floating-point values, you're referring to input/output pin density, but I'm not entirely sure.

vpr/src/analytical_place/flat_placement_density_manager.cpp
```diff
@@ -45,13 +49,68 @@ static PrimitiveVector calc_bin_underfill(const PrimitiveVector& bin_utilization
...
+ *
+ * This will automatically select good target densisities, but will allow the
+ * user to override these values from the command line.
+ *
+ *  @param target_density_arg_strs
+ *      The command-line arguments provided by the user.
+ *  @param physical_tile_types
+ *      A vector of all physical tile types in the architecture.
+ */
+static std::vector<float> get_phy_ty_target_densities(const std::vector<std::string>& target_density_arg_strs,
```
amin1377: In other functions, for example, in arch_utils, we use physical_type in the function name instead of abbreviating it as ph_ty. I'm not sure why you decided to shorten it here, but it might be harder to interpret for someone reading the code for the first time.

vpr/src/analytical_place/partial_legalizer.h
```diff
@@ -308,6 +308,16 @@ struct PartitionedWindow {
  *        regions.
  */
 class PerPrimitiveDimPrefixSum2D {
+    // This class stores the prefix sum as a fixed-point number instead of a
+    // floating point number. This is to prevent error accumulation which can
+    // occur due to numerical imprecisions. This variable selects how many bits
+    // the fractional component of the fixed-point number should have.
+    static constexpr unsigned num_fractional_bits_ = 10;
```
amin1377, line 315: I think this might be beyond the scope of this PR, but it could be better to create a separate class for fixed-point numbers (perhaps under vpt_types?) rather than handling it within this class. If you agree, please create an issue for it so we can track it.

AlexandreSinger: Issue has been raised: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3153

doc/src/vpr/command_line_usage.rst
AlexandreSinger: Added the following paragraph: 
image

AlexandreSinger:
Thanks for the review @amin1377 , please let me know if you have any further comments!

PR #3138, opened 2025-06-12
Change QoR parsing configs to use the fixed channel width variant

AmirhosseinPoolad:
Changes the parse_file attribute of tests that use fixed channel widths instead of doing the minimum channel width search to use vpr_fixed_chan_width.txt instead of vpr_standard.txt

PR #3149, opened 2025-06-18
Regenerate nightly test golden results

AmirhosseinPoolad:
This PR regenerates all the golden results for vtr_reg_nightly_test1-7. To summarize, the reasons for doing this would be:

- There were some unimportant failures on a limited number of circuits that needed new golden results
- The way some test results were being parsed was wrong. #3138 Fixed that but necessitated new golden results
- The new[ly added to the tasklist] vtr_xilinx_qor test in vtr_reg_nightly_test2 had old and invalid golden results
- Yosys upgrade which also changed a lot of results

Since we have a mostly clean slate otherwise, regenerating golden results should be safe. One potential issue is that this would leave the following commits untested:

image

Given that these commits are either things we want to do anyway e.g. yosys upgrade or they don't change VTR functionality in any way, I would not be too concerned about it.


AlexandreSinger (approved):
LGTM, thank you so much for doing this Amir! This is somewhat healthy to do every so often so long as we verify that the large circuits that we care about do not regress in quality (which Logan already did when updating the Yosys version).

PR #3151, opened 2025-06-18
Fixed Route Drawing Function

SamuelHo10:
When selecting the clustered nets option in the GUI to visualize routing, certain lines and connections were not being drawn. The code was looping through each net depth first and traversing through each node once. However, when the tree branches, the connection between the branched node and the main tree was skipped in the original code. I fixed this by checking when a new branch is being drawn, and adding the parent node to the vector.

| Old| New |
| ----------- | ----------- |
| image| image       |


vpr/src/draw/draw_basic.cpp
```diff
@@ -593,6 +593,10 @@ void draw_routed_net(ParentNetId net_id, ezgl::renderer* g) {
             draw_state->draw_rr_node[inode].color = DEFAULT_RR_NODE_COLOR;
         }
 
+        if(rr_nodes_to_draw.empty() && rt_node.parent().has_value()) {
```
soheilshahrouz: It's not clear from the code why this conition needs to be check. Please add a comment explining why this is needed.

soheilshahrouz (changes requested):
Thanks for this PR!

C++ formatting fails. You can run `make format` to make the code consistent with coding style rules.

soheilshahrouz (approved):
Thanks for adding the comment!

PR #3152, opened 2025-06-19
[AP][Solver] Ignored Disconnected Blocks in AP Solver

AlexandreSinger:
After investigating some of the slowest running testcases, I realized that we were not handling disconnected blocks in the solver.

Especially after we started thresholding out high-fanout nets, some circuits were taking far longer to solve than they should. They especially took a long time to set up the matrices. After investigating, I realized that there were many blocks which we completely disconnected from the rest of the circuit. There is no reason to optimize the location of these blocks since the AP objective is formulated based on net connectivity. As such, these disconnected blocks should be completely ignored during placement.

Ignoring these blocks reduces the number of variables in the A matrix, which can greatly improve runtime. Early results on Titan show up to a 3.5x improvement in GP runtime and a 20% improvement in GP runtime on average.

Future work is to be more methodical on what nets to mark as ignored. The AP flow currently does not directly set signals like clocks as ignored, which may be able to allow us to label more blocks as disconnected.

### DIFF
--- a/vpr/src/analytical_place/analytical_solver.h
+++ b/vpr/src/analytical_place/analytical_solver.h
@@ -63,6 +63,7 @@ class AnalyticalSolver {
     AnalyticalSolver(const APNetlist& netlist,
                      const AtomNetlist& atom_netlist,
                      const PreClusterTimingManager& pre_cluster_timing_manager,
+                     const DeviceGrid& device_grid,
                      float ap_timing_tradeoff,
                      int log_verbosity);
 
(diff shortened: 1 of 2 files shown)


vpr/src/analytical_place/analytical_solver.h
```diff
@@ -136,6 +137,21 @@ class AnalyticalSolver {
...
+    /// @brief A vector of blocks in the netlist which are not connected to any
+    ///        non-ignored nets. Since the blocks have no connection to any
+    ///        other blocks in the netlist, the are not considered movable or
+    ///        fixed. The solver will just leave these blocks wherever they were
+    ///        legalized.
+    std::vector<APBlockId> disconnected_blocks_;
+
+    /// @brief The width of the device grid. Used for randomly generating points
+    ///        on the grid.
+    size_t device_grid_width_;
```
vaughnbetz, line 149: Why do you keep cached copies of the grid width and height in the solver instead of asking the grid for width and height?

vaughnbetz (commented):
LGTM but see one question on width/height

vpr/src/analytical_place/analytical_solver.h
AlexandreSinger: During Global Placement, the device size is assumed to be fixed. We also only really use this information in the first iteration. I preferred to just pass the device size in instead of holding on to a reference to the device itself.

PR #3154, opened 2025-06-19
[AP] Marking Global Nets in AP

AlexandreSinger:
After investigating how the placer and packer decide which nets are global, I realized that we were not marking any nets as global. The Atom Netlist annotates nets as global after packing, when loading the clustered netlist. It decides if a net is global or not if the net enters a global port in the clustered netlist.

We can replicate this behaviour during AP by looking at the Atom Netlist and doing something similar. This allows us to tell if a net is global or not.

In AP, we always want to ignore global nets since their cost does not play a role on the overall objective. By ignoring these nets, we can further improve the performance and quality of the solution.

Now that we know which nets are global, I have also updated the wirelength estimator in the AP flow to make it align better with the placer's wirelength estimator. They now line up very closely, but are still slightly off due to some factors being unable to be ported over easily.


vpr/src/analytical_place/partial_placement.cpp
```diff
@@ -38,23 +39,23 @@ double PartialPlacement::estimate_post_placement_wirelength(const APNetlist& net
...
-        double max_y = std::numeric_limits<unsigned>::lowest();
+        // To align with other wirelength estimators in VTR (for example in the
+        // placer), we do not include global nets (clocks, etc.) in the wirelength
+        // calculation.
+        if (netlist.net_is_global(net_id))
+            continue;
+
+        // Similar to the placer, weight the wirelength of this net as a function
+        // of its fanout.
+        // TODO: Should update the costs to 3D.
```
vaughnbetz, line 60: We should comment that these fanouts are at the AP netlist (unclustered) level so the correction factors will lead to a somewhat higher hpwl prediction than after clustering.

vpr/src/analytical_place/partial_placement.cpp
```diff
@@ -66,10 +67,13 @@ double PartialPlacement::estimate_post_placement_wirelength(const APNetlist& net
 
         // Floor the positions to get the x and y coordinates of the tiles each
         // block belongs to.
-        unsigned tile_dx = std::floor(max_x) - std::floor(min_x);
-        unsigned tile_dy = std::floor(max_y) - std::floor(min_y);
+        // TODO: In the placer, the x and y dimensions are multiplied by cost
+        //       factors based on the channel width. Should somehow bring these
+        //       in here.
```
vaughnbetz: Probably don't want to bring those into a hpwl estimate but could bring them into a cost estimate.

vaughnbetz (approved):

PR #3155, opened 2025-06-20
[AP][Solver] Per-Connection Timing Optimization

AlexandreSinger:
The B2B solver originally used a simple net weighting model to incorperate timing into the objective. This works well, but a more direct approach to timing should provide a better solution.

Added extra edges to the B2B formulation which encode per-connection delays into the objective.

Simply, these edges incentivice the solver to shrink the distance between driver and sinks where shrinking that distance would improve the delay the most (focusing more on currently critical edges).

To encode this objective, I compute the instantaneuous derivative of the delay of an connection with respect to the distance (normalized such that the units are not in seconds) and use that as a weight for a pin to pin connection between each driver and its sinks.

More tuning is likely need to find good factors for each term. For now this was what worked well on the VTR benchmarks.

### DIFF
--- a/vpr/src/analytical_place/analytical_solver.cpp
+++ b/vpr/src/analytical_place/analytical_solver.cpp
@@ -21,8 +21,11 @@
 #include "flat_placement_types.h"
 #include "partial_placement.h"
 #include "ap_netlist.h"
+#include "place_delay_model.h"
+#include "timing_info.h"
 #include "vpr_error.h"
 #include "vtr_assert.h"
+#include "vtr_math.h"
 #include "vtr_time.h"
 #include "vtr_vector.h"
 
@@ -49,6 +52,7 @@ std::unique_ptr<AnalyticalSolver> make_analytical_solver(e_ap_analytical_solver 
                                                          const DeviceGrid& device_grid,
                                                          const AtomNetlist& atom_netlist,
                                                          const PreClusterTimingManager& pre_cluster_timing_manager,
+                                                         std::shared_ptr<PlaceDelayModel> place_delay_model,
                                                          float ap_timing_tradeoff,
                                                          unsigned num_threads,
                                                          int log_verbosity) {
@@ -81,6 +85,7 @@ std::unique_ptr<AnalyticalSolver> make_analytical_solver(e_ap_analytical_solver 
             (void)device_grid;
             (void)atom_netlist;
             (void)pre_cluster_timing_manager;
+            (void)place_delay_model;
             (void)ap_timing_tradeoff;
             (void)log_verbosity;
             VPR_FATAL_ERROR(VPR_ERROR_AP,
@@ -93,6 +98,7 @@ std::unique_ptr<AnalyticalSolver> make_analytical_solver(e_ap_analytical_solver 
                                                device_grid,
                                                atom_netlist,
                                                pre_cluster_timing_manager,
+                                               place_delay_model,
                                                ap_timing_tradeoff,
                                                log_verbosity);
 #else
@@ -110,7 +116,6 @@ std::unique_ptr<AnalyticalSolver> make_analytical_solver(e_ap_analytical_solver 
 
 AnalyticalSolver::AnalyticalSolver(const APNetlist& netlist,
                                    const AtomNetlist& atom_netlist,
-                                   const PreClusterTimingManager& pre_cluster_timing_manager,
                                    const DeviceGrid& device_grid,
                                    float ap_timing_tradeoff,
                                    int log_verbosity)
@@ -159,40 +164,6 @@ AnalyticalSolver::AnalyticalSolver(const APNetlist& netlist,
         row_id_to_blk_id_[new_row_id] = blk_id;
         current_row_id++;
         num_moveable_blocks_++;
-    }
-
-    update_net_weights(pre_cluster_timing_manager);
-}
-
-void AnalyticalSolver::update_net_weights(const PreClusterTimingManager& pre_cluster_timing_manager) {
-    // If the pre-cluster timing manager has not been initialized (i.e. timing
-    // analysis is off), no need to update.
-    if (!pre_cluster_timing_manager.is_valid())
-        return;
-
-    // For each of the nets, update the net weights.
-    for (APNetId net_id : netlist_.nets()) {
-        // Note: To save time, we do not compute the weights of nets that we
-        //       do not care about for AP. This leaves their weights at 1.0 just
-        //       in case they are accidentally used.
-        if (netlist_.net_is_ignored(net_id))
-            continue;
-
-        AtomNetId atom_net_id = netlist_.net_atom_net(net_id);
-        VTR_ASSERT_SAFE(atom_net_id.is_valid());
-
-        float crit = pre_cluster_timing_manager.calc_net_setup_criticality(atom_net_id, atom_netlist_);
-
-        // When optimizing for WL, the net weights are just set to 1 (meaning
-        // that we want to minimize the WL of nets).
-        // When optimizing for timing, the net weights are set to the timing
-        // criticality, which is based on the lowest slack of any edge belonging
-        // to this net.
-        // The intuition is that we care more about shrinking the wirelength of
-        // more critical connections than less critical ones.
-        // Use the AP timing trade-off term to linearly interpolate between these
-        // weighting terms.
-        net_weights_[net_id] = ap_timing_tradeoff_ * crit + (1.0f - ap_timing_tradeoff_);
     }
 }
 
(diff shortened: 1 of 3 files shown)


AlexandreSinger:
Result on Titan, timing driven with no fixed blocks:

| Metric | Value normalized to AP baseline |
| ------ | --------------------------------- |
| post_fl_hpwl | 1.04 |
| post_dp_hpwl | 1.00 |
| total_wirelength | 1.00 |
| post_gp_cpd | 1.01 |
| post_fl_cpd | 1.02 |
| post_dp_cpd | 0.98 |
| crit_path_delay | 0.98 |

Wintermute was loaded yesterday, so I do not really trust the runtime results; but locally, I have found that GP runtime increased by around 10-20%.

These results match what I have been seeing on VTR. The post-GP and post-FL results appear to get a bit worse; however overall the result gets better. I believe what is causing this behavior is that the timing analysis and wirelength analysis I am performing during GP is so early that it is more of a guide than truth. Since my timing objective that I added is focused more on the router lookahead than the criticalities that I compute, it appears to do worse in GP but actually does better overall.

Much more tuning is needed. but I think this is good enough to replace the net weighting approach for now. I want to continue to tune and prod this over time.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -850,7 +856,180 @@ void B2BSolver::add_connection_to_system(APBlockId first_blk_id,
...
+    // be equal in magnitude and opposite. This means the central difference will
+    // be 0. This is not good since it would cause the objective to ignore the
+    // delay of blocks within the same cluster (making them more incentivized to
+    // not be in the same cluster together). To prevent this, we set the derivative
+    // to be the forward difference in that case. It must be the forward difference
+    // since the backward difference will likely be negative. This basically sets
+    // the derivative to be the penalty for putting the driver and sink in different
+    // tiles.
+    if (tile_dx == 0) {
+        VTR_ASSERT_SAFE_MSG(forward_difference_x == -1.0f * backward_difference_x,
```
vaughnbetz: This is a bit of a dangerous assert.  Currently true given the way that we compute the placement delay model, but it could change in the future (it's not fundamental).  Not sure if you should keep the insert in that case .... it could fire in the future for no good reason.  So if you don't really need it, I suggest removing it.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -850,7 +856,180 @@ void B2BSolver::add_connection_to_system(APBlockId first_blk_id,
...
+    return std::make_pair(d_delay_x, d_delay_y);
+}
+
+std::pair<double, double> B2BSolver::get_delay_normalization_facs(APBlockId driver_blk) {
+    // We want to find normalization factors for the delays along connections.
+    // A simple normalization factor to use is 1 over the delay of leaving a
+    // tile. This should be able to remove the units without changing the value
+    // too much.
+
+    // Similar to calcuting the derivative, we want to use the legalized position
```
vaughnbetz: calculating

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -871,7 +1050,10 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement) {
         size_t num_pins = netlist_.net_pins(net_id).size();
         VTR_ASSERT_SAFE_MSG(num_pins > 1, "net must have at least 2 pins");
 
-        double net_w = net_weights_[net_id];
+        // ====================================================================
+        // Wirelength Connections
+        // ====================================================================
+        double wl_net_w = (1.0f - ap_timing_tradeoff_) * net_weights_[net_id];
```
vaughnbetz, line 1057: Line 529 already adjusted the net_weights array with ap_timing_tradeoff.  Should it be adjusted again here?  I don't quite get the desired interaction between the two lines.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -883,19 +1065,85 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement) {
...
+        add_connection_to_system(net_bounds.max_x_blk, net_bounds.min_x_blk, num_pins, wl_net_w, p_placement.block_x_locs, triplet_list_x, b_x);
+        add_connection_to_system(net_bounds.max_y_blk, net_bounds.min_y_blk, num_pins, wl_net_w, p_placement.block_y_locs, triplet_list_y, b_y);
+
+        // ====================================================================
+        // Timing Connections
+        // ====================================================================
+        // Only add timing connection if timing analysis is on and we are not
+        // in the first iteration. The current timing flow needs legalized
+        // positions to compute the delay derivative, which do not exist until
+        // the next iteration. Its fine to do one wirelength driven iteration first.
```
vaughnbetz, line 1089: Its -> It's

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -883,19 +1065,85 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement) {
...
+        if (pre_cluster_timing_manager_.is_valid() && iteration != 0) {
+            // Create connections from each driver pin to each of its sink pins.
+            // This will incentivize shrinking the distance from drivers to sinks
+            // of connections which would improve the timing.
+            APPinId driver_pin = netlist_.net_driver(net_id);
+            APBlockId driver_blk = netlist_.pin_block(driver_pin);
+            for (APPinId net_pin : netlist_.net_pins(net_id)) {
+                if (net_pin == driver_pin)
```
vaughnbetz: Cleaner and slightly faster to just use .net_sinks(net_id) instead (removes the continue). See https://docs.verilogtorouting.org/en/latest/api/vpr/netlist/#_CPPv4NK7Netlist9net_sinksEK5NetId

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -883,19 +1065,85 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement) {
...
+
+                // Since the delay between two blocks may not monotonically increase
+                // (it may go down with distance due to different length wires), it
+                // is possible for the derivative of delay to be negative. The weight
+                // terms in this formulation should not be negative to prevent infinite
+                // answers. To prevent this, clamp the derivative to 0.
+                // TODO: If this is negative, it means that the sink should try to move
+                //       away from the driver. Perhaps add an anchor point to pull the
+                //       sink away.
+                if (d_delay_x < 0)
```
vaughnbetz: could just use std::max with 0
Not sure if it will save time to use 0. or not (would avoid a conversion if the compiler didn't convert at compile time).

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -883,19 +1065,85 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement) {
...
+                    d_delay_x = 0;
+                if (d_delay_y < 0)
+                    d_delay_y = 0;
+
+                // The units for delay is in seconds; however the units for
```
vaughnbetz: is -> are

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -883,19 +1065,85 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement) {
...
+                // the wirelength term is in tile. To ensure the units match,
```
vaughnbetz: are in tiles

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -883,19 +1065,85 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement) {
...
+                // Get the criticality of this timing edge from driver to sink.
+                double crit = pre_cluster_timing_manager_.get_timing_info().setup_pin_criticality(netlist_.pin_atom_pin(net_pin));
+
+                // Set the weight of the connection from driver to sink equal to:
+                //      weight_tradeoff_terms * (1 + crit) * d_delay * delay_norm
+                // The intuition is that we want the solver to shrink the distance
+                // from drivers to sinks (which would improve timing) for edges
+                // with the best tradeoff between delay and wire, with a focus
+                // on the more critical edges.
+                double timing_net_w = ap_timing_tradeoff_ * net_weights_[net_id] * timing_slope_fac_ * (1.0 + crit);
```
vaughnbetz, line 1136: I think net_weights have also already been adjusted for timing earlier, if the quadratic objective is used.  Is that what you want (double-adjustment?).  Some comment seems appropriate on that anyway.

vpr/src/analytical_place/analytical_solver.h
```diff
@@ -569,6 +597,38 @@ class B2BSolver : public AnalyticalSolver {
                                   std::vector<Eigen::Triplet<double>>& triplet_list,
                                   Eigen::VectorXd& b);
 
+    /**
+     * @brief Get the instantaneous derivative of delay for the given driver
+     *        and sink pair.
+     *
+     * The instantaneous derivative gives the amount delay would increase or
```
vaughnbetz, line 604: for a change in distance *by one unit*.  (I think that's the real definition).

vpr/src/analytical_place/analytical_solver.h
```diff
@@ -642,6 +702,14 @@ class B2BSolver : public AnalyticalSolver {
     ///        loop so far. This includes creating the CG solver object and
     ///        actually solving for a solution.
     float total_time_spent_solving_linear_system_ = 0.0f;
+
+    /// @brief Timing manager object used for calculating the criticality of
+    ///        edges in the graph.
+    const PreClusterTimingManager& pre_cluster_timing_manager_;
+
+    /// @breif The place delay model used for calculating the delay between
```
vaughnbetz: @brief

vpr/src/analytical_place/analytical_solver.h
```diff
@@ -642,6 +702,14 @@ class B2BSolver : public AnalyticalSolver {
...
+    ///        to tiles on the FPGA. Used for computing the timing terms.
```
vaughnbetz: to -> two

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: The base solver class sets all of the weights to 1 by default. Each child class will implement their own net weighting function. For the quadratic solver I left it with the criticality-based net weighting approach from before (it takes some work to bring what I have here into the quadratic formulation due to the matrices not being recomputed). The B2B solver currently leaves these weights as all 1s, but in the future I hope to investigate also adding criticality to the mix here as well.

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: The code you are looking at here is in the B2B solver and only the B2B solver. I have not thought about generalizing this to all potential solver types mainly because this is actually the inner-loop of the B2B solver (where it reconstructs the matrix with the new positions of blocks for the linearization terms). I will try to add comments to make this more clear.

PR #3156, opened 2025-06-20
Converting Yosys to Submodule

loglav03:
#### Description
Converted Yosys from vendored in subdirectory to submodule. Yosys submodule points to version 0.53.
**Important Notes About Yosys Submodule:**

- Yosys now lives in libs/EXTERNAL instead of in the VTR root.
- Points to Yosys version 0.53.
- The CmakeLists in libs/EXTERNAL does the process for initializing and updating submodules within yosys, and building yosys.

#### Motivation and Context

Converting yosys to a submodule makes it easier to update yosys to a newer version if needed. It also makes it easier to implement yosys-slang to replace synlig/surelog.

Ran the required CI tests -> ALL PASSED (with updated golden results)

#### Types of changes

#### Checklist:


AlexandreSinger:
Thank you so much @loglav03 !

I am going to frame this and put this on my wall for the reason why submodules are necessary:
image

I have a few comments, but I cannot put them in the standard review spots since your PR is so large that it crashes the GitHub UI for me...

Please resolve the following and it looks good to me!

- [ ] `CMakeLists.txt` : Please delete the line you commented out. Please also delete everything else in that if statement. I believe it goes unused; however, be careful with the set(MAKE_PROGRAM) line. I think that is supposed to be in libs/EXTERNAL/CmakeLists.txt with your new changes. That define for "-D_YOSYS_" is interesting, I am not sure who is using this; but if it is used by parmys it should be moved to the if statement below that adds the parmys directory, if its used by YOSYS it should be moved to the new section you added; however I am not sure.
- [ ] `dev/subtree_config.xml` : Now that you have made Yosys a submodule, please remove it from the subtree config. This is as simple as just deleting the following lines: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/7840430dde3c3d48d903c660f8bdcdb69e558cfd/dev/subtree_config.xml#L52-L56
- [ ] `libs/EXTERNAL/CMakeLists.txt` : What you added here is very interesting. It builds Yosys from our repo instead of adding its CMakeLists (which is a good choice in my opinion; their CMakeLists would interfere with our build system I think). If you got this from somewhere else, can you please add a comment with a link to where you got it. The consequence of this is that now it looks like we have two builds happening simultaneously:
image

I do **not** think this is a problem. From prior investigations into our build system, I have found that two builds were happening simultaneously for a long time; it is just now being more explicit about it. I do not want this to block your PR; however can you please raise an issue about this so we can track it. We really should not have two builds happening simultaneously as it can throttle machines when multiple cores are used to compile. I am not sure if the "MAKE_PROGRAM" variable is the culprit for this, but I am pretty sure this is not an easy problem to fix.

I like this PR a lot and want to see it in ASAP, so please just resolve the three comments above and we can merge this in. Thank you so much for doing this.

AlexandreSinger:
@vaughnbetz LGTM!

Issue #2729, opened 2024-09-20
[Packer] Setting Higher Target Pin Utilization When Regions are Full


AlexandreSinger:
The packer currently has a fallback where, if it could not pack after 4 iterations, it will increase the target pin utilization of the clusters to give it a better chance of finding a valid clustering (at the expense of runtime):
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/e38c0f3d975decbafc112d18877a99506c7e845f/vpr/src/pack/pack.cpp#L195-L208

As the TODO reads, this was implemented in a very hard-coded way. It assumes the logic block type which was having issues is the clbs and the name is "clb".

This should either increase the target pin utilization for the overused block types or it should just increase the target pin utilization for all logical block types.

PR #3158, opened 2025-06-21
[Arch] Added Zero ASIC's Z1000 eFPGA Architecture

AlexandreSinger:
Zero ASIC uses VPR as part of its open-source FPGA CAD flow (Logik). Because of this, its architecture XML file is also open-source.

This brings in Zero ASIC's Z1000 architecture. A 2048 4-LUT/FF architecture with 1024 GPIOs and can support up to 4 unique clock domains.

I had to slightly modify the architecture file since the Z1000 uses custom register primitives which its synthesis front-end can target instead of regular ".latch" primitives. To get around this, I added a custom mode to the BLE to use the ".latch" primitive instead of the register primitives. Since the pins for the ".latch" primitive and the register primitives have one-to-one correspondance, this should have no affect on VPR's execution.

In order for these circuits to route properly, the z1000 RR graph must also be provided. Included these as zip files and added a make command which will extract the RR graphs. Zipped, this file is around 2 MB, unzipped it is 90 MB.

Added a tasks to NightlyTest7 which runs the MCNC benchmarks that could fit on this device. There is currently an issue in the packer causing it to give up too early and not pack as much as it can; the circuits affected by this bug have been commented out and will be uncommented once that bug is fixed.


AlexandreSinger:
More details about the architecture can be found here: https://github.com/siliconcompiler/logiklib/tree/main/logiklib/zeroasic/z1000

image

image

AlexandreSinger:
The bug in the packer is described in issue https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2729

Once that is resolved, much more circuits can be run on this architecture. The target external pin utilization of {0.8, 0.6} is far too strict for this architecture and causes it to be unable to pack more than around 50/60% of the device. The packer should relax the target pin utilization if it fails to pack, but it is currently not doing that correctly.

vaughnbetz (approved):
Looks great -- thanks for adding this arch and these tests!

PR #3159, opened 2025-06-22
Flat Routing Visualization

SamuelHo10:
Several changes were made for flat routing visualization.

Firstly, I modified the draw intra-logic block code so that the margins between blocks are absolute and not based on relative size. The intra-logic blocks are also now drawn to be not long and skinny, and instead must have a reasonable width-to-height ratio. Furthermore, the intra-logic blocks of maximum depth were previously not drawn. That has also been fixed.

To get the pin locations of intracluster RRNodeId, I made helper functions to convert RRNodeId into cluster_blk_id and pb_graph_pin. I also separated the code which draws edges and pins, primarily for readability and also to avoid repetition. The loops with clusternets were replaced with atomnets when flat routing is enabled. Lastly, for testing and debugging purposes, the ability to click on intracluster pins to highlight the net was added.


vpr/src/draw/draw_rr.h
```diff
@@ -27,6 +27,7 @@ void draw_rr_edges(RRNodeId from_node, ezgl::renderer* g);
 
 void draw_rr_chan(RRNodeId inode, const ezgl::color color, ezgl::renderer* g);
 
+void draw_rr_intrapin(RRNodeId inode, const ezgl::color& color, ezgl::renderer* g);
```
vaughnbetz: Add a comment describing what the function does.  I think it would be best to make it a Doxygen comment so we can choose to build Doxygen comments for more drawing code in the future.

General principle: comment what a function does in the .h file, or (if it is a static function) in its declaration in the .cpp file.  Use Doxygen formatting in either case.

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
 
-    static vtr::OffsetMatrix<int> chanx_track; /* [1..device_ctx.grid.width() - 2][0..device_ctx.grid.height() - 2] */
-    static vtr::OffsetMatrix<int> chany_track; /* [0..device_ctx.grid.width() - 2][1..device_ctx.grid.height() - 2] */
-    if (draw_state->draw_route_type == e_route_type::GLOBAL) {
-        /* Allocate some temporary storage if it's not already available. */
-        int width = (int)device_ctx.grid.width();
-        int height = (int)device_ctx.grid.height();
-        if (chanx_track.empty()) {
-            chanx_track = vtr::OffsetMatrix<int>({{{1, width - 1}, {0, height - 1}}});
+    // Draw Pins
```
vaughnbetz: Isn't this drawing rr nodes (not just pins)?  Suggest changing the comment.

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
+        // Draw intra-cluster nodes
+        if (!is_inode_inter_cluster) {
+            draw_rr_intrapin(inode, color, g);
+            continue;
         }
 
-        for (int i = 1; i < width - 1; i++)
-            for (int j = 0; j < height - 1; j++)
-                chanx_track[i][j] = (-1);
+        // Draw IO Pins
```
vaughnbetz: Draw cluster-level IO pins

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
 
-        if (!is_inter_cluster_node(rr_graph, prev_node) || !is_inter_cluster_node(rr_graph, inode)) {
+        // For 3D architectures, draw only visible layers
+        if (!draw_state->draw_layer_display[current_node_layer].visible || !edge_visibility.visible) {
             continue;
         }
 
-        auto iedge = find_edge(prev_node, inode);
-        auto switch_type = rr_graph.edge_switch(RRNodeId(prev_node), iedge);
+        // Skip drawing sources and sinks
```
vaughnbetz: Skip drawing edges to or from sources and sinks

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
             case e_rr_type::IPIN: {
-                draw_rr_pin(inode, color, g);
-                if (edge_visibility.visible) {
-                    g->set_color(color, edge_visibility.alpha);
-                    if (rr_graph.node_type(prev_node) == e_rr_type::OPIN) {
-                        draw_pin_to_pin(prev_node, inode, g);
-                    } else {
-                        draw_pin_to_chan_edge(inode, prev_node, g);
-                    }
+                if (rr_graph.node_type(prev_node) == e_rr_type::OPIN) {
```
vaughnbetz: To match the structure of the rest of the code and shorten this routine I think this switch statement should be moved into a helper router (maybe called draw_inter_cluster_rr_edge )

amin1377:
@SamuelHo10:

Thank you for your changes! Could you please share some screenshots showing how the intra-logic connections are drawn? Also, it would be helpful if you could include a screenshot for the case when the flat-router is not enabled, just to ensure the default flow remains intact.

I’m currently testing this PR locally, but having the screenshots here would be useful for future reference.

vpr/src/draw/draw_rr.cpp
```diff
@@ -683,6 +703,30 @@ RRNodeId draw_check_rr_node_hit(float click_x, float click_y) {
...
+            ezgl::point2d p = draw_coords->get_absolute_pin_location(blk_id_pin_id.first, blk_id_pin_id.second);
+
+            if (click_x >= p.x - draw_coords->pin_size && click_x <= p.x + draw_coords->pin_size && click_y >= p.y - draw_coords->pin_size && click_y <= p.y + draw_coords->pin_size) {
+                hit_node = inode;
+                return hit_node;
+            }
+
+            continue;
+        }
         switch (rr_graph.node_type(inode)) {
```
vaughnbetz, line 732: I'd add a comment here: check for inter-cluster nodes

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -279,6 +279,65 @@ void draw_chanx_to_chany_edge(RRNodeId chanx_node, RRNodeId chany_node, enum e_e
...
+    t_draw_coords* draw_coords = get_draw_coords_vars();
+
+    if (!draw_state->is_flat) {
+        return;
+    }
+
+    auto blk_id_pin_id1 = get_rr_node_cluster_blk_id_pb_graph_pin(inode);
+    auto blk_id_pin_id2 = get_rr_node_cluster_blk_id_pb_graph_pin(prev_node);
+
+    ezgl::point2d p2 = draw_coords->get_absolute_pin_location(blk_id_pin_id1.first, blk_id_pin_id1.second);
```
vaughnbetz: What do the .first and .second represent?  I think it would be more clear to use a structure with named elements instead of just a pair (which has essentially unnamed elements).

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -279,6 +279,65 @@ void draw_chanx_to_chany_edge(RRNodeId chanx_node, RRNodeId chany_node, enum e_e
...
+void draw_intrapin_to_intrapin(RRNodeId inode, RRNodeId prev_node, ezgl::renderer* g) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    t_draw_coords* draw_coords = get_draw_coords_vars();
+
+    if (!draw_state->is_flat) {
+        return;
+    }
+
+    auto blk_id_pin_id1 = get_rr_node_cluster_blk_id_pb_graph_pin(inode);
+    auto blk_id_pin_id2 = get_rr_node_cluster_blk_id_pb_graph_pin(prev_node);
```
vaughnbetz: _id and _id_prev would be more meaningful than _id1 and _id2 (you've lost which is the previous and current in that case).

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -279,6 +279,65 @@ void draw_chanx_to_chany_edge(RRNodeId chanx_node, RRNodeId chany_node, enum e_e
...
+
+    ezgl::point2d p2 = draw_coords->get_absolute_pin_location(blk_id_pin_id1.first, blk_id_pin_id1.second);
+    ezgl::point2d p1 = draw_coords->get_absolute_pin_location(blk_id_pin_id2.first, blk_id_pin_id2.second);
+
+    g->draw_line(p1, p2);
+
+    float xend = p2.x + (p1.x - p2.x) / 10.;
+    float yend = p2.y + (p1.y - p2.y) / 10.;
```
vaughnbetz: Names are confusing.
prev->id2->p1.  Change them all to end / start or from/to.

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -279,6 +279,65 @@ void draw_chanx_to_chany_edge(RRNodeId chanx_node, RRNodeId chany_node, enum e_e
...
+void draw_intrapin_to_pin(RRNodeId inode, RRNodeId prev_node, ezgl::renderer* g) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    t_draw_coords* draw_coords = get_draw_coords_vars();
+
+    if (!draw_state->is_flat) {
+        return;
+    }
+    const auto& rr_graph = g_vpr_ctx.device().rr_graph;
+
+    bool swapped = false;
```
vaughnbetz: I think this is confusing.  I'd just have a bit of code that gets the p_start and p_end with a small helper function, and then the rest of the code can be straight line code.  Again, using 1, 2 is not very clear for something with a start and end.  Use more direct names.  Swapping the variables makes the code harder to read by obscuring what is what.

vpr/src/draw/draw_rr_edges.h
```diff
@@ -20,6 +20,8 @@
 void draw_chany_to_chany_edge(RRNodeId from_node, RRNodeId to_node, short switch_type, ezgl::renderer* g);
 void draw_chanx_to_chanx_edge(RRNodeId from_node, RRNodeId to_node, short switch_type, ezgl::renderer* g);
 void draw_chanx_to_chany_edge(RRNodeId chanx_node, RRNodeId chany_node, enum e_edge_dir edge_dir, short switch_type, ezgl::renderer* g);
+void draw_intrapin_to_pin(RRNodeId inode, RRNodeId prev_node, ezgl::renderer* g);
```
vaughnbetz: Add Doxygen comments for these new functions.  If you know what some other routines do as well, this would be a good time to add Doxygen comments for them too (but that's up to you ... I don't want to swamp you!).

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -131,40 +131,59 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
...
+void highlight_net(char* message, RRNodeId hit_node) {
     auto& cluster_ctx = g_vpr_ctx.clustering();
+    auto& atom_ctx = g_vpr_ctx.atom();
     auto& route_ctx = g_vpr_ctx.routing();
 
     /* Don't crash if there's no routing */
     if (route_ctx.route_trees.empty())
         return;
 
+    if (route_ctx.is_flat) {
```
vaughnbetz, line 143: I think this could be simplified by using something like:

// declare nets as a reference of the type of .nets
if (route_ctx.is_flat) 
   nets = atom_ctx.netlist().nets()
else 
   nets = cluster_ctx.clb_nlist.nets()

// rest of code just uses nets

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -131,40 +131,59 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
...
-                            size_t(net_id),
-                            cluster_ctx.clb_nlist.net_name(net_id).c_str());
+    for (auto& rt_node : route_ctx.route_trees[parent_id].value().all_nodes()) {
+        RRNodeId inode = rt_node.inode;
+        if (draw_state->draw_rr_node[inode].color == ezgl::MAGENTA) {
+            draw_state->net_color[parent_id] = draw_state->draw_rr_node[inode].color;
+            if (inode == hit_node) {
+                std::string orig_msg(message);
+                std::string net_name;
+                if (!route_ctx.is_flat) {
```
vaughnbetz: I suggest taking this code and making it into a helper function: e.g. draw_get_net_name_string ()

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -131,40 +131,59 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
...
+                        size_t(parent_id),
+                        net_name.c_str());
             }
+        } else if (draw_state->draw_rr_node[inode].color
+                   == ezgl::WHITE) {
+            // If node is de-selected.
+            draw_state->net_color[parent_id] = ezgl::BLACK;
+            break;
         }
     }
```
vaughnbetz, line 187: Don't you still need the appliation->update_message() call to make this name show up in the status line at the bottom of the window?

vpr/src/draw/draw_searchbar.h
```diff
@@ -28,7 +28,10 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type, ClusterBlockId b
 
 /* If an rr_node has been clicked on, it will be highlighted in MAGENTA.
  * If so, and toggle nets is selected, highlight the whole net in that colour.*/
-void highlight_nets(char* message, RRNodeId hit_node, bool is_flat);
+void highlight_net(char* message, RRNodeId hit_node);
+
+/* Checks if a node is part of a net, and highlights the net if it is. */
+void check_node_highlight_net(char* message, ParentNetId parent_id, RRNodeId hit_node);
```
vaughnbetz: Make comment Doxygen format.  Document the functionality and each of the inputs as @param. parent_id isn't too obvious so explain it well.

I think this routine also updates (or diplays) message.  You should say that.
Why do we pass message in?  Are we updating it and returning it?  Or just displaying it on the screen?  If just displaying it, make message a local variable instead of passed in.

vpr/src/draw/draw_types.cpp
```diff
@@ -151,6 +151,22 @@ ezgl::rectangle t_draw_coords::get_absolute_pb_bbox(const ClusterBlockId clb_ind
     return result;
 }
 
+ezgl::point2d t_draw_coords::get_absolute_pin_location(const ClusterBlockId clb_index, const t_pb_graph_pin* pb_graph_pin) {
+
+    t_pb_graph_node* pb_gnode = pb_graph_pin->parent_node;
+    ezgl::rectangle pb_bbox = this->get_absolute_pb_bbox(clb_index, pb_gnode);
+    int num_pins = pb_gnode->num_pins();
+
```
vaughnbetz, line 159: Explain the logic of how you are dividing up the graphics area.  Explain the choice of 0.01 below.  Probably it should be a const float <meaningful_name> = 0.01 to make it more readable and easier to change.

vpr/src/draw/draw_types.h
```diff
@@ -432,6 +432,11 @@ struct t_draw_coords {
      */
     ezgl::rectangle get_absolute_pb_bbox(const ClusterBlockId clb_index, const t_pb_graph_node* pb_gnode);
 
+    /**
+     * @brief returns a 2D point for the absolute location of the given pb_graph_pin
+     */
```
vaughnbetz, line 446: Explain @param.
Explain anything the user should know about this location (pixel loc in the entire graphics area?)

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -38,7 +38,7 @@
 
 static void draw_internal_load_coords(int type_descrip_index, t_pb_graph_node* pb_graph_node, float parent_width, float parent_height);
 static int draw_internal_find_max_lvl(const t_pb_type& pb_type);
-static void draw_internal_calc_coords(int type_descrip_index, t_pb_graph_node* pb_graph_node, int num_pb_types, int type_index, int num_pb, int pb_index, float parent_width, float parent_height, float* blk_width, float* blk_height);
+static void draw_internal_calc_coords(int type_descrip_index, t_pb_graph_node* pb_graph_node, int num_block, int num_columns, int num_rows, float parent_width, float parent_height, float* blk_width, float* blk_height);
```
vaughnbetz: Add a doxygen comment documenting the overall functionality (what does this set up?  when should it be called) and all parameters (with @param).

Not essential for this PR, but we could change blk_width and blk_height to vectors and that would be nicer.

I think num_columns and num_rows are the number for drawing things inside this type of tile.  That needs to be explained. At first I thought they were from the grid (they aren't!).

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -237,11 +237,30 @@ static void draw_internal_load_coords(int type_descrip_index, t_pb_graph_node* p
             /* Find the number of instances for each child pb_type. */
             int num_pb = mode.pb_type_children[j].num_pb;
 
+            int num_blocks = num_pb * num_children;
+
+            // determine an optimal number of columns
```
vaughnbetz: Explain this algorithm.  Here's what I think it is:
Find a number or rows and columns that will be as nearly equal as possible (to make a squarish 2D grid) and where num_rows * num_columns gives us exactly the right number of blocks.

vaughnbetz: Avoid using the term optimal, as that means the provably best -- can't prove this is the best visually :).

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -237,11 +237,30 @@ static void draw_internal_load_coords(int type_descrip_index, t_pb_graph_node* p
...
+            int num_columns = 1;
+            for (int k = 1; k * k <= num_blocks; ++k) {
+                if (num_blocks % k == 0) {
+                    num_columns = k;
+                }
+            }
+            int num_rows = num_blocks / num_columns;
+
+            const int MAX_WIDTH_HEIGHT_RATIO = 2;
```
vaughnbetz, line 302: Explain this.  Trying to make a squarish layout again?  Should this also consider num_rows and num_cols?

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -237,11 +237,30 @@ static void draw_internal_load_coords(int type_descrip_index, t_pb_graph_node* p
...
+            if (parent_width > parent_height * MAX_WIDTH_HEIGHT_RATIO) {
+                std::swap(num_columns, num_rows);
+            }
+
             for (int k = 0; k < num_pb; ++k) {
+
+                int num_block = j * num_pb + k;
```
vaughnbetz: I think this is a block index.  In VTR we usually use num_<something> to mean the maximum number of something (e.g. a loop limit or vector size).  If this is a block index, I would use blk_num or iblk or blk_idx instead of num_block.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -258,56 +277,53 @@ static void draw_internal_load_coords(int type_descrip_index, t_pb_graph_node* p
...
-        // that should test for io blocks, and setting capacity_divisor > 1
-        // will squish every thing down
-        capacity_divisor = capacity - 1;
+    // add safety check to ensure that the dimensions will never be below zero
+    if (parent_width <= 2 * abs_parent_padding || parent_height <= 2 * abs_parent_padding - abs_text_padding) {
+        abs_parent_padding = 0;
+        abs_text_padding = 0;
     }
 
     /* Draw all child-level blocks in just most of the space inside their parent block. */
```
vaughnbetz: in just --> using

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -400,7 +416,7 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
             if (draw_state->draw_block_text) {
                 g->draw_text(
                     ezgl::point2d(abs_bbox.center_x(),
-                                  abs_bbox.top() - (abs_bbox.height()) / 15.0),
+                                  abs_bbox.top() - draw_coords->get_tile_height() * 0.01),
```
vaughnbetz: Suggest tying this 0.01 to one of the const drawing values above, or making a new one.  It'll be hard to  find and update/align this 0.01 with other drawing offsets in future changes if you don't give it a name (or even better, compute it from the constants already present).

vpr/src/util/vpr_utils.cpp
```diff
@@ -858,13 +858,13 @@ t_pb_graph_pin* get_pb_graph_node_pin_from_model_port_pin(const t_model_ports* m
 }
 
 //Retrieves the atom pin associated with a specific CLB and pb_graph_pin
+// Warning: Not all pb_graph_pins are associated with an atom pin!
```
vaughnbetz: I'd expand on this. Some represents nodes/pins within the pb_graph hierarchy, but not on a primitive.

vpr/src/util/vpr_utils.cpp
```diff
@@ -1755,6 +1755,42 @@ RRNodeId get_atom_pin_rr_node_id(AtomPinId atom_pin_id) {
...
+    VTR_ASSERT(sub_tile && sub_tile_num >= 0 && sub_tile_num < physical_tile->capacity);
+
+    ClusterBlockId blk_id = place_ctx.grid_blocks().block_at_location({x, y, sub_tile_num, layer});
+    VTR_ASSERT(blk_id != ClusterBlockId::INVALID());
+
+    t_pb_graph_pin* pb_graph_pin = physical_tile->pin_num_to_pb_pin.at(pin_physical_num);
+
+    VTR_ASSERT(pb_graph_pin);
+
+    return {blk_id, pb_graph_pin};
```
vaughnbetz, line 1777: Need to very clearly comment this return type in the Doxygen comment in the .h file (use @return).  Consider using a named structure. A pair could be OK with enough commenting. If you stick with a pair, callers should extract the data into named variables for clarity, or (even better) use something like:

auto [blk_id, pb_graph_pin] = get_rr_node_cluster_blk_id_pb_graph_pin ( ...)

vpr/src/util/vpr_utils.h
```diff
@@ -256,6 +256,22 @@ RRNodeId get_pin_rr_node_id(const RRSpatialLookup& rr_spatial_lookup,
...
+ * **Warning**: This function should be called only if flat-router is enabled,
+ * since, otherwise, the routing resources inside clusters are not added to the RR graph.
+ * @param rr_node_id The RR node ID.
+ */
+std::pair<ClusterBlockId, t_pb_graph_pin*> get_rr_node_cluster_blk_id_pb_graph_pin(RRNodeId rr_node_id);
+
+/**
+ * @brief Returns the atom pin ID for the given RR node ID.
+ * **Warning**: This function should be called only if flat-router is enabled,
+ * since, otherwise, the routing resources inside clusters are not added to the RR graph. Also, not all RRNodes have an AtomPinId associated with them.
```
vaughnbetz: What do you do if the rr_node_id does not correspond to an atom pin?  Return an invalid ID?  assert?  (Bad: silently corrupt memory or you don't know!).
Comment this ... OK to either return the invalid ID if this isn't an atom_pin, or OK to assert but you need to comment which.  I lean toward returning the invalid id in that case since it might be useful for some if statements in the future, but your call.

vpr/src/util/vpr_utils.cpp
```diff
@@ -1755,6 +1755,42 @@ RRNodeId get_atom_pin_rr_node_id(AtomPinId atom_pin_id) {
...
+}
+
+AtomPinId get_rr_node_atom_pin_id(RRNodeId rr_node_id) {
+    ClusterBlockId blk_id;
+    t_pb_graph_pin* pb_graph_pin;
+    std::tie(blk_id, pb_graph_pin) = get_rr_node_cluster_blk_id_pb_graph_pin(rr_node_id);
+
```
vaughnbetz, line 1783: Should have some VTR_ASSERT_SAFE calls here to verify that this rr_node_id does in fact  correspond to an atom pin.
VTR_ASSERT_SAFE is turned off in release build; you could also use VTR_ASSERT if this code isn't very hot and you always want it checked.

vaughnbetz (changes requested):
Looks good; some changes requested though.
As we discussed in the meeting, you should:
- fix the arrow direction (done I think)
- test more architecture files.  
- Stratix IV: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/titan/stratixiv_arch.timing.xml (need to use with a titan benchmark; neuron is the smallest normal one, but titan_other has some smaller ones still)
- Stratix 10: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/titan/stratix10_arch.timing.xml (need to use a Stratix 10 mapped Titan design or a Titanium design)
- Artix 7: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/xilinx/7series_BRAM_DSP_carry.xml
- Z1000: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/zeroasic/z1000/z1000.xml
- The arch we use with Koios (hopefully the developer guide points you at the right task)

vpr/src/draw/draw_searchbar.cpp
SamuelHo10: The problem is that the two nets are different types. The atom netlist has elements of AtomNetId type and the cluster Netlist has elements of ClusterNetId type. Is there an easy way to cast all the elements to a ParentNetId type since I think it would work then?

vaughnbetz: @amin1377 : any ideas? Do we have to explicitly cast to a parentNetID or are there implicit conversion operators?

If this would be messy we can leave the current code.

vpr/src/draw/draw_searchbar.cpp
SamuelHo10: I have it in the highlight_net(char* message, RRNodeId hit_node) function after all the loops.

soheilshahrouz:
@SamuelHo10 
Thansk for this PR.

We recently added a [coding style guide](https://docs.verilogtorouting.org/en/latest/dev/coding_style/) to the documentation. Please read the guide and make sure that the code is consistent with the style rules.

vpr/src/draw/intra_logic_block.cpp
SamuelHo10: Yeah I agree we should use const drawing values, but currently all the const drawing variables are declared locally, so I could not access them. Do you think would be a good idea to have the const variables to be declared globally somewhere like in the t_draw_coords struct or in a separate namespace? I am not sure of the best way to do this.

vpr/src/draw/draw_searchbar.cpp
amin1377: I just discussed it with Samuel, and I think we should leave it as is. Both ClusterNetId and AtomNetId inherit from ParentNetId. However, the issue lies with the .nets() function, which returns a range based on ClusterNetId or AtomNetId. Although there are potential workarounds, I think implementing them would make the code a bit messy.

vpr/src/draw/intra_logic_block.cpp
vaughnbetz: You could use constexpr to make constants at the top of the file for values that are used in multiple places in the file.  If a constant value is used only in one routine, it is good to declare it there.  If in more than one routine but in one file, you can use constexpr at the top of the file. If in multiple files, then use constexpr in a header file they all include (e.g. draw_types.h or could make a new one).

vpr/src/draw/draw_searchbar.cpp
amin1377: I think the cleanest approach would be to encapsulate these functions within a class, with the parent netlist stored as a member field that’s initialized to either the ClusterNetlist or AtomNetlist when the class is created. However, implementing this might be beyond the scope of this PR.

vpr/src/util/vpr_utils.cpp
SamuelHo10: Moved comment to header file using doxygen format

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
-        /* Allocate some temporary storage if it's not already available. */
-        int width = (int)device_ctx.grid.width();
-        int height = (int)device_ctx.grid.height();
-        if (chanx_track.empty()) {
-            chanx_track = vtr::OffsetMatrix<int>({{{1, width - 1}, {0, height - 1}}});
+    // Draw Pins
+    for (size_t i = 1; i < rr_nodes_to_draw.size(); ++i) {
+        RRNodeId inode = rr_nodes_to_draw[i];
+        auto rr_type = rr_graph.node_type(inode);
+        bool is_inode_inter_cluster = is_inter_cluster_node(rr_graph, inode);
```
amin1377, line 619: I think the name of this variable should be changed to inter_cluster_inode. The prefix "is_" is typically used for functions, not variables.

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
+        // Skip drawing sources and sinks
+        if (rr_type == e_rr_type::SINK || rr_type == e_rr_type::SOURCE) {
+            continue;
         }
 
-        if (chany_track.empty()) {
-            chany_track = vtr::OffsetMatrix<int>({{{0, width - 1}, {1, height - 1}}});
+        // Draw intra-cluster nodes
+        if (!is_inode_inter_cluster) {
+            draw_rr_intrapin(inode, color, g);
```
amin1377: I think the name of this function should be changed to something like draw_intra_cluster_pin or something similar.

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
+            continue;
         }
 
-        for (int i = 1; i < width - 1; i++)
-            for (int j = 0; j < height - 1; j++)
-                chanx_track[i][j] = (-1);
+        // Draw IO Pins
+        if (rr_type == e_rr_type::OPIN || rr_type == e_rr_type::IPIN) {
+            draw_rr_pin(inode, color, g);
```
amin1377: I assume this function only draws cluster pins. To make the contrast clearer between this function and the one that draws intra-cluster pins, I’d suggest renaming it to something like draw_cluster_pin.

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
+            continue;
+        }
     }
 
+    // Draw Edges
     for (size_t i = 1; i < rr_nodes_to_draw.size(); ++i) {
+
         RRNodeId inode = rr_nodes_to_draw[i];
         auto rr_type = rr_graph.node_type(inode);
+        bool is_inode_inter_cluster = is_inter_cluster_node(rr_graph, inode);
```
amin1377: Same comment as above

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
+        int current_node_layer = rr_graph.node_layer(inode);
 
         RRNodeId prev_node = rr_nodes_to_draw[i - 1];
         auto prev_type = rr_graph.node_type(RRNodeId(prev_node));
+        bool is_prev_node_inter_cluster = is_inter_cluster_node(rr_graph, prev_node);
```
amin1377: Same comment as above

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,133 +612,139 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
 
-        int current_node_layer = rr_graph.node_layer(inode);
-        int prev_node_layer = rr_graph.node_layer(prev_node);
-        t_draw_layer_display edge_visibility = get_element_visibility_and_transparency(prev_node_layer, current_node_layer);
+        g->set_color(color, edge_visibility.alpha);
 
-        //Don't draw node if the layer of the node is not set to visible on screen
-        if (!draw_state->draw_layer_display[current_node_layer].visible) {
+        if (!is_inode_inter_cluster && !is_prev_node_inter_cluster) {
+            draw_intrapin_to_intrapin(inode, prev_node, g);
```
amin1377: Based on the function name (draw_intrapin_to_intrapin), it seems to draw the connection between two intra-cluster pins. I’d suggest renaming it to draw_intra_cluster_connection for clarity.

amin1377 (changes requested):
Thanks, Samuel!

vpr/src/util/vpr_utils.h
SamuelHo10: I went with an assert statement for now.

SamuelHo10:
Thanks for the code reviews @amin1377 and @vaughnbetz. I've tried my best to improve the code documentation and clarity based on your suggestions. I still have yet to test the other architectures, so I will try to get that done as soon as possible.

@AlexandreSinger At our last meeting, you suggested revising the naming of the internal blocks. Previously, the format was pb_type->name [pb->name]. I've updated the code to align more closely with the hierarchical_type_name() function in vpr_types.h, so it now prints as pb_type->name[placement_index][mode_name].
Would this naming scheme be more helpful for your use case?
vpr_graphics_name

vaughnbetz:
I agree; I think using : would be better than [] to separate mode names and instance names from the hierarchical name of the sub-block types.

vpr/src/draw/draw_rr_edges.h
```diff
@@ -17,12 +17,53 @@
...
+ * @param chanx_node The horizontal channel node
+ * @param chany_node The vertical channel node
+ * @param edge_dir The direction of the edge, FROM_X_TO_Y or FROM_Y_TO_X
+ * @param switch_type The type of switch used for the connection
+ * @param g The ezgl renderer 
+ */
 void draw_chanx_to_chany_edge(RRNodeId chanx_node, RRNodeId chany_node, enum e_edge_dir edge_dir, short switch_type, ezgl::renderer* g);
+
+/**
+ * @brief Draws the edge between an intra-cluster pin and a pin when flat routing is enabled. It does not matter whether prev_node is the intra-cluster pin or whether inode is the intra-cluster pin.
```
vaughnbetz: "between an intra-cluster pin and a pin"
I think the second pin is an iner-cluster pin (or equivalently, you could call that a top-level cluster pin if you prefer).  I think the comment should make clear what the type of the second pin is.

vpr/src/draw/draw_searchbar.cpp
vaughnbetz: OK, I'm fine with leaving this alone then.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -34,11 +34,38 @@
 #include "draw.h"
 #include "draw_triangle.h"
 
+constexpr float FRACTION_TEXT_PADDING = 0.01;
```
vaughnbetz, line 38: Should put a comment above this briefly explaining what it is for and what it is a fraction of.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -34,11 +34,38 @@
...
+
+/**
+ * @brief Finds the maximum depth of sub-blocks for a given pb_type.
+ * @param pb_type The top-level pb_type.
+ * @return The maximum level for the given pb_type.
+ */
 static int draw_internal_find_max_lvl(const t_pb_type& pb_type);
-static void draw_internal_calc_coords(int type_descrip_index, t_pb_graph_node* pb_graph_node, int num_pb_types, int type_index, int num_pb, int pb_index, float parent_width, float parent_height, float* blk_width, float* blk_height);
+/**
+ * @brief A helper function for draw_internal_load_coords. Calculates the coordinates of a internal block and stores its bounding box inside global variables. The calculated width and height of the block are also assigned to the pointers blk_width and blk_height.
```
vaughnbetz: Line is too long; you should break it into multiple lines so it displays nicely.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -237,11 +264,34 @@ static void draw_internal_load_coords(int type_descrip_index, t_pb_graph_node* p
             /* Find the number of instances for each child pb_type. */
             int num_pb = mode.pb_type_children[j].num_pb;
 
+            // Determine how we want to arrange the sub-blocks in the parent block. We want the blocks to be squarish, and not too wide or too tall. In other words, we want the number of rows to be as close to the number of columns as possible such that num_rows * num_columns =  num_blocks.
```
vaughnbetz: Break into multiple lines so it displays nicely.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -375,43 +421,38 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
...
-            // just label its type, and put it up at the top so we can see it
-            if (draw_state->draw_block_text) {
-                g->draw_text(
-                    ezgl::point2d(abs_bbox.center_x(),
-                                  abs_bbox.top() - (abs_bbox.height()) / 15.0),
-                    pb_type->name,
-                    abs_bbox.width(),
-                    abs_bbox.height());
-            }
+    pb_type_name += "[" + std::to_string(pb->pb_graph_node->placement_index) + "]";
```
vaughnbetz: Put a comment at the top of the overall logic explaining the convention you are following to make names (i.e. how do I interpret these names?)

vpr/src/util/vpr_utils.h
```diff
@@ -166,6 +166,9 @@ const t_pb_graph_pin* find_pb_graph_pin(const t_pb_graph_node* pb_gnode, const s
 
 const t_pb_graph_pin* find_pb_graph_pin(const AtomNetlist& netlist, const AtomPBBimap& atom_pb_lookup, const AtomPinId pin_id);
 
+/**
```
vaughnbetz, line 169: If you know what happens if you call this with a pb_graph_pin that does not match an atom pin, it is good to add that to the comment. Is that an error or is an invalid AtomPinId returned?

vaughnbetz (changes requested):
Thanks for all the updates ... they look good @SamuelHo10 !
I have a few more requests. After addressing those and checking the listed archs, this is good to merge from my perspective.

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,138 +612,168 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
-    if (draw_state->draw_route_type == e_route_type::GLOBAL) {
-        /* Allocate some temporary storage if it's not already available. */
-        int width = (int)device_ctx.grid.width();
-        int height = (int)device_ctx.grid.height();
-        if (chanx_track.empty()) {
-            chanx_track = vtr::OffsetMatrix<int>({{{1, width - 1}, {0, height - 1}}});
+    // Draw RR Nodes
+    for (size_t i = 1; i < rr_nodes_to_draw.size(); ++i) {
+        RRNodeId inode = rr_nodes_to_draw[i];
+        auto rr_type = rr_graph.node_type(inode);
```
soheilshahrouz: auto -> explicit type name

vpr/src/draw/draw_basic.cpp
```diff
@@ -616,138 +612,168 @@ void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::ren
...
-                                      "Unexpected connection from an rr_node of type %d to one of type %d.\n",
-                                      prev_type, rr_type);
-                        }
-                    }
-                }
+void draw_inter_cluster_rr_edge(RRNodeId inode, RRNodeId prev_node, e_rr_type rr_type, e_rr_type prev_type, const RRGraphView& rr_graph, ezgl::renderer* g) {
 
-                break;
+    auto iedge = find_edge(prev_node, inode);
+    auto switch_type = rr_graph.edge_switch(RRNodeId(prev_node), iedge);
```
soheilshahrouz: auto -> typename

vpr/src/draw/draw_basic.h
```diff
@@ -57,6 +57,12 @@ void draw_routed_net(ParentNetId net, ezgl::renderer* g);
 void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw,
                         ezgl::renderer* g);
 
+/** @brief Draws an edge between two rr_nodes, which are both intra-cluster nodes.
+ * @param inode The current rr_node id
+ * @param prev_node The previous rr_node id
+ * @param g The ezgl renderer
+ */
+void draw_inter_cluster_rr_edge(RRNodeId inode, RRNodeId prev_node, e_rr_type rr_type, e_rr_type prev_type, const RRGraphView& rr_graph, ezgl::renderer* g);
```
soheilshahrouz: add an empty line between functions
isn't possible to retrieve node types from rr_graph instead of passing them as args?

vpr/src/draw/draw_rr.cpp
```diff
@@ -513,10 +513,30 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
 
+void draw_rr_intra_cluster_pin(RRNodeId inode, const ezgl::color& color, ezgl::renderer* g) {
+    t_draw_state* draw_state = get_draw_state_vars();
+    t_draw_coords* draw_coords = get_draw_coords_vars();
+
+    if (!draw_state->is_flat) {
+        return;
+    }
+
+    auto blk_id_pin_id = get_rr_node_cluster_blk_id_pb_graph_pin(inode);
```
soheilshahrouz: structured binding or std::tie makes the code more readable

vpr/src/draw/draw_rr.cpp
```diff
@@ -683,6 +703,32 @@ RRNodeId draw_check_rr_node_hit(float click_x, float click_y) {
...
+        }
+
+        // Check for intra cluster nodes
+        if (!is_inter_cluster_node(rr_graph, inode)) {
+
+            if (!draw_state->is_flat) {
+                continue;
+            }
+
+            auto blk_id_pin_id = get_rr_node_cluster_blk_id_pb_graph_pin(inode);
```
soheilshahrouz: structured binding or std::tie makes the code more readable

vpr/src/draw/draw_rr.h
```diff
@@ -27,10 +27,14 @@ void draw_rr_edges(RRNodeId from_node, ezgl::renderer* g);
 
 void draw_rr_chan(RRNodeId inode, const ezgl::color color, ezgl::renderer* g);
 
+/**
+ * @brief Draws the intra-cluster pin for a given RRNodeId when flat routing is enabled.
+ */
+void draw_rr_intra_cluster_pin(RRNodeId inode, const ezgl::color& color, ezgl::renderer* g);
```
soheilshahrouz, line 33: add empty line between functions

vpr/src/draw/draw_rr_edges.h
```diff
@@ -17,12 +17,62 @@
...
+
+/**
+ * @brief Draws the edge between an intra-cluster pin and an inter-cluster pin when flat routing is enabled.
+ * @param intra_cluster_node The intra-cluster pin node
+ * @param inter_cluster_node The inter-cluster pin node
+ * @param pin_edge_dir The direction of the edge, FROM_INTER_CLUSTER_TO_INTRA_CLUSTER or FROM_INTRA_CLUSTER_TO_INTER_CLUSTER
+ * @param pin_side The side of the inter-cluster pin (e.g. TOP, RIGHT, BOTTOM, LEFT)
+ * @param g The ezgl renderer
+ */
+void draw_intra_cluster_pin_to_pin(RRNodeId intra_cluster_node, RRNodeId inter_cluster_node, e_pin_edge_dir pin_edge_dir, e_side pin_side, ezgl::renderer* g);
```
soheilshahrouz, line 48: add empty line

vpr/src/draw/draw_rr_edges.h
```diff
@@ -17,12 +17,62 @@
...
 void draw_pin_to_sink(RRNodeId ipin_node, RRNodeId sink_node, ezgl::renderer* g);
 void draw_source_to_pin(RRNodeId source_node, RRNodeId opin_node, ezgl::renderer* g);
+
+/**
+ * @brief Determines the side of clb, the inter-cluster pin is located on based on the channel node.
+ * @param pin_node The inter-cluster pin node
+ * @param chan_node The channel node (CHANX or CHANY) that the pin is connected to
+ * @return The side of the clb that the pin is located on (e.g. TOP, RIGHT, BOTTOM, LEFT)
+ */
+e_side get_pin_side(RRNodeId pin_node, RRNodeId chan_node);
```
soheilshahrouz, line 74: adding an empty line between functions makes the visually separate and more readable

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -131,40 +131,61 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
...
     auto& cluster_ctx = g_vpr_ctx.clustering();
+    auto& atom_ctx = g_vpr_ctx.atom();
     auto& route_ctx = g_vpr_ctx.routing();
 
     /* Don't crash if there's no routing */
     if (route_ctx.route_trees.empty())
         return;
 
+    if (route_ctx.is_flat) {
+        for (auto net_id : atom_ctx.netlist().nets()) {
```
soheilshahrouz: auto -> typename

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -131,40 +131,61 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
...
+            check_node_highlight_net(message, net_id, hit_node);
+        }
+
+    } else {
+        for (auto net_id : cluster_ctx.clb_nlist.nets()) {
```
soheilshahrouz: auto -> typename

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -131,40 +131,61 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
...
-                    sprintf(message, "%s  ||  Net: %zu (%s)", orig_msg.c_str(),
-                            size_t(net_id),
-                            cluster_ctx.clb_nlist.net_name(net_id).c_str());
-                }
-            } else if (draw_state->draw_rr_node[inode].color
-                       == ezgl::WHITE) {
-                // If node is de-selected.
-                draw_state->net_color[net_id] = ezgl::BLACK;
-                break;
+    for (auto& rt_node : route_ctx.route_trees[parent_net_id].value().all_nodes()) {
```
soheilshahrouz: auto -> typename

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -131,40 +131,61 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
...
+        if (draw_state->draw_rr_node[inode].color == ezgl::MAGENTA) {
+            draw_state->net_color[parent_net_id] = draw_state->draw_rr_node[inode].color;
+            if (inode == hit_node) {
+                std::string orig_msg(message);
+                sprintf(message, "%s  ||  Net: %zu (%s)", orig_msg.c_str(),
+                        size_t(parent_net_id),
+                        draw_get_net_name(parent_net_id).c_str());
             }
+        } else if (draw_state->draw_rr_node[inode].color
+                   == ezgl::WHITE) {
```
soheilshahrouz: Why is this condition broken into two lines?
It is not that long

vpr/src/draw/draw_types.h
```diff
@@ -109,13 +109,18 @@ enum e_draw_net_type {
     ALL_NETS,
     HIGHLIGHTED
 };
-
-/* Chanx to chany or vice versa? */
-enum e_edge_dir {
+/// Chanx to chany or vice versa?
```
soheilshahrouz, line 113: empty line

SamuelHo10:
@vaughnbetz I got the Z1000 architecture to work with the help of Alex. However, when visualizing intra-cluster routing, the assertion on vpr_utils.cpp:1771 gets called.
```
ClusterBlockId blk_id = place_ctx.grid_blocks().block_at_location({x, y, sub_tile_num, layer});
VTR_ASSERT(blk_id != ClusterBlockId::INVALID());
```
Inter-cluster route visualization still appears to work. I suspect the issue arises because we’re loading the RR graph externally. Other UI features, like toggling RR, also seem to break on this architecture.

vaughnb-cerebras:
OK, we can merge without waiting for that.  But we should debug this. Can you file an issue to track it, and enlist help from Alex and Amin as needed?

vaughnbetz:
Looks good!
One more change: 
1. Change the names shown in the pb_type architecture hierarchy. Currently it seems you are printing pb_name:mode_name[index].
Change that to 
pb_name[index]:mode_name.

Also, if pb_name == mode_name (before you add the [index]) you should omit the :mode_name part.

2. Probably post this PR:
- How do we show what is placed at a lowest level pb_type location?  This can hold a primitive instance.  If there is no primitive instance (from the atom netlist) show the final pb_type in what / empty.  If there is something there, colour it in (I think you already do this).  Confirm that clicking on that primitive shows its name.  We could also print its name in brackets in that location, e.g. 
lut (bob_lut_to_fix_adder)

If this looks silly due to the instance names being long we can not do it.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -434,19 +434,22 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
...
     if (!pb->is_primitive()) {
-        // primitives have no modes
+        // Format for non-primitives: <block_type_name>[<placement_index>]:<mode_name>
+        pb_type_name += "[" + std::to_string(pb->pb_graph_node->placement_index) + "]";
         std::string mode_name = pb->pb_graph_node->pb_type->modes[pb->mode].name;
         pb_type_name += ":" + mode_name;
+    } else {
+        // Format for primitives: <block_type_name>(<block_name>)
+        if (pb->name != nullptr) {
+            std::string pb_name(pb->name);
```
vaughnbetz, line 453: @SamuelHo10 : could you show an example of this?  I'm looking for the name of the atom placed there for primitives (if there is one) which should require an atom netlist lookup.
Again, if it looks messy due to being too long a name we can not do it.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -434,19 +434,22 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
...
-    // draw text for each physical block.
-    // format: <block_type_name>:<mode_name>[<placement_index>]
+    // Display text for each physical block.
     std::string pb_type_name(pb_type->name);
 
-    // get the mode of the physical block
     if (!pb->is_primitive()) {
-        // primitives have no modes
+        // Format for non-primitives: <block_type_name>[<placement_index>]:<mode_name>
+        pb_type_name += "[" + std::to_string(pb->pb_graph_node->placement_index) + "]";
```
vaughnbetz: Can we also not print the mode if the mode_name == pb_type_name (i.e. it provides redundant info)?  Would require that if statement.

vaughnbetz: Looks like you just did that in the next commit :).

vpr/src/draw/intra_logic_block.cpp
SamuelHo10: The names seem short for this architecture at least.

PR #3160, opened 2025-06-22
[AP][DP] Added Option to Scale Init T in Annealer

AlexandreSinger:
Since the AP flow produces an initial placement that has 4x improved estimated wirelength and 2x improved CPD, the initial temperature of the annealer is too high. From experimentation, I have found that reducing the initial temperature in half improves quality slightly and reduces runtime.

Added an option to the Placer to scale up or down the initial temperature. The default flow just sets this scale to 1.0, but the AP flow sets this scale to 0.5 currently.

I have also exposed this option to the command-line so this option can be swept more easily.

Results on VTR with AP (timing driven, no fixed IOs):

Init T Scale | Post-DP HPWL | Post-DP CPD | Final Wirelength | Final CPD | DP Run Time | Routing Run Time | Total Run Time
1.0000 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00
0.5000 | 0.99 | 0.99 | 0.99 | 1.00 | 0.85 | 1.01 | 0.96
0.2500 | 1.02 | 1.00 | 1.01 | 1.01 | 0.70 | 1.01 | 0.92
0.1250 | 1.04 | 1.00 | 1.03 | 1.01 | 0.57 | 1.02 | 0.88
0.0625 | 1.05 | 1.00 | 1.04 | 1.01 | 0.47 | 1.07 | 0.84

Scaling the initial temperature does hurt wirelength and CPD; however, reducing it by 0.5 hits a minimum point. Refarding run time, decreasing the initial temperature greatly reduces the runtime of DP which results in an overall runtime improvement. Routing takes more and more time, which implies to me that the circuit may be becoming more difficult to route, but that may just be caused by the increase in wirelength.

Overall, setting the initial temperature scale to 0.5 for AP makes sense since it improves on practically all metrics. I am running Titan now. I predict that Titan may prefer to set the scale to 0.25 (which I would prefer too).

Future work is to investigate sweeping inner_num with this scale to find a good combination of the two to get better quality and run time.

### DIFF
--- a/doc/src/vpr/command_line_usage.rst
+++ b/doc/src/vpr/command_line_usage.rst
@@ -810,6 +810,22 @@ If any of init_t, exit_t or alpha_t is specified, the user schedule, with a fixe
 
     **Default:** ``circuit``
 
+.. option:: --anneal_auto_init_t_scale <float>
+
+    A scale on the starting temperature of the anneal for the automatic annealing
+    schedule.
+
+    When in the automatic annealing schedule, the annealer will select a good
+    initial temperature based on the quality of the initial placement. This option
+    allows you to scale that initial temperature up or down by multiplying the
+    initial temperature by the given scale. Increasing this number
+    will increase the initial temperature which will have the annealer potentially
+    explore more of the space at the expense of run time. Depending on the quality
+    of the initial placement, this may improve or hurt the quality of the final
+    placement.
+
+    **Default:** ``1.0``
+
 .. option:: --init_t <float>
 
     The starting temperature of the anneal for the manual annealing schedule.
--- a/vpr/src/analytical_place/detailed_placer.cpp
+++ b/vpr/src/analytical_place/detailed_placer.cpp
@@ -75,6 +75,19 @@ AnnealerDetailedPlacer::AnnealerDetailedPlacer(const BlkLocRegistry& curr_cluste
         }
     }
 
+    // The solution produced by the AP flow is significantly better than the
+    // initial placement solution produced by the initial placer in the default
+    // flow. Even though the annealer auto-selects its initial temperature based
+    // on the quality of the placement, we found that the initial temperatute was
+    // still too high and it was hurting quality. This scales down the initial
+    // temperature auto-selected by the annealer to prevent this and improve
+    // runtime.
+    // Here we still scale by whatever the user passed in to scale the initial
+    // temperature by, but we also scale it down further. This allows AP to scale
+    // the initial temperature down by default, while still allowing the user
+    // some control.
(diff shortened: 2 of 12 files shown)


doc/src/vpr/command_line_usage.rst
```diff
@@ -810,6 +810,21 @@ If any of init_t, exit_t or alpha_t is specified, the user schedule, with a fixe
...
+.. option:: --anneal_auto_init_t_scale <float>
+
+    A scale on the starting temperature of the anneal for the automatic annealing
+    schedule.
+
+    When in the automatic annealing schedule, the annealer will select a good
+    initial temperature based on the quality of the initial placement. This option
+    allows you to scale that initial temperature up or down. Increasing this number
+    will increase the initial temperature which will have the annealer potentially
+    explore more of the space at the expense of run time. Depending on the quality
```
vaughnbetz, line 823: Say what the adjustment is -- multiply by this parameter.

vpr/src/analytical_place/detailed_placer.cpp
```diff
@@ -75,6 +75,15 @@ AnnealerDetailedPlacer::AnnealerDetailedPlacer(const BlkLocRegistry& curr_cluste
...
     }
 
+    // The solution produced by the AP flow is significantly better than the
+    // initial placement solution produced by the initial placer in the default
+    // flow. Even though the annealer auto-selects its initial temperature based
+    // on the quality of the placement, we found that the initial temperatute was
+    // still too high and it was hurting quality. This scales down the initial
+    // temperature auto-selected by the annealer to prevent this and improve
+    // runtime.
+    float anneal_auto_init_t_scale = vpr_setup.PlacerOpts.place_auto_init_t_scale * 0.5f;
```
vaughnbetz, line 89: Explain that we use an init_t that is half as big, and still scale by what the user sent in.

vpr/src/base/read_options.cpp
```diff
@@ -2200,6 +2200,21 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
...
 
+    place_grp.add_argument(args.place_auto_init_t_scale, "--anneal_auto_init_t_scale")
+        .help(
+            "A scale on the starting temperature of the anneal for the automatic annealing "
+            "schedule.\n"
+            "\n"
+            "When in the automatic annealing schedule, the annealer will select a good "
+            "initial temperature based on the quality of the initial placement. This option "
+            "allows you to scale that initial temperature up or down. Increasing this number "
+            "will increase the initial temperature which will have the annealer potentially "
```
vaughnbetz, line 2212: Same comment as earlier: say multiplicative.

vaughnbetz:
Looks good!

PR #3161, opened 2025-06-24
[RouterLookahead] Router Lookahead Report

AlexandreSinger:
Added an option to generate a report on the router lookahead. This reports profiles the router lookahead by performing a set of trial routes from source RR nodes to all target RR nodes and comparing the estimated cost from the router lookahead to the actual cost of the path.

Notably, this report tries to show the mean squared error of the router lookahead and the maximum overestimation of the router lookahead.

The mean squared error is a measure of how accurate the router lookahead is. The more accurate the router lookahead is, the faster the path search will be (with less loss in quality). The max overestimation is a measure of the admissibility of the router lookahead heuristic. The more the router lookahead overestimates, the worse the quality of the path found (theoretically).


AlexandreSinger:
@amin1377 FYI, you might find this option interesting.

libs/librrgraph/src/base/rr_node_types.h
```diff
@@ -32,6 +32,31 @@ enum class e_rr_type : unsigned char {
...
+            return "CHANX";
+        case e_rr_type::CHANY:
+            return "CHANY";
+        case e_rr_type::NUM_RR_TYPES:
+            return "NUM_RR_TYPES";
+        default:
+            return "UNKNOWN";
+    }
+}
+
```
soheilshahrouz: `rr_node_typename` in rr_node_types.h does the same thing.

vpr/src/route/router_lookahead/router_lookahead_report.cpp
```diff
@@ -0,0 +1,314 @@
...
+    cost_params.criticality = 1.;
+    cost_params.astar_fac = router_opts.astar_fac;
+    cost_params.astar_offset = router_opts.astar_offset;
+    cost_params.post_target_prune_fac = router_opts.post_target_prune_fac;
+    cost_params.post_target_prune_offset = router_opts.post_target_prune_offset;
+    cost_params.bend_cost = router_opts.bend_cost;
+
+    // Create a NOP router lookahead to be used during the all-destination dijkstra search.
+    t_det_routing_arch temp_det_routing_arch;
+    auto temp_router_lookahead = make_router_lookahead(temp_det_routing_arch, e_router_lookahead::NO_OP,
```
soheilshahrouz: replace auto with explicit type name

vpr/src/route/router_lookahead/router_lookahead_report.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    // We are going to collect info on the overestimation from each RR node type.
+    struct t_overestimation_info {
+        float overestimation = 0.0f;
+        RRNodeId src_node = RRNodeId::INVALID();
+        RRNodeId target_node = RRNodeId::INVALID();
+        float heuristic_delay = 0.0f;
+        float path_delay = 0.0f;
+    };
+    std::vector<t_overestimation_info> max_overestimation_per_type((int)e_rr_type::NUM_RR_TYPES);
```
soheilshahrouz: You can use vtr::vector and index it directly with e_rr_type

vpr/src/route/router_lookahead/router_lookahead_report.cpp
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/**
+ * @brief This method profiles the given router lookahead by trying several
+ *        routes with no lookahead and comparing the "ideal" path costs to the
+ *        costs estimated by the router lookahead. This information can be
+ *        valuable since over-estimations in the router lookahead cause the
+ *        heuristic to be inadmissible which can hurt quality.
+ */
+static void profile_lookahead_overestimation(std::ofstream& os,
```
soheilshahrouz: This functio is quite long. Consider breaking it down

vpr/src/route/router_lookahead/router_lookahead_report.cpp
```diff
@@ -0,0 +1,314 @@
...
+    const RRGraphView& rr_graph = device_ctx.rr_graph;
+    RoutingContext& route_ctx = g_vpr_ctx.mutable_routing();
+
+    print_header(os, "Lookeahead Overestimation Report");
+
+    // Variables for the profiling.
+    //  The target number of random source (sample) nodes to use.
+    size_t target_num_trials = 100;
+    //  The maximum number of attempts.
+    size_t max_attempts = 10000;
```
soheilshahrouz: use constexpr if there is no plan to modify these values through command line options

libs/librrgraph/src/base/rr_node_types.h
AlexandreSinger: LOL, thanks. I knew a function like this had to have existed I just did not see it because the line was too long and I did not see it on my terminal.

PR #3162, opened 2025-06-24
[Packer] Reorganized Iterative Packer Algorithm

AlexandreSinger:
The packer iteratively restarts itself with different options if the packer fails to find a dense enough clustering. The way this was written was very hard to follow and was error prone.

Rewrote the iterative packer code to be more like a state-machine. This makes it easier to add new states in the future and maintain the current states.

My goal here is to be accurate to what the packer was doing before (i.e. no results will change, the packer will behave exactly as it did before). So some of the state transitions may make you say "why do that, why not this?"; those changes will come in a future PR. This is just cleanup.

### DIFF
--- a/vpr/src/pack/pack.cpp
+++ b/vpr/src/pack/pack.cpp
@@ -22,10 +22,121 @@
 #include "vtr_assert.h"
 #include "vtr_log.h"
 
+namespace {
+
+/**
+ * @brief Enumeration for the state of the packer.
+ *
+ * If the packer fails to find a dense enough packing, depending on how the
+ * packer failed, the packer may iteratively retry packing with different
+ * settings to try and find a denser packing. These states represent a
+ * different type of iteration.
+ */
+enum class e_packer_state {
+    /// @brief Default packer state.
+    DEFAULT,
+    /// @brief Succcess state for the packer. The packing looks feasible to
+    ///        fit on the device (does not exceed number of blocks of each
+    ///        type in the grid) and meets floorplanning constraints.
+    SUCCESS,
+    /// @brief Standard fallback where there are no region constraints. Turns
+    ///        on unrelated clustering and balanced packing if it can.
+    UNRELATED_AND_BALANCED,
+    /// @brief Region constraints: Turns on attraction groups for overfilled regions.
+    ATTRACTION_GROUPS,
+    /// @brief Region constraints: Turns on more attraction groups for overfilled regions.
+    MORE_ATTRACTION_GROUPS,
+    /// @brief Region constraints: Turns on more attraction groups for all regions.
+    ATTRACTION_GROUPS_ALL_REGIONS,
+    /// @brief Region constraints: Turns on more attraction groups for all regions
+    ///        and increases the target density of "clb" blocks.
+    ///        TODO: This should increase the target density of all overused blocks.
+    ATTRACTION_GROUPS_ALL_REGIONS_AND_INCREASED_TARGET_DENSITY,
+    /// @brief The failure state.
+    FAILURE
+};
+
+} // namespace
+
 static bool try_size_device_grid(const t_arch& arch,
                                  const std::map<t_logical_block_type_ptr, size_t>& num_type_instances,
                                  float target_device_utilization,
                                  const std::string& device_layout_name);
+
+/**
+ * @brief The packer iteratively re-packes the netlist if it fails to find a
+ *        valid clustering. Each iteration is a state the packer is in, where
+ *        each state uses a different set of options. This method gets the next
+ *        state of the packer given the current state of the packer.
+ *
+ *  @param current_packer_state
+ *      The current state of the packer (the state used to make the most recent
+ *      clustering).
+ *  @param fits_on_device
+ *      Whether the current clustering fits on the device or not.
+ *  @param floorplan_regions_overfull
+ *      Whether the current clustering has overfilled regions.
+ *  @param floorplan_not_fitting
+ *      Whether the current clustering is fitting on the current floorplan.
+ */
+static e_packer_state get_next_packer_state(e_packer_state current_packer_state,
+                                            bool fits_on_device,
+                                            bool floorplan_regions_overfull,
+                                            bool floorplan_not_fitting) {
+    // Next packer state logic
+    e_packer_state next_packer_state = e_packer_state::FAILURE;
+    if (fits_on_device && !floorplan_regions_overfull) {
+        // If everything fits on the device and the floorplan regions are
+        // not overfilled, the next state is success.
+        next_packer_state = e_packer_state::SUCCESS;
+    } else {
+        if (floorplan_not_fitting) {
+            // If there are overfilled region constraints.
+
+            // When running with tight floorplan constraints, some regions may become overfull with clusters (i.e.
+            // the number of blocks assigned to the region exceeds the number of blocks available). When this occurs, we
+            // cluster more densely to be able to adhere to the floorplan constraints. However, we do not want to cluster more
+            // densely unnecessarily, as this can negatively impact wirelength. So, we have iterative approach. We check at the end
+            // of every iteration if any floorplan regions are overfull. In the first iteration, we run
+            // with no attraction groups (not packing more densely). If regions are overfull at the end of the first iteration,
+            // we create attraction groups for partitions with overfull regions (pack those atoms more densely). We continue this way
+            // until the last iteration, when we create attraction groups for every partition, if needed.
+
+            switch (current_packer_state) {
+                case e_packer_state::DEFAULT:
+                    next_packer_state = e_packer_state::ATTRACTION_GROUPS;
+                    break;
+                case e_packer_state::UNRELATED_AND_BALANCED:
+                case e_packer_state::ATTRACTION_GROUPS:
+                    next_packer_state = e_packer_state::MORE_ATTRACTION_GROUPS;
+                    break;
+                case e_packer_state::MORE_ATTRACTION_GROUPS:
+                    next_packer_state = e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS;
+                    break;
+                case e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS:
+                    next_packer_state = e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS_AND_INCREASED_TARGET_DENSITY;
+                    break;
+                case e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS_AND_INCREASED_TARGET_DENSITY:
+                default:
+                    next_packer_state = e_packer_state::FAILURE;
+                    break;
+            }
+        } else {
+            // If there are no overfilled region constraints, but some block
+            // types on the grid are overfilled.
+            switch (current_packer_state) {
+                case e_packer_state::DEFAULT:
+                    next_packer_state = e_packer_state::UNRELATED_AND_BALANCED;
+                    break;
+                default:
+                    next_packer_state = e_packer_state::FAILURE;
+                    break;
+            }
+        }
+    }
+
+    return next_packer_state;
+}
 
 bool try_pack(const t_packer_opts& packer_opts,
               const t_analysis_opts& analysis_opts,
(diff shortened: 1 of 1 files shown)


vpr/src/pack/pack.cpp
```diff
@@ -145,7 +145,39 @@ bool try_pack(const t_packer_opts& packer_opts,
...
+     * packer failed, the packer may iteratively retry packing with different
+     * setting to try and find a denser packing. These states represent a
+     * different type of iteration.
+     */
+    enum class e_packer_state {
+        /// @brief Default packer state.
+        DEFAULT,
+        /// @brief Succcess state for the packer.
+        SUCCESS,
+        /// @brief Standard fallback where there is not region constraints. Turns
```
AmirhosseinPoolad: There are no*

vpr/src/pack/pack.cpp
```diff
@@ -170,48 +202,98 @@ bool try_pack(const t_packer_opts& packer_opts,
 
         bool floorplan_not_fitting = (floorplan_regions_overfull || g_vpr_ctx.floorplanning().constraints.get_num_partitions() > 0);
 
+        // Next packer state logic
```
AmirhosseinPoolad, line 287: I think it would be really nice if this entire section was put in a helper function that takes the current state and the other needed variables and returns the new state. You'll probably need to define the enum in file scope instead of function scope but I think it's worth it to make the loop more readable.

vpr/src/pack/pack.cpp
```diff
@@ -170,48 +202,98 @@ bool try_pack(const t_packer_opts& packer_opts,
...
+                        break;
+                }
             }
-            if (packer_opts.balance_block_type_utilization == e_balance_block_type_util::AUTO) {
-                VTR_ASSERT(balance_block_type_util == false);
-                balance_block_type_util = true;
+        }
+
+        // Set up for the next packer state.
+        switch (next_packer_state) {
```
AmirhosseinPoolad, line 297: There's a little bit of a safety issue here in that if in the future someone decides to skip one of the stages and go straight to the next stage (by modifying the previous section) this code might leave the packer in an invalid state. Essentially there's a coupling between the state diagram/transitions in the state machine and the effect of each state.

This isn't a massive issue and it's probably tricky to avoid it since I bet the attraction_groups methods are not idempotent so if you feel like it would be too much work, it's okay.

vpr/src/pack/pack.cpp
```diff
@@ -145,7 +145,39 @@ bool try_pack(const t_packer_opts& packer_opts,
 
     g_vpr_ctx.mutable_atom().mutable_lookup().set_atom_pb_bimap_lock(true);
 
-    while (true) {
+    /**
+     * @brief Enumeration for the state of the packer.
+     *
+     * If the packer fails to find a dense enough packing, depending on how the
+     * packer failed, the packer may iteratively retry packing with different
+     * setting to try and find a denser packing. These states represent a
```
vaughnbetz: setting -> settings

vpr/src/pack/pack.cpp
```diff
@@ -145,7 +145,39 @@ bool try_pack(const t_packer_opts& packer_opts,
...
+     * different type of iteration.
+     */
+    enum class e_packer_state {
+        /// @brief Default packer state.
+        DEFAULT,
+        /// @brief Succcess state for the packer.
```
vaughnbetz: Suggest defining success: packing looks feasible to fit on the device (doesn't exceed number of blocks of each type in the grid) and to meet floorplanning constraints.

vpr/src/pack/pack.cpp
```diff
@@ -145,7 +145,39 @@ bool try_pack(const t_packer_opts& packer_opts,
...
+        ///        on unrelated clustering and balanced packing if it can.
+        UNRELATED_AND_BALANCED,
+        /// @brief Region constraints: Turns on attraction groups for overfilled regions.
+        ATTRACTION_GROUPS,
+        /// @brief Region constraints: Turns on more attraction groups for overfilled regions.
+        MORE_ATTRACTION_GROUPS,
+        /// @brief Region constraints: Turns on more attraction groups for all regions.
+        ATTRACTION_GROUPS_ALL_REGIONS,
+        /// @brief Region constraints: Turns on more attraction groups for all regions
+        ///        and increases the target density of clb blocks.
```
vaughnbetz: clb blocks?  Probably should say "any overused block types"

vpr/src/pack/pack.cpp
```diff
@@ -170,48 +202,98 @@ bool try_pack(const t_packer_opts& packer_opts,
...
+                        next_packer_state = e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS_AND_INCREASED_TARGET_DENSITY;
+                        break;
+                    case e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS_AND_INCREASED_TARGET_DENSITY:
+                    default:
+                        next_packer_state = e_packer_state::FAILURE;
+                        break;
+                }
+            } else {
+                // If there are no overfilled region constraints.
+                switch (current_packer_state) {
```
vaughnbetz: What if I had a floorplan_not_fitting state that I've now resolved (with attraction groups), but then I still find I don't fit on the device.  I think we only try turning UNRELATED clustering on if I didn't create any attraction groups (I was in state DEFAULT).  Those shouldn't be coupled; we should try all density-improving options in that case.

Not sure if that is easiest to implement with more state checks in the else, or if two separate states (one monitoring attraction group status and another the overall density-targeting status) would be better.

vpr/src/pack/pack.cpp
```diff
@@ -170,48 +202,98 @@ bool try_pack(const t_packer_opts& packer_opts,
...
+                    case e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS:
+                        next_packer_state = e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS_AND_INCREASED_TARGET_DENSITY;
+                        break;
+                    case e_packer_state::ATTRACTION_GROUPS_ALL_REGIONS_AND_INCREASED_TARGET_DENSITY:
+                    default:
+                        next_packer_state = e_packer_state::FAILURE;
+                        break;
+                }
+            } else {
+                // If there are no overfilled region constraints.
```
vaughnbetz: augment comment: "but some block types on the grid are overfilled."

vpr/src/pack/pack.cpp
```diff
@@ -224,9 +306,18 @@ bool try_pack(const t_packer_opts& packer_opts,
                 //       do it for all block types. Doing it only for a clb
                 //       string is dangerous -VB.
                 cluster_legalizer.get_target_external_pin_util().set_block_pin_util("clb", pin_util);
+                break;
```
vaughnbetz, line 349: Yes, we should fix this.

vaughnbetz: We should really check which types of blocks are overused and increase their target_external_pin_util.

vpr/src/pack/pack.cpp
```diff
@@ -170,48 +202,98 @@ bool try_pack(const t_packer_opts& packer_opts,
...
+                 */
+                switch (current_packer_state) {
+                    case e_packer_state::DEFAULT:
+                        next_packer_state = e_packer_state::ATTRACTION_GROUPS;
+                        break;
+                    case e_packer_state::UNRELATED_AND_BALANCED:
+                    case e_packer_state::ATTRACTION_GROUPS:
+                        next_packer_state = e_packer_state::MORE_ATTRACTION_GROUPS;
+                        break;
+                    case e_packer_state::MORE_ATTRACTION_GROUPS:
```
vaughnbetz: I think right now if we fail to fit on the grid after UNRELATED_AND_BALANCED we will go to the MORE_ATTRACTION_GROUPS state. That will eventually lead to turning on INCREASED_TARGET_DENSITY, which is good, but it will also result in at least a couple of useless clustering attempts (with MORE_ATTRACTION_GROUPS and ATTRACTION_GROUPS_ALL_REGIONS, which won't change anything if there are no attraction groups).  Should change the control flow / state machine to avoid that.

vaughnbetz (changes requested):
I suggest some changes.

vpr/src/pack/pack.cpp
AlexandreSinger: The comment follows what is currently happening in this packer state. This one currently actually does do specifically "clb" blocks. This will change in a future PR, the goal of this PR is organization. It will be much easier to fix after this is in.

vpr/src/pack/pack.cpp
AlexandreSinger: I agree with you, this is something I noticed as well. Attraction groups should be turned on if we have overfilled region constraints, but we want unrelated clustering to turn on if we have too many of a given block type. I am planning on fixing this in the next PR.

vpr/src/pack/pack.cpp
AlexandreSinger: I already do this on the AP changes I am working on now. It will be very easy to bring in later in a future PR.

vpr/src/pack/pack.cpp
AlexandreSinger: Not exactly, I looked into this code to see exactly what it was doing. The current flow does the following:

1. Pack with default parameters (what I call default state)
2. If failed, if the region constraints are not overfilled (or if you do not have region constraints) turn on unrelated clustering and balanced block type usage. If you do have an overfilled region constraint it  will turn on attraction groups.
3. If failed, if you have no overused block types it fails hard. If you have overfilled regions it creates more attraction groups.
4. If failed, again if you have no overused block types it fails hard, if you do it creates more attraction groups.

This repeats with the attraction groups until the last try where it raises the target pin utilization. But this only occurs when there is an overfilled region constraint.

The code that did this was incredibly confusing and what you are noticing now is the issues with the iterative packing algorithm (since this cleanup makes it more apparent what it is doing).

I am avoiding making any changes to this flow right now since it requires a lot of testing and its a lot easier to clean the code first and then make the changes. There are already issues raised tracking this issue.

vpr/src/pack/pack.cpp
```diff
@@ -170,48 +202,98 @@ bool try_pack(const t_packer_opts& packer_opts,
...
-            if (packer_opts.allow_unrelated_clustering == e_unrelated_clustering::AUTO) {
-                VTR_ASSERT(allow_unrelated_clustering == false);
-                allow_unrelated_clustering = true;
+            // If everything fits on the device and the floorplan regions are
+            // not overfilled, the next state is success.
+            next_packer_state = e_packer_state::SUCCESS;
+        } else {
+            if (floorplan_not_fitting) {
+                // If there are overfilled region constraints.
+                /*
```
soheilshahrouz: To be consistent with the coding style guide, don't use `/* */` for implementation comments.

vpr/src/pack/pack.cpp
```diff
@@ -170,48 +202,98 @@ bool try_pack(const t_packer_opts& packer_opts,
...
-                VTR_ASSERT(balance_block_type_util == false);
-                balance_block_type_util = true;
+        }
+
+        // Set up for the next packer state.
+        switch (next_packer_state) {
+            case e_packer_state::UNRELATED_AND_BALANCED: {
+                //1st pack attempt was unsuccessful (i.e. not dense enough) and we have control of unrelated clustering
+                //
+                //Turn it on to increase packing density
```
soheilshahrouz: add space after `//` to be consistent with the coding style guide

vpr/src/pack/pack.cpp
```diff
@@ -255,14 +346,23 @@ bool try_pack(const t_packer_opts& packer_opts,
             VPR_FATAL_ERROR(VPR_ERROR_OTHER, "Failed to find device which satisfies resource requirements required: %s (available %s)", resource_reqs.c_str(), resource_avail.c_str());
         }
 
-        //Reset floorplanning constraints for re-packing
-        g_vpr_ctx.mutable_floorplanning().cluster_constraints.clear();
+        // If the packer was unsuccessful, reset the packed solution and try again.
+        if (next_packer_state != e_packer_state::SUCCESS) {
+            //Reset floorplanning constraints for re-packing
```
soheilshahrouz: space after `//`

vpr/src/pack/pack.cpp
AlexandreSinger: I agree, but this is not something I am able to fix right now. The options used for the next iteration of the packer requires information from the previous iteration. It is much cleaner to just set it up here. I am planning on leaving the way it is.

vpr/src/pack/pack.cpp
vaughnb-cerebras: OK, sounds good. It would be good to fix this!

PR #3164, opened 2025-06-24
RR Edge ID Verification

amin1377:
#### Description
In a `.route` file, each edge between two nodes includes a switch ID, which is used to retrieve the electrical characteristics of that edge. When reading the `.route` file, this switch ID is compared against the corresponding switch ID from the RR Graph for consistency. If they do not match, an error is raised.

This consistency check is generally helpful. However, when analyzing multiple timing corners, each corner may have a different RR Graph with varying timing information (graph topologies are the same). As a result, the number and ordering of switch **types** may differ between corners. To allow reuse of the same `.route` file for fair comparisons across different timing corners, this strict verification can become a limitation.

To address this issue, a new CLI parameter `verify_rr_switch_id` has been introduced. By default, it is set to `on`, enforcing the consistency check. When set to `off`, the switch ID from the `.route` file is ignored, and the switch ID from the RR Graph is used instead.

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -1637,6 +1637,16 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("on")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    gen_grp.add_argument<bool, ParseOnOff>(args.verify_rr_switch_id, "--verify_rr_switch_id")
+        .help(
+            "Verify that the switch IDs in the routing file are consistent with those in the RR Graph."
+            "Set this to false when switch IDs in the routing file may differ from the RR Graph."
+            "For example, when analyzing different timing corners using the same netlist, placement, and routing files,"
+            "the RR switch IDs in the RR Graph may differ due to changes in delays."
+            "In such cases, set this option to false so that the switch IDs from the RR Graph are used, and those in the routing file are ignored.")
+        .default_value("on")
+        .show_in(argparse::ShowIn::HELP_ONLY);
+
     gen_grp.add_argument(args.target_device_utilization, "--target_utilization")
         .help(
             "Sets the target device utilization."
--- a/vpr/src/base/read_options.h
+++ b/vpr/src/base/read_options.h
@@ -73,6 +73,7 @@ struct t_options {
     argparse::ArgValue<e_timing_update_type> timing_update_type;
     argparse::ArgValue<bool> CreateEchoFile;
     argparse::ArgValue<bool> verify_file_digests;
+    argparse::ArgValue<bool> verify_rr_switch_id;
     argparse::ArgValue<std::string> device_layout;
     argparse::ArgValue<float> target_device_utilization;
     argparse::ArgValue<e_constant_net_method> constant_net_method;
--- a/vpr/src/base/read_route.cpp
+++ b/vpr/src/base/read_route.cpp
@@ -63,6 +63,8 @@ static void process_nets(const Netlist<>& net_list,
                          bool verify_rr_switch_id,
                          bool is_flat);
 
+static void update_rr_switch_id(t_trace* trace, std::set<int>& seen_rr_nodes);
+
 static void process_global_blocks(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, const char* filename, int& lineno, bool is_flat);
 static void format_coordinates(int& layer_num, int& x, int& y, std::string coord, ClusterNetId net, const char* filename, const int lineno);
 static void format_pin_info(std::string& pb_name, std::string& port_name, int& pb_pin_num, const std::string& input);
@@ -422,10 +424,65 @@ static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterN
         oldpos = fp.tellg();
     }
 
+    if (verify_rr_switch_id) {
+        VTR_ASSERT(validate_traceback(head_ptr));
+    } else {
+        std::set<int> seen_rr_nodes;
+        update_rr_switch_id(head_ptr, seen_rr_nodes);
+    }
+
     /* Convert to route_tree after reading */
-    VTR_ASSERT(validate_traceback(head_ptr));
     route_ctx.route_trees[inet] = TracebackCompat::traceback_to_route_tree(head_ptr);
     free_traceback(head_ptr);
+}
+
+static void update_rr_switch_id(t_trace* trace, std::set<int>& seen_rr_nodes) {
+    if (!trace) {
+        return;
+    }
+
+    seen_rr_nodes.insert(trace->index);
+
+    t_trace* next = trace->next;
+
+    if (next) {
+        if (trace->iswitch == OPEN) { //End of a branch
+
+            //Verify that the next element (branch point) has been already seen in the traceback so far
+            if (!seen_rr_nodes.count(next->index)) {
+                VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Traceback branch point %d not found", next->index);
+            } else {
+                //Recurse along the new branch
+                update_rr_switch_id(next, seen_rr_nodes);
+                return;
+            }
+        } else { //Midway along branch
+
+            //Check there is an edge connecting trace and next
+
+            auto& device_ctx = g_vpr_ctx.device();
+            const auto& rr_graph = device_ctx.rr_graph;
+            bool found = false;
+            for (t_edge_size iedge = 0; iedge < rr_graph.num_edges(RRNodeId(trace->index)); ++iedge) {
+                int to_node = size_t(rr_graph.edge_sink_node(RRNodeId(trace->index), iedge));
+                if (to_node == next->index) {
+                    found = true;
+
+                    //Verify that the switch matches
+                    int rr_iswitch = rr_graph.edge_switch(RRNodeId(trace->index), iedge);
+                    trace->iswitch = rr_iswitch;
+                    break;
+                }
+            }
+            if (!found) {
+                VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Traceback no RR edge between RR nodes %d -> %d\n", trace->index, next->index);
+            }
+            //Recurse
+            update_rr_switch_id(next, seen_rr_nodes);
+            return;
+        }
+    }
+    VTR_ASSERT(!next);
 }
 
 /**
--- a/vpr/src/base/read_route.cpp
+++ b/vpr/src/base/read_route.cpp
@@ -39,9 +39,30 @@
 #include "old_traceback.h"
 
 /*************Functions local to this module*************/
-static void process_route(const Netlist<>& net_list, std::ifstream& fp, const char* filename, int& lineno, bool is_flat);
-static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, const char* filename, int& lineno);
-static void process_nets(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, std::string name, std::vector<std::string> input_tokens, const char* filename, int& lineno, bool is_flat);
+static void process_route(const Netlist<>& net_list,
+                          std::ifstream& fp,
+                          const char* filename,
+                          int& lineno,
+                          bool verify_rr_switch_id,
+                          bool is_flat);
+
+static void process_nodes(const Netlist<>& net_list,
+                          std::ifstream& fp,
+                          ClusterNetId inet,
+                          const char* filename,
+                          bool verify_rr_switch_id,
+                          int& lineno);
+
+static void process_nets(const Netlist<>& net_list,
+                         std::ifstream& fp,
+                         ClusterNetId inet,
+                         std::string name,
+                         std::vector<std::string> input_tokens,
+                         const char* filename,
+                         int& lineno,
+                         bool verify_rr_switch_id,
+                         bool is_flat);
+
 static void process_global_blocks(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, const char* filename, int& lineno, bool is_flat);
 static void format_coordinates(int& layer_num, int& x, int& y, std::string coord, ClusterNetId net, const char* filename, const int lineno);
 static void format_pin_info(std::string& pb_name, std::string& port_name, int& pb_pin_num, const std::string& input);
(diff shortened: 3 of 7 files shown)


vpr/src/base/read_options.cpp
```diff
@@ -1637,6 +1637,16 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("on")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    gen_grp.add_argument<bool, ParseOnOff>(args.verify_rr_switch_id, "--verify_rr_switch_id")
+        .help(
+            "Verify that the switch IDs in the routing file are consistent with those in the RR Graph."
```
soheilshahrouz: Add a space or '\n' at the end of each line.

soheilshahrouz: document this new option in the rst file

vpr/src/base/read_options.cpp
```diff
@@ -1637,6 +1637,16 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("on")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    gen_grp.add_argument<bool, ParseOnOff>(args.verify_rr_switch_id, "--verify_rr_switch_id")
```
soheilshahrouz: If this is a routing option, add it to `route_grp`

vpr/src/base/read_options.h
```diff
@@ -73,6 +73,7 @@ struct t_options {
     argparse::ArgValue<e_timing_update_type> timing_update_type;
     argparse::ArgValue<bool> CreateEchoFile;
     argparse::ArgValue<bool> verify_file_digests;
+    argparse::ArgValue<bool> verify_rr_switch_id;
```
soheilshahrouz: I think this option is only applicable to the routing stage and its variable should move there

vpr/src/base/read_route.cpp
```diff
@@ -39,9 +39,32 @@
 #include "old_traceback.h"
 
 /*************Functions local to this module*************/
-static void process_route(const Netlist<>& net_list, std::ifstream& fp, const char* filename, int& lineno, bool is_flat);
-static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, const char* filename, int& lineno);
-static void process_nets(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, std::string name, std::vector<std::string> input_tokens, const char* filename, int& lineno, bool is_flat);
+static void process_route(const Netlist<>& net_list,
```
soheilshahrouz, line 54: The definition of some of these have doxygen comments. Can you move them to the declaration?

vpr/src/base/read_route.cpp
```diff
@@ -39,9 +39,32 @@
...
+                         std::ifstream& fp,
+                         ClusterNetId inet,
+                         std::string name,
+                         std::vector<std::string> input_tokens,
+                         const char* filename,
+                         int& lineno,
+                         bool verify_rr_switch_id,
+                         bool is_flat);
+
+static void update_rr_switch_id(t_trace* trace, std::set<int>& seen_rr_nodes);
```
soheilshahrouz: Doxygen comment

vpr/src/base/read_route.cpp
soheilshahrouz: Thank you for breaking long lines. Can you do the same for the definition of these functions?

vpr/src/base/read_route.cpp
```diff
@@ -158,15 +186,15 @@ static void process_route(const Netlist<>& net_list, std::ifstream& fp, const ch
...
+            process_nets(net_list, fp, inet, tokens[2], tokens, filename, lineno, verify_rr_switch_id, is_flat);
         }
     }
 
     tokens.clear();
 }
 
 ///@brief Check if the net is global or not, and process appropriately
-static void process_nets(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, std::string name, std::vector<std::string> input_tokens, const char* filename, int& lineno, bool is_flat) {
+static void process_nets(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, std::string name, std::vector<std::string> input_tokens, const char* filename, int& lineno, bool verify_rr_switch_id, bool is_flat) {
```
soheilshahrouz: This line is too long.

soheilshahrouz: move the doxygen comment to declaration

vpr/src/base/read_route.cpp
```diff
@@ -199,12 +227,12 @@ static void process_nets(const Netlist<>& net_list, std::ifstream& fp, ClusterNe
...
         }
 
-        process_nodes(net_list, fp, inet, filename, lineno);
+        process_nodes(net_list, fp, inet, filename, verify_rr_switch_id, lineno);
     }
     input_tokens.clear();
 }
 
-static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, const char* filename, int& lineno) {
+static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterNetId inet, const char* filename, bool verify_rr_switch_id, int& lineno) {
```
soheilshahrouz: This line is too long

vpr/src/base/read_route.cpp
```diff
@@ -396,12 +424,67 @@ static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterN
...
+static void update_rr_switch_id(t_trace* trace, std::set<int>& seen_rr_nodes) {
+    if (!trace) {
+        return;
+    }
+
+    seen_rr_nodes.insert(trace->index);
+
+    t_trace* next = trace->next;
+
+    if (next) {
```
soheilshahrouz: I think you can use `if (next == nullptr)` to return early and get rid of this if statement. Less nestedness increases readablity

vpr/src/base/read_route.cpp
```diff
@@ -396,12 +424,67 @@ static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterN
...
+        if (trace->iswitch == OPEN) { //End of a branch
+
+            //Verify that the next element (branch point) has been already seen in the traceback so far
+            if (!seen_rr_nodes.count(next->index)) {
+                VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Traceback branch point %d not found", next->index);
+            } else {
```
soheilshahrouz: Is this else statement necessary?
I think we can remove it. Then, both branches of `if (trace->iswitch == OPEN)` call `update_rr_switch_id(next, seen_rr_nodes)`. So, we can move it outside the conditional statement.

vpr/src/base/read_route.cpp
```diff
@@ -396,12 +424,67 @@ static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterN
...
+                //Recurse along the new branch
+                update_rr_switch_id(next, seen_rr_nodes);
+                return;
+            }
+        } else { //Midway along branch
+
+            //Check there is an edge connecting trace and next
+
+            auto& device_ctx = g_vpr_ctx.device();
+            const auto& rr_graph = device_ctx.rr_graph;
```
soheilshahrouz: The only usage of `device_ctx` is to access `rr_graph`. 
`const auto& rr_graph = g_vpr_ctx.device().rr_graph` would be more concise.

vpr/src/base/read_route.cpp
```diff
@@ -396,12 +424,67 @@ static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterN
...
+            if (!seen_rr_nodes.count(next->index)) {
+                VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Traceback branch point %d not found", next->index);
+            } else {
+                //Recurse along the new branch
+                update_rr_switch_id(next, seen_rr_nodes);
+                return;
+            }
+        } else { //Midway along branch
+
+            //Check there is an edge connecting trace and next
```
soheilshahrouz: To be consistent with the coding style guide, add a space after `//`

vpr/src/base/vpr_types.h
```diff
@@ -1343,6 +1343,8 @@ struct t_router_opts {
 
     bool with_timing_analysis;
 
+    bool verify_rr_switch_id;
```
soheilshahrouz: doxygen comment

vpr/src/base/read_route.cpp
```diff
@@ -396,12 +424,67 @@ static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterN
...
+                    //Verify that the switch matches
+                    int rr_iswitch = rr_graph.edge_switch(RRNodeId(trace->index), iedge);
+                    trace->iswitch = rr_iswitch;
+                    break;
+                }
+            }
+            if (!found) {
+                VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Traceback no RR edge between RR nodes %d -> %d\n", trace->index, next->index);
+            }
+            //Recurse
```
soheilshahrouz: I think an iterative function that uses `std::stack` is easier to understand and eliminates the possiblity of stack overflow.
With an iterative function, you don't need to pass `std::set` as an argument, making the function self-contained.

soheilshahrouz (changes requested):
Thanks @amin1377.

I left a few comment for improving the readability a bit.

vpr/src/base/read_route.cpp
vaughnbetz: When moving the comment, see if you can improve it.  What does "process appropriately" mean?

vpr/src/base/read_route.cpp
```diff
@@ -396,12 +424,67 @@ static void process_nodes(const Netlist<>& net_list, std::ifstream& fp, ClusterN
         oldpos = fp.tellg();
     }
 
+    if (verify_rr_switch_id) {
+        VTR_ASSERT(validate_traceback(head_ptr));
+    } else {
+        std::set<int> seen_rr_nodes;
+        update_rr_switch_id(head_ptr, seen_rr_nodes);
```
vaughnbetz: It looks like update_rr_switch_id has much of the functionality of validate_traceback inside it.  I think it would be more clear to rename validate_traceback to validate_and_update_traceback and pass it a flag remap_switch_id which makes it do the remapping behaviour as it validates.

If it is safe to remap before validation (which it might not be if the traceback is invalid) you could alternatively make a more limited remapping function that you call before traceback validation.

vpr/src/base/vpr_types.h
vaughnbetz: I think you could name this better.  Isn't this controlling whether you verify the switch id in route files?  Hence a better name would be --verify_route_file_switch_id  (and you'd change this in the command line, documentation and code).

vpr/src/base/read_options.cpp
amin1377: I actually intentionally didn’t add it to the .rst file. Since it’s an advanced option, I wasn’t sure how much visibility we want to give users for this feature.

vpr/src/base/read_options.cpp
amin1377: Strictly speaking, it’s not a routing option. If set to false, it disables a default check and uses the switch ID from the RR graph instead. Because of this, I grouped it with other checks (like verify_file_digests, etc.). However, if you still think it would be better to include it under the router group, please let me know.

vpr/src/base/read_options.h
amin1377: Same response as above.

vaughnbetz:
I’d leave it grouped with verification options. I’d also document it in the .rst for consistency.VaughnOn Jul 1, 2025, at 11:35 AM, Amin Mohaghegh ***@***.***> wrote:﻿
@amin1377 commented on this pull request.



In vpr/src/base/read_options.cpp:
> @@ -1637,6 +1637,16 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("on")
 
+    gen_grp.add_argument<bool, ParseOnOff>(args.verify_rr_switch_id, "--verify_rr_switch_id")

Strictly speaking, it’s not a routing option. If set to false, it disables a default check and uses the switch ID from the RR graph instead. Because of this, I grouped it with other checks (like verify_file_digests, etc.). However, if you still think it would be better to include it under the router group, please let me know.

vpr/src/base/old_traceback.cpp
```diff
@@ -141,70 +141,64 @@ void print_traceback(const t_trace* trace) {
...
 
-    if (next) {
-        if (trace->iswitch == OPEN) { //End of a branch
+        if (next == nullptr) {
+            continue;
+        }
 
+        if (trace->iswitch == OPEN) { //End of a branch
             //Verify that the next element (branch point) has been already seen in the traceback so far
             if (!seen_rr_nodes.count(next->index)) {
```
vaughnb-cerebras, line 165: Should check that index is a valid rr_node index (otherwise the validator could crash on invalid input). Print a helpful error message if it is out of range.

vpr/src/base/old_traceback.h
```diff
@@ -46,4 +46,14 @@ class TracebackCompat {
 t_trace* alloc_trace_data();
 void free_traceback(t_trace* trace);
 void print_traceback(const t_trace* trace);
-bool validate_traceback(t_trace* trace);
+
+/**
+ * @brief Validate the routing traceback if verify_switch_id is true, otherwise update the
```
vaughnb-cerebras: Should update this comment.

"Validate the integrity of the traceback rooted a trace: it should contain only valid rr nodes, branches in the routing tree
should link to existing routing, and edges in the traceback should exist in the RRGraph.  
If verify_switch_id is true, this routine also checks that the switch types used in the traceback match those in the 
RRGraph. If verify_switch_id is false, the switch ids (types) are remapped to those in the RRGraph. This switch remapping is 
useful when an RRGraph with a more detailed delay model (and hence more switch types) is used with a prior routing."

vaughnb-cerebras (changes requested):
Thanks.  I suggest a couple more changes.

One more thought: should we validate the switch_ids sometimes and not others (what the code is currently doing) or should we just always recompute them on reading in a route file?  We don't have to change it, but thought I'd ask ... maybe we're writing out redundant information and then correcting it.  In that case not writing out the redundant information at all might be better (although changing the .route format would cause some backwards-incompatibility).

vpr/src/base/old_traceback.cpp
amin1377: `validate_and_update_traceback` is called within the `process_nodes` function. In this function, all relevant checks related to the node are performed when creating the trace.

PR #3167, opened 2025-06-26
[Packer] Updated the Next State Logic of Iterative Re-Packing

### LINKED ISSUE #2729: [Packer] Setting Higher Target Pin Utilization When Regions are Full

AlexandreSinger:
The packer iteratively re-packs the netlist if it fails to create a dense enough packing. The way it does this can be thought of like states, where each state turns on a new set of packer options that can create a denser clustering.

The prior state transition logic for the iterative re-packing did not turn on enough options to make the clustering as dense as VTR can manage. This updates the state transition logic to create even more dense clusterings.

The next state logic is as follows:
- If there is an over-utilized block type, turn on unrelated clustering and balanced block type utilization if able.
- If there are overfilled floorplan regions, create more attraction groups than the prior iteration (or increase the pull). If all regions have attraction groups and the pull is strong, do not increase further.
- Finally, if all else fails, increase the target external pin utilization if any of the overfilled block types have target pin utilizations less than 1.0.
- If all prior fails, fail hard.

Found that for the Z1000 architecture, this increased the maximum LUT utilization from around 68% to around 86%.

Some data from the Z1000 architecture:

- Before this change, only the 4 smallest MCNC benchmarks could pack on the Z1000 architecture:

| circuit | num LUTs | Lut Utilization |
| ------ | ----------- | ------------- |
| tseng | 1046 | 51.1% |
| ex5p | 1064 | 52.0% |
| apex4 | 1262 | 61.6% |
| misex3 | 1397 | 68.2% |

- After this change, 6 more circuits were able to be packed on the Z1000 architecture:

| circuit | num LUTs | Lut Utilization |
| ------ | ----------- | ------------- |
| dsip | 1362 | 66.5% |
| diffeq | 1494 | 73.0% |
| alu4 | 1522 | 74.3% |
| des | 1591 | 77.7% |
| bigkey | 1699 | 83.0% |
| seq | 1750 | 85.5% |

- Two MCNC circuits can still theoretically fit on the Z1000, but are still failing to be packed:

| circuit | num LUTs | Lut Utilization |
| ------ | ----------- | ------------- |
| apex2 | 1878 | 91.7% |
| s298 | 1930 | 94.0% |

Resolves #2729


AlexandreSinger:
@petergrossmann21 FYI

PR #3170, opened 2025-06-27
Add CHANZ type

soheilshahrouz:
Adds CHANZ RR node type for inter-layer connections. Previously, these connection were modeled with CHANX nodes whose ptc number started from the maximum channel width.


vaughnbetz:
Should that issue (CHANZ rr_indexed_data is wrong) be fixed in this PR too?

libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -437,6 +432,14 @@ void check_rr_node(const RRGraphView& rr_graph,
             }
             break;
 
+        case e_rr_type::CHANZ:
+            if (xhigh != xlow || yhigh != ylow || xhigh > int(grid.width()) - 2 || ylow < 1 || yhigh > int(grid.height()) - 2) {
```
vaughnbetz: Change to width -1, height -1.
We'll allow CHANZ wires right up to the edge of the grid.

libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -437,6 +432,14 @@ void check_rr_node(const RRGraphView& rr_graph,
...
+                VPR_FATAL_ERROR(VPR_ERROR_ROUTE,
+                                "Error in check_rr_node: CHANZ out of range for endpoints (%d,%d) and (%d,%d)\n", xlow, ylow, xhigh, yhigh);
+            }
+            // TODO: handle global routing case
```
vaughnbetz, line 443: Can remove this TODO (this code would work for global routing).

libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -503,14 +495,18 @@ void check_rr_node(const RRGraphView& rr_graph,
             }
             break;
 
+        case e_rr_type::CHANZ:
+            // TODO: do checks for CHANZ type
```
vaughnbetz: can make tracks_per_node = 1 if we don't have global routing.
If global routing, not sure .... leave a TODO for that case.
Once you have a way to specify how many chanz nodes per switch block / channel, that is probably what you'd use to set the tracks_per_node for this case too for global routing.

vaughnbetz: Or just leave it unchecked in the global case forever ...

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -236,13 +236,19 @@ class RRGraphView {
...
 
     /** @brief Return the length (number of grid tile units spanned by the wire, including the endpoints) of a specified node.
-     * @note node_length() only applies to CHANX or CHANY and is always a positive number
+     * @note node_length() only applies to CHANX or CHANY or CHANZ and is always a positive number
      */
     inline int node_length(RRNodeId node) const {
-        VTR_ASSERT(node_type(node) == e_rr_type::CHANX || node_type(node) == e_rr_type::CHANY);
+        VTR_ASSERT(node_type(node) == e_rr_type::CHANX || node_type(node) == e_rr_type::CHANY || node_type(node) == e_rr_type::CHANZ);
+
+        if (node_type(node) == e_rr_type::CHANZ) {
```
vaughnbetz, line 250: Put a comment here ... it's a bit mysterious otherwise. Explain the why the length is always 1.

libs/librrgraph/src/base/rr_node_types.h
```diff
@@ -29,18 +29,22 @@ enum class e_rr_type : unsigned char {
     OPIN,       ///<Output pin of a block
     CHANX,      ///<x-directed routing wire, or an x-directed segment of a channel for global routing
     CHANY,      ///<y-directed routing wire, or a y-directed segment of a channel for global routing
+    CHANZ,      ///<z-directed routing wire used to connect two different layers
```
vaughnbetz: Comment that xlow will always == xhigh, and ylow == yhigh. To save storage it also has only one layer, which indicates the starting layer if it is a unidirectional wire.

vpr/src/route/check_route.cpp
```diff
@@ -463,14 +478,34 @@ static bool check_adjacent(RRNodeId from_node, RRNodeId to_node, bool is_flat) {
...
+                VPR_FATAL_ERROR(VPR_ERROR_ROUTE,
+                                "in check_adjacent: %d and %d are not adjacent", from_node, to_node);
+            }
+            break;
+
+        case e_rr_type::CHANZ:
+            if (to_type == e_rr_type::CHANX || to_type == e_rr_type::CHANY) {
+                num_adj += chanxy_chanz_adjacent(to_node, from_node);
+            } else if (to_type == e_rr_type::CHANZ) {
+                if (from_xlow == to_xlow && from_xhigh == to_xhigh && from_ylow == to_ylow && from_yhigh == to_yhigh && std::abs(from_layer - to_layer) == 1) {
```
vaughnbetz: Put a TODO here: I think you'll have to widen the range of how close CHANX and CHANY can be eventually.
TODO: also allow CHANZ to CHANZ, for future proofing.

vpr/src/route/check_route.cpp
```diff
@@ -463,14 +478,34 @@ static bool check_adjacent(RRNodeId from_node, RRNodeId to_node, bool is_flat) {
...
+                    num_adj++;
+                }
             } else {
+                std::cout << describe_rr_node(device_ctx.rr_graph, device_ctx.grid, device_ctx.rr_indexed_data, from_node, is_flat) << std::endl;
+                std::cout << describe_rr_node(device_ctx.rr_graph, device_ctx.grid, device_ctx.rr_indexed_data, to_node, is_flat) << std::endl;
```
vaughnbetz: Remove duplicate line.

vpr/src/route/check_route.cpp
```diff
@@ -463,14 +478,34 @@ static bool check_adjacent(RRNodeId from_node, RRNodeId to_node, bool is_flat) {
                     /* UDSD Modification by WMF End */
                 }
             } else if (to_type == e_rr_type::CHANX) {
-                num_adj += chanx_chany_adjacent(to_node, from_node);
+                num_adj += rr_graph.nodes_are_adjacent(to_node, from_node);
+            } else if (to_type == e_rr_type::CHANZ) {
+                num_adj += chanxy_chanz_adjacent(from_node, to_node);
```
vaughnbetz: Let's push this code into .nodes_are_adjacent. Can have it call the current helper after figuring out there is a CHANZ and swapping nodes (or do the swapping at the top of the current routine).

vpr/src/route/check_route.cpp
```diff
@@ -866,6 +887,30 @@ void check_net_for_stubs(const Netlist<>& net_list,
     }
 }
 
+static bool chanxy_chanz_adjacent(RRNodeId chanxy_node, RRNodeId chanz_node) {
```
vaughnbetz: Move to rr_graph_utils or something (or make externally visible and call from nodes_are_adjacent).

vpr/src/route/rr_graph_generation/rr_graph_area.cpp
```diff
@@ -361,17 +351,13 @@ void count_unidir_routing_transistors(std::vector<t_segment_inf>& /*segment_inf*
...
-        num_inputs_to_cblock[c] = 0;
-
     maxlen = std::max(device_ctx.grid.width(), device_ctx.grid.height());
     cblock_counted = new bool[maxlen];
     for (auto k = 0; k < maxlen; k++)
         cblock_counted[k] = 0;
 
     ntrans = 0;
-    for (const RRNodeId& from_rr_node : device_ctx.rr_graph.nodes()) {
+    for (const RRNodeId from_rr_node : device_ctx.rr_graph.nodes()) {
```
vaughnbetz, line 360: Add CHANZ to the switch for from_rr_type and to_rr_type

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
```diff
@@ -516,6 +515,24 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
                                           y,
                                           describe_rr_node(rr_graph, grid, rr_indexed_data, inode, is_flat).c_str());
                             }
+                        } else if (rr_graph.node_type(inode) == e_rr_type::CHANZ) {
+                            VTR_ASSERT_MSG(rr_graph.node_xlow(inode) == rr_graph.node_xhigh(inode), "CHANZ should move only along layers");
+                            VTR_ASSERT_MSG(rr_graph.node_ylow(inode) == rr_graph.node_yhigh(inode), "CHANZ should move only along layers");
+
+                            if (x != rr_graph.node_xlow(inode)) {
```
vaughnbetz, line 542: Check if this cross check of node.x (index_lookup (x,y,ptc) == x can be factored out into a general loop (same for CHANX, CHANY, etc. --> loop over rr_node indices and make sure you find the right node, loop over nodes and make sure the index is rr_node_indices.

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
```diff
@@ -516,6 +515,24 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
...
+                                          describe_rr_node(rr_graph, grid, rr_indexed_data, inode, is_flat).c_str());
+                            }
+
+                            if (y != rr_graph.node_ylow(inode)) {
+                                VPR_ERROR(VPR_ERROR_ROUTE, "RR node y position does not agree between rr_nodes (%d) and rr_node_indices (%d): %s",
+                                          rr_graph.node_xlow(inode),
+                                          y,
+                                          describe_rr_node(rr_graph, grid, rr_indexed_data, inode, is_flat).c_str());
+                            }
+
```
vaughnbetz, line 555: 1. Add a check for layer matching between the node and look up (one check should work for all rr_types, so just one if).
2. We could factor out the code at the end of this routine  that checks rr_node_counts into a local/static helper function that just takes a ref to rr_node_counts as an input.

libs/librrgraph/src/base/check_rr_graph.cpp
soheilshahrouz: What about CHANX and CHANY?

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
soheilshahrouz: I am a bit confused about this. Can you elaborate on this?

soheilshahrouz:
3d_sb_titan_other_auto_bb

|  | place_time | total_swaps | placed_wirelength_est | placed_CPD_est | route_time | CPD | heap_pops | WL |
|--------|------------|-------------|------------------------|----------------|------------|-----|-----------|-----|
| ratio  | 0.9863     | 1.0129      | 1.0136                 | 1.1523         | 0.3736     | 0.9958  | 0.2993     | 0.9733 |

libs/librrgraph/src/base/check_rr_graph.cpp
vaughnbetz: Can probably relax them too.

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
vaughnbetz: OK to leave if it doesn't make sense. I was thinking that the cross-check of rr_node_indices vs. rr_node could be more generic (check that every node's x,y leads to lookups that match for that (x,y)).  But if it's a pain it can be left alone. The current code both checks consistency and some properties of CHANX, CHANY, etc.; if recoded to separate out the consistency part, then the consistency part could be more generic.

vaughnbetz:
Only failure; expected QoR failure on a 3D placement estimated delay. Update and merge!

soheilshahrouz:
I ran mcnc benchmarks to make sure that adding CHANZ does not impact 2D architectures. I get the same QoR on both branches.

PR #3172, opened 2025-06-30
[NetlistWriter] Fixed Failing Nightly Test

AlexandreSinger:
The nightly test for the netlist writer was failing due to changes in the clock to Q delays. After some digging, I found that there was an oversight in the netlist writer code when it generates ram black boxes.

Someone hard-coded the name of the clock port to "clock", when in reality the port's name is "clk". This was a very simple fix.

This resolves the failures in NightlyTest2:

image


AmirhosseinPoolad:
LGTM

PR #3180, opened 2025-07-02
Stratix 10 Quad Port Ram Support

kimiatkh:
This pull request adds support for quad-port RAM in the Stratix 10 architecture file. It also updates the vqm2blif tool to process quad-port RAMs from VQM files and generate the appropriate operation modes in the BLIF output. Additionally, it makes a small modification to the MLAB primitives in the architecture capture by adding the class=memory attribute, enabling VPR to correctly pack MLABs slices in to MLAB primitives.


utils/vqm2blif/src/base/vqm2blif_util.cpp
```diff
@@ -568,6 +568,16 @@ void generate_opname_ram (t_node* vqm_node, const LogicalModels& arch_models, st
                     cout << "   port A output sequential: " << bool(ram_info.port_a_output_clock) << "\n";
                     cout << "   port B output sequential: " << bool(ram_info.port_b_output_clock) << "\n";
                 }
+            } else {
+                if (ram_info.port_a_output_clock && ram_info.port_b_output_clock) {
+                    reg_outputs = true; //Sequential output
+                } else if (!ram_info.port_a_output_clock && !ram_info.port_b_output_clock) {
+                    reg_outputs = false; //Comb output
```
soheilshahrouz, line 575: Please edit the comments in your code to be consistent the coding style guide:

1) Add a space after `//`
2) Avoid using block comments for implementation-related commnets

utils/vqm2blif/src/base/vqm2blif_util.cpp
```diff
@@ -613,8 +625,9 @@ void generate_opname_ram (t_node* vqm_node, const LogicalModels& arch_models, st
         }
 
     //A dual port memory, both port A and B params have been found
-    } else {
-        VTR_ASSERT(ram_info.mode == "dual_port" || ram_info.mode == "bidir_dual_port");
+    } else if (ram_info.mode == "dual_port" || ram_info.mode == "bidir_dual_port"){
```
soheilshahrouz, line 628: missing space ){

utils/vqm2blif/src/base/vqm2blif_util.cpp
```diff
@@ -645,6 +658,53 @@ void generate_opname_ram (t_node* vqm_node, const LogicalModels& arch_models, st
...
             tmp_mode_hash.append(".port_b_data_width{" + std::to_string(ram_info.port_b_data_width) + "}");
 
+            LogicalModelId arch_model_id = arch_models.get_model_by_name(tmp_mode_hash);
+            if (!arch_model_id.is_valid()) {
+                //3a) Not found, use the default name (no specific address/data widths)
+                ; // do nothing
+            } else {
+                //3b) Use the more detailed name, since it was found in the architecture
+                mode_hash = tmp_mode_hash;
+            }
```
soheilshahrouz, line 668: re-write this if statement get rid of the branch where nothing is done

utils/vqm2blif/src/base/vqm2blif_util.cpp
```diff
@@ -568,6 +568,16 @@ void generate_opname_ram (t_node* vqm_node, const LogicalModels& arch_models, st
                     cout << "   port A output sequential: " << bool(ram_info.port_a_output_clock) << "\n";
                     cout << "   port B output sequential: " << bool(ram_info.port_b_output_clock) << "\n";
                 }
+            } else {
```
vaughnbetz, line 571: Suggest a comment of what the else is:
// quad port mode

vtr_flow/arch/titan/stratix10_arch.timing.xml
```diff
@@ -4889,6 +4887,919 @@
...
+        <port clock="clk_portain" combinational_sink_ports="eccstatus portadataout portbdataout " name="portaaddr"/>
+        <port clock="clk_portain" combinational_sink_ports="eccstatus portadataout portbdataout " name="portaaddr2"/>
+        <port clock="clk_portain" combinational_sink_ports="eccstatus portadataout portbdataout " name="portawe"/>
+        <port clock="clk_portain" combinational_sink_ports="eccstatus portadataout portbdataout " name="portare"/>
+        <port clock="clk_portain" combinational_sink_ports="eccstatus portadataout portbdataout " name="portabyteenamasks"/>
+        <port clock="clk_portain" combinational_sink_ports="eccstatus portadataout portbdataout " name="portaaddrstall"/>
+        <port clock="clk_portbin" combinational_sink_ports="eccstatus portadataout portbdataout " name="portbdatain"/>
+        <port clock="clk_portbin" combinational_sink_ports="eccstatus portadataout portbdataout " name="portbaddr"/>
+        <port clock="clk_portbin" combinational_sink_ports="eccstatus portadataout portbdataout " name="portbaddr2"/>
+        <port clock="clk_portbin" combinational_sink_ports="eccstatus portadataout portbdataout " name="portbwe"/>
```
vaughnbetz: I think it is worth putting a comment at the top of the quad port section explaining what a quad port S10 RAM is.  It is really simple quad port (2 read ports, 2 write ports).  So you have 4 address buses, two datain and two dataout.  
It looks like when you have combinational outputs (not registered outputs) you assume portaaddr affects both portadataout and portbdata.  Same for portaaddr2. We also have the portbaddr and portbaddr2 affecting both portadtaouta and portbdataout.  That is overly conservative: we'll only use one address port to choose the read out data.  Not sure if it is always the same one or not; vqm2blif could probably always write it on one of the portaaaddr buses for the data a side, and one of the portbaddr buses for the port b side.  Then portadataout and portbdataout would only be affected by the one chosen address port each.

Right now portadatain and portbdatain are always affecting the timing of portadataout and portbdataout when the outputs aren't registered.  That is also overly conservative; the datain values will be stored in the RAM but (by default at least) Quartus assumes they won't flow through to the outputs the same cycle (that's called read-during-write behaviour -- Quartus has settings to choose which one you assume, but I think we should assume the less conservative ones where dataina does not affect dataouta etc.

Suggest updating the timing section here.

vaughnbetz:
Adding @AlexandreSinger as he may be able to help with this, plus documenting the Titanium benchmarks and how to run them etc. We'd like to get this in soon @kimiatkh so we can refer to Titanium in an upcoming paper.

AlexandreSinger:
I think this PR is good to merge just to get the architecture fixes into Master. We can work on releasing a new version of Titan after (since these architecture changes work with the old version of Titanium as well). Documentation can come in another PR and should not block this PR in my opinion. Will merge once CI completes.

kimiatkh:
@vaughnbetz I've already added the comments you requested for the quad-port RAMs.

I've also verified the MLAB packing. For example, in cholesky_bdti, there are 342 memory slices packed into 172 MLAB cells, with each cell able to hold up to two slices. This suggests that the packing is working correctly.

However, with the new version merged in, some of the benchmarks that contain MLABs in Titanium and Titan23 will now fail. I’ve prepared the newly generated netlist archive, but before uploading it I need to make a small change to the script that downloads the Titan BLIFs. Since this branch is now closed, I’ll open a new branch for that update.

The golden results also need to be updated with the new MLAB statistics. I already have the new golden results for Titan23, and the run for Titanium is still ongoing but should finish by the end of today.

vaughnbetz:
Hi @kimiatkh : There are ten mlab primitives in an MLAB (ram slices), each of which can be up to 2 bits wide.  So we should get a bunch of 20-bit wide RAMs in MLABs.  Is that happening now?  If not, we likely have a regression.
@AlexandreSinger : tagging you too.

vaughnbetz:
Thanks @kimiatkh !  That sounds like proper behaviour.
There will be some narrower RAMs, so having some at 15 bits etc. seems fine.

Issue #3108, opened 2025-06-04
Embed slang-yosys plugin as a SystemVerilog frontend for yosys


tarikgraba:
Is it possible to embed the slang base frontend for yosys?

https://github.com/povik/yosys-slang

It has better support for SystemVerilog constructs, which would make v-to-r easier to use with pure SystemVerilog project.

It need a more recent version of yosys (>0.44) and I do not now if this is a hard contraint.

Thanks

vaughnbetz:
@loglav03 is working on updating the yosys front end. Logan any thoughts?

loglav03:
Yosys is currently being updated to the latest version (0.53) and there will be a discussion next week about possibly implementing yosys-slang.

vaughnbetz:
We did update yosys and add vtr-flow documentation for yosys-slang so I think this is done. It's marked experimental for now though.

Issue #2821, opened 2024-11-21
System Verilog support is broken due to compilation error in F4PGA plugin


MohamedElgammal:
The System verilog is broken as it requires enabling `YOSYS_F4PGA_PLUGINS` and doing so is getting a compilation error as follows:

vaughnbetz:
@soheilshahrouz : in case you feel like a tour of some yosys + C++ stuff ...

hzeller:
Isn't the systemverilog plugin called [synlig](https://github.com/chipsalliance/synlig) for a while now ? I suspect this pulls in a somewhat old version...

amirarjmand93:
I have been testing Synlig for integration, and so far it appears to be a suitable tool for SystemVerilog, rather than using it as just a standalone parser and elaborator, like Surlog. I will provide updates here along with some screenshots of my progress asap.

amirarjmand93:
![image_2024-11-24_13-59-09](https://github.com/user-attachments/assets/0a4e6988-0e3c-4e8d-83ad-5ac19569e798)

@vaughnbetz I executed some SystemVerilog designs of the VTR flow benchmark. It passed from Synlig, Parmys, Yosys, and VPR end-to-end successfully. It seems OK, but it takes more time to integrate fully. I am trying to solve it.

vaughnbetz:
That's good news! Thanks @amirarjmand93.

amirarjmand93:
I am writing to provide an update regarding the integration of **Synlig**.

#1
 It seems that a race condition is causing a nondeterministic situation, leading to errors with some nested submodules during the synchronization and update process. In some cases, I encounter no errors with the same configuration.

#2
Sometimes, I encounter the error (RPC failed) and similar issues when cloning(non-fatal error), even though the network is functioning well and the Postbuffer size has been increased. 

![image_2024-12-03_14-20-21](https://github.com/user-attachments/assets/db57412b-df40-46e4-aa62-c601d5f869ae)

It's in progress.

vaughnbetz:
The clone error sounds like a timeout/disconnect issue with the chipsalliance repo.  Adding @hzeller in case he knows why that repo is disconnecting before the clone is done. @amirarjmand93 : I imagine we could switch from a submodule to a subtree for this dependency if necessary, as I believe a subtree would actually put the files in the VTR repo, and hence cloning should happen from our repo and hopefully be robust.

vaughnbetz:
[...]

PR #3181, opened 2025-07-03
Switching from Synlig to Yosys-Slang

loglav03:
#### Description

Replaced Synlig/Surelog with the Yosys-Slang plugin for Yosys for reading SystemVerilog.

Yosys-Slang is installed by default in the build system and depends on Yosys. To use Yosys-Slang as the parser instead of the default Yosys parser, add the '-parser slang' option at the end of your 'run_vtr_flow.py \<circuit\> \<arch\>' or add it to the script parameters in the config file for regression test suites.

Yosys-Slang can be used to parse both verilog and SystemVerilog but may require specific syntax when parsing with verilog.

If you disable the cmake param "-DWITH_PARMYS" you should also disable the "-DSLANG_SYSTEMVERILOG" param as well to avoid cmake errors.
#### Related Issue

1. https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2821
2. https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3108

#### Motivation and Context

Synlig has a lot of overhead when building with the vtr build system leading to issues when building in CI and sometimes locally. Synlig is also poorly maintained. Yosys-Slang has significantly less overhead, can easily be implemented as a plugin for yosys, and is regularly maintained.

Has been tested with the vtr_reg_system_verilog tests [PASSED]

#### Types of changes

#### Checklist:

- [ x] I have added tests to cover my changes


.github/workflows/test.yml
```diff
@@ -96,7 +96,7 @@ jobs:
 
     - name: Build
       env:
-        CMAKE_PARAMS: "-DCMAKE_COMPILE_WARNING_AS_ERROR=on -DVTR_IPO_BUILD=off -DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on -DWITH_PARMYS=OFF -DWITH_ODIN=on"
+        CMAKE_PARAMS: "-DCMAKE_COMPILE_WARNING_AS_ERROR=on -DVTR_IPO_BUILD=off -DVTR_ASSERT_LEVEL=3 -DWITH_BLIFEXPLORER=on -DWITH_PARMYS=OFF -DSLANG_SYSTEMVERILOG=OFF -DWITH_ODIN=on"
```
AlexandreSinger: Since we are deciding to make SLANG_SYSTEMVERILOG on by default. I think it would make sense to have "-DWITH_PARMYS=OFF" also imply "-DSLANG_SYSTEMVERILOG=OFF". I think it would be a bit confusing to have to turn them both off.

CMakeLists.txt
```diff
@@ -438,7 +438,7 @@ if(${WITH_ODIN})
 endif()
 
 # handle cmake params to compile Yosys SystemVerilog/UHDM plugins
-if(${SYNLIG_SYSTEMVERILOG})
+if(${SLANG_SYSTEMVERILOG})
```
AlexandreSinger, line 441: Related to my comment above. Here you can emit a warning saying something like "Parmys is set to off, turning SLANG off too". You can use another variable to organize this to prevent people from using the wrong variable. Like "USE_SLANG_SYSTEMVERILOG" or something.

libs/EXTERNAL/CMakeLists.txt
```diff
@@ -84,93 +83,40 @@ if (${WITH_PARMYS})
...
-            GIT_REPOSITORY https://github.com/chipsalliance/synlig.git
-            GIT_TAG main
-            GIT_PROGRESS TRUE
-            GIT_SHALLOW TRUE
-
-            # Set source and build directories
-            SOURCE_DIR "${SYNLIG_SOURCE_DIR}"
-            BUILD_IN_SOURCE FALSE
-            INSTALL_DIR " "
+    if (${SLANG_SYSTEMVERILOG})
```
AlexandreSinger, line 88: Looking at the release build, I am seeing a lot of warnings coming from yosys-slang: https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/16204611463/job/45751741631?pr=3181

We really do not care about these warnings since this is an external project, we do not really care about these warnings and we want to keep the build warning clean. Can you please suppress these warnings?

libs/EXTERNAL/CMakeLists.txt
```diff
@@ -84,93 +83,40 @@ if (${WITH_PARMYS})
...
+        set(SLANG_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/yosys-slang)
+        set(SLANG_BUILD_DIR ${CMAKE_BINARY_DIR}/yosys-slang)
+        set(YOSYS_CONFIG_EXECUTABLE ${CMAKE_BINARY_DIR}/bin/yosys-config CACHE STRING "Path to the yosys-config helper" FORCE)
+        #Initialize yosys-slang submodules
+        execute_process(
+	        COMMAND git submodule update --init
```
AlexandreSinger: This `git` should probably be `${GIT_EXECUTABLE}`, right?

This is also a problem in the yosys commands above.

libs/EXTERNAL/CMakeLists.txt
```diff
@@ -84,93 +83,40 @@ if (${WITH_PARMYS})
...
+	        WORKING_DIRECTORY ${SLANG_SRC_DIR}
+	    )
+	    execute_process(
+	        COMMAND ${GIT_EXECUTABLE} submodule foreach --recursive git\ submodule\ update\ --init
```
AlexandreSinger: This `git` should also be `${GIT_EXECUTABLE}` at the end of this line.

libs/EXTERNAL/CMakeLists.txt
```diff
@@ -84,93 +83,40 @@ if (${WITH_PARMYS})
...
-            
-            LOG_BUILD ON
-            LOG_UPDATE ON
-            LOG_INSTALL ON
-            LOG_CONFIGURE OFF
-            LOG_OUTPUT_ON_FAILURE ON
-
-            # Ensure dependencies like Yosys are built first
-            DEPENDS yosys  # Ensure submodule sync runs before synlig build
+            BUILD_COMMAND ${CMAKE_COMMAND} --build ${SLANG_BUILD_DIR} -- -j${PROCESSOR_COUNT}
```
AlexandreSinger: "-j${PROCESSOR_COUNT}" should not be set here. The number of threads is enhereted by the make that creates it. I think this should not be here. This can have some odd affects. This is what is causing this warning:
image

libs/EXTERNAL/CMakeLists.txt
AlexandreSinger: Yosys also has some warnings. Can you also suppress those warnings as well?

libs/EXTERNAL/CMakeLists.txt
```diff
@@ -84,93 +83,40 @@ if (${WITH_PARMYS})
...
+	        COMMAND ${GIT_EXECUTABLE} submodule foreach --recursive git\ submodule\ update\ --init
+	        WORKING_DIRECTORY ${SLANG_SRC_DIR}
+	    )
+        set(SLANG_FE "${SLANG_SRC_DIR}/src/slang_frontend.cc")
+        ExternalProject_Add(
+            yosys-slang
+            SOURCE_DIR ${SLANG_SRC_DIR}
+            BINARY_DIR ${SLANG_BUILD_DIR}
+
+            #Disabling UndrivenPass in slang_frontend.cc
```
AlexandreSinger, line 106: Could you add some more comments here as to why we need to patch slang. Is this something that we can bring back to the slang master? Or is there something integral to VTR that causes problems?

libs/EXTERNAL/patch_slang.cmake
```diff
@@ -0,0 +1,9 @@
+if(NOT DEFINED IN)
```
AlexandreSinger, line 10: I would add a header to this file to explain what it is doing and why.

vtr_flow/tasks/regression_tests/vtr_reg_system_verilog/task_list.txt
```diff
@@ -1,3 +1,4 @@
 regression_tests/vtr_reg_system_verilog/f4pga_button_controller/
+#regression_tests/vtr_reg_system_verilog/koios_sv/
```
AlexandreSinger: Is this supposed to be commented out?

libs/EXTERNAL/CMakeLists.txt
loglav03: It actually uses -j1 if I remove the "-j${PROCESSOR_COUNT}". I get this when I remove it:
gmake[4]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.

.github/workflows/test.yml
loglav03: @AlexandreSinger Should be good now

CMakeLists.txt
loglav03: @AlexandreSinger Should be good now

libs/EXTERNAL/CMakeLists.txt
loglav03: @AlexandreSinger Yosys and yosys-slang warnings are now suppressed.

libs/EXTERNAL/CMakeLists.txt
loglav03: @AlexandreSinger These are now `${GIT_EXECUTABLE}`. Same goes for the comment below

libs/EXTERNAL/CMakeLists.txt
loglav03: I currently have it patching slang to disable the UndrivenPass because it was failing benchmarks due to unsupported synchronous rules in vtr_primitives.v and wouldn't let me run any benchmark with read_slang without disabling that pass.

I believe Peter is working on this in his work to clean up benchmarks to be compatible with read_slang. So if that ends up being fixed, I'll probably be able to remove the patch step for slang. I think we can leave this in for now just so the CI is able to pass the SystemVerilog reg test.

vtr_flow/tasks/regression_tests/vtr_reg_system_verilog/task_list.txt
loglav03: No, thank you for pointing that out. I uncommented it and generated golden results for it.

libs/EXTERNAL/patch_slang.cmake
loglav03: Header is now added.

AlexandreSinger (changes requested):
Thank you for resolving my prior comments! I appreciate it. This PR is looking very close to being able to be merged. I just have two final comments:

First, regarding the "-j" issue. When you remove it, it was asking you to prefix your call to the make subprocess with "+". So that would look like `BUILD_COMMAND +${CMAKE_COMMAND} ...`. This will enable the job server of our CMake to communicate with the sub-cmake that you are calling. I wonder if this is the cause of the multiple builds we are seeing for the Yosys build? May be worth looking into. But for this PR, can you please remove the `-j...` option and prepend the build command with `+`. I want to avoid confusing warnings from showing up to users of VTR.

Second, thank you for adding back the koios SV regression test. It is actually very common to forget this. I actually made a CI test to check for this type of thing automatically. I only have it run on currently run test suites. Can you please add `vtr_reg_system_verilog` to this config file:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/a8e00024ef85f05797bdf7dc2c5b756a50025363/dev/vtr_test_suite_verifier/test_suites_info.json#L14-L17

I do not think you need to mark any tests as ignored, so you can just copy the entry for `vtr_reg_valgrind_small`

loglav03:
@AlexandreSinger All comments should be resolved now! Simply adding a + to the build_command for slang didn't work so I had to find a workaround for it. It should be inheriting threads properly now without emitting a warning.

AlexandreSinger (approved):
LGTM

PR #3183, opened 2025-07-03
[Build] Fixed Issue in Rhel9 Build

AlexandreSinger:
The rhel9 build of VTR is unable to resolve the "onetbb" prefix in the include path for tbb. This is easily resolved by just removing this prefix.

Other tbb includes do not use this prefix. This is only found in this one place.

### DIFF
--- a/vpr/src/route/route_utils.cpp
+++ b/vpr/src/route/route_utils.cpp
@@ -20,8 +20,8 @@
 #include "timing_util.h"
 
 #ifdef VPR_USE_TBB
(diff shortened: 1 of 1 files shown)


amin1377 (approved):

AmirhosseinPoolad:
Oh hey my quest to use Fedora might actually prove useful to someone! Kind of surprising that it didn't work on RHEL9 but also I've only used Fedora 40 and 41 and RHEL 9 is AFAIK based on Fedora 34.

LGTM.

PR #3184, opened 2025-07-03
[CI] Improved Compatibility Build Speed

AlexandreSinger:
The compatibility builds do not run any tests. As such, they do not need to use compiler optimizations or IPO. By turning these features off, we should be able to speed up the compilation process on these builds.

### DIFF
--- a/vpr/src/route/route_utils.cpp
+++ b/vpr/src/route/route_utils.cpp
@@ -20,8 +20,8 @@
 #include "timing_util.h"
 
 #ifdef VPR_USE_TBB
-#include <oneapi/tbb/combinable.h>
-#include <oneapi/tbb/parallel_for_each.h>
+#include <tbb/combinable.h>
+#include <tbb/parallel_for_each.h>
 #endif // VPR_USE_TBB
 
 #ifndef NO_GRAPHICS


AmirhosseinPoolad:
I only checked 'B: GCC 11 (Ubuntu Noble - 24.04' but it appears to be around 4 minutes faster. Good change! Since we're not using the debug symbols in any way, can we can get a tiny bit more speed here by not generating debug symbols at all? Something like this:

```
cmake -DCMAKE_BUILD_TYPE=Debug \
      -DCMAKE_C_FLAGS_DEBUG="-O0" \
      -DCMAKE_CXX_FLAGS_DEBUG="-O0" \
```
By default cmake also adds the -g flag for the Debug build type.

You should also add a comment which says that these builds are compiled with no optimizations and are used only for checking compatibility. Otherwise I feel like someone in the future might accidentally copy paste the config here for another set of tests that should be compiled with release configuration.

AlexandreSinger:
Hi @AmirhosseinPoolad , thanks for the feedback! I agree, this should be commented for the exact reason you give.

I would actually like to keep the symbols on in this case. This should make debugging a little easier and symbols do not add too much run time. But it is something we can investigate later. For now, the lowest hanging fruit is making this debug.

AlexandreSinger:
Also, FYI. The build times have been very unstable due to how we have been using the CCache action. It turns out we can make these much much more stable. I am working on a much larger PR which should cut the CI time down by at least 30 mins per run.

AmirhosseinPoolad:
> I would actually like to keep the symbols on in this case. This should make debugging a little easier and symbols do not add too much run time. 

I do agree that this is probably not very impactful runtime-wise but how would it make debugging easier or harder? Aren't these builds basically thrown away after they are compiled and we only care if the compilation succeeded or not?

AlexandreSinger:
> > I would actually like to keep the symbols on in this case. This should make debugging a little easier and symbols do not add too much run time.
> 
> I do agree that this is probably not very impactful runtime-wise but how would it make debugging easier or harder? Aren't these builds basically thrown away after they are compiled and we only care if the compilation succeeded or not?

Thats a good point. I was not thinking. But I have a feeling that someone may want to add "basic_tests" to these builds in the future. I feel like future proofing this since the cost of having debug symbols is so small.

PR #3185, opened 2025-07-03
Remove system-wide installation of Sphinx

AmirhosseinPoolad:
Installing Sphinx using the system package manager goes against VTR's common practice. Both VTR and VTR documentation advises users to use a python virtualenv to install python dependencies and have a requirements.txt file to install any python dependencies. The current system-wide dependency installer scripts install Sphinx using the system package managers (APT and DNF) which might be undesirable for users that wish to keep their global python packages clean.

Other than that, it is not the only dependency for building the documentation and users still need to use pip and $VTR_ROOT/docs/requirements.txt to install the other dependencies for building the docs, which include the correct version of Shpinx. This PR removes the redundant and dangerous Sphinx installation from VTR's system-wide package installer scripts.


AmirhosseinPoolad:
Here's hoping this doesn't break the CI.

AlexandreSinger (approved):
LGTM

Issue #3039, opened 2025-05-14
Clarify pack, place, route, analysis on vs. off reporting


vaughnb-cerebras:
#### Proposed Behaviour
Currently the reporting of what stages are run in vpr is confusing. For example, if you specify --route, you will show all stages as being enabled, when really packing and placement are loading from file.

We should instead show one of three values for pack, place, route and analysis: ENABLED, DISABLED, or LOAD

#### Current Behaviour
Here's an example of a couple of current outputs:

----------------
The first file (xx.route.log) only did routing:

Packing was skipped (read from file):
# Load packing
Begin loading packed FPGA netlist file.
Netlist generated from file '[wb_conmax_stratixiv_arch_timing.net](http://wb_conmax_stratixiv_arch_timing.net/)'.
Detected 0 constant generators (to see names run with higher pack verbosity)
Finished loading packed FPGA netlist file

Placement was skipped (read from file):
# Load Placement
Reading ref_wb_conmax_stratixiv_arch_timing.place.

Successfully read ref_wb_conmax_stratixiv_arch_timing.place.

The command line specifies --route (interpreted as only perform routing) so this makes sense.  The following output is confusing though, so we should probably fix it Amin:

Packer: ENABLED 
Placer: ENABLED
Router: ENABLED
Analysis: ENABLED

===========================

The second ile (xx.place.log) only performed placement. This makes sense, since --place was specified, and that's what it is implies.

# Load packing
Begin loading packed FPGA netlist file.
Netlist generated from file '[wb_conmax_stratixiv_arch_timing.net](http://wb_conmax_stratixiv_arch_timing.net/)'.
Detected 0 constant generators (to see names run with higher pack verbosity)
Finished loading packed FPGA netlist file

<a bunch of output from placement, and then routing isn't run>

The output below is again misleading though:

Packer: ENABLED
Placer: ENABLED
Router: DISABLED
Analysis: DISABLED

The packer wasn't really enabled (it was loaded).  It looks like our logic just checks what the last stage to be run is and marks it and everything before it as "enabled".  We should probably also have a "loaded from file" state to explain what is going on more.

#### Possible Solution
Print ENABLED, DISABLED or LOAD in the command line flow summary.

Adding @amin1377 to make sure he's not simultaneously doing this.

PR #3188, opened 2025-07-04
Stage reporting update

SamuelHo10:
See issue #3039.

The updated values for each stage are now:

#### --pack

Packer: ENABLED
Placer: DISABLED
Analytical Placer: DISABLED
Router: DISABLED
Analysis: DISABLED

#### --place

Packer: LOAD
Placer: ENABLED
Analytical Placer: DISABLED
Router: DISABLED
Analysis: DISABLED

#### --route

Packer: LOAD
Placer: LOAD
Analytical Placer: DISABLED
Router: ENABLED
Analysis: AUTO

#### --analysis

Packer: LOAD
Placer: LOAD
Analytical Placer: DISABLED
Router: LOAD
Analysis: ENABLED


vpr/src/base/ShowSetup.h
```diff
@@ -33,4 +34,12 @@ struct ClusteredNetlistStats {
 };
 
 void ShowSetup(const t_vpr_setup& vpr_setup);
+
+/**
+ * @brief Converts the enum e_stage_action to a string.
+ * @param action The stage action to convert.
+ * @return A string representation of the stage action.
+ */
+std::string stage_action_to_string(e_stage_action action);
```
soheilshahrouz: move the declaration to vpr_types.h next to e_stage_action's declaration. This way, if someone adds a new value to e_stage_action, he/she can easily see that this function should be updated as well.

soheilshahrouz: If you have the energy for it, can you make e_stage_action an `enum class` instead of `enum`?

soheilshahrouz: With enum class, you can also think about using `constexpr vtr::array<e_stage_action, const char*, (size_t)e_stage_action::NUM_STAGE_ACTIONS>` to store names. The code for this would be more concise compared to a function. For example, check `rr_node_typename` in `rr_node_types.h`

soheilshahrouz (changes requested):

vpr/src/base/vpr_types.h
```diff
@@ -692,13 +692,18 @@ struct t_netlist_opts {
 };
 
 ///@brief Should a stage in the CAD flow be skipped, loaded from a file, or performed
-enum e_stage_action {
+enum class e_stage_action {
```
vaughnbetz, line 704: Please explain the states more.
Skip // Do not perform at all (end flow early)
Load // Load a previous result from file
Do // Run the specified algorithm
Auto // Run the specified algorithm if possible. Currently used to avoid analysis if we don't succeed at routing

vaughnbetz: Auto -> Skip_if_prior_fail

PR #3190, opened 2025-07-07
[CI] Moved Sanitized Test from Debug to RelWithDebInfo

AlexandreSinger:
I expect the sanitized test to be the tall pole for the speed of the CI due to its long run time on the testcases. After investigating with Vaughn, we realized that this test was using the debug build of VTR. This was likely making it much slower than it needed to be.

To improve the speed of this test, moved it to use the release build. We still want symbols to give good error messages.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -483,7 +483,7 @@ jobs:
     - name: Test
       env:
         CMAKE_PARAMS: ${{ matrix.params }}
-        BUILD_TYPE: debug
+        BUILD_TYPE: RelWithDebInfo
         LSAN_OPTIONS: 'exitcode=42' #Use a non-standard exit code to ensure LSAN errors are detected
         # In Ubuntu 20240310.1.0, the entropy of ASLR has increased (28 -> 32). LLVM 14 in this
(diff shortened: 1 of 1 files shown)


AmirhosseinPoolad:
Nice gains! I'm seeing an absurd amount of compiler warnings in the CI run for this PR, do you know what's the reason for those?
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/16127998530/job/45509408278?pr=3190

LGTM other than that.

AlexandreSinger:
> Nice gains! I'm seeing an absurd amount of compiler warnings in the CI run for this PR, do you know what's the reason for those? https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/16127998530/job/45509408278?pr=3190
> 
> LGTM other than that.

Good eye! I did not notice that. I gave it a quick look and it seems to be coming from a "maybe uninitialized" warning. I am not concerned about it; however its good to be aware of.

I believe this is similar to the IPO warnings that we suppressed which get revealed when more optimizations are performed. Since we do not care about being warning clean for this build, I do not think any action is necessary. Thanks for pointing it out!

Issue #3169, opened 2025-06-27
Yosys-Slang Parsing Errors


loglav03:
@petergrossmann21 I now have yosys-slang set up to be used as a frontend for yosys, and I'm experiencing some errors that I think you would expect from your experience with slang.

I ran **./run_reg_test.py vtr_reg_system_verilog -show_failures** and I got the following error below:

![Image](https://github.com/user-attachments/assets/23e2d56a-1c8b-44da-a969-8ea9b280521f)

This is the section of verilog from **vtr_primitives.v** (in module duel_port_ram) that I think is tripping up slang:

![Image](https://github.com/user-attachments/assets/7afb8487-f5f8-4500-bf69-d6092cd4cf38)

I believe slang doesn't like it whenever **'always@(posedge clk)'** is used or similar rules.

I also noticed the benchmarks you provided that had error/warnings for parsing with slang also had these 'always@(posedge clk)' lines. Is this something that is fixable?

The error is coming from **line 3604** in this file from yosys-slang: https://github.com/povik/yosys-slang/blob/master/src/slang_frontend.cc

loglav03:
I ended up commenting out the UndrivenPass in the file where it was causing it to break, and changed the .tcl file like Amir demonstrated, and that seemed to fix the UndrivenPass failures. I'm going to see if I can comment out that pass on the fly in cmake before it builds since I brought yosys-slang in as a submodule. Thanks @amirarjmand93  and @alirezazd  for help on this fix.

I also got it to choose between read_verilog or read_slang based on file extensions when yosys-slang is enabled as a parser.

I'm now testing whether yosys-slang works with multiple input files and it seems to be working in that regard. However, I'm getting some errors within some multi-input .sv files. I'm currently not sure if this is yosys-slang causing these errors or the syntax of the .sv files themselves, or maybe both.

I will probably need some 1-on-1 help with debugging this section @amirarjmand93 @alirezazd @petergrossmann21

amirarjmand93:
Glad to hear that @loglav03 
[...]

PR #3195, opened 2025-07-09
Resolve slang errors in VTR benchmarks

petergrossmann21:
This PR attempts to resolve errors and warnings encountered when running slang on the VTR benchmarks.

In broad strokes, there are four types of errors that were fixed:

1.  Eliminating defparam statements in favor of a more contemporary parameter override syntax
2. Making output ports type `reg` where needed because they are used in always blocks.
3. Reordering wire declarations and assign statements so that the wire declaration for a signal always precedes its use in an assign statement.
4.  Refactoring combinational always blocks (particularly those with large case statements) to eliminate scenarios where a branch in the always block can prevent a signal from being updated -- these missing updates are what cause synthesis tools to infer latches instead of inferring combinational logic as designers typically intend.

Every attempt was made to make these edits minimally invasive, up to and including leaving some (not all) odd whitespace intact along with any number of other syntax/style choices that would not normally be acceptable in professional-grade HDL code.

In addition to these changes, a redundant compiler directive was converted to a parameter in mkDelayWorker32B.v to attempt to preserve design intent.  Depending upon exactly how yosys was processing the redundant directive, this may or may not change synthesis results for this benchmark.

#### Related Issue

This work supports https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3169

#### Motivation and Context

The goal of this PR is to make the circuits slang-clean so they can be reliably used with that plugin in yosys.

Initial testing has been done only running slang locally on each Verilog file and by obtaining the following result

```
/tools/vpr/vtr-verilog-to-routing-fix_benchmark_slang/vtr-verilog-to-routing/vtr_flow/tasks/regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain$ ../../../../../vtr_flow/scripts/run_vtr_task.py . -check_golden

regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain...[Pass]
```

#### Types of changes

#### Checklist:


petergrossmann21:
One small additional update.  After installing the slang plugin from source on top of Yosys 0.54 and hiding the specify blocks in vtr_flow/primitives.v behind `ifdef wrappers, I was able to do the following simple experiment:

* execute Yosys from VPR repo root directory
* run the following commands at the Yosys tcl prompt:

```
plugin -i slang
read_slang vtr_flow/primitives.v vtr_flow/benchmarks/verilog/boundtop.v 
```

This command sequence completed successfully.  So I think we need to focus on what issues in the test setup are interfering with a successful parse.

petergrossmann21:
After more digging, I found an open issue in yosys-slang that describes similar results to what I am seeing:

https://github.com/povik/yosys-slang/issues/72

Note that the proposed workaround involves always reading the library verilog and the user verilog with the same command.  I'm not clear on whether that solution is workable for the VTR tests or not, @loglav03:  thoughts?

povik:
@petergrossmann21 I confirm read_slang won't see the parameters on modules loaded into Yosys with read_verilog. If you require to mix read_verilog and read_slang, the following workaround is available:

First you load the modules as required with read_verilog.

Second you load the user design with read_slang, at the same time you describe the pre-existing modules as blackboxes. E.g. for your dual_port_ram primitive you would add the following to the read_slang file list:

```
`celldefine
module dual_port_ram #(
    parameter ADDR_WIDTH = 1,
    parameter DATA_WIDTH = 1
) (
    input clk,

    input [ADDR_WIDTH-1:0] addr1,
    input [ADDR_WIDTH-1:0] addr2,
    input [DATA_WIDTH-1:0] data1,
    input [DATA_WIDTH-1:0] data2,
    input we1,
    input we2,
    output reg [DATA_WIDTH-1:0] out1,
    output reg [DATA_WIDTH-1:0] out2
);
endmodule
`endcelldefine
```
This way slang understand the parameters and their influence over port widths which is critical for correct elaboration.

When involving blackboxes make sure to specify the top with `--top` (related to an issue povik/yosys-slang#115). Let me know if you have trouble with this method.

povik:
One point which surprises me about your scripts is that you apply `read_verilog` to your primitives instead of `read_verilog -lib`. This means the modules are read including their behavioral content as if they were a user design. (The workaround from above would work with `read_verilog -lib` too.)

petergrossmann21:
@loglav03 @vaughnbetz What are your thoughts about reducing the proposed workaround to practice?  Are there any issues precluding doing as suggested in the synthesis scripts?

vaughnbetz:
The workaround seems fine to me. I'd just check the resource count and cpd results before and after the changes to ensure we haven't accidentally broken something. If the scripts are using an odd command or an order that doesn't always work we should change them.

povik:
> This resulted in this: ERROR: Attempted to add new module named '\dual_port_ram', but a module by that name already exists

@loglav03 The key part is that for slang the module is wrapped in
```
`celldefine
...
`endcelldefine
```
Alternatively it should have the `(* blackbox *)` attribute. With this change you shouldn't be getting the error above.

povik:
The "module by that name already exists" error might have also been caused by leaving out `--top <module-name>` from the command line. This is pending a fix in yosys-slang.

loglav03:
> The "module by that name already exists" error might have also been caused by leaving out `--top <module-name>` from the command line. This is pending a fix in yosys-slang.

@povik @petergrossmann21 - I was able to get this working without throwing the "module by that name already exists" error and shows
```
2. Executing SLANG frontend.
Top level design units:
    dual_port_ram
    single_port_ram
```
However, I'm running into the same issue I was having above. These modules get read by read_slang and throw 
```
Source line vtr_blackboxes.v:60:5:     specify
ERROR: Feature unimplemented at /home/llavign1/Gits/vtr-clone/libs/EXTERNAL/yosys-slang/src/slang_frontend.cc:2854, see AST and code line dump above
```
I'm not experienced with verilog, so I'm not sure if there's a possible workaround with the specify blocks that won't break anything.

povik:
Hi @loglav03 

If you're getting this output:

```
2. Executing SLANG frontend.
Top level design units:
    dual_port_ram
    single_port_ram
```

I think you haven't specified the top with `--top`.

loglav03:
This is what I have set up in the tcl script:
```
set readfile [file join [pwd] "filelist.txt"]
#Writing names of circuit files to file list
build_filelist {XXX} $readfile
puts "Using Yosys read_slang command"
read_slang --top single_port_ram --top dual_port_ram -C $readfile
```
The file that defines the single and dual port ram modules as blackboxes is included in the $readfile

loglav03:
@povik Thank you for taking the time to help with this. @petergrossmann21 I managed to get the synthesis script set up in a way that passes verilog benchmarks that were previously failing with yosys-slang. 

When slang is enabled, it does this in the synthesis script where -DNO_SPECIFY is a macro that prevents slang from reading the specify blocks in vtr_primitives. So, with this macro, I was able to read the user design verilog along with vtr_primitives with the same read_slang command. This resulted in the benchmarks (listed below) passing.

For the test I used, I made my own copy of the vtr_reg_nightly_test3/vtr_reg_qor with your fixed benchmarks substituted in for the old ones to test these benchmarks. It doesn't cover all of the ones you modified, but I would say the other ones would probably pass as well if we were to test them. I haven't checked the golden results with these yet so that's something I still have to do.

Modified portion of synthesis.tcl:
```
set readfile [file join [pwd] "filelist.txt"]
#Writing names of circuit files to file list
build_filelist {XXX} $readfile
puts "Using Yosys read_slang command"
read_slang -DNO_SPECIFY -C $readfile
```
Example of the guard on specify blocks:
```
`ifndef NO_SPECIFY
    specify
        (in *> out) = "";
    endspecify
`endif // NO_SPECIFY
```
Benchmarks that passed:
```
mcml.v
LU32PEEng
LU8PEEng
stereovision3
or1200.v
raygentop.v
mkSMAdapter4B.v
mkPktMerge.v
mkDelayWorker32B.v
ch_intrinsics.v
boundtop.v
bgm.v 
arm_core.v
```

povik:
Can you provide details on the nature of the failure? Is this a QoR check or a functional check which is failing?

loglav03:
You are correct these results are much smaller than expected. There's even some that are showing `relative value 0.0 outside of range [0.9,1.1]`. I assumed we're getting these results because of the specify blocks I have set to not be read by slang in vtr_primitives to avoid throwing the "specify ERROR: Feature unimplemented" errors. Below are the specify blocks in single and dual port ram modules that I have guarded against slang.

Single port ram:
```
localparam MEM_DEPTH = 2 ** ADDR_WIDTH;

reg [DATA_WIDTH-1:0] Mem[MEM_DEPTH-1:0];

`ifndef NO_SPECIFY
    specify
        (clk*>out)="";
        $setup(addr, posedge clk, "");
        $setup(data, posedge clk, "");
        $setup(we, posedge clk, "");
        $hold(posedge clk, addr, "");
        $hold(posedge clk, data, "");
        $hold(posedge clk, we, "");
    endspecify
`endif // NO_SPECIFY
   
    always@(posedge clk) begin
        if(we) begin
            Mem[addr] = data;
        end
    	out = Mem[addr]; //New data read-during write behaviour (blocking assignments)
    end
```
Dual port ram:
```
localparam MEM_DEPTH = 2 ** ADDR_WIDTH;

reg [DATA_WIDTH-1:0] Mem[MEM_DEPTH-1:0];

`ifndef NO_SPECIFY
    specify
        (clk*>out1)="";
        (clk*>out2)="";
        $setup(addr1, posedge clk, "");
        $setup(addr2, posedge clk, "");
        $setup(data1, posedge clk, "");
        $setup(data2, posedge clk, "");
        $setup(we1, posedge clk, "");
        $setup(we2, posedge clk, "");
        $hold(posedge clk, addr1, "");
        $hold(posedge clk, addr2, "");
        $hold(posedge clk, data1, "");
        $hold(posedge clk, data2, "");
        $hold(posedge clk, we1, "");
        $hold(posedge clk, we2, "");
    endspecify
`endif // NO_SPECIFY
   
    always@(posedge clk) begin //Port 1
        if(we1) begin
            Mem[addr1] = data1;
        end
        out1 = Mem[addr1]; //New data read-during write behaviour (blocking assignments)
    end

    always@(posedge clk) begin //Port 2
        if(we2) begin
            Mem[addr2] = data2;
        end
        out2 = Mem[addr2]; //New data read-during write behaviour (blocking assignments)
    end
```
It might also be because I don't have it reading the vtr_primitives correctly with the user design? I get this at the start of the output for yosys, and it looks like its reading it properly, but I'm not sure what the warning is for ch_intrinsics.v. Also vtr_primitives is being read by read_slang along with the user design in the same command, so I'm not sure if it would need an equivalent -lib. Maybe that could be a reason?

```
1. Executing SLANG frontend.
ch_intrinsics.v:49:9: warning: non-edge sensitivity on a signal will be synthesized as @* sensitivity
always @( tag or memory_controller_address or memory_controller_write_enable or memory_controller_in)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ch_intrinsics.v:278:10: warning: non-edge sensitivity on a signal will be synthesized as @* sensitivity
always @(cur_state)
         ^~~~~~~~~
Top level design units:
    DFF
    LUT_K
    adder
    dual_port_ram
    fpga_interconnect
    memset
    multiply
    mux
```

loglav03:
So I ran this `read_slang -DNO_SPECIFY --top memset -C $readfile` where the full vtr_primitives.v file is included in the readfile.

This passed through the entire yosys stage, but when it reached the VPR stage (packing, placing, routing) it threw this error:
```
# Building complex block graph
Warning 1: io[0].clock[0] unconnected pin in architecture.
# Building complex block graph took 0.01 seconds (max_rss 24.3 MiB, delta_rss +5.5 MiB)
Circuit file: ch_intrinsics.pre-vpr.blif
# Load circuit
# Load circuit took 0.00 seconds (max_rss 26.4 MiB, delta_rss +2.1 MiB)
Error 1: 
Type: Blif file
File: ch_intrinsics.pre-vpr.blif
Line: 171
Message: Only rising edge latches supported
```
Not sure what this means or how it can be fixed. @vaughnbetz @AlexandreSinger @povik - Any thoughts?

vaughnbetz:
From the message, it looks like it found a falling edge register, which we don't currently support. I looked at ch_intrinsics.v and I don't see any negative edge triggered registers, so this looks like an improper synthesis if there is a negative edge register. Can you attach the ch_intrinsics.pre-vpr.blif file so I can look at line 171?

loglav03:
It's not only ch_intrinsics that is getting that error. There are a few others that are getting the exact same error so I agree that it is improper synthesis causing these. There's also a few other benchmarks that are failing in the abc stage and a couple failing in the yosys stage (I'm assuming these are also caused by improper synthesis).

ch_intrinsics (ch_intrinsics.pre-vpr.blif): 
Screenshot from 2025-07-21 12-03-10

raygentop.v failure in abc (abc0.out):
Screenshot from 2025-07-21 13-20-45

LU8PEEng.v failure in yosys (parmys.out):
Screenshot from 2025-07-21 13-38-27

petergrossmann21:
> The ch_intrinsics.pre-vpr.blif file has an active low (al) latch, which we don't support (and which is unusual -- most designs use registers, not latches, and commercial FPGA CAD tools warn about latches and do a conservative timing analysis).
> 
> The other two are crashes (seg fault for one) so something looks quite broken.

@loglav03 The always block starting in this line 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3ad299e1429e2fb3ba738bf5507673bcfc7b792e/vtr_flow/benchmarks/verilog/ch_intrinsics.v#L50

is the culprit here.

The block is using an older coding style in which it takes advantage of the partial sensitivity list to prevent `tag` and `prevTag` from being treated like latch clock signals so that the block can infer combinational logic.  slang's choice to convert the sensitivity list to `always @(*)` is causing this intent to be ignored.  When the block is modeled with `always @(*)` the overall block syntax becomes sequential, with `tag` and `prevTag` being treated like clocks.

There are ways to recode the HDL to avoid this issue; many of the solutions that will work would also represent an improvement in overall HDL coding style for this block.

loglav03:
That makes a lot of sense, thanks @petergrossmann21. I was hoping it was something to do with the code and not the way I have slang set up because I'm running out of ways to have it set up. Would this mean it is likely the other failures I'm getting for other benchmarks would be caused by outdated coding styles?

petergrossmann21:
> That makes a lot of sense, thanks @petergrossmann21. I was hoping it was something to do with the code and not the way I have slang set up because I'm running out of ways to have it set up. Would this mean it is likely the other failures I'm getting for other benchmarks would be caused by outdated coding styles?

@loglav03 We'll have to handle them on a case by case basis, but so far when I have seen issues parsing these benchmark there have been ways to refactor outdated coding styles to resolve them.

vaughnbetz (approved):
LGTM! Thanks.

PR #3196, opened 2025-07-10
Fix slang errors and warnings in Koios benchmarks

petergrossmann21:
Resolve slang errors/warnings in the Koios benchmarks.

#### Description

The errors/warnings fell into the following buckets:

1.  Add reg keyword to output ports where outputs are used in LHS of always blocks.
2. Move timescale directives to top of files.
3. Add empty output port connection declarations to eliminate warnings (no functional change)
4. Change instance name to prevent namespace conflict with SystemVerilog keyword
5. Comment out an initial block that used a variable that had also been commented out.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3169

#### Motivation and Context

Eliminating slang errors allows these benchmarks to continue to be used in flows that make use of the yosys-slang plugin.

Primary method of testing has been running slang directly at the command line.  Quick research suggested there was not a single VTR regression test I could execute to try all four of these benchmarks in one push button; due to this and test runtime concerns I have deferred running VTR regression tests to validate that no synthesis results have changed.

From a synthesis standpoint, the risk of results changing due to these edits is very low.

#### Types of changes

#### Checklist:

### DIFF
--- a/vtr_flow/benchmarks/verilog/koios/attention_layer.v
+++ b/vtr_flow/benchmarks/verilog/koios/attention_layer.v
@@ -164,7 +164,7 @@ wordwise_bram_2 out_buffer34
 );
 
 //Softmax layer has a parallelism of 4
-softmax soft(
+softmax softmax(
 	.inp(data_to_softmax),
 	.sub0_inp(data_to_softmax),
 	.sub1_inp(data_to_softmax),
--- a/vtr_flow/benchmarks/verilog/koios/conv_layer_hls.v
+++ b/vtr_flow/benchmarks/verilog/koios/conv_layer_hls.v
@@ -18,6 +18,7 @@
 // Abridged for VTR by: Daniel Rauch
 //////////////////////////////////////////////////////////////////////////////
 
+`timescale 1 ns / 1 ps
 
 module dpram (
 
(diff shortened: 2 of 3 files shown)


povik:
@gadfort Now that you have pinged me I've taken a look. The two below are worth changing but I'd say it's of low priority.

> * Add reg keyword to output ports where outputs are used in LHS of always blocks.

This could be made into a demotable error.
 
> * Move timescale directives to top of files.

Ideally synthesis would ignore timescales. I suppose there was an error "time scale declaration must come before all other items in scope".

petergrossmann21:
> @gadfort Now that you have pinged me I've taken a look. The two below are worth changing but I'd say it's of low priority.
> 
> > * Add reg keyword to output ports where outputs are used in LHS of always blocks.
> 
> This could be made into a demotable error.
> 
> > * Move timescale directives to top of files.
> 
> Ideally synthesis would ignore timescales. I suppose there was an error "time scale declaration must come before all other items in scope".

@povik regarding timescales you are correct that this is the error we see.

See also comments on #3195; what are your thoughts regarding handling parameters for modules read using previous `read_verilog` commands?

PR #3198, opened 2025-07-11
[CI] Updated Install Dependencies for Format Tests

AlexandreSinger:
Explicitly installed the apt packages required for the tests and the
python requirements.

In the future, we should update the version of pylint and black used by
the CI to allow us to use a more up to date version of Python for
linting. It looks like linting has to be done on CI which is not ideal.


vaughnbetz (approved):
LGTM

PR #3199, opened 2025-07-11
[AP][MassLegalizer] Resolved Issues in Partitioner

AlexandreSinger:
Found some low-hanging fruit in the partial legalizer regarding how we merge the windows and how we partition the blocks.

I found that the windows were not being fully merged (i.e. there still existed overlap between windows). I updated the algorithm to iteratively retry to merge if the bounding box of the window changes. To maintain good performance, I added a spatial lookup to narrow down the region where overlap can occur.

The partitioning of blocks had a major bug in it where it was not actually moving blocks over most of the time. This was causing the partial legalizer to struggle to resolve of a legal solution. To resolve this, I completely rewrote the partitioning algorithm to make it more stable. The new algorithm starts by putting the blocks on the side they want (if their solved position is on the left of the partition, the will go into the left partition). It then moves the blocks closest to the partition line until the overfill of both partitions is as balanced as can be.

These two optimizations together seem to have improved WL by around 2% and CPD by around 1%.


vpr/src/analytical_place/partial_legalizer.cpp
```diff
@@ -1196,52 +1197,121 @@ std::vector<SpreadingWindow> BiPartitioningPartialLegalizer::get_min_windows_aro
...
-                finished_window[j] = true;
+
+        // After merging with another window, the bounds of the window is updated
+        // and another window may now become merged. We need to recheck again after
+        // this occurs.
+        while (true) {
+            // Get the lower and upper bounds in the spatial lookup. This will
+            // give a list of all windows which could possibly overlap right now
+            // (at least in the x dimension).
+            // NOTE: There may be diplicates; but thats ok since these checks
```
vaughnbetz: diplicates -> duplicates

vpr/src/analytical_place/partial_legalizer.cpp
```diff
@@ -1587,72 +1669,158 @@ void BiPartitioningPartialLegalizer::partition_blocks_in_window(
...
+
+    // Try to move things from upper to lower greedily.
+    std::vector<APBlockId> upper_blocks_to_move;
+    upper_blocks_to_move.reserve(upper_contained_blocks.size());
+    for (size_t i = 0; i < upper_contained_blocks.size(); i++) {
+        // If the upper window has no overfill, there is no reason to move anything.
+        if (upper_window_overfill.is_zero())
+            break;
+
+        // Check what would be the gain of moving this block to the upper partition.
```
vaughnbetz: upper partition -> lower partition

vpr/src/analytical_place/partial_legalizer.cpp
```diff
@@ -1587,72 +1669,158 @@ void BiPartitioningPartialLegalizer::partition_blocks_in_window(
...
+        upper_blocks_to_move.push_back(upper_blk);
+        upper_contained_blocks[i] = APBlockId::INVALID();
+    }
+
+    // Move the blocks into the windows.
+    //  Here we maintain the order the best we can to make future sorts faster.
+    // Compress the lower contained blocks which have not moved.
+    lower_window.contained_blocks.reserve(lower_contained_blocks.size() - lower_blocks_to_move.size() + upper_blocks_to_move.size());
+    for (APBlockId lower_blk : lower_contained_blocks) {
+        if (!lower_blk.is_valid())
```
vaughnbetz: Would be a bit shorter and more clear to reverse this to 
if (lower_blk.is_valid()) {
   lower_window.contained ...
}

vpr/src/analytical_place/partial_legalizer.cpp
```diff
@@ -1587,72 +1669,158 @@ void BiPartitioningPartialLegalizer::partition_blocks_in_window(
...
+    // they were inserted in reverse order, so we reverse them before inserting.
+    // TODO: Investigate if this actually has an improvements.
+    upper_window.contained_blocks.reserve(upper_contained_blocks.size() - upper_blocks_to_move.size() + lower_blocks_to_move.size());
+    std::reverse(lower_blocks_to_move.begin(), lower_blocks_to_move.end());
+    upper_window.contained_blocks.insert(upper_window.contained_blocks.end(),
+                                         lower_blocks_to_move.begin(),
+                                         lower_blocks_to_move.end());
+    // Compress the upper contained blocks which have not moved.
+    for (APBlockId upper_blk : upper_contained_blocks) {
+        if (!upper_blk.is_valid())
```
vaughnbetz: Same as comment above.

vpr/src/analytical_place/partial_legalizer.cpp
```diff
@@ -1587,72 +1669,158 @@ void BiPartitioningPartialLegalizer::partition_blocks_in_window(
...
+        lower_window.contained_blocks = std::move(lower_contained_blocks);
+        upper_window.contained_blocks = std::move(upper_contained_blocks);
+        return;
+    }
+
+    // 2) Resolve any overfill by moving blocks over the partition line.
+    // Here, we try to move blocks that are closest to the partition line such
+    // that the overfill on both sides is balanced.
+
+    // Try to move things from lower to upper greedily.
```
vaughnbetz, line 1807: The code for the lower and upper overfill and the block moving code for lower and upper are very similar, with just upper / lower transposed.  Could you refactor it to a helper routine or two that you pass references to the from list and to list t to and the list of blocks to move etc.?  And then just call the helper(s) twice with the appropriate references.

vpr/src/analytical_place/partial_legalizer.cpp
vaughnbetz: Why don't you inert region.xmax into the map, and make upper a function of region.xmax, instead of using region.xmax-1 on both insertion and checking for the bound?  Aren't the regions inclusive where they include their xmax too?  
If there's a reason you're not including the xmax boundary you should comment what it is.

vaughnbetz (changes requested):
A few possible changes and a commenting clarification (or maybe a change to xmax behaviour, depending on what the answer is).

vpr/src/analytical_place/partial_legalizer.cpp
AlexandreSinger: Ill improve the comments. It is correct to subtract the 1 here. You are correct that the regions are inclusive (they spill right over to the crust from xmin to xmax); however we do not consider two windows to overlap if the xmax of one window is the same as the xmin of another window (i.e. they are touching on one side). Hence, for this map, we can exclude searching for any windows that begin at the xmax coordinate since we do not care if any windows are found there.

Thinking about this more though, this only works if the windows are always at integer locations (which they are right now, but I would like to keep this general if I can). So really this should be xmax - epsilon, where epsilon is some reasonably small number.

AlexandreSinger: This was a really good point. Looking at my code, this was very unclear. I have used a more general way of handling this which is much better now. We now exclude the border from being included in this map.

vpr/src/analytical_place/partial_legalizer.cpp
```diff
@@ -1504,6 +1599,108 @@ PartitionedWindow BiPartitioningPartialLegalizer::partition_window(
...
+ * @brief Helper method for trying to move a block with the given mass from its
+ *        window to the other window if, by moving this block, we can balance
+ *        the overfill of the two windows better.
+ *
+ * Returns true if the block was moved.
+ *
+ * If this block is moved, the utilizations and the overfills will be updated
+ * accordingly, but the block itself will remain unmoved.
+ * NOTE: This may seem strange, it is; this is to prevent updating the primitive
+ *       vectors too much, but tends to take a long time.
```
vaughnbetz: but tends -> which tends

vaughnbetz (approved):
LGTM. Just one typo ...

PR #3201, opened 2025-07-13
Update Release Procedures With Git Tag Memo

AlexandreSinger:
I have been running into issues with how VTR handles its versioning code. I have tracked the issue down to the tagged v9.0.0 release of VTR being off of another branch than Master.

I added a small note to the release procedures to warn about this for the future.

### DIFF
--- a/README.developers.md
+++ b/README.developers.md
@@ -1644,23 +1644,23 @@ to output the html analysis to a specific folder, run `scan-build make -o /some/
 ## General Principles
 
 We periodically make 'official' VTR releases.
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
@vaughnbetz Its really not that pressing of an issue that v9.0.0 is a tag of another branch than Master; but I have been running into issues with how downstream tools (like SiliconCompiler) track the version of VTR. Some users have been experiencing the version being set to "v9.0.0-candidate" which causes issues.

Just added a small note to the release procedure about this since its not an obvious thing.

vaughnbetz (approved):
LGTM

PR #3202, opened 2025-07-14
[AP][GP] Tuning Parameters

AlexandreSinger:
Tuned some of the parameters for the global placement stage. The goal of this tuning was to speed up the global placer as much as reasonable, while retaining similar quality.

Reduced the B2B convergence gap slightly to allow the B2B solver to converge on a more accurate solution.

Reduced the number of B2B convergences required before stopping from 2 to 1. I found that this did not provide as much quality, and reducing this greatly improved the solver's speed.

Reduced the max number of CG iterations. This should further improve the speed of the solver.

Reduced the relative gap between the upper and lower bound solutions in the global placer from 5% to 1%.


vaughnbetz (approved):
LGTM

PR #3203, opened 2025-07-18
Deterministic parallel router

AlexPoupakis:
This PR provides everything necessary to make the parallel router deterministic without having to offset the lookahead costs.

- Added floating point tolerance to the post target pruning condition
  Note: the vtr::isclose() function cannot used because it checks the absolute difference (it is symmetric), whereas we can only tolerate error on the left hand side (assymetric). Using this function makes the parallel router non-deterministic!
- Created SIMPLE lookahead which can only load (not compute) a cost map and query it based on distance.

By using appropriate (admissible) cost maps, both of these changes allow the parallel router to be deterministic. Attached are admissible cost maps for the flagship VTR architecture (`k6_frac_N10_frac_chain_mem32K_40nm`) devices.
[vtr_devices_admissible_lookahead_maps.zip](https://github.com/user-attachments/files/21318973/vtr_devices_admissible_lookahead_maps.zip)


vpr/src/base/read_options.cpp
```diff
@@ -2938,6 +2942,7 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
             " to reduce the run-time to build the router lookahead and also its memory footprint\n"
             " * extended_map: A more advanced and extended lookahead which accounts for a more\n"
             "                 exhaustive node sampling method\n"
+            " * simple: A purely distance-based lookahead loaded from an external file\n"
```
AlexandreSinger, line 2977: Could you please also add this to the documentation here:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/3ad299e1429e2fb3ba738bf5507673bcfc7b792e/doc/src/vpr/command_line_usage.rst?plain=1#L1844-L1851

This code only changes the help message, not the documentation. Its something I hope to unify at some point.
I recognize that the compressed and extended maps are not there either. If you have the energy to add those as well to the documentation I would appreciate it.

In the documentation for the simple router lookhaed, can you add a bit on how to load this from an external file. I think it would be valuable to include the VPR option that loads the map from an external file (which i think is read_router_lookahead).

vpr/src/route/router_lookahead/router_lookahead_simple.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include "intra_cluster_serdes.h"
+#include "mmap_file.h"
+#include "serdes_utils.h"
+#endif /* VTR_ENABLE_CAPNPROTO */
+
+
+/******** File-Scope Variables ********/
+
+//Look-up table from CHANX/CHANY (to SINKs) for various distances
+t_simple_cost_map simple_cost_map;
```
AlexandreSinger: Can you please make this a private member variable of the SimpleLookahead class? File-scope variables can be dangerous and we are trying to move away from them in VTR.

I recognize that the other lookaheads do this. I am hoping to move them to using class variables in the future.

vpr/src/base/read_options.cpp
AlexandreSinger: I think it would also be valueable to explain what is different about this lookahead compared to the other lookaheads (are the others not distance based? for example.).

vpr/src/route/router_lookahead/router_lookahead_simple.h
```diff
@@ -0,0 +1,45 @@
...
+#include "vtr_ndmatrix.h"
+#include "router_lookahead.h"
+#include "router_lookahead_map_utils.h"
+
+class SimpleLookahead : public RouterLookahead {
+  protected:
+    float get_expected_cost(RRNodeId node, RRNodeId target_node, const t_conn_cost_params& params, float R_upstream) const override;
+    std::pair<float, float> get_expected_delay_and_cong(RRNodeId from_node, RRNodeId to_node, const t_conn_cost_params& params, float R_upstream) const override;
+    
+    void compute(const std::vector<t_segment_inf>& /*segment_inf*/) override {
```
AlexandreSinger, line 19: An error should be thrown in this case. If someone is computing the lookahead when they are not supposed to, the code should stop them.

vpr/src/route/router_lookahead/router_lookahead_simple.cpp
```diff
@@ -0,0 +1,189 @@
...
+// instead.
+//
+#ifndef VTR_ENABLE_CAPNPROTO
+
+#define DISABLE_ERROR                               \
+    "is disabled because VTR_ENABLE_CAPNPROTO=OFF." \
+    "Re-compile with CMake option VTR_ENABLE_CAPNPROTO=ON to enable."
+
+void read_router_lookahead(const std::string& /*file*/) {
+    VPR_THROW(VPR_ERROR_PLACE, "MapLookahead::read_router_lookahead " DISABLE_ERROR);
```
AlexandreSinger: Should this and the one below be "MapLookahead" or "SimpleLookhead". I think it should be SimpleLookahead.

vpr/src/route/router_lookahead/router_lookahead_simple.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include "vtr_ndmatrix.h"
+#include "router_lookahead.h"
+#include "router_lookahead_map_utils.h"
+
+class SimpleLookahead : public RouterLookahead {
```
AlexandreSinger, line 14: In VTR, we like to add doxygen comments before classes. Could you please add a doxygen comment here. Something like:

```cpp
/**
 * @brief <short description>
 *
 * <long description>
 */
```

vpr/src/route/router_lookahead/router_lookahead_simple.h
```diff
@@ -0,0 +1,45 @@
...
+    } 
+
+    void write(const std::string& file) const override;
+
+    void write_intra_cluster(const std::string& /*file*/) const override {
+      VPR_THROW(VPR_ERROR_ROUTE, "SimpleLookahead::write_intra_cluster unimplemented");
+    }
+
+    float get_opin_distance_min_delay(int /*physical_tile_idx*/, int /*from_layer*/, int /*to_layer*/, int /*dx*/, int /*dy*/) const override {
+      return -1;
```
AlexandreSinger: Should this throw an error instead? If not, is returning "-1" the best idea? Having a negative minimum delay seems strange to me.

AlexandreSinger:
There is also some minor formatting issues:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/16373929193/job/46342261017?pr=3203

You can run `make format` from the root directory or just fix them by hand.

vpr/src/route/router_lookahead/router_lookahead_simple.cpp
AlexPoupakis: Absolutely, this was a migration issue

vpr/src/route/router_lookahead/router_lookahead_simple.h
AlexPoupakis: For this I did the same thing that CLASSIC, NO_OP, EXTENDED_MAP, and COMPRESSED do. All of them return -1 and if I were to throw an error (which I tried at first by instinct), then the code breaks.

vpr/src/route/router_lookahead/router_lookahead_simple.cpp
AlexPoupakis: Making the cost map a private member breaks the 
`dump_readable_router_lookahead_map(file, simple_cost_map_size, get_wire_cost_entry);` 
call, since `get_wire_cost_entry` cannot access the cost map. How do you want to handle this?

AlexandreSinger: Ah, thats not something that we should fix in this PR. This is something that is a larger issue with these files.

For this PR, can you please just make the variable `static`. This will at least prevent link-time issues from creeping in.

AlexandreSinger: If making it static causes an issue (which I do not expect), feel free to leave as is.

AlexPoupakis: Done!

AlexandreSinger (approved):
LGTM

PR #3204, opened 2025-07-21
Add --place_freq command line option

soheilshahrouz:
Adds a new command line option to do placement for each channel width trial when searching for minimum channel width.

### DIFF
--- a/vtr_flow/scripts/python_libs/vtr/vpr/vpr.py
+++ b/vtr_flow/scripts/python_libs/vtr/vpr/vpr.py
@@ -114,8 +114,8 @@ def run_relax_w(
         vpr_args["place"] = True
 
     # Remove arguments related to minimum channel width binary search
-    vpr_args.pop("min_route_chan_width_hint")
-    vpr_args.pop("verify_binary_search")
+    vpr_args.pop("min_route_chan_width_hint", None)
+    vpr_args.pop("verify_binary_search", None)
 
     if write_rr_graph:  # Write out rr_graph with known W
         vpr_args["write_rr_graph"] = write_rr_graph


vpr/src/base/vpr_types.h
```diff
@@ -375,23 +375,23 @@ constexpr int NUM_PL_MOVE_TYPES = 7;
 constexpr int NUM_PL_NONTIMING_MOVE_TYPES = 3;
 
 /* Timing data structures end */
+
+// Annealing schedule
```
vaughnbetz: Could expand this comment. AUTO_SCHED: computes initial temperature, exit criterion and temperature update rate from statistics computing during the anneal.
USER_SCHED: the user has specified explicit numbers for the key annealing schedule parameters

vpr/src/base/vpr_types.h
```diff
@@ -375,23 +375,23 @@ constexpr int NUM_PL_MOVE_TYPES = 7;
...
     PLACEMENT,
     ROUTING
 };
-/* What's on screen? */
 
-enum pfreq {
-    PLACE_NEVER,
-    PLACE_ONCE,
-    PLACE_ALWAYS
+enum class e_place_freq {
```
vaughnbetz, line 397: Put a comment on this (doxygen)

vpr/src/place/place.cpp
```diff
@@ -52,6 +51,10 @@ void try_place(const Netlist<>& net_list,
     // Initialize the variables in the placement context.
     mutable_placement.init_placement_context(placer_opts, directs);
 
+    if (mutable_floorplanning.cluster_constraints.empty()) {
```
vaughnbetz, line 57: Comment why this only needs to be loaded once.

vaughnbetz:
@soheilshahrouz : nag, nag.
A few things to change, and then we can merge this. It would be good if we could merge this in our meeting Monday.

PR #3206, opened 2025-07-22
Cleaned up RRGraph Drawing Code

SamuelHo10:
Edges for each side are now drawn between inter-cluster and intra-cluster nodes. I’ve simplified the RRGraph drawing code by introducing reusable helper functions draw_rr_edge and draw_rr_node, while preserving all existing functionality and color schemes. I’ve also disabled the code that was drawing sources and sinks, as it didn’t seem to render correctly to begin with. Would it make sense to reimplement source/sink visualization, or is it something we want to leave out for now? In the next PR, I will probably hook up the new buttons to RRGraph drawing, so I left the toggle_RR variable unchanged for now.


vaughnbetz:
It's probably good to add SOURCE and SINK viewing, probably with another checkbox to control clutter. 

You'd probably draw sources and sinks in some visually distinct way (circle for one and ellipse for the other?) with edges to pins where they connect. Physically they are inside the block, so in the interior of cluster-level blocks makes sense. For primitive level grey would ideally be drawn in the primitive they belong to. But you'll have to figure out which primitive that is ... the net_rr_terminals data structure loader also figures that out so you could look at it.

vpr/src/draw/draw_rr.cpp
```diff
@@ -275,14 +236,13 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
-
-    int transparency_factor = get_rr_node_transparency(rr_node);
 
     e_rr_type from_type, to_type;
-    short switch_type;
 
-    from_type = rr_graph.node_type(rr_node);
+    from_type = rr_graph.node_type(inode);
+    if (from_type == e_rr_type::SOURCE || from_type == e_rr_type::SINK || !is_inter_cluster_node(rr_graph, inode)) {
+        return;
```
vaughnbetz, line 245: Add a short comment 
// Currently don't visualize sources or sinks

vpr/src/draw/draw_rr.cpp
```diff
@@ -291,228 +251,101 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
 
     for (t_edge_size iedge = 0, l = rr_graph.num_edges(inode); iedge < l; iedge++) {
-        RRNodeId to_node = rr_graph.edge_sink_node(rr_node, iedge);
+        RRNodeId to_node = rr_graph.edge_sink_node(inode, iedge);
         to_type = rr_graph.node_type(to_node);
         bool edge_configurable = rr_graph.edge_is_configurable(inode, iedge);
 
-        if (!is_edge_valid_to_draw(RRNodeId(to_node), rr_node))
-            continue; // skip drawing if edge is not valid to draw
+        if (to_type == e_rr_type::SOURCE || to_type == e_rr_type::SINK || !is_inter_cluster_node(rr_graph, to_node)) {
```
vaughnbetz: Add a short comment
// Currently don't visualize sources or sinks

vpr/src/draw/draw_rr.cpp
```diff
@@ -291,228 +251,101 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
-                            g->set_color(ezgl::MEDIUM_PURPLE, transparency_factor);
-                        }
-                        draw_pin_to_pin(inode, to_node, g);
-                        break;
-                    default:
-                        vpr_throw(VPR_ERROR_OTHER, __FILE__, __LINE__,
-                                  "in draw_rr_edges: node %d (type: %d) connects to node %d (type: %d).\n",
-                                  inode, from_type, to_node, to_type);
-                        break;
+                if (to_type == e_rr_type::CHANX || to_type == e_rr_type::CHANY) {
```
vaughnbetz: Should add type MUX to this (Amin just added it).  Can use the same colour as if it were a CHANX or CHANY.

vpr/src/draw/draw_rr.cpp
```diff
@@ -291,228 +251,101 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
-                        vpr_throw(VPR_ERROR_OTHER, __FILE__, __LINE__,
-                                  "in draw_rr_edges: node %d (type: %d) connects to node %d (type: %d).\n",
-                                  inode, from_type, to_node, to_type);
-                        break;
+            case e_rr_type::CHANX:
+            case e_rr_type::CHANY:
+                if (to_type == e_rr_type::IPIN) {
+                    color = blk_LIGHTSKYBLUE;
+                    if (draw_state->draw_rr_node[to_node].node_highlighted && draw_state->draw_rr_node[inode].color == DEFAULT_RR_NODE_COLOR) {
+                        // If the IPIN is clicked on, draw connection to all the CHANX
```
vaughnbetz: I think the overall highlighting code needs to be tested and this comment needs to be hoisted higher and updated.
I believe you can right click on an rr-node, and its fanin and fanout  will be highlighted in certain colours.  The highlighted node itself is drawn in magenta. You should test exactly what happens now, and see if you like it.  I think the code can be simplified (which you've done) but you need to test we're doing some reasonable highlighting (idea: right click highlights and rr-node and its fan-in edges and fan-out edges).

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -300,39 +301,48 @@ void draw_intra_cluster_edge(RRNodeId inode, RRNodeId prev_node, ezgl::renderer*
...
+        // determine which coord is first based on the pin edge direction
+        ezgl::point2d prev_coord, icoord;
+        if (pin_edge_dir == FROM_INTRA_CLUSTER_TO_INTER_CLUSTER) {
+            prev_coord = intra_cluster_coord;
+            icoord = {inter_cluster_x, inter_cluster_y};
+        } else if (pin_edge_dir == FROM_INTER_CLUSTER_TO_INTRA_CLUSTER) {
+            prev_coord = {inter_cluster_x, inter_cluster_y};
+            icoord = intra_cluster_coord;
+        } else {
+            VPR_ERROR(VPR_ERROR_DRAW, "Invalid pin edge direction: %d", pin_edge_dir);
```
vaughnbetz, line 338: Can I have edges that connect two INTRA_CLUSTER pins?  I believe we can at intermediate hierarchy levels in a clustered block / pb_type. You should update the code to handle that case if I'm right.

vpr/src/draw/draw_rr.cpp
vaughnbetz: Maybe the highlighting can be simplified to something that just looks at the state of the prior and current rr-nodes (without paying attention to their type) and decides if an edge should be highlighted or not.  I suspect we're being overly complex with the code that checks the node types (CHANX etc.).  If we can simplify, we could just move it to a helper function to find the colour.

vaughnbetz (changes requested):
Nice cleanup!
Some comments embedded. 
Biggest one: check the highlighting of rr_nodes works properly; you can simplify the code or change how it works, but we should be able to highlight nodes by (right) clicking on them and we should see their fanin and fanout edges highlighted too.
Future-proofing / next: @amin1377 has added support for a new MUX type.  After this PR is in, you should test with an openFPGA tileable rr-graph on a small device to ensure it draws properly, and you should add MUX support.  MUXes are like 0-length wires; we can draw them as points in a switch block I think but we'll have to allocate space for them etc. so it is probably a reasonable amount of work.
Future proofing 2: @soheilshahrouz has added support for CHANZ nodes. You should make sure those draw too in a future PR.  At least we shouldn't crash, and most code that checks for CHANX and CHANY should have a CHANZ case too.

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -300,39 +301,48 @@ void draw_intra_cluster_edge(RRNodeId inode, RRNodeId prev_node, ezgl::renderer*
...
+    const auto& rr_graph = device_ctx.rr_graph;
 
     if (!draw_state->is_flat) {
         return;
     }
 
-    // determine the location of the pins
-    float inter_cluster_x, inter_cluster_y;
-    ezgl::point2d intra_cluster_coord;
+    for (const e_side& pin_side : TOTAL_2D_SIDES) {
```
soheilshahrouz: side is char or int
don't need to iterate over it by reference

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -610,4 +620,116 @@ void draw_pin_to_chan_edge(RRNodeId pin_node, RRNodeId chan_node, ezgl::renderer
...
+void draw_rr_edge(RRNodeId inode, RRNodeId prev_node, ezgl::color color, ezgl::renderer* g) {
+    auto& device_ctx = g_vpr_ctx.device();
+    auto& rr_graph = device_ctx.rr_graph;
+    t_draw_state* draw_state = get_draw_state_vars();
+
+    e_rr_type rr_type = rr_graph.node_type(inode);
+    bool inode_inter_cluster = is_inter_cluster_node(rr_graph, inode);
+    int current_node_layer = rr_graph.node_layer(inode);
+
+    e_rr_type prev_type = rr_graph.node_type(RRNodeId(prev_node));
```
soheilshahrouz: Why is `prev_node` casted here?

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -610,4 +620,116 @@ void draw_pin_to_chan_edge(RRNodeId pin_node, RRNodeId chan_node, ezgl::renderer
...
+        return;
+    }
+
+    draw_inter_cluster_rr_edge(inode, prev_node, rr_type, prev_type, g);
+}
+
+void draw_inter_cluster_rr_edge(RRNodeId inode, RRNodeId prev_node, e_rr_type rr_type, e_rr_type prev_type, ezgl::renderer* g) {
+    const RRGraphView& rr_graph = g_vpr_ctx.device().rr_graph;
+    t_edge_size iedge = find_edge(prev_node, inode);
+    short switch_type = rr_graph.edge_switch(RRNodeId(prev_node), iedge);
```
soheilshahrouz: is casting `prev_node` required?

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -610,4 +620,116 @@ void draw_pin_to_chan_edge(RRNodeId pin_node, RRNodeId chan_node, ezgl::renderer
...
+
+    switch (rr_type) {
+        case e_rr_type::IPIN: {
```
soheilshahrouz: I think braces are not necessary for `case` statements.
Everything that comes after `case` is executed no matter they are surrounded by braces or not.

vpr/src/draw/draw_rr.cpp
```diff
@@ -275,14 +236,13 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
     const auto& rr_graph = device_ctx.rr_graph;
-    auto rr_node = inode;
-
-    int transparency_factor = get_rr_node_transparency(rr_node);
 
     e_rr_type from_type, to_type;
-    short switch_type;
 
-    from_type = rr_graph.node_type(rr_node);
+    from_type = rr_graph.node_type(inode);
```
soheilshahrouz: It is usually a good idea to have the declaration and definition in the same line, if possible.

vpr/src/draw/draw_rr.cpp
SamuelHo10: Will do in another PR if I have time.

vpr/src/draw/draw_rr.cpp
SamuelHo10: I checked that the RRNode highlighting works. I think currently we have a global variable vtr::vector<RRNodeID, t_draw_rr_node> draw_rr_node; which stores information on whether the RRNode is highlighted or not.

vaughnbetz:
Thanks.  Let's merge this and fine-tune any rough edges after.
For intra-cluster rr-graph dumping: @amin1377 is there an echo file for that?  We should make one if not I think. I think it's intentionally that you don't write it to the rr-graph.xml file, as the nodes are design specific after you have optimized the rr-graph I think.
@SamuelHo10 : how are you drawing intracluster pins?  blue for inputs, pink for outputs, across the top of each sub-block?  Can you click on those rr-nodes too to highlight them and their fan-in and fan-out (that would be very useful!).
For the fle[0]n2_lut5 sub-block, it appears to have only one input, but there would be more in the .arch file.  Are its rr-nodes rendering correctly?

image

vaughnbetz (approved):
LGTM

amin1377:
@vaughnbetz: Sorry for the late reply, I’m not sure how I missed your comment. You’re right: we don’t write out the RR Graph with intra-cluster resources because it’s design-specific, and we don’t have an echo file for that. It would be very helpful to have one, so I’ve created an issue for it and assigned it to myself.

amin1377:
@vaughnbetz: Currently, the GUI does not provide an option to display only intra-cluster routing or only inter-cluster routing (when the flat router is used). Do you think it would be useful to add such an option?

PR #3207, opened 2025-07-23
Capnp Interchange File Directory

amin1377:
Use a relative path based on the current CMake directory to access interchange files. This enables VTR to be used as a submodule in other repositories.


AlexandreSinger (approved):
LGTM.

This is not the ideal way of handling this, but this is better than what we currently do. What we currently do is just as bad.

The better way of doing this is in `libs/CMakeLists.txt` we set a variable like `VTR_EXTERNAL_INTERCHANGE_DIR` and then in this CMakeLists we just use that. It makes the interface a bit more stable. Generally you want to limit how the project structure influences our code. But this is not urgent. Feel free to raise an issue on this point. I am good with merging this PR as is.

PR #3208, opened 2025-07-23
OpenFPGA Patch

amin1377:
This PR updates the code to work with the VTR master branch. Specifically, it removes the static identifier from a few functions and reintroduces two methods in the RR graph builder that I had previously removed by mistake (they were part of VTR but not explicitly visible earlier).


AlexandreSinger (approved):
LGTM

PR #3212, opened 2025-07-24
Add Parsing Code For Scatter-Gather Patterns

AmirhosseinPoolad:
Adds xml parsing code for scatter-gather patterns, with a sample architecture (That doesn't quite make sense) and test to make sure the code actually runs.

Note that scatter-gather patterns are not implemented in VPR and adding the tag does nothing other than parsing the information in the architecture file.

Information about what scatter-gather patterns and what the tag means is available in the documentation for this PR.


libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -137,49 +128,74 @@ void read_sb_wireconns(const std::vector<t_arch_switch_inf>& switches,
...
 
 static t_wireconn_inf parse_wireconn_inline(pugi::xml_node node,
                                             const pugiutil::loc_data& loc_data,
-                                            const std::vector<t_arch_switch_inf>& switches) {
+                                            const std::vector<t_arch_switch_inf>& switches,
+                                            bool can_skip_from_to) {
 
     // Parse an inline wireconn definition, using attributes
-    expect_only_attributes(node, {"num_conns", "from_type", "to_type", "from_switchpoint", "to_switchpoint", "from_order", "to_order", "switch_override"}, loc_data);
+    expect_only_attributes(node, {"num_conns", "from_type", "to_type", "from_switchpoint", "to_switchpoint", "from_order", "to_order", "switch_override", "side"}, loc_data);
```
soheilshahrouz: This line is too long.

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -60,11 +47,15 @@ static void parse_switchpoint_order(const char* order, SwitchPointOrder& switchp
  * @param node             XML node containing inline wireconn attributes.
  * @param loc_data         Location data for error reporting.
  * @param switches         List of architecture switch definitions (used for switch overrides).
+ * @param can_skip_from_to Determines if the from or to attributes are optional or mandatory.
+ * <wireconn> tags for switch blocks require mandatory from/to attributes while those for scatter-gather
+ * patterns are optional.
```
soheilshahrouz: I don't think `from` and `to` are optional for scatter-gather patterns. Depending on whether `<wireconn>` is under scatter or gather tag , one of `from` or `to` tags are required.

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -137,49 +128,74 @@ void read_sb_wireconns(const std::vector<t_arch_switch_inf>& switches,
...
+        parse_comma_separated_wire_points(char_prop, wc.from_switchpoint_set);
+    }
 
     // get the destination wire point
-    char_prop = get_attribute(node, "to_switchpoint", loc_data).value();
-    parse_comma_separated_wire_points(char_prop, wc.to_switchpoint_set);
+    char_prop = get_attribute(node, "to_switchpoint", loc_data, from_to_required).value();
+    if (!can_skip_from_to) {
+        parse_comma_separated_wire_points(char_prop, wc.to_switchpoint_set);
+    }
```
soheilshahrouz, line 201: More error checking is needed. For example, for scatter-gather patterns, a combination of `from` and `to` in <`wireconn`> is illegal. I think the caller of this function should specify what fields are allowed in `<wireconn>`

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -550,6 +551,12 @@ void xml_read_arch(const char* ArchFile,
             process_noc_tag(Next, arch, loc_data);
         }
 
+        // Process scatter-gather patterns (optional)
+        Next = get_single_child(architecture, "scatter_gather_list", loc_data, pugiutil::OPTIONAL);
+        if (Next) {
+            process_sg_tag(Next, arch, loc_data, arch->switches);
```
soheilshahrouz, line 557: Why `arch->switches` is needed for parsing sg tag?

libs/libarchfpga/src/read_xml_arch_file_sg.h
```diff
@@ -0,0 +1,18 @@
...
+ * This function should be called after read_xml_arch_file.cpp:process_switches().
+ * 
+ * @param sg_list_tag XML node pointing to the <scatter_gather_patterns> tag.
+ * @param arch High-level architecture information. This function fills
+ * arch->scatter_gather_patterns with scatter-gather-related information.
+ * @param loc_data Points to the location in the architecture file where the parser is reading. Used for priting error messages.
+ * @param switches Contains all the architecture switches. Usually same as arch->switches.
+ */
+void process_sg_tag(pugi::xml_node sg_list_tag,
+                    t_arch* arch,
```
soheilshahrouz, line 16: Don't pass arch by pointer. We shouldn't follow C-style coding used in libarchfpga.

libs/libarchfpga/src/scatter_gather_types.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <string>
+#include "switchblock_types.h"
+
+/**
+ * @brief Enumeration for the type field of an <sg_pattern> tag.
+ * 
+ */
+enum class e_scatter_gather_type {
+    UNIDIR, // Unidirectional connection
```
soheilshahrouz: doxygen comment for values using `///<`

libs/libarchfpga/src/scatter_gather_types.h
```diff
@@ -0,0 +1,51 @@
...
+ * @brief Struct containing information of a <sg_link> tag. This tag describes how and where the scatter (fanout) happens relative to the gather (fanin).
+ * 
+ */
+struct t_sg_link {
+    std::string name;     // Name of the sg_link.
+    std::string mux_name; // Name of the multiplexer used to gather connections.
+    std::string seg_type; // Segment/wire used to move through the device to the scatter location.
+    int x_offset;         // X offset of where the scatter happens relative to the gather. If set, the y and z offsets must be zero.
+    int y_offset;         // y offset of where the scatter happens relative to the gather. If set, the x and z offsets must be zero.
+    int z_offset;         // z offset of where the scatter happens relative to the gather. If set, the x and y offsets must be zero.
```
soheilshahrouz: doxygen comment for member variables

libs/libarchfpga/src/switchblock_types.h
```diff
@@ -122,6 +145,8 @@ struct t_wireconn_inf {
                                     *          larger than 'from'), or some 'from' elements driving to 'to' elements (if 'from' is
                                     *          larger than 'to')
                                     */
+
+    std::vector<e_side> sides; // Used for scatter-gather wireconns determining which sides to gather from / scatter to, ignored in other usages.
```
soheilshahrouz: use `///` or `///<`

soheilshahrouz: If you have the energy, please chany the style of other comments in this struct.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
     The 'to' set is all L4 switchpoint 0's.
     Note that since different switchpoints are selected from different segment types it is not possible to specify this without using ``<from>`` sub-tags.
 
+.. _scatter_gathern_patterns:
+
+Scatter-Gather Patterns
+---------------------
+
+The content under the ``<scatter_gather_list>`` tag consists of one or more ``<sg_pattern>`` tags that are used to specify a scatter-gather pattern.
+
```
vaughnbetz, line 2655: Give high-level context on why we need this (specify multi-level switch patterns, rather than the direct wire-to-wire switch patterns of conventional switch blocks). These additional switches, wires and/or muxes will be added to the architecture, augmenting wires created using segment specifications and swiches created using switch box specifications.
The number of any additional wires or muxes created by scatter-gather specifications will not vary with routing channel width.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+When instantiated, a scatter-gather pattern gathers connections from a switchblock and passes the connection through a multiplexer and a wire segment, then scatters or fans out somewhere else in the device. These patterns can be used to define 3D switchblocks. An example is shown below:
+
+    .. code-block:: xml
+
+        <scatter_gather_list>
+            <sg_pattern name="name" type="unidir"> <!-- segment: attribute or tag? -->
```
vaughnbetz: Delete self-comment.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+                <gather>
+                    <wireconn num_conns="30" from_type="L16" from_switchpoint="0,12,8,4" side="rltb"/> <!-- Illegal to have to_type and to_switchpoint --> <!-- No above/under side -->
```
vaughnbetz: Clarify in the comment that rltb allowed as source sides, but not other layers (above/under).

vaughnbetz: Do you want this example to describe what it is doing and describe limitations later in the detailed sections?

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+                <gather/>
+
+                <scatter>
+                    <wireconn num_conns="30" to_type="L16" to_switchpoint="0" side="rtlb"/> <!-- Illegal to have from_type and from_switchpoint -->
```
vaughnbetz: Suggest describing what you are doing here, and leaving legality rules for the detailed section later.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+                <scatter/>
+                
+                <sg_link_list>
+                    <sg_link name="L_UP" z_offset="1" x_offset="0" y_offset="0" mux="3D_SB_MUX" seg_type="TSV"/> <!-- One hot offsets -->
```
vaughnbetz: Suggest explaining what you're doing.  Instead of one-hot, either leave for detailed legality later or say only one of x,y,z offset can be non-zero.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+                <sg_location type="EVERYWHERE" num="10" sg_link="L_UP"/> <!-- type="EVERYWHERE/PERIMETER/CORNER/FRINGE/CORE"-->
+                <sg_location type="EVERYWHERE" num="10" sg_link="L_DOWN"/>
+            <sg_pattern/>
+
+            <sg_pattern name="interposer_conn_sg" type="bidir">
+                ... <!-- Another scatter-gather pattern specification-->
+            <sg_pattern/>
+        <scatter_gather_list/>
+
+This format allows users to specify complex switchblock patterns without the verbosity of custom switchblocks.
```
vaughnbetz: Expand on this or move earlier and give higher-level.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+
+.. arch:tag:: <sg_pattern name="string" type={unidir|bidir}>
+
+    :req_param name: A unique alphanumeric string
+    :req_param type: ``unidir`` or ``bidir``.
```
vaughnbetz, line 2695: explain unidir and bidir in this context (briefly) or put a link if you can find the exact text you need elsewhere.

vaughnbetz: unidir: all the gather connections are combined in a mux that then drives the scatter-gather node which in turn drives the wires specified in the scatter specification.
bidir: the gather and scatter connections are mirrored; the same scatter pattern is implemented at each end of the scatter-gather pattern. This implies the two muxes driving the scatter-gather node can have their outputs tri-stated.

doc/src/arch/reference.rst
vaughnbetz: High-level spec needs a picture showing the gather pattern, the scatter-gather node, and the scatter pattern. Link that to the high-level description.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+
+    Contains a <wireconn> tag specifying how the fan-in or gather connections are selected. See ``wireconn`` for the relevant specification.
+
+.. arch:tag:: <scatter>
+
+    Contains a <wireconn> tag specifying how the fan-out or scatter connections are selected. See ``wireconn`` for the relevant specification.
+
+.. arch:tag:: <sg_link_list>
+
+    Contains one or more <sg_link> tags specifying how the connections move from the gather location to the scatter location.
```
vaughnbetz: specifying the pattern of how connections move from

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+    
+    .. note:: These <sg_link> tags are not instantiations of the pattern. instead, the instantiations select one of the sg_link tags to use.
+
+.. arch:tag:: <sg_link  name="string" x_offset="int" y_offset="int" z_offset="int" mux="string" seg_type="string">
+
+    :req_param name: A unique alphanumeric string
+    :req_param mux: Name of the multiplexer used to gather connections
+    :req_param seg_type: Name of the segment/wire used to move through the device to the scatter location
```
vaughnbetz: add "(i.e. the type of the scatter-gather node)"

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+
+    :opt_param x_offset: Offset of the scatter relative to the gather in the x-axis
+    :opt_param x_offset: Offset of the scatter relative to the gather in the y-axis
+    :opt_param x_offset: Offset of the scatter relative to the gather in the z-axis
+
+    .. note:: One and only one of the offset fields for the sg_link tag must be set.
```
vaughnbetz: The magnitude of the offset will generally be chosen by the architecture file creator to match the length of the sg_link segment type.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
...
+
+.. arch:tag:: <sg_location type="string" num="int" sg_link_name="string">
+
+    Instantiates the scatter-gather pattern with the specified sg_link.
+
+    :req_param num: number of scatter-gather instances per switch_block
```
vaughnbetz: per matching location (e.g. per switch block)

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2634,90 @@ The full format is documented below.
     The 'to' set is all L4 switchpoint 0's.
     Note that since different switchpoints are selected from different segment types it is not possible to specify this without using ``<from>`` sub-tags.
 
+.. _scatter_gathern_patterns:
```
vaughnbetz: Should also make an architecture example (or more than one) in the architecture modeling tutorial/examples section.

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -60,11 +47,15 @@ static void parse_switchpoint_order(const char* order, SwitchPointOrder& switchp
  * @param node             XML node containing inline wireconn attributes.
  * @param loc_data         Location data for error reporting.
  * @param switches         List of architecture switch definitions (used for switch overrides).
+ * @param can_skip_from_to Determines if the from or to attributes are optional or mandatory.
+ * <wireconn> tags for switch blocks require mandatory from/to attributes while those for scatter-gather
```
vaughnbetz: ... require both from and to attributes while those for scatter-gather use one or the other ...

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -60,11 +47,15 @@ static void parse_switchpoint_order(const char* order, SwitchPointOrder& switchp
...
+ * patterns are optional.
  * @return                 A `t_wireconn_inf` structure populated with parsed data.
  */
 static t_wireconn_inf parse_wireconn_inline(pugi::xml_node node,
                                             const pugiutil::loc_data& loc_data,
-                                            const std::vector<t_arch_switch_inf>& switches);
+                                            const std::vector<t_arch_switch_inf>& switches,
+                                            bool can_skip_from_to);
```
vaughnbetz: can_skip_from_or_to

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -315,66 +331,24 @@ static void parse_num_conns(std::string num_conns, t_wireconn_inf& wireconn) {
...
+    
+    e_side from_side = CHAR_SIDE_MAP.at(func_type[0]);
+    e_side to_side = CHAR_SIDE_MAP.at(func_type[1]);
+
+    // Can't go from side to same side
+    if (to_side == from_side) {
+        archfpga_throw(__FILE__, __LINE__, "Unknown permutation function specified, cannot go from side to same side: %s\n", func_type);
     }
+
+    // Can't go from above/under to above/under
```
vaughnbetz: Explain why.
"We don't allow specification of patterns that imply edges going over 2 or more layers (this doesn't seem electrically logical), so we disallow going from above/under to above/under."

libs/libarchfpga/src/parse_switchblocks.h
```diff
@@ -42,6 +44,22 @@ void read_sb_wireconns(const std::vector<t_arch_switch_inf>& switches,
...
+/**
+ * @brief Parses a wireconn node and returns a `t_wireconn_inf` structure.
+ *
+ * Determines whether the wireconn is in inline or multi-node format and dispatches
+ * to the appropriate parsing subroutine.
+ *
+ * @param node             XML node representing the wireconn.
+ * @param loc_data         Location data for error reporting.
+ * @param switches         List of architecture switch definitions (used for switch overrides).
+ * @return                 A `t_wireconn_inf` structure populated with parsed data.
```
vaughnbetz, line 59: add can_skip_from_or_to comment.

libs/libarchfpga/src/physical_types.h
```diff
@@ -2242,4 +2244,7 @@ struct t_arch {
 
     // added for vib
     std::vector<VibInf> vib_infs;
+
+    /// Stores information for scatter-gather patterns
```
vaughnbetz: that can be used to define some of the rr-graph connectivity

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "pugixml_util.hpp"
+#include "arch_error.h"
+#include "switchblock_types.h"
+
+/**
+ * @brief Parses all <sg_link> tags under a <sg_link_list> tag.
+ * 
+ * @param sg_link_list_tag XML node pointing to the <sg_link_list> tag.
+ * @param loc_data Points to the location in the architecture file where the parser is reading. Used for priting error messages.
+ * @return std::vector<t_sg_link> containing information of scatter gather links.
```
vaughnbetz: containing -> the information for the sg_links in this list

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+ */
+static std::vector<t_sg_link> parse_sg_link_tags(pugi::xml_node sg_link_list_tag,
+                                                 const pugiutil::loc_data& loc_data) {
+    std::vector<t_sg_link> sg_link_list;
+    pugiutil::expect_only_children(sg_link_list_tag, {"sg_link"}, loc_data);
+    for (pugi::xml_node node : sg_link_list_tag.children()) {
+        if (strcmp(node.name(), "sg_link") != 0) continue;
```
vaughnbetz: Shouldn't this be an error (with a message), not a skip?

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+        const std::vector<std::string> expected_attributes = {"name", "x_offset", "y_offset", "z_offset", "mux", "seg_type"};
+        pugiutil::expect_only_attributes(node, expected_attributes, loc_data);
+
+        t_sg_link sg_link;
+        sg_link.name = pugiutil::get_attribute(node, "name", loc_data).as_string();
+        sg_link.mux_name = pugiutil::get_attribute(node, "mux", loc_data).as_string();
+        sg_link.seg_type = pugiutil::get_attribute(node, "seg_type", loc_data).as_string();
+
+        int x_offset = pugiutil::get_attribute(node, "x_offset", loc_data, pugiutil::OPTIONAL).as_int();
```
vaughnbetz: Just checking -- does get_attribute.as_int() with an OPTIONAL flag return 0 if it is not in the xml?  I'd comment that if so. And if not, I'm not sure how this works ...
@soheilshahrouz thinks there's another default_value method that might be easier to read.

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+        int y_offset = pugiutil::get_attribute(node, "y_offset", loc_data, pugiutil::OPTIONAL).as_int();
+        int z_offset = pugiutil::get_attribute(node, "z_offset", loc_data, pugiutil::OPTIONAL).as_int();
+
+        if (x_offset == 0 && y_offset == 0 && z_offset == 0) {
+            archfpga_throw(loc_data.filename_c_str(), loc_data.line(node), "All offset fields in the <sg_link> are non-zero or missing.");
+        }
+
```
vaughnbetz, line 36: Comment why.  We expect exactly one non-zero offset so the gather and scatter points are joined by a node that can be represented by a straight wire.

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return sg_link_list;
+}
+
+/**
+ * @brief Parses all <sg_location> tags under a <sg_pattern> tag.
+ * 
+ * @param sg_pattern_tag XML node pointing to the <sg_pattern> tag.
+ * @param loc_data Points to the location in the architecture file where the parser is reading. Used for priting error messages.
+ * @return std::vector<t_sg_location> containing information of scatter gather locations.
```
vaughnbetz: containing -> contains

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+ */
+static std::vector<t_sg_location> parse_sg_location_tags(pugi::xml_node sg_pattern_tag,
+                                                         const pugiutil::loc_data& loc_data) {
+    std::vector<t_sg_location> sg_location_list;
+    for (pugi::xml_node node : sg_pattern_tag.children()) {
+        if (strcmp(node.name(), "sg_location") != 0) continue;
```
vaughnbetz, line 74: Shouldn't this be an error message instead of a skip?

vaughnbetz: Or an assert if schema checking would have already caught this?

libs/libarchfpga/src/scatter_gather_types.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include <string>
+#include "switchblock_types.h"
+
+/**
+ * @brief Enumeration for the type field of an <sg_pattern> tag.
```
vaughnbetz: Explain what it controls ... as in the reference, explain that with unidir the gather pattern makes a mux which is connected by a node to the scatter pattern of edges. With bidir, the structure is made symmetric with a gather mux and a scatter edge pattern on each end of the node linking them.

libs/libarchfpga/src/scatter_gather_types.h
```diff
@@ -0,0 +1,51 @@
...
+ * 
+ */
+enum class e_scatter_gather_type {
+    UNIDIR, // Unidirectional connection
+    BIDIR   // Bidirectional connection
+};
+
+/**
+ * @brief Struct containing information of an <sg_location> tag. An <sg_location> tag instantiates the scatter-gather pattern in some switchblocks around the device.
```
vaughnbetz: switchblocks -> switchblock locations (i.e. routing channel intersections)

libs/libarchfpga/src/switchblock_types.h
```diff
@@ -24,6 +26,20 @@ enum e_side : unsigned char {
     NUM_3D_SIDES = 6,
 };
 
+const std::unordered_map<char, e_side> CHAR_SIDE_MAP = {
+    {'T', TOP},
```
vaughnbetz, line 35: I'd comment what these are.  E.g. "At the intersection of routing channels, left, right, top and bottom specify the x- and y-directed channels while above and under specify the switch block wires one a layer above or below the current one.  above and below are only used for multi-layer FPGAs."

vtr_flow/arch/scatter-gather/k6_frac_N10_frac_chain_mem32K_40nm.xml
```diff
@@ -0,0 +1,1597 @@
...
+    <local_interconnect C_wire="2.5e-10"/>
+    <mux_transistor_size mux_transistor_size="3"/>
+    <FF_size FF_size="4"/>
+    <LUT_transistor_size LUT_transistor_size="4"/>
+  </power>
+  <clocks>
+    <clock buffer_size="auto" C_wire="2.5e-10"/>
+  </clocks>
+  
+    <scatter_gather_list>
```
vaughnbetz: Comment what this is doing.
This is a 2D architecture, so this 3D pattern should eventually be moved to a 3D arch (or maybe moved now?)

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
AmirhosseinPoolad: This loop is actually going through all children of the <sg_pattern> tag that might be gather, scatter or sg_link_list.

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
AmirhosseinPoolad: This is what the documentation for as_int says:

> Get attribute value as a number, or the default value if conversion did not succeed or attribute is empty

Default value is one of the optional function arguments of as_int with a default value of 0. So yes it will result in a default value of zero if an attribute does not exist. Added a comment and also explicitly passed 0 to as_int.

vtr_flow/arch/scatter-gather/k6_frac_N10_frac_chain_mem32K_40nm.xml
AmirhosseinPoolad: The purpose of this architecture file is simply testing the parsing code to make sure it doesn't crash and is not intended for usage. I think this makes sense since right now since there's no other functionality to test and parsing is all we can do. After the RR-Graph changes are added I will make a proper 3D architecture instead of this.

Added comments to the file.

libs/libarchfpga/src/parse_switchblocks.cpp
AmirhosseinPoolad: This checking is done in the "process_sg_tag" function. I don't think it makes sense for the wireconn tag to legalize itself based on what's around it, the parent tag should be the one doing the checking.

libs/libarchfpga/src/read_xml_arch_file.cpp
AmirhosseinPoolad: Parsing wireconn requires it for name of the from_type etc. Technically process_sg_tag could get arch->switches from arch itself (Since arch is also passed to it) but doing it that way means that the user of process_sg_tag must take care to call the function after switches are parsed. By adding it as an argument the compiler enforces that.

libs/libarchfpga/src/read_xml_arch_file_sg.h
AmirhosseinPoolad: arch is a pointer outside of this function too. Changing it means changing a fair amount of libarchfpga, VPR and probably parmys too and I don't think it's in the scope of this PR. I could dereference arch and pass it as a reference but I'm not sure if it makes a lot of sense.

doc/src/arch/reference.rst
```diff
@@ -2592,6 +2592,10 @@ The full format is documented below.
         By using a zero-delay and zero-resistance switch one can also create T and L shaped wire segments.
         
         **Default:** If no override is specified, the usual wire_switch that drives the ``to`` wire will be used. 
+    
+    :opt_param side:
+       Specifies the sides that connections are gathered from or scattered to. Valid sides are right, left, top and bottom and are represented by characters 'r', 'l', 't' and 'b'.
```
AlexandreSinger, line 2597: You may want to link forward to the "_scatter_gather_patterns" tag so it can easily be linked. Some thing like "See ..."

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
     The 'to' set is all L4 switchpoint 0's.
     Note that since different switchpoints are selected from different segment types it is not possible to specify this without using ``<from>`` sub-tags.
 
+.. _scatter_gathern_patterns:
```
AlexandreSinger: Small nit-pick, should this be "_scatter_gather_patterns". There seems to be an extra "n". This literally has no affects but it can prevent bugs when people try to use this reference for documentation.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
...
+The content under the ``<scatter_gather_list>`` tag consists of one or more ``<sg_pattern>`` tags that are used to specify a scatter-gather pattern.
+Scatter-gather patterns can be used to specify multi-level switch patterns, rather than the direct wire-to-wire switch patterns of conventional switch blocks.
+These additional switches, wires and/or muxes will be added to the architecture, augmenting wires created using segment specifications and swiches created using switch box specifications.
+The number of any additional wires or muxes created by scatter-gather specifications will not vary with routing channel width.
+
+.. note:: Scatter-Gather patterns are work in progress and experimental. Currently, VPR does not support this specification and using this tag would not result in any modifications to the RR-Graph.
+
+When instantiated, a scatter-gather pattern gathers connections from a switchblock and passes the connection through a multiplexer and a wire segment, then scatters or fans out somewhere else in the device. These patterns can be used to define 3D switchblocks. An example is shown below:
+
+    .. code-block:: xml
```
AlexandreSinger, line 2660: This may take a bit of work, but it may be a good idea to make a figure of this code block. Something simple to help the reader visualize what the scatter-gather patterns do at a high-level.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
...
+            <sg_pattern/>
+        <scatter_gather_list/>
+
+.. arch:tag:: <sg_pattern name="string" type={unidir|bidir}>
+
+    :req_param name: A unique alphanumeric string
+    :req_param type: ``unidir`` or ``bidir``.
+
+Unidir: Added connections are unidirectional; all the gather connections are combined in a mux that then drives the scatter-gather node which in turn drives the wires specified in the scatter specification.
+Bidir: The gather and scatter connections are mirrored; the same scatter pattern is implemented at each end of the scatter-gather pattern. This implies the two muxes driving the scatter-gather node can have their outputs tri-stated.
```
AlexandreSinger: I think you may want to add a return carriage between these two points. It looks a little weird when rendered:
image

I also recommend using single parentheses (``) for the names and not capitilizing them so they match what a user would write. Something like:

`unidir`: ...

`bidir`: ...

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
...
+                ... <!-- Another scatter-gather pattern specification -->
+            <sg_pattern/>
+        <scatter_gather_list/>
+
+.. arch:tag:: <sg_pattern name="string" type={unidir|bidir}>
+
+    :req_param name: A unique alphanumeric string
+    :req_param type: ``unidir`` or ``bidir``.
+
+Unidir: Added connections are unidirectional; all the gather connections are combined in a mux that then drives the scatter-gather node which in turn drives the wires specified in the scatter specification.
```
AlexandreSinger: You never defined what a "scatter-gather node" is. If you add a high-level figure, it may be a good idea to label it.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
...
+.. arch:tag:: <scatter>
+
+    Contains a <wireconn> tag specifying how the fan-out or scatter connections are selected. See ``wireconn`` for the relevant specification.
+    This <wireconn> tag supports an additioinal 'side' attribute which selects the sides that connections are scattered to (e.g. any of the top, bottom, left and right).
+
+.. arch:tag:: <sg_link_list>
+
+    Contains one or more <sg_link> tags specifying how the pattern of how connections move from the gather location to the scatter location.
+    
+    .. note:: These <sg_link> tags are not instantiations of the pattern. instead, the instantiations select one of the sg_link tags to use.
```
AlexandreSinger: This is a little confusing to me. I am not quite sure what this is saying.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
...
+
+    :req_param name: A unique alphanumeric string
+    :req_param mux: Name of the multiplexer used to gather connections
+    :req_param seg_type: Name of the segment/wire used to move through the device to the scatter location (i.e. the type of the scatter-gather node)
+
+    :opt_param x_offset: Offset of the scatter relative to the gather in the x-axis
+    :opt_param x_offset: Offset of the scatter relative to the gather in the y-axis
+    :opt_param x_offset: Offset of the scatter relative to the gather in the z-axis
+
+    .. note:: One and only one of the offset fields for the sg_link tag must be set. The magnitude of the offset will generally be chosen by the architecture file creator to match the length of the sg_link segment type.
```
AlexandreSinger: Nit: "One and only one of the offset ... **can** be set." I think it would sound a little better that way.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
...
+    :req_param num: number of scatter-gather instances per matching location (e.g. per switch block)
+    :req_param sg_link_name: name of the sg_link used in the instantiation
+    :req_param type: Can be one of the following strings:
+
+        * ``EVERYWHERE`` – at each switch block of the FPGA
+        * ``PERIMETER`` – at each perimeter switch block (x-directed and/or y-directed channel segments may terminate here)
+        * ``CORNER`` – only at the corner switch blocks (both x and y-directed channels terminate here)
+        * ``FRINGE`` – same as PERIMETER but excludes corners
+        * ``CORE`` – everywhere but the perimeter
+    Sets the location on the FPGA where the connections described by this scatter-gather pattern be instantiated.
```
AlexandreSinger, line 2742: I am a little confused how this corresponds with "location". For me, location would imply an (x, y) tile position of some sort, not general groups.

Is there any way that we can allow this specification to allow the user to only allow a particular SG pattern to exist at a specified tile coordinate? Perhaps it can be similar to the fixed-layout specification of tiles. It seems a little rigid to assign locations this way.

doc/src/arch/reference.rst
```diff
@@ -2634,6 +2638,104 @@ The full format is documented below.
     The 'to' set is all L4 switchpoint 0's.
     Note that since different switchpoints are selected from different segment types it is not possible to specify this without using ``<from>`` sub-tags.
 
+.. _scatter_gathern_patterns:
+
+Scatter-Gather Patterns
```
AlexandreSinger, line 2643: I recall during the weekly VTR meetings, we discussed putting this in the layout tag. I believe that this tag should be a part of the layout in some way, since the location of these SG patterns do depend on the layout of the device (see my comment on sg_location). This should be updated in the text somewhere in here.

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -137,49 +128,77 @@ void read_sb_wireconns(const std::vector<t_arch_switch_inf>& switches,
...
+    }
+
+    // Parse the optional "side" field of the <wireconn> tag
+    std::string sides_string = get_attribute(node, "side", loc_data, pugiutil::OPTIONAL).as_string();
+
+    if (sides_string.find_first_not_of("rtlbRTLB") != std::string::npos) {
+        archfpga_throw(loc_data.filename_c_str(), loc_data.line(node), "Unknown side specified: %s\n", sides_string.c_str());
+    }
+    for (char side_char : sides_string) {
+        wireconn.sides.push_back(CHAR_SIDE_MAP.at(side_char));
```
AlexandreSinger: Isnt this a little dangerous? What if the user specifies "rrrrrrrrrrrrrrrr... (repeat until out of RAM)"

Or even more realisitically someone wrote "rtlRTL" where they accidentally did caps as well as lower case. This would not catch this case and may cause duplicates to come into the code. I recommend deduping somehow here.

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -137,49 +128,77 @@ void read_sb_wireconns(const std::vector<t_arch_switch_inf>& switches,
...
-    parse_comma_separated_wire_types(char_prop, wc.from_switchpoint_set);
+    char_prop = get_attribute(node, "from_type", loc_data, from_to_required).value();
+    if (!can_skip_from_or_to) {
+        parse_comma_separated_wire_types(char_prop, wc.from_switchpoint_set);
+    }
 
     // get to type
-    char_prop = get_attribute(node, "to_type", loc_data).value();
-    parse_comma_separated_wire_types(char_prop, wc.to_switchpoint_set);
+    char_prop = get_attribute(node, "to_type", loc_data, from_to_required).value();
```
AlexandreSinger, line 186: What if someone did not specify either a from_type or a to_type? Is that acceptable? If not, we should throw an error here during parsing.

libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -315,66 +334,27 @@ static void parse_num_conns(std::string num_conns, t_wireconn_inf& wireconn) {
...
-    } else if (0 == strcmp(func_type, "ur")) {
-        conn.set_sides(UNDER, RIGHT);
-    } else if (0 == strcmp(func_type, "ub")) {
-        conn.set_sides(UNDER, BOTTOM);
-    } else {
-        /* unknown permutation function */
-        archfpga_throw(__FILE__, __LINE__, "Unknown permutation function specified: %s\n", func_type);
+
+    // Only valid sides are right, top, left, bottom, above and under
+    if (std::string(func_type).find_first_not_of("rtlbauRTLBAU") != std::string::npos) {
```
AlexandreSinger, line 343: This is a clever way of doing this.

As well as this check, you may also want to check that the number of charaters in the string is the expected num (2). If not, the user may think they can specify "ltrt" for example as a short-hand (similar to your new interface). It would be nice to give a true error or at least a warning saying the later characters are ignored. I think it should be an error.

AlexandreSinger (commented):
It is looking good so far. I just had some comments on documentation (a figure for this would be really nice) and I had some concerns on error checking in the parser.

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+ * @param sg_link_list_tag XML node pointing to the <sg_link_list> tag.
+ * @param loc_data Points to the location in the architecture file where the parser is reading. Used for priting error messages.
+ * @return std::vector<t_sg_link> the information for the sg_links in the sg_link_list.
+ */
+static std::vector<t_sg_link> parse_sg_link_tags(pugi::xml_node sg_link_list_tag,
+                                                 const pugiutil::loc_data& loc_data) {
+    std::vector<t_sg_link> sg_link_list;
+    pugiutil::expect_only_children(sg_link_list_tag, {"sg_link"}, loc_data);
+    for (pugi::xml_node node : sg_link_list_tag.children()) {
+        const std::vector<std::string> expected_attributes = {"name", "x_offset", "y_offset", "z_offset", "mux", "seg_type"};
```
AlexandreSinger, line 20: Can this be made a constexpr instead of a const? A constexpr may allow the compiler to directly inline this code instead of creating a variable.

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -0,0 +1,139 @@
...
+        sg_link.mux_name = pugiutil::get_attribute(node, "mux", loc_data).as_string();
+        sg_link.seg_type = pugiutil::get_attribute(node, "seg_type", loc_data).as_string();
+
+        // Since the offset attributes are optional and might not exist, the as_int method will return a value of zero if the attribute is empty
+        int x_offset = pugiutil::get_attribute(node, "x_offset", loc_data, pugiutil::OPTIONAL).as_int(0);
+        int y_offset = pugiutil::get_attribute(node, "y_offset", loc_data, pugiutil::OPTIONAL).as_int(0);
+        int z_offset = pugiutil::get_attribute(node, "z_offset", loc_data, pugiutil::OPTIONAL).as_int(0);
+
+        if (x_offset == 0 && y_offset == 0 && z_offset == 0) {
+            archfpga_throw(loc_data.filename_c_str(), loc_data.line(node), "All offset fields in the <sg_link> are non-zero or missing.");
```
AlexandreSinger: Are they non-zero? This error message may be misleading. Shouldn't this say "all offset fields are zero or missing"

libs/libarchfpga/src/scatter_gather_types.h
```diff
@@ -0,0 +1,53 @@
...
+    BIDIR   ///< Bidirectional connection
+};
+
+/**
+ * @brief Struct containing information of an <sg_location> tag. An <sg_location> tag instantiates the scatter-gather pattern in some switchblock locations around the device.
+ * 
+ */
+struct t_sg_location {
+    e_sb_location type;
+    int num;
```
AlexandreSinger: You may want to add a comment explaining these variables. "num" may be a bit vague in this context.

libs/libarchfpga/src/switchblock_types.h
```diff
@@ -24,13 +30,30 @@ enum e_side : unsigned char {
     NUM_3D_SIDES = 6,
 };
 
-constexpr std::array<e_side, NUM_2D_SIDES> TOTAL_2D_SIDES = {{TOP, RIGHT, BOTTOM, LEFT}};                     //Set of all side orientations
-constexpr std::array<const char*, NUM_2D_SIDES> TOTAL_2D_SIDE_STRINGS = {{"TOP", "RIGHT", "BOTTOM", "LEFT"}}; //String versions of side orientations
+const std::unordered_map<char, e_side> CHAR_SIDE_MAP = {
```
AlexandreSinger: Although creating this unordered map is very clever to do this lookup, since this is a const instead of a constexpr this can actually be quite slow to lookup into since the compiler will likely turn this into a global variable in the memory.

I recommend just creating a simple function here which converts a char into the correct "e_side" using a switch statement. This function is likely to be inlined, so it should be performant, and its less prone to errors, as people may index this global map in the wrong way.

Something like:

```cpp
inline e_side char_to_e_side(char c) {
  switch (c) {
    case 'T':
    case 't':
      return e_side::TOP;
    case 'R':
    case 'r':
      return e_side::RIGHT;
    ...
  }
}
```

libs/libarchfpga/src/switchblock_types.h
```diff
@@ -40,6 +63,13 @@ enum class e_sb_location {
     E_XY_SPECIFIED
 };
 
+const std::unordered_map<std::string, e_sb_location> SB_LOCATION_STRING_MAP = {{"EVERYWHERE", e_sb_location::E_EVERYWHERE},
```
AlexandreSinger: Same comment here as e_side.

vtr_flow/arch/scatter-gather/k6_frac_N10_frac_chain_mem32K_40nm.xml
```diff
@@ -0,0 +1,1606 @@
...
+    <local_interconnect C_wire="2.5e-10"/>
+    <mux_transistor_size mux_transistor_size="3"/>
+    <FF_size FF_size="4"/>
+    <LUT_transistor_size LUT_transistor_size="4"/>
+  </power>
+  <clocks>
+    <clock buffer_size="auto" C_wire="2.5e-10"/>
+  </clocks>
+  
+  <!-- Specify a scatter-gather pattern for 3D connections. However, do that this is not a 3D architecture and this file is only used to test the parsing logic for scatter-gather patterns. -->
```
AlexandreSinger: I am confused what this comment trying to say?

vtr_flow/arch/scatter-gather/k6_frac_N10_frac_chain_mem32K_40nm.xml
```diff
@@ -0,0 +1,1606 @@
...
+        
+        <sg_link_list>
+          <!-- Link going up one layer, using the 'sw' multiplexer to gather connections from the bottom layer and using the 'wire' node/wire to move up one layer -->
+          <sg_link name="L_UP" z_offset="1" x_offset="0" y_offset="0" mux="sw" seg_type="wire"/>
+          <!-- Link going down one layer, using the 'sw' multiplexer to gather connections from the top layer and using the 'wire' node/wire to down up one layer -->
+          <sg_link name="L_DOWN" z_offset="-1" mux="sw" seg_type="wire"/>
+        </sg_link_list>
+        
+        <!-- Instantiate 10 'L_UP' sg_links per switchblock location everywhere on the device -->
+        <sg_location type="EVERYWHERE" num="10" sg_link_name="L_UP"/>
```
AlexandreSinger: I still stand by that this would be better suited in the layout somehow and not in the "scatter_gather_list".

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_scatter_gather/config/config.txt
```diff
@@ -0,0 +1,27 @@
...
+circuits_dir=benchmarks/verilog
+
+# Path to directory of architectures to use
+archs_dir=arch/scatter-gather
+
+# Add circuits to list to sweep
+circuit_list_add=ch_intrinsics.v
+
+# Add architectures to list to sweep
+arch_list_add=k6_frac_N10_frac_chain_mem32K_40nm.xml
```
AlexandreSinger: Although you did put this in a special "scatter-gather" folder (which I think is a good idea), I would also modify the name of the arch file to be extra safe. Maybe something like:

`k6_frac_N10_frac_chain_mem32K_sg_40nm.xml`

Just to prevent confusion when people are quickly reading these task files.

AlexandreSinger (commented):
Some final comments. I was only able to get half-way through the PR last time.

doc/src/arch/reference.rst
AmirhosseinPoolad: Added a figure giving a visual overview of scatter-gather patterns.

doc/src/arch/reference.rst
AmirhosseinPoolad: This uses the same syntax as the custom switch blocks. There is an XY_SPECIFIED option which offers more control but since it's not documented (yet) I didn't want to add it here.

doc/src/arch/reference.rst
AmirhosseinPoolad: I think 'must' makes sense here since 'can' implies that it's optional to do so.

doc/src/arch/reference.rst
AmirhosseinPoolad: This is the way custom switchblocks work. It's also a bit easier to implement since it leaves the layout parsing code untouched.

libs/libarchfpga/src/parse_switchblocks.cpp
AmirhosseinPoolad: Good catch, I changed the sides variable to a set instead of a vector. I don't think the memory is a problem (what if you pass a 12 terabyte .c file to gcc? Is it a security issue if it crashes?) but It did allow duplicate sides before.

libs/libarchfpga/src/parse_switchblocks.cpp
AmirhosseinPoolad: So the way I did this is the can_skip_from_or_to argument. By default it is set to false and from/to_type are required attributes, this is used by custom switch blocks and therefore the behavior for those aren't changed at all. For scatter-gather specifications, the error checking is done outside of this function and in process_sg_tag.

Added checking for completely empty fields in process_sg_tag.

libs/libarchfpga/src/parse_switchblocks.cpp
AmirhosseinPoolad: It's pretty clear in the [docs](https://docs.verilogtorouting.org/en/latest/arch/reference/#tag-%3Cfunctype=) what is allowed in that field, however I do agree with enforcing it in the code. Added error checking for func_type that's not of length 2.

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
AmirhosseinPoolad: Oops. Good catch.

libs/libarchfpga/src/scatter_gather_types.h
AmirhosseinPoolad: Added doxygen to the struct.

libs/libarchfpga/src/switchblock_types.h
AmirhosseinPoolad: I actually made a function for this during development but Soheil convinced me to use a map here. I don't think this code matters that much to try and hyper optimize it and we're already doing this approach in a lot of other places in VTR (especially since switch cases can only be used because the input type is char and not string). The proper way to do this would be using compile-time reflection but that doesn't exist in C++ yet.

libs/libarchfpga/src/switchblock_types.h
AmirhosseinPoolad: Yeah since this is not char this would be a mess of a function with a thousand if/elses and comparing strings vs hashing strings isn't very different performance wise.

vtr_flow/arch/scatter-gather/k6_frac_N10_frac_chain_mem32K_40nm.xml
AmirhosseinPoolad: Do note*

What I mean is that the specification is nonsense and the purpose is just testing the parsing code (in the strong tests). I'll add a proper 3D architecture when the RR-Graph changes are implemented.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_scatter_gather/config/config.txt
AmirhosseinPoolad: That makes sense!

libs/libarchfpga/src/switchblock_types.h
vaughnbetz: We don't do string lookups in performance-critical code, so I don't think it matters.

AlexandreSinger: To me its not the performance I care about, it is the unnecessary global variable. Generally its good to avoid global variables. Just a small nitpick since doing this too much can cause bloat to the program for no reason.

AlexandreSinger: Also, the lack of safety from a user giving an invalid character can be a bit of a problem. When the input is a custom enum its not as big of an issue, but the set of all possible chars is quite large.

AlexandreSinger: Its not super important though, so if it causes too much overhead its fine to keep as is. I think compilers should be smart enough to handle this. Especially since you did not declare this variable as "inline", and it compiled without a linker error; which implies that at least the symbol is not global.

libs/libarchfpga/src/read_xml_arch_file_sg.cpp
AmirhosseinPoolad: I tried this but apparently you can't do constexpr vector of strings. I guess vector<string_view> works but I'm not sure if it's worth doing all the work.

AlexandreSinger: Not worth the work! Just honestly curios. Odd that C++ does not allow that.

libs/libarchfpga/src/parse_switchblocks.cpp
AmirhosseinPoolad: Oops, wrong thing in the clipboard. Link fixed.

vaughnbetz:
We can defer the tutorial until after the implementation is done and you have a sample architecture or two, which will likely form the  basis for the tutorial.

libs/libarchfpga/src/switchblock_types.h
AmirhosseinPoolad: > it is the unnecessary global variable

It's a constant though. You can not change it any way so the main issue with global variables (piece of state that anyone and everyone can change at any time which makes the program become a complicated mess) doesn't apply. It's basically equivalent to a macro.

doc/src/arch/reference.rst
AlexandreSinger: This is a matter of perspective; for me "one and only one" implies that one must be set, using the word "can" removes the redundancy. Its really a nit, so use whichever word you prefer!

libs/libarchfpga/src/parse_switchblocks.cpp
AlexandreSinger: I do not think the memory is a problem either, I was just making a joke lol. The duplicates is a problem though since it may be possible that people made a mistake.

libs/libarchfpga/src/switchblock_types.h
AlexandreSinger: "const" is a compiler directive which the compiler does not need to follow. Its to prevent the user from modifying the variable. Since this is defined as a const and not an inline, it is possible that the compiler could make it a global variable (i.e. store it in dedicated memory to be read from). What I am worried about is collisions since this is defined in a header file; so if two different cpp files include the header file they may collide depending on the will of the compiler. I am actually not sure why this does not crash already... Perhaps there is something in the C-standard about const variables defined in header files being implicitly inline...

vaughnbetz: That is worrisome.  Linker errors where you get two global variables with the same name are very nasty.  I would get rid of this then, moving the definition of the global to a single .cpp file, or otherwise changing the implementation to avoid any potential issue.

AlexandreSinger: I found a stack overflow discussion on this topic:
https://stackoverflow.com/questions/12042549/define-constant-variables-in-c-header

It looks like there is some C++ standard magic going on that allows this. According to the post "const variables in a namespace scope has internal linkage". This basically leads to each include getting their own private copy of this variable in this case. This is just confusing in my opinion (I did not even know this fact) and it is a bit wasteful.

I think if we do not create a helper function, we should at least declare this variable as `inline`. This explicitly ensures static linkage without relying on the compiler to infer it. Inline also has a fun property where (I think) it will only allocate one copy in the executable for all usages; but this I am not sure of.

AmirhosseinPoolad: > "const" is a compiler directive which the compiler does not need to follow. 

If a compiler ignores the const keyword it would not be compliant to the C++ standard. A compliant compiler can't just ignore things that aren't implementation defined or undefined and const is neither of these. 

> I think if we do not create a helper function, we should at least declare this variable as inline. This explicitly ensures static linkage without relying on the compiler to infer it. Inline also has a fun property where (I think) it will only allocate one copy in the executable for all usages; but this I am not sure of.

Agreed. I think using inline gives us the best of all worlds here. I'll add that to the maps' definition.

AlexandreSinger: Just wanted to close this thread on my comment on "const". When I say that the compiler does not have to follow it, what I mean is the following. Lets say we made the following code:
```cpp
const int x = 1;
// NOTE: x = x + 1 would yield a semantic error
int y = x + 1;
return y;
```

Since x is a const, the user cannot assign another value to it; however, the compiler is allowed to optimize this code into something functionally equivalent to the following:
```cpp
int x = 1;
x += 1;
return x;
```
Where x is some register. Notice that the compiler is not treating x like a const anymore. In compilers, we actually throw out the "const" keyword after semantic checking (essentially, sometimes its kept around). The "constness" of the variable is guaranteed by the functional correctness of the compilation process. Hence, this is what I mean by "const" being a compiler directive that the compiler does not need to follow.

When you defined this map as a const global variable, you are only ensuring that the user never modifies it; you are not ensuring that the compiler will not, so the compiler may have to create multiple copies of this variable to maintain functional correctness (I think). This is one of those annoying compiler things that may be implementation dependent. We should avoid this pattern the best that we can.

AmirhosseinPoolad: Made the map variables inline. Inline variables fix the duplicate copies issue (which, to emphasize, was strictly a "the binary might be 2 kilobytes larger at worst" issue and nothing functionally broken or slow) and was added in C++17 to help with this specific usecase.
image

The reason I so strongly advocate for using maps instead of functions is:

- The code for any function we would write for this would be a bad implementation of a map anyway (O(n) linear search instead of O(1) lookup. not that performance matters much here.)
- Data is easier to understand than code, and the map implementation is pretty much purely data (from a programmer's perspective)
- Most importantly, the map is defined right next to the enum definition. Anyone changing the enum in the future would see the map right underneath the enum and would update it. In case they don't, it's much easier to catch during review. Contrast this to a function definition which is much more likely to be forgotten to be updated.

vaughnbetz (approved):
LGTM

PR #3214, opened 2025-07-28
Throw if BLIF model doesn't exist in the arch xml

gigeresk:
Adds: Throw an exception if  a BLIF model doesn't exist in the architecture file. Previous behaviour was a segfault.
Also modifies a function that uses `static_cast`ing for invalid state to rather use std::optional.

### DIFF
--- a/vpr/src/base/read_blif.cpp
+++ b/vpr/src/base/read_blif.cpp
@@ -539,7 +539,7 @@ struct BlifAllocCallback : public blifparse::Callback {
         LogicalModelId arch_model_id = models_.get_model_by_name(blif_model.netlist_name());
 
         if(!arch_model_id.is_valid()) {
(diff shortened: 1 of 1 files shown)


AlexandreSinger (approved):
These changes LGTM! Very welcome change, this segmentation fault has hit me a couple of times. I appreciate that you fixed it!

Please see my other comment on the lineno. Its up to you if you want to raise an issue to make this more clear, but I think its OK.

AlexandreSinger:
P.S. Regarding the failing SV testcase. It looks like its nothing. I think the test just needs to be rerun. Sometimes the GH servers have some issues.

gigeresk:
@AlexandreSinger Thanks Alex-- I'll file an issue an re-trigger the tests

PR #3216, opened 2025-07-29
Skip Invalid Atom Blocks

amin1377:
When echo file generation is enabled, a crash occurs during molecule printing. This is caused by not checking the validity of an atom's block while iterating over the atoms in a molecule. This PR adds a check to ensure the block is valid before accessing it, preventing the crash.


AlexandreSinger (approved):
LGTM, yes this is definately needed. Almost all molecules will have valid atom IDs inside of them, but larger molecules (like carry chains for example) may have invalid atom IDs within them (due to a site location being unused in the molecule).

PR #3221, opened 2025-08-01
Upgrading to Yosys v0.55 and Updating Slang

loglav03:
Upgraded both yosys and yosys-slang to more recent versions, and provided explicit top module selection for HDL designs.

#### Description

Upgraded Yosys to v0.55 and upgraded yosys-slang to more recent version (55e3a2a). Modified synthesis flow to explicitly select the top modules for the given HDL instead of solely relying on the yosys `hierarchy -auto-top` command to automatically select top modules. Also added some common koios benchmarks to the top_map array for top module selection.

#### Related Issue

Supports: #3195

#### Motivation and Context

Upgrading yosys and slang to more recent versions could potentially help solve some of the issues with latch inference occurring in some HDL designs. And providing explicit top module selection is required by slang to avoid errors, and is better practice than relying on automatic selection of top modules.

Ran tests locally: These were my own test clones of vtr_reg_nightly_test3/vtr_reg_qor and vtr_reg_nightly_test4/koios_medium
Ran and passed CI tests.

Local tests still have failures because still waiting on benchmark modifications in #3195

#### Types of changes

#### Checklist:

#### Summary of QoR

Tests Used: **vtr_reg_qor**, **vtr_reg_qor_chain**, **koios_medium**, **koios_other**
Tables Used: 
[QoR_Tables.zip](https://github.com/user-attachments/files/21688631/QoR_Tables.zip)
[Summary_of_Tables.xlsx](https://github.com/user-attachments/files/21688630/Summary_of_Tables.xlsx)

Most metrics have minimal differences. At most they will differ by 1% - 2%


libs/EXTERNAL/CMakeLists.txt
AlexandreSinger, line 1: Since we are removing the patch from occuring, are you also able to remove the `patch_slang.cmake` file?

vtr_flow/misc/yosys/synthesis.tcl
```diff
@@ -1,10 +1,10 @@
...
 plugin -i parmys
 
-read_verilog -nomem2reg +/parmys/vtr_primitives.v
-setattr -mod -set keep_hierarchy 1 single_port_ram
-setattr -mod -set keep_hierarchy 1 dual_port_ram
-setattr -mod -set keep 1 dual_port_ram
+#read_verilog -nomem2reg +/parmys/vtr_primitives.v
+#setattr -mod -set keep_hierarchy 1 single_port_ram
+#setattr -mod -set keep_hierarchy 1 dual_port_ram
+#setattr -mod -set keep 1 dual_port_ram
```
AlexandreSinger: Commented out code should be removed. Also it looks like this code was moved, but this last line was not moved over. Was that intentional?

AlexandreSinger (changes requested):
Thanks @loglav03 , I just have a couple of comments.

The self-hosted CI runner is currently down right now and I am working on getting it up and running again. Once its back up, I want to run the weekly tests on this branch before we merge it. I do not expect much change to the golden results based on the changes in this PR, but its always good to check.

libs/EXTERNAL/CMakeLists.txt
loglav03: That's correct. The file is now removed

vtr_flow/misc/yosys/synthesis.tcl
loglav03: Thanks, I missed that. That was intentional. The lines that were kept were in the original synthesis.tcl file before my changes, and the line that wasn't moved over was put there when I was dealing the initial slang errors with the primitives file.

loglav03:
@AlexandreSinger - Before we merge, I'm looking into replacing the -auto-top option with -top <module(s)> in the yosys `hierarchy` command in synthesis.tcl. This for some reason is giving me a syntax error when using -top, so this may take a bit to get working.

loglav03:
@AlexandreSinger - That plan sounds good to me! Good news is I have most of the benchmarks already mapped to their top modules for slang usage so it shouldn't be too difficult to get it set up the way you described.

loglav03:
> > @AlexandreSinger - I've completed the QoR comparisons for this. Should vtr_reg_qor, vtr_reg_qor_chain, and koios_medium be enough tests for this?
> 
> Hi @loglav03 , those are pretty good. Koios_medium does not do all of the koios testcases though, and they are all a bit small. To be extra safe, could you please also do the `koios_other` tests? These include a few of the larger Koios benchmarks which I am interested how they changed.
> 
> These can be found in `vtr_reg_nightly_test6/koios_other/`.
> 
> This should give us a good breadth of how the change to the versions changed the Koios benchmark suite.

Koios_other has now been added!

AlexandreSinger:
Thanks @loglav03 , the koios_other results looks great! If anything this is improving the quality quite substantially. WL improved by around 1% and CPD improved by around 4% (however, CPD tends to be noisy for these circuits). Overall it looks like this change is consistent across these benchmarks. I am ok with merging. I will update the branch and merge when it passes the CI again (good to be safe and the CI is quick now).

AlexandreSinger (approved):
LGTM

vaughnbetz:
Thanks @loglav03 and @AlexandreSinger !  Great to be up to date.

PR #3222, opened 2025-08-05
[WIP] Change CI Containers Workflow to the Modern Versions

AmirhosseinPoolad:
WIP, hoping it would result in better caching and faster CI.


AmirhosseinPoolad:
Looking at the runtimes for previous action runs, this shaves a minute or two from the containers workflow at best, since the only thing we can cache is package installation. Most of the time is spent building VTR from scratch. Somehow making the docker build use ccache would be the way to go here. Closing since I don't think it's worth the work.

PR #3223, opened 2025-08-06
[CI] Updated Weekly Golden Tests

AlexandreSinger:
Four of the NightlyTests were failing for known reasons.

The chain predictor being off likely slowed down sligthly which is not of a concern.

The 3D sb testcases were known to have changed results.

The AP Titan testcases expectedly changed.


AmirhosseinPoolad:
Ooh I didn't catch that. Well yeah let's merge this now. LGTM.

PR #3224, opened 2025-08-06
[APPack] Initial Placement Fallback

AlexandreSinger:
This is the first PR bringing back changes from the AP tuning branch into the main branch of VTR.

This PR adds the ability for the AP initial placer of APPack to fallback on the original initial placer implementation if it fails to find a valid initial placement. This only occured in two testcases from the Koios benchmark suite.


vpr/src/place/initial_placement.cpp
```diff
@@ -1713,7 +1713,7 @@ static inline std::vector<ClusterBlockId> get_sorted_clusters_to_place(
  * Subsequent passes will then try to place clusters at exponentially farther
  * distances.
  */
-static inline void place_blocks_min_displacement(std::vector<ClusterBlockId>& clusters_to_place,
+static inline bool place_blocks_min_displacement(std::vector<ClusterBlockId>& clusters_to_place,
```
amin1377, line 1750: Just a nitpick: I think the function’s comment should mention that if it cannot find a valid initial placement, it will return false.

vpr/src/place/initial_placement.cpp
```diff
@@ -1862,7 +1863,7 @@ static inline void place_blocks_min_displacement(std::vector<ClusterBlockId>& cl
  * each stage, the clusters are ordered based on heuristics such that the most
  * impactful clusters get first dibs on placement.
  */
-static inline void place_all_blocks_ap(enum e_pad_loc_type pad_loc_type,
+static inline bool place_all_blocks_ap(enum e_pad_loc_type pad_loc_type,
```
amin1377, line 1901: Same comment as above

vpr/src/place/initial_placement.cpp
```diff
@@ -1892,13 +1893,17 @@ static inline void place_all_blocks_ap(enum e_pad_loc_type pad_loc_type,
...
-        place_blocks_min_displacement(constrained_clusters,
+        bool all_clusters_placed = place_blocks_min_displacement(constrained_clusters,
                                       pad_loc_type,
                                       blk_loc_registry,
                                       place_macros,
                                       cluster_netlist,
                                       flat_placement_info);
         VTR_LOG("\n");
+        if (!all_clusters_placed) {
+            VTR_LOG("Could not place all constrained clusters, falling back on original IP.\n");
```
amin1377: Another nitpick: I think it’s better to avoid abbreviations in logs whenever possible. I suggest changing “IP” to “initial placement.”

vpr/src/place/initial_placement.cpp
```diff
@@ -1924,13 +1929,17 @@ static inline void place_all_blocks_ap(enum e_pad_loc_type pad_loc_type,
...
-        place_blocks_min_displacement(large_macro_clusters,
+        bool all_clusters_placed = place_blocks_min_displacement(large_macro_clusters,
                                       pad_loc_type,
                                       blk_loc_registry,
                                       place_macros,
                                       cluster_netlist,
                                       flat_placement_info);
         VTR_LOG("\n");
+        if (!all_clusters_placed) {
+            VTR_LOG("Could not place all large macros, falling back on original IP.\n");
```
amin1377: Same as above

vpr/src/place/initial_placement.cpp
```diff
@@ -1991,13 +2006,29 @@ void initial_placement(const t_placer_opts& placer_opts,
...
-                                flat_placement_info);
-        } else {
-            //Assign scores to blocks and placement macros according to how difficult they are to place
+            all_blocks_placed = place_all_blocks_ap(placer_opts.pad_loc_type,
+                                                    blk_loc_registry,
+                                                    place_macros,
+                                                    flat_placement_info);
+
+            // If AP failed to place all of the blocks, reset the placement solution
+            // so we can fall back on the original initial placement algorithm.
```
amin1377, line 2054: I was thinking it might be better to find a different name than “original initial placement.” I don’t have a good alternative in mind at the moment, but I wanted to bring it up with you.

amin1377 (changes requested):
Thanks, Alex!

vpr/src/place/initial_placement.cpp
AmirhosseinPoolad: Would non-AP work here?

AlexandreSinger: I like non-AP

AlexandreSinger:
Thanks for the comments @amin1377 !

While testing this code, I also found an old bug where the centroid of clusters was calculated to be off-device which was causing some accesses out of bounds. I have resolved this.

Let me know if you have any further comments!

PR #3225, opened 2025-08-07
[Build][Version] Reverted Version Tagging Change

AlexandreSinger:
A recent cleanup I made of the way VTR does its versioning added the "--tags" argument to the git describe command. The hope of this option was that it would enable VTR to point to the v9.0.0 tag instead of the v8.0.0 tag; however, this was not the case.

In some cases, it was able to find the v9.0.0 tag; however, since this tag is on another branch, many users are unable to find that tag from git describe. This caused the versioning code to fall back on v9.0.0-candidate which is the last tagged version on the branch. This tag is the branching point of v9.0.0, but we do not want to use this tag since 1) it does not use the correct naming scheme (the "-candidate" causes problems with downstream tools) and 2) this is not a complete tag (i.e. it lacks metadata).

I reverted this change, so now git describe should stably point to the last full tagged release of VTR on the master branch (which is v8.0.0). This will at least ensure that the version VTR produces is stable. We prefer a stable versioning system based off v8.0.0 over an unstable versioning system based off v9.0.0.

### DIFF
--- a/libs/libvtrutil/CMakeLists.txt
+++ b/libs/libvtrutil/CMakeLists.txt
@@ -54,7 +54,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 include(GetGitRevisionDescription)
 
 # Get the VCS revision.
-git_describe_working_tree(VTR_VCS_REVISION --tags --always --long)
+git_describe_working_tree(VTR_VCS_REVISION --always --long)
 string(FIND ${VTR_VCS_REVISION} "NOTFOUND" GIT_DESCRIBE_VTR_REVISION_NOTFOUND)
 if (NOT ${GIT_DESCRIBE_VTR_REVISION_NOTFOUND} EQUAL -1)
     # Git describe failed, usually this means we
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
@AmirhosseinPoolad The issue we are running into downstream is that the versioning provided by VTR is not consistent. The version needs to have a consistent pattern so downstream tools can check if a user has a version of the tool which is equal or greater than some known good version.

The issue we are running into is that sometimes the user would have one of 4 possible versions:

- v8.0.0-<commits_since_v8.0.0>-...
- v9.0.0-<commits_since_v9.0.0>-...
- v9.0.0-candidate-<commits_since_v9.0.0-candidate>-...
- The hash of the commit (and only the hash of the commit

The last one appears to sometimes occur; which is a major problem. The issue is that these cases are too inconsistent and it is very challenging for downstream tools to handle this.

I agree with you that having v9.0.0 would be much better, but due to how the release happened it causes too many issues.
I propose we return to v8.0.0 for the time being (merge this current PR), and then (in a separate PR) investigate moving to a higher version. I think the simplest solution would be to release a v9.0.1 extremely close to the v9.0.0 branching point which contains all the changes we want.

AmirhosseinPoolad:
LGTM, let's discuss further actions sometime later.

PR #3226, opened 2025-08-07
[CI] Moved Sanitized Build Test from Odin to Parmys

### LINKED ISSUE #3191: [CI] Move Sanitized Build from Odin to Yosys
The current sanitized build test is testing Odin only and not Yosys. This should use Yosys instead since that is the default flow.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/605fb00eeeaff7679831866b5d03bb589ac578ef/.github/workflows/test.yml#L454-L456

AlexandreSinger:
The sanitized build test was testing the Odin build and not the parmys build (which is the default). Moved the test to Parmys.

Resolves #3191

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -471,8 +471,8 @@ jobs:
         include: [
           {
             name: 'Basic',
-            params: '-DVTR_ASSERT_LEVEL=3 -DVTR_ENABLE_SANITIZE=on -DVTR_IPO_BUILD=off -DWITH_BLIFEXPLORER=on -DWITH_PARMYS=OFF -DWITH_ODIN=on',
-            suite: 'vtr_reg_basic_odin'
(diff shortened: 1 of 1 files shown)


AmirhosseinPoolad:
I investigated this a little bit. For some reason parmys is being compiled with '-g  -fsanitize=address  -fsanitize=leak  -fuse-ld=gold  -static-libasan' (Look at build/build.ninja) when by all means it shouldn't (we're adding the parmys subdirectory before setting the various compile options for VPR). I'm not quite sure why but my suspicion is that all the PUBLIC stuff in various cmakefiles has something to do with it (parmys depends on a bunch of vtr internal libraries which do get compiled with asan).

I'm also not entirely sure why it's erroring out. I think we're compiling yosys without asan (which is good and intended) but yosys is trying to dynamically load parmys.so which has some asan function declarations without any definitions. The loader then complains about this and crashes the program.

I don't really have any idea how to fix any of this. We might need a cmake wizard.

PR #3230, opened 2025-08-08
[OpenSTA] Updated SDF Generation with Internal Registers

AlexandreSinger:
Fixed an issue with how the post-implementation SDF annotations are generated for black boxes with internal registers on the input or output ports (but not both).

When neither of the input or output ports are registered, then the circuit would be fully combinatorial. This behavior is unchanged.

mae-timing-model-combinational

When both of the input and output ports are registered, then the input ports would have setup and hold constraints. The output ports would have clock to Q delays. This did not annotate correctly for multi-bit ports, so I updated the code to apply the annotations per pin in the port.

mae-timing-model-full-registered

When the input port is registered and the output port is not, then the input ports would have setup and hold constraints; however originally the output ports did not have clock to Q delays. In a black-box model, the output ports would have a clock to Q delay equal to the clock to Q delay of the input ports plus the combinatorial delays from the input ports to the output ports. I have added this logic to the SDF generation code to include these delays.

mae-timing-model-input-registered

When the output port is registered and the input port is not, then the output ports would have clock to Q delays (as normal); however originally the input ports were not given setup or hold constraints (these were applied to the output ports by how the architecture file traditionally defines these ports). For a black-box, the input ports would have setup and hold constraints, there is just a bit of delay before the timing arc hits the registers. Added the proper setup and hold constraints and applied port delays on the input ports equal to their propagation delay to the output ports.

mae-timing-model-output-registered


vpr/src/base/netlist_writer.cpp
```diff
@@ -1813,16 +1892,110 @@ class NetlistWriterVisitor : public NetlistVisitor {
...
+                                                     pin->associated_clock_pin->pin_number,
+                                                     delay_triple);
+                    }
+                } else if (top_pb_route.count(cluster_pin_idx) > 0) {
+                    // If this pin is not sequential, but has a timing path which
+                    // ultimately leads to an internal register, then this pin
+                    // will have setup and hold constraints. This pin will absorb
+                    // the maximum setup and hold constraints of all paths. To
+                    // account for the internal delays of the black-box, we add
+                    // port delays equal to the max/min path delays to registers.
```
AlexandreSinger, line 1920: Create a bolded comment explaining assumptions on the timing graph within primitive blocks. -Peter

vaughnbetz: An alternative way to achieve this is to have a combinational delay from the input port to the (registered) output port, which has a uTsu and uTh as well as a uTco.  That's how they are defined in the architecture file, and how Tatum models and analyzes them.    
Can liberty models do the same thing?  If so, you could consider just using that syntax as it is closer to what VPR does internally. If not, I would explicitly write a comment explaining that you are converting from one style to the other.

vaughnbetz: See https://docs.verilogtorouting.org/en/latest/tutorials/arch/timing_modeling/ for some examples of the modeling I'm talking about.

AlexandreSinger:
@vaughnbetz After discussions with Peter, we decided that keeping this change in timing model to be the best approach at this time.

The issue we are running into is that in order to create a proper timing model similar to Tatum, we need to define internal timing nodes in the liberty model. The issue is that these internal timing nodes need names (so the SDF file can annotate them) and those names need to have correspondance with VPR in some way. The problem is that there are no standard naming convention for these internal timing nodes (i.e. suppose an input/output port is registered, then this would be two timing nodes: the "D" part of the register and the "Q" part of the register). We need VPR to have some common naming convention for this or allow it to be specified in the architecture file (which would be preferred).

For the time being we will keep this current approach since it is pessimistic, so accurate for our use case, and it is better than what was there before (prior, no annotations were created at all). In the future, I hope to revisit this since having a timing model more similar to Tatum would be ideal.

vaughnbetz (approved):
Thanks for the detailed block comment. This looks good.

PR #3231, opened 2025-08-10
[APPack] Tuned Up APPack Code

AlexandreSinger:
This is the second PR that brings in private tunings for the Triple-AP work into VTR master.

These are all the tuning changes to the APPack algorithm. This comes with changes to the max displacement threshold values, improved fall-back logic, and better unrelated clustering logic.

Also fixed a small issue in the delay derivative calculation of the analytical solver to explicitly handle boundary conditions of the derivative.


doc/src/vpr/command_line_usage.rst
```diff
@@ -1384,6 +1384,38 @@ Analytical Placement is generally split into three stages:
 
     **Default:** ``auto``
 
+.. option:: --appack_max_dist_th { auto | <regex>:<float>,<float> }
```
amin1377: I’d suggest changing the name to something like --appack_unrelated_cluster_max_dist_th (or similar) to make it clear that this threshold applies when unrelated clustering is used.

doc/src/vpr/command_line_usage.rst
```diff
@@ -1384,6 +1384,38 @@ Analytical Placement is generally split into three stages:
...
+
+   Sets parameters used for unrelated clustering for the logical block types
```
amin1377: I don’t think mentioning “logical block types” is necessary here, as it could confuse users who are not familiar with these terms. Also, the first sentence should provide a quick summary of what this parameter does (i.e., set the distance for unrelated clustering of blocks), as the current wording feels a bit vague.

doc/src/vpr/command_line_usage.rst
```diff
@@ -1384,6 +1384,38 @@ Analytical Placement is generally split into three stages:
...
+   used by APPack. APPack uses the primitive-level placement produced by the
+   global placer to cluster primitives together. APPack uses this information
+   to help increase the density of clusters (if needed) by searching for
+   unrelated molecules to pack together. It does this by searching out from
+   the centroid of the cluster being created until it finds a valid molecule.
+   This argument allows the user to select the maximum distance the code will
+   search and how many attempts it will try to search for each cluster.
```
amin1377, line 1399: I also suggest mentioning why *multiple* attempts are made, as it’s not clear to someone unfamiliar with the algorithm where and why these attempts occur.

libs/libarchfpga/src/arch_util.cpp
```diff
@@ -1029,6 +1029,32 @@ bool pb_type_contains_blif_model(const t_pb_type* pb_type, const std::string& bl
     return false;
 }
 
+bool pb_type_contains_memory_pbs(const t_pb_type* pb_type) {
+    // TODO: This should eb a graph traversal instead of a recursive function.
```
amin1377: *be a graph traversal

amin1377: I think modifying this function to use a queue and a while loop to pop from it shouldn’t be too hard. If it’s easy to implement, I’d suggest fixing it now; otherwise, leave a TODO comment.

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -899,8 +910,10 @@ std::pair<double, double> B2BSolver::get_delay_derivative(APBlockId driver_blk,
 
     int tile_dx = sink_block_loc.x - driver_block_loc.x;
     int tile_dy = sink_block_loc.y - driver_block_loc.y;
-    VTR_ASSERT_SAFE(tile_dx < (int)device_grid_width_);
-    VTR_ASSERT_SAFE(tile_dy < (int)device_grid_height_);
+    // The following asserts are to protect the "delay" calls used for the forward
+    // difference calculations below.
+    VTR_ASSERT_SAFE(tile_dx + 1 < (int)device_grid_width_);
+    VTR_ASSERT_SAFE(tile_dy + 1 < (int)device_grid_height_);
```
amin1377, line 916: One unrelated question: I assume that in the CI tests, we still have cases where VPR is built with an elevated assertion level; is that correct?

vpr/src/pack/appack_context.h
```diff
@@ -157,4 +115,9 @@ struct APPackContext : public Context {
     // When selecting candidates, what distance from the cluster will we
     // consider? Any candidate beyond this distance will not be proposed.
     APPackMaxDistThManager max_distance_threshold_manager;
+
+    // When performing unrelated clustering, the following manager class decides
+    // how far we should search for unrelated candidates and how many attempts
+    // we should perform.
+    APPackUnrelatedClusteringManager unrelated_clustering_manager;
```
amin1377, line 122: A much cleaner solution!

vpr/src/pack/appack_max_dist_th_manager.h
```diff
@@ -58,16 +58,20 @@ class APPackMaxDistThManager {
     static constexpr float io_max_dist_th_scale_ = 0.5f;
     static constexpr float io_max_dist_th_offset_ = 15.0f;
 
+  public:
+    // When packing fails, it may try to increase the max distance threshold to
+    // resolve the failure. This scales the max distance threshold of the failing
+    // logical block type. Increasing this value will reduce the number of fallbacks
+    // but may yield worse quality. This is expected to be a number larger than 0.
+    static constexpr float max_dist_th_fail_increase_scale = 10.0f;
```
amin1377, line 66: Scaling the scale by 10 after each failure seems a bit excessive, doesn’t it?

vpr/src/pack/appack_unrelated_clustering_manager.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // that are very close to the cluster being created.
+    // NOTE: Molecules within the same tile as the centroid are considered to have
+    //       0 distance. The distance is computed relative to the bounds of the
+    //       tile containing the centroid.
+    max_unrelated_tile_distance_.resize(logical_block_types.size(),
+                                        default_max_unrelated_tile_distance_);
+    max_unrelated_clustering_attempts_.resize(logical_block_types.size(),
+                                              default_max_unrelated_clustering_attempts_);
+
+    // For memories, we do not perform unrelated clustering.
```
amin1377: I assume you are concerned only with BRAMs here, not MLABs; is that correct? In any case, I’d suggest mentioning whether MLABs are considered or not.

amin1377: Also mentioning it in the function’s comment (pb_type_contains_memory_pbs)

vpr/src/pack/appack_unrelated_clustering_manager.h
```diff
@@ -0,0 +1,152 @@
...
+#include <vector>
+#include "physical_types.h"
+#include "vtr_assert.h"
+
+/**
+ * @brief Manager class for unrelated clustering in APPack.
+ *
+ * After searching for candidates by connectivity and timing, the user may
+ * turn on unrelated clustering, which will allow molecules which are
+ * unrelated to the cluster being created to be attempted to be packed in.
```
amin1377: I’d suggest briefly mentioning what you mean by “unrelated cluster.”

amin1377 (changes requested):
Thanks, Alex! Overall, it looks good to me. I just mainly asked to expand on some comments so that a novice like me can also understand what’s going on :)

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: Hi Amin, yes that is correct. I actually have all builds (that run anything) build with assert level 3:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/18f2b8265cc09e179a008aad985734c70369157f/.github/workflows/test.yml#L56-L63

This will exercise all asserts that are SAFE or harder:
https://docs.verilogtorouting.org/en/latest/api/vtrutil/logging/#vtr-assertion

Note: VTR_ASSERT_DEBUG (which is at assert level 4) is exercised only for the basic tests:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/18f2b8265cc09e179a008aad985734c70369157f/.github/workflows/test.yml#L247-L256

This particular assert that I am adding here is very rare to occur, however, so I wanted to solve it in this PR. It happened on LU32 randomly while working on this PR.

vpr/src/pack/appack_max_dist_th_manager.h
AlexandreSinger: This is a trade-off between run time and quality. For some circuits, we must use unrelated clustering / increasing pin density in order to create a dense enough clustering. If we increase this too slowly, it would take a large number of iterations to reach this solution.

The choice of 10 was originally tied to the lut max distance threshold being 0.01 (at the time), so this would take at most 2 iterations of increasing before moving on. I have since changed the lut max distance threshold to 0.02, so we could decrease this scale to ~7; or even better make it exponentially increase (multiply by 2, then multiply by 20). I would need to experiment to find what would work the best! Its not a bad point. I will do this exploration in another PR.

vpr/src/pack/appack_unrelated_clustering_manager.cpp
AlexandreSinger: This is both BRAMs and MLABs. I found that memory blocks (the logical blocks in the memory class) do not do well with unrelated clustering. I have not directly tested MLABs directly though. I will clean up the comments.

doc/src/vpr/command_line_usage.rst
AlexandreSinger: Thanks for catching this! It looks like I forgot to update the name.

libs/libarchfpga/src/arch_util.cpp
AlexandreSinger: I do agree it would not be hard; however, I do not want to debug this again since I know this works. My though when making this todo was to actually change how the pb_type graph is stored to make it easier to traverse (similar to the netlist objects). That would be a much bigger change that is beyond the scope of AP.

AlexandreSinger: I will leave as a TODO for now.

vpr/src/pack/appack_max_dist_th_manager.h
AlexandreSinger: Actually, thinking back on this; I tried picking default values which make it relatively uncommon for the max_distance threshold to be just barely too low. It may not reveal much quality by reducing the scale factor here. My biggest worry is iteratively doing packing too often (which will really hurt runtime).

PR #3234, opened 2025-08-12
[3D] Added Basic Test and Fixed Bugs in 3D Flow

AlexandreSinger:
While I was experimenting with the 3D flow, I ran into some simple issues in the placement code for a basic 3D FPGA. The issue is that this FPGA has all of the IO tiles on one die and all the CLBs on another. The code was not handling the case when no tile exists on a given die and was hitting assertions and seg faults.

I have fixed these issues and added a basic test to exercise this on a set of basic circuits.


vpr/src/place/initial_placement.cpp
```diff
@@ -421,12 +421,18 @@ static bool find_centroid_neighbor(ClusterBlockId block_id,
                                                              centroid_loc,
                                                              num_layers);
 
+    // If no compressed location can be found on this layer, return false.
+    // TODO: Maybe search in the layers above or below.
+    const t_physical_tile_loc& compressed_loc_on_layer = compressed_centroid_loc[centroid_loc.layer];
+    if (compressed_loc_on_layer.x == OPEN && compressed_loc_on_layer.y == OPEN && compressed_loc_on_layer.layer_num == OPEN)
```
amin1377: If one of the coordinates is OPEN, doesn’t it imply that the others should also be OPEN? If so, I’d suggest checking only one of them and adding an assertion to ensure that the others are likewise set to OPEN.

vpr/src/place/move_utils.cpp
```diff
@@ -852,6 +852,12 @@ bool find_to_loc_centroid(t_logical_block_type_ptr blk_type,
                                                                                          centroid,
                                                                                          num_layers);
 
+    // If no compressed location can be found on this layer, return false.
+    // TODO: Maybe search in the layers above or below.
+    const t_physical_tile_loc& compressed_loc_on_layer = centroid_compressed_loc[to_layer_num];
+    if (compressed_loc_on_layer.x == OPEN && compressed_loc_on_layer.y == OPEN && compressed_loc_on_layer.layer_num == OPEN)
```
amin1377: Same comment as above

vpr/src/place/initial_placement.cpp
AlexandreSinger: This is a good point. I cleaned up the code to make this more clear. I found that the layer does not need to be checked for being OPEN in this context so I rewrote this to only use the x and y positions and added the assert you recommended.

PR #3236, opened 2025-08-13
Updated Route and Net UI Elements

SamuelHo10:
#### Description
I’ve updated the RRGraph UI elements to give users greater flexibility in controlling what’s displayed. Previously, the route UI used dropdown menus that only allowed selecting specific combinations of RR nodes/edges. These have now been replaced with checkboxes and a master toggle, allowing more granular control. I also added options to enable or disable highlighting for both nets and routing, along with checkboxes to show or hide intracluster and intercluster nets.

Additionally, I cleaned up parts of the code related to highlighting, net transparency, and callback functions. Several element names have been renamed. Please review them to ensure they make sense.

#### Testing
image

image

image

image

image

There are some issues with the highlighting when turning both show nets and show RR on, and I think it will be a bit difficult to fix. Should I make these two options mutually exclusive? I think it might be useful though in some cases to show both the RRGraph and routing at once.

Messed up highlighting when both highlight rr edges, display rr, and display nets are turned on:
image


vpr/src/draw/draw_basic.cpp
```diff
@@ -609,19 +606,40 @@ void draw_routed_net(ParentNetId net_id, ezgl::renderer* g) {
 //Draws the set of rr_nodes specified, using the colors set in draw_state
 void draw_partial_route(const std::vector<RRNodeId>& rr_nodes_to_draw, ezgl::renderer* g) {
     t_draw_state* draw_state = get_draw_state_vars();
+    auto& rr_graph = g_vpr_ctx.device().rr_graph;
```
AlexandreSinger: Style: I would not use auto here. Please write out the full type. I think this is also implicitly const, should be written explicitly.

vpr/src/draw/draw_basic.cpp
```diff
@@ -609,19 +606,40 @@ void draw_routed_net(ParentNetId net_id, ezgl::renderer* g) {
...
         RRNodeId inode = rr_nodes_to_draw[i];
         ezgl::color color = draw_state->draw_rr_node[inode].color;
 
+        bool inter_cluster_node = is_inter_cluster_node(rr_graph, inode);
+
+        if (inter_cluster_node && !draw_state->draw_inter_cluster_nets) {
+            continue;
+        }
+
+        if (!inter_cluster_node && !draw_state->draw_intra_cluster_nets) {
```
AlexandreSinger: This if statement is almost identical to the one in line 618. Based on the current logic, the always continue if we are not drawing inter cluster nets, we do not need to check inter_cluster_node. Is this intended?

vpr/src/draw/draw_basic.cpp
```diff
@@ -609,19 +606,40 @@ void draw_routed_net(ParentNetId net_id, ezgl::renderer* g) {
...
         RRNodeId inode = rr_nodes_to_draw[i];
         RRNodeId prev_node = rr_nodes_to_draw[i - 1];
+        bool inter_cluster_node = is_inter_cluster_node(rr_graph, inode);
+        bool prev_inter_cluster_node = is_inter_cluster_node(rr_graph, prev_node);
+
+        if ((inter_cluster_node && prev_inter_cluster_node) && !draw_state->draw_inter_cluster_nets) {
+            continue;
+        }
+
+        if ((!inter_cluster_node || !prev_inter_cluster_node) && !draw_state->draw_intra_cluster_nets) {
```
AlexandreSinger: Please add comments over this if statement and the one at 636. These are a bit hard to follow.

vpr/src/draw/draw_rr.cpp
```diff
@@ -45,45 +45,52 @@ void draw_rr(ezgl::renderer* g) {
...
-                default:
-                    break;
-            }
+
+        e_rr_type node_type = rr_graph.node_type(inode);
+        bool inter_cluster_node = is_inter_cluster_node(rr_graph, inode);
+        bool node_highlighted = draw_state->draw_rr_node[inode].node_highlighted;
+
+        // Node colors by types
+        static const std::map<e_rr_type, ezgl::color> node_colors = {
```
AlexandreSinger: Making this static is a bit strange. static as a particular meaning in this context when it is used within a function. You should just use const in this case. Also, we do not care about order in this case, so an unordered_map would be more appropriate.

However, I would consider not using a map at all in this case. The e_rr_type is an enum that goes from 0 to the number of e_rr_types. I recommed using an array in this case. You can then make it a constexpr, which will have the desired affect. Something like:
```cpp
constexpr vtr::array<e_rr_type, const char*, (size_t)e_rr_type::NUM_RR_TYPES> node_colors ...
```

AlexandreSinger: If this is too complicated, it may just be easiest to make a helper function for "get_node_color". That way you can add safety checks and other debug information.

vpr/src/draw/draw_rr.cpp
```diff
@@ -262,31 +263,65 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
 
         ezgl::color color = DEFAULT_RR_NODE_COLOR;
+        e_edge_type edge_type;
+        bool draw_edge = true;
+        bool inode_inter_cluster = is_inter_cluster_node(rr_graph, inode);
+        bool to_node_inter_cluster = is_inter_cluster_node(rr_graph, to_node);
 
         // Color map for edges based on {from_type, to_type}
-        static const std::map<std::pair<e_rr_type, e_rr_type>, ezgl::color> edge_color_map = {
+        static const std::map<std::pair<e_rr_type, e_rr_type>, e_edge_type> edge_type_map = {
```
AlexandreSinger: Same comment here. The use of the `static` keyword in a function definition creates a global variable. I do not recommend using it here. I recognize it is not your code, but if you could please fix it I would appreciate it! I am trying to track down cases like this in VTR.

vpr/src/draw/draw_rr.cpp
AlexandreSinger: Note: The main thing I am concerned about here is the use of the `static` keyword.

vpr/src/draw/draw_rr.cpp
```diff
@@ -262,31 +263,65 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
+            draw_edge = false;
+        }
+
+        // Special case for direct connections between output pins and input pins of clb.
+        if (edge_type == e_edge_type::PIN_TO_IPIN && inode_inter_cluster && to_node_inter_cluster) {
+            draw_edge = draw_state->draw_connection_box_edges;
+        }
+
+        // Select edge colors
+        static const std::map<e_edge_type, ezgl::color> edge_color_map = {
```
AlexandreSinger: Please remove `static` keyword.

There are times that using this keyword can be helpful; but I do not think it is helping much in this case since this map is so simple.

vpr/src/draw/draw_toggle_functions.cpp
```diff
@@ -18,70 +18,84 @@
...
+        *data->toggle_state = false;
+    }
+
+    data->app->refresh_drawing();
+}
+
+void toggle_show_nets_cbk(GtkSwitch*, gboolean state, ezgl::application* app) {
+    t_draw_state* draw_state = get_draw_state_vars();
+
+    bool switch_state = state ? true : false;
```
AlexandreSinger: Why is this line necessary? Why not just cast to a boolean when it is used?

vpr/src/draw/draw_toggle_functions.cpp
```diff
@@ -18,70 +18,84 @@
...
-    else if (strcmp(combo_box_content, "Nodes SBox") == 0)
-        new_state = DRAW_NODES_SBOX_RR;
-    else if (strcmp(combo_box_content, "Nodes SBox CBox") == 0)
-        new_state = DRAW_NODES_SBOX_CBOX_RR;
-    else if (strcmp(combo_box_content, "Nodes SBox CBox Internal") == 0)
-        new_state = DRAW_NODES_SBOX_CBOX_INTERNAL_RR;
-    else
-        // all rr
-        new_state = DRAW_ALL_RR;
+    bool switch_state = state ? true : false;
```
AlexandreSinger: Same comment as above.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -612,9 +618,8 @@ void draw_logical_connections(ezgl::renderer* g) {
 
     g->set_line_dash(ezgl::line_dash::none);
 
-    //constexpr float NET_ALPHA = 0.0275;
-    float NET_ALPHA = draw_state->net_alpha;
-    int transparency_factor;
+    int transparency;
+    ezgl::color color;
```
AlexandreSinger: I would move these variables down to where they are used.

vpr/src/draw/search_bar.cpp
```diff
@@ -207,51 +207,51 @@ bool highlight_rr_nodes(RRNodeId hit_node) {
...
-            VTR_LOG("%s\n", describe_rr_node(device_ctx.rr_graph, device_ctx.grid, device_ctx.rr_indexed_data, node, draw_state->is_flat).c_str());
-        }
+    if (!hit_node) {
+        application.update_message(draw_state->default_message);
+        rr_highlight_message = "";
+        application.refresh_drawing();
+        return false;
+    }
+
+    const auto& device_ctx = g_vpr_ctx.device();
```
AlexandreSinger: Can you please write out the type instead of using auto here.

vpr/src/draw/search_bar.cpp
```diff
@@ -207,51 +207,51 @@ bool highlight_rr_nodes(RRNodeId hit_node) {
...
 
-        //Show info about *only* hit node to graphics
-        std::string info = describe_rr_node(device_ctx.rr_graph, device_ctx.grid, device_ctx.rr_indexed_data, hit_node, draw_state->is_flat);
+    // Highlight neighboring non_configurable nodes in magenta as well
+    auto nodes = draw_expand_non_configurable_rr_nodes(hit_node);
```
AlexandreSinger: Same comment here. If nodes is a simple type, it can be helpful to have the full type here.

vpr/src/draw/search_bar.cpp
```diff
@@ -207,51 +207,51 @@ bool highlight_rr_nodes(RRNodeId hit_node) {
...
 
-        sprintf(message, "Selected %s", info.c_str());
-        rr_highlight_message = message;
+    for (auto node : nodes) {
```
AlexandreSinger: Same comment here.

vpr/src/draw/search_bar.cpp
```diff
@@ -207,51 +207,51 @@ bool highlight_rr_nodes(RRNodeId hit_node) {
...
+            draw_state->draw_rr_node[node].node_highlighted = false;
         }
-    } else {
-        application.update_message(draw_state->default_message);
-        rr_highlight_message = "";
-        application.refresh_drawing();
-        return false; //No hit
+
+        //Print info about all nodes to terminal including neighboring non-configurable nodes
+        VTR_LOG("%s\n", describe_rr_node(device_ctx.rr_graph, device_ctx.grid, device_ctx.rr_indexed_data, node, draw_state->is_flat).c_str());
```
AlexandreSinger, line 234: Just double checking, does this produce a lot of log messages? Or does it produce a reasonable amount?

AlexandreSinger (commented):
Left some minor coding comments. Otherwise looks good.

vpr/src/draw/draw_basic.cpp
SamuelHo10: The first statement checks for if we've enabled draw INTER cluster nets, and the second if statement checks for whether draw INTRA cluster nets is enabled. I've added comments to make it more clear.

vpr/src/draw/search_bar.cpp
SamuelHo10: It prints out a reasonable amount as long as there are not too many neighboring non-configurable RR nodes I believe.

SamuelHo10:
Thanks for the code review Alex!

vpr/src/draw/draw.cpp
```diff
@@ -189,28 +189,21 @@ static void draw_main_canvas(ezgl::renderer* g) {
...
-        switch (draw_state->show_nets) {
-            case DRAW_CLUSTER_NETS:
-                drawnets(g);
-                break;
-            case DRAW_PRIMITIVE_NETS:
-                break;
-            default:
-                break;
+        if (draw_state->draw_nets == DRAW_FLYLINES && draw_state->show_nets) {
+            drawnets(g);
```
vaughnbetz: Will this let you draw flylines after routing?  It seems it would only enable flylines after placement, but it would be better to allow the user to select flylines after routing too, in case the user wanted to go between the routed and flyline views.
It seems like the flyline drawing checks should be outside of the pic_on_screen if.

vpr/src/draw/draw_toggle_functions.cpp
```diff
@@ -18,70 +18,80 @@
...
+    gchar* setting = gtk_combo_box_text_get_active_text(GTK_COMBO_BOX_TEXT(self));
+
     // assign corresponding enum value to draw_state->show_nets
-    if (strcmp(setting, "None") == 0)
-        new_state = DRAW_NO_NETS;
-    else if (strcmp(setting, "Cluster Nets") == 0) {
-        new_state = DRAW_CLUSTER_NETS;
-    } else { // Primitive Nets - Used to be called "Logical Connections"
-        new_state = DRAW_PRIMITIVE_NETS;
+    if (strcmp(setting, "Routing") == 0) {
```
vaughnbetz, line 55: Does something elsewhere prevent selecting the 'Routing' option if we've only completed PLACEMENT (and hence no routing exists)?

vpr/src/draw/draw_toggle_functions.h
```diff
@@ -20,9 +20,27 @@
 
 #include "ezgl/application.hpp"
 
-/* Callback function for main.ui created toggle_nets button in ui_setup.cpp. Controls whether or not nets are visualized.
- * Toggles value of draw_state->show_nets.*/
-void toggle_nets_cbk(GtkComboBox* self, ezgl::application* app);
+/**
+ * @brief Generalized callback function for checkboxes.
+ * Toggles the state of a boolean variable based on the checkbox state.
+ */
```
vaughnbetz, line 28: Good to document the @param too

vpr/src/draw/draw_types.h
```diff
@@ -41,21 +42,8 @@ enum e_draw_crit_path {
 
 enum e_draw_nets {
```
vaughnbetz, line 44: Comment this enum and all the values in it; this is an important one.  The comment can briefly explain what a flyline is (straight line from src to sink)

vpr/src/draw/draw_types.h
```diff
@@ -134,15 +122,41 @@ enum e_draw_noc {
     DRAW_NOC_LINK_USAGE
 };
 
+/// Different types of edges in the routing resource graph
```
vaughnbetz: Explain what type the edges within a cluster are (PIN_TO_OPIN and PIN_TO_IPIN?)

vpr/src/draw/draw_types.h
```diff
@@ -134,15 +122,41 @@ enum e_draw_noc {
...
+    CHAN_TO_CHAN,
+    NUM_EDGE_TYPES
+};
+
+/**
+ * @brief Structure used to hold data passed into the toggle checkbox callback function.
+ */
+typedef struct {
+    ezgl::application* app;
+    bool* toggle_state;
```
vaughnbetz: I think this could just be a bool, not a bool* (with suitable code modifications).  gtk needs a pointer to a struct, but inside the struct you can use pointers or values as you like.  I think a straight bool would be cleaner here.

vaughnbetz: Good to explicitly comment the member variables, even though in this case the reader can probably guess what they are for.

vpr/src/draw/draw_types.h
```diff
@@ -134,15 +122,41 @@ enum e_draw_noc {
...
+} t_checkbox_data;
+
 /* Structure which stores state information of a rr_node. Used
  * for controling the drawing each rr_node when ROUTING is on screen.
  * color: Color of the rr_node
  * node_highlighted: Whether the node is highlighted. Useful for
  *					 highlighting routing resources on rr_graph
  */
+/**
```
vaughnbetz, line 143: Repeated or nearly repeated comment. Delete the old one.

vpr/src/draw/draw_types.h
```diff
@@ -134,15 +122,41 @@ enum e_draw_noc {
...
+ * @brief Structure used to store the state information of an rr_node. 
+ * Used to control drawing each rr_node when ROUTING is on screen.
+ * color: Color of the rr_node
+ * node_highlighted: Whether the node is highlighted. Useful for highlighting routing resources on rr_graph.
+ * node_hit: Whether the node is hit by a mouse click. Required since multiple RR nodes can be clicked on at the same time.
```
vaughnbetz: Do we need node_hit?  Wouldn't the node_highlighted field be enough?
If there's a good reason, add a longer comment to explain it.

vaughnbetz (commented):
Looks good; I just have a few things I suggest documenting and/or checking.
As we discussed in the meeting, things to check / update:
- Should be able to select flylines or routing after the routing is complete; only flylines before that.
- Not sure if we need a hit and a highlighted value for each rr_node; should explain why both are needed in the comments more, or try to get rid of the 'hit' one to save memory.

vpr/src/draw/draw.cpp
SamuelHo10: I think the original function names are a bit confusing. draw_nets is a function which only draws flylines during the placement stage. I've renamed the function to draw_flylines_placement.

vpr/src/draw/draw_types.h
SamuelHo10: I used a pointer since I need to modify a variable in draw_state.

vpr/src/draw/draw_types.h
SamuelHo10: I've changed it to a set to save memory.

SamuelHo10:
> It seems like (in another issue) we could do the same vec-> change on the .highlighted data member. What do you think?

There are usually a lot more highlighted nodes than hit nodes, and I'm worried it might impact performance since searching through a set is more costly than looking it up in an array.

Issue #3242, opened 2025-08-21
There might be three coding problems


victroix:
1.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/246053dfea69af32e7ff0fdbe3c9b702b38a7bf1/vpr/src/place/delay_model/compute_delta_delays_utils.cpp#L194
    const size_t high_y = (longest_length <= device_height) ? std::max(device_width - longest_length, mid_y) : mid_y;
Should be: const size_t high_y = (longest_length <= device_height) ? std::max(device_height - longest_length, mid_y) : mid_y;
2.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/246053dfea69af32e7ff0fdbe3c9b702b38a7bf1/libs/libvtrutil/src/vtr_math.h#L42
    auto iter = first + len / 2;
Might add after this line: if(len == 2)  {iter = first;}
at least i add these when generating delay table with reducer::median. It might be due to the fact that I'm using VS2015 for compilation.
 i tested if reducing two delays by median, for example, [0.8, 0.9], delay was 0.45, not 0.85.
3.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/246053dfea69af32e7ff0fdbe3c9b702b38a7bf1/vpr/src/place/move_generators/simpleRL_move_generator.cpp#L234
        last_action_ = action_type_q_pos;
might be:         last_action_ = std::min((size_t)action_type_q_pos, num_available_actions - 1); ?

AlexandreSinger:
Hi @victroix thanks for the issue!

(1) certainly looks like a bug for me, @amin1377 what do you think?

(2) I do not think this issue was ever actually being hit. Due to some confusing code in vtr::math a different median function was being used. I have raised a PR ( #3243 ) which resolves this issue anyways since this method was definately broken.

(3) I am not sure what is the issue with this line, can you please elaborate what you think is wrong?

AlexandreSinger:
All issues have been resolved. Closing issue.

AlexandreSinger:
@victroix Thanks again for pointing these issues out!

victroix:
@AlexandreSinger @vaughnbetz Thank you for your reply. It is my honor to communicate with seniors from the VTR project.

PR #3247, opened 2025-08-21
[Place][MoveGenerator] Fixed Rare Issue

AlexandreSinger:
Found that it may be possible for a random move to not be selected due to numerical precision issues. Added a check to handle this.

See issue #3242

### DIFF
--- a/vpr/src/place/move_generators/simpleRL_move_generator.cpp
+++ b/vpr/src/place/move_generators/simpleRL_move_generator.cpp
@@ -251,7 +251,7 @@ t_propose_action EpsilonGreedyAgent::propose_action() {
             // chosen. In this case, just choose the last action.
             // For example, for three actions:
             //      [0.33, 0.66, 0.99]
-            //   Notice that the last ation does not perfectly add up to 1.0. To
+            //   Notice that the last action does not perfectly add up to 1.0. To
             //   get around this, we just pretend that it rounded up to 1.0.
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
Another solution to this is that we just directly set the last element to 1.0 for the CDF (since it will always be 1.0); however, we will always want a check like this in the code for safety.

vpr/src/place/move_generators/simpleRL_move_generator.cpp
```diff
@@ -227,9 +227,34 @@ t_propose_action EpsilonGreedyAgent::propose_action() {
...
+        if (itr != cumm_epsilon_action_prob_.end()) {
+            action_type_q_pos = itr - cumm_epsilon_action_prob_.begin();
+        } else {
+            // Due to numerical precision (and dumb luck) its possible that the
+            // CDF does not fully add up to 1.0 (for example 0.9999) and the
+            // random number chosen is 1.0. In this case, no action will be
+            // chosen. In this case, just choose the last action.
+            // For example, for three actions:
+            //      [0.33, 0.66, 0.99]
+            //   Notice that the last ation does not perfectly add up to 1.0. To
```
vaughnb-cerebras: ation -> action

vaughnb-cerebras (approved):
Looks good. One typo to fix; good to merge once CI finishes.

PR #3251, opened 2025-08-23
[Logging][Warnings] Suppressed Overly Verbose Warnings

AlexandreSinger:
There were many occurrences of warning messages that were either overly verbose or unnecessary.

In the RR graph reading code, there were several warnings that were generated when nodes could not be found for a cost index. This warning is good to keep around I think; however we do not need to print the warning for each cost index.

In check netlist, we were printing a warning when the netlist does not contain any global nets to non-global pins, which is a strange thing to warn about.

In the annealer, we have been warning whenever the initial temperature was being calculated. This warning was saying that some of the swaps attempted was rejected; but I do not think this needs to be warned about.

In the router, there were two warnings which are not of any use to a user, but for developers they may be useful. I have raised the verbosity level of these warnings. This required that I update the router code to pass around the route_verbosity variable.

Found similar warnings in the router lookahead, which I suppressed to higher route verbosity levels.

### DIFF
--- a/vpr/src/base/check_netlist.cpp
+++ b/vpr/src/base/check_netlist.cpp
@@ -63,7 +63,7 @@ void check_netlist(int verbosity) {
     }
     free_hash_table(net_hash_table);
     if (global_to_non_global_connection_count > 0) {
-        VTR_LOG_WARN("Netlist contains %d global net to non-global architecture pin connections\n", global_to_non_global_connection_count);
+        VTR_LOG("Netlist contains %d global net to non-global architecture pin connections\n", global_to_non_global_connection_count);
     }
 
     auto& device_ctx = g_vpr_ctx.device();
--- a/vpr/src/route/connection_router.tpp
+++ b/vpr/src/route/connection_router.tpp
@@ -83,7 +83,7 @@ std::tuple<bool, bool, RTExploredNode> ConnectionRouter<Heap>::timing_driven_rou
     if (std::isinf(rr_node_route_inf_[sink_node].path_cost)) {
         //Found no path, that may be due to an unlucky choice of existing route tree sub-set,
         //try again with the full route tree to be sure this is not an artifact of high-fanout routing
-        VTR_LOGV_WARN(route_verbosity_ > 1, "No routing path found in high-fanout mode for net %zu connection (to sink_rr %d), retrying with full route tree\n", size_t(conn_params.net_id_), sink_node);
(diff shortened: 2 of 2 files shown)


vpr/src/base/check_netlist.cpp
```diff
@@ -62,7 +62,9 @@ void check_netlist(int verbosity) {
         }
     }
     free_hash_table(net_hash_table);
-    VTR_LOG_WARN("Netlist contains %d global net to non-global architecture pin connections\n", global_to_non_global_connection_count);
+    if (global_to_non_global_connection_count > 0) {
+        VTR_LOG_WARN("Netlist contains %d global net to non-global architecture pin connections\n", global_to_non_global_connection_count);
```
vaughnbetz: Probably this should just be an info message; not really sure why we made this a warning.

vpr/src/route/connection_router.tpp
```diff
@@ -83,7 +83,7 @@ std::tuple<bool, bool, RTExploredNode> ConnectionRouter<Heap>::timing_driven_rou
     if (std::isinf(rr_node_route_inf_[sink_node].path_cost)) {
         //Found no path, that may be due to an unlucky choice of existing route tree sub-set,
         //try again with the full route tree to be sure this is not an artifact of high-fanout routing
-        VTR_LOG_WARN("No routing path found in high-fanout mode for net %zu connection (to sink_rr %d), retrying with full route tree\n", size_t(conn_params.net_id_), sink_node);
+        VTR_LOGV_WARN(route_verbosity_ > 1, "No routing path found in high-fanout mode for net %zu connection (to sink_rr %d), retrying with full route tree\n", size_t(conn_params.net_id_), sink_node);
```
vaughnbetz: I think this should be an info message; it's not an unusual result.

vpr/src/route/connection_router.tpp
```diff
@@ -154,7 +154,7 @@ bool ConnectionRouter<Heap>::timing_driven_route_connection_common_setup(
         }
 
         // Otherwise, leave unrouted and bubble up a signal to retry this net with a full-device bounding box
-        VTR_LOG_WARN("No routing path for connection to sink_rr %d, leaving unrouted to retry later\n", sink_node);
+        VTR_LOGV_WARN(route_verbosity_ > 1, "No routing path for connection to sink_rr %d, leaving unrouted to retry later\n", sink_node);
```
vaughnbetz: I think this should be an info message; it's not an unusual result.

PR #3252, opened 2025-08-23
[CI] Moved to Use Standard Package Installer Script

AlexandreSinger:
Found that the CI was using a separate script which was indirectly calling the standard package install script. This can cause the standard package install script to regress, but it was also causing pressure on the GitHub runners by having them install more packages than they needed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,6 +93,7 @@ jobs:
     - name: Install dependencies
       run: |
         ./install_apt_packages.sh
+        sudo apt-get update
         sudo apt-get install qtbase5-dev
 
     - uses: hendrikmuhs/ccache-action@v1.2


install_apt_packages.sh
```diff
@@ -1,53 +1,63 @@
...
 # Base packages to compile and run basic regression tests
-sudo apt-get install -y \
-    make \
-    cmake \
-    build-essential \
-    pkg-config \
-    bison \
-    flex \
-    python3-dev \
+packages_to_install=(
```
AmirhosseinPoolad, line 4: This is pretty cool but I feel like the bash syntax might be unfamiliar to many people. Is it beneficial to the CI to use this syntax vs the old syntax?

.github/workflows/test.yml
```diff
@@ -87,8 +87,13 @@ jobs:
...
       id: cpu-cores
 
+    # NOTE: BlifExplorer requires Qt5 to build. We do not currently actually
+    #       run the executable as part of the CI, but we at least want to test
+    #       that it can be compiled.
     - name: Install dependencies
-      run: ./.github/scripts/install_dependencies.sh
+      run: |
+        ./install_apt_packages.sh
+        sudo apt-get install qtbase5-dev
```
AmirhosseinPoolad, line 97: Should we do a sudo apt update here to be safe?

AmirhosseinPoolad (commented):
Thanks for the PR Alex. The changes sound very good to me. Had a couple of thoughts about some not super important things. Looks good otherwise.

install_apt_packages.sh
AlexandreSinger: Yes, I think. The reason I made this PR is because over the last week the CI has been having issues installing packages. For some runs installing packages takes hours. I think what is going on is that we are calling "apt-get install" many times for each job and we are getting throttled by GitHub somehow.

For this script, there is no reason to call apt-get that many times; instead we just create a bash array.

AmirhosseinPoolad: Makes sense!

.github/workflows/test.yml
AlexandreSinger: Not a bad idea! Technically we dont since install_apt_packages does as well, but we might as well.

PR #3255, opened 2025-08-26
Fix regression testing false negative

### LINKED ISSUE #3254: [CI] Strong tests not failing if there's a mismatch in architecture in golden results vs test results
The architecture name for strong_scatter_gather golden results used to be different than the actual architecture file (Changed to arch file name without regenerating golden results) and this passed CI without while actually printing out an error.

The piece of code responsible for it is in parse_vtr_task.py:428. The issue (I think) is that the run_vtr_task file actually returns with a zero exit code:

Image

(Blue circle next to the command means the return code of the command was zero)

How to reproduce:

change the architecture name in 'vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_scatter_gather/config/golden_results.txt' to anything else, run './run_vtr_task.py regression_tests/vtr_reg_strong/strong_scatter_gather' then './run_vtr_task.py regression_tests/vtr_reg_strong/strong_scatter_gather -check_golden'

AmirhosseinPoolad:
Fixes #3254 . See this action run:

https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/17196322572/job/48779145751

The test has passed the CI but there are actually errors:
Image

Why did it pass? Well there this piece of code:
Image

But that's not the root cause. The QoR parsing script returned -1 in case of a mismatch between arch file name in the actual run and the golden results. In case of  1 run failure, the -1 would cancel the 1 and as a result the regression tests would pass while having errors. This commit fixes that by not throwing an exception in that specific case (Therefore not returning -1) and taking the absolute value of the QoR and run failures before summing them up.

### DIFF
--- a/run_reg_test.py
+++ b/run_reg_test.py
@@ -199,8 +199,10 @@ def vtr_command_main(arg_list, prog=None):
         print("All tests passed")
     elif tests_run and total_num_func_failures != 0 or total_num_qor_failures != 0:
         print("Error: {} tests failed".format(total_num_func_failures + total_num_qor_failures))
-
-    sys.exit(total_num_func_failures + total_num_qor_failures)
+    
+    # If the QoR parsing script throws an exception, it returns -1. This could potentially cancel a run failure and result in a false negative.
+    # Absolute value is taken to avoid that.
+    sys.exit(abs(total_num_func_failures) + abs(total_num_qor_failures))
 
 
 def display_qor(reg_test):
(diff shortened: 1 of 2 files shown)


AmirhosseinPoolad:
Just to add, the only reason all the strong tests were actually ran in the linked CI run was that the scatter gather tests were at the end of the task list. If you moved it to the beginning, the script would crash at the start, no other tasks would be run, and CI would've still passed.

AlexandreSinger (approved):
LGTM

Issue #2266, opened 2023-03-30
Wrong blif for subtractor for different architecture files


FPGA1234:
We have tried to execute the following RTL for subtracting two 4-bit numbers using the architecture file which supports carry chain adder. But the blif generated by the tools is giving the errors.

RTL code is shown below -
![image](https://user-images.githubusercontent.com/129363702/228753220-cfa99436-d893-4a8d-95e0-d28a92dec951.png)

The tools is generating a blif file with the following logic -
![image](https://user-images.githubusercontent.com/129363702/228753321-0df8dd7c-f85e-4a58-b51d-4eca06e00273.png)

The behavior of the tool output vs expected behavior is elaborated is shown below -
![image](https://user-images.githubusercontent.com/129363702/228753397-8c8fb6a7-3c10-412c-bdcc-33ac96657090.png)

In the image, one could see that subtractor implemented by tool has the adder (highlighted in red circle) with inputs as GND and VCC, CIN as GND, thus giving COUT as GND.
This results in computation of 2's complement as 1's complement instead of (1's complement+1) as 2's complement = (1's complement + 1).
Also, one could see in the blif that the sign bit of output out, i.e., out[4] is assigned value of 0.

We have analyzed and found the working solution as shown in the diagram on the right.
The changes are highlighted in green and blue circle.
Now, we have adder (highlighted in green circle) with inputs as VCC and VCC, CIN as GND, thus giving COUT as VCC to make 2's complement = (1's complement + 1).
Also, we have taken an additional adder (highlighted in blue circle) with inputs as GND and VCC, CIN as COUT of previous adder. Thus, generating correct sign bit.

The architecture file is added below -
[k6_frac_N10_adder_chain_40nm.txt](https://github.com/verilog-to-routing/vtr-verilog-to-routing/files/11108686/k6_frac_N10_adder_chain_40nm.txt)

One can reproduce the error by giving by calling vtr-verilog-to-routing/vtr_flow/scripts/run_vtr_flow.py using the architecture file, RTL provided above.

alirezazd:
@vaughnbetz Please assign me to this issue.

vaughnbetz:
Sending to Logan to see if he can test if this is still an issue with yosys+parmys.

vaughnbetz:
@loglav03

loglav03:
[...]

PR #3261, opened 2025-08-29
[WIP] Fixing Subtraction CIN Bug

loglav03:
#### Description

When doing a 4-bit input a and b subtraction to get a 5 bit output, we must compute as 2's complement = (1's complement + 1)
See as a + (1's comp. of b) + 1 = a + (2's comp. of b).

The subtraction carry chain was originally computing 2's complement as 1's complement (a + (1's comp. of b)). The reason for this is that when the original wide hard adder was split into a 1-bit adder chain, the head adder of the chain was supposed to seed the rest of the chain with a carry-in of 1. Instead it computed GND (0) + VCC (1) + GND (0), which gives a cout of 0, seeding the rest of the chain with a carry-in of 0, resulting in a + (1's comp. of b). **This was fixed to compute VCC + VCC + GND, which seeds a carry-in of 1 to the rest of the chain.**

There is also an additional adder added to the tail end of the chain. This adder takes in GND and VCC and CIN as inputs and is supposed to output the correct sign bit out[4]. **This is currently broken** in the code as I was having trouble connecting the correct output back to the original wide hard adder.

See the issue below for a diagram of the bug and further explanation.

#### Related Issue

https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2266

No in depth testing done yet as the code still needs some fixing up.

#### Types of changes

#### Checklist:

### DIFF
--- a/parmys/parmys-plugin/parmys.cc
+++ b/parmys/parmys-plugin/parmys.cc
@@ -817,17 +817,6 @@ struct ParMYSPass : public Pass {
         log("\nTechmap Time: ");
         log_time(techmap_time);
         log("\n--------------------------------------------------------------------\n");
-
-        fprintf(stderr, "[BLIF-OUT] POs=%d\n", odin_netlist->num_top_output_nodes);
-        for (int k=0; k<odin_netlist->num_top_output_nodes; ++k) {
-            npin_t* in = odin_netlist->top_output_nodes[k]->input_pins[0];
-            fprintf(stderr, "  PO[%d]: net=%s driver=%s.pin%d\n",
-                k,
-                in->net && in->net->name ? in->net->name : "(noname)",
-                (in->net && in->net->driver_pins[0]->node &&
-                in->net->driver_pins[0]->node->name) ? in->net->driver_pins[0]->node->name : "(null)",
-                in->net ? in->net->driver_pins[0]->pin_node_idx : -1);
-        }
     }
 
     static void report(netlist_t *odin_netlist)
--- a/vtr_flow/misc/yosys/synthesis.tcl
+++ b/vtr_flow/misc/yosys/synthesis.tcl
@@ -80,15 +80,11 @@ techmap -map +/parmys/aldffe2dff.v
 
 opt -full
 
-write_verilog -noexpr pre_parmys.v
-
 # Separate options for Parmys execution (Verilog or SystemVerilog)
 if {$env(PARSER) == "default" || $env(PARSER) == "slang"} {
     # For Verilog, use -nopass for a simpler, faster flow
     parmys -a QQQ -nopass -c CCC YYY
 } 
-
-write_verilog -noexpr post_parmys.v
 
 opt -full
 
@@ -102,7 +98,5 @@ opt -fast -noff
 stat
 
 hierarchy -check -auto-top -purge_lib
-
-write_verilog -noexpr post_everything.v
 
 write_blif -true + vcc -false + gnd -undef + unconn -blackbox ZZZ


loglav03:
@vaughnbetz - Opened this PR for the next person to take over the issue.

vaughnbetz:
Thanks @loglav03 . @KennethKent : anyone to take on this bug fix?

loglav03:
Currently, the newly updated code for splitting an adder for subtraction matches the structure depicted in the solution diagram provided by the related issue: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2266

However, even though the adder chain structure matches the provided solution, it is still failing vpr. I found that the old code was passing through vpr even though the subtraction adder chain was buggy. So, I investigated this to see what was causing it to pass, hoping it would give insight into what is causing the new code to fail.

I found that in the parmys elaboration pass, it walks the netlist and calls the function ```equalize_ports_size``` for nodes with the ```MINUS``` type. In the case of the provided test verilog, the original wide hard adder node had a port 'a' and 'b' size of 4 each and an out port size of 5. In this function, the original code had a line ```int new_out_size = port_a_size;``` which would create a new node, remap the input pins of the original node to the new node, and only remap the first 4 output pins to the new node and exclude the 5th. **(See function starting at line 1365 in vtr-root/parmys/parmys-plugin/netlist/netlist_utils.cc)**

At the end of this function, it has the conditional block shown below, which would remap the 5th output pin to a buffer node whose input pin is fed by the PAD node, which essentially makes the last output pin a constant 0. 
```
if (new_out_size < port_y_size) {
        for (int i = new_out_size; i < port_y_size; i++) {
            /* need to drive extra output pins with PAD */
            nnode_t *buf_node = make_1port_gate(BUF_NODE, 1, 1, node, traverse_mark_number);
            /* hook a pin from PAD node into the buf node */
            add_input_pin_to_node(buf_node, get_pad_pin(netlist), 0);
            /* remap the extra output pin to buf node */
            remap_pin_to_new_node(node->output_pins[i], buf_node, 0);
        }
 }
```
The issue I'm seeing with the new code is that the output .names seem to be getting dropped from the pre-vpr.blif, which trips up the vpr stage because it can't see any primary outputs for the subtraction adder chain. The output .names appear in the parmys.blif so it looks like abc ends up dropping them before pre-vpr.blif is created.

So, it seems that every primary output being driven by the sumout pin of each adder slice gets dropped at some point after yosys+parmys and before producing the pre-vpr.blif (likely in abc). I believe this is being caused by the adder chain being blackboxed so abc won't touch it (see .blackbox attribute in .blif files) and results in abc not being able to see the primary outputs since they're driven by the adder chain and drops them.

The old code was able to produce a single primary output seen by abc and pass vpr because the final output pin of the original wide adder was remapped to a buffer node driven by the pad node, which later in the parmys techmap pass, the buffer node gets eliminated and the nets between the pad node-to-buf node and the buf node-to-primary output gets joined together **(see function starting at line 545 in vtr-root/parmys/parmys-plugin/mapping/partial_map.cc)**, so the PO for the last output pin is driven by the pad node and not the adder chain itself. (See difference in .blif outputs below for old vs. new code for outputs .names)

Old parmys.blif:
[temp.parmys.txt](https://github.com/user-attachments/files/22914135/temp.parmys.txt)

Old pre-vpr.blif:
[temp.pre-vpr.txt](https://github.com/user-attachments/files/22914143/temp.pre-vpr.txt)

_**Note**: In the new code out files, the ordering is out of place for the adder sub circuits. The final adder sub circuit appears first in the sub circuit list. This doesn't affect anything, it's just the way they happened to be ordered._

New parmys.blif:
[temp.parmys.txt](https://github.com/user-attachments/files/22914150/temp.parmys.txt)

New pre-vpr.blif:
[temp.pre-vpr.txt](https://github.com/user-attachments/files/22914167/temp.pre-vpr.txt)

vaughnbetz:
I'm a bit out of my depth in parmys operation here ... @KennethKent any ideas?

alirezazd:
Closing this PR. Fix in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3493

PR #3265, opened 2025-09-05
[AP][Solver][3D] 3D AP Solver Support

AlexandreSinger:
Upgraded the solver in the AP flow to support a "Z" dimension representing the layer. This code only kicks on when the architecture has more than one layer, since in the single layer case there is no point to compute the Z dimension.

## Release Notes

* **New Features**
  * Added support for multi-die 3D placement optimization in the analytical solver, enabling placements across multiple device layers.

* **Tests**
  * Introduced new test configurations to validate 3D placement with IO constraints.
  * Updated regression test baselines to reflect current solver performance metrics.


vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -1125,12 +1204,26 @@ void B2BSolver::init_linear_system(PartialPlacement& p_placement, unsigned itera
                 add_connection_to_system(blk_id, net_bounds.max_y_blk, num_pins, wl_net_w, p_placement.block_y_locs, triplet_list_y, b_y);
                 add_connection_to_system(blk_id, net_bounds.min_y_blk, num_pins, wl_net_w, p_placement.block_y_locs, triplet_list_y, b_y);
             }
+            if (is_multi_die() && blk_id != net_bounds.max_z_blk && blk_id != net_bounds.min_z_blk) {
+                // For multi-die FPGAs, we apply extra weight in the layer dimension
+                // since moving between layers tends to cost more wiring than moving
+                // within the same layer.
+                double multidie_net_w = wl_net_w * layer_distance_cost_fac_;
```
vaughnbetz: How is this set?  Ideally we'd set these as 1/wiring supply for each dimension. Note that x and y should have different factors too when the wiring supply in each dimension is different (e.g. S10, 7-series)

vpr/src/analytical_place/analytical_solver.h
```diff
@@ -531,6 +531,16 @@ class B2BSolver : public AnalyticalSolver {
     ///        number, the solver will focus more on timing and less on wirelength.
     static constexpr double timing_slope_fac_ = 0.75;
 
+    /// @brief For most FPGA architectures, the cost of moving horizontally is
+    ///        equivalent to the cost moving vertically (i.e. moving in increasing
+    ///        x-dimension has the same cost as moving the same amount in the
```
vaughnbetz: Not true for S10 and 7-series (different wiring supplies).  Also may not be true for timing in many architectures.

vaughnbetz: Should be able to get these cost factors from some Soheil/Amin code in the annealer (they compute prefix sums of wiring supply in each dimension; should be able to ask for total wiring in x or y or z from them.

vpr/src/analytical_place/analytical_solver.h
```diff
@@ -720,22 +756,28 @@ class B2BSolver : public AnalyticalSolver {
...
     // for the next call to solve.
     vtr::vector<APBlockId, double> block_x_locs_solved;
     vtr::vector<APBlockId, double> block_y_locs_solved;
+    vtr::vector<APBlockId, double> block_z_locs_solved;
 
     // The following are the legalized solution coming into the analytical solver
     // (other than the first iteration). These are stored to be used as anchor
     // blocks during the solver.
     vtr::vector<APBlockId, double> block_x_locs_legalized;
     vtr::vector<APBlockId, double> block_y_locs_legalized;
```
vaughnbetz, line 770: Probably want to say that for speed, the various _z values are ignored / not updated or used if a device has only one die.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/basic_3d_ap/constraints/mm9a_io_constraint.xml
```diff
@@ -0,0 +1,70 @@
...
+            <add_atom name_pattern="out:p_44" />
+            <add_atom name_pattern="out:p_45" />
+        </partition>
+        <partition name="p_46-47">
+            <add_region subtile="-1" layer_low="1" layer_high="1" x_low="9" x_high="9" y_low="6" y_high="6" />
+            <add_atom name_pattern="out:p_44" />
+            <add_atom name_pattern="out:p_45" />
+        </partition>
+        <partition name="p_48">
+            <add_region subtile="-1" layer_low="1" layer_high="1" x_low="9" x_high="9" y_low="8" y_high="8" />
```
vaughnbetz, line 65: Do you want all the IOs on layer 1?  Or would it be a stronger test if there were some on each layer?

vaughnbetz (approved):
LGTM, but a few comments embedded. 
Should also add a before and after runtime etc. comparison on the 2D case.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/basic_3d_ap/constraints/mm9a_io_constraint.xml
AlexandreSinger: For this specific example, the IOs are only located on layer 1 and the CLBs are located on layer 0. This is a simple test architecture to just ensure that different blocks are placed on different layers.

vaughnbetz:
I think this one is almost good to go ... some comments to address, and long term at least the cost scaling factors should come from rr_graph computations / refactored annealing code that gets the scaling factors based on routing supply.

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: You are correct. @amin1377 pointed me to how the placer does this and they have cost multipliers in each dimension:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/9ebcd5c3a8aa2a0896973899b10293fa203c207d/vpr/src/place/net_cost_handler.h#L547-L566

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/9ebcd5c3a8aa2a0896973899b10293fa203c207d/vpr/src/place/net_cost_handler.cpp#L1445-L1460

One issue that I am running into is that these cost handlers are inside the of the NetCostHandler object which requires the placer state to use. I want to avoid copying the code, so I may investigate cleaning up this interface to expose a simpler way for me to use this in AP as well.

AlexandreSinger: However, this code may be simple enough that it can be copied. I will investigate.

PR #3266, opened 2025-09-08
Interposer tag

AmirhosseinPoolad:
This PR contains an initial proposal for a new interposer tag. To define an interposer cut, you would add an <interposer_cut dim=x|y loc=val> tag under a `<layout>` tag that defines the location of the cut and the axis of the cut. This tag would then have multiple <interdie_wire> children specifying the connections between the two sides of the cut-line.

> <interdie_wire sg_name="interposer_conn_sg" sg_link="X4" offset_start="0" offset_end="2" offset_increment="1" num=50/>

- sg_name: name of the scatter-gather pattern to be used for the connections
- sg_link: name of the sg_link to be used for the connections
- offset_start,end,increment: Connections can be laid out in various offsets from the cut-line, these options
- num: number of scatter-gather patterns to be instantiated at each switchblock location

Example:

image

Note that in this example we have 3 different switchblock locations and 3*50=150 sg patterns would be instantiated.

### DIFF
--- a/doc/src/api/libarchfpga/physical_types.rst
+++ b/doc/src/api/libarchfpga/physical_types.rst
@@ -2,7 +2,7 @@
 Physical Types
 ========
 
-These types are used to capture user's intended architecture specification and will be later turned into a flattened device grid according to the device's size.
+These types are used to capture user's intended architecture specification.
 
 .. doxygenstruct:: t_arch
    :project: vpr
--- a/doc/src/api/libarchfpga/scatter_gather_types.rst
+++ b/doc/src/api/libarchfpga/scatter_gather_types.rst
@@ -2,7 +2,7 @@
 Scatter-Gather Types
 ========
 
-These types store information about about scatter-gather patterns
+These types store information about about scatter-gather routing patterns
 
 .. doxygenstruct:: t_scatter_gather_pattern
    :project: vpr
--- a/doc/src/api/libarchfpga/grid_types.rst
+++ b/doc/src/api/libarchfpga/grid_types.rst
@@ -2,7 +2,7 @@
 Grid Specification Types
 ========
 
-These types are used to capture user's intended architecture specification and will be later turned into a flattened device grid according to the device's size.
+These types are used to capture user's intended grid specification, i.e., which tiles go where in the device. These specifications will be later turned into a flattened device grid according to the device's size.
 
 .. doxygenstruct:: t_grid_loc_spec
    :project: vpr


libs/libarchfpga/src/interposer_types.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <unordered_map>
+#include <vector>
+#include <string>
+#include "vtr_vector.h"
+
+enum class e_interposer_cut_dim {
```
vaughnbetz, line 19: Doxygen

libs/libarchfpga/src/interposer_types.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <unordered_map>
```
vaughnbetz, line 10: Doxygen at top of file ... what is this file for?

libs/libarchfpga/src/interposer_types.h
```diff
@@ -0,0 +1,34 @@
...
+#include <vector>
+#include <string>
+#include "vtr_vector.h"
+
+enum class e_interposer_cut_dim {
+    X,
+    Y
+};
+
+inline const std::unordered_map<char, e_interposer_cut_dim> CHAR_INTERPOSER_DIM_MAP = {
```
vaughnbetz, line 25: Doxygen comment for each struct.

AmirhosseinPoolad:
I think the code is ready for review now! @vaughnbetz @soheilshahrouz

doc/src/arch/reference.rst
```diff
@@ -581,6 +581,32 @@ Grid Layout Example
 
     Example FPGA grid
 
+
+.. arch:tag:: <interposer_cut dim=x|y loc="int"/>
+
+    :req_param dim: Dimension or axis of the cut. 'X' or 'x' means a horizontal cut while 'Y' or 'y' means a vertical cut.
+    :req_param loc: Location of the cut. Currently only absolute positions are supported.
```
vaughnbetz: Can this be an equation, or does it have to be a number?  Should say ...

vaughnbetz: Or maybe that's what absolute positions means?

doc/src/arch/reference.rst
```diff
@@ -581,6 +581,32 @@ Grid Layout Example
...
+    :req_param offset_increment: Increment/distance between scatter-gather instantiations.
+    :req_param num: Number of scatter-gather instantiations per switchblock location.
+
+    Defines the interdie wiring between the two sides of the cut. Connectivity is defined using scatter-gather patterns. Starting at 'offset_start' from location of the cut and moving by 'offset_increment' until we reach the location of 'offset_end' away from the cut, 'num' scatter-gather patterns defined by 'sg_name' and 'sg_link' will be instantiated.
+    Note that these offset points always define the starting point of the scatter-gather pattern's sg_link.
+
+    .. figure:: scatter_gather_images/interposer_diagram.png
+        
+        An example of how specifying interposers in VTR works. Connections between the two sides of a cut are first severed. The two sides are then reconnected using scatter_gather patterns. Note that there are 'num' of each pattern at each switchblock locations.
+
```
vaughnbetz, line 609: 1. I think the figure should show the die-crossing wires at each x location, since that's what would happen.
2. On the figure or in the text, say the length of the sg_link wire in this case is 3.

libs/libarchfpga/src/interposer_types.h
```diff
@@ -0,0 +1,51 @@
...
+    int num;              ///< Number of scatter-gather instantiations per switchblock location
+};
+
+/**
+ * @brief Struct containing information of an interposer cut
+ * 
+ */
+struct t_interposer_cut_inf {
+    e_interposer_cut_dim dim;                        ///< Dimension or axis of interposer cut
+    int loc;                                         ///< Location of the cut on the grid
```
vaughnbetz: Give an example for this one.  If the interposer cut is at y = 13, is the cut above or below the tiles with coordinate y = 13?

libs/libarchfpga/src/interposer_types.h
```diff
@@ -0,0 +1,51 @@
...
+    {'y', e_interposer_cut_dim::Y}};
+
+/**
+ * @brief Struct containing information of interdire wires i.e. connections between the dies on an interposer
+ * 
+ */
+struct t_interdie_wire_inf {
+    std::string sg_name;  ///< Name of the scatter-gather pattern to be used for the interdie connection
+    std::string sg_link;  ///< Name of the scatter-gather link to be used for the interdie connection
+    int offset_start;     ///< Starting point of scatter-gather instantiations
```
vaughnbetz: I think you can beef up these comments a bit.  If the interposer cut is at y = 13, is an offset of 0 for an upward going sg_link a switch block / wire start point at y = 13?  For a downward going wire with a cut at y = 13, is an offset of 0 a sg_link wire start point at y = 14?
Are the start and end points inclusive (I think they are, but you should say that).

doc/src/arch/reference.rst
vaughnbetz: I assume the interposer cuts all layers?  Should say that too.  Or say it only supports 1 layer architectures for now.

libs/libarchfpga/src/read_xml_arch_file_interposer.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    std::string interposer_dim = pugiutil::get_attribute(interposer_cut_tag, "dim", loc_data).as_string();
+    if (interposer_dim.size() != 1 || !CHAR_INTERPOSER_DIM_MAP.contains(interposer_dim[0])) {
+        archfpga_throw(loc_data.filename_c_str(), loc_data.line(interposer_cut_tag), "Interposer tag dimension must be a single character of either X, x, Y or y.");
+    }
+
+    interposer.dim = CHAR_INTERPOSER_DIM_MAP.at(interposer_dim[0]);
+
+    interposer.loc = pugiutil::get_attribute(interposer_cut_tag, "loc", loc_data).as_int();
+    if (interposer.loc < 0) {
```
vaughnbetz, line 21: What does it mean if the interposer location is 0?  Goes back to the question of which side of an interposer the specified value goes on.  If the lower side, then a loc of 0 makes sense. Otherwise it doesn't make sense.

vaughnbetz (changes requested):

doc/src/arch/reference.rst
AmirhosseinPoolad: Yes, it was absolute versus relative [to device height/width] position, as in numbers only. Changed the comment to say "integer values" instead.

doc/src/arch/reference.rst
AmirhosseinPoolad: I'm not entirely sure about what happens with 3D architectures and interposers. The idea I had was that each layer could have it's own independent cut lines to keep everything flexible. This might make physical sense for a 2 layer face-to-face stacked chip, but I'm not sure about that either.

Right now the code does independent cut lines for each layer, but I could change it if you think it wouldn't make physical sense.

vaughnbetz: I'm fine with that.  I'd mention that somewhere in the docs, unless it is super-obvious from its position in the various tags.

libs/libarchfpga/src/grid_types.h
```diff
@@ -0,0 +1,131 @@
...
+ *                      +----+                +----+           +----+
+ *                      |    |                |    |           |    |
+ *                      |    |                |    |    ...    |    |
+ *                      |    |                |    |           |    |
+ *                      +----+                +----+           +----+
+ *       (startx,starty)
+ *                            <-------------->
+ *                                 repeatx
+ *
+ * startx/endx and endx/endy define a rectangular region instances dimensions.
```
vaughnbetz: instances -> instance's

libs/libarchfpga/src/grid_types.h
```diff
@@ -0,0 +1,131 @@
...
+    t_grid_loc_def(std::string block_type_val, int priority_val)
+        : block_type(std::move(block_type_val))
+        , priority(priority_val)
+        , x{"0", "W-1", "max(w+1,W)", "w"} //Fill in x direction, no repeat, incr by block width
+        , y{"0", "H-1", "max(h+1,H)", "h"} //Fill in y direction, no repeat, incr by block height
+    {}
+
+    std::string block_type; //The block type name
+
+    int priority = 0; //Priority of the specification.
```
vaughnbetz: Make these member comments Doxygen format

libs/libarchfpga/src/grid_types.h
```diff
@@ -0,0 +1,131 @@
...
+
+struct t_metadata_dict;
+
+/**
+ * @brief Grid location specification
+ * Each member is a formula evaluated in terms of 'W' (device width),
+ * and 'H' (device height). Formulas can be evaluated using parse_formula()
+ * from expr_eval.h.
+ */
+struct t_grid_loc_spec {
```
vaughnbetz, line 21: I think we should build these into the grid part of the VPR API documentation on the web. You'd need to edit some doxyfile for that, and add an overview at the top of this struct saying this is used to capture the user's specification of the architecture will eventually be turned into a flattened grid.

libs/libarchfpga/src/grid_types.h
```diff
@@ -0,0 +1,131 @@
...
+    // When 1 metadata tag is split among multiple t_grid_loc_def, one
+    // t_grid_loc_def is arbitrarily chosen to own the metadata, and the other
+    // t_grid_loc_def point to the owned version.
+    std::unique_ptr<t_metadata_dict> owned_meta;
+    t_metadata_dict* meta = nullptr; // Metadata for this location definition. This
+                                     // metadata may be shared with multiple grid_locs
+                                     // that come from a common definition.
+};
+
+enum class GridDefType {
```
vaughnbetz: Add Doxygen comment.

libs/libarchfpga/src/physical_types.h
```diff
@@ -248,126 +250,10 @@ typedef enum e_power_estimation_method_ t_power_estimation_method;
...
-};
-
-enum GridDefType {
-    AUTO,
-    FIXED
-};
 
 struct t_layer_def {
-    std::vector<t_grid_loc_def> loc_defs; //The list of block location definitions for this layer specification
+    std::vector<t_grid_loc_def> loc_defs;              ///< List of block location definitions for this layer specification
```
vaughnbetz, line 255: It would be good to make an overall Doxygen comment for what this struct is intended for.

vaughnbetz (approved):
I suggest a few more commenting updates, and it would be good to get the grid_def data structure in the online documentation (maybe as another item in the grid tab).

Or, maybe we should put the architecture file capturing structs (grid_def and interposer) in a new Doxygen tab for architecture constructs?

AmirhosseinPoolad:
Added the requested changes. For now, I added the doxygen documentation for the grid structs in the grid tab of VPR, but it might be useful to have a LibArchFPGA specific tab. Unfortunately much of LibArchFPGA is not Doxygen documented so it won't be much of a tab.

soheilshahrouz:
Is there a test for parsing this new tag?

vaughnbetz:
I like the idea of a libarchfpga tab, even if it is currently sparse.  Could you add that?

@soheilshahrouz is also right that you should add a parsing test.

AmirhosseinPoolad:
Added a test and also a libarchfpga tab to docs. The added tab is pretty sparse at the moment, but I think fixing that would go beyond the scope of this PR.

doc/src/api/libarchfpga/physical_types.rst
```diff
@@ -0,0 +1,33 @@
+========
+Physical Types
+========
+
+These types are used to capture user's intended architecture specification and will be later turned into a flattened device grid according to the device's size.
```
vaughnbetz: This comment isn't right ; this goes with the grid defs, but not the physical types.

doc/src/api/libarchfpga/scatter_gather_types.rst
```diff
@@ -0,0 +1,17 @@
+========
+Scatter-Gather Types
+========
+
+These types store information about about scatter-gather patterns
```
vaughnbetz: about scatter-gather routing patterns (I'd add the word routing).

vaughnbetz (approved):
Looks good, but I have two minor nits about the high-level doxygen comments. If you fix those, it's good to merge.

doc/src/api/libarchfpga/physical_types.rst
AmirhosseinPoolad: I used the grid defs page as a template for the other pages and forgot to change this one's description. Fixed it now.

PR #3268, opened 2025-09-09
RR Graph / Architecture Pin Side Consistency Check

amin1377:
Adding a check to the RR Graph node validation to raise an error if the pin sides stored in the RR Graph don’t correspond to the architecture file.


libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -511,6 +513,19 @@ void check_rr_node(const RRGraphView& rr_graph,
                 VPR_FATAL_ERROR(VPR_ERROR_ROUTE,
                                 "in check_rr_node: inode %d (type %d) has a capacity of %d.\n", inode, rr_type, capacity);
             }
+            rr_graph_sides = rr_graph.node_sides(rr_node);
+            std::tie(std::ignore, std::ignore, arch_side_vec) = get_pin_coordinates(type, ptc_num, std::vector<e_side>(TOTAL_2D_SIDES.begin(), TOTAL_2D_SIDES.end()));
+            // Number of sides in arch_side_vec may be higher than rr_graph sides since there may be duplicates (a pin may have different x/y offset on the same side)
+            // Because of that, we itearte over arch_side_vec and check if the side is in rr_graph_sides
```
vaughnbetz: itearate -> iterate

PR #3269, opened 2025-09-09
Fix bug where sg_link offsets would be uninitialized

AmirhosseinPoolad:
This PR simplifies the logic and fixes a bug where the zero offsets would be left uninitialized.

### DIFF
--- a/libs/libarchfpga/src/read_xml_arch_file_sg.cpp
+++ b/libs/libarchfpga/src/read_xml_arch_file_sg.cpp
@@ -4,6 +4,7 @@
 #include "pugixml_util.hpp"
 #include "arch_error.h"
 #include "switchblock_types.h"
+#include "vtr_util.h"
 
 /**
  * @brief Parses all <sg_link> tags under a <sg_link_list> tag.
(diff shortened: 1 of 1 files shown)


libs/libarchfpga/src/read_xml_arch_file_sg.cpp
```diff
@@ -26,32 +27,12 @@ static std::vector<t_sg_link> parse_sg_link_tags(pugi::xml_node sg_link_list_tag
...
-            if (x_offset != 0 || z_offset != 0) {
-                archfpga_throw(loc_data.filename_c_str(), loc_data.line(node), "More than one of the offset fields in the <sg_link> are non-zero.");
-            }
-        }
-        if (z_offset != 0) {
-            sg_link.z_offset = z_offset;
-            if (y_offset != 0 || x_offset != 0) {
-                archfpga_throw(loc_data.filename_c_str(), loc_data.line(node), "More than one of the offset fields in the <sg_link> are non-zero.");
-            }
+        if (vtr::exactly_k_conditions(1, sg_link.x_offset != 0, sg_link.y_offset != 0, sg_link.z_offset != 0)) {
```
soheilshahrouz: I think the condition should be negated

PR #3270, opened 2025-09-09
Add test for undefined values equality checking

AmirhosseinPoolad:
Adds a static_assert to make sure the three undefined value constants are all equal to -1.


vpr/test/test_undefined_value.cpp
```diff
@@ -0,0 +1,6 @@
+#include "vpr_types.h"
+#include "arch_types.h"
+#include "librrgraph_types.h"
+
+// Make sure the 'unkown value' constants in VPR, LibArchFPGA and LibRRGraph are all equal to the same value (-1)
```
AlexandreSinger: "unkown" -> "unknown"

Also I think you add a bit on why. Just something simple like "these should be equal since they are sometimes intermixed by mistake".

PR #3271, opened 2025-09-10
[Place] Estimating Starting T Using Equilibrium T

AlexandreSinger:
I found that the way that the annealer estimates the initial temperature to be too high when the initial placement is of very good quality (for example, after AP).

Added a new way of estimating the starting temperature by setting it to an estimation of the equilibrium temperature. The equilibrium temperature is the temperature at which the change in cost after an annealing iteration would be 0.

The old way (of using the variance of the change in cost) is still the default; however, this new method can be turned on in the command-line.

### DIFF
--- a/vpr/src/analytical_place/full_legalizer.cpp
+++ b/vpr/src/analytical_place/full_legalizer.cpp
@@ -59,6 +59,9 @@
 
 #include "setup_grid.h"
 #include "stats.h"
+#ifndef NO_GRAPHICS
+#include "draw_global.h"
+#endif
 
 std::unique_ptr<FullLegalizer> make_full_legalizer(e_ap_full_legalizer full_legalizer_type,
                                                    const APNetlist& ap_netlist,
@@ -1003,6 +1006,7 @@ void FlatRecon::legalize(const PartialPlacement& p_placement) {
 
     // Perform the initial placement on created clusters.
     place_clusters(p_placement);
+    update_drawing_data_structures();
 }
 
 void NaiveFullLegalizer::create_clusters(const PartialPlacement& p_placement) {
@@ -1219,6 +1223,7 @@ void NaiveFullLegalizer::legalize(const PartialPlacement& p_placement) {
     //       made part of the placement and verify placement should check for
     //       it.
     post_place_sync();
+    update_drawing_data_structures();
 }
 
 void APPack::legalize(const PartialPlacement& p_placement) {
@@ -1304,4 +1309,14 @@ void APPack::legalize(const PartialPlacement& p_placement) {
 
     // Synchronize the pins in the clusters after placement.
     post_place_sync();
+    update_drawing_data_structures();
+}
+
+void FullLegalizer::update_drawing_data_structures() {
+#ifndef NO_GRAPHICS
+    // update graphic resources incase of clustering changes
+    if (get_draw_state_vars()) {
+        get_draw_state_vars()->refresh_graphic_resources_after_cluster_change();
+    }
+#endif
 }
--- a/vpr/src/analytical_place/full_legalizer.h
+++ b/vpr/src/analytical_place/full_legalizer.h
@@ -59,6 +59,9 @@ class FullLegalizer {
      *                      device grid and fixed blocks are observed.
      */
     virtual void legalize(const PartialPlacement& p_placement) = 0;
+    
+    /// @brief Update drawing data structure for current placement
+    void update_drawing_data_structures();
 
   protected:
     /// @brief The AP Netlist to fully legalize the flat placement of.
--- a/vpr/src/draw/draw_types.cpp
+++ b/vpr/src/draw/draw_types.cpp
@@ -43,6 +43,24 @@ void t_draw_state::reset_block_colors() {
               true);
 }
 
+void t_draw_state::refresh_graphic_resources_after_cluster_change() {
+    const ClusteringContext& cluster_ctx = g_vpr_ctx.clustering();
+    const AtomContext& atom_ctx = g_vpr_ctx.atom();
+
+    // Resize block color vectors to match the possibly new clustered size
+    block_color_.resize(cluster_ctx.clb_nlist.blocks().size());
+    use_default_block_color_.resize(cluster_ctx.clb_nlist.blocks().size());
+    reset_block_colors();
+
+    // Resize net color as well since they might be rebuild in analytical placement
+    if (is_flat) {
+        net_color.resize(atom_ctx.netlist().nets().size());
+    } else {
+        net_color.resize(cluster_ctx.clb_nlist.nets().size());
+    }
+    std::fill(net_color.begin(), net_color.end(), ezgl::BLACK);
+}
+
 /**************************************************
  * begin t_draw_pb_type_info function definitions *
  **************************************************/
--- a/vpr/src/draw/draw_types.h
+++ b/vpr/src/draw/draw_types.h
@@ -333,6 +333,9 @@ struct t_draw_state {
     void reset_block_color(ClusterBlockId blk);
     void reset_block_colors();
 
+    ///@brief Refresh graphics resources' size after update to cluster blocks size
+    void refresh_graphic_resources_after_cluster_change();
+
     std::vector<std::pair<t_pl_loc, ezgl::color>> colored_locations;
 
     /// @brief Stores UI checkbox struct for passing into the callback functions


vpr/src/place/annealer.cpp
```diff
@@ -345,7 +482,7 @@ float PlacementAnnealer::estimate_starting_temperature_() {
     return init_temp;
 }
 
-e_move_result PlacementAnnealer::try_swap_(MoveGenerator& move_generator,
```
AmirhosseinPoolad, line 348: If I understand correctly, this new function could possibly add one more copy (cost and outcome vs outcome only) to every single annealing move. You should definitely benchmark to see if runtime of the default flow is hurt by this change.

AlexandreSinger: One copy on the scale of a function as expensive as try_swap should have no affect imo.

doc/src/vpr/command_line_usage.rst
```diff
@@ -840,6 +840,18 @@ If any of init_t, exit_t or alpha_t is specified, the user schedule, with a fixe
...
 
+.. option:: ----anneal_auto_init_t_estimator {cost_variance, equilibrium_est}
+
+   Controls which estimation method is used when selecting the starting temperature
+   for the automatic annealing schedule.
+
+   The options for estimators are:
+
+   * ``cost_variance``: Estimates the initial temperature using the variance of cost after a set of trial swaps.
+   * ``equilibrium_est``: Estimates the initial temperature by trying to predict the equilibrium temperature for the initial placement (i.e. the temperature that would result in no change in cost).
```
vaughnbetz: I'd just called this 'equilibrium' or 'equilibrium_t'.  The _est doesn't add value in my opinion.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_place_init_t_est/config/config.txt
```diff
@@ -0,0 +1,28 @@
...
+circuits_dir=benchmarks/blif/wiremap6
+
+# Path to directory of architectures to use
+archs_dir=arch/timing
+
+# Add architectures to list to sweep
+arch_list_add=k6_frac_N10_40nm.xml
+
+# Add circuits to list to sweep
+circuit_list_add=alu4.pre-vpr.blif
```
vaughnbetz, line 11: Doesn't matter much because these are probably fast, but one circuit is probably enough to test both these flows work.

vpr/src/place/annealer.cpp
```diff
@@ -291,11 +294,145 @@ PlacementAnnealer::PlacementAnnealer(const t_placer_opts& placer_opts,
...
+    // Perform N trial swaps and collect the change in cost for each of these
+    // swaps. Accepted swaps are swaps which resulted in a negative change in
+    // cost, rejected swaps are swaps which resulted in a positive change in
+    // cost.
+    std::vector<double> accepted_swaps;
+    std::vector<double> rejected_swaps;
+    accepted_swaps.reserve(move_lim);
+    rejected_swaps.reserve(move_lim);
+    for (int i = 0; i < move_lim; i++) {
+        bool manual_move_enabled = false;
```
vaughnbetz: I'd delete this. No need to enable manual moves during initial temperature calculations -- this is intended to let people directly control the annealer, but just doing that in the main annealer is enough.  

There should be code like this at the start of the main annealer -- if it is there, there isn't any need to duplicate it here.  If it isn't there, I think we should move this there anyway.

vpr/src/place/annealer.cpp
```diff
@@ -291,11 +294,145 @@ PlacementAnnealer::PlacementAnnealer(const t_placer_opts& placer_opts,
...
+    // exists within this range.
+    //      Initialize the lower bound temperature to 0. The temperature cannot
+    //      be less than 0.
+    double lower_bound_temp = 0.0;
+    //      Initialize the upper bound temperature to 1e10. It is possible for
+    //      the equilibrium temperature to be infinite if the initial placement
+    //      is so bad that no swaps are accepted. In that case this value will
+    //      be returned instead of infinity.
+    //      TODO: Find what a reasonable value for this should be.
+    double upper_bound_temp = 1e10;
```
vaughnbetz: Could be smarter about this ... e.g. could use 3* absolute value of the largest cost increase in rejected_swaps (that swap would have had a 71% chance of being accepted at that temperature in that case).

vpr/src/place/annealer.cpp
```diff
@@ -291,11 +294,145 @@ PlacementAnnealer::PlacementAnnealer(const t_placer_opts& placer_opts,
...
+        double trial_temp = (lower_bound_temp + upper_bound_temp) / 2.0;
+
+        // Calculate the expected change in cost at this temperature (which we
+        // call the residual here).
+        double expected_total_post_rejected_cost = 0.0;
+        for (double rejected_cost : rejected_swaps) {
+            // Expected change in cost after a rejected swap is the change in
+            // cost multiplied by the probability that this swap is accepted at
+            // this temperature.
+            double accpetance_prob = std::exp((-1.0 * rejected_cost) / trial_temp);
```
vaughnbetz: typo: acceptance_prob

amin1377:
@AlexandreSinger: For the results you reported on Titan and VTR benchmarks, are you using the default placement flow with just the equilibrium approach for init temp, or is this with AP? Also, if the results are for AP, when you say “place time,” are you referring to the total placement time or just the detailed placement (i.e., SA) portion? Thanks!

doc/src/vpr/command_line_usage.rst
```diff
@@ -840,6 +840,18 @@ If any of init_t, exit_t or alpha_t is specified, the user schedule, with a fixe
...
     **Default:** ``1.0``
 
+.. option:: ----anneal_auto_init_t_estimator {cost_variance, equilibrium_est}
+
+   Controls which estimation method is used when selecting the starting temperature
+   for the automatic annealing schedule.
+
+   The options for estimators are:
+
+   * ``cost_variance``: Estimates the initial temperature using the variance of cost after a set of trial swaps.
```
amin1377: Although I know this isn’t done for all parameters, the current documentation for options doesn’t really explain how the initial temperature is chosen. I think it would be helpful if you could expand the documentation a bit. Later, once you publish this work, we can remove the explanation and simply point to the paper.

AlexandreSinger:
> @AlexandreSinger: Also, since this PR makes equilibrium the default init temp estimator, I’d suggest running the Koios benchmark as well. It’s become almost standard to run all three (VTR, Titan, Koios) to ensure there are no regressions.

That was a typo... The default is variance in the code. I just messed up the docs. Thanks for pointing that out!

doc/src/vpr/command_line_usage.rst
AlexandreSinger: *if* I publish this work (: 

And sounds good. TBH the initial temperature is just proportional to the variance of the cost. But I can make the documentation more explicit here.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_place_init_t_est/config/config.txt
AlexandreSinger: Not a problem, I will add a few extra!

vpr/src/place/annealer.cpp
AlexandreSinger: I agree, this should not be here assuming that the entire purpose of this method is to estimate the starting temperature.

Looking at the code, however, I do not believe that the manual move feature is working at all. It appears to only ever turn on in the initial temperature estimation. I do not believe that this should be resolved in this PR, but I can raise an issue if you would like?

In the meantime, I will remove this from the equilibrium estimator (which I have added), but I will leave it the default flow for now so we do not regress the feature further.

AlexandreSinger:
@vaughnbetz I have updated the PR with your comments. Let me know if you have any further comments. Once we merge this in, I want to experiment more with accepting the swaps or not during the estimation.

vpr/src/place/annealer.cpp
AlexandreSinger: Ill raise an issue to look into this!

vaughnbetz (approved):
LGTM

AlexandreSinger:
@vaughnbetz I will present these results at the VTR meeting today, but I ran the original temperature estimator (variance) and my proposed equilibrium estimator on a set of placements across the Titan Benchmark Suite:

- Default VPR's Initial Placement
- Default VPR's Final Placement
- AP's Initial Placement
- AP's Final Placement

And got the final results:

  | First Iter Cost Ratio Average | First Iter Cost Ratio Variance | First Iter Cost Ratio Average Absolute Error | Geomean Initial Temperature | Geomean Place Time | Geomean Final WL | Geomean Final CPD
Variance (default) | 1.4045 | 0.4593 | 0.4048 | 1.00 | 1.00 | 1.00 | 1.00
Equilibrium | 0.9995 | 0.0039 | 0.0319 | 0.22 | 0.81 | 1.00 | 1.01


I wanted to post these here to keep organized. As you can see, across all of these placements, the ratio of the cost before and after first iteration is MUCH closer to 1.00 and the variance is much tighter. This results in an initial temperature that is more than 4x lower than default with little to no loss in WL or CPD.

I wrote a script to collect these results automatically, so I am starting to use it to explore variations on my equilibrium approach. Will discuss in the VTR meeting.

@amin1377 FYI

PR #3273, opened 2025-09-12
[bug fix] analytical place with graphics no longer segfault

robluo:
Analytical placement update clustering global context later compared to original placement, but the graphic resources are initialized earlier in vpr flow. This causes a segmentation fault when enabling both display on and analytical place. 
This PR fixes the issue. 

#### Description

Resized graphic resources (block_color and net_color) in post_place_sync().

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/base/place_and_route.cpp
+++ b/vpr/src/base/place_and_route.cpp
@@ -20,9 +20,6 @@
 #include "route.h"
 #include "route_export.h"
 #include "draw.h"
-#ifndef NO_GRAPHICS
-#include "draw_global.h"
-#endif
 #include "rr_graph.h"
 #include "read_xml_arch_file.h"
 #include "route_common.h"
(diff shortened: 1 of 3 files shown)


vpr/src/base/place_and_route.cpp
```diff
@@ -604,4 +607,11 @@ void post_place_sync() {
     for (const ClusterBlockId block_id : cluster_ctx.clb_nlist.blocks()) {
         blk_loc_registry.place_sync_external_block_connections(block_id);
     }
+
+#ifndef NO_GRAPHICS
+    // update graphic resources incase of clustering changes
+    if (get_draw_state_vars()) {
+        get_draw_state_vars()->refresh_graphic_resources_after_cluster_change();
```
AlexandreSinger: I do not think this bit of code belongs in this function. I recommend moving it somewhere else. Perhaps even into its own function. Post-place sync is not doing things related to drawing directly.

robluo: I moved this to a full legalizer member function and it is called at the end of all versions of legalize().

AlexandreSinger (approved):
LGTM

vaughnbetz:
@robluo : one remaining thing to investigate, to see if we can have a simpler solution: Do we need a refresh function like this, or could we just move where the drawing state is initialized? It seems like we usually initialize it after clustering, but with analytical placement enabled somehow it gets initialized before we've completed clustering. Could we just move it so it is called after clustering, whether analytical placement is enabled or not?

PR #3276, opened 2025-09-16
[WIP] Scatter-gather patterns for 3D architectures

soheilshahrouz:
This PR includes the following changes:
1) Creates nodes and edges in RR graph based on scatter-gather patterns specifed in the architecture file.
2) Updates router lookahead map to handle CHANZ nodes properly.
3) Computes indexed data for CHANZ nodes.
4) Removes 3-d (with above and under) switch blocks from code


libs/libarchfpga/src/parse_switchblocks.cpp
```diff
@@ -328,19 +328,18 @@ static void parse_comma_separated_wire_points(const char* ch, std::vector<t_wire
...
-//set sides for a specific conn for custom switch block pattern
-static void set_switch_func_type(SBSideConnection& conn, const char* func_type) {
+static void set_switch_func_type(SBSideConnection& conn, std::string_view func_type) {
 
-    if (std::string(func_type).length() != 2) {
+    if (func_type.length() != 2) {
         archfpga_throw(__FILE__, __LINE__, "Custom switchblock func type must be 2 characters long: %s\n", func_type);
     }
 
     // Only valid sides are right, top, left, bottom, above and under
```
AlexandreSinger: Since you removed above and under, this comment should be updated.

Is it documented that we want to remove the above and under from this?

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -902,13 +913,15 @@ class t_rr_graph_storage {
...
-    vtr::vector<RRNodeId, short> node_layer_;
+    // Layer number refers to the die that the node belongs to.
+    // The layer number of base die is zero and die above it one, etc.
+    // This data is also considered as a hot data since it is used in inner loop of router,
+    // but since it didn't fit nicely into t_rr_node_data due to alignment issues, we had to store it in a separate vector.
+
+    /// @brief The lowest layer number where a given node is located at.
+    vtr::vector<RRNodeId, char> node_layer_low_;
+    /// @brief The highest layer number where a given node is located at.
+    vtr::vector<RRNodeId, char> node_layer_high_;
```
AlexandreSinger: I know that the RR graph is an extremely large data structure and is sensitive to data types. I am concerned about turning a vector of shorts into two vectors of chars. TBH I have no idea what would happen in this case lol.

However, if we think that the low and high layers will be access together usually, it may be a better idea to create a vector of structs, where we pack the low and high together into a short type. Since you already have accessor methods, it should be simple to change this. Just an idea, its up to you.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -1091,14 +1106,14 @@ class t_rr_graph_view {
...
-     */
-    short node_layer(RRNodeId id) const{
-        return node_layer_[id];
+    /// @brief Retrieve the lowest layer (die) number where the given RRNodeId is located.
+    short node_layer_low(RRNodeId id) const {
+        return node_layer_low_[id];
+    }
+
+    /// @brief Retrieve the highest layer (die) number where the given RRNodeId is located.
+    short node_layer_high(RRNodeId id) const {
```
AlexandreSinger: Should this return a short or a char? I am cautious of implicit casts like this, even if it is larger.

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -225,10 +228,14 @@ class RRGraphView {
         return node_storage_.node_yhigh(node);
     }
 
-    /** @brief Return the layer num of a specified node.
-    */
-    inline short node_layer(RRNodeId node) const {
-        return node_storage_.node_layer(node);
+    /// @brief Returns the highest layer where a node is located at.
+    inline short node_layer_high(RRNodeId node) const {
```
AlexandreSinger: Same comment here. Should this be char?

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -225,10 +228,14 @@ class RRGraphView {
...
+        return node_storage_.node_layer_high(node);
+    }
+
+    /// @brief Returns the lowest layer where a node is located at.
+    inline short node_layer_low(RRNodeId node) const {
```
AlexandreSinger: Same here.

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -717,8 +721,11 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         temp_string_ = rr_graph_builder_->node_ptc_nums_to_string(node.id());
         return temp_string_.c_str();
     }
-    inline int get_node_loc_layer(const t_rr_node& node) final {
-        return rr_graph_->node_layer(node.id());
+    inline int get_node_loc_layer_low(const t_rr_node& node) final {
```
AlexandreSinger, line 723: This is now being cast to an integer. Is this intentional?

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -733,13 +740,20 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         return rr_graph_->node_ylow(node.id());
     }
 
-    inline void set_node_loc_layer(int layer_num, int& inode) final {
+    inline void set_node_loc_layer_low(int layer_num, int& inode) final {
         auto node = (*rr_nodes_)[inode];
         RRNodeId node_id = node.id();
 
+        VTR_ASSERT(layer_num >= 0);
+        rr_graph_builder_->set_node_layer_low(node_id, layer_num);
```
AlexandreSinger: Layer num is an integer type here. If you plan to keep that the case, you should add an assert to ensure its not larger than the size of a char.

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -733,13 +740,20 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
+    }
+
+    inline void set_node_loc_layer_high(int layer_num, int& inode) final {
+        auto node = (*rr_nodes_)[inode];
+        RRNodeId node_id = node.id();
 
         VTR_ASSERT(layer_num >= 0);
-        rr_graph_builder_->set_node_layer(node_id, layer_num);
+        rr_graph_builder_->set_node_layer_high(node_id, layer_num);
```
AlexandreSinger: Same comment here.

libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp
```diff
@@ -139,10 +164,12 @@ void alloc_and_load_rr_indexed_data(const RRGraphView& rr_graph,
 
     load_rr_indexed_data_base_costs(rr_graph, rr_indexed_data, base_cost_type, echo_enabled, echo_file_name);
 
-    if (echo_enabled) {
+    if (true) {
```
AlexandreSinger: A bit of a nit-pick, this should just be inlined then. If (true) should not be here.

vpr/src/route/router_lookahead/router_lookahead_map.cpp
```diff
@@ -482,16 +496,26 @@ float MapLookahead::get_opin_distance_min_delay(int physical_tile_idx, int from_
...
     VTR_ASSERT_SAFE(to_layer_num < static_cast<int>(f_wire_cost_map.dim_size(1)));
     VTR_ASSERT_SAFE(seg_index < static_cast<int>(f_wire_cost_map.dim_size(3)));
     VTR_ASSERT_SAFE(delta_x < static_cast<int>(f_wire_cost_map.dim_size(4)));
     VTR_ASSERT_SAFE(delta_y < static_cast<int>(f_wire_cost_map.dim_size(5)));
 
-    int chan_index = 0;
-    if (rr_type == e_rr_type::CHANY) {
-        chan_index = 1;
+    int chan_index;
+    switch (rr_type) {
```
AlexandreSinger: I recommend making this a helper method. This is used a couple of times in this file.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3815,20 +3817,20 @@ static std::vector<t_segment_inf> process_segments(pugi::xml_node Parent,
     pugi::xml_node SubElem;
     pugi::xml_node Node;
 
-    /* Count the number of segs and check they are in fact
-     * of segment elements. */
+    // Count the number of segs and check they are in fact of segment elements.
     int NumSegs = count_children(Parent, "segment", loc_data);
```
AmirhosseinPoolad: snake_case

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3815,20 +3817,20 @@ static std::vector<t_segment_inf> process_segments(pugi::xml_node Parent,
     pugi::xml_node SubElem;
     pugi::xml_node Node;
 
-    /* Count the number of segs and check they are in fact
-     * of segment elements. */
+    // Count the number of segs and check they are in fact of segment elements.
```
AmirhosseinPoolad: This comment was very poorly worded imo, can you rewrite it?

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -3885,16 +3887,19 @@ static std::vector<t_segment_inf> process_segments(pugi::xml_node Parent,
             } else if (strcmp(tmp, "y") == 0) {
                 Segs[i].parallel_axis = e_parallel_axis::Y_AXIS;
                 y_axis_seg_found = true;
+            } else if (strcmp(tmp, "z") == 0) {
```
AmirhosseinPoolad, line 3911: Can you use string_view for tmp instead to avoid doing C style string stuff?

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -4164,7 +4174,11 @@ static void process_bend(pugi::xml_node Node, t_segment_inf& segment, const int
         part_len.push_back(list.size() + 1 - sum_len);
 }
 
-static void calculate_custom_SB_locations(const pugiutil::loc_data& loc_data, const pugi::xml_node& SubElem, const int grid_width, const int grid_height, t_switchblock_inf& sb) {
+static void calculate_custom_SB_locations(const pugiutil::loc_data& loc_data,
+                                          const pugi::xml_node& SubElem,
```
AmirhosseinPoolad, line 4199: snake_case

libs/libarchfpga/src/switchblock_types.h
```diff
@@ -15,6 +15,28 @@ enum e_directionality {
...
+    UNIVERSAL,
+
+    /// The FULL switch block type allows for complete connectivity between tracks.
+    FULL,
+
+    /// A CUSTOM switch block has also been added which allows a user to describe custom permutation functions and connection patterns.
+    /// See comment at top of SRC/route/build_switchblocks.c
+    CUSTOM
+};
+
```
AmirhosseinPoolad, line 39: Convert comments to doxygen style (///<)

libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -347,7 +347,8 @@ void check_rr_node(const RRGraphView& rr_graph,
     int xhigh = rr_graph.node_xhigh(rr_node);
     int ylow = rr_graph.node_ylow(rr_node);
     int yhigh = rr_graph.node_yhigh(rr_node);
-    int layer_num = rr_graph.node_layer(rr_node);
+    int layer_low = rr_graph.node_layer_low(rr_node);
+    int layer_high = rr_graph.node_layer_low(rr_node);
```
AmirhosseinPoolad: Both are rr_graph_node_layer_low, is that intended?

libs/librrgraph/src/base/get_parallel_segs.cpp
```diff
@@ -6,7 +6,9 @@ std::vector<t_segment_inf> get_parallel_segs(const std::vector<t_segment_inf>& s
                                              bool keep_original_index /* = false */) {
     std::vector<t_segment_inf> result;
     for (size_t i = 0; i < segment_inf.size(); ++i) {
-        if (segment_inf[i].parallel_axis == parallel_axis || segment_inf[i].parallel_axis == e_parallel_axis::BOTH_AXIS) {
+        e_parallel_axis seg_axis = segment_inf[i].parallel_axis;
+        if (seg_axis == parallel_axis
+            || (parallel_axis != e_parallel_axis::Z_AXIS && seg_axis == e_parallel_axis::BOTH_AXIS)) {
```
AmirhosseinPoolad, line 15: Add comment explaining why we exclude z_axis here.

libs/librrgraph/src/base/rr_graph_utils.cpp
```diff
@@ -164,13 +165,19 @@ void rr_set_sink_locs(const RRGraphView& rr_graph, RRGraphBuilder& rr_graph_buil
...
         // and its new location has already been set
-        if ((node_xhigh - node_xlow) == 0 && (node_yhigh - node_ylow) == 0)
+        if ((node_xhigh - node_xlow) == 0 && (node_yhigh - node_ylow) == 0) {
+            continue;
+        }
+
+        const int node_layer = rr_graph.node_layer_low(node_id);
+        // Only CHANZ nodes spand across multiple layers. We're processing SINK nodes
+        if (node_layer !=  rr_graph.node_layer_high(node_id)) {
             continue;
```
AmirhosseinPoolad: Shouldn't you assert/crash here?

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -401,19 +407,19 @@ class RRGraphView {
...
                 end_y = std::to_string(node_ylow(node)) + ",";
-                end_layer_str = std::to_string(node_layer_num) + ")";
+                end_layer_str = std::to_string(node_layer_low(node)) + ")";
             }
 
             else {                                                      // signal travels in increasing direction, stays at same point, or can travel both directions
                 start_x = " (" + std::to_string(node_xlow(node)) + ","; //start coordinates have smaller value
                 start_y = std::to_string(node_ylow(node)) + ",";
-                start_layer_str = std::to_string(node_layer_num) + ")";
+                start_layer_str = std::to_string(node_layer_high(node)) + ")";
```
AmirhosseinPoolad: I think this should be node_layer_high

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -423,12 +423,15 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
             segment_inf_vec_ptr = &segment_inf_x_;
-        else
+        } else if (axis == e_parallel_axis::Y_AXIS) {
             segment_inf_vec_ptr = &segment_inf_y_;
+        } else {
+            segment_inf_vec_ptr = &segment_inf_z_;
+        }
 
-        for(std::vector<t_segment_inf>::size_type i=0; i < (*segment_inf_vec_ptr).size(); i++){
+        for(std::vector<t_segment_inf>::size_type i=0; i < segment_inf_vec_ptr->size(); i++){
```
AmirhosseinPoolad: Can you change the std::vector::size_type to size_t? This looks overly complicated.

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -733,13 +740,23 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
         return rr_graph_->node_ylow(node.id());
     }
 
-    inline void set_node_loc_layer(int layer_num, int& inode) final {
+    inline void set_node_loc_layer_low(int layer_num, int& inode) final {
         auto node = (*rr_nodes_)[inode];
```
AmirhosseinPoolad: Auto

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -733,13 +740,23 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
...
 
+        // Currently, we only support two layers
+        VTR_ASSERT(layer_num >= 0 && layer_num <= 1);
+        rr_graph_builder_->set_node_layer_low(node_id, layer_num);
+    }
 
-        VTR_ASSERT(layer_num >= 0);
-        rr_graph_builder_->set_node_layer(node_id, layer_num);
+    inline void set_node_loc_layer_high(int layer_num, int& inode) final {
+        auto node = (*rr_nodes_)[inode];
```
AmirhosseinPoolad: auto

libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp
```diff
@@ -19,19 +19,48 @@
...
 
 static float get_delay_normalization_fac(const vtr::vector<RRIndexedDataId, t_rr_indexed_data>& rr_indexed_data, const bool echo_enabled, const char* echo_file_name);
 
 static void load_rr_indexed_data_T_values(const RRGraphView& rr_graph, vtr::vector<RRIndexedDataId, t_rr_indexed_data>& rr_indexed_data);
 
-static void calculate_average_switch(const RRGraphView& rr_graph, int inode, double& avg_switch_R, double& avg_switch_T, double& avg_switch_Cinternal, int& num_switches, int& num_shorts, short& buffered, vtr::vector<RRNodeId, std::vector<RREdgeId>>& fan_in_list);
+/**
+ * @brief Computes average R, Tdel, and Cinternal of fan-in switches for a given node.
+ *
+ * Iterates over all incoming edges of @p inode, skipping SHORT switches,
```
AmirhosseinPoolad, line 36: What is '@[no space here]p'?

soheilshahrouz: It points to function argument

libs/libarchfpga/src/switchblock_types.h
soheilshahrouz: `///` adds a doxygen comment for what follows the comment, while `///<` comments what came before the comment.

vpr/src/route/overuse_report.cpp
```diff
@@ -149,6 +155,8 @@ void report_overused_nodes(const Netlist<>& net_list,
                 report_sinks = true;
                 break;
 
+                // TODO: handle CHANZ nodes
+
```
AmirhosseinPoolad: TODO remaining in final code

vpr/src/route/router_lookahead/router_lookahead_compressed_map.cpp
```diff
@@ -422,8 +422,10 @@ std::pair<float, float> CompressedMapLookahead::get_expected_delay_and_cong(RRNo
     auto& device_ctx = g_vpr_ctx.device();
     auto& rr_graph = device_ctx.rr_graph;
 
-    int from_layer_num = rr_graph.node_layer(from_node);
-    int to_layer_num = rr_graph.node_layer(to_node);
+    // TODO: handle CHANZ nodes that span multiple layers
```
AmirhosseinPoolad, line 425: TODO again. Not saying that you have to fix these right now, but a reminder that it is there. I sometimes forget about my TODOs so I'm just making sure that it's intended.

vpr/src/route/router_lookahead/router_lookahead_map.cpp
```diff
@@ -504,29 +514,37 @@ static void compute_router_wire_lookahead(const std::vector<t_segment_inf>& segm
     const auto& device_ctx = g_vpr_ctx.device();
     const auto& grid = device_ctx.grid;
 
+    const size_t num_layers = grid.get_num_layers();
+    const size_t chan_type_dim_size = (num_layers == 1) ? 2 : 3;
```
AmirhosseinPoolad, line 523: Add comment explaining this.

vpr/src/route/router_lookahead/router_lookahead_simple.cpp
```diff
@@ -68,8 +68,9 @@ std::pair<float, float> SimpleLookahead::get_expected_delay_and_cong(RRNodeId fr
     e_rr_type from_type = rr_graph.node_type(from_node);
     util::Cost_Entry cost_entry(0, 0);
     if (is_chanxy(from_type) || is_chanz(from_type)) {
-        int from_layer_num = rr_graph.node_layer(from_node);
-        int to_layer_num = rr_graph.node_layer(to_node);
+        // TODO: handle CHANZ nodes
```
AmirhosseinPoolad, line 71: TODO. Again, just reminding that it exists.

AmirhosseinPoolad (commented):
Reviewed from DecompNetlistRouter to build_scatter_gathers.cpp.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+//
+// Static Function Definitions
+//
+
+static void index_to_correct_channels(const t_wireconn_inf& pattern,
+                                      const t_physical_tile_loc& loc,
+                                      const t_chan_details& chan_details_x,
+                                      const t_chan_details& chan_details_y,
+                                      std::vector<t_chan_loc>& correct_channels) {
+    correct_channels.clear();
```
AmirhosseinPoolad, line 64: If you're clearing this vector at the start of the function, why not just have it as return value then?

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+    for (e_side side : pattern.sides) {
+        t_physical_tile_loc chan_loc;
+        e_rr_type chan_type;
+
+        index_into_correct_chan(loc, side, chan_details_x, chan_details_y, chan_loc, chan_type);
```
AmirhosseinPoolad, line 70: The name for these two functions are way to similar, I thought it was recursive at first glance. I think 'index_to_correct_channels' should have sg or scatter_gather somewhere in the name to differentiate it better.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+    for (const auto [chan_loc, chan_type, chan_side] : channels) {
+        int seg_coord = (chan_type == e_rr_type::CHANY) ? chan_loc.y : chan_loc.x;
+
+        const t_wire_type_sizes& wire_type_sizes = (chan_type == e_rr_type::CHANX) ? wire_type_sizes_x : wire_type_sizes_y;
+        const t_chan_seg_details* chan_details = (chan_type == e_rr_type::CHANX) ? chan_details_x[chan_loc.x][chan_loc.y].data() : chan_details_y[chan_loc.x][chan_loc.y].data();
+
+        for (const t_wire_switchpoints& wire_switchpoints : wire_switchpoints_vec) {
+            auto wire_type = vtr::string_view(wire_switchpoints.segment_name);
+
+            if (wire_type_sizes.find(wire_type) == wire_type_sizes.end()) {
```
AmirhosseinPoolad: Prefer using contains from c++20

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+    std::vector<t_sg_candidate> candidates;
+
+    for (const auto [chan_loc, chan_type, chan_side] : channels) {
+        int seg_coord = (chan_type == e_rr_type::CHANY) ? chan_loc.y : chan_loc.x;
+
+        const t_wire_type_sizes& wire_type_sizes = (chan_type == e_rr_type::CHANX) ? wire_type_sizes_x : wire_type_sizes_y;
+        const t_chan_seg_details* chan_details = (chan_type == e_rr_type::CHANX) ? chan_details_x[chan_loc.x][chan_loc.y].data() : chan_details_y[chan_loc.x][chan_loc.y].data();
+
+        for (const t_wire_switchpoints& wire_switchpoints : wire_switchpoints_vec) {
+            auto wire_type = vtr::string_view(wire_switchpoints.segment_name);
```
AmirhosseinPoolad: Why do we have a vtr::string_view? Maybe it's from when the project was not using C++17? Anyway I don't see why we need to use it now. Also, auto.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+                        }
+                        if (seg_direction == Direction::INC && is_dest) {
+                            continue;
+                        }
+                    }
+
+                    int wire_switchpoint = get_switchpoint_of_wire(chan_type, chan_details[iwire], seg_coord, chan_side);
+
+                    // Check if this wire belongs to one of the specified switchpoints; add it to our 'wires' vector if so
+                    if (wire_switchpoint != valid_switchpoint) continue;
```
AmirhosseinPoolad: The loop seems backwards to me. You're going through all valid switchpoints then all wires, meaning that you do all the processing for the wire only to throw it away because in the outer loop we were looking for switchpoint number 1 instead of 2, even though switchpoint 2 might also be valid in the next iteration of the outer loop.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+                }
+            }
+        }
+    }
+
+    return candidates;
+}
+
+//
+// Non-satatic Function Definitions
```
AmirhosseinPoolad: static*

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+    std::vector<t_chan_loc> scatter_channels;
+
+    vtr::FormulaParser formula_parser;
+    vtr::t_formula_data formula_data;
+
+    std::vector<t_bottleneck_link> bottleneck_links;
+
+    interdie_3d_links.clear();
+    if (grid.get_num_layers() > 1) {
+        interdie_3d_links.resize({grid.width(), grid.height()});
```
AmirhosseinPoolad, line 184: Why not return a pair of std::vector<t_bottleneck_link> and interdie_3d_links instead? Sure the function signature would look terrible but I just don't like output arguments if we can help it and I think they make the code hard to understand.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    const auto [wire_type_sizes_x, wire_type_sizes_y] = count_wire_type_sizes(chan_details_x, chan_details_y, nodes_per_chan);
+
+    for (const t_scatter_gather_pattern& sg_pattern : scatter_gather_patterns) {
+        VTR_ASSERT(sg_pattern.type == e_scatter_gather_type::UNIDIR);
```
AmirhosseinPoolad, line 190: Should probably add to docs that only unidirectional patterns are currently supported.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+                t_physical_tile_loc scatter_loc;
+                scatter_loc.x = gather_loc.x + sg_link.x_offset;
+                scatter_loc.y = gather_loc.y + sg_link.y_offset;
+                scatter_loc.layer_num = gather_loc.layer_num + sg_link.z_offset;
+
+                index_to_correct_channels(sg_pattern.gather_pattern, gather_loc, chan_details_x, chan_details_y, gather_channels);
+                index_to_correct_channels(sg_pattern.scatter_pattern, scatter_loc, chan_details_x, chan_details_y, scatter_channels);
+
+                if (gather_channels.empty() || scatter_channels.empty()) {
+                    continue;
```
AmirhosseinPoolad, line 233: We should probably at least warn the user here that there's an ill formed pattern.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+                                                                   sg_pattern.gather_pattern.from_switchpoint_set,
+                                                                   chan_details_x, chan_details_y,
+                                                                   wire_type_sizes_x, wire_type_sizes_y,
+                                                                   /*is_dest=*/false);
+
+                auto scatter_wire_candidates = find_candidate_wires(scatter_channels,
+                                                                    sg_pattern.scatter_pattern.to_switchpoint_set,
+                                                                    chan_details_x, chan_details_y,
+                                                                    wire_type_sizes_x, wire_type_sizes_y,
+                                                                    /*is_dest=*/true);
```
AmirhosseinPoolad: auto

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -0,0 +1,286 @@
...
+                                                                   formula_data,
+                                                                   sg_pattern.scatter_pattern.num_conns_formula,
+                                                                   gather_wire_candidates.size(),
+                                                                   scatter_wire_candidates.size());
+
+                bottleneck_fanin = std::min<int>(bottleneck_fanin, gather_wire_candidates.size());
+                bottleneck_fanout = std::min<int>(bottleneck_fanout, scatter_wire_candidates.size());
+
+                if (bottleneck_fanin == 0 || bottleneck_fanout == 0) {
+                    continue;
```
AmirhosseinPoolad, line 281: ditto here, maybe we should warn the user. There's a potential problem of printing way too much stuff that should be considered.

vpr/src/route/rr_graph_generation/build_scatter_gathers.h
```diff
@@ -0,0 +1,55 @@
...
+    e_side side;                  ///< Side of the reference switch block the channel lies on
+};
+
+/// Represents a wire candidate for scatter/gather connections.
+struct t_sg_candidate {
+    t_chan_loc chan_loc;                 ///< Channel location (coordinates, type, side) where the wire lies.
+    t_wire_switchpoint wire_switchpoint; ///< Wire index and its valid switchpoint
+};
+
+/// Represents a scatter/gather bottleneck connection between two locations.
```
AmirhosseinPoolad, line 28: I think we referred to this as the "scatter-gather node" in the docs. I don't have an issue with the name bottleneck but I think you should say in the comments that the bottleneck connection is the same thing as scatter-gather node in the docs.

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -1544,7 +1536,7 @@ void alloc_and_load_rr_switch_inf(RRGraphBuilder& rr_graph_builder,
                                   const float R_minW_nmos,
                                   const float R_minW_pmos,
                                   const int wire_to_arch_ipin_switch,
-                                  int* wire_to_rr_ipin_switch) {
+                                  RRSwitchId* wire_to_rr_ipin_switch) {
```
AmirhosseinPoolad: Can you pass this by reference instead?

vtr_flow/arch/multi_die/stratixiv_3d/3d_SB_inter_die_stratixiv_arch.timing.xml
```diff
@@ -48368,4 +48326,30 @@
...
+          <sg_link name="L_UP" z_offset="1" x_offset="0" y_offset="0" mux="seg4_inter_die_driver" seg_type="L4"/>
+          <!-- Link going down one layer, using the 'sw' multiplexer to gather connections from the top layer and using the 'wire' node/wire to down up one layer -->
+          <sg_link name="L_DOWN" z_offset="-1" mux="seg4_inter_die_driver" seg_type="L4"/>
+        </sg_link_list>
+
+        <!-- Instantiate 10 'L_UP' sg_links per switchblock location everywhere on the device -->
+        <sg_location type="EVERYWHERE" num="1" sg_link_name="L_UP"/>
+        <sg_location type="EVERYWHERE" num="1" sg_link_name="L_DOWN"/>
+    </sg_pattern>
+  </scatter_gather_list>
```
AmirhosseinPoolad, line 48354: Comments here should be removed

AmirhosseinPoolad (commented):
And that concludes the review! Thanks for the PR Soheil.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
soheilshahrouz: `clear()` does not free the allocated memory, so the same memory is reused in later calls.  
If instead we create a local vector and return it, new memory will be allocated each time.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
soheilshahrouz: Resolved in #3280

vpr/src/route/rr_graph_generation/rr_graph.cpp
soheilshahrouz: I raised another PR that moves this funtion to another file. Changing the pointer to reference would cause a conflict. I'll change it in the other PR.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
soheilshahrouz: I think the return type would become too complicated.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
soheilshahrouz: I understand that this order is not the most efficient one, but it allows us to support fixed switch point order. When the order is fixed, switch point that are specified first have a higher priority. I'll add a comment to explain this in the code.

vpr/src/route/router_lookahead/router_lookahead_compressed_map.cpp
soheilshahrouz: Changed the code to error out when compressed router lookahead is used on a 3D arhcitecture.

soheilshahrouz:
Titan 2D

| Metric                                                     | master.txt | feature.txt       |
|-------------------------------------------------------------|------------|-------------------|
| vtr_flow_elapsed_time                                      | 1          | 1.01962283887379  |
| max_vpr_mem                                                | 1          | 0.988996538527656 |
| pack_time                                                  | 1          | 1.0159114629532   |
| placed_wirelength_est                                      | 1          | 1                 |
| place_time                                                 | 1          | 1.03434272913768  |
| placed_CPD_est                                             | 1          | 1                 |
| routed_wirelength                                          | 1          | 1                 |
| critical_path_delay                                        | 1          | 1                 |
| crit_path_route_time                                       | 1          | 1.01068496559616  |

soheilshahrouz:
Comparison between 3D and 2D architectures on this branch

Titan other
| Pack Time   | Place Time  | Route Time  | WL          | CPD         |
|-------------|-------------|-------------|-------------|-------------|
| 0.988 | 0.997 | 1.009 | 0.925 | 0.968 |

Titan 23
| Pack Time   | Place Time  | Route Time  | WL          | CPD         |
|-------------|-------------|-------------|-------------|-------------|
| 1.013 | 0.985 | 0.963| 0.898| 0.899|

vaughnbetz:
Looks good!  I'm reviewing now; if you can resolve the conflicts that would be good.

doc/src/arch/reference.rst
```diff
@@ -2001,16 +2001,16 @@ Wire Segments
...
     :opt_param axis:
-        Specifies if the given segment applies to either x or y channels only. If this tag is not given, it is assumed that the given segment
-        description applies to both x-directed and y-directed channels.
+        Specifies if the given segment applies to x, y, or z channels only. If this tag is not given, it is assumed that the given segment
+        description applies to both x-directed and y-directed channels (and not to z-directed channels).
 
         .. note:: It is required that both x and y segment axis details are given or that at least one segment within ``segmentlist`` 
             is specified without the ``axis`` tag (i.e. at least one segment applies to both x-directed and y-directed 
-            chanels). 
+            chanels). For 3-d architectures, it is required that at least one write segment with `axis="z"` is defined.
```
vaughnbetz: write -> wire

libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -441,7 +442,7 @@ void check_rr_node(const RRGraphView& rr_graph,
             break;
 
         case e_rr_type::CHANZ:
-            if (xhigh != xlow || yhigh != ylow || xhigh > grid_width - 1 || ylow < 1 || yhigh > grid_height - 1) {
+            if (xhigh != xlow || yhigh != ylow || xhigh > grid_width - 1 || yhigh > grid_height - 1) {
```
vaughnbetz: Why don't we also check for xlow < 0 and ylow < 0 here?

vaughnbetz: Might also want to relax the CHANX case yhigh > grid_height -2 to grid_height - 1 (allow channels across top) and CHANY case xhigh > grid_width -2 to grid_width -1. Reason: from our discussion about interposer cuts today,  we may want to allow channels to be the same on all tiles, without assuming special I/O or other tiles that don't have routing channels above them.  So relaxing this check moves us a little bit toward that.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -903,13 +914,15 @@ class t_rr_graph_storage {
...
-     *in a separate vector.
-     */
-    vtr::vector<RRNodeId, short> node_layer_;
+    // Layer number refers to the die that the node belongs to.
+    // The layer number of base die is zero and die above it one, etc.
+    // This data is also considered as a hot data since it is used in inner loop of router,
+    // but since it didn't fit nicely into t_rr_node_data due to alignment issues, we had to store it in a separate vector.
+
+    /// @brief The lowest layer number where a given node is located at.
+    vtr::vector<RRNodeId, char> node_layer_low_;
```
vaughnbetz: Does most code get node_layer_low and node_layer_high in sequence (right after the other)?  If so, we'd probably get slightly better performance if we put the two elements in a struct (one char each) and stored a vector of that struct.
Something to consider if it is not too much work ....

libs/librrgraph/src/base/rr_graph_utils.cpp
```diff
@@ -240,8 +244,8 @@ bool inter_layer_connections_limited_to_opin(const RRGraphView& rr_graph) {
     for (const RRNodeId from_node : rr_graph.nodes()) {
         for (t_edge_size edge : rr_graph.edges(from_node)) {
             RRNodeId to_node = rr_graph.edge_sink_node(from_node, edge);
-            int from_layer = rr_graph.node_layer(from_node);
-            int to_layer = rr_graph.node_layer(to_node);
+            int from_layer = rr_graph.node_layer_low(from_node);
+            int to_layer = rr_graph.node_layer_low(to_node);
 
             if (from_layer != to_layer) {
                 if (rr_graph.node_type(from_node) != e_rr_type::OPIN) {
```
vaughnbetz, line 252: Suggest putting a TODO on this OPIN code since I think you're going to redo this later to have OPINs only connect to CHANZ's that span their layer. At that point this code won't be needed, or at least won't work properly.

libs/librrgraph/src/base/rr_spatial_lookup.h
```diff
@@ -80,7 +80,7 @@ class RRSpatialLookup {
      *   @param layer specifies which FPGA die the node is located at (e.g. multi-die(3D) FPGA)
      *   @param (xlow, ylow) is the lower left corner of the grid location range to search within the FPGA
      *   @param (xhigh, yhigh) is the top right corner of the grid location range to search within the FPGA
-     *   @param rr_type specifies the type of resource,
+     *   @param type specifies the type of resource,
```
vaughnbetz, line 83: Routine above also should have its parameter comment updated from rr_type to type.

libs/librrgraph/src/base/rr_spatial_lookup.h
```diff
@@ -102,7 +102,7 @@ class RRSpatialLookup {
      *
      *   @param layer specified which FPGA die the node is located at (e.g. multi-die(3D) FPGA)
      *   @param (x, y) is the coordinate of the routing channel within the FPGA
-     *   @param rr_type specifies the type of routing channel, either x-direction or y-direction
+     *   @param type specifies the type of routing channel, either x-direction or y-direction
      *
      * @note 
      * - Return an empty list if there are no routing channel at the given (layer,x,y) location
```
vaughnbetz, line 108: Does this routine only work for CHANX and CHANY nodes?  If so, we should say that a bit more explicitly. (type must be CHANX or CHANY).  

Do we need this routine, as it is simply a wrapper around find_nodes()?  Now might be a good time to replace it in the code.

libs/librrgraph/src/base/check_rr_graph.cpp
soheilshahrouz: `xlow < 0`, `ylow < 0`, `xhigh > grid_width - 1`, and `yhigh > grid_height - 1` are checked above for all node types, so I'll remove these redundant checks for specific node types.

libs/librrgraph/src/base/rr_graph_utils.cpp
soheilshahrouz: I think we'll need this function, but the implementation would change. We should check whether CHANZ nodes are driven only by OPINs or by CHANX/CHANY wires or a combination of both.

libs/librrgraph/src/base/rr_spatial_lookup.h
soheilshahrouz: It also works for CHANZ nodes.

`find_nodes()` is a private member function. It has an optional `side` argument that must be set to `TOP` when the function is called to find CHANX/CHANY/CHANZ nodes. I think making this method public is a bit dangerous as it allows the used to pass illegal `side` values.

Issue #3275, opened 2025-09-16
[Draw][Place] Manual Moves


AlexandreSinger:
I do not believe that manual moves are working as intended.

Currently, the `try_swap_` method takes a boolean argument which decides whether manual moves are on for the given swap or not. This boolean is only ever set to true during the initial temperature estimation as far as I can tell:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/112d358b0c5a848720b6a514746d8adea157587b/vpr/src/place/annealer.cpp#L312-L321

This happens before any annealing iteration occurs. During the actual inner-loop of the annealer, you can see that this is always set to false:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/112d358b0c5a848720b6a514746d8adea157587b/vpr/src/place/annealer.cpp#L683-L690

I wonder if something was removed from the inner-loop method which no longer allows for manual moves during the actual annealing algorithm.

See #3271 for context.

AlexandreSinger:
@vaughnbetz FYI

vaughnbetz:
Whoops, meant @haydar-c . Thank you for flagging the mistake!

haydar-c:
Sure, closing. Thanks @AlexandreSinger

PR #3278, opened 2025-09-19
[Draw][Place] Fix Manual Moves

haydar-c:
This PR fixes a segmentation fault when using manual moves in placement and enables using manual moves in the inner annealing loop  as well. Previously, manual moves worked only before the inner annealing loop and would segfault, since reward calculation was still invoked even though no move generator was used.

#### Changes

- Guard reward calculation when manual_move_enabled is true
- Add explicit checking of the manual move option inside the inner annealing loop.

#### Related Issue
 https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3275


AlexandreSinger (approved):
LGTM! Thanks!

vaughnbetz:
Just to be paranoid, can you run one design before and after this change (a reasonable size one) and compare CPU time in placement to make sure it doesn't slow down?  I think the overhead of the extra check should be minimal but it is good to be paranoid.

vaughnbetz:
The pause button and the debug buttons basically worked.
One issue with the debug button: breaking at router iteration 3 didn't actually stop at iteration 3 most of the time, although it did say it stopped sometimes. Stopping in placement seemed OK.
==> Debug the router iteration / debug options.

==> Other issue: update the graphics documentation and debug documentation on the vtr website to explain these buttons.

haydar-c:
Adding the largest 2 VTR Chain benchmark place times normalized to without change. They seem negligible.


| Circuit       | Normalized Placement Time (With Change / Without Change) |
|---------------|-------------------------------------------|
| LU32PEEng.v   | 0.9972                                    |
| mcml.v        | 1.0086                                    |


Planning to raise another PR for the debug/breakpoint options and graphics documentation (including updating the start of the graphics documentation).

PR #3280, opened 2025-09-24
Remove vtr::string_view

soheilshahrouz:
Removes vtr::string_view class. All instances are replaced with std::string_view.


libs/libvtrutil/src/vtr_expr_eval.cpp
```diff
@@ -318,7 +318,7 @@ static void get_formula_object(const char* ch, int& ichar, const t_formula_data&
             //A number
             fobj->type = E_FML_NUMBER;
             fobj->data.num = mydata.get_var_value(
-                vtr::string_view(
+                std::string_view(
```
AmirhosseinPoolad: You can pass an std::string to string_view now

libs/libvtrutil/src/vtr_expr_eval.h
```diff
@@ -44,18 +44,10 @@ class t_formula_data {
...
-        return get_var_value(vtr::string_view(var.data(), var.size()));
-    }
+    void set_var_value(std::string_view var, int value) { vars_[var] = value; }
 
     ///@brief get the value of a specific part of the formula (the var can be c-style string)
-    int get_var_value(vtr::string_view var) const {
+    int get_var_value(std::string_view var) const {
         auto iter = vars_.find(var);
         if (iter == vars_.end()) {
             std::string copy(var.data(), var.size());
```
AmirhosseinPoolad: This should also just work by passing var to std::string's constructor.

AmirhosseinPoolad: Nevermind, look at the line below this. You might even be able to avoid doing the copy entirely. Although I'm not sure if the string_view is necessarily null-terminated.

PR #3283, opened 2025-09-26
[Draw][Breakpoint] Fix debug breakpoints to work in placement and routing

haydar-c:
This PR fixes the debug breakpoint so that breakpoints are triggered under the intended conditions in both placement and routing.

- Previously, the router iteration breakpoint was not triggered correctly because the counter was not being updated properly.
- In addition, if placement breakpoints were still active during routing, none of the routing breakpoints would trigger since they weren’t being guarded against.

Both issues are addressed in this PR.

#### Related Issue
See comments in the PR #3278

### DIFF
--- a/vpr/src/draw/breakpoint.cpp
+++ b/vpr/src/draw/breakpoint.cpp
@@ -61,15 +61,15 @@ bool check_for_breakpoints(bool in_placer) {
     //goes through the breakpoints vector
     t_draw_state* draw_state = get_draw_state_vars();
     for (size_t i = 0; i < draw_state->list_of_breakpoints.size(); i++) {
-        if (draw_state->list_of_breakpoints[i].type == BT_MOVE_NUM && draw_state->list_of_breakpoints[i].active)
+        if (draw_state->list_of_breakpoints[i].type == BT_MOVE_NUM && draw_state->list_of_breakpoints[i].active && in_placer)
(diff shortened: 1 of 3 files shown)


AlexandreSinger (commented):
LGTM, so long as the feature works locally!

haydar-c:
Based on comments, I ran the largest 2 VTR benchmarks on my desktop and putting CPU time comparisons here.

### Build: **No Graphics**

| Circuit       | Normalized Placement Time (With Change / Without Change) | Normalized Placement Time (With Change / Without Change) |
|---------------|-----------------------------------------------------------|--------------|
| mcml.v              |  1.001724667  | 1.005574136  |
|LU32PEEng.v    |  1.009246933 | 1.001697073  |


### Build: **With Graphics** (`--disp off`)
| Circuit       | Normalized Placement Time (With Change / Without Change) | Normalized Placement Time (With Change / Without Change) |
|---------------|-----------------------------------------------------------|--------------|
| mcml.v              |  1.008656351  | 1.015185602  |
|LU32PEEng.v    |  1.016414734   | 1.005959983  |

The latest change shows a small increase in runtime, but it’s marginal. I’ll defer to the reviewer on whether we should optimize it for CPU time.

haydar-c:
Yes, the second columns is meant to be normalized routing time. Sorry for the confusion, updated now.

PR #3285, opened 2025-09-29
Feature ap draw

robluo:
Now able to draw analytical placement

#### Description
1. Drawing analytical placement logical location before and after the solving and legalization.
2. No net connection or clickable elements.
3. Can transition to other graphic displays after finishing analytical placement. 

#### Types of changes

#### Checklist:


vpr/src/analytical_place/global_placer.cpp
AlexandreSinger, line 1: I would like to keep the drawing code pretty isolated from the AP code as much as possible. In the future, we may decide to make another global placer as well, and I would like to avoid duplicate code.

I suggest that you create a new class in the `analytical_place` directory for drawing. Something like APDrawManager or something. This class would then have methods for drawing the pre and post legalized placements. This would remove the global accesses from this method. The global placer can then take this as an argument, which it can use to draw whenever it needs.

vpr/src/analytical_place/global_placer.cpp
```diff
@@ -422,6 +440,8 @@ PartialPlacement SimPLGlobalPlacer::place() {
 
         if (hpwl_relative_gap < target_hpwl_relative_gap_)
             break;
+
```
AlexandreSinger: Why add these extra lines?

vpr/src/analytical_place/global_placer.cpp
```diff
@@ -352,6 +356,10 @@ PartialPlacement SimPLGlobalPlacer::place() {
     PartialPlacement best_p_placement(ap_netlist_);
     double best_ub_hpwl = std::numeric_limits<double>::max();
 
+#ifndef NO_GRAPHICS
```
AlexandreSinger: See my comments above. I would like to keep these ifdefs and the draw state out of this class.

vpr/src/draw/draw.cpp
```diff
@@ -302,16 +306,30 @@ void update_screen(ScreenUpdatePriority priority, const char* msg, enum pic_type
...
+            if (canvas != nullptr) {
+                canvas->get_camera().set_world(initial_world);
+            }
         }
 
         draw_state->setup_timing_info = setup_timing_info;
         draw_state->pic_on_screen = pic_on_screen_val;
     }
 
+    // What is this? Always true!
```
AlexandreSinger: I would mark this as a FIXME, or raise an issue so we can resolve. A comment like this will get missed.

vpr/src/draw/draw.cpp
```diff
@@ -481,7 +499,18 @@ void init_draw_coords(float clb_width, const BlkLocRegistry& blk_loc_registry) {
     //Margin beyond edge of the drawn device to extend the visible world
     //Setting this to > 0.0 means 'Zoom Fit' leave some fraction of white
     //space around the device edges
+#else
+    (void)clb_width;
+    (void)blk_loc_registry;
+#endif /* NO_GRAPHICS */
+}
+
+#ifndef NO_GRAPHICS
```
AlexandreSinger, line 517: I would be careful with this ifdef. You are not defining these functions when graphics is disabled. However, they are still available in the header file...

I would put the ifdef inside of the method and have the method do nothing when graphics is not enabled.

vpr/src/draw/draw.h
```diff
@@ -52,6 +52,9 @@ void update_screen(ScreenUpdatePriority priority, const char* msg, enum pic_type
  */
 void init_draw_coords(float clb_width, const BlkLocRegistry& blk_loc_registry);
 
+void set_initial_world_ap();
```
AlexandreSinger, line 63: These new methods need comments. Also see my prior comments on their definitions.

vpr/src/draw/draw_basic.h
```diff
@@ -29,6 +29,8 @@
  * Blocks are drawn in layer order (so that semi-transparent blocks/grids render well)*/
 void drawplace(ezgl::renderer* g);
 
+void draw_analytical_place(ezgl::renderer* g);
```
AlexandreSinger, line 35: Needs comment.

vpr/src/draw/draw_basic.cpp
```diff
@@ -193,6 +192,60 @@ void drawplace(ezgl::renderer* g) {
...
+    for (int layer = 0; layer < total_layers; ++layer) {
+        const auto& layer_disp = draw_state->draw_layer_display[layer];
+        if (!layer_disp.visible) continue;
+
+        for (int x = 0; x < (int)device_ctx.grid.width(); ++x) {
+            for (int y = 0; y < (int)device_ctx.grid.height(); ++y) {
+                // Only draw at the root of a non-unit tile
+                int w_off = device_ctx.grid.get_width_offset({x, y, layer});
+                int h_off = device_ctx.grid.get_height_offset({x, y, layer});
+                if (w_off > 0 || h_off > 0) continue;
```
soheilshahrouz: There is a method in DeviceGrid to check if a given location is at the root of block

vpr/src/analytical_place/analytical_draw_manager.h
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @class AnalyticalDrawManager
+ * @brief Manages graphics updates during analytical placement operations.
+ *
+ * This class provides a clean interface for updating the screen during
+ * analytical placement without requiring the placement code to be littered
+ * with NO_GRAPHICS conditional compilation directives.
+ */
+class AnalyticalDrawManager {
```
AlexandreSinger: Minor nitpick: I would call this APDrawManager instead. The reason I say this is that this will be more tricky to change later, and "AnalyticalDrawManager" makes me think that we are using analytical techniques to improve drawing.

vpr/src/analytical_place/global_placer.cpp
```diff
@@ -11,13 +11,18 @@
...
 #include "analytical_solver.h"
 #include "ap_flow_enums.h"
 #include "ap_netlist.h"
 #include "ap_netlist_fwd.h"
 #include "atom_netlist.h"
 #include "device_grid.h"
+#include "draw.h"
+#ifndef NO_GRAPHICS
+#include "draw_global.h"
+#endif
```
AlexandreSinger: I think you can remove these 4 includes now.

vpr/src/analytical_place/global_placer.cpp
```diff
@@ -361,12 +370,18 @@ PartialPlacement SimPLGlobalPlacer::place() {
         solver_->solve(i, p_placement);
         float solver_end_time = runtime_timer.elapsed_sec();
         double lb_hpwl = p_placement.get_hpwl(ap_netlist_);
+        
+        // Update graphics after analytical solver
+        draw_manager.update_graphics("Iteration " + std::to_string(i) + " After Solver");
```
AlexandreSinger: Since we have this class now, I recommend that you move this logic into the draw manager itself! That way we only pass in the iteration. This will absorb the implementation details of drawing away from the logic of the global placer.

It will also remove some of the duplicate code below.

AlexandreSinger:
@robluo Your PR broke CI. This can happen when there is a merge conflict post-the-fact. When you rebased your branch, you may have rebased onto an older version of VTR! That can cause some funny stuff like that to occur.

I had to revert your PR to keep master working. You will need to re-raise this code in another PR. Sorry for the pain.

Issue #3284, opened 2025-09-26
Place and route graphical breakpoints updates


vaughnbetz:
#### Expected Behaviour
route_net_id >= 0 crashes with an unknown expression error if route_net_id is misspelt (e.g. route_net_ids)..
  - should give an error message and proceed if we can't parse a formula, instead of crashing
  
Setting a router breakpoint every N iterations leads to next stopping after every net
   - Should just stop next time a breakpoint condition is met, which is at the start of that routing iteration
   
route_net_id is misspelt in the pop-up when you stop (fix text)

PR #3286, opened 2025-09-29
[Draw][Breakpoints] Place and route graphical breakpoints updates

haydar-c:
The updates in this PR are:

- Added early validation of breakpoint expressions when the user sets them. Invalid expressions are now rejected immediately with an “Invalid Expression” message, preventing crashes later.
- Fixed router iteration breakpoints: previously, when a router_iter condition was met, the debugger stopped on every net in that iteration. Now, it stops once per router iteration, and will only stop again when the condition is met in a future iteration.
- Fixed a typo in the breakpoint pop-up message.

**Related Issue**
#3284


vaughnbetz (approved):
LGTM. I assume the user gets a message of some sort if he/she types in an invalid expression -- is that correct?

vaughnbetz:
Thanks!

PR #3287, opened 2025-09-30
Add check that model exists during read_blif to avoid seg fault

gigeresk:
As described in title. This offers feedback when feeding vtr malformed inputs.


AlexandreSinger (approved):
LGTM! Thanks @gigeresk

PR #3281, opened 2025-09-25
Refactor: Move intra-cluster RR graph code into dedicated files

soheilshahrouz:
Moves all functions that create nodes and edges for intra-cluster routing resources to separate files.

PR #3289, opened 2025-09-30
Address remaining comments from #3281

soheilshahrouz:
Addresses Vaughn's comments from #3281

### DIFF
--- a/libs/libarchfpga/src/physical_types.h
+++ b/libs/libarchfpga/src/physical_types.h
@@ -1532,9 +1532,7 @@ struct t_pb_graph_pin_power {
 /* FPGA Routing architecture                                                                     */
 /*************************************************************************************************/
 
-/* Description of routing channel distribution across the FPGA, only available for global routing
- * Width is standard dev. for Gaussian.  xpeak is where peak     *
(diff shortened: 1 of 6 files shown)


AlexandreSinger (approved):
LGTM

libs/libarchfpga/src/physical_types.h
```diff
@@ -1532,22 +1532,21 @@ struct t_pb_graph_pin_power {
...
- * occurs. dc is the dc offset for Gaussian and pulse waveforms. */
+/// @brief Description of routing channel distribution across the FPGA, only available for global routing
 enum class e_stat {
     UNIFORM,
     GAUSSIAN,
     PULSE,
     DELTA
 };
 
+/// @brief Parameters describing a channel distribution.
```
vaughnbetz, line 1543: Please add:
/// If detailed routing is performed, only a uniform (all channels in a given direction are the same width) distribution is supported.

Under peak, add:
/// Peak value. For a UNIFORM distribution, this is the value for all channels (in a given direction).

vaughnbetz (approved):
I suggest one minor addition; after that it's good to merge.

PR #3293, opened 2025-10-03
[Draw] Remove pause button

haydar-c:
This PR removes the Pause button from the GUI. The button only had an effect if VPR was already paused (for example, at a breakpoint, at the end of a stage, or due to a previous pause).

**Changes**

- Remove pause button from GUI and related code.
- Restore writing of the route checkpoint file when the program is terminated.
- Updated outdated screenshots in the documentation for the menu bar and Misc tab sections.


haydar-c:
The button seems to be working after the changes of @soheilshahrouz in the PR's
#3295 and #3290, thank you Soheil!

Planning to close this PR and raise a new one for the old menu bar update and writing to checkpoint file when stopped with signal handler enabled.

PR #3297, opened 2025-10-08
Draw nodes on top of edges

soheilshahrouz:
This PR updates the drawing order of edges and nodes. All edges are now drawn first, followed by nodes, ensuring that nodes (and their muxes) appear on top of edges. Previously, some incoming edges of a mux were rendered both beneath and above the mux.


soheilshahrouz:
Screenshot from 2025-10-08 12-18-51

PR #3299, opened 2025-10-08
Draw interposer cuts

soheilshahrouz:
This PR updates the graphics code to draw interposer cut lines.

### DIFF
--- a/libs/libarchfpga/src/device_grid.h
+++ b/libs/libarchfpga/src/device_grid.h
@@ -197,12 +197,16 @@ class DeviceGrid {
         return &grid_.get(n);
     }
 
-    /// Returns the list of horizontal interposer cut locations for each layer.
+    /// Returns the list of horizontal interposer cut locations for each layer,
+    /// i.e. y value of the tile row just below each cut
+    /// Accessed as [layer][cut_idx]
     inline const std::vector<std::vector<int>>& get_horizontal_interposer_cuts() const {
         return horizontal_interposer_cuts_;
     }
 
-    /// Returns the list of vertical interposer cut locations for each layer.
+    /// Returns the list of vertical interposer cut locations for each layer,
+    /// i.e. x value of the tile column just to the left each cut
+    /// Accessed as [layer][cut_idx]
     inline const std::vector<std::vector<int>>& get_vertical_interposer_cuts() const {
         return vertical_interposer_cuts_;
     }
@@ -229,8 +233,12 @@ class DeviceGrid {
 
     std::vector<t_logical_block_type_ptr> limiting_resources_;
 
-    /// Horizontal interposer cut locations in each layer.
+    /// Horizontal interposer cut locations in each layer,
+    /// i.e. y value of the tile row just below each cut
+    /// Accessed as [layer][cut_idx]
     std::vector<std::vector<int>> horizontal_interposer_cuts_;
-    /// Vertical interposer cust location in each layer.
+    /// Vertical interposer cut location in each layer,
+    /// i.e. x value of the tile column just to the left each cut
+    /// Accessed as [layer][cut_idx]
     std::vector<std::vector<int>> vertical_interposer_cuts_;
 };
--- a/vpr/src/draw/draw_interposer.cpp
+++ b/vpr/src/draw/draw_interposer.cpp
@@ -43,8 +43,12 @@ void draw_interposer_cuts(ezgl::renderer* g) {
         for (int cut_y : horizontal_cuts[layer]) {
             float y;
             if (draw_state->pic_on_screen == e_pic_type::PLACEMENT) {
+                // Draw the interposer line exactly in the middle of routing channel.
                 y = (draw_coords->tile_y[cut_y + 1] + draw_coords->tile_y[cut_y] + draw_coords->get_tile_height()) / 2.0f;
             } else if (draw_state->pic_on_screen == e_pic_type::ROUTING) {
+                // Draw the interposer above the last track and below the next tile.
+                // The row below the cut owns the channel, so the cut is drawn closer
+                // to the row above the cut to make this ownership explicit.
                 y = draw_coords->tile_y[cut_y + 1] - 0.5f;
             } else {
                 VTR_ASSERT(false);
@@ -56,8 +60,12 @@ void draw_interposer_cuts(ezgl::renderer* g) {
         for (int cut_x : vertical_cuts[layer]) {
             float x;
             if (draw_state->pic_on_screen == e_pic_type::PLACEMENT) {
+                // Draw the interposer line exactly in the middle of routing channel.
                 x = (draw_coords->tile_x[cut_x + 1] + draw_coords->tile_x[cut_x] + draw_coords->get_tile_width()) / 2.0f;
             } else if (draw_state->pic_on_screen == e_pic_type::ROUTING) {
+                // Draw the interposer line between the last track and the next column.
+                // The channel belongs to the column on the left of the cut, so the cut is
+                // positioned closer to the column on the right to make this ownership clear.
                 x = draw_coords->tile_x[cut_x + 1] - 0.5f;
             } else {
                 VTR_ASSERT(false);


doc/src/arch/reference.rst
```diff
@@ -582,10 +582,12 @@ Grid Layout Example
     Example FPGA grid
 
 
-.. arch:tag:: <interposer_cut dim=x|y loc="int"/>
+.. arch:tag:: <interposer_cut x="int" y="int"/>
 
-    :req_param dim: Dimension or axis of the cut. 'X' or 'x' means a horizontal cut while 'Y' or 'y' means a vertical cut.
-    :req_param loc: Location of the cut. Cuts are done above or to the right of the tiles at coordinate 'loc'. For example a cut with dim=x and loc=0 would cut the vertical wires above tiles in the 0th row. Currently only integer values are supported.
+    :opt_param x: Specifies the x-coordinate of a vertical interposer cut.
+    :opt_param y: Specifies the y-coordinate of a horizontal interposer cut.opt_param
```
AmirhosseinPoolad: extra "opt_param" at the end

libs/libarchfpga/src/interposer_types.h
```diff
@@ -4,33 +4,20 @@
...
- * @brief Enum for direction of an interposer cut. X means horizontal cut and Y means vertical cut.
- * 
- */
-enum class e_interposer_cut_dim {
-    X,
-    Y
+/// @brief Enum for direction of an interposer cut.
+enum class e_interposer_cut_type {
+    HORZ,
+    VERT
```
AmirhosseinPoolad, line 16: I would prefer if it was simply HORIZONTAL and VERTICAL. But it's not a strong preference, feel free to leave it as is.

libs/libarchfpga/src/read_xml_arch_file_interposer.cpp
```diff
@@ -8,36 +10,48 @@
...
+    if (x != ARCH_FPGA_UNDEFINED_VAL) {
+        interposer.loc = x;
+        interposer.dim = e_interposer_cut_type::VERT;
+    } else if (y != ARCH_FPGA_UNDEFINED_VAL) {
+        interposer.loc = y;
+        interposer.dim = e_interposer_cut_type::HORZ;
+    } else {
+        archfpga_throw(loc_data.filename_c_str(), loc_data.line(interposer_cut_tag),
+                       "Interposer cut tag must specify where the cut is to appear using `x` or `y` attributes.");
+    }
```
AmirhosseinPoolad, line 33: This doesn't check if both x and y are defined simultaneously.

libs/libarchfpga/src/read_xml_arch_file_interposer.cpp
```diff
@@ -8,36 +10,48 @@
...
         t_interdie_wire_inf interdie_wire;
 
-        interdie_wire.sg_name = pugiutil::get_attribute(interdie_wire_tag, "sg_name", loc_data).as_string();
-        interdie_wire.sg_link = pugiutil::get_attribute(interdie_wire_tag, "sg_link", loc_data).as_string();
+        interdie_wire.sg_name = pugiutil::get_attribute(interdie_wire_tag, "sg_name", loc_data).as_string(nullptr);
+        interdie_wire.sg_link = pugiutil::get_attribute(interdie_wire_tag, "sg_link", loc_data).as_string(nullptr);
+
+        interdie_wire.offset_definition.start_expr = pugiutil::get_attribute(interdie_wire_tag, "offset_start", loc_data).as_string(nullptr);
+        interdie_wire.offset_definition.end_expr = pugiutil::get_attribute(interdie_wire_tag, "offset_end", loc_data).as_string(nullptr);
+        interdie_wire.offset_definition.incr_expr = pugiutil::get_attribute(interdie_wire_tag, "offset_increment", loc_data).as_string(nullptr);
```
AmirhosseinPoolad: This is actually more unsafe than before. xml_attribute::as_string without the optional argument will return an empty string, not a nullptr. The C++ standard requires the input character array to std::string's constructor to not be null and it is undefined behaviour in C++ < 23.  in C++23 and newer, basic_string( std::nullptr_t ) is explicitly deleted.

libs/libarchfpga/src/read_xml_arch_file_interposer.cpp
```diff
@@ -8,36 +10,48 @@
...
 
-        interdie_wire.offset_definition.start_expr = pugiutil::get_attribute(interdie_wire_tag, "offset_start", loc_data).as_string();
-        interdie_wire.offset_definition.end_expr = pugiutil::get_attribute(interdie_wire_tag, "offset_end", loc_data).as_string();
-        interdie_wire.offset_definition.incr_expr = pugiutil::get_attribute(interdie_wire_tag, "offset_increment", loc_data).as_string();
+        interdie_wire.num = pugiutil::get_attribute(interdie_wire_tag, "num", loc_data).as_int(ARCH_FPGA_UNDEFINED_VAL);
```
AmirhosseinPoolad: Also do note that pugiutil::get_attribute by default will check for attribute existence and will error out if the attribute doesn't exist, which makes the default values unnecessary.

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,55 @@
...
+    const DeviceGrid& grid = device_ctx.grid;
+    t_draw_coords* draw_coords = get_draw_coords_vars();
+
+    const std::vector<std::vector<int>>& horizontal_cuts = grid.get_horizontal_interposer_cuts();
+    const std::vector<std::vector<int>>& vertical_cuts = grid.get_vertical_interposer_cuts();
+
+    std::vector<std::pair<ezgl::point2d, ezgl::point2d>> lines_to_draw;
+
+    const ezgl::rectangle world{{draw_coords->tile_x.front() - draw_coords->get_tile_width(), draw_coords->tile_y.front() - draw_coords->get_tile_height()},
+                                {draw_coords->tile_x.back() + 2 * draw_coords->get_tile_width(), draw_coords->tile_y.back() + 2 * draw_coords->get_tile_height()}};
```
AmirhosseinPoolad: Should add comment explaining the formulas

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    g->set_color(ezgl::BLACK, 255);
+    g->set_line_dash(ezgl::line_dash::asymmetric_5_3);
+    g->set_line_width(2);
+
+    const float offset_factor = draw_state->pic_on_screen == e_pic_type::PLACEMENT ? -0.5f : -0.5f / device_ctx.chan_width.max;
```
AmirhosseinPoolad: Add comment explaining why this exists.

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,55 @@
...
+#ifndef NO_GRAPHICS
+
+#include "draw_interposer.h"
+
+#include "draw_global.h"
+#include "vtr_assert.h"
+#include "globals.h"
+
+void draw_interposer_cuts(ezgl::renderer* g) {
+    t_draw_state* draw_state = get_draw_state_vars();
```
AmirhosseinPoolad: Should be marked as const

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,55 @@
...
+    const DeviceContext& device_ctx = g_vpr_ctx.device();
+    const DeviceGrid& grid = device_ctx.grid;
+    t_draw_coords* draw_coords = get_draw_coords_vars();
```
AmirhosseinPoolad: ditto

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    if (draw_state->pic_on_screen == e_pic_type::PLACEMENT || draw_state->pic_on_screen == e_pic_type::ROUTING) {
+        for (const auto& [start, end] : lines_to_draw) {
+            g->draw_line(start, end);
+        }
+    } else {
+        VTR_ASSERT(draw_state->pic_on_screen == e_pic_type::NO_PICTURE);
+    }
```
AmirhosseinPoolad: I think you should check this before you do all the calculations for drawing.

libs/libarchfpga/src/device_grid.h
```diff
@@ -208,4 +226,7 @@ class DeviceGrid {
     std::vector<std::map<t_physical_tile_type_ptr, size_t>> instance_counts_; /* [layer_num][physical_tile_type_ptr] */
 
     std::vector<t_logical_block_type_ptr> limiting_resources_;
+
+    std::vector<std::vector<int>> horizontal_interposer_cuts_;
```
vaughnbetz, line 239: Doxygen comments

libs/libarchfpga/src/device_grid.h
```diff
@@ -189,15 +197,25 @@ class DeviceGrid {
         return &grid_.get(n);
     }
 
+    inline const std::vector<std::vector<int>>& get_horizontal_interposer_cuts() const {
```
vaughnbetz, line 203: Doxygen comment

libs/libarchfpga/src/device_grid.h
```diff
@@ -189,15 +197,25 @@ class DeviceGrid {
...
+        return horizontal_interposer_cuts_;
+    }
+
+    inline const std::vector<std::vector<int>>& get_vertical_interposer_cuts() const {
```
vaughnbetz, line 210: Doxygen comment

libs/libarchfpga/src/device_grid.h
```diff
@@ -208,4 +226,7 @@ class DeviceGrid {
     std::vector<std::map<t_physical_tile_type_ptr, size_t>> instance_counts_; /* [layer_num][physical_tile_type_ptr] */
 
     std::vector<t_logical_block_type_ptr> limiting_resources_;
```
vaughnbetz, line 234: Not your code but comment this one too if you understand it.

vpr/src/draw/draw.h
```diff
@@ -40,7 +40,10 @@ extern ezgl::application application;
 
 #endif /* NO_GRAPHICS */
 
-void update_screen(ScreenUpdatePriority priority, const char* msg, enum pic_type pic_on_screen_val, std::shared_ptr<const SetupTimingInfo> timing_info);
+void update_screen(ScreenUpdatePriority priority,
```
vaughnbetz, line 43: Not your code, but this is a good oppportunity to add a Doxygen comment on this one. Should explain all the parameters.

vpr/src/draw/draw_interposer.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#ifndef NO_GRAPHICS
```
vaughnbetz, line 11: Add a Doxygen comment about what should go in this file.

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,64 @@
...
+    const std::vector<std::vector<int>>& vertical_cuts = grid.get_vertical_interposer_cuts();
+    std::vector<std::pair<ezgl::point2d, ezgl::point2d>> lines_to_draw;
+
+    for (size_t layer = 0; layer < grid.get_num_layers(); layer++) {
+        if (!draw_state->draw_layer_display[layer].visible) {
+            continue;
+        }
+
+        for (int cut_y : horizontal_cuts[layer]) {
+            float y = draw_coords->tile_y[cut_y + 1] + offset_factor * draw_coords->get_tile_height();
```
vaughnbetz: For routing, wouldn't it be safer to draw this 0.5 units (tracks) below the next tile?  We don't know that the number of tracks per channel and the tile height or width will be exactly correlated (although they are currently correlated some as we set the tile width and height to be some multiple of the typical or max channel width I think).
That seems cleaner than the current code, which shifts down by 0.5 / chan_width_max * tile_height (a bit more random where this winds up).
Placement also seems a bit risky; we could instead just compute the middle of the channel, or half-way between the bottom of the tile above and the top of the tile below.

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,64 @@
...
+            lines_to_draw.push_back({{world.left(), y}, {world.right(), y}});
+        }
+
+        for (int cut_x : vertical_cuts[layer]) {
+            float x = draw_coords->tile_x[cut_x + 1] + offset_factor * draw_coords->get_tile_width();
+            lines_to_draw.push_back({{x, world.bottom()}, {x, world.top()}});
```
vaughnbetz, line 74: Same comment as for the horiz cuts.

libs/libarchfpga/src/device_grid.h
vaughnbetz: I'd say exactly what they are: y value of the tile row just below the cut. Accessed as [layer][cut_idx].

libs/libarchfpga/src/device_grid.h
```diff
@@ -208,4 +228,9 @@ class DeviceGrid {
...
+
+    /// Horizontal interposer cut locations in each layer.
+    std::vector<std::vector<int>> horizontal_interposer_cuts_;
+    /// Vertical interposer cust location in each layer.
+    std::vector<std::vector<int>> vertical_interposer_cuts_;
```
vaughnbetz, line 243: I'd say exactly what they are: x value of the tile column just to the left of each cut.

cust -> cuts

vaughnbetz: Accessed as [layer][cut_idx].

vaughnbetz: Probably should give similar levels of detail for the accessor functions too.

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::vector<std::pair<ezgl::point2d, ezgl::point2d>> lines_to_draw;
+
+    for (size_t layer = 0; layer < grid.get_num_layers(); layer++) {
+        if (!draw_state->draw_layer_display[layer].visible) {
+            continue;
+        }
+
+        for (int cut_y : horizontal_cuts[layer]) {
+            float y;
+            if (draw_state->pic_on_screen == e_pic_type::PLACEMENT) {
```
vaughnbetz, line 45: Should put a brief comment here of why you're doing this / where this positions the lines.

vpr/src/draw/draw_interposer.cpp
```diff
@@ -0,0 +1,75 @@
...
+            } else {
+                VTR_ASSERT(false);
+            }
+
+            lines_to_draw.push_back({{world.left(), y}, {world.right(), y}});
+        }
+
+        for (int cut_x : vertical_cuts[layer]) {
+            float x;
+            if (draw_state->pic_on_screen == e_pic_type::PLACEMENT) {
```
vaughnbetz, line 62: Should put a brief comment here of why you're doing this / where this positions the lines. (or put the comment above in a spot where it applies to both)

PR #3301, opened 2025-10-09
Fix GUI message consistency, restore signal-handler checkpointing, and update documentation image

haydar-c:
Changes:
- Previously, GUI messages (e.g., during SA placement or routing) were only refreshed upon user interaction or stage changes (being stale 1 step). Bringing back the message update on stage.
- The signal handler checkpointing logic had been non-functional and giving compilation errors; restoring its functionality when VPR_USE_SIGNAL_HANDLER is set.
- Updated the old menu bar image

Overall_view


vpr/src/base/vpr_signal_handler.cpp
```diff
@@ -92,7 +78,10 @@ void checkpoint() {
     VTR_LOG("Attempting to checkpoint current placement to file: %s\n", placer_checkpoint_file.c_str());
     print_place(nullptr, nullptr, placer_checkpoint_file.c_str(), g_vpr_ctx.placement().block_locs());
```
soheilshahrouz: In the placement stage, g_vpr_ctx.placement().block_locs() does not contain block locations. It is stored locally withing a Placer object.

vpr/src/base/vpr_signal_handler.cpp
```diff
@@ -44,23 +44,9 @@ void vpr_install_signal_handler() {
...
-        } else if (uncleared_sigint_count == 2) {
-            VTR_LOG("Recieved two uncleared SIGINTs: Attempting to checkpoint and exit...\n");
-            checkpoint();
-            std::quick_exit(INTERRUPTED_EXIT_CODE);
-        } else if (uncleared_sigint_count == 3) {
-            //Really exit (e.g. SIGINT while checkpointing)
-            VTR_LOG("Recieved three uncleared SIGINTs: Exiting...\n");
-            std::quick_exit(INTERRUPTED_EXIT_CODE);
-        }
+        VTR_LOG("Recieved SIGINT: Attempting to checkpoint then exit...\n");
```
soheilshahrouz: I’m not sure if it’s safe to call VTR_LOG from a signal handler. It probably calls standard library functions internally, which are not safe to use in a signal handler.

vpr/src/base/vpr_signal_handler.cpp
```diff
@@ -27,7 +29,9 @@
 void vpr_signal_handler(int signal);
 void checkpoint();
 
-std::atomic<int> uncleared_sigint_count(0);
+static inline void safe_write(const char* msg) {
```
vaughnbetz, line 39: Need a doxygen comment; what is this for?

vaughnbetz (approved):
LGTM

PR #3302, opened 2025-10-09
Extract channel width from RR graph

soheilshahrouz:
This PR computes channel widths directly from the RR graph. These computed values now replace t_chan_width usages outside of the RR graph generation.


vpr/src/base/stats.cpp
```diff
@@ -223,66 +188,94 @@ void length_and_bends_stats(const Netlist<>& net_list, bool is_flat) {
 static void get_channel_occupancy_stats(const Netlist<>& net_list, bool /***/) {
     const auto& device_ctx = g_vpr_ctx.device();
 
-    auto chanx_occ = vtr::Matrix<int>({{
-                                          device_ctx.grid.width(),     //[0 .. device_ctx.grid.width() - 1] (length of x channel)
-                                          device_ctx.grid.height() - 1 //[0 .. device_ctx.grid.height() - 2] (# x channels)
-                                      }},
-                                      0);
+    auto chanx_occ = vtr::NdMatrix<int, 3>({{
```
amin1377, line 191: I’d suggest adding a comment explaining that the reason the y-dimension for x-channels and the x-dimension for y-channels are one unit less than the device height/width is that there are no routing channels on the device perimeter.

vpr/src/base/stats.cpp
```diff
@@ -223,66 +188,94 @@ void length_and_bends_stats(const Netlist<>& net_list, bool is_flat) {
...
-        int max_occ = -1;
-
-        for (size_t i = 1; i < device_ctx.grid.width(); ++i) {
-            max_occ = std::max(chanx_occ[i][j], max_occ);
-            ave_occ += chanx_occ[i][j];
+    VTR_LOG("X - Directed channels: layer   j   max occ   ave occ   ave cap\n");
+    VTR_LOG("                        ----- ---- -------- -------- --------\n");
+
+    for (size_t layer = 0; layer < device_ctx.grid.get_num_layers(); ++layer) {
+        for (size_t j = 0; j < device_ctx.grid.height() - 1; ++j) {
```
amin1377: I think it would be better to use i to iterate over rows (up to height) and j to iterate over columns (up to width) to stay consistent with standard matrix notation.

soheilshahrouz: In RR graph generation code, `i` is always used for columns while `j` is used for columns. I think that this convention can lead to confusion, so I'll replace `i` and `j` with `x` and `y`.

vpr/src/base/stats.cpp
```diff
@@ -220,69 +185,97 @@ void length_and_bends_stats(const Netlist<>& net_list, bool is_flat) {
...
+    VTR_LOG("X - Directed channels: layer   y   max occ   ave occ   ave cap\n");
+    VTR_LOG("                        ----- ---- -------- -------- --------\n");
+
+    for (size_t layer = 0; layer < device_ctx.grid.get_num_layers(); ++layer) {
+        for (size_t y = 0; y < device_ctx.grid.height() - 1; y++) {
+            float ave_occ = 0.0f;
+            float ave_cap = 0.0f;
+            int max_occ = -1;
+
+            for (size_t x = 1; x < device_ctx.grid.width(); x++) {
```
vaughnbetz, line 226: Pre-existing code, but do we still not build x-directed routing channels at x = 0 and x = width?  Seems to be an assumption (an old one) that there are I/Os around the edges and we have no wires there.  Can you check if that's still the case.  If it is, we can leave it and comment it here.
It would be a good cleanup someday to get rid of that assumption though ...

vpr/src/base/stats.cpp
```diff
@@ -57,7 +57,7 @@ static void write_channel_occupancy_table(const std::string_view filename,
 static void length_and_bends_stats(const Netlist<>& net_list, bool is_flat);
 
 ///@brief Determines how many tracks are used in each channel.
-static void get_channel_occupancy_stats(const Netlist<>& net_list, bool /***/);
+static void get_channel_occupancy_stats(const Netlist<>& net_list);
```
vaughnbetz, line 60: Should probably say "and prints out statistics" since otherwise this routine doesn't return anything / do anything.

vpr/src/base/stats.cpp
```diff
@@ -293,64 +286,72 @@ static void write_channel_occupancy_table(const std::string_view filename,
...
                 for (int i = rr_graph.node_xlow(inode); i <= rr_graph.node_xhigh(inode); i++)
-                    chanx_occ[i][j]++;
+                    chanx_occ[layer][i][j]++;
             } else if (rr_type == e_rr_type::CHANY) {
                 int i = rr_graph.node_xlow(inode);
+                int layer = rr_graph.node_layer_low(inode);
                 for (int j = rr_graph.node_ylow(inode); j <= rr_graph.node_yhigh(inode); j++)
-                    chany_occ[i][j]++;
+                    chany_occ[layer][i][j]++;
             }
```
vaughnbetz, line 357: Should we update these routines to calculate and print z-directed channel occupancies for 3D FPGAs?

vpr/src/base/vpr_context.h
```diff
@@ -263,6 +263,14 @@ struct DeviceContext : public Context {
...
     int delayless_switch_idx = UNDEFINED;
 
+    /// Stores the number of CHANX wire segments in each routing channel at [layer][x][y]
+    vtr::NdMatrix<int, 3> rr_chanx_width;
+    /// Stores the number of CHANY wire segments in each routing channel at [layer][x][y]
+    vtr::NdMatrix<int, 3> rr_chany_width;
+
+    std::vector<int> rr_chanx_list;
+    std::vector<int> rr_chany_list;
+
```
vaughnbetz, line 275: Should we move a similar structure here for z?  I think there is a similar calculation buried in the placer cost function set up right now.

vpr/src/base/vpr_context.h
```diff
@@ -263,6 +263,14 @@ struct DeviceContext : public Context {
 
     int delayless_switch_idx = UNDEFINED;
 
+    /// Stores the number of CHANX wire segments in each routing channel at [layer][x][y]
+    vtr::NdMatrix<int, 3> rr_chanx_width;
+    /// Stores the number of CHANY wire segments in each routing channel at [layer][x][y]
+    vtr::NdMatrix<int, 3> rr_chany_width;
+
+    std::vector<int> rr_chanx_list;
```
vaughnbetz: Comment what these lists are.

vaughnbetz: I think they are channel widths, in which case rr_chanx_width and rr_chany_width would be better names.

vpr/src/base/vpr_context.h
```diff
@@ -263,6 +263,14 @@ struct DeviceContext : public Context {
 
     int delayless_switch_idx = UNDEFINED;
 
+    /// Stores the number of CHANX wire segments in each routing channel at [layer][x][y]
```
vaughnbetz: routing channel -> portion of a routing channel
(or you could say "routing channel segment"

vpr/src/base/vpr_context.h
soheilshahrouz: `rr_chanx_width` and `rr_chany_width` are already used. I'll rename them to `rr_chany_segment_width` and `rr_chanx_segment_width` so I can use the suggested names for these vectors.

vpr/src/base/stats.cpp
soheilshahrouz: CHANX/CHANY are not created at x/y=0  and x=W-1/y=H-1

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/269e85c8ca3bef62f532338a8b76788bd2144696/vpr/src/route/rr_graph_generation/rr_graph.cpp#L1521-L1539

vpr/src/base/stats.cpp
soheilshahrouz: I'll do it in another PR after I implement 3d opin connectivity though CHANZ.

vaughnbetz:
Weird, but can happen. Could change them to run at a fixed channel width if you're pretty sure all is OK.

soheilshahrouz:
> Weird, but can happen. Could change them to run at a fixed channel width if you're pretty sure all is OK.

I changed the seed to resolve this.

soheilshahrouz:
Nightly test 1, which was the only failing nightly test now passes.

@amin1377 If you don't have further comments, could you please approve this PR"?

amin1377 (approved):

PR #3304, opened 2025-10-12
[AP][DP] Undid the Initial T Scaling

AlexandreSinger:
Originally in AP, we simply scaled the initial temperature relative to the temperature produced by the baseline flow. Now that we moved to a better initial temperature estimator, this is no longer necessary. Removed this for now.

It may still be worth investigating this, but I do not think a simple scalaing is sufficient for AP using the new equilibrium initial temperature estimator.

NOTE: In all prior results that I showed for the equilibrium initial temperature estimator, I did remove this line; I just forgot to push it in my last PR.


vaughnbetz (approved):
LGTM

PR #3305, opened 2025-10-14
Add interposer wires 

soheilshahrouz:
This PR adds support for inter-die wires defined using the `<interdie_wire>` tag under `<interposer_cut>` in the architecture file. The graphics code has been updated accordingly to visualize these wires correctly.

### DIFF
--- a/vpr/src/draw/draw_rr_edges.cpp
+++ b/vpr/src/draw/draw_rr_edges.cpp
@@ -60,9 +60,9 @@ void draw_chany_to_chany_edge(RRNodeId from_node, RRNodeId to_node, RRSwitchId r
                     y2 = to_chan.bottom();
                     y1 = draw_coords->tile_y[to_ylow] + 0.1 * draw_coords->get_tile_height();
                 } else {
+                    // since no U-turns from_tracks must be INC as well
                     VTR_ASSERT_SAFE(rr_graph.node_direction(from_node) == Direction::INC);
                     y2 = to_chan.bottom();
-                    // since no U-turns from_tracks must be INC as well
                     y1 = draw_coords->tile_y[to_ylow - 1] + draw_coords->get_tile_height();
                 }
             } else { // DEC wire starts at top edge
--- a/libs/librrgraph/src/base/check_rr_graph.cpp
+++ b/libs/librrgraph/src/base/check_rr_graph.cpp
@@ -556,7 +556,7 @@ void check_rr_node(const RRGraphView& rr_graph,
     float C = rr_graph.node_C(rr_node);
     float R = rr_graph.node_R(rr_node);
 
-    if (rr_type == e_rr_type::CHANX || rr_type == e_rr_type::CHANY) {
+    if (rr_type == e_rr_type::CHANX || rr_type == e_rr_type::CHANY || rr_type == e_rr_type::CHANZ) {
         if (C < 0. || R < 0.) {
             VPR_ERROR(VPR_ERROR_ROUTE,
                       "in check_rr_node: node %d of type %d has R = %g and C = %g.\n", inode, rr_type, R, C);


libs/libarchfpga/src/physical_types.h
```diff
@@ -2067,7 +2067,11 @@ struct t_arch {
     // types of connections requires a different switch, all names should correspond to a switch in Switches.
     std::vector<std::string> ipin_cblock_switch_name;
 
-    std::vector<t_grid_def> grid_layouts; //Set of potential device layouts
+    /// Set of potential device layouts
+    std::vector<t_grid_def> grid_layouts;
+
+    /// Returns the selected layout
+    const t_grid_def& grid_layout() const;
```
AmirhosseinPoolad: Add some more explanation here e.g. that the selected layout is either auto or (if defined) whatever --device is.

libs/libarchfpga/src/scatter_gather_types.h
```diff
@@ -14,17 +14,19 @@ enum class e_scatter_gather_type {
 };
 
 /**
- * @brief Struct containing information of an <sg_location> tag. An <sg_location> tag instantiates the scatter-gather pattern in some switchblock locations around the device.
+ * @brief Struct containing information of an <sg_location> tag.
+ * An <sg_location> tag instantiates the scatter-gather pattern in some switchblock locations around the device.
+ * `x_spec` and `y_spec` are only valid if type == e_sb_location::E_XY_SPECIFIED.
```
AmirhosseinPoolad: What are x_spec and y_spec?

vpr/src/draw/draw.cpp
```diff
@@ -462,54 +462,64 @@ void init_draw_coords(float clb_width, const BlkLocRegistry& blk_loc_registry) {
...
+    // space around the device edges
+    constexpr float VISIBLE_MARGIN = 0.01;
+
+    float draw_width = draw_coords->tile_x[grid.width() - 1] + draw_coords->get_tile_width();
+    float draw_height = draw_coords->tile_y[grid.height() - 1] + draw_coords->get_tile_height();
+
+    initial_world = ezgl::rectangle(
+        {-VISIBLE_MARGIN * draw_width, -VISIBLE_MARGIN * draw_height},
+        {(1. + VISIBLE_MARGIN) * draw_width, (1. + VISIBLE_MARGIN)
+                                                 * draw_height});
```
AmirhosseinPoolad: I'm having trouble understanding this. It's copy pasted from set_initial_world so you should use that, but other than that I also don't understand why your code would need this here at all (set_initial_world is called in update_screen anyway).

Also I think the comment actually belongs to set_initial_world and was here (in master) by mistake.

vpr/src/draw/draw_basic.cpp
```diff
@@ -170,17 +166,12 @@ void drawplace(ezgl::renderer* g) {
                                              abs_clb_bbox.height());
                             }
                             /* Draw text for block type so that user knows what block */
-                            if (width_offset == 0
-                                && height_offset == 0) {
+                            if (grid.is_root_location({i, j, layer_num})) {
```
AmirhosseinPoolad: Didn't we check for that already at line 90? Not sure how this could ever be not true.

vpr/src/draw/draw.cpp
soheilshahrouz: I just updated the comments. I'll remove the duplicate code

vpr/src/draw/draw.cpp
```diff
@@ -301,10 +301,11 @@ void update_screen(ScreenUpdatePriority priority,
     
     strcpy(draw_state->default_message, msg);
 
-    if (!draw_state->show_graphics)
+    if (!draw_state->show_graphics) {
         ezgl::set_disable_event_loop(true);
-    else
+    } else {
```
vaughnbetz, line 306: Personally I prefe
if () {
  // stuff
}
else {  // I broke this line.

}

vpr/src/draw/draw_basic.cpp
```diff
@@ -70,119 +70,101 @@ const std::vector<ezgl::color> kelly_max_contrast_colors = {
...
+                    ClusterBlockId bnum = grid_blocks.block_at_location({i, j, k, layer_num});
+                    // Fill background for the clb. Do not fill if "show_blk_internal" is toggled.
+
+                    // Determine the block color and logical type
+                    ezgl::color block_color;
+                    t_logical_block_type_ptr logical_block_type = nullptr;
+                    // flag whether the current location is highlighted with a special color or not
+                    bool current_loc_is_highlighted = false;
+
+                    if (placer_breakpoint_reached()) {
```
vaughnbetz: Could move this into a routine with a proper comment.  I think this is checking if we've stopped at a placer breakpoint and if the block is in a special list of blocks to be coloured (due to being part of a move).

vaughnbetz: Could move this into a routine with a proper comment. I think this is checking if we've stopped at a placer breakpoint and if the block is in a special list of blocks to be coloured (due to being part of a move).

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -113,31 +121,40 @@ void draw_chanx_to_chanx_edge(RRNodeId from_node, RRNodeId to_node, RRSwitchId r
...
     else {
         if (rr_graph.node_direction(to_node) != Direction::BIDIR) {
             // must connect to to_node's wire beginning at x2
             if (rr_graph.node_direction(to_node) == Direction::INC) { // INC wire starts at leftmost edge
-                VTR_ASSERT(from_xlow < to_xlow);
-                x2 = to_chan.left();
-                // since no U-turns from_tracks must be INC as well
-                x1 = draw_coords->tile_x[to_xlow - 1]
-                     + draw_coords->get_tile_width();
+                if (rr_graph.node_direction(from_node) == Direction::DEC) {
```
vaughnbetz, line 141: I think commenting why you don't draw from the ends in this case would be good.

vpr/src/route/rr_graph_generation/build_scatter_gathers.h
```diff
@@ -66,3 +66,6 @@ std::vector<t_bottleneck_link> alloc_and_load_scatter_gather_connections(const s
                                                                          const t_chan_width& nodes_per_chan,
                                                                          vtr::RngContainer& rng,
                                                                          vtr::NdMatrix<std::vector<t_bottleneck_link>, 2>& interdie_3d_links);
+
+void convert_interposer_cuts_to_sg_patterns(const std::vector<t_layer_def>& interposer_inf,
```
vaughnbetz, line 77: Doxygen comment.

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -376,6 +378,14 @@ static void build_rr_graph(e_graph_type graph_type,
 static int get_delayless_switch_id(const t_det_routing_arch& det_routing_arch,
                                    bool load_rr_graph);
 
+static void add_and_connect_non_3d_sg_links(RRGraphBuilder& rr_graph_builder,
```
vaughnbetz, line 397: Doxygen comment.
Make sure you mention this should only be called once the scatter-gather links/nodes are in the spatial lookup. 
And say what the function does.

vaughnbetz: Make sure it says you're passing data back through the non_3d_... reference.

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
```diff
@@ -363,6 +363,78 @@ void alloc_and_load_inter_die_rr_node_indices(RRGraphBuilder& rr_graph_builder,
...
+                                                                                       int& index) {
+    const DeviceContext& device_ctx = g_vpr_ctx.device();
+    const DeviceGrid& grid = device_ctx.grid;
+
+    vtr::NdMatrix<int, 3> chanx_ptc({grid.get_num_layers(), grid.width(), grid.height()}, chan_width_inf.x_max);
+    vtr::NdMatrix<int, 3> chany_ptc({grid.get_num_layers(), grid.width(), grid.height()}, chan_width_inf.y_max);
+
+    std::vector<std::pair<RRNodeId, int>> node_indices;
+    node_indices.reserve(bottleneck_links.size());
+
```
vaughnbetz, line 379: Suggest outlining the overall flow here ... what are the sequence of steps to implement an sg_link rr node?

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
```diff
@@ -363,6 +363,78 @@ void alloc_and_load_inter_die_rr_node_indices(RRGraphBuilder& rr_graph_builder,
...
+            xlow = xhigh = dst_loc.x;
+            ylow = dst_loc.y + 1;
+            yhigh = src_loc.y;
+        } else {
+            VTR_ASSERT(false);
+        }
+
+        vtr::NdMatrix<int, 3>& ptc_matrix = (chan_type == e_rr_type::CHANX) ? chanx_ptc : chany_ptc;
+
+        int ptc = 0;
```
vaughnbetz, line 397: Explain why you get the max (want to ensure ptc_num is unique along the length of the rr_node we're creating for this sg_link.

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
vaughnbetz: This routine works out the node properties and adds it to the spatial lookup.  Later, you'll ...

vpr/src/route/rr_graph_generation/rr_node_indices.h
```diff
@@ -41,6 +41,11 @@ void alloc_and_load_inter_die_rr_node_indices(RRGraphBuilder& rr_graph_builder,
                                               const vtr::NdMatrix<std::vector<t_bottleneck_link>, 2>& interdie_3d_links,
                                               int* index);
 
+std::vector<std::pair<RRNodeId, int>> alloc_and_load_non_3d_sg_pattern_rr_node_indices(RRGraphBuilder& rr_graph_builder,
```
vaughnbetz, line 52: Doxygen comment.

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -2011,6 +2041,91 @@ static void build_rr_chan(RRGraphBuilder& rr_graph_builder,
...
+
+        rr_graph_builder.set_node_layer(node_id, layer, layer);
+        rr_graph_builder.set_node_coordinates(node_id, xlow, ylow, xhigh, yhigh);
+        rr_graph_builder.set_node_capacity(node_id, 1);
+
+        const size_t cons_index = link.chan_type == e_rr_type::CHANX ? CHANX_COST_INDEX_START + link.parallel_segment_index
+                                                                     : CHANX_COST_INDEX_START + num_seg_types_x + link.parallel_segment_index;
+        rr_graph_builder.set_node_cost_index(node_id, RRIndexedDataId(cons_index));
+
+        float R = 0;
```
vaughnbetz: This should come from the segment defined to be used by the sg_link.
Also do the same for CHANZ wires.

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -2011,6 +2041,91 @@ static void build_rr_chan(RRGraphBuilder& rr_graph_builder,
...
+        const t_bottleneck_link& link = sg_links[i];
+
+        int xlow, xhigh, ylow, yhigh;
+        Direction direction;
+        const t_physical_tile_loc& src_loc = link.gather_loc;
+        const t_physical_tile_loc& dst_loc = link.scatter_loc;
+
+        VTR_ASSERT_SAFE(src_loc.layer_num == dst_loc.layer_num);
+        const int layer = src_loc.layer_num;
+
```
vaughnbetz: could this be refactored with another block of code (which I think was the spatial lookup insertion code) to have a helper once)?

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -2011,6 +2041,91 @@ static void build_rr_chan(RRGraphBuilder& rr_graph_builder,
...
+                                            const t_chan_details& chan_details_x,
+                                            const t_chan_details& chan_details_y,
+                                            size_t num_seg_types_x,
+                                            t_rr_edge_info_set& non_3d_sg_rr_edges_to_create) {
+    VTR_ASSERT(sg_links.size() == sg_node_indices.size());
+    const size_t num_links = sg_links.size();
+
+    for (size_t i = 0; i < num_links; i++) {
+
+        const t_bottleneck_link& link = sg_links[i];
```
vaughnbetz, line 2073: Comment the overall flow / algorithm of this routine, or the major code blocks.

vaughnbetz: Unless the comment at the func declaration has enough detail (once it's written).

vaughnbetz (changes requested):

vpr/src/draw/draw_rr_edges.cpp
```diff
@@ -56,24 +56,32 @@ void draw_chany_to_chany_edge(RRNodeId from_node, RRNodeId to_node, RRSwitchId r
...
-                // since no U-turns from_tracks must be INC as well
-                y1 = draw_coords->tile_y[to_ylow - 1]
-                     + draw_coords->get_tile_width();
+                if (rr_graph.node_direction(from_node) == Direction::DEC) {
+                    y2 = to_chan.bottom();
+                    y1 = draw_coords->tile_y[to_ylow] + 0.1 * draw_coords->get_tile_height() ;
+                } else {
+                    VTR_ASSERT_SAFE(rr_graph.node_direction(from_node) == Direction::INC);
+                    y2 = to_chan.bottom();
+                    // since no U-turns from_tracks must be INC as well
```
AmirhosseinPoolad: I think comment is supposed to be above to assert

AmirhosseinPoolad: Actually a better way to do this would be explaining that we need to handle U-turns in an special way and adding a comment explaining the general logic of this section of code. It's a bit vague at the moment.

soheilshahrouz:
@vaughnbetz I can't merge this PR myself. If you don't have further comments, can you merge it?

PR #3306, opened 2025-10-14
librrgraph: add correct array operator

heshpdx:
This is the correct fix for gcc-16's build error. It works and doesn't crash on two different machines tested by SPEC CPU folks.

```
error: no match for 'operator[]' (operand types are 'edge_sort_iterator' and 'long int')
```

### DIFF
--- a/libs/librrgraph/src/base/rr_graph_storage.cpp
+++ b/libs/librrgraph/src/base/rr_graph_storage.cpp
@@ -235,7 +235,7 @@ class edge_sort_iterator {
         return lhs.swapper_.idx_ <= rhs.swapper_.idx_;
     }
 
-    const t_rr_edge_info operator[] (ssize_t n) const {
+    t_rr_edge_info operator[] (ssize_t n) const {
       edge_sort_iterator ret = *this;
       ret.swapper_.idx_ += n;
       return ret.swapper_;


amin1377:
@heshpdx: Since you’re now returning by value, does it still need to be const? In the previous PR, I mentioned making it a const reference because it was returned by reference. For the [] operator, it’s common to return a const reference for better memory efficiency.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -235,6 +235,12 @@ class edge_sort_iterator {
         return lhs.swapper_.idx_ <= rhs.swapper_.idx_;
     }
 
+    const t_rr_edge_info operator[] (ssize_t n) const {
```
heshpdx:
```suggestion
    t_rr_edge_info operator[] (ssize_t n) const {
```

AmirhosseinPoolad:
> The build errors are seen with gcc-16 ToT. Which compiler version did you use?

I used the master branch of gcc for x86_64 on linux, which I assume is the latest in development version (Running gcc --version reports gcc 16). I'm not familiar with ToT. If you could send me the exact branch of gcc you used I could test the PR using that branch.

AmirhosseinPoolad:
I think I figured out the issue. edge_sort_iterator advertises itself as a `std::random_access_iterator_tag` but it's a lie. random access iterators require operator[] but in their infinite wisdom C++ compilers don't actually check that and will only error out if some algorithm using the iterator e.g. std::sort tries to use operator[]. I can't be sure, but I think what happened here is that my STL implementation didn't use operator[] while yours did.

I do agree with this PR, this iterator should definitely have an operator[] and it's honestly a total coincidence that it compiles on any machine right now. I'm not sure if this implementation is correct, however. each time you call operator[n] the internal index is added to by n. This is similar to what operator+= does. I don't think operator[] should change the internal index. I'll take a look at the relevant code to see what a correct implementation would look like, and maybe port this iterator to use concepts instead.

vaughnbetz:
Thanks for the careful review @AmirhosseinPoolad !

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -235,6 +235,12 @@ class edge_sort_iterator {
...
+      edge_sort_iterator ret = *this;
+      ret.swapper_.idx_ += n;
+      return ret.swapper_;
+    }
+
```
AmirhosseinPoolad: Would this compile successfully on your system if it returned an edge_swapper& like operator* does? operator[n] is (or should be) equivalent to \*(it + n). We basically need a combination of operator+ and operator* here.

AmirhosseinPoolad: I actually think the previous PR was fine. The function was const and that was correct since (*this + n) constructs a new iterator and doesn't modify 'this' at all.

AmirhosseinPoolad (commented):
Sorry for all the rants I did in this PR @heshpdx! Thank you for flagging this issue. I'm not entirely sure if the current implementation of operator[] is correct. I had some thoughts on it. You should definitely test this PR locally on the machine you're having build issues on (since my machine and the CI machine doesn't use operator[] in std::stable_sort) to make sure it's completely correct. I'm sure the strong tests would suffice:

> ./run_reg_test.py vtr_reg_strong -j[num_cores]

libs/librrgraph/src/base/rr_graph_storage.cpp
heshpdx: [!3298](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3298) did not work as it still led to segfaults, so Amin reverted it.

ToT stands for Top of Tree, sorry about that. The newer compilers appear to be more strict.

The current PR as is, has now been tested successfully across a wide variety of compilers and systems (gcc, llvm, nvidia/intel/amd/cray/ibm compilers, and their respective hardware).  SPEC CPU 2026 is built with `-std=c++17`; builds with newer standards might work but we are focusing on C++17 as the baseline. The successful tests are NOT using upstream VPR's unit test or regressions, but rather just the few titan cmdlines of place+route that we run for the benchmarks.

I'm not a tester, I'm just the project leader and messenger 😄.  Maybe the community waits to merge this until the problem is exposed by a more mature gcc-16, and maybe the real solution is to use `ranges::zip_view`. I don't have an opinion on whether or not this change should be accepted here, I am just sharing that it works for SPEC. Thanks!

heshpdx: Our compiler team did a bisect and said that the "first bad commit" on gcc-16 was the following, so maybe that gives a clue to the behavioral change for VPR which necessitates this PR.
https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=aaeca77a79a9a897c97f00f83f2471e7a8bd6685

vaughnbetz:
I'm fine with merging this as everyone seems to agree the old behaviour was not really guaranteed by the standard.  @AmirhosseinPoolad : I think you're also saying this is fine to merge, but we may someday be able to do it in a cleaner way (but not yet ... would have to move to a later C++ standard).  @AmirhosseinPoolad : I'd still like a quick CPU time check, which I expect won't show any change, but I like to be paranoid ...

libs/librrgraph/src/base/rr_graph_storage.cpp
heshpdx: Some early testing reports from our next integration hint at needing to remove the const. We will get back to you.

vaughnbetz:
Had a discussion with @AmirhosseinPoolad and @AlexandreSinger . They believe the code we're fixing (not the new code, the existing code) that stores edges in multiple parallel arrays and sorts them is fragile in its use of iterators / subscript operator. @AmirhosseinPoolad is going to rewrite the edge sorting code to be simpler so we don't need to make a complex iterator / subscript operator here. Basic idea: 
- Make a struct with (sort_val, edge_index) for each edge. sort_val will be computed from the various edge data arrays (probably 2 of them) in the same way as now.
- Call stable_sort with this new data structure
- Make a new copy of each of the edge_data_arrays, with the indices re-ordered as specified by the sort output
- Delete the old copies / move the new edge_data_arrays into the main storage.
- Delete the temporary sort_val,edge_index vector

It might be slightly less efficient, but this is not very hot code anyway (we do one sort after making an rr-graph and before routing). It also shouldn't be the peak memory footprint as the router data structures shouldn't be loaded yet. We should measure that though, and we could do the edge_data_array copy one at a time if it was the peak memory footprint.

heshpdx:
I acknowledge the better solution is #3316.
SPEC will have to keep this one though, because CPU2026 is declared to support C++17.
Thanks for your help. Feel free to close this PR!

heshpdx:
> Ok, will close. If there’s some C++20 stuff in #3316 a more basic version of it could be created for spec if you wished.

The more basic version would be to put back the iterators 🤣

Don't worry about it.

PR #3313, opened 2025-10-16
Add needed includes to vpr_signal_handler

AmirhosseinPoolad:
vpr_signal_handler.cpp used some header files that were indirectly included. This commit directly adds those header files to vpr_signal_handler.

I also removed the unused includes. This file previously had some issues with the pre-release GCC-16 (See #3306)  because unistd.h was not directly included. The ifdef for unix is based on similar code in vtr_util.cpp:

```
#if defined(__unix__)
#include <unistd.h> //For getpid()
#endif
```


amin1377 (approved):
Looks good. Thanks, Amir!

PR #3316, opened 2025-10-21
Add edge sorting method to rr_graph_storage

AmirhosseinPoolad:
Previously we used a non-conformant proxy iterator to sort the rr edges (which were stored in a struct-of-array fashion). That was very fragile and resulted in multiple problems when updating compilers and standard libraries. This PR addresses the issues we had with the previous methods we used to sort edges by using a slower but conformant method.

Sorting edges is now much slower and requires building 4 entire new arrays for edges, but since we only sort two times when building the RR graph it should be fine.

### DIFF
--- a/libs/librrgraph/src/base/rr_graph_storage.h
+++ b/libs/librrgraph/src/base/rr_graph_storage.h
@@ -801,8 +801,15 @@ class t_rr_graph_storage {
     void mark_edges_as_rr_switch_ids();
 
     /** @brief
-     * Sorts edge data such that configurable edges appears before
-     * non-configurable edges.
+     * Sorts all edges based on source node with configurable coming first, or equivalently
+     * partitions the edges based on their source node and configurability. Afterwards, it
+     * builds a list of indices to each RRNodeId's first edge. With this, it can then easily
+     * say which edges belong to which node.
+     * 
+     * @param rr_switches Information of switches, used to figure out if edge is configurable or not.
+     * 
+     * @note Must be called after constructing the RR Graph. You can not use most of the edge accessor
+     * methods before this method is called since the relevant data structures are not set up yet.
      */
     void partition_edges(const vtr::vector<RRSwitchId, t_rr_switch_inf>& rr_switches);
 
@@ -829,31 +836,32 @@ class t_rr_graph_storage {
         std::vector<RREdgeId> edge_indices(edge_range.begin(), edge_range.end());
 
         std::ranges::stable_sort(edge_indices, comparison_function);
-
-        vtr::vector<RREdgeId, RRNodeId> new_edge_src_node_(num_edges, RRNodeId::INVALID());
-        vtr::vector<RREdgeId, RRNodeId> new_edge_dest_node_(num_edges, RRNodeId::INVALID());
-        vtr::vector<RREdgeId, short> new_edge_switch_(num_edges, LIBRRGRAPH_UNDEFINED_VAL);
-        vtr::vector<RREdgeId, bool> new_edge_remapped_(num_edges, false);
         
-        size_t new_index = 0;
-        for (RREdgeId edge_index : edge_indices) {
-    
-            RREdgeId new_edge_index = RREdgeId(new_index);
+        // Generic lambda that allocates a 'vec'-sized new vector with all elements set to default value,
+        // then builds the new vector to have rearranged elements from 'vec' and finaly move the new vector
+        // to replace vec. Essentially does a permutation on vec based on edge_indices.
+        auto array_rearrage = [&edge_indices] (auto& vec, auto default_value) {
 
-            new_edge_src_node_[new_edge_index] = edge_src_node_[edge_index];
-            new_edge_dest_node_[new_edge_index] = edge_dest_node_[edge_index];
-            new_edge_switch_[new_edge_index] = edge_switch_[edge_index];
-            new_edge_remapped_[new_edge_index] = edge_remapped_[edge_index];
-            
-            new_index++;
-        }
+            // Since vec could have any type, we need to figure out it's type to allocate new_vec.
+            // The scary std::remove_reference stuff does exactly that. This does nothing other than building a new 'vec' sized vector.
+            typename std::remove_reference<decltype(vec)>::type new_vec(vec.size(), default_value);
 
-        VTR_ASSERT(new_index == num_edges);
+            size_t new_index = 0;
+            for (RREdgeId edge_index : edge_indices) {
+                RREdgeId new_edge_index = RREdgeId(new_index);
+                new_vec[new_edge_index] = vec[edge_index];
 
-        edge_src_node_ = std::move(new_edge_src_node_);
-        edge_dest_node_ = std::move(new_edge_dest_node_);
-        edge_switch_ = std::move(new_edge_switch_);
-        edge_remapped_ = std::move(new_edge_remapped_);
+                new_index++;
+            }
+            VTR_ASSERT(new_index == vec.size());
+
+            vec = std::move(new_vec);
+        };
+
+        array_rearrage(edge_src_node_, RRNodeId::INVALID());
+        array_rearrage(edge_dest_node_, RRNodeId::INVALID());
+        array_rearrage(edge_switch_, LIBRRGRAPH_UNDEFINED_VAL);
+        array_rearrage(edge_remapped_, false);
     }
 
     /******************


libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -419,6 +137,26 @@ void t_rr_graph_storage::init_fan_in() {
     }
 }
 
+
+// Functor for sorting edges according to destination node's ID.
```
soheilshahrouz: // ---> /// (doxygen)

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -527,6 +262,35 @@ void t_rr_graph_storage::mark_edges_as_rr_switch_ids() {
     remapped_edges_ = true;
 }
 
+// Functor for sorting edges according to source node, with configurable edges coming first
```
soheilshahrouz: /// (doxygen)

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -809,6 +813,47 @@ class t_rr_graph_storage {
...
+     * and invalidates all the RREdgeIds. This is not an inplace sort, and it is very expensive.
+     * You should not be calling this method more than once or twice in the entire program, definitely do not use it in a hot loop.
+     * @tparam t_comp_func callable object with two size_t arguments. See 'edge_compare_dest_node' for example.
+     * @param comparison_function Comparison function to order edges with.
+     */
+    template <typename t_comp_func>
+    void sort_edges(t_comp_func comparison_function) {
+
+        size_t num_edges = edge_src_node_.size();
+        std::vector<size_t> edge_indices(num_edges, 0);
```
soheilshahrouz: In fucntors, size_t is casted into RREdgeId. I think the code would be more concise if you create a vector of RREdgeId. std::iota might not work with StrongId, but you can use vtr::StrongIdRagne to initlize the vector.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -527,6 +262,35 @@ void t_rr_graph_storage::mark_edges_as_rr_switch_ids() {
...
+class edge_compare_src_node_and_configurable_first {
+  public:
+    edge_compare_src_node_and_configurable_first(const vtr::vector<RRSwitchId, t_rr_switch_inf>& rr_switch_inf, const t_rr_graph_storage& rr_graph_storage)
+        : rr_switch_inf_(rr_switch_inf),
+          rr_graph_storage_(rr_graph_storage) {}
+
+    bool operator()(size_t lhs_idx, size_t rhs_idx) const {
+        RREdgeId lhs = RREdgeId(lhs_idx);
+        RREdgeId rhs = RREdgeId(rhs_idx);
```
soheilshahrouz: You can probably use arguments of type RREdgeId directly and get rid of these two lines.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -809,6 +813,47 @@ class t_rr_graph_storage {
...
+
+        std::iota(edge_indices.begin(), edge_indices.end(), 0);
+
+        std::ranges::stable_sort(edge_indices, comparison_function);
+
+
+
+        vtr::vector<RREdgeId, RRNodeId> new_edge_src_node_(num_edges, RRNodeId::INVALID());
```
soheilshahrouz: Run make foramt. It should remove these blank lines.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -809,6 +813,47 @@ class t_rr_graph_storage {
...
+        vtr::vector<RREdgeId, RRNodeId> new_edge_dest_node_(num_edges, RRNodeId::INVALID());
+        vtr::vector<RREdgeId, short> new_edge_switch_(num_edges, LIBRRGRAPH_UNDEFINED_VAL);
+        vtr::vector<RREdgeId, bool> new_edge_remapped_(num_edges, false);
+        
+        size_t new_index = 0;
+        for (size_t edge_index : edge_indices) {
+            new_edge_src_node_[RREdgeId(new_index)] = edge_src_node_[RREdgeId(edge_index)];
+            new_edge_dest_node_[RREdgeId(new_index)] = edge_dest_node_[RREdgeId(edge_index)];
+            new_edge_switch_[RREdgeId(new_index)] = edge_switch_[RREdgeId(edge_index)];
+            new_edge_remapped_[RREdgeId(new_index)] = edge_remapped_[RREdgeId(edge_index)];
```
soheilshahrouz: If you use a vtr::vector<RREdgeId, RRedgeId> for edge_indices, you can use vtr::vector::pairs() method to automatically  iterate over new_index and edge_index. This can also help get rid of type casting.

soheilshahrouz (commented):
Thank @AmirhosseinPoolad.

I suggested a few minor changes.

libs/librrgraph/src/base/rr_graph_storage.h
AmirhosseinPoolad: I think this would kind of obfuscate what the code is doing and make it harder to understand. I did the new_index casting in a separate line to make these four lines more readable.

vaughnbetz:
The largest circuits look like a larger men increase — maybe 8%. We could reallocate the vectors  one by one at some cpu cost but probably less peak memory. I’d zoom in on LU32 or neuron and try that.VaughnOn Oct 21, 2025, at 8:44 PM, Amir Poolad ***@***.***> wrote:﻿AmirhosseinPoolad left a comment (verilog-to-routing/vtr-verilog-to-routing#3316)
Ran the VTR circuits, results look very good.



Metric
Vs Master




Total VPR runtime
0.968


Max mem. usage
1.030


RRG generation time
0.805



This appears to be a pretty good change overall. New code is (imo) more readable, less fragile and faster with only a slight memory cost.
@vaughnbetz

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -809,6 +820,49 @@ class t_rr_graph_storage {
...
+        std::ranges::stable_sort(edge_indices, comparison_function);
+        
+        // Generic lambda that allocates a 'vec'-sized new vector with all elements set to default value,
+        // then builds the new vector to have rearranged elements from 'vec' and finaly move the new vector
+        // to replace vec. Essentially does a permutation on vec based on edge_indices.
+        auto array_rearrage = [&edge_indices] (auto& vec, auto default_value) {
+
+            // Since vec could have any type, we need to figure out it's type to allocate new_vec.
+            // The scary std::remove_reference stuff does exactly that. This does nothing other than building a new 'vec' sized vector.
+            typename std::remove_reference<decltype(vec)>::type new_vec(vec.size(), default_value);
```
vaughnbetz, line 847: I wonder if we have to do a .reserve or .shrink_to_fit to make sure this new_vec does not have any extra capacity? We're careful to do that in the RR graph, so I imagine the original edge_src_node_ etc. have done that.  If we don't do it for the new one, there's some chance the standard library created a vector with extra capacity for us, which we don't want with these big vectors.

Seems worth testing, and also a good practice to use .shrink_to_fit for safety in case some versions of the standard library default to capacity == requested constructor size for big vectors and some default to something else (e.g. 2x size for any constructor that asks for a size).

vaughnbetz (commented):
Nice!  Code looks good, but I have a question / concern about .shrink_to_fit and .capacity.  I think we can make the code more memory size safe (and maybe it will save some memory, depending on what the standard library is doing).

libs/librrgraph/src/base/rr_graph_storage.h
AmirhosseinPoolad: t_rr_graph_storage actually has a shrink_to_fit method that does this for all the vectors in t_rr_graph_storage. Currently it's only used for building tileable RR graphs. I think we should test calling t_rr_graph_storage::shrink_to_fit at the end of RR graph generation in a separate PR, since doing it in the sorting function wouldn't actually change the peak memory usage of the sorting function and would only help in "long term." It might also have an extra memory and runtime cost (The underlying allocator might have to allocate a new smaller buffer and copy all the elements to it.)

vaughnbetz:
@heshpdx : FYI.  Here is a PR that makes the funky [] operator unnecessary.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -809,6 +820,49 @@ class t_rr_graph_storage {
...
+            }
+            VTR_ASSERT(new_index == vec.size());
+
+            vec = std::move(new_vec);
+        };
+
+        array_rearrage(edge_src_node_, RRNodeId::INVALID());
+        array_rearrage(edge_dest_node_, RRNodeId::INVALID());
+        array_rearrage(edge_switch_, LIBRRGRAPH_UNDEFINED_VAL);
+        array_rearrage(edge_remapped_, false);
```
heshpdx, line 864: Are these typos?
`rearrage` --> `rearrange`

AmirhosseinPoolad: Thanks for catching this! Opened #3351 to fix it.

PR #3320, opened 2025-10-27
Now direct connection builder checks the port name of destination subtile

tangxifan:
#### Description
This is a follow-up PR on PR #3319 

It adds the dedicated testcase to validate

- When the destination sub tile is out of the range of subtiles in a tile, the direct connection builder should skip instead of aborting

The testcase will trigger another bug:

The ``clb_to_clb_directs`` relies on pin index to find the destination pin of subtiles. This is not reliable, which may cause wrong direct connections to be added.

Take the example of the testcases in this PR, 

```
<direct name="cim_direct1" from_pin="cim8_16k.out[15:0]" to_pin="cim8_16k.d_in[31:16]" x_offset="0" y_offset="0" z_offset="-1"/>
```

The line expects to build direct connections from ``MULT[1]`` to ``RAM[0]``. However, when checking if there are any connections required from ``MULT[2]`` to ``MULT[1]``, the current builder will mistakenly add an edge.
This is due to that the pin index of ``RAM[0]`` is always within the range of ``MULT[1]`` and ``MULT[2]`` (RAM has more pins than MULT).

```
+---------------+
|                         |
|    RAM[0]         |<-+----+
+---------------+    |        |
|    MULT[1]       |---+       |
+---------------+             |
|    MULT[2]       |--------+
+---------------+
```

This PR adds the port name to be checked when building the direct connection, to avoid such bug.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/pack/cluster_legalizer.h
+++ b/vpr/src/pack/cluster_legalizer.h
@@ -11,6 +11,7 @@
  * externally to the Packer in VPR.
  */
 
+#include <string>
 #include <vector>
 #include "atom_netlist_fwd.h"
 #include "noc_data_types.h"
@@ -431,6 +432,13 @@ class ClusterLegalizer {
         VTR_ASSERT_SAFE(cluster_id.is_valid() && (size_t)cluster_id < legalization_clusters_.size());
         const LegalizationCluster& cluster = legalization_clusters_[cluster_id];
         return cluster.pb;
+    }
+
+    /// @brief Gets the name of the given cluster.
+    inline std::string get_cluster_name(LegalizationClusterId cluster_id) const {
+        VTR_ASSERT_SAFE(cluster_id.is_valid() && (size_t)cluster_id < legalization_clusters_.size());
+        const LegalizationCluster& cluster = legalization_clusters_[cluster_id];
+        return cluster.pb->name;
     }
 
     /// @brief Gets the logical block type of the given cluster.
--- a/vpr/src/pack/cluster_util.cpp
+++ b/vpr/src/pack/cluster_util.cpp
@@ -39,8 +39,9 @@ static void echo_clusters(char* filename, const ClusterLegalizer& cluster_legali
     }
 
     for (auto& cluster_atom : cluster_atoms) {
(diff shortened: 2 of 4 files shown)


vpr/src/route/rr_graph_generation/clb2clb_directs.cpp
```diff
@@ -62,6 +62,8 @@ std::vector<t_clb_to_clb_directs> alloc_and_load_clb_to_clb_directs(const std::v
         clb_to_clb_directs[i].to_clb_type = physical_tile;
 
         tile_port = find_tile_port_by_name(physical_tile, port_name);
+        // Cache the destination port name as the pin index is not enough to identify if the destination subtile is the one we want!!!
```
vaughnbetz: Minor nit: we don't need the !!! at the end.  
If there's something you want to emphasize about this test I think you'll have to explain it. But just adding !!! in comments will be a distraction.

vpr/src/route/rr_graph_generation/tileable_rr_graph/tileable_rr_graph_gsb.cpp
```diff
@@ -1774,6 +1774,15 @@ void build_direct_connections_for_one_gsb(const RRGraphView& rr_graph,
                         }
                     }
                     VTR_ASSERT(to_sub_tile != nullptr);
+                    /* Check if the to port is the one from the subtile, if not, pass as this is not the destination */
```
vaughnbetz: Minor nit: we've moved to 
// Use double slash for implementation comments

so if you can change to that style it would be good.

vpr/src/route/rr_graph_generation/clb2clb_directs.cpp
tangxifan: Understood. The exclamation has been removed.

vpr/src/route/rr_graph_generation/tileable_rr_graph/tileable_rr_graph_gsb.cpp
tangxifan: Sure. I will keep this convention in future code changes.

PR #3329, opened 2025-11-01
[Pack] Cleaned Up Read Netlist Code

AlexandreSinger:
The code for reading a netlist file was very hard to follow and made it challenging to try and improve the pb interface. Went through and cleaned up the code to make it a bit more obvious what it is trying to do.

### DIFF
--- a/vpr/src/base/read_netlist.cpp
+++ b/vpr/src/base/read_netlist.cpp
@@ -44,11 +44,21 @@
 
 static const char* netlist_file_name = nullptr;
 
+/**
+ * @brief XML parser to populate the routing represented on the ports of the pb.
+ */
 static void process_ports(pugi::xml_node Parent,
                           t_pb* pb,
                           t_pb_routes& pb_route,
                           const pugiutil::loc_data& loc_data);
 
+/**
+ * @brief XML parser to populate pb info and to update internal nets of the parent CLB
+ *
+ *   @param Parent     XML tag for this pb_type
+ *   @param pb         physical block to use
+ *   @param loc_data   xml location info for error reporting
+ */
 static void process_pb(pugi::xml_node Parent,
                        const ClusterBlockId index,
                        t_pb* pb,
@@ -56,6 +66,13 @@ static void process_pb(pugi::xml_node Parent,
                        int& num_primitives,
                        const pugiutil::loc_data& loc_data);
 
+/**
+ * @brief  XML parser to populate CLB info and to update nets with the nets of this CLB
+ *
+ *   @param clb_nlist  Array of CLBs in the netlist
+ *   @param index      index of the CLB to allocate and load information into
+ *   @param loc_data   xml location info for error reporting
+ */
 static void process_complex_block(pugi::xml_node Parent,
                                   const ClusterBlockId index,
                                   int& num_primitives,
@@ -63,20 +80,37 @@ static void process_complex_block(pugi::xml_node Parent,
                                   const std::unordered_map<std::string, int>& logical_block_type_name_to_index,
                                   ClusteredNetlist& clb_nlist);
 
+/**
+ * @brief After the clustered netlist has been created, many of the atom
+ *        netlist datastructures will become out of date. Need to sync the
+ *        clustered netlist with the atom netlist.
+ */
 static void sync_clustered_and_atom_netlists(ClusteredNetlist& clb_nlist,
                                              AtomNetlist& atom_netlist,
                                              AtomLookup& atom_lookup);
 
+/**
+ * @brief This function updates the nets list and the connections between
+ *        that list and the complex block
+ */
 static void load_external_nets_and_cb(ClusteredNetlist& clb_nlist, const AtomNetlist& atom_netlist);
 
 static void load_internal_to_block_net_nums(const t_logical_block_type_ptr type, t_pb_routes& pb_route);
 
 static void load_atom_index_for_pb_pin(t_pb_routes& pb_route, int ipin);
 
+/**
+ * @brief Counts the number of constant generators in the design and displays
+ *        this information to the log files.
+ */
 static void mark_constant_generators(const ClusteredNetlist& clb_nlist, int verbosity);
 
 static size_t mark_constant_generators_rec(const t_pb* pb, const t_pb_routes& pb_route, int verbosity);
 
+/**
+ * @brief Walk through the atom netlist looking up and storing the t_pb_graph_pin
+ *        associated with each connected AtomPinId
+ */
 static void load_atom_pin_mapping(const ClusteredNetlist& clb_nlist);
 
 /**
@@ -88,8 +122,8 @@ ClusteredNetlist read_netlist(const char* net_file,
                               const t_arch* arch,
                               bool verify_file_digests,
                               int verbosity) {
-    AtomNetlist& atom_netlist = g_vpr_ctx.mutable_atom().mutable_netlist();
-    AtomLookup& atom_lookup = g_vpr_ctx.mutable_atom().mutable_lookup();
+    AtomNetlist& mutable_atom_netlist = g_vpr_ctx.mutable_atom().mutable_netlist();
+    AtomLookup& mutable_atom_lookup = g_vpr_ctx.mutable_atom().mutable_lookup();
     const auto& logical_block_types = g_vpr_ctx.device().logical_block_types;
 
     vtr::ScopedStartFinishTimer read_netlist_timer("Loading packed FPGA netlist file.");
@@ -174,11 +208,11 @@ ClusteredNetlist read_netlist(const char* net_file,
             //Note that we currently don't require that the atom_netlist_id exists,
             //to remain compatible with old .net files
             std::string atom_nl_id = atom_netlist_id.value();
-            if (atom_nl_id != atom_netlist.netlist_id()) {
+            if (atom_nl_id != mutable_atom_netlist.netlist_id()) {
                 std::string msg = vtr::string_fmt(
                     "Netlist was generated from a different atom netlist file"
                     " (loaded atom netlist ID: %s, packed netlist atom netlist ID: %s)",
-                    atom_nl_id.c_str(), atom_netlist.netlist_id().c_str());
+                    atom_nl_id.c_str(), mutable_atom_netlist.netlist_id().c_str());
                 if (verify_file_digests) {
                     vpr_throw(VPR_ERROR_NET_F, netlist_file_name, loc_data.line(top), msg.c_str());
                 } else {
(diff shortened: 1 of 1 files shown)


vpr/src/base/read_netlist.cpp
```diff
@@ -27,27 +35,37 @@
...
-#include "hash.h"
 #include "globals.h"
 #include "atom_netlist.h"
 #include "read_netlist.h"
 #include "pb_type_graph.h"
 
 static const char* netlist_file_name = nullptr;
 
-static void processPorts(pugi::xml_node Parent, t_pb* pb, t_pb_routes& pb_route, const pugiutil::loc_data& loc_data);
+static void processPorts(pugi::xml_node Parent,
```
AmirhosseinPoolad: I think while we're here we should rename these functions to snake_case

vpr/src/base/read_netlist.cpp
```diff
@@ -27,27 +35,37 @@
...
+                         t_pb* pb,
+                         t_pb_routes& pb_route,
+                         const pugiutil::loc_data& loc_data);
 
-static void processPb(pugi::xml_node Parent, const ClusterBlockId index, t_pb* pb, t_pb_routes& pb_route, int* num_primitives, const pugiutil::loc_data& loc_data, ClusteredNetlist* clb_nlist);
+static void processPb(pugi::xml_node Parent,
```
AmirhosseinPoolad: snake_case. There's a bunch more in this file but I won't write a message for each.

vpr/src/base/read_netlist.cpp
AmirhosseinPoolad: Also needs doxygen.

vpr/src/base/read_netlist.cpp
AmirhosseinPoolad: Doxygen.

vpr/src/base/read_netlist.cpp
```diff
@@ -27,27 +35,37 @@
...
                                 const ClusterBlockId index,
-                                int* num_primitives,
+                                int& num_primitives,
                                 const pugiutil::loc_data& loc_data,
-                                ClusteredNetlist* clb_nlist);
+                                const std::unordered_map<std::string, int>& logical_block_type_name_to_index,
+                                ClusteredNetlist& clb_nlist);
 
-static int add_net_to_hash(t_hash** nhash, const char* net_name, int* ncount);
+static void sync_clustered_and_atom_netlists(ClusteredNetlist& clb_nlist,
```
AmirhosseinPoolad, line 88: Doxygen

vpr/src/base/read_netlist.cpp
```diff
@@ -27,27 +35,37 @@
...
+                                             AtomNetlist& atom_netlist,
+                                             AtomLookup& atom_lookup);
 
-static void load_external_nets_and_cb(ClusteredNetlist& clb_nlist);
+static void load_external_nets_and_cb(ClusteredNetlist& clb_nlist, const AtomNetlist& atom_netlist);
```
AmirhosseinPoolad, line 96: Doxygen

vpr/src/base/read_netlist.cpp
```diff
@@ -70,17 +86,23 @@ ClusteredNetlist read_netlist(const char* net_file,
...
-    size_t bcount = 0;
-    std::vector<std::string> circuit_inputs, circuit_outputs, circuit_clocks;
-
-    auto& atom_ctx = g_vpr_ctx.mutable_atom();
+    AtomNetlist& atom_netlist = g_vpr_ctx.mutable_atom().mutable_netlist();
+    AtomLookup& atom_lookup = g_vpr_ctx.mutable_atom().mutable_lookup();
+    const auto& logical_block_types = g_vpr_ctx.device().logical_block_types;
 
-    int num_primitives = 0;
+    clock_t begin = clock();
```
AmirhosseinPoolad: I think you should use vtr::ScopedActionTimer instead

vpr/src/base/read_netlist.cpp
```diff
@@ -70,17 +86,23 @@ ClusteredNetlist read_netlist(const char* net_file,
...
                               bool verify_file_digests,
                               int verbosity) {
-    clock_t begin = clock();
-    size_t bcount = 0;
-    std::vector<std::string> circuit_inputs, circuit_outputs, circuit_clocks;
-
-    auto& atom_ctx = g_vpr_ctx.mutable_atom();
+    AtomNetlist& atom_netlist = g_vpr_ctx.mutable_atom().mutable_netlist();
+    AtomLookup& atom_lookup = g_vpr_ctx.mutable_atom().mutable_lookup();
+    const auto& logical_block_types = g_vpr_ctx.device().logical_block_types;
```
AmirhosseinPoolad, line 127: Can you get rid of the autos here? There's a lot of them in this file/PR, don't want to spam.

vpr/src/base/read_netlist.cpp
```diff
@@ -393,213 +461,142 @@ static void processPb(pugi::xml_node Parent, const ClusterBlockId index, t_pb* p
...
+            vpr_throw(VPR_ERROR_NET_F, netlist_file_name, loc_data.line(child),
+                      "Unknown mode %s for cb %s #%d.\n", mode.value(),
+                      new_child_pb->name, pb_index);
+        }
 
-    hash_value = insert_in_hash_table(nhash, net_name, *ncount);
-    if (hash_value->count == 1) {
-        VTR_ASSERT(*ncount == hash_value->index);
-        (*ncount)++;
+        processPb(child, index, new_child_pb, pb_route, num_primitives, loc_data);
```
AmirhosseinPoolad: I think we should refactor these functions to net be recursive.  If we want to clean up the packer we really need to get rid of the recursions in these massive functions.

vpr/src/base/read_netlist.cpp
```diff
@@ -789,7 +789,9 @@ static void processPorts(pugi::xml_node Parent, t_pb* pb, t_pb_routes& pb_route,
                     loc += 2; //Skip over the '->'
                     std::string interconnect_name = pins[i].substr(loc, std::string::npos);
 
-                    pin_node = alloc_and_load_port_pin_ptrs_from_string(
+                    int* num_ptrs = nullptr;
+                    int num_sets = 0;
+                    t_pb_graph_pin*** pin_node = alloc_and_load_port_pin_ptrs_from_string(
```
AmirhosseinPoolad, line 810: Is this a 3D array or a 2D array of pointers? We should convert this to std::vector.

vpr/src/base/read_netlist.cpp
```diff
@@ -1053,36 +1050,34 @@ static void mark_constant_generators(const ClusteredNetlist& clb_nlist, int verb
...
+    if (pb_type->blif_model == nullptr) {
         // Check the model name to see if it exists, iterate through children if it does
-        for (i = 0; i < pb->pb_graph_node->pb_type->modes[pb->mode].num_pb_type_children; i++) {
-            pb_type = &(pb->pb_graph_node->pb_type->modes[pb->mode].pb_type_children[i]);
-            for (j = 0; j < pb_type->num_pb; j++) {
+        for (int i = 0; i < pb_type->modes[pb->mode].num_pb_type_children; i++) {
+            const t_pb_type* child_pb_type = &(pb_type->modes[pb->mode].pb_type_children[i]);
+            for (int j = 0; j < child_pb_type->num_pb; j++) {
                 if (pb->child_pbs[i][j].name != nullptr) {
                     const_gen_count += mark_constant_generators_rec(&(pb->child_pbs[i][j]), pb_route, verbosity);
```
AmirhosseinPoolad, line 1076: Same as above. It's incredibly hard to understand how exactly these functions work with the recursion.

AmirhosseinPoolad (changes requested):
Thanks for the PR Alex. Had some thoughts. About the autos, I didn't tag all of them here so you might want to do a CTRL+F for auto to see all of them. I also didn't tag all the functions that were missing doxygen docs.

vpr/src/base/read_netlist.cpp
AlexandreSinger: This one has doxygen, it is labeling the function not the forward declaration. I decided to keep it as it is. I will go through and ensure that all the functions have doxygen though, since some were missing it.

vpr/src/base/read_netlist.cpp
AlexandreSinger: Went through the file and removed most of them. Some I left behind since using auto was useful.

vpr/src/base/read_netlist.cpp
AlexandreSinger: This function is used for parsing XML. Unfortunately removing this recursion will be very challenging to do and may actually take away from the readability of this function. I think it will be best to leave this update behind for now. In the future when the t_pb type is cleaned up a bit more, I hope to absorb this method in a cleaner way.

vpr/src/base/read_netlist.cpp
AlexandreSinger: This is a 2D array of pins. I also do not like this, but I think this will be better cleaned up in a future PR. This is drilled into how pb types are stored in VTR right now. I would prefer to leave it alone in this PR.

vpr/src/base/read_netlist.cpp
AlexandreSinger: I also agree that recursion can make code a bit challenging to work with; however, the way that the pb_types are currently designed, recursion is the cleanest way of iterating over them. I hope to remove this limitation in a future PR.

vpr/src/base/read_netlist.cpp
```diff
@@ -165,49 +188,58 @@ ClusteredNetlist read_netlist(const char* net_file,
         }
 
         //Collect top level I/Os
-        auto top_inputs = pugiutil::get_single_child(top, "inputs", loc_data);
-        circuit_inputs = vtr::StringToken(top_inputs.text().get()).split(" \t\n");
+        pugi::xml_node top_inputs = pugiutil::get_single_child(top, "inputs", loc_data);
+        auto circuit_inputs = vtr::StringToken(top_inputs.text().get()).split(" \t\n");
```
soheilshahrouz: Replace auto with explicit type name

vpr/src/base/read_netlist.cpp
```diff
@@ -216,42 +248,45 @@ ClusteredNetlist read_netlist(const char* net_file,
...
-    for (auto net_id : atom_ctx.netlist().nets()) {
-        atom_ctx.mutable_lookup().remove_atom_net(net_id);
+    return clb_nlist;
+}
+
+/**
+ * @brief After the clustered netlist has been created, many of the atom
+ *        netlist datastructures will become out of date. Need to sync the
+ *        clustered netlist with the atom netlist.
+ */
```
soheilshahrouz: Mode doxygen comment to declartion to be compliant with the coding style guide.

vpr/src/base/read_netlist.cpp
```diff
@@ -261,19 +296,17 @@ ClusteredNetlist read_netlist(const char* net_file,
  *   @param index      index of the CLB to allocate and load information into
  *   @param loc_data   xml location info for error reporting
  */
-static void processComplexBlock(pugi::xml_node clb_block,
+static void process_complex_block(pugi::xml_node clb_block,
```
soheilshahrouz, line 321: Move the doxygen comment

vpr/src/base/read_netlist.cpp
```diff
@@ -1210,35 +1202,36 @@ static void load_atom_pin_mapping(const ClusteredNetlist& clb_nlist) {
 }
 
 void set_atom_pin_mapping(const ClusteredNetlist& clb_nlist, const AtomBlockId atom_blk, const AtomPortId atom_port, const t_pb_graph_pin* gpin) {
-    auto& atom_ctx = g_vpr_ctx.mutable_atom();
+    const AtomNetlist& atom_netlist = g_vpr_ctx.atom().netlist();
+    AtomLookup& atom_lookup = g_vpr_ctx.mutable_atom().mutable_lookup();
```
soheilshahrouz: These references are usually called mutable_??? in the codebase. I think it is better if we  stick to this convention.

vpr/src/base/read_netlist.cpp
```diff
@@ -377,266 +442,203 @@ void processAttrsParams(pugi::xml_node Parent, const char* child_name, T& atom_n
...
                       "Unknown port %s for pb %s[%d].\n", port_name.value(),
-                      pb->pb_graph_node->pb_type->name,
+                      pb_type->name,
                       pb->pb_graph_node->placement_index);
         }
 
         //Extract all the pins for this port
-        pins = vtr::StringToken(Cur.text().get()).split(" \t\n");
-        num_tokens = pins.size();
+        auto pins = vtr::StringToken(Cur.text().get()).split(" \t\n");
```
soheilshahrouz: typename

soheilshahrouz (changes requested):

AlexandreSinger:
Thanks for the feedback @soheilshahrouz , resolved. Please let me know if you have any further comments.

PR #3331, opened 2025-11-04
fix the issue for pack pattern to recognize register connection

saaramahmoudi:
This branch has a fix for a previous PR made by @amin1377:
[Issue 2991](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2991)

In that PR, we are trying to prevent a pre-packer to form molecule that both registered and unregistered outputs are used but only one of them can be routed through LE output based on the architecture. However, the condition only checks for .latch in the blif file and will not catch FFs as a register. This PR only changes the if condition to consider FFs, as well.

### DIFF
--- a/vpr/src/pack/prepack.cpp
+++ b/vpr/src/pack/prepack.cpp
@@ -1144,6 +1144,8 @@ static AtomBlockId get_sink_block(const AtomBlockId block_id,
     AtomBlockId pattern_sink_block_id = AtomBlockId::INVALID();
     for (const auto& sink_pin_id : net_sinks) {
         auto sink_block_id = atom_nlist.pin_block(sink_pin_id);
+        // If the sink block has a clock, it is considered stateful (e.g., a latch or flip-flop).
+        // Mark this so we can later decide whether to drop the block based on the net’s fanout.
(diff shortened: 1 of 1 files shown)


vpr/src/pack/prepack.cpp
```diff
@@ -1141,11 +1141,10 @@ static AtomBlockId get_sink_block(const AtomBlockId block_id,
     // Iterate through all sink blocks and check whether any of them
     // is compatible with the block specified in the pack pattern.
     bool connected_to_latch = false;
-    LogicalModelId latch_model_id = LogicalModels::MODEL_LATCH_ID;
     AtomBlockId pattern_sink_block_id = AtomBlockId::INVALID();
     for (const auto& sink_pin_id : net_sinks) {
         auto sink_block_id = atom_nlist.pin_block(sink_pin_id);
-        if (atom_nlist.block_model(sink_block_id) == latch_model_id) {
+        if (!atom_nlist.block_is_combinational(sink_block_id)) {
```
amin1377, line 1149: I think it would be better to add a comment explaining why this API call is used (we’re assuming that if a block has a clock, it’s stateful and therefore falls into the category we want to detect).

PR #3332, opened 2025-11-04
Remove an unsed hash functor

soheilshahrouz:
Removes a hash functor that is not used.

### DIFF
--- a/libs/libvtrutil/src/vtr_strong_id.h
+++ b/libs/libvtrutil/src/vtr_strong_id.h
@@ -211,9 +211,9 @@ class StrongId {
      * Note that since these are templated functions we provide an empty set of template parameters
      * after the function name (i.e. <>)
      */
-    friend constexpr bool operator== <>(const StrongId<tag, T, sentinel>& lhs, const StrongId<tag, T, sentinel>& rhs);
+    friend constexpr bool operator== <>(const StrongId<tag, T, sentinel>& lhs, const StrongId<tag, T, sentinel>& rhs) noexcept;
     ///@brief != operator
-    friend constexpr bool operator!= <>(const StrongId<tag, T, sentinel>& lhs, const StrongId<tag, T, sentinel>& rhs);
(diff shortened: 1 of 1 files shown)


libs/libarchfpga/src/physical_types_util.cpp
```diff
@@ -687,18 +682,18 @@ std::vector<std::string> block_type_class_index_to_pin_names(t_physical_tile_typ
         pin_info.push_back(block_type_pin_index_to_pin_inst(type, pin_physical_num, is_flat));
     }
 
-    auto cmp = [](const t_pin_inst_port& lhs, const t_pin_inst_port& rhs) {
+    auto cmp = [](const t_pin_inst_port& lhs, const t_pin_inst_port& rhs) noexcept {
```
AmirhosseinPoolad: This should be renamed to something more explanatory. What is it comparing? How is the comparison done?

vpr/src/util/vpr_utils.cpp
```diff
@@ -1479,18 +1476,18 @@ void print_switch_usage() {
...
+            if (!inward_switch_inf[to_node_index].contains(switch_index))
                 inward_switch_inf[to_node_index][switch_index] = 0;
             //VTR_ASSERT(from_node.type != OPIN);
             inward_switch_inf[to_node_index][switch_index]++;
         }
     }
 
     for (const RRNodeId rr_id : device_ctx.rr_graph.nodes()) {
-        for (const auto [rr_switch_id, node_switch_fanin] : inward_switch_inf[rr_id]) {
+        for (const RRSwitchId rr_switch_id : inward_switch_inf[rr_id] | std::views::keys) {
```
AmirhosseinPoolad, line 1487: I don't have a strong preference, but I think the previous approach was simpler to reason about (no pipe operator) and more readable.

AmirhosseinPoolad: To be clear, I know what the new code does but I have absolutely no idea how exactly it works underneath and if I understand correctly it's a ton of STL and compiler magic that might or might not result in optimal code gen. The old code is incredibly simple in comparison.

PR #3333, opened 2025-11-04
[Pack] Removed Pb Pin Scratchpad

AlexandreSinger:
Found another use of scratchpad memory in the pb type graph data structure which was only used in one place.

Removed it and replaced it with proper data structures being passed around.


AmirhosseinPoolad (commented):
Thanks for the PR Alex. Looks good, but I think you should add comments explaining the purpose of seen_pins and pin_marker.

AmirhosseinPoolad:
Lol. LGTM.

PR #3338, opened 2025-11-10
[Router] Fix intra-cluster pin handling for tiles with heterogeneous subtiles

amin1377:
There was a bug in the router when adding intra-cluster resources. Specifically, while retrieving the index of tile-level pins for blocks, there was an incorrect assumption that all sub-tile instances are of the same type. This PR fixes that bug and adds a clarifying comment to the function that returns this list.

To ensure the fix doesn’t introduce issues in existing architectures, I ran the titan_other benchmarks with the flat router enabled, and the results remained unchanged (as expected).

### DIFF
--- a/vpr/src/route/rr_graph_generation/rr_graph_intra_cluster.cpp
+++ b/vpr/src/route/rr_graph_generation/rr_graph_intra_cluster.cpp
@@ -197,33 +197,33 @@ static std::vector<int> get_cluster_block_pins(t_physical_tile_type_ptr physical
     // A counter to keep track of number of tile-level pins for the sub-tiles before the current sub-tile.
     int seen_sub_tiles_num_cluser_pins = 0;
     // The number of tile-level pins for the sub-tile instance that the cluster block is mapped to.
-    int cluster_sub_tile_isnt_num_pins;
+    int cluster_sub_tile_inst_num_pins;
     // A flag to check if the sub-tile instance that the cluster block is mapped to has been found.
     bool found_sub_tile = false;
 
     // Iterate over all the sub-tiles to find the sub-tile instance that the cluster block is mapped to.
-    for (const t_sub_tile& tmp_sub_tile: physical_tile->sub_tiles) {
-        if (tmp_sub_tile.capacity.is_in_range(sub_tile_index)) {
+    for (const t_sub_tile& sub_tile: physical_tile->sub_tiles) {
+        if (sub_tile.capacity.is_in_range(sub_tile_index)) {
             // This sub-tile type is the one that the cluster block is mapped to.
             found_sub_tile = true;
             // The number of tile-level pins for all isntances of the same sub-tile type is the same. Thus,
             // we can the the number of tile-level pins for the sub-tile instance by dividing the total number of pins
             // for the given sub-tile type by the number of instances.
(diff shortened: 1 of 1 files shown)


vpr/src/util/vpr_utils.h
```diff
@@ -328,7 +328,12 @@ std::vector<int> get_cluster_netlist_intra_tile_pins_at_loc(const t_physical_til
...
                                                             const vtr::vector<ClusterBlockId, std::unordered_set<int>>& pin_chains_num,
                                                             t_physical_tile_type_ptr physical_type);
-
+/**
+ * @brief Returns the list of pins (both cluster-level and intra-cluster-level) of the given cluster block.
+ * @param physical_tile The physical tile type that the cluster block is mapped to.
+ * @param cluster_blk_id The cluster block ID.
+ * @param abs_cap The absolute capacity number of the sub-tile that the cluster block is mapped to.
+ */
 std::vector<int> get_cluster_block_pins(t_physical_tile_type_ptr physical_tile,
```
soheilshahrouz: Is this function used anywhere outside of RR graph generation?
If not, I suggest moving it to rr_graph_intra_cluster.cpp

soheilshahrouz: Why is `abs_cap` named that way?
It seems to be the subtile index to which the given clustered block is mapped.

vpr/src/util/vpr_utils.cpp
```diff
@@ -1909,17 +1909,47 @@ std::vector<int> get_cluster_netlist_intra_tile_pins_at_loc(const t_physical_til
...
-    int max_num_pin = get_tile_total_num_pin(physical_tile) / physical_tile->capacity;
-    int num_tile_pin_per_inst = physical_tile->num_pins / physical_tile->capacity;
-    std::vector<int> pin_num_vec(num_tile_pin_per_inst);
-    std::iota(pin_num_vec.begin(), pin_num_vec.end(), abs_cap * num_tile_pin_per_inst);
+    // To get the list of pins, we first add the pins on the tile-level, then add the pins on the intra-tile-level.
+
+    // A counter to keep track of number of tile-level pins for the sub-tiles before the current sub-tile.
+    int seen_sub_tiles_num_cluser_pins = 0;
+    // The number of tile-level pins for the sub-tile instance that the cluster block is mapped to.
+    int cluster_sub_tile_isnt_num_pins;
```
soheilshahrouz: typo
isnt --> inst

vpr/src/util/vpr_utils.cpp
```diff
@@ -1909,17 +1909,47 @@ std::vector<int> get_cluster_netlist_intra_tile_pins_at_loc(const t_physical_til
...
+    // A flag to check if the sub-tile instance that the cluster block is mapped to has been found.
+    bool found_sub_tile = false;
+
+    // Iterate over all the sub-tiles to find the sub-tile instance that the cluster block is mapped to.
+    for (const t_sub_tile& tmp_sub_tile: physical_tile->sub_tiles) {
```
soheilshahrouz: I suggest renaming `tmp_sub_tile` to just `sub_tile`

soheilshahrouz (commented):
Thansk Amin!

Is there a CI test for this?

amin1377:
@soheilshahrouz: Thanks! I’ve addressed your comments. We don’t have a test for this yet, but I’ve created an issue and will add one soon.

soheilshahrouz (approved):

PR #3339, opened 2025-11-10
Adding routing struggle metric

amin1377:
WIP

### DIFF
--- a/vpr/src/route/route.cpp
+++ b/vpr/src/route/route.cpp
@@ -582,6 +582,8 @@ bool route(const Netlist<>& net_list,
         // profiling::time_on_criticality_analysis();
     }
 
+    // Calculates the number of iterations with high heap ops.
+    // This is defined as an iteration that has more heap operations than the maximum of the previous three iterations.
     int num_high_heap_ops_iters = 0;
     if (per_iter_heap_ops_count.size() > 4) {
         for (int i = 3; i < static_cast<int>(per_iter_heap_ops_count.size()); i++) {


amin1377:
@vaughnbetz: I’ve included the results for various architectures and benchmarks. For most circuits, the routing struggle is zero, likely because, unlike in the paper, the utilization isn’t as high and there are ample routing resources available. Therefore, for the benchmarks included, this metric might not be particularly informative.

However, I did test it on some architectures and circuits where the router struggled with congestion, and in those cases, the routing struggle metric was above 20% (as mentioned in the paper). I don’t have a strong preference about whether to include this metric or not; I just thought it might be interesting to have a metric that allows us to compare how difficult routing was across different architectures and circuits.

vpr/src/route/route.cpp
```diff
@@ -578,6 +582,31 @@ bool route(const Netlist<>& net_list,
         // profiling::time_on_criticality_analysis();
     }
 
+    // Calculates the number of iterations with high heap ops.
+    // This is defined as an iteration that has more heap operations than the maximum of the previous three iterations.
+    int num_high_heap_ops_iters = 0;
+    if (per_iter_heap_ops_count.size() > 4) {
+        for (int i = 3; i < static_cast<int>(per_iter_heap_ops_count.size()); i++) {
```
AlexandreSinger, line 589: Minor comment for readability: Instead of casting the size into an int, why not make "i" and "j" both size_t type?

vpr/src/route/route.cpp
```diff
@@ -578,6 +582,31 @@ bool route(const Netlist<>& net_list,
         // profiling::time_on_criticality_analysis();
     }
 
+    // Calculates the number of iterations with high heap ops.
+    // This is defined as an iteration that has more heap operations than the maximum of the previous three iterations.
+    int num_high_heap_ops_iters = 0;
+    if (per_iter_heap_ops_count.size() > 4) {
```
AlexandreSinger, line 588: Why is this if statement needed? Is there a problem when the size is equal to 4? The below for loop will not execute if the size is 3 or less anyways due to the upper-bound condition.

vpr/src/route/route.cpp
```diff
@@ -578,6 +582,31 @@ bool route(const Netlist<>& net_list,
...
+                num_high_heap_ops_iters++;
+            }
+        }
+    }
+
+    // Calculates the routing struggle ratio.
+    // This ratio is defined as the number of iterations with high heap ops
+    // (i.e., iterations having more heap operations than the maximum of the previous three iterations
+    // divided by the total number of iterations.
+    // For more details, see TODO: routing struggle paper is not published yet.
```
AlexandreSinger, line 604: Has the paper not been published yet?

PR #3342, opened 2025-11-11
RR Graph - Cut interposer crossing wires

AmirhosseinPoolad:
This PR adds the last piece of functionality for modeling 2.5D architectures in upstream VPR: Actually cutting down the wires and connections that cross between the dice. 

I had an issue with code duplication and there are still some functions that look very similar but do different things. I tried my best to keep the duplication to the lowest level functions and have more general higher level functions that wrap them but there's still some remaining. If you have an idea to do this in a better way please let me know.

### DIFF
--- a/libs/librrgraph/src/base/rr_graph_builder.h
+++ b/libs/librrgraph/src/base/rr_graph_builder.h
@@ -352,10 +352,11 @@ class RRGraphBuilder {
         node_storage_.alloc_and_load_edges(rr_edges_to_create);
     }
     
-    /** @brief Removes a given list of RREdgeIds for the RR Graph.
+    /** @brief Removes a given list of RREdgeIds from the RR Graph.
      * This method does not preserve the order of edges. If you're
-     * calling it after partition_edges has been called, you need
-     * to call it again.
+     * calling it after partition_edges has been called, you will
+     * need to call partition_edges again.
+     * This operation is O(#RR Graph edges) and should not be called frequently.
      *
      * @param rr_edges_to_remove list of RREdgeIds to be removed
      */
--- a/libs/librrgraph/src/base/rr_graph_storage.h
+++ b/libs/librrgraph/src/base/rr_graph_storage.h
@@ -401,7 +401,7 @@ class t_rr_graph_storage {
     }
 
     /** @brief Returns a range of all edges in the RR Graph.
-     * This method does not depend on the edges begin correctly
+     * This method does not depend on the edges being correctly
      * sorted and can be used before partition_edges is called.
      */
     inline vtr::StrongIdRange<RREdgeId> all_edges() const {
@@ -784,10 +784,11 @@ class t_rr_graph_storage {
     /** @brief Adds a batch of edges.*/
     void alloc_and_load_edges(const t_rr_edge_info_set* rr_edges_to_create);
 
-    /** @brief Removes a given list of RREdgeIds for the RR Graph.
+    /** @brief Removes a given list of RREdgeIds from the RR Graph.
      * This method does not preserve the order of edges. If you're
-     * calling it after partition_edges has been called, you need
-     * to call it again.
+     * calling it after partition_edges has been called, you will
+     * need to call partition_edges again.
+     * This operation is O(#RR Graph edges) and should not be called frequently.
      *
      * @param rr_edges_to_remove list of RREdgeIds to be removed
      */
--- a/vpr/src/route/rr_graph_generation/rr_graph.cpp
+++ b/vpr/src/route/rr_graph_generation/rr_graph.cpp
@@ -1671,7 +1671,7 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
 
     // If there are any interposer cuts, remove the edges and shorten the wires that cross interposer cut lines.
     if (grid.has_interposer_cuts()) {
-        std::vector<RREdgeId> interposer_edges = mark_interposer_cut_edges_for_removal(rr_graph, grid);
+        std::vector<RREdgeId> interposer_edges = get_interposer_cut_edges_for_removal(rr_graph, grid);
         rr_graph_builder.remove_edges(interposer_edges);
 
         update_interposer_crossing_nodes_in_spatial_lookup_and_rr_graph_storage(rr_graph, grid, rr_graph_builder, sg_node_indices);
--- a/vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
+++ b/vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
@@ -1,3 +1,8 @@
+/**
+ * @file interposer_cut.h
+ * @brief This file implements functions can be used to generate RR Graphs for 2.5D architectures. See rr_graph_interposer.h for more details.
+ */
+
 #include <cstddef>
 #include <utility>
 #include <vector>
@@ -19,7 +24,7 @@
 // Static function declarations
 
 /**
- * @brief Takes location of a source and a sink and determines wether it crosses cut_loc or not.
+ * @brief Takes location of a source and a sink and determines whether it crosses cut_loc or not.
  * For example, the interval (1, 4) is cut by 3, while it is not cut by 5 or 0.
  */
 static bool should_cut(int src_loc, int sink_loc, int cut_loc);
@@ -35,11 +40,23 @@ static short node_xstart(const RRGraphView& rr_graph, RRNodeId node);
 static short node_ystart(const RRGraphView& rr_graph, RRNodeId node);
 
 /**
- * @brief Update a CHANY node's bounding box in RRGraph and SpatialLookup entries.
+ * @brief Update a CHANY node's bounding box in RR Graph and Spatial Lookup entries.
  * This function assumes that the channel node actually crosses the cut location and
  * might not function correctly otherwise.
  *
  * This is a low level function, you should use cut_channel_node that wraps this up in a nicer API.
+ *
+ * @param node Node's ID
+ * @param x_low Node's lower x position
+ * @param y_low Node's lower y position
+ * @param x_high Node's higher x position
+ * @param y_high Node's higher y position
+ * @param layer Node's layer
+ * @param ptc_num Node's ptc_num
+ * @param cut_loc_y y location of horizontal interposer cut line
+ * @param node_direction Node's direction
+ * @param rr_graph_builder RR Graph builder for mutating the RR Graph
+ * @param spatial_lookup RR Graph spatial lookup
  */
 static void cut_chan_y_node(RRNodeId node,
                             int x_low,
@@ -59,6 +76,18 @@ static void cut_chan_y_node(RRNodeId node,
  * might not function correctly otherwise.
  *
  * This is a low level function, you should use cut_channel_node that wraps this up in a nicer API.
+ *
+ * @param node Node's ID
+ * @param x_low Node's lower x position
+ * @param y_low Node's lower y position
+ * @param x_high Node's higher x position
+ * @param y_high Node's higher y position
+ * @param layer Node's layer
+ * @param ptc_num Node's ptc_num
+ * @param cut_loc_x x location of vertical interposer cut line
+ * @param node_direction Node's direction
+ * @param rr_graph_builder RR Graph builder for mutating the RR Graph
+ * @param spatial_lookup RR Graph spatial lookup
  */
 static void cut_chan_x_node(RRNodeId node,
                             int x_low,
@@ -167,11 +196,10 @@ static short node_ystart(const RRGraphView& rr_graph, RRNodeId node) {
     }
 }
 
-std::vector<RREdgeId> mark_interposer_cut_edges_for_removal(const RRGraphView& rr_graph, const DeviceGrid& grid) {
+std::vector<RREdgeId> get_interposer_cut_edges_for_removal(const RRGraphView& rr_graph, const DeviceGrid& grid) {
     std::vector<RREdgeId> edges_to_be_removed;
 
-    // Loop over all RREdgeIds and mark ones that cross a cutline to be removed
-
+    // Loop over all RREdgeIds and add ones that cross a cutline to a list for removal
     for (RREdgeId edge_id : rr_graph.all_edges()) {
         RRNodeId src_node = rr_graph.edge_src_node(edge_id);
         RRNodeId sink_node = rr_graph.edge_sink_node(edge_id);
@@ -190,6 +218,8 @@ std::vector<RREdgeId> mark_interposer_cut_edges_for_removal(const RRGraphView& r
 
         int layer = rr_graph.node_layer_low(src_node);
 
+        // If the start of the source node and the start of the sink node of this edge are on different sides of an interposer cut line
+        // the edge starts on a piece of wire that would be cut by the interposed and should be added to the removal list.
         for (int cut_loc_y : grid.get_horizontal_interposer_cuts()[layer]) {
             int src_start_loc_y = node_ystart(rr_graph, src_node);
             int sink_start_loc_y = node_ystart(rr_graph, sink_node);
@@ -223,8 +253,11 @@ static void cut_chan_y_node(RRNodeId node,
                             Direction node_direction,
                             RRGraphBuilder& rr_graph_builder,
                             RRSpatialLookup& spatial_lookup) {
+
+    VTR_ASSERT_SAFE(should_cut(y_low, y_high, cut_loc_y));
+
     if (node_direction == Direction::INC) {
-        // Anything above cut_loc_y shouldn't exist
+        // This node should not exist at locations above cut_loc_y
         rr_graph_builder.set_node_coordinates(node, x_low, y_low, x_high, cut_loc_y);
 
         // Do a loop from cut_loc_y to y_high and remove node from spatial lookup
@@ -232,7 +265,7 @@ static void cut_chan_y_node(RRNodeId node,
             spatial_lookup.remove_node(node, layer, x_low, y_loc, e_rr_type::CHANY, ptc_num);
         }
     } else if (node_direction == Direction::DEC) {
-        // Anything below cut_loc_y (inclusive) shouldn't exist
+        // This node should not exist at locations below cut_loc_y (inclusive)
         rr_graph_builder.set_node_coordinates(node, x_low, cut_loc_y + 1, x_high, y_high);
 
         // Do a loop from y_low to cut_loc_y and remove node from spatial lookup
@@ -255,8 +288,10 @@ static void cut_chan_x_node(RRNodeId node,
                             Direction node_direction,
                             RRGraphBuilder& rr_graph_builder,
                             RRSpatialLookup& spatial_lookup) {
+    VTR_ASSERT_SAFE(should_cut(x_low, x_high, cut_loc_x));
+
     if (node_direction == Direction::INC) {
-        // Anything to the right of cut_loc_x shouldn't exist
+        // This node should not exist at locations to the right of cut_loc_x
         rr_graph_builder.set_node_coordinates(node, x_low, y_low, cut_loc_x, y_high);
 
         // Do a loop from cut_loc_x to x_high and remove node from spatial lookup
@@ -264,7 +299,7 @@ static void cut_chan_x_node(RRNodeId node,
             spatial_lookup.remove_node(node, layer, x_loc, y_low, e_rr_type::CHANX, ptc_num);
         }
     } else if (node_direction == Direction::DEC) {
-        // Anything to the left of cut_loc_x (inclusive) shouldn't exist
+        // This node should not exist at locations to the left of cut_loc_x (inclusive)
         rr_graph_builder.set_node_coordinates(node, cut_loc_x + 1, y_low, x_high, y_high);
 
         // Do a loop from x_low to cut_loc_x - 1 and remove node from spatial lookup
--- a/vpr/src/route/rr_graph_generation/rr_graph_interposer.h
+++ b/vpr/src/route/rr_graph_generation/rr_graph_interposer.h
@@ -3,9 +3,53 @@
 /**
  * @file interposer_cut.h
  * @brief This file implements functions that:
- * (1) Marks all edges that cross an interposer cut for removal
+ *
+ * (1) Get a list of all edges that cross an interposer cut for removal. Edges whose source and sink nodes start on opposite
+ * sides of an interposer cut line are considered to cross the cut line and will be removed.
+ *
  * (2) Makes the channel nodes that cross an interposer cut shorter to have them not cross the interposer anymore
+ *
  * Using these two functions and combined with 2D scatter-gather patterns, you can model and implement 2.5D FPGA RR Graphs.
+ *
+ * Below you can see a diagram of a device with a vertical interposer cut at x = 2 (Shown with X) and a horizontal interposer cut at y = 2 (shown with =)
+ * X(x,y) denotes the ChanX at location (x,y) and Y(x,y) denotes the ChanX at location (x,y). Consistent with VPR's coordinate system, each tile "owns"
+ * the channels above and to the right of it, therefore interposer cut lines will leave the channels on the side of the tile it belongs to.
+ *
+ *
+ *     X(0,3)           X(1,3)        X  X(2,3)           X(3,3)           X(4,3)
+ *    _________        _________      X _________        _________        _________
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    |       | Y(0,3) |       |Y(1,3)X |       | Y(2,3) |       | Y(3,3) |       | Y(4,3)
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    ---------        ---------      X ---------        ---------        ---------
+ *                                    X
+ *     X(0,2)           X(1,2)        X  X(2,2)           X(3,2)           X(4,2)
+ *                                    X
+ *    _________        _________      X _________        _________        _________
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    |       | Y(0,2) |       |Y(1,2)X |       | Y(2,2) |       | Y(3,2) |       | Y(4,2)
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    ---------        ---------      X ---------        ---------        ---------
+ * ===================================X=================================================
+ *     X(0,1)           X(1,1)        X  X(2,1)           X(3,1)           X(4,1)
+ *                                    X
+ *    _________        _________      X _________        _________        _________
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    |       | Y(0,1) |       |Y(1,1)X |       | Y(2,1) |       | Y(3,1) |       | Y(4,1)
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    ---------        ---------      X ---------        ---------        ---------
+ *                                    X
+ *     X(0,0)           X(1,0)        X  X(2,0)           X(3,0)           X(4,0)
+ *                                    X
+ *    _________        _________      X _________        _________        _________
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    |       | Y(0,0) |       |Y(1,0)X |       | Y(2,0) |       | Y(3,0) |       | Y(4,0)
+ *    |       |        |       |      X |       |        |       |        |       |
+ *    ---------        ---------      X ---------        ---------        ---------
+ *                                    X
+ *                                    X
+ *
+ *
  */
 
 #include <vector>
@@ -15,14 +59,22 @@
 
 /**
  * @brief Goes through all edges in the RR Graph and returns a list of the edges that cross an interposer cut line.
+ * Edges whose source and sink nodes start on opposite sides of an interposer cut line are considered to cross the
+ * cut line and will be removed.
  * 
  * @return std::vector<RREdgeId> List of all edges that cross an interposer cut line.
  */
-std::vector<RREdgeId> mark_interposer_cut_edges_for_removal(const RRGraphView& rr_graph, const DeviceGrid& grid);
+std::vector<RREdgeId> get_interposer_cut_edges_for_removal(const RRGraphView& rr_graph, const DeviceGrid& grid);
 
 /**
- * @brief Shortens the channel nodes that cross an interposer cut line
+ * @brief Shortens the channel nodes that cross an interposer cut line. 
  * 
+ * @details When a wire crosses an interposer cut line, it is shortened to instead end at the interposed cut line
+ * unless it is in the list of sg_node_indices. The wire's drive point and the wire up to the interposer cut is
+ * retained and will not be modified. "Shortening a wire" in this context means modifying the information in the
+ * RR Graph itself (Changing the bounding box of the corresponding RR Graph node) and the spatial lookup
+ * (Making sure there is no reference to the shortened wire in locations it doesn't exist at anymore.)
+ *
  * @param rr_graph RRGraphView, used to read the RR Graph.
  * @param grid Device grid, used to access interposer cut locations.
  * @param rr_graph_builder RRGraphBuilder, to modify the RRGraph.


libs/librrgraph/src/base/rr_graph_builder.h
```diff
@@ -352,6 +352,10 @@ class RRGraphBuilder {
         node_storage_.alloc_and_load_edges(rr_edges_to_create);
     }
 
+    inline void remove_edges(std::vector<RREdgeId>& rr_edges_to_remove) {
```
AlexandreSinger, line 363: Should add dox string to match the rest of the document.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -57,6 +59,33 @@ void t_rr_graph_storage::alloc_and_load_edges(const t_rr_edge_info_set* rr_edges
     }
 }
 
+void t_rr_graph_storage::remove_edges(std::vector<RREdgeId>& rr_edges_to_remove) {
+    size_t starting_edge_count = edge_dest_node_.size();
+
+    vtr::uniquify(rr_edges_to_remove);
+    
+    size_t edge_list_end = edge_dest_node_.size() - 1;
```
AlexandreSinger, line 76: I think this code (including the loop) deserves a small comment. I am struggling to follow what this code is doing a bit.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -400,6 +400,11 @@ class t_rr_graph_storage {
         return vtr::StrongIdRange<RREdgeId>(first_edge(id), last_edge(id));
     }
 
+
+    inline vtr::StrongIdRange<RREdgeId> all_edges() const {
```
AlexandreSinger, line 407: This needs a dox string.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -776,6 +781,8 @@ class t_rr_graph_storage {
     /** @brief Adds a batch of edges.*/
     void alloc_and_load_edges(const t_rr_edge_info_set* rr_edges_to_create);
 
+    void remove_edges(std::vector<RREdgeId>& rr_edges_to_remove);
```
AlexandreSinger, line 795: Dox string.

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -584,14 +586,19 @@ class RRGraphView {
...
-     * for (RREdgeId edge : rr_graph.edges(node)) {
+     * for (t_edge_size edge : rr_graph.edges(node)) {
      *     // Do something with the edge
      * }
      */
     inline edge_idx_range edges(const RRNodeId& id) const {
         return vtr::make_range(edge_idx_iterator(0), edge_idx_iterator(num_edges(id)));
     }
 
+    inline vtr::StrongIdRange<RREdgeId> all_edges() const {
```
AlexandreSinger, line 603: Dox string.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,271 @@
...
+    // Return early for OPIN and IPIN types (Some BIDIR pins would trigger the assertion below)
+    if (rr_graph.node_type(node) == e_rr_type::OPIN || rr_graph.node_type(node) == e_rr_type::IPIN) {
+        VTR_ASSERT(rr_graph.node_xlow(node) == rr_graph.node_xhigh(node));
+        return rr_graph.node_xlow(node);
+    }
+
+    switch (rr_graph.node_direction(node)) {
+        case Direction::DEC:
+            return rr_graph.node_xhigh(node);
+            break;
```
AlexandreSinger: These breaks are not necessary since return is an end in control sequence. Not a huge deal, but technically its good style to remove since these lines of code cannot ever be hit.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
```diff
@@ -0,0 +1,271 @@
...
+            return (rr_graph.node_xlow(node));
+            break;
+
+        case Direction::BIDIR:
+            VTR_ASSERT_MSG(false, "Bidir node has no starting point");
+            break;
+
+        default:
+            VTR_ASSERT(false);
+            break;
```
AlexandreSinger, line 163: This is invalid and will crash if the assert level is set too low I think...

If the assert level is set low enough, this function will not return anything. You should consider making this a stronger error or returning something.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,271 @@
...
+static short node_ystart(const RRGraphView& rr_graph, RRNodeId node) {
+    // Return early for OPIN and IPIN types (Some BIDIR pins would trigger the assertion below)
+    if (rr_graph.node_type(node) == e_rr_type::OPIN || rr_graph.node_type(node) == e_rr_type::IPIN) {
+        return rr_graph.node_ylow(node);
+    }
+
+    switch (rr_graph.node_direction(node)) {
+        case Direction::DEC:
+            return rr_graph.node_yhigh(node);
+            break;
```
AlexandreSinger: Same comments as above.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,271 @@
...
+    std::vector<RREdgeId> edges_to_be_removed;
+
+    // Loop over all RREdgeIds and mark ones that cross a cutline to be removed
+
+    for (RREdgeId edge_id : rr_graph.all_edges()) {
+        RRNodeId src_node = rr_graph.edge_src_node(edge_id);
+        RRNodeId sink_node = rr_graph.edge_sink_node(edge_id);
+
+        if (src_node == RRNodeId(5866) && sink_node == RRNodeId(5604)) {
+            VTR_LOG("HI\n");
```
AlexandreSinger: Hi :)

(Remove temp code).

vpr/src/route/rr_graph_generation/rr_graph.h
```diff
@@ -17,7 +17,7 @@ enum {
     RR_GRAPH_NO_WARN = 0x00,
     RR_GRAPH_WARN_FC_CLIPPED = 0x01,
     RR_GRAPH_WARN_CHAN_X_WIDTH_CHANGED = 0x02,
-    RR_GRAPH_WARN_CHAN_Y_WIDTH_CHANGED = 0x03
+    RR_GRAPH_WARN_CHAN_Y_WIDTH_CHANGED = 0x04
```
AlexandreSinger: Why change this value? Is there a benefit to making this one-hot?

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
```diff
@@ -496,111 +498,104 @@ bool verify_rr_node_indices(const DeviceGrid& grid,
                             const t_rr_graph_storage& rr_nodes,
                             bool is_flat) {
     std::unordered_map<RRNodeId, int> rr_node_counts;
+    for (t_physical_tile_loc tile_loc : grid.all_locations()) {
```
AlexandreSinger: I think this should be a const reference. Tile_locs are not simple types I think.

vpr/src/route/rr_graph_generation/rr_graph.h
AmirhosseinPoolad: It's used in the code as if it was one hot (you can see people OR-ing RR_GRAPH_WARN_CHAN_X_WIDTH_CHANGED into a variable and using it as a bit mask throughout the code), but it wasn't actually one hot. I think it was a bug from when RR Graph generation was modified to support different CHANY and CHANX.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
AmirhosseinPoolad: Ah! totally missed this.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
AmirhosseinPoolad: I know but I always liked to have the breaks anyways in case someone decides to change this to something more complicated in the future. I'll remove them if it's more common to not have the break in the codebase.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
AmirhosseinPoolad: The intended behaviour is crashing if this function is called with invalid inputs. Here it's even worse than just invalid input since it means there's an RR Node with direction of Direction::NUM_DIRECTIONS. At that point the entire RR Graph generation code is FUBAR. Anyway, we don't really have an assertion level that will always trigger. even VTR_ASSERT_OPT is removed with VTR_ASSERT_LEVEL == 0. I think it's fine to stick to the default assert level here.

vpr/src/route/rr_graph_generation/rr_node_indices.cpp
AmirhosseinPoolad: It's just 3 integers but yeah you're right it should be const ref.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
AlexandreSinger: Why not use a proper error, like VTR_ERROR or VPR_ERROR?

vpr/src/route/rr_graph_generation/rr_graph.h
AlexandreSinger: You should have a comment here to document it. Just in case someone wants to modify it.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
AlexandreSinger: Its really up to you; generally, I try not to add code which will never get hit. It hurts the code coverage lol.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
AmirhosseinPoolad: Makes sense!

vpr/src/route/rr_graph_generation/rr_graph.h
AmirhosseinPoolad: Added a comment and changed the coding style to make it super obvious it's a bit mask.

doc/src/arch/reference.rst
```diff
@@ -2751,7 +2750,7 @@ The number of any additional wires or muxes created by scatter-gather specificat
     Overview of how scatter-gather patterns work. First, connections from a switchblock location are selected according to the specification.
     These selected connection are then muxed and passed through the scatter-gather node, which is typically a wire segment. The scatter-gather node then fans out or scatters in another switchblock location.
 
-.. note:: Scatter-Gather patterns are only supported for 3D architectures where the scatter-gather links are unidirectional. They are not currently supported in 2D architectures or with bidirectional sg_links.
+.. note:: Scatter-Gather patterns are only supported for 3D architectures where the scatter-gather links are uni-directional and uni-directional 2D architectures. Bidirectional sg_links are not currently supported.
```
amin1377: I think scatter-gather tag is not specific to 3D architecture.

libs/librrgraph/src/base/rr_graph_builder.h
```diff
@@ -351,6 +351,17 @@ class RRGraphBuilder {
...
+    
+    /** @brief Removes a given list of RREdgeIds for the RR Graph.
+     * This method does not preserve the order of edges. If you're
+     * calling it after partition_edges has been called, you need
+     * to call it again.
+     *
+     * @param rr_edges_to_remove list of RREdgeIds to be removed
+     */
+    inline void remove_edges(std::vector<RREdgeId>& rr_edges_to_remove) {
+        node_storage_.remove_edges(rr_edges_to_remove);
```
amin1377, line 364: I think we set a flag after partitioning edges. I'd suggest adding an assertion to ensure that the flag is not set when this function is called.

libs/librrgraph/src/base/rr_graph_view.cpp
```diff
@@ -129,5 +129,3 @@ bool RRGraphView::validate_in_edges() const {
     }
     return true;
 }
```
amin1377, line 131: I think @AlexandreSinger would like to see an empty line at the end. :)

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -584,14 +586,23 @@ class RRGraphView {
      * @example
      * RRGraphView rr_graph; // A dummy rr_graph for a short example
      * RRNodeId node; // A dummy node for a short example
-     * for (RREdgeId edge : rr_graph.edges(node)) {
+     * for (t_edge_size edge : rr_graph.edges(node)) {
      *     // Do something with the edge
      * }
      */
     inline edge_idx_range edges(const RRNodeId& id) const {
```
amin1377: I think the comments above this function don’t clearly explain that the returned ID is not the RREdgeId, but rather the index among the node’s outgoing edges. If you could update the comments to make that distinction clearer, and also show one of the functions that uses this index instead of RREdgeId, that would be great.

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -1667,6 +1670,13 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
         }
     }
 
+    if (!grid.get_horizontal_interposer_cuts().empty() || !grid.get_vertical_interposer_cuts().empty()) {
```
amin1377: I think it would be better if you add a comment explaining what is happening here. It would be helpful to mention why the extra edges are created and why the following step (removing them) is necessary.

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -1667,6 +1670,13 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
...
+        std::vector<RREdgeId> interposer_edges = mark_interposer_cut_edges_for_removal(rr_graph, grid);
+        rr_graph_builder.remove_edges(interposer_edges);
+
+        update_interposer_crossing_nodes_in_spatial_lookup_and_rr_graph_storage(rr_graph, grid, rr_graph_builder, sg_node_indices);
+    }
+
     add_and_connect_non_3d_sg_links(rr_graph_builder, sg_links, sg_node_indices, chan_details_x, chan_details_y, num_seg_types_x, rr_edges_to_create);
```
amin1377, line 1682: In the comment above, please also include that all edges crossing the interposer line are removed, and that this function adds the inter-die connections based on the sg tag.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
```diff
@@ -0,0 +1,266 @@
...
+std::vector<RREdgeId> mark_interposer_cut_edges_for_removal(const RRGraphView& rr_graph, const DeviceGrid& grid) {
+    std::vector<RREdgeId> edges_to_be_removed;
+
+    // Loop over all RREdgeIds and mark ones that cross a cutline to be removed
+
+    for (RREdgeId edge_id : rr_graph.all_edges()) {
+        RRNodeId src_node = rr_graph.edge_src_node(edge_id);
+        RRNodeId sink_node = rr_graph.edge_sink_node(edge_id);
+
+        // TODO: ignoring ChanZ nodes for now
```
amin1377, line 207: Can you elaborate a bit more on why chanz is ignored? My guess is that since this code is focused on the 2.5D architecture, the assumption is that chanz is not stretched across the 2D plane but along the Z-axis, so the interposer line wouldn’t apply to it. If that’s the case, I’d also add an ASSERT_DEBUG here to ensure that x_low == x_high and y_low == y_high.

AmirhosseinPoolad: I haven't thought about 3D architectures that also have interposer cut lines so I'm honestly not sure what should happen to edges to connect to/from chanz nodes. I think this needs further discussions about what it even means to have an architecture like this.

libs/librrgraph/src/base/rr_graph_view.cpp
AmirhosseinPoolad: I'm pretty sure the CI make format test has been broken for the past couple of months, because this really shouldn't have passed CI. Also if you run make format on master you'll see a ton of changes. @AlexandreSinger Can you check this? I looked into it and couldn't really find out what's going on.

AlexandreSinger: I think I found it: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3344

Testing it now.

FYI, I did some research about the newline characters at the end of files and found something annoying. Many editors now imply that the new line character is at the end of the file and does not show it. So, even though it looks like there is no end of line character on this file there may be! This is why we need this in the auto-formatter unfortunately. Almost all modern IDE automatically put newlines at the end of files I think; but I need to do more looking around to be sure.

AlexandreSinger: Fixed. Once that PR comes into Master, the CI will be checking for CPP and Python formatting again.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -57,6 +59,44 @@ void t_rr_graph_storage::alloc_and_load_edges(const t_rr_edge_info_set* rr_edges
...
+
+    // Sort and make sure all edge indices are unique
+    vtr::uniquify(rr_edges_to_remove);
+    VTR_ASSERT_SAFE(std::is_sorted(rr_edges_to_remove.begin(), rr_edges_to_remove.end()));
+    
+    // Index of the last edge
+    size_t edge_list_end = edge_dest_node_.size() - 1;
+
+    // Iterate backwards through the list of indices we want to remove.
+    for (auto it = rr_edges_to_remove.rbegin(); it != rr_edges_to_remove.rend(); ++it) {
```
soheilshahrouz: range-based for loop with std::views::reverse

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -57,6 +59,44 @@ void t_rr_graph_storage::alloc_and_load_edges(const t_rr_edge_info_set* rr_edges
     }
 }
 
+void t_rr_graph_storage::remove_edges(std::vector<RREdgeId>& rr_edges_to_remove) {
+    size_t starting_edge_count = edge_dest_node_.size();
+
+    // Sort and make sure all edge indices are unique
+    vtr::uniquify(rr_edges_to_remove);
+    VTR_ASSERT_SAFE(std::is_sorted(rr_edges_to_remove.begin(), rr_edges_to_remove.end()));
```
soheilshahrouz, line 70: I suggest adding an assert to check that the last ID is within a valid range.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
soheilshahrouz, line 1: I think rr_graph_interposer is more consistent with the naming style of the other files in this directory.

libs/librrgraph/src/base/rr_graph_builder.h
```diff
@@ -351,6 +351,17 @@ class RRGraphBuilder {
     inline void alloc_and_load_edges(const t_rr_edge_info_set* rr_edges_to_create) {
         node_storage_.alloc_and_load_edges(rr_edges_to_create);
     }
+    
+    /** @brief Removes a given list of RREdgeIds for the RR Graph.
+     * This method does not preserve the order of edges. If you're
+     * calling it after partition_edges has been called, you need
```
vaughnbetz: "you need to call it again" --> you will need to call partition_edges again

libs/librrgraph/src/base/rr_graph_builder.h
vaughnbetz: I think the routine is legal to call after partition_edges; it should just unset the edges are partitioned flag.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -57,6 +59,44 @@ void t_rr_graph_storage::alloc_and_load_edges(const t_rr_edge_info_set* rr_edges
...
+
+    // We have a new index to the end of the list, call erase on the elements past that index
+    // to update the std::vector and shrink the actual data structures.
+    edge_dest_node_.erase(edge_dest_node_.begin() + edge_list_end + 1, edge_dest_node_.end());
+    edge_src_node_.erase(edge_src_node_.begin() + edge_list_end + 1, edge_src_node_.end());
+    edge_switch_.erase(edge_switch_.begin() + edge_list_end + 1, edge_switch_.end());
+    edge_remapped_.erase(edge_remapped_.begin() + edge_list_end + 1, edge_remapped_.end());
+
+    VTR_ASSERT(edge_dest_node_.size() == (starting_edge_count - rr_edges_to_remove.size()));
+}
```
vaughnbetz, line 104: Should set partitioned = false here

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -400,6 +400,14 @@ class t_rr_graph_storage {
         return vtr::StrongIdRange<RREdgeId>(first_edge(id), last_edge(id));
     }
 
+    /** @brief Returns a range of all edges in the RR Graph.
+     * This method does not depend on the edges begin correctly
```
vaughnbetz: begin -> being

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -776,6 +784,15 @@ class t_rr_graph_storage {
     /** @brief Adds a batch of edges.*/
     void alloc_and_load_edges(const t_rr_edge_info_set* rr_edges_to_create);
 
+    /** @brief Removes a given list of RREdgeIds for the RR Graph.
```
vaughnbetz: 'for the RR' -> 'from the RR'
'has been called, you need to call it again' -> 'has been called, you need to call partition_edges again'

vaughnbetz: Probably should say this operation is O(num_edges) in the RR graph, so it should not be called often and edges should be removed in bulk if at all possible.

vpr/src/route/rr_graph_generation/rr_graph_interposer.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include <vector>
+#include "rr_graph_fwd.h"
+#include "rr_graph_view.h"
+#include "device_grid.h"
+
+/**
+ * @brief Goes through all edges in the RR Graph and returns a list of the edges that cross an interposer cut line.
```
vaughnbetz, line 61: What's the definition of an 'edge' that crosses an interposer cut line?  I think you need to define that here, since really it is parts of nodes that are cut, and the dangling edges are removed.

vaughnbetz: If this returns a list of edges (which I think it does) without marking them elsewhere, you should rename it 'get_interposer_cut_edges_for_removal'

vpr/src/route/rr_graph_generation/rr_graph_interposer.h
```diff
@@ -0,0 +1,23 @@
...
+ * 
+ * @return std::vector<RREdgeId> List of all edges that cross an interposer cut line.
+ */
+std::vector<RREdgeId> mark_interposer_cut_edges_for_removal(const RRGraphView& rr_graph, const DeviceGrid& grid);
+
+/**
+ * @brief Shortens the channel nodes that cross an interposer cut line
```
vaughnbetz: Explain what it does more (when a wire crosses an interposer cut line it is shortened to instead end at the interposer cut line, unless it is in the list of sg_node_indices that we should not cut).  Its drive point and the wire up to the interposer is retained.  Does this only work for unidirectional architectures / wires?  If so, say that.

amin1377 (approved):

libs/librrgraph/src/base/rr_graph_view.h
```diff
@@ -584,14 +586,25 @@ class RRGraphView {
...
      * RRGraphView rr_graph; // A dummy rr_graph for a short example
      * RRNodeId node; // A dummy node for a short example
-     * for (RREdgeId edge : rr_graph.edges(node)) {
+     * for (t_edge_size edge : rr_graph.edges(node)) {
      *     // Do something with the edge
      * }
+     *
+     * @note Iterating on the range returned by this function will not give you an RREdgeId, but instead gives you the index among a node's outgoing edges
      */
     inline edge_idx_range edges(const RRNodeId& id) const {
```
soheilshahrouz: pass the argument by value

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include "rr_graph_view.h"
+#include "rr_node_types.h"
+#include "rr_spatial_lookup.h"
+#include "vpr_error.h"
+#include "vtr_assert.h"
+
+#include "interposer_cut.h"
+
+/**
+ * @brief Takes location of a source and a sink and determines wether it crosses cut_loc or not. For example, the interval (1, 4) is cut by 3, while it is not cut by 5 or 0.
```
soheilshahrouz: We don't have a line width limit, but I suggest avoiding lines longer than 120 characters. Fort this, I needed to scroll horizontally.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
```diff
@@ -0,0 +1,266 @@
...
+        return false;
+    } else {
+        return true;
+    }
+}
+
+/**
+ * @brief Calculates the starting x point of node based on it's directionality.
+ */
+static short node_xstart(const RRGraphView& rr_graph, RRNodeId node) {
```
soheilshahrouz, line 134: `get_adjusted_rr_position()` already implements this logic, but it is a static function.
Might worth it to re-use that function.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
```diff
@@ -0,0 +1,266 @@
...
+        default:
+            VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Invalid RR node direction.");
+            break;
+    }
+}
+
+/**
+ * @brief Calculates the starting y point of node based on it's directionality.
+ */
+static short node_ystart(const RRGraphView& rr_graph, RRNodeId node) {
```
soheilshahrouz, line 167: `get_adjusted_rr_position`

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
soheilshahrouz: The coding style guide recommends placing the Doxygen comment on the declaration of a static function.

If all static functions are declared at the top of the file, their definition order no longer matters. However, without declarations, any new static function must appear after all the functions it calls. This makes adding new functions to the file a bit challenging.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+/**
+ * @brief Update a CHANY node's bounding box in RRGraph and SpatialLookup entries.
+ * This function assumes that the channel node actually crosses the cut location and
+ * might not function correctly otherwise.
+ *
+ * This is a low level function, you should use cut_channel_node that wraps this up in a nicer API.
+ */
+static void cut_chan_y_node(RRNodeId node, int x_low, int y_low, int x_high, int y_high, int layer, int ptc_num, int cut_loc_y, Direction node_direction, RRGraphBuilder& rr_graph_builder, RRSpatialLookup& spatial_lookup) {
```
soheilshahrouz: too long

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+/**
+ * @brief Update a CHANX node's bounding box in RRGraph and SpatialLookup entries.
+ * This function assumes that the channel node actually crosses the cut location and
+ * might not function correctly otherwise.
+ *
+ * This is a low level function, you should use cut_channel_node that wraps this up in a nicer API.
+ */
+static void cut_chan_x_node(RRNodeId node, int x_low, int y_low, int x_high, int y_high, int layer, int ptc_num, int cut_loc_x, Direction node_direction, RRGraphBuilder& rr_graph_builder, RRSpatialLookup& spatial_lookup) {
```
soheilshahrouz: too long

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * @brief Update a CHANX or CHANY node's bounding box in RRGraph and SpatialLookup entries if it crosses cut_loc
+ * 
+ * @param node Channel segment RR graph node that might cross the interposer cut line
+ * @param cut_loc location of vertical interposer cut line
+ * @param interposer_cut_type Type of the interposer cut line (Horizontal or vertical)
+ * @param sg_node_indices Sorted list of scatter-gather node IDs. We do not want to cut these nodes as they're allowed to cross an interposer cut line.
+ */
+static void cut_channel_node(RRNodeId node, int cut_loc, e_interposer_cut_type interposer_cut_type, const RRGraphView& rr_graph, RRGraphBuilder& rr_graph_builder, RRSpatialLookup& spatial_lookup, const std::vector<std::pair<RRNodeId, int>>& sg_node_indices) {
```
soheilshahrouz: too long

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    if (interposer_cut_type == e_interposer_cut_type::HORZ) {
+        cut_chan_y_node(node, x_low, y_low, x_high, y_high, layer, ptc_num, cut_loc, node_direction, rr_graph_builder, spatial_lookup);
+    } else if (interposer_cut_type == e_interposer_cut_type::VERT) {
+        cut_chan_x_node(node, x_low, y_low, x_high, y_high, layer, ptc_num, cut_loc, node_direction, rr_graph_builder, spatial_lookup);
+    }
+}
+
+void update_interposer_crossing_nodes_in_spatial_lookup_and_rr_graph_storage(const RRGraphView& rr_graph, const DeviceGrid& grid, RRGraphBuilder& rr_graph_builder, const std::vector<std::pair<RRNodeId, int>>& sg_node_indices) {
```
soheilshahrouz: the line is too long
the function name is too long

vpr/src/route/rr_graph_generation/interposer_cut.h
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @brief Shortens the channel nodes that cross an interposer cut line
+ * 
+ * @param rr_graph RRGraphView, used to read the RR Graph.
+ * @param grid Device grid, used to access interposer cut locations.
+ * @param rr_graph_builder RRGraphBuilder, to modify the RRGraph.
+ * @param sg_node_indices list of scatter-gather node IDs. We do not want to cut these nodes as they're allowed to cross an interposer cut line.
+ */
+void update_interposer_crossing_nodes_in_spatial_lookup_and_rr_graph_storage(const RRGraphView& rr_graph, const DeviceGrid& grid, RRGraphBuilder& rr_graph_builder, const std::vector<std::pair<RRNodeId, int>>& sg_node_indices);
```
soheilshahrouz: too long

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -46,6 +47,8 @@
 #include "rr_types.h"
 #include "rr_node_indices.h"
 
+#include "interposer_cut.h"
```
soheilshahrouz: add it somewhere next to some related inclusions

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -1667,6 +1670,15 @@ static std::function<void(t_chan_width*)> alloc_and_load_rr_graph(RRGraphBuilder
...
 
+    // If there are any interposer cuts, remove the edges and shorten the wires that cross interposer cut lines.
+    if (!grid.get_horizontal_interposer_cuts().empty() || !grid.get_vertical_interposer_cuts().empty()) {
+        std::vector<RREdgeId> interposer_edges = mark_interposer_cut_edges_for_removal(rr_graph, grid);
+        rr_graph_builder.remove_edges(interposer_edges);
+
+        update_interposer_crossing_nodes_in_spatial_lookup_and_rr_graph_storage(rr_graph, grid, rr_graph_builder, sg_node_indices);
+    }
+
+    // Add 2D scatter-gather link edges (the nodes have already been created at this point). These links are mostly used for interposer-crossing connections, but could also be used for other things.
```
soheilshahrouz: break the comment into multiple lines

vpr/src/route/rr_graph_generation/rr_graph.h
```diff
@@ -11,13 +11,15 @@
 #include "rr_graph_type.h"
 #include "clb2clb_directs.h"
 
-/* Warnings about the routing graph that can be returned.
- * This is to avoid output messages during a value sweep */
+/** @brief Warnings about the routing graph that can be returned.
+ * This is to avoid output messages during a value sweep.
```
vaughnbetz, line 15: If you know what these are, comment what each value controls.  
Is value sweep a channel width sweep?

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
AmirhosseinPoolad: It doesn't quite work the same way. get_adjusted_rr_position does return the driver's position for uni-directional chanxy nodes, but it also returns the midpoint of bidirectional wires (I want to crash if this happens). I also return OPIN/IPIN positions without changing them while get_adjusted_rr_position adjusts those position to be the same as their adjacent channel (e.g. left-side pins would return (x-1, y)).

did some slight refactoring of the functions to make them more readable.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
vaughnbetz: wether -> whether

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
vaughnbetz: I'd update the comment to say that we don't allow interposers in the Z dimension (z is for die-stacking, and an interposer in a die stacking dimension is not currently manufacturable).

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
```diff
@@ -0,0 +1,266 @@
...
+        // Currently 3D connection block edges cross layers. We don't want to cut these edges.
+        if (rr_graph.node_layer_low(src_node) != rr_graph.node_layer_low(sink_node)) {
+            continue;
+        }
+        VTR_ASSERT(rr_graph.node_layer_low(src_node) == rr_graph.node_layer_high(src_node));
+        VTR_ASSERT(rr_graph.node_layer_low(sink_node) == rr_graph.node_layer_high(sink_node));
+
+        int layer = rr_graph.node_layer_low(src_node);
+
+        for (int cut_loc_y : grid.get_horizontal_interposer_cuts()[layer]) {
```
vaughnbetz, line 223: Explain this a bit.  If the start (drive point) of the src node of the edge is on a different side of the interposer than the sink node of the edge drive point, the edge starts on a piece of wire that should be cut by the interposer. Mark it for removal.

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    return edges_to_be_removed;
+}
+
+/**
+  * @brief Update a CHANY node's bounding box in RRGraph and SpatialLookup entries.
+  * This function assumes that the channel node actually crosses the cut location and
+  * might not function correctly otherwise.
+  *
```
vaughnbetz: comment the params -- there are a lot so good to document.

vaughnbetz: Could you put an assert in for the channel node crosses the cut location?

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+}
+
+/**
+  * @brief Update a CHANX node's bounding box in RRGraph and SpatialLookup entries.
+  * This function assumes that the channel node actually crosses the cut location and
+  * might not function correctly otherwise.
+  *
+  * This is a low level function, you should use cut_channel_node that wraps this up in a nicer API.
+  */
```
vaughnbetz: Document params

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+static void cut_chan_x_node(RRNodeId node, int x_low, int y_low, int x_high, int y_high, int layer, int ptc_num, int cut_loc_x, Direction node_direction, RRGraphBuilder& rr_graph_builder, RRSpatialLookup& spatial_lookup) {
+    if (node_direction == Direction::INC) {
+        // Anything to the right of cut_loc_x shouldn't exist
+        rr_graph_builder.set_node_coordinates(node, x_low, y_low, cut_loc_x, y_high);
+
```
vaughnbetz, line 296: Add assert that the node crosses the cut location (or just return without doing anything and update the comment).

vpr/src/route/rr_graph_generation/interposer_cut.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include "device_grid.h"
+#include "interposer_types.h"
+#include "rr_graph_builder.h"
+#include "rr_graph_fwd.h"
+#include "rr_graph_view.h"
+#include "rr_node_types.h"
+#include "rr_spatial_lookup.h"
+#include "vpr_error.h"
+#include "vtr_assert.h"
+
```
vaughnbetz, line 21: Comment what this file is for.
Add an ascii art showing where the interposer boundary is / what wires are cut / where channels are vs. the interposer for the same coordinate (e.g. y).

vaughnbetz (changes requested):
Looks good; some commenting upgrades and a few changes suggested.

libs/librrgraph/src/base/rr_graph_builder.h
AmirhosseinPoolad: We actually have 2 flags, one is "edges_read_" and the other is "partitioned_". This method should definitely set partitioned_ to false, but if edges have been read, using this method would make the read edge indices invalid which would be very dangerous. I think the current behaviour (assert edges_read_ equals false) plus setting partitioned_ to false would be correct.

vpr/src/route/rr_graph_generation/rr_graph.h
AmirhosseinPoolad: I did a quick check, this enum and the warning system is completely unused. The RR Graph generation code does set these warning bits, but the results are not used anywhere.  I think I'll do a follow up PR to remove it.

vpr/src/route/rr_graph_generation/rr_graph_interposer.cpp
AmirhosseinPoolad: It's actually a bit more concerning. We could, for example, have an L16 wire that crosses an interposer cut line, gets to a 3D switch block on the other side and connects to a chan z node. The current version of the code does not remove that edge while ideally it should be removed.

I think this should be addressed in a follow up PR.

vaughnbetz (approved):
Thanks, LGTM. You should open an issue for the 3D case + interposer edge removal that you mention as problematic above.

PR #3343, opened 2025-11-14
[LibArchFPGA] Moved CAD Types out of Lib

AlexandreSinger:
The CAD types header file does not belong in libarchfpga since it was only really used in the prepacker of VPR. It makes several references to the pb_graph structure which I would ultimately like to move into VPR entirely.

This change only moves the declarations of types into VPR. No changes on functionality was made.


AmirhosseinPoolad (approved):
LGTM

PR #3344, opened 2025-11-15
[CI] Fix CPP Format Issue

AlexandreSinger:
The format test for CPP was done through a bash script which calls make
format under the hood. The issue was that if make format fails, it
thinks that the formatting is good, which is bad. I have fixed the
script so that it actually fails if make format fails.

The thing that was causing make format to fail was that the dependencies
that VPR needs to build were not installed. Added these dependencies.

### DIFF
--- a/libs/librtlnumber/src/include/internal_bits.hpp
+++ b/libs/librtlnumber/src/include/internal_bits.hpp
@@ -27,14 +27,14 @@ constexpr short integer_t_size = (sizeof(integer_t) * 8);
     }
 
 #define unroll_1d(lut) \
-    {lut[_0], lut[_1], lut[_x], lut[_z]}
+    { lut[_0], lut[_1], lut[_x], lut[_z] }
 #define unroll_2d(lut) \
-    {unroll_1d(lut[_0]), unroll_1d(lut[_1]), unroll_1d(lut[_x]), unroll_1d(lut[_z])}
+    { unroll_1d(lut[_0]), unroll_1d(lut[_1]), unroll_1d(lut[_x]), unroll_1d(lut[_z]) }
 
 #define unroll_1d_invert(lut) \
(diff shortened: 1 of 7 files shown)


AlexandreSinger:
Ok, I was close. Its not a Python issue (or maybe its not). Its a dependencies issue:
image

image

image

I did not install all of the dependencies since, why would you need dependencies if you are just doing formatting... But it looks like VTR cannot even be formatted without installing the dependencies. Thats kinda annoying but whatever.

odin_ii/regression_test/tools/ODIN_CONFIG.py
```diff
@@ -1542,7 +1542,7 @@ def buildChildren(self, child_, nodeName_):
 
 
 def usage():
-    print USAGE_TEXT
```
AmirhosseinPoolad, line 1545: Unrelated to this PR:

Woah! is that python 2? Do we even run this anywhere?

AlexandreSinger: Many of the scripts in VTR are not run by the CI. I do not have any hard data on this, but I think most Python scripts outside of vtr_task's python libs are not run at all.

Another issue is that Python is an interpreted language; so a Python2 syntax file can be executed by a Python3 interpreter and not crash unless it hits a magic if statement which prints something. I actually saw a file with this exact problem.

AlexandreSinger:
> Thanks for doing this Alex! LGTM.
> 
> P.S. We have so many Python 2 scripts! Is there a tool that can check a .py file and check if it's compatible with python 2 or 3?

@AmirhosseinPoolad I agree. I do not like it either, I am not aware of any tool; however, I think the Python linter was running the python code in some form since it drew my attention to this Python2 syntax issues (and had me fix them) without actually executing the code. However, this process has to be limited since Python is interpreted and the types of variables are not known until run time.

PR #3346, opened 2025-11-17
Setting IPIN Tdel

amin1377:
When reading a custom RR graph, I noticed that the routing results differed between reading an RR graph from file and generating the RR graph internally. After investigating, it became apparent that the discrepancy is caused by the router lookahead assigning different base costs for IPINs in the two cases.

When reading an RR graph, the Tdel assigned to an IPIN’s rr_indexed_data is set to the delay of the most frequent switch connected to that IPIN.
However, when the RR graph is generated, the Tdel is taken from the switch specified for IPINs in the architecture file.

This doesn't cause a problem for VPR's default RR Graph generator, but when building a custom RR graph (which I am implementing in a separate PR), the switches connected to IPIN nodes may not match those specified in the architecture file, leading to inconsistent Tdel values and therefore inconsistent routing results.

This PR updates the behavior so that reading and generating RR graphs use a consistent approach for assigning Tdel to IPINs.


amin1377:
> How does the custom RR graph generator decide which switch to use for different IPINs?

The switch to be used is determined by the switch pattern file provided to VTR, which specifies the connection patterns and the corresponding switch delays.

soheilshahrouz:
If different inputs to an IPIN have different delays, I think using the average is more accurate. I assume each edge with a distinct delay corresponds to a separate RR switch. Multiple of these switches may appear the same number of times, and using the mode causes one of them to be selected randomly.

amin1377:
For RR graphs built using VPR’s default RR graph builder, we don’t expect any QoR changes (since the same switch type is used to connect channels to IPINs). Just to be sure, I ran the Titan_other benchmarks and compared the results with the master branch, and they were the same: [Link](https://docs.google.com/spreadsheets/d/1EujuYvAnrOjjQPB8QrMDKcRzwwN44pB2ESqal0RIzCo/edit?usp=sharing)

PR #3347, opened 2025-11-17
Move node_tilable_track_nums_ to rr_graph_storage and add remove_node

amin1377:
**This PR depends on PR #3342. Please wait until that PR is merged.**

In this PR, node_tilable_track_nums_ is moved into the RR graph storage instead of residing in the RR graph builder, since this data structure (when using a tileable RR graph) needs to remain valid for the entire lifetime of the RR graph.

Additionally, a method is added to facilitate removing nodes from the RR graph.


libs/librrgraph/src/base/rr_graph_builder.cpp
```diff
@@ -309,8 +265,9 @@ void RRGraphBuilder::add_track_node_to_lookup(RRNodeId node) {
             // Routing channel nodes may have different ptc num 
             // Find the track ids using the x/y offset  
             if (e_rr_type::CHANX == node_type || e_rr_type::CHANY == node_type) {
-                ptc = (node_type == e_rr_type::CHANX) ? node_tilable_track_nums_[node][x - node_storage_.node_xlow(node)] : 
-                    node_tilable_track_nums_[node][y - node_storage_.node_ylow(node)];
+                const std::vector<short>& track_nums = node_storage_.node_tilable_track_nums(node);
+                ptc = (node_type == e_rr_type::CHANX) ? track_nums[x - node_storage_.node_xlow(node)] : 
+                      track_nums[y - node_storage_.node_ylow(node)];
```
AmirhosseinPoolad: I really don't like the way the ternary operator is used here. We have an assertion at the beginning of this function is CHANX or CHANY, so I think we could do something like:

```
ptc_num = undefined
if (chanx) ptc_num = something
else if (chany) ptc_num = something_else
add_node
```

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -178,7 +182,7 @@ bool t_rr_graph_storage::verify_first_edges() const {
 
 void t_rr_graph_storage::init_fan_in() {
     //Reset all fan-ins to zero
-    edges_read_ = true;
```
AmirhosseinPoolad, line 181: I agree that this method shouldn't set this flag, but I think I need a couple more eyes to look at this.

libs/librrgraph/src/base/rr_graph_storage.cpp
AmirhosseinPoolad: Can you change this code to use .pairs() instead? Something like:

> for (const auto& [edge_id, edge_destination_node] : edge_dest_node_.pairs())

Looks much nicer imo.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -699,6 +705,25 @@ class t_rr_graph_storage {
...
+            return node_tilable_track_nums_[node].size() > 1;
+        }
+    }
+
+    std::string node_ptc_nums_to_string(RRNodeId node) const;
+
+    const std::vector<short>& node_tilable_track_nums(RRNodeId node) const {
+        return node_tilable_track_nums_[node];
+    }
+
```
AmirhosseinPoolad: Doxygen for all of these

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -83,6 +83,10 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
         rr_graph_builder->add_rr_segment(segment_inf[(iseg)]);
     }
 
+    if (graph_type == e_graph_type::UNIDIR_TILEABLE) {
+        rr_graph_builder->set_tileable(true);
+    }
+
```
AmirhosseinPoolad, line 87: Does this fix a bug?

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -625,6 +629,54 @@ void t_rr_graph_storage::set_node_direction(RRNodeId id, Direction new_direction
     node_storage_[id].dir_side_.direction = new_direction;
 }
 
+void t_rr_graph_storage::set_node_ptc_nums(RRNodeId node, const std::string& ptc_str) {
+    VTR_ASSERT(size_t(node) < node_storage_.size());
+    std::vector<std::string> ptc_tokens = vtr::StringToken(ptc_str).split(",");
```
AmirhosseinPoolad: This method (and the corresponding method in RRGraphBuilder should get a nice, parsed vector of ints instead of parsing the string themselves. As of now this method is doing way too many things. Whoever calls this method should do the parsing themselves.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -625,6 +629,54 @@ void t_rr_graph_storage::set_node_direction(RRNodeId id, Direction new_direction
...
+    VTR_ASSERT(ptc_tokens.size() >= 1);
+    set_node_ptc_num(node, std::stoi(ptc_tokens[0]));
```
AmirhosseinPoolad: Add comment explaining why the first ptc_num is stored somewhere, then all ptc_nums are stored in another place. Why do we have this duplication?

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -625,6 +629,54 @@ void t_rr_graph_storage::set_node_direction(RRNodeId id, Direction new_direction
     node_storage_[id].dir_side_.direction = new_direction;
 }
 
+void t_rr_graph_storage::set_node_ptc_nums(RRNodeId node, const std::string& ptc_str) {
+    VTR_ASSERT(size_t(node) < node_storage_.size());
+    std::vector<std::string> ptc_tokens = vtr::StringToken(ptc_str).split(",");
+    VTR_ASSERT(ptc_tokens.size() >= 1);
```
AmirhosseinPoolad: VTR_ASSERT(!ptc_tokens.empty());

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -625,6 +629,54 @@ void t_rr_graph_storage::set_node_direction(RRNodeId id, Direction new_direction
...
+    VTR_ASSERT_MSG(node_type(node) == e_rr_type::CHANX || node_type(node) == e_rr_type::CHANY,
+                   "Track number valid only for CHANX/CHANY RR nodes");
+
+    size_t node_length = std::abs(node_xhigh(node) - node_xlow(node))
+                       + std::abs(node_yhigh(node) - node_ylow(node));
+    if (node_length + 1 != node_tilable_track_nums_[node].size()) {
+        node_tilable_track_nums_[node].resize(node_length + 1);
+    }
+
+    VTR_ASSERT(node_offset < node_tilable_track_nums_[node].size());
```
AmirhosseinPoolad: I think this would be clearer if it compared node_offset with node_length (can't have an offset bigger that's bigger than how long the node actually is) and it was before resizing node_tilable_track_nums_ (after line 653).

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -625,6 +629,54 @@ void t_rr_graph_storage::set_node_direction(RRNodeId id, Direction new_direction
...
+        return std::to_string(size_t(node_ptc_num(node)));
+    }
+    std::string ret;
+    for (size_t iptc = 0; iptc < node_tilable_track_nums_[node].size(); iptc++) {
+        ret += std::to_string(size_t(node_tilable_track_nums_[node][iptc])) + ",";
+    }
+    // Remove the last comma
+    ret.pop_back();
+    return ret;
+}
```
AmirhosseinPoolad: t_rr_graph_storage (and RRGraphBuilder) should not be doing string manipulations at all. I would move this to a free helper function in rr_graph_utils which takes a node ID and a reference to RRGraphView.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+        node_name_.erase(node);
+    }
+    if (is_tileable_) {
+        node_bend_start_.erase(node_bend_start_.end()-num_nodes_to_remove, node_bend_start_.end());
+        node_bend_end_.erase(node_bend_end_.end()-num_nodes_to_remove, node_bend_end_.end());
+        node_tilable_track_nums_.erase(node_tilable_track_nums_.end()-num_nodes_to_remove, node_tilable_track_nums_.end());
+    }
+
+    std::vector<RREdgeId> removed_edges;
+    auto adjust_edges = [&](vtr::vector<RREdgeId, RRNodeId>& edge_nodes) {
```
AmirhosseinPoolad: Please explicitly add the variables you want to the capture list.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+void t_rr_graph_storage::remove_nodes(const std::vector<RRNodeId>& nodes) {
+    VTR_ASSERT(!edges_read_);
+    // To remove the nodes, we first sort them in ascending order. This makes it easy 
+    // to calculate the offset by which other node IDs need to be adjusted. 
+    // For example, after sorting the nodes to be removed, if a node ID falls between 
+    // the first and second element, its ID should be reduced by 1. 
+    // If a node ID is larger than the last element, its ID should be reduced by 
+    // the total number of nodes being removed.
+    std::vector<RRNodeId> sorted_nodes = nodes;
+    std::sort(sorted_nodes.begin(), sorted_nodes.end());
```
AmirhosseinPoolad: How much do you care about not mutating the list of nodes for removal? This copy seems unnecessary to me but it's your call.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
     }
 }
 
+void t_rr_graph_storage::remove_nodes(const std::vector<RRNodeId>& nodes) {
```
AmirhosseinPoolad: Rename "nodes" to something like "rr_nodes_to_remove"

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+    VTR_ASSERT(!edges_read_);
+    // To remove the nodes, we first sort them in ascending order. This makes it easy 
+    // to calculate the offset by which other node IDs need to be adjusted. 
+    // For example, after sorting the nodes to be removed, if a node ID falls between 
+    // the first and second element, its ID should be reduced by 1. 
+    // If a node ID is larger than the last element, its ID should be reduced by 
+    // the total number of nodes being removed.
+    std::vector<RRNodeId> sorted_nodes = nodes;
```
AmirhosseinPoolad: This should also be renamed accordingly.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+            node_ptc_[new_node] = node_ptc_[old_node];
+            node_layer_[new_node] = node_layer_[old_node];
+            node_name_[new_node] = node_name_[old_node];
+            if (is_tileable_) {
+                node_bend_start_[new_node] = node_bend_start_[old_node];
+                node_bend_end_[new_node] = node_bend_end_[old_node];
+                node_tilable_track_nums_[new_node] = node_tilable_track_nums_[old_node];
+            }
+        }
+    }
```
AmirhosseinPoolad: This looks sound from what I can see be we should definitely have a CI test that uses this function.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+    // For example, after sorting the nodes to be removed, if a node ID falls between 
+    // the first and second element, its ID should be reduced by 1. 
+    // If a node ID is larger than the last element, its ID should be reduced by 
+    // the total number of nodes being removed.
+    std::vector<RRNodeId> sorted_nodes = nodes;
+    std::sort(sorted_nodes.begin(), sorted_nodes.end());
+    
+    // Iterate over the nodes to be removed and adjust the IDs of nodes 
+    // that fall between them. 
+    for (size_t i = 0; i < sorted_nodes.size(); ++i) {
```
AmirhosseinPoolad: i and j are very vague. Maybe something like "removal_index" and "node_index" would be clearer? Not sure.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+    // Now that the data structures are adjusted, we can shrink the size of them
+    size_t num_nodes_to_remove = sorted_nodes.size();
+    VTR_ASSERT(num_nodes_to_remove <= node_storage_.size());
+    node_storage_.erase(node_storage_.end()-num_nodes_to_remove, node_storage_.end());
+    node_ptc_.erase(node_ptc_.end()-num_nodes_to_remove, node_ptc_.end());
+    node_layer_.erase(node_layer_.end()-num_nodes_to_remove, node_layer_.end());
+    for (size_t node_index = node_name_.size()-num_nodes_to_remove; node_index < node_name_.size(); ++node_index) {
+        RRNodeId node = RRNodeId(node_index);
+        node_name_.erase(node);
+    }
```
AmirhosseinPoolad, line 741: Add comment explaining why this is correct (node_name is a hashmap not a vector and all the RRNodeIds past \#nodes - \#removal_nodes are invalid)

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
     }
 }
 
+void t_rr_graph_storage::remove_nodes(const std::vector<RRNodeId>& nodes) {
+    VTR_ASSERT(!edges_read_);
```
AmirhosseinPoolad, line 697: Add assertion for edges_partitioned_

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+    if (is_tileable_) {
+        node_bend_start_.erase(node_bend_start_.end()-num_nodes_to_remove, node_bend_start_.end());
+        node_bend_end_.erase(node_bend_end_.end()-num_nodes_to_remove, node_bend_end_.end());
+        node_tilable_track_nums_.erase(node_tilable_track_nums_.end()-num_nodes_to_remove, node_tilable_track_nums_.end());
+    }
+
+    std::vector<RREdgeId> removed_edges;
+    auto adjust_edges = [&](vtr::vector<RREdgeId, RRNodeId>& edge_nodes) {
+        for (size_t edge_index = 0; edge_index < edge_nodes.size(); ++edge_index) {
+            RREdgeId edge_id = RREdgeId(edge_index);
```
AmirhosseinPoolad: use all_edges() instead.

libs/librrgraph/src/base/rr_graph_storage.cpp
AmirhosseinPoolad: Add a comment explaining what this lambda does.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+            RRNodeId node = edge_nodes[edge_id];
+    
+            // Find insertion point in the sorted vector
+            auto node_it = std::lower_bound(sorted_nodes.begin(), sorted_nodes.end(), node);
+    
+            if (node_it != sorted_nodes.end() && *node_it == node) {
+                // Node exists in sorted_nodes, mark edge for removal
+                removed_edges.push_back(edge_id);
+            } else {
+                size_t node_offset = std::distance(sorted_nodes.begin(), node_it);
```
AmirhosseinPoolad: Not sure what's going on here. A comment would be nice.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -712,6 +737,17 @@ class t_rr_graph_storage {
...
+     * @brief Removes a given list of RRNodes from the RR Graph
+     * This method should be called before partition_edges has been called.
+     * If init_fan_in has been called, you need to call it again after removing the nodes.
+     * @note This a very expensive method, so should be called only when necessary. It is better
+     * to not add nodes in the first place, instead of relying on this method to remove nodes.
+     *
+     * @param nodes list of RRNodes to be removed
+     */
+    void remove_nodes(const std::vector<RRNodeId>& nodes);
+
```
AmirhosseinPoolad, line 761: From what I can understand, this function is O(|V| + |E|logk) where k is the number of nodes you want to remove. You should add that to the doxygen.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -651,6 +703,80 @@ void t_rr_graph_storage::set_virtual_clock_network_root_idx(RRNodeId virtual_clo
...
+                size_t new_node_index = size_t(node) - node_offset;
+                edge_nodes[edge_id] = RRNodeId(new_node_index);
+            }
+        }
+    };
+
+    adjust_edges(edge_src_node_);
+    adjust_edges(edge_dest_node_);
```
AmirhosseinPoolad, line 772: What about edge_switch_ and edge_remapped_?

libs/librrgraph/src/base/rr_graph_builder.cpp
amin1377: I applied the change you requested. But just out of curiosity, what’s the issue with using the ternary operator?

libs/librrgraph/src/io/rr_graph_reader.cpp
amin1377: Yes, I realize that when we read the RR graph, this variable isn’t set to true (unlike when the tileable RR graph is built).

libs/librrgraph/src/base/rr_graph_storage.cpp
amin1377: added the following comment:
```
// The default VTR RR graph generator assigns only one PTC number per node, which is
// stored in the node_ptc_ field of rr_graph_storage. However, when the tileable RR
// graph is used, CHANX/CHANY nodes can have multiple PTC numbers.
// 
// To satisfy VPR's requirements, we store the PTC number for offset = 0 in the
// node_ptc_ field, and store all PTC numbers assigned to the node in the
// node_tileable_track_nums_ field.
```

libs/librrgraph/src/base/rr_graph_storage.cpp
amin1377: I couldn’t care less about it 🙂 Just change the parameter to be passed by value.

libs/librrgraph/src/base/rr_graph_storage.cpp
amin1377: I agree. I will add the CI tests with CRR PR which is actaully using these API calls. For the tests that I curretnly have, I need to be able to build CRR.

libs/librrgraph/src/base/rr_graph_storage.cpp
amin1377: I’m iterating over the vector that’s passed to the lambda function… I think I’m not fully understanding what you mean here.

libs/librrgraph/src/base/rr_graph_storage.cpp
amin1377: I need to modify data structures that use RRNodeId (either as a key or value). The two data structures you mentioned only deal with edge IDs.

amin1377:
@AmirhosseinPoolad: I’ve addressed your comments. I’d appreciate it if you could take another look at the PR and let me know if you have any further feedback. It would be great if we could finalize this PR today!

libs/librrgraph/src/base/rr_graph_storage.cpp
amin1377: @soheilshahrouz: If you get a chance, I’d appreciate your thoughts on this. The reason I believe this needs to be removed is that edge_read should only be set when we are modifying data structures related to RR edges; for example, when updating edge_sink_node_ or the data structure that stores the first edge for each node.

In this case, we are only modifying the data structure that tracks the number of fan-ins for each node, and I should be able to call this without requiring edge_read to be set to true (since that would imply that edges must have been partitioned, etc.).

libs/librrgraph/src/base/rr_graph_builder.cpp
AmirhosseinPoolad: It's just very hard to read if the statements aren't super short and trivial imo.

libs/librrgraph/src/base/rr_graph_storage.cpp
AmirhosseinPoolad: Oh okay didn't see that, sorry. Why not iterate over .pairs() then?

> for (auto [edge_id, node] : edge_nodes.pairs())

This is a very minor nitpick tbh, I just don't like the cast. Feel free to ignore if you disagree.

AmirhosseinPoolad (commented):
Sorry for being late! This looks good other than the lambda capture list (super dangerous imo to capture  *everything* by non-const reference and it wouldn't be clear to future maintainers which outside variables the lambda is actually using without going through the lambda code). I still need another person to look at the flag change.

amin1377:
Thanks, @AmirhosseinPoolad . I’ve applied the changes you recommended. Please let me know if you have any further comments.

PR #3348, opened 2025-11-19
Removing 3D Attributes Script

amin1377:
Adding a bash script that takes an RR graph file as input and removes all attributes related to 3D architecture. I considered implementing this as a Python script, but based on my experience, even with streaming, Python doesn't handle very large RR graph files efficiently. For that reason, I decided to implement it as a bash script instead.

I tested this script by writing out the RR graph for one of the titan_other benchmarks, running the script on it, verifying that there were no occurrences of "layer", "layer_high", or "layer_low" in the modified file, and then running VTR again using the updated RR graph file. The results were identical in both runs.

### DIFF
--- a/vtr_flow/scripts/remove_3d_attributes.sh
+++ b/vtr_flow/scripts/remove_3d_attributes.sh
@@ -5,6 +5,11 @@
 # to remove all attributes related to 3D FPGA support (layer="0",
 # layer_low="0", layer_high="0"). In VPR, these values default to 0.
 
+# This also helps convert older RR graph files that used the "layer" attribute
+# instead of "layer_low" and "layer_high" for nodes, by removing the "layer" attribute
+# altogether. As a result, VPR can read the file without issues, and it slightly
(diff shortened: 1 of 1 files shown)


vtr_flow/scripts/remove_3d_attributes.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+# This script is intended to be used on RR graph XML files generated by VPR.
+# If the architecture is a 2D architecture, you can simply call this script
+# to remove all attributes related to 3D FPGA support (layer="0",
+# layer_low="0", layer_high="0"). In VPR, these values default to 0.
```
vaughnbetz, line 6: I’d add a comment saying why this is useful.
1. Convert an older rr graph with layer set to 2D format as we now require either no layer or both layer_low and layer_high.
2. 2. It also modestly reduces the file size for 2D archs.

PR #3353, opened 2025-11-20
[Spelling] Fixed Spelling in Documentation

AlexandreSinger:
Used codespell to fix all of the spelling in the documentation. This is basically all files which are not in the vpr or libs folder.


AlexandreSinger:
@vaughnbetz FYI. I am actually surprised how few spelling mistakes there were in the documentation. 174 lines is not bad. VPR and libs is much worse at around 600. That's coming in later.

dev/tutorial/ToNewVTRAdmin.txt
```diff
@@ -22,7 +22,7 @@ Whenever anyone commits code, our buildbot setup will automatically compile the
 code then run some regression tests.  Buildbot flags problems and reports who 
 made which commit.  The VTR buildbot website is located here: 
 http://canucks.eecg.toronto.edu:8080/ It gives you regular updates on the state 
-of thet trunk. As the administrator, you should check buildbot regularly (eg.
```
AmirhosseinPoolad, line 25: Should we remove this entire folder? The information is quite outdated.

dev/ammend-format.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-#Reformats and then ammends the currently checked-out commit
```
AmirhosseinPoolad, line 2: Should probably also rename the file itself. Or maybe delete it, I don't think it's used by anyone.

dev/rebase-format.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-#Reformat and ammend each commit since the provided 
+#Reformat and amend each commit since the provided
```
AmirhosseinPoolad, line 3: Same with this file, I think we might as well just remove it.

dev/tutorial/ToNewVTRAdmin.txt
vaughnbetz: We're not using buildbot anymore, so we should remove outdated info (maybe all is outdated here).

PR #3358, opened 2025-11-22
Remove ipin rr switch

amin1377:
Removing ipin_rr_switch_id

### DIFF
--- a/vpr/src/route/rr_graph_generation/rr_graph_area.cpp
+++ b/vpr/src/route/rr_graph_generation/rr_graph_area.cpp
@@ -290,9 +290,7 @@ static void count_bidir_routing_transistors(int num_switch, float R_minW_nmos, f
     /* Now add in the input connection block transistors. */
 
     // Get most frequent ipin switch
-    auto most_frequent_ipin_switch_it = std::ranges::max_element(
-        ipin_switch_count,
-        [](const auto& a, const auto& b) noexcept { return a.second < b.second; });
+    auto most_frequent_ipin_switch_it = std::ranges::max_element(ipin_switch_count, {}, [](const auto& p) noexcept { return p.second; });
     VTR_ASSERT(most_frequent_ipin_switch_it != ipin_switch_count.end());
     RRSwitchId most_frequent_ipin_switch = most_frequent_ipin_switch_it->first;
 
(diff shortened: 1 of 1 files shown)


libs/librrgraph/src/utils/alloc_and_load_rr_indexed_data.cpp
```diff
@@ -159,8 +157,7 @@ void alloc_and_load_rr_indexed_data(const RRGraphView& rr_graph,
     }
 
     load_rr_indexed_data_T_values(rr_graph,
-                 wire_to_ipin_switch,
-                 rr_indexed_data);
+                                  rr_indexed_data);
```
soheilshahrouz: I suggest calling this function in a single line. It's too short to break it into multiple lines.

vpr/src/route/rr_graph_generation/rr_graph_area.cpp
```diff
@@ -471,8 +490,15 @@ static void count_unidir_routing_transistors(std::vector<t_segment_inf>& /*segme
 
     /* Now add in the input connection block transistors. */
 
-    input_cblock_trans = get_cblock_trans(num_inputs_to_cblock, wire_to_ipin_switch,
-                                          max_inputs_to_cblock, trans_sram_bit);
+    // Get most frequent ipin switch
+    RRSwitchId most_frequent_ipin_switch = std::max_element(ipin_switch_count.begin(), ipin_switch_count.end(), [](const auto& a, const auto& b) {
```
soheilshahrouz: std::ranges::max_element with projection would be more readable

vpr/src/route/rr_graph_generation/rr_graph_area.cpp
```diff
@@ -285,8 +289,17 @@ static void count_bidir_routing_transistors(int num_switch, RRSwitchId wire_to_i
 
     /* Now add in the input connection block transistors. */
 
-    input_cblock_trans = get_cblock_trans(num_inputs_to_cblock, wire_to_ipin_switch,
-                                          max_inputs_to_cblock, trans_sram_bit);
+    // Get most frequent ipin switch
+    RRSwitchId most_frequent_ipin_switch = std::max_element(ipin_switch_count.begin(),
```
soheilshahrouz: ranges + projection

amin1377:
@soheilshahrouz: I applied the changes you requested. Regarding the test for variable IPIN delays, I already have one for CRR in the CRR PR. As for the default RR graph generator, since it hasn’t been tested before, implementing it will take some time. If you agree, I suggest we handle that in a separate PR.

vpr/src/route/rr_graph_generation/rr_graph_area.cpp
```diff
@@ -471,8 +490,17 @@ static void count_unidir_routing_transistors(std::vector<t_segment_inf>& /*segme
 
     /* Now add in the input connection block transistors. */
 
-    input_cblock_trans = get_cblock_trans(num_inputs_to_cblock, wire_to_ipin_switch,
-                                          max_inputs_to_cblock, trans_sram_bit);
+    // Get most frequent ipin switch
+    auto most_frequent_ipin_switch_it = std::ranges::max_element(
+        ipin_switch_count,
+        [](const auto& a, const auto& b) noexcept { return a.second < b.second; });
```
soheilshahrouz: I sill think the projection signitaure of max_element is more readable and fits into a single line.

`auto it = std::ranges::max_element(ipin_switch_count, {}, [](auto& p){ return p.second; });`

PR #3359, opened 2025-11-25
[Spelling] Fixed Simple Spelling Mistakes in Pack and Power

AlexandreSinger:
Fixed the simple spelling mistakes in most of VPR. The only VPR folder left is route which has quite a few since it is so documented. That needs to come in its own PR.

### DIFF
--- a/vpr/src/pack/cluster_router.cpp
+++ b/vpr/src/pack/cluster_router.cpp
@@ -246,7 +246,7 @@ static bool check_edge_for_route_conflicts(std::unordered_map<const t_pb_graph_n
                 pb_graph_node->illegal_modes.push_back(result.first->second->index);
             }
 
-            // If the number of illegal modes equals the number of available mode for a specific pb_graph_node it means that no cluster can be generated. This resuts
+            // If the number of illegal modes equals the number of available mode for a specific pb_graph_node it means that no cluster can be generated. This results
             // in a fatal error.
             if ((int)pb_graph_node->illegal_modes.size() >= pb_graph_node->pb_type->num_modes) {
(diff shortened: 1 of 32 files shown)


vpr/src/pack/cluster_router.cpp
```diff
@@ -984,10 +983,10 @@ static void fix_duplicate_equivalent_pins(t_lb_router_data* router_data, const A
                 VTR_ASSERT(lb_type_graph[pin_index].num_fanout[0] == 1);
                 int sink_index = lb_type_graph[pin_index].outedges[0][0].node_index;
                 VTR_ASSERT(lb_type_graph[sink_index].type == LB_SINK);
-                VTR_ASSERT_MSG(sink_index == lb_nets[ilb_net].terminals[iterm], "Remapped pin must be connected to original sink");
+                VTR_ASSERT_MSG(sink_index == lb_nets[ilb_net].terminals[term_idx], "Remapped pin must be connected to original sink");
 
                 //Change the target
-                lb_nets[ilb_net].terminals[iterm] = pin_index;
+                lb_nets[ilb_net].terminals[term_idx] = pin_index;
```
AmirhosseinPoolad, line 989: This entire function is actually terrifying the way it uses and reuses term_idx. Can you please put a note somewhere to refactor it in a later PR?

AlexandreSinger: I actually already have this on my radar. This will be part of future packer cleanups.

This reuse of the variable is not too bad, I tried to resolve it but its a bit more involved for this PR. Its just searching for a terminal index which matches some attribute.

PR #3366, opened 2025-12-01
Add Schema ID to RR Graph Formats

amin1377:
This PR adds a schema file ID to the RR graph attributes. The ID is included in both the XML and Cap’n Proto (binary) formats. When reading an RR graph, VTR now loads the schema ID stored in its internal schema and compares it with the ID found in the XML or binary file. If the IDs do not match, an error is raised.

@AlexandreSinger: I believe this also addresses your earlier concern about including an ID for the XML file.

Below is the first line of the RR graph XML, showing the added schema ID:

image

For the Cap’n Proto file: each time the schema is generated, Cap’n Proto assigns a random 64-bit number in the format `@0x...`. This number is stored as the schema ID in both the XML and binary RR graph files.

### DIFF
--- a/libs/librrgraph/src/io/rr_graph_reader.cpp
+++ b/libs/librrgraph/src/io/rr_graph_reader.cpp
@@ -73,11 +73,7 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
                   bool do_check_rr_graph,
                   bool echo_enabled,
                   const char* echo_file_name,
-#ifdef VTR_ENABLE_CAPNPROTO
                   const int route_verbosity,
-#else
-                  bool /*route_verbosity*/,
-#endif
                   bool is_flat) {
     vtr::ScopedStartFinishTimer timer("Loading routing resource graph");
 
@@ -100,6 +96,9 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
     ::capnp::Schema schema = ::capnp::Schema::from<ucap::RrGraph>();
     schema_file_id = schema.getProto().getScopeId();
     VTR_LOGV(route_verbosity > 1, "Schema file ID: 0x%016lx\n", schema_file_id);
+#else
+    // Suppress the warning about unused variable 'route_verbosity'
+    (void)route_verbosity;
 #endif
 
     RrGraphSerializer reader(
--- a/libs/librrgraph/src/io/rr_graph_writer.cpp
+++ b/libs/librrgraph/src/io/rr_graph_writer.cpp
@@ -38,11 +38,7 @@ void write_rr_graph(RRGraphBuilder* rr_graph_builder,
                     const char* file_name,
                     bool echo_enabled,
                     const char* echo_file_name,
-#ifdef VTR_ENABLE_CAPNPROTO
                     const int route_verbosity,
-#else
-                    bool /*route_verbosity*/,
-#endif
                     bool is_flat) {
 
     // If Cap'n Proto is enabled, a unique ID is assigned to the schema used to serialize the RR graph.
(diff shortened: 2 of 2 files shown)


libs/librrgraph/src/io/rr_graph.xsd
```diff
@@ -37,6 +37,7 @@
        attribute should be the name of a xs:simpleType that defines an
        xs:restriction / xs:enumeration.  This will be demonstrated later.
   -->
+
```
AlexandreSinger: Is this line-feed intentional?

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -85,6 +86,13 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
         rr_graph_builder->set_tileable(true);
     }
 
+    unsigned long schema_file_id = 0;
+#ifdef VTR_ENABLE_CAPNPROTO
+    ::capnp::Schema schema = ::capnp::Schema::from<ucap::RrGraph>();
+    schema_file_id = schema.getProto().getScopeId();
+    VTR_LOG("Schema file ID: 0x%016lx\n", schema_file_id);
```
AlexandreSinger: Should we make this a verbose message? I am not sure if this message will make sense for most users of VPR.

libs/librrgraph/src/io/rr_graph_writer.cpp
```diff
@@ -39,6 +40,13 @@ void write_rr_graph(RRGraphBuilder* rr_graph_builder,
                     const char* echo_file_name,
                     bool is_flat) {
 
+    unsigned long schema_file_id = 0;
+#ifdef VTR_ENABLE_CAPNPROTO
+    ::capnp::Schema schema = ::capnp::Schema::from<ucap::RrGraph>();
+    schema_file_id = schema.getProto().getScopeId();
+    VTR_LOG("Schema file ID: 0x%016lx\n", schema_file_id);
```
AlexandreSinger: Same comment here. I am wondering if we should make this a verbose log message.

libs/librrgraph/src/io/rr_graph_reader.cpp
amin1377: I don't have a strong opinion on it, but I think if you are reading or writing the RR Graph, this is relevant information to print even when verbose mode is not enabled.

vaughnbetz: I think it’s ok to print it only when there’s a mismatch (already done above) or verbose is set.

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
```diff
@@ -1727,6 +1729,17 @@ class RrGraphSerializer final : public uxsd::RrGraphBase<RrGraphContextTypes> {
      *     <xs:attribute name="tool_comment" type="xs:string" />
      *   </xs:complexType>
      */
+    inline void set_rr_graph_schema_file_id(unsigned long schema_file_id, void*& /*ctx*/) final {
+        // Only check if the schema file ID is not 0. If it is 0, it means capnproto is not enabled.
+        // Thus, we cannot check the schema file ID mismatch.
```
vaughnbetz: I think this comment moment is incomplete. Don’t you also allow older rr graphs with no schema to come in? Might also want the to explain that we’re checking the file is compatible when can, but alllowing it to be read if we can’t determine its schema

libs/librrgraph/src/io/rr_graph_reader.cpp
```diff
@@ -85,6 +86,13 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
         rr_graph_builder->set_tileable(true);
     }
 
+    unsigned long schema_file_id = 0;
```
vaughnbetz, line 94: Should comment the purpose of the schema file id

vaughnbetz (changes requested):
A few minor changes

libs/librrgraph/src/io/rr_graph_uxsdcxx_serializer.h
amin1377: Changed the comment to the following:

```
// Only check if schema_file_id_ (set when initializing the class) is not 0.
// If it is 0, it means Cap'n Proto is not enabled, so we cannot check for a schema file ID mismatch.
// This function is only called when the RR graph file being read contains a schema file ID.
// If it does not, this function is not called, and the RR graph can be read without performing
// the schema file ID check.
```

AlexandreSinger (approved):
Thanks @amin1377 LGTM!

I am happy that the RR graph writer now has access to the route verbosity as well.

PR #3370, opened 2025-12-12
[Spelling] Fixed Spelling Mistakes in VTR Flow

AlexandreSinger:
Fixed the spelling mistakes in the VTR flow. We exclude the pearl libraries and benchtracker since these projects are not being maintained at this time.

We also ignore the benchmarks and tasks directories since these are mainly used for testing and contain a lot of false positives. We do not want to make it challenging to add more tests.

This should be the last major spelling fix-up, unless there is a directory that is currently not being checked that we actually want to be checked.

### DIFF
--- a/vtr_flow/primitives.lib
+++ b/vtr_flow/primitives.lib
@@ -84,7 +84,7 @@ library (VTRPrimitives) {
      *
      *  INPUTS:
      *      datain
-     *  OUPUTS:
+     *  OUTPUTS:
      *      dataout
      */
     cell (fpga_interconnect) {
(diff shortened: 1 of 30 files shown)


vtr_flow/primitives.lib
```diff
@@ -84,7 +84,7 @@ library (VTRPrimitives) {
      *
      *  INPUTS:
      *      datain
-     *  OUPUTS:
+     *  OUTPUTS:
```
AmirhosseinPoolad, line 87: I'm not familiar with this .lib file so I'm not sure if it's safe to change this. Is the functionality here tested to make sure it's okay?

AlexandreSinger: This is a Liberty file. They are a very strange "standard" and they are not often written by hand this way. They allow C++ style block-comments. This is within a block-comment. So this will not make a functional change.

PR #3382, opened 2025-12-28
Update 7series

WhiteNinjaZ:
#### Description

This PR provides an updated version of the switch blocks within the seven series. While running several comparisons between the RR graph generated by VTR and the actual 7-series we found that although the custom switchblocks within our old architecture captured the general trends of the 7-series, several interconnects between wires had incorrect counts. This PR fixes those issues and provides a more accurate capture. 

The following is a heatmap of the generated SB connections in VTR:
image

This connectivity pattern is very close to the actual 7-series switchblock as published in the Netcracker paper (included here for reference):
image

When looking at the 7 largest benchmarks, both CPD and Min W where either as good as or better than the previous architecture (LU8 and Stereo0 had a 3% improvement on minW and LU32 had a 7% improvement):
image

#### Types of changes

#### Checklist:


vaughnbetz:
Hi @WhiteNinjaZ : can you add a before vs. after QoR ratio comparison?  I think the table above is just from the VTR 9 paper, and isn't showing the change due to this PR (please correct me if I'm wrong).

vaughnbetz:
QoR looks good -- nice improvements!

PR #3390, opened 2026-01-10
RR GSB Reordering Incoming Edges

amin1377:
When generating bitstreams in OpenFPGA, depending on the target device, it is sometimes necessary to reorder incoming edges so that all edges originating from OPINs are placed first. This PR updates the incoming-edge creation logic to enforce this ordering.

### DIFF
--- a/vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
+++ b/vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
@@ -901,8 +901,8 @@ void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
         for (size_t side = 0; side < get_num_sides(); ++side) {
             /* Edges from grid outputs are the 1st part */
             for (size_t opin_id = 0; opin_id < opin_node_[side].size(); ++opin_id) {
-                if ((0 < from_grid_edge_map.contains(side))
-                    && (0 < from_grid_edge_map.at(side).contains(opin_id))) {
+                if ((from_grid_edge_map.contains(side))
+                    && (from_grid_edge_map.at(side).contains(opin_id))) {
                     chan_node_in_edges_[size_t(chan_side)][track_id].push_back(from_grid_edge_map[side][opin_id]);
                 }
             }
@@ -910,8 +910,8 @@ void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
         for (size_t side = 0; side < get_num_sides(); ++side) {
             /* Edges from routing tracks are the 2nd part */
             for (size_t itrack = 0; itrack < chan_node_[side].get_chan_width(); ++itrack) {
-                if ((0 < from_track_edge_map.contains(side))
-                    && (0 < from_track_edge_map.at(side).contains(itrack))) {
+                if ((from_track_edge_map.contains(side))
+                    && (from_track_edge_map.at(side).contains(itrack))) {
                     chan_node_in_edges_[size_t(chan_side)][track_id].push_back(from_track_edge_map[side][itrack]);
                 }
             }
@@ -922,16 +922,16 @@ void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
         for (size_t side = 0; side < get_num_sides(); ++side) {
             /* Edges from grid outputs are the 1st part */
             for (size_t opin_id = 0; opin_id < opin_node_[side].size(); ++opin_id) {
-                if ((0 < from_grid_edge_map.contains(side))
-                    && (0 < from_grid_edge_map.at(side).contains(opin_id))) {
+                if ((from_grid_edge_map.contains(side))
+                    && (from_grid_edge_map.at(side).contains(opin_id))) {
                     chan_node_in_edges_[size_t(chan_side)][track_id].push_back(from_grid_edge_map[side][opin_id]);
                 }
             }
 
             /* Edges from routing tracks are the 2nd part */
             for (size_t itrack = 0; itrack < chan_node_[side].get_chan_width(); ++itrack) {
-                if ((0 < from_track_edge_map.contains(side))
-                    && (0 < from_track_edge_map.at(side).contains(itrack))) {
+                if ((from_track_edge_map.contains(side))
+                    && (from_track_edge_map.at(side).contains(itrack))) {
                     chan_node_in_edges_[size_t(chan_side)][track_id].push_back(from_track_edge_map[side][itrack]);
                 }
             }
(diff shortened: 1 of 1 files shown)


vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
```diff
@@ -831,7 +831,8 @@ void RRGSB::add_mux_node(const RRNodeId& mux_node) {
 
 void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
                                     const e_side& chan_side,
-                                    const size_t& track_id) {
+                                    const size_t& track_id,
```
AlexandreSinger: Nit: size_t is a simple type and should not be passed by reference.

I know this is not your code, but while youre here lol.

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
```diff
@@ -831,7 +831,8 @@ void RRGSB::add_mux_node(const RRNodeId& mux_node) {
...
+                                    const bool reorder_incoming_edges) {
```
AlexandreSinger, line 835: Why make this a const bool?

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.h
```diff
@@ -221,7 +221,8 @@ class RRGSB {
     /* Sort all the incoming edges for one channel rr_node */
     void sort_chan_node_in_edges(const RRGraphView& rr_graph,
                                  const e_side& chan_side,
-                                 const size_t& track_id);
+                                 const size_t& track_id,
+                                 const bool reorder_incoming_edges);
```
AlexandreSinger, line 232: I would add something in the docstring for what reorder_incoming_edges is doing and why it is needed.

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
```diff
@@ -891,29 +892,51 @@ void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
...
-    for (size_t side = 0; side < get_num_sides(); ++side) {
-        /* Edges from grid outputs are the 1st part */
-        for (size_t opin_id = 0; opin_id < opin_node_[side].size(); ++opin_id) {
-            if ((0 < from_grid_edge_map.count(side))
-                && (0 < from_grid_edge_map.at(side).count(opin_id))) {
-                chan_node_in_edges_[size_t(chan_side)][track_id].push_back(from_grid_edge_map[side][opin_id]);
+    // Sorting incoming edges. If reorder_incoming_edges is true,
+    // the edges from OPINs will be put first. This is required for bitstream generation.
+    if (reorder_incoming_edges) {
+        for (size_t side = 0; side < get_num_sides(); ++side) {
```
AlexandreSinger, line 901: This took me a lot of time to realize what the difference between the true and false case is. It looks like when we are reordering the incoming edges, we split the for loop to ensure that the edges come first. That makes sense, but I recommend adding some comments here to make that more obvious. From first glance, this is very tricky to see.

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
```diff
@@ -891,29 +892,51 @@ void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
...
+            /* Edges from grid outputs are the 1st part */
+            for (size_t opin_id = 0; opin_id < opin_node_[side].size(); ++opin_id) {
+                if ((0 < from_grid_edge_map.count(side))
+                    && (0 < from_grid_edge_map.at(side).count(opin_id))) {
```
AlexandreSinger: These if statements are a bit confusing. Are you OK with upgrading them to C++20's ".contains" method instead?
https://en.cppreference.com/w/cpp/container/unordered_map/contains.html

I think it will just make this code easier to follow.

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
amin1377: It’s for safety. This is set by the user on the OpenFPGA side, so it doesn’t make sense to allow it to be changed in this function.

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
```diff
@@ -891,29 +892,56 @@ void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
...
+    // Sorting incoming edges. If reorder_incoming_edges is true,
+    // the edges from OPINs will be put first. This is required for bitstream generation.
+    if (reorder_incoming_edges) {
+        // If reorder_incoming_edges is true, the edges from OPINs will be put first.
+        // So, in the first loop, we process the edges from OPINs. Then, in the second loop,
+        // we process the edges from routing tracks.
+        for (size_t side = 0; side < get_num_sides(); ++side) {
+            /* Edges from grid outputs are the 1st part */
+            for (size_t opin_id = 0; opin_id < opin_node_[side].size(); ++opin_id) {
+                if ((0 < from_grid_edge_map.contains(side))
```
AmirhosseinPoolad: You don't need the comparison with 0 anymore

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
```diff
@@ -891,29 +892,56 @@ void RRGSB::sort_chan_node_in_edges(const RRGraphView& rr_graph,
...
+                    && (0 < from_grid_edge_map.at(side).contains(opin_id))) {
+                    chan_node_in_edges_[size_t(chan_side)][track_id].push_back(from_grid_edge_map[side][opin_id]);
+                }
+            }
+        }
+        for (size_t side = 0; side < get_num_sides(); ++side) {
+            /* Edges from routing tracks are the 2nd part */
+            for (size_t itrack = 0; itrack < chan_node_[side].get_chan_width(); ++itrack) {
+                if ((0 < from_track_edge_map.contains(side))
+                    && (0 < from_track_edge_map.at(side).contains(itrack))) {
```
AmirhosseinPoolad: Same here

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
```diff
@@ -1038,15 +1075,17 @@ void RRGSB::sort_ipin_node_in_edges(const RRGraphView& rr_graph,
...
+            if (0 < from_track_edge_map[iside].contains(itrack)) {
+                ipin_node_in_edges_[size_t(ipin_side)][ipin_id].push_back(from_track_edge_map[iside][itrack]);
+            }
         }
     }
 
     for (e_side iside : {TOP, RIGHT, BOTTOM, LEFT}) {
         for (size_t ipin = 0; ipin < get_num_opin_nodes(iside); ++ipin) {
-            if (0 < from_opin_edge_map[size_t(iside)].count(ipin)) {
+            if (0 < from_opin_edge_map[size_t(iside)].contains(ipin)) {
```
AmirhosseinPoolad: ditto

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_gsb.cpp
amin1377: Good catch! Thanks!

PR #3392, opened 2026-01-14
Skip RR graph for packer-only

amin1377:
When the packer is the only stage being run, there is no need to build the RR graph. This PR implements that optimization.

### DIFF
--- a/vpr/src/base/vpr_api.cpp
+++ b/vpr/src/base/vpr_api.cpp
@@ -1316,11 +1316,7 @@ static void free_routing() {
 static void free_noc() {}
 
 static bool is_pack_only(const t_vpr_setup& vpr_setup) {
-    return vpr_setup.PackerOpts.doPacking != e_stage_action::SKIP &&
-        vpr_setup.PlacerOpts.doPlacement == e_stage_action::SKIP &&
-        vpr_setup.APOpts.doAP == e_stage_action::SKIP &&
-        vpr_setup.RouterOpts.doRouting == e_stage_action::SKIP &&
-        vpr_setup.AnalysisOpts.doAnalysis == e_stage_action::SKIP;
+    return vpr_setup.PackerOpts.doPacking != e_stage_action::SKIP && vpr_setup.PlacerOpts.doPlacement == e_stage_action::SKIP && vpr_setup.APOpts.doAP == e_stage_action::SKIP && vpr_setup.RouterOpts.doRouting == e_stage_action::SKIP && vpr_setup.AnalysisOpts.doAnalysis == e_stage_action::SKIP;
 }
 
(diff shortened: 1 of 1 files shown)


vpr/src/base/vpr_api.cpp
```diff
@@ -1310,6 +1315,10 @@ static void free_routing() {
  */
 static void free_noc() {}
 
+static bool is_pack_only(const t_vpr_setup& vpr_setup) {
+    return vpr_setup.PackerOpts.doPacking != e_stage_action::SKIP && vpr_setup.PlacerOpts.doPlacement == e_stage_action::SKIP && vpr_setup.APOpts.doAP == e_stage_action::SKIP && vpr_setup.RouterOpts.doRouting == e_stage_action::SKIP && vpr_setup.AnalysisOpts.doAnalysis == e_stage_action::SKIP;
```
AmirhosseinPoolad, line 1319: Add a line break here, too long.

vpr/src/base/vpr_api.cpp
```diff
@@ -533,15 +538,15 @@ bool vpr_flow(t_vpr_setup& vpr_setup, t_arch& arch) {
...
+void vpr_create_device(t_vpr_setup& vpr_setup, const t_arch& arch, const bool pack_only) {
     vtr::ScopedStartFinishTimer timer("Create Device");
     vpr_create_device_grid(vpr_setup, arch);
 
     vpr_setup_clock_networks(vpr_setup, arch);
 
     vpr_setup_noc(vpr_setup, arch);
 
-    if (vpr_setup.PlacerOpts.place_chan_width != NO_FIXED_CHANNEL_WIDTH) {
+    if (vpr_setup.PlacerOpts.place_chan_width != NO_FIXED_CHANNEL_WIDTH && !pack_only) {
```
AmirhosseinPoolad, line 549: You might as well have vpr_create_device call is_pack_only instead of passing it as an argument.

AmirhosseinPoolad (commented):
Looks good, had some minor thoughts.

vpr/src/base/vpr_api.cpp
amin1377: I tried that, but running make format changed the line to this format.

vpr/src/base/vpr_api.cpp
amin1377: I thought about that, but I decided to pass it as an argument because this function is called in different places. Depending on how the parameters are set, calling this function could result in pack_only being set to true unintentionally. Passing it as a parameter makes the behavior explicit and ensures that the logic determining whether the packer is the only enabled stage is invoked exactly where I expect it to be.

PR #3393, opened 2026-01-14
[CI] Regenerate some NightlyTestManual golden results

AmirhosseinPoolad:
Regenerates golden results for some small test circuits. Please double check to see if I put the correct golden results in the correct place.


AlexandreSinger (approved):
LGTM thanks Amir.

PR #3394, opened 2026-01-15
Generate RR graph when requested in pack-only flow

amin1377:
If writing the RR graph is requested, it must still be generated even when only the packing stage is being run. This PR updates PR #3392 to handle this case correctly.


vpr/src/base/vpr_api.cpp
```diff
@@ -1316,7 +1316,12 @@ static void free_routing() {
 static void free_noc() {}
 
 static bool is_pack_only(const t_vpr_setup& vpr_setup) {
-    return vpr_setup.PackerOpts.doPacking != e_stage_action::SKIP && vpr_setup.PlacerOpts.doPlacement == e_stage_action::SKIP && vpr_setup.APOpts.doAP == e_stage_action::SKIP && vpr_setup.RouterOpts.doRouting == e_stage_action::SKIP && vpr_setup.AnalysisOpts.doAnalysis == e_stage_action::SKIP;
+    return vpr_setup.RoutingArch.write_rr_graph_filename.empty()
```
AlexandreSinger: I do not like this change since it breaks the semantics of the function. This function is supposed to return true if only the pack stage is being run. If someone uses this function in another context, it will not work as expected.

Instead of adding this check in this function, why not just add the check in the line that actually loads the RR graph?
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/582e2f72f6e8541fd786f2dbfe765f252f0dbb8f/vpr/src/base/vpr_api.cpp#L541-L554
This is the only place that uses pack_only in all of VPR, so you can just add a check if the RR graph was requested here.

amin1377: That's actually a good suggestion! Thanks!

vaughnbetz (approved):
Looks good to me; thanks.

PR #3399, opened 2026-01-21
[Pack] Added New Floorplanning Constraint for Constraining Partitions to Logical Block Types

AlexandreSinger:
The packer currently has a bug where if you have two logical blocks that are in different tiles which can implement the same primitive, and you try to constrain that primitive to a region that only contains one of the logical block types, it fails to select the correct logical block type during pack time (and fails to find a solution, even though the solution is trivial).

The issue is caused by the packer not taking into account the constraints on the logical block types in the partitions that atoms are constrained to.

To resolve this issue, I have extended the floorplanning placement constraints to include a new tag "add_logical_block". This new tag adds constraints onto the partition. Any atom in the partition will only be able to implement a logical block that the partition is constrained to. Note: You can add multiple of these "add_logical_block" tags to constrain to multiple types. An example of this new constraint is shown below:
```xml
<vpr_constraints tool_name="vpr">
<partition_list>
    <partition name="clbA_part">
        <add_atom name_pattern="\[.*\]"/>
        <add_region x_low="1" y_low="1" x_high="22" y_high="44"/>
        <add_logical_block name_pattern="clbA"/>
    </partition>
    <partition name="clbB_part">
        <add_atom name_pattern="n_n[0-9]+"/>
        <add_region x_low="23" y_low="1" x_high="44" y_high="44"/>
        <add_logical_block name_pattern="clbB"/>
    </partition>
</partition_list>
</vpr_constraints>
```
This example is added to the strong tests. It is on an architecture with two types of CLB (clbA and clbB), where one type is on the left half of the chip and the other is on the right half of the chip:
Screenshot from 2026-01-21 13-48-35
Without these logical block constraints, the packer cannot find a solution.

Another use-case for this is for RAMs. Suppose you have an architecture with two types of RAMs (M9K and M144K, for example), this interface would allow the user to constrain the RAM primitives to a specific RAM type (if they know better than the packer for example).

## Release Notes

* **New Features**
  * Added support for constraining partitions to specific logical block types in placement constraints, enabling finer-grained control over atom-to-cluster assignments during packing.

* **Documentation**
  * Updated placement constraints documentation to describe logical block type constraints, including matching by name and regex pattern.

* **Tests**
  * Added regression test validating logical block type constraints with multi-variant architectures.

### DIFF
--- a/vpr/src/base/user_place_constraints.h
+++ b/vpr/src/base/user_place_constraints.h
@@ -105,7 +105,7 @@ class UserPlaceConstraints {
      * @brief Constrain the atoms in the given partition to be within blocks of a specific type
      *
      *   @param part_id   The id of the partition being constrained
-     *   @param lb_type   The logical block type the partition is constrained to
+     *   @param lb_type   The logical block type (i.e. cluster type) the partition is constrained to
      */
     void constrain_part_lb_type(PartitionId part_id, t_logical_block_type_ptr lb_type);
 
@@ -147,7 +147,9 @@ class UserPlaceConstraints {
 
     /**
      * This is an empty set that is returned as a reference by the get_part_lb_type_constraints
-     * function. This is to allow that function to return all results by reference.
+     * function if a partition does not have a constraint on the logical block types into which
+     * its primitives can be packed into. This is to allow that function to return all results
+     * by reference.
      */
     std::unordered_set<t_logical_block_type_ptr> EMPTY_SET_;
 };
--- a/vpr/src/pack/verify_clustering.cpp
+++ b/vpr/src/pack/verify_clustering.cpp
@@ -386,7 +386,7 @@ static unsigned check_clustering_floorplanning_consistency(
                 }
             }
 
-            // Check that each atom in the cluster can implement this logical block type.
+            // Check that each atom in the cluster can be implemented in this logical block type.
             for (AtomBlockId atom_blk_id : atoms_in_clb) {
                 PartitionId atom_part_id = constraints.get_atom_partition(atom_blk_id);
                 if (atom_part_id.is_valid() && constraints.is_part_constrained_to_lb_types(atom_part_id)) {


vpr/src/base/user_place_constraints.cpp
```diff
@@ -61,6 +64,33 @@ PartitionRegion& UserPlaceConstraints::get_mutable_partition_pr(PartitionId part
...
+    auto it = constrained_part_lb_types_.find(part_id);
+    if (it == constrained_part_lb_types_.end()) {
+        // If this partition is not already constrained, create a new entry.
+        constrained_part_lb_types_.insert({part_id, {lb_type}});
+    } else {
+        // If this partition is already constrained to a block type, add it
+        // to the set.
+        it->second.insert(lb_type);
+    }
+}
```
AmirhosseinPoolad, line 80: constrained_part_lb_types_ is an unordered_map, so I think you can do constrained_part_lb_types_[part_id].insert(lb_type) and it will just work.

vpr/src/base/user_place_constraints.cpp
```diff
@@ -61,6 +64,33 @@ PartitionRegion& UserPlaceConstraints::get_mutable_partition_pr(PartitionId part
...
+    return constrained_part_lb_types_.contains(part_id);
+}
+
+const std::unordered_set<t_logical_block_type_ptr>& UserPlaceConstraints::get_part_lb_type_constraints(PartitionId part_id) const {
+    auto it = constrained_part_lb_types_.find(part_id);
+
+    VTR_ASSERT_SAFE(it != constrained_part_lb_types_.end());
+
+    return it->second;
+}
```
AmirhosseinPoolad, line 94: Not a huge fan of this API, shouldn't be *this* easy to crash the program. You have a function whose name implies "hey what LB types is this partition constrained to?" and a very reasonable assumption is that if it's not constrained to any LB types the function will tell you that it's not constrained to any LB types. This however crashes which is super unexpected!

You can't do an optional since you're returning a reference and returning whatever constrained_part_lb_types_[part_id] is (which could be empty) isn't possible since the function is constant. I don't have an easy solution for this tbh. I think the safest option here is to have an empty set as a const private field in this class and if part_id does not exist in constrained_part_lb_types_ this method returns a reference to that empty set instead.

Anyway, you definitely don't want VTR_ASSERT_SAFE here. This check should be done at all times imo, so VTR_ASSERT or even VTR_ASSERT_OPT would be better choices.

vpr/src/pack/verify_clustering.cpp
```diff
@@ -385,6 +385,27 @@ static unsigned check_clustering_floorplanning_consistency(
...
+                if (atom_part_id.is_valid() && constraints.is_part_constrained_to_lb_types(atom_part_id)) {
+                    // If the atom partition exists and is constrained to some lb types,
+                    // check that the cluster's logical block type is one of the types
+                    // that this atom can implement.
+                    const auto& atom_lb_type_constraints = constraints.get_part_lb_type_constraints(atom_part_id);
+                    t_logical_block_type_ptr clb_block_type = clb_nlist.block_type(clb_blk_id);
+                    if (!atom_lb_type_constraints.contains(clb_block_type)) {
+                        VTR_LOG_ERROR(
+                            "Cluster block %zu contains an atom %zu which is constrained "
+                            "to be unable to implement the logical block type of the cluster: %s\n",
```
AmirhosseinPoolad: I don't understand this sentence.

vtr_flow/arch/reg_test_archs/fix_clusters_multiple_clb.xml
```diff
@@ -0,0 +1,275 @@
+<!--
```
AmirhosseinPoolad, line 1: Add a comment to this file that explains why this architecture exists and the general idea of it

vpr/src/base/user_place_constraints.cpp
AmirhosseinPoolad: unordered_map::operator[] not being const and default constructing new keys has it's benefits!

doc/src/vpr/placement_constraints.rst
```diff
@@ -121,3 +122,17 @@ Partitions, Atoms, and Regions
...
 		there are not enough resources in a partition. Overlapping partitions cause some physical blocks to be counted in more
 		than one partition, which will degrade the packing algorithm's ability to create a clustering that can be placed given the floorplan constraints.
+
+	.. arch:tag:: <add_logical_block name_pattern="string">
+
+		An ``<add_logical_block>`` tag is used to constrain the atoms in the partition to specific logical block types. This tag is optional and can be repeated multiple times within a partition to specify multiple allowed logical block types.
+		When logical block type constraints are specified, atoms in the partition will only be packed into clusters of the specified logical block types.
+		The ``<add_logical_block>`` tag has the following attribute:
+
+		:req_param name_pattern:
```
amin1377, line 135: I understand that this would give users more flexibility, but using regex to specify block types feels a bit weird. It might be better to let users specify the exact logical type they want, which would help prevent unintentional mistakes.

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -55,6 +56,8 @@
 struct partition_info {
     Partition part;
     std::vector<AtomBlockId> atoms;
+    // NOTE: This needs to be a vector for writing to make sense since an order needs to exist.
```
amin1377: I guess I didn’t quite understand why the order matters here. Isn’t this only used to check what logic blocks are available? If so, wouldn’t an unordered_map be a better choice?

vpr/src/base/user_place_constraints.cpp
AlexandreSinger: Discussed in person. In summary this syntax is used throughout this file. I also think its better to be more explicit here since some people reading this code may not be as familiar with the map containers insert-if-missing semantics.

vpr/src/base/user_place_constraints.cpp
AlexandreSinger: I ended up doing the returning of an empty set. Since this is part of the UserPlaceConstraints class, we can make a private static member which can be referenced. I added a warning to the function declaration to make it clear that unconstrained partitions will return an empty set.

doc/src/vpr/placement_constraints.rst
AlexandreSinger: Discussed in person. Although you are correct, I still think that having this flexibility is better than not having it.

vpr/src/base/user_place_constraints.h
```diff
@@ -98,7 +101,40 @@ class UserPlaceConstraints {
      */
     PartitionRegion& get_mutable_partition_pr(PartitionId part_id);
 
+    /**
+     * @brief Constrain the atoms in the given partition to be within blocks of a specific type
+     *
+     *   @param part_id   The id of the partition being constrained
+     *   @param lb_type   The logical block type the partition is constrained to
```
vaughnbetz: I wonder if we should mention ('cluster type') here ... (to differentiate from primitives) ?

vpr/src/base/user_place_constraints.h
```diff
@@ -108,6 +144,12 @@ class UserPlaceConstraints {
      * Store all partitions
      */
     vtr::vector<PartitionId, Partition> partitions;
+
+    /**
+     * This is an empty set that is returned as a reference by the get_part_lb_type_constraints
+     * function. This is to allow that function to return all results by reference.
```
vaughnbetz: ... function 'if a partition does not have a constraint on the logical block types into which its primitives can be packed into.'

vpr/src/pack/verify_clustering.cpp
```diff
@@ -385,6 +385,28 @@ static unsigned check_clustering_floorplanning_consistency(
                     break;
                 }
             }
+
+            // Check that each atom in the cluster can implement this logical block type.
```
vaughnbetz: 'can be implemented in this logical block type'

PR #3405, opened 2026-01-26
[Scripts] Fixed Timeout Issue

AlexandreSinger:
The timeout argument for run_vtr_flow was not working correctly. If a test ran longer than the timeout, the Python script will not terminate the process. The run will continue until it is finished and then get the proper timeout error. This defeats the whole purpose of having a timeout.

Tracked the problem down to a utility script which was trying to kill the process and was unable to do so.

### DIFF
--- a/vtr_flow/scripts/python_libs/vtr/util.py
+++ b/vtr_flow/scripts/python_libs/vtr/util.py
@@ -214,7 +214,13 @@ class CommandRunner:
                     # Abort if over time limit
                     elapsed_time = time.time() - start_time
                     if self._timeout_sec and elapsed_time > self._timeout_sec:
+                        # Send SIGTERM command to all processes in the process
+                        # group. We need to do this since Popen may create a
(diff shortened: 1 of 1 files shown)


vtr_flow/scripts/python_libs/vtr/util.py
```diff
@@ -212,7 +214,12 @@ def run_system_command(
                     # Abort if over time limit
                     elapsed_time = time.time() - start_time
                     if self._timeout_sec and elapsed_time > self._timeout_sec:
-                        proc.terminate()
+                        os.killpg(proc.pid, signal.SIGTERM)
+                        try:
+                            proc.wait(timeout=5)
+                        except subprocess.TimeoutExpired:
+                            os.killpg(proc.pid, signal.SIGKILL)
+                        break
```
AmirhosseinPoolad, line 228: Add comments explaining what this is, then LGTM.

Issue #3406, opened 2026-01-27
[LibArchFPGA] Unknown attributes are not being warned / errored


AlexandreSinger:
See #3403 

There are three examples shown in this PR:

1. "num_pins" attribute in mode
2. "num_pins" attribute in interconnect tags
3. Missing "Cout" attribute in switch tag.

This attributes are all ignored / given default values; however, these should throw an error to prevent people writing syntax they do not intend.

PR #3407, opened 2026-01-27
[LibArchFPGA] Added Attribute Checks to Modes and Interconnects

AlexandreSinger:
LibArchFPGA was not producing an error when extra attributes were added to the mode and interconnect tags. These attributes were just getting thrown away, however we should error to ensure the user was not expecting them to do anything.

See https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3406


AlexandreSinger:
Weekly tests passed. Ready for review.

PR #3409, opened 2026-01-29
Reduce CRR build time

amin1377:
This PR adds several fixes to reduce the time required to build the CRR.

### DIFF
--- a/libs/libarchfpga/src/read_xml_arch_file.cpp
+++ b/libs/libarchfpga/src/read_xml_arch_file.cpp
@@ -1852,6 +1852,8 @@ static void process_interconnect(vtr::string_internment& strings,
         pugi::xml_node cur = get_first_child(Parent, child_name, loc_data, ReqOpt::OPTIONAL);
 
         while (cur != nullptr) {
+            expect_only_attributes(cur, {"name", "input", "output"}, loc_data);
+
             if (0 == strcmp(cur.name(), "complete")) {
                 mode->interconnect[interconnect_idx].type = COMPLETE_INTERC;
             } else if (0 == strcmp(cur.name(), "direct")) {
@@ -1932,6 +1934,7 @@ static void process_mode(pugi::xml_node Parent,
     if (implied_mode) {
         mode->name = vtr::strdup("default");
     } else {
+        expect_only_attributes(Parent, {"name", "disable_packing"}, loc_data);
         const char* prop = get_attribute(Parent, "name", loc_data).value();
         mode->name = vtr::strdup(prop);
     }


vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_pattern_matcher.h
```diff
@@ -125,9 +126,14 @@ class CRRPatternMatcher {
             return name == pattern;
         }
 
-        // Compile regex for this pattern
-        std::string regex_str = pattern_to_regex(pattern);
-        std::regex compiled_regex(regex_str);
+        // Cache compiled regex objects to avoid recompilation on every call.
+        static std::unordered_map<std::string, std::regex> regex_cache;
```
AlexandreSinger: I wouldn't use a static variable here since the memory will never be returned to the system. Since this is a mapping between strings and regexes, the data will be stored here for the rest of the program.

Instead, why not make this method non-static and store this cache as part of the CRRPatternMatcher class? That way the memory will be freed when this object is deleted.

AlexandreSinger: This cache is an interesting idea, but is creating the regexes really that big of a bottleneck? If it gave you great performance I am ok with it; but if the gains are small I would recommend removing this.

libs/librrgraph/src/base/rr_graph_storage.cpp
```diff
@@ -10,6 +10,7 @@
 #include <algorithm>
 #include <cstddef>
 #include <ranges>
+#include <unordered_map>
```
AlexandreSinger: Is unordered map used in this file? I do not see any added maps.

libs/libvtrutil/src/vtr_util.h
AlexandreSinger, line 1: Just double checking, is this uniquify method used anywhere else? If it is, we should ensure that it does not change functionality beyond what we would expect. This is a strange issue.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -287,25 +289,30 @@ CRRConnectionBuilder::SegmentInfo CRRConnectionBuilder::parse_segment_info(const
...
-    auto it = node_lookup.find(hash);
-    if (it != node_lookup.end()) {
-        return it->second;
+    auto it = col_nodes.find(hash);
+    if (it == col_nodes.end()) {
+        it = row_nodes.find(hash);
+        if (it == row_nodes.end()) {
+            return RRNodeId::INVALID();
+        }
     }
```
AlexandreSinger, line 308: Just for cleanliness to make this more readable, you may want to consider writing it like this:
```cpp
auto col_it = col_nodes.find(hash);
if (col_it != col_nodes.end())
  return it->second;

auto row_it = row_nodes.find(hash);
if (row_it != row_nodes.end())
  return it->second;

return RRNodeId::INVALID();
```

The way it is currently written is a bit confusing.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -349,15 +356,18 @@ RRNodeId CRRConnectionBuilder::process_channel_node(const SegmentInfo& info,
...
+    if (it == col_nodes.end()) {
+        it = row_nodes.find(hash);
+        if (it == row_nodes.end()) {
+            VTR_LOGV(verbosity_ > 1, "Node not found: %s [%s] (%d,%d) -> (%d,%d)\n", seg_type_label.c_str(),
+                     seg_sequence.c_str(), x_low, y_low, x_high, y_high);
+            return RRNodeId::INVALID();
+        }
     }
+
+    return it->second;
```
AlexandreSinger: Same exact comment as above. The way this is written is a bit confusing.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_edge_builder.cpp
```diff
@@ -52,6 +54,14 @@ static t_arch_switch_inf create_crr_switch(const int delay_ps, const std::string
 static RRSwitchId find_or_create_crr_switch_id(const int delay_ps,
                                                const std::string& sw_template_id,
                                                const int verbosity) {
+    // Cache template_id -> switch_id for O(1) lookup instead of
+    // iterating over the entire all_sw_inf map (O(n)) on every call.
+    static std::unordered_map<std::string, int> template_id_cache;
```
AlexandreSinger: Be very careful with static variables like this. These variables are never freed! You are basically never getting this memory back so at some point you need to think about clearing them!

I strongly recommend creating a cache variable and storing it somewhere. That way the memory can be freed when you are done parsing the CRR!

libs/libvtrutil/src/vtr_util.h
amin1377: It was used in the CRR generator and in the remove_edge method that Amir developed for the interposer to remove edges.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_edge_builder.cpp
amin1377: I agree. When I initially made it a static variable, it was just a quick hack to see how much performance gain I’d get. I’ve since changed it so it gets freed once the caller function is done using it.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_pattern_matcher.h
AlexandreSinger: As we discussed in-person. This static variable is not going to be good long term since this unordered map will exist forever. If this is not giving much performance, it should just be removed.

AmirhosseinPoolad: Please refactor this to also work like template_id_cache.

AmirhosseinPoolad (commented):
Had the same comment as Alex, looks good otherwise.

AlexandreSinger (approved):
LGTM

PR #3414, opened 2026-02-01
[AP][PartialLegalizer] Basic 3D Support

AlexandreSinger:
Added basic 3D support to the partial legalizer by updating data
structures.

The basic algorithm is to always partition along the layers first, and
then partition along the 2D layer second.

We now should support non-timing driven AP on 3D architectures.

## Summary by CodeRabbit

* **New Features**
  - Multi-layer support added across placement binning, density management and legalization.
  - Wirelength estimation extended to include vertical (z) dimension for 3D placements.

* **Improvements**
  - Placement legalization and spreading/partitioning enhanced to operate across multiple layers.
  - Placement export now preserves per-bin layer information.

* **Tests**
  - Updated 3D analytical placement test configuration (WL-driven scenario).

### DIFF
--- a/vpr/src/analytical_place/partial_legalizer.cpp
+++ b/vpr/src/analytical_place/partial_legalizer.cpp
@@ -1194,7 +1194,7 @@ std::vector<SpreadingWindow> BiPartitioningPartialLegalizer::get_min_windows_aro
             region = vtr::bounding_box(region,
                                        density_manager_->flat_placement_bins().bin_region(overfilled_bin_cluster[i]));
             new_window.layer_low = std::min(new_window.layer_low, density_manager_->flat_placement_bins().bin_layer(overfilled_bin_cluster[i]));
-            new_window.layer_high = std::min(new_window.layer_high, density_manager_->flat_placement_bins().bin_layer(overfilled_bin_cluster[i]));
+            new_window.layer_high = std::max(new_window.layer_high, density_manager_->flat_placement_bins().bin_layer(overfilled_bin_cluster[i]));
         }
 
         // Grow the region until it is just large enough to not overfill
@@ -1550,7 +1550,8 @@ PartitionedWindow BiPartitioningPartialLegalizer::partition_window(
     if (window.layer_low != window.layer_high) {
         // Set the pivot to be the middle of the layer.
         // TODO: We should do a proper search here to find a good partition line.
-        double pivot = (window.layer_high - window.layer_low) / 2.0;
+        size_t pivot_layer = window.layer_low + (window.layer_high - window.layer_low) / 2;
+        double pivot = static_cast<double>(pivot_layer) + 0.5;
 
         partitioned_window.partition_dir = e_partition_dir::PLANAR;
         partitioned_window.pivot_pos = pivot;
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
@coderabbitai review

AlexandreSinger:
Results on the largest 8 VTR benchmarks (on a 2D architecture to ensure that the 2D results do not change):

  | baseline_parse_results.txt | parse_results.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.975925
odin_synth_time |   |  
parmys_synth_time |   |  
abc_depth |   |  
abc_synth_time |   |  
num_clb | 1 | 1
num_memories | 1 | 1
num_mult | 1 | 1
max_vpr_mem | 1 | 1.000711
num_pre_packed_blocks |   |  
device_grid_tiles | 1 | 1
pack_time |   |  
placed_wirelength_est | 1 | 1
place_time | 1 | 0.973308
placed_CPD_est | 1 | 1
routed_wirelength | 1 | 1
critical_path_delay | 1 | 1
crit_path_route_time | 1 | 1.001435


The overall flow runtime did not change a noticeale amount and he QoR is identical.

vpr/src/analytical_place/partial_legalizer.cpp
```diff
@@ -689,51 +703,56 @@ void FlowBasedLegalizer::legalize(PartialPlacement& p_placement) {
...
-                                              const vtr::Rect<double>& region) const {
+                                              const vtr::Rect<double>& region,
+                                              size_t layer) const {
     VTR_ASSERT_SAFE(dim.is_valid());
     // Get the sum over the given region.
-    uint64_t sum = dim_prefix_sum_[dim].get_sum(region.xmin(),
-                                                region.ymin(),
-                                                region.xmax() - 1,
-                                                region.ymax() - 1);
+    uint64_t sum = layer_dim_prefix_sum_[layer][dim].get_sum(region.xmin(),
```
amin1377, line 739: Was there a reason you explicitly used uint64_t instead of size_t? I know size_t is platform-dependent, but on 64-bit machines it should be 64 bits.

vpr/src/analytical_place/partial_placement.cpp
```diff
@@ -21,20 +21,25 @@ double PartialPlacement::get_hpwl(const APNetlist& netlist) const {
...
         }
         // TODO: In the placer, the x and y dimensions are multiplied by cost
         //       factors based on the channel width. Should somehow bring these
         //       in here.
         //       Vaughn thinks these may make sense in the objective HPWL, but
         //       not the in the estimated post-placement wirelength.
-        VTR_ASSERT_SAFE(max_x >= min_x && max_y >= min_y);
-        hpwl += max_x - min_x + max_y - min_y;
+        VTR_ASSERT_SAFE(max_x >= min_x && max_y >= min_y && max_z >= min_z);
+        // TODO: Should dz be added here? Or should we add a penalty?
```
amin1377, line 41: I personally think that for wirelength cost, no additional cost should be added when blocks are located on different layers, since timing already accounts for that.

vpr/src/analytical_place/partial_legalizer.cpp
AlexandreSinger: The prefix sum is stored in fixed point. I need to keep an eye on the number of bits and the sign of the value stored. I forced it to uint64_t to lock it to an unsigned 64-bit number. If we used size_t, the number of bits may change depending on the machine which we do not want here.

vpr/src/analytical_place/partial_placement.cpp
AlexandreSinger: That makes sense! I am going to keep the TODO for me to do a deeper exploration into it later.

PR #3415, opened 2026-02-02
[SDC] Updated to new TCL-based SDC parser

AlexandreSinger:
Updated the code on the VTR side to support the new SDC parser.

All of the changes are isolated into the file that reads the SDC file (i.e. declares the callbacks used by LibSDCParse). The prior functionality of VTR should not have changed; however, now that we are using a TCL-based parser, the syntax coverage should be much better.

## Summary by CodeRabbit

* **New Features**
  * Improved SDC parsing: per-target clock creation and explicit clock-source resolution for ports/pins/nets.
* **Bug Fixes**
  * Stricter validation and clearer errors/warnings for timing commands (multicycle, false path, uncertainties, latencies); unsupported constructs now reported.
* **Documentation**
  * Reformatted SDC sample files for consistent selector spacing and semicolon formatting.
* **Chores**
  * Updated SDC parsing submodule reference and added Tcl/SWIG packages to installers.


AlexandreSinger:
Once this is closer to merging, we will need to run the weekly tests on this branch.

AlexandreSinger:
@coderabbitai review

AlexandreSinger:
Results on titan_quick_qor (the exact script from nightly test 2):

  | baseline.txt | parse_results.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.997192
num_LAB | 1 | 1
num_DSP | 1 | 1
num_M9K | 1 | 1
num_M144K | 1 | 1
max_vpr_mem | 1 | 1.027392
num_pre_packed_blocks | 1 | 1
device_grid_tiles | 1 | 1
pack_time | 1 | 0.98536
placed_wirelength_est | 1 | 1
place_time | 1 | 0.986697
placed_CPD_est | 1 | 1
routed_wirelength | 1 | 1
critical_path_delay | 1 | 1
crit_path_route_time | 1 | 0.999193

The important thing to notice here is that the quality of results is identical, meaning that all of the SDC commands should be interpreted in the same way as before.

I did a spot check on a few of the circuits and found that all of the old warnings are the same and there is only one new warning:
```
Warning XXX: set_clock_groups: '-exclusive' argument is deprecated. Use '-asynchronous' instead.
```
Which is expected.

These results are hiding something somewhat important. The cost of using a TCL interpreter here is performance. The time it takes to parse the SDC files has gone up:
- on the smallest circuit (stereovision), the time to load timing constraints went from 0.01 seconds to 3.86 seconds
- on one of the largest circuits (directrf), the time to load timing constraints went from 0.09 seconds to  54.20 seconds

This is not terrible, we still parse in less than a minute for a circuit that takes 2.5 hours to run. The issue comes from how we are doing the glob search (we are entirely doing it in TCL). We can probably get the runtime back by optimizing that code heavily.

@vaughnbetz Let me know what you think.

vpr/src/timing/read_sdc.cpp
```diff
@@ -1,8 +1,11 @@
 #include "read_sdc.h"
 
+#include <filesystem>
 #include <limits>
 #include <regex>
+#include <unordered_map>
 
+#include "tatum/TimingGraphFwd.hpp"
 #include "vtr_log.h"
```
vaughnbetz, line 29: Make sure there are big picture comments somewhere (and point at them) or add them here.  Why are you adding all these objects to a database?  Where should I go to see the overall flow commented?

vpr/src/timing/read_sdc.cpp
```diff
@@ -81,84 +134,120 @@ class SdcParseCallback : public sdcparse::Callback {
...
+                VTR_ASSERT(cmd.targets.strings.size() == 1);
+                clock_name = cmd.name;
+            }
+
+            // Get the clock source associated with this pin.
+            tatum::NodeId clock_source = get_clock_source(clock_pin);
+            VTR_ASSERT(clock_source.is_valid());
+
+            // Create netlist clock
+            create_clock_object(clock_name, clock_source, cmd);
```
vaughnbetz, line 251: Make sure netlist_clock is a defined term, or define it in a comment.  I think it means the clock nets (drives a clock pin on some block) in the design netlist.

vpr/src/timing/read_sdc.cpp
```diff
@@ -1052,6 +1265,11 @@ class SdcParseCallback : public sdcparse::Callback {
     std::set<AtomPinId> netlist_clock_drivers_;
     std::map<std::string, AtomPinId> netlist_primary_ios_;
 
+    std::unordered_map<sdcparse::PortObjectId, AtomPinId> object_to_port_id_;
```
vaughnbetz, line 1290: Comment the mappings -- names are good but I really like to have the data made explicit.

vaughnbetz (approved):
A few commenting suggestions to consider, but LGTM.

PR #3417, opened 2026-02-03
[AP][Solver] Basic 3D Timing Support

AlexandreSinger:
Upgraded the solver to allow for basic 3D support when timing is enabled. With this, the full AP flow should now work with 3D architectures.

## Summary by CodeRabbit

* **New Features**
  * 3D multi-die FPGA placement support extended across delay derivative and normalization flows for multi-layer placements.

* **Bug Fixes / Reliability**
  * Improved boundary checks and safeguards for multi-layer delay calculations; prevents invalid values and divisions by zero.

* **Tests**
  * Added a regression test configuration covering the full 3D placement flow with default arguments.

### DIFF
--- a/vpr/src/analytical_place/analytical_solver.cpp
+++ b/vpr/src/analytical_place/analytical_solver.cpp
@@ -1021,7 +1021,6 @@ std::tuple<double, double, double> B2BSolver::get_delay_derivative(APBlockId dri
     // difference calculations below.
     VTR_ASSERT_SAFE(tile_dx + 1 < (int)device_grid_width_);
     VTR_ASSERT_SAFE(tile_dy + 1 < (int)device_grid_height_);
-    VTR_ASSERT_SAFE(tile_dlayer + 1 < (int)device_grid_num_layers_);
 
     // Get the delay of a wire going from the given driver block location to the
     // given sink block location. This should only use the physical tile type of


vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -12,6 +12,7 @@
 #include <limits>
 #include <memory>
 #include <thread>
+#include <tuple>
```
amin1377, line 15: Just out of curiosity, what’s the best practice here? For libraries that are already included in the header file (analytical_solver.h), is it common to also include them in the .cpp file?

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -1005,7 +1003,7 @@ std::pair<double, double> B2BSolver::get_delay_derivative(APBlockId driver_blk,
     // can be implemented, which it currently does).
     t_physical_tile_loc driver_block_loc(block_x_locs_legalized[driver_blk],
                                          block_y_locs_legalized[driver_blk],
-                                         layer_num);
+                                         is_multi_die() ? block_z_locs_legalized[driver_blk] : 0);
```
amin1377, line 1006: For a single-layer architecture, isn’t block_z_locs_legalized already set to zero for all blocks?

vpr/src/analytical_place/analytical_solver.cpp
```diff
@@ -1014,10 +1012,11 @@ std::pair<double, double> B2BSolver::get_delay_derivative(APBlockId driver_blk,
     // in x/y is not larger than the width/height of the device.
     t_physical_tile_loc sink_block_loc(driver_block_loc.x + flat_dx,
                                        driver_block_loc.y + flat_dy,
-                                       layer_num);
+                                       is_multi_die() ? driver_block_loc.layer_num + flat_dlayer : 0);
```
amin1377, line 1015: Same as above; wouldn’t flat_dlayer already be zero for a single-layer architecture?

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: I usually include it in both. This allows you to remove the libraries from the header files in the future without having to modify the .cpp files. This follows include-what-you-use principals, which just makes the code more stable to changes.

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: That is correct, but I wanted to make this more stable. Specifically in this case, block_z_locs_legalized may not be initialized at all.

vpr/src/analytical_place/analytical_solver.cpp
AlexandreSinger: Same comment here. This just makes it safer to prevent weird errors from creeping in due to assumptions.

PR #3418, opened 2026-02-03
Add atom model name to net echo file

amin1377:
Update atom net echo file to include the atom model name

### DIFF
--- a/vpr/src/base/atom_netlist_utils.cpp
+++ b/vpr/src/base/atom_netlist_utils.cpp
@@ -314,8 +314,8 @@ void print_netlist_as_blif(FILE* f, const AtomNetlist& netlist, const LogicalMod
             ports.push_back(port_id);
         }
 
-        fprintf(f, "# Subckt Model Name: %s\n", model.name);
-        fprintf(f, ".subckt %s \\\n", netlist.block_name(blk_id).c_str());
+        fprintf(f, "# Subckt %zu: %s\n", size_t(blk_id), netlist.block_name(blk_id).c_str());
+        fprintf(f, ".subckt %s \\\n", model.name);
         for (size_t i = 0; i < ports.size(); i++) {
             auto width = netlist.port_width(ports[i]);
             for (size_t j = 0; j < width; ++j) {


vpr/src/base/atom_netlist_utils.cpp
```diff
@@ -306,14 +306,15 @@ void print_netlist_as_blif(FILE* f, const AtomNetlist& netlist, const LogicalMod
...
+        const t_model& model = models.get_model(blk_model);
 
         std::vector<AtomPortId> ports;
         for (auto port_id : netlist.block_ports(blk_id)) {
             VTR_ASSERT(netlist.port_width(port_id) > 0);
             ports.push_back(port_id);
         }
 
-        fprintf(f, ".subckt %s \\\n", netlist.block_name(blk_id).c_str());
+        fprintf(f, ".subckt %s %s \\\n", model.name, netlist.block_name(blk_id).c_str());
```
AlexandreSinger: I am confused, the function is named `print_netlist_as_blif`; is this still proper blif syntax?
image
image
from: https://course.ece.cmu.edu/~ee760/760docs/blif.pdf

I think we should be careful with making this change. To be honest, I am not even sure if block_name is even correct here...

amin1377: I understand your concern, and I agree that if we are generating BLIF files intended for consumption, we should definitely adhere to the BLIF standard. However, `print_netlist_as_blif` is currently used only to generate **echo files** (see [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/base/read_circuit.cpp#L71) and [here](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/base/read_circuit.cpp#L85)).

Given that this is an echo file, I’m not sure that strict adherence to the BLIF standard is required here. The intent of this change was simply to make the output more descriptive and easier to understand.

Regarding the block name: I don't know what you mean by it being “correct.” Before this change, we were printing only the instance name, without any indication of the model type. In this PR, the model name is printed first, followed by the instance name, which provides more information than before.

amin1377:
@AlexandreSinger: As discussed, I took another look at the BLIF echo file, and it doesn’t seem to violate any BLIF standards. Because of that, I added the model name as a comment so the BLIF standard isn’t violated. Please let me know what you think.

AlexandreSinger (approved):
I like this solution a lot better! Thank you, LGTM.

PR #3420, opened 2026-02-05
Add interposer legality checking to placement

AmirhosseinPoolad:
Previously placer would place placement macros across interposer cutlines. This would have resulted in unroutable placements. This PR adds legality checking for that case, making the placer respect interposer cutlines for placement macros.

### DIFF
--- a/libs/libarchfpga/src/grid_util.h
+++ b/libs/libarchfpga/src/grid_util.h
@@ -1,6 +1,8 @@
+#include <algorithm>
 #include <concepts>
 #include <vector>
 
+#include "vtr_assert.h"
 #include "vtr_log.h"
 #include "vtr_ndmatrix.h"
 
@@ -23,8 +25,8 @@
  * 
  * @param device_width Width of the device
  * @param device_height Height of the device
- * @param horizontal_lines Horizontal lines defining blocks in the device
- * @param vertical_lines Vertical lines defining blocks in the device
+ * @param horizontal_lines Sorted list of horizontal lines defining blocks in the device
+ * @param vertical_lines Sorted list of vertical lines defining blocks in the device
  * @param reduced_matrix Reduced matrix with values for each block
  */
 template<typename T>
@@ -34,6 +36,14 @@ vtr::NdMatrix<T, 2> get_device_sized_matrix_from_reduced(size_t device_width,
                                                          const std::vector<int>& vertical_lines,
                                                          const vtr::NdMatrix<T, 2>& reduced_matrix) {
     vtr::NdMatrix<T, 2> device_matrix({device_width, device_height}, T());
+
+    // Reduced matrix size and the lines defining blocks in the device should match
+    VTR_ASSERT(horizontal_lines.size() == reduced_matrix.dim_size(1) - 1);
+    VTR_ASSERT(vertical_lines.size() == reduced_matrix.dim_size(0) - 1);
+
+    // Horizontal and vertical lines should be sorted
+    VTR_ASSERT(std::ranges::is_sorted(horizontal_lines));
+    VTR_ASSERT(std::ranges::is_sorted(vertical_lines));
 
     for (size_t i = 0; i < vertical_lines.size() + 1; i++) {
         for (size_t j = 0; j < horizontal_lines.size() + 1; j++) {


libs/libarchfpga/src/grid_util.h
```diff
@@ -0,0 +1,51 @@
+#include <concepts>
+#include <vector>
+
+#include "vtr_log.h"
+#include "vtr_ndmatrix.h"
+
+template<typename T>
+vtr::NdMatrix<T, 2> get_device_sized_matrix_from_reduced(size_t x_size,
```
amin1377: Doxygen comment for the function

libs/libarchfpga/src/device_grid.h
```diff
@@ -18,6 +18,11 @@ struct t_grid_tile {
     const t_metadata_dict* meta = nullptr;
 };
 
+// A unique identifier for a die on device
+// Used in multi-die 2.5D and 3D architectures
```
AlexandreSinger, line 22: Just to confirm, for a 3D die with 2 layers, would each layer be considered different dice?

libs/libarchfpga/src/grid_util.h
```diff
@@ -0,0 +1,51 @@
...
+                                                         size_t y_size,
```
AlexandreSinger: I would use `device_width` and `device_height` to be consistent with other functions. This function is named "device_sized_matrix" so it would make it a bit more clear what x and y mean here. Also doxygen comments for this function would be helpful.

libs/libarchfpga/src/grid_util.h
```diff
@@ -0,0 +1,51 @@
...
+                                                         const std::vector<int>& horizontal_lines,
+                                                         const std::vector<int>& vertical_lines,
+                                                         const vtr::NdMatrix<T, 2>& reduced_matrix) {
+    vtr::NdMatrix<T, 2> device_matrix({x_size, y_size});
+    device_matrix.fill(T());
```
AlexandreSinger: Does the NdMatrix class not have a constructor that already does this? Is this fill necessary?

vtr_flow/arch/interposer/titan_interposer/stratixiv_arch.timing_interposer.xml
AlexandreSinger, line 1: You should add a comment at the top of this file with at least the date the Stratix-IV was copied into this file. It may be useful if the base Stratix-IV architecture capture changes. You should also add a comment for what makes this architecture different from the base architecture.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_interposer/titan_interposer/config/config.txt
```diff
@@ -0,0 +1,32 @@
...
+# Parse info and how to parse
+parse_file=vpr_titan.txt
+
+# How to parse QoR info
+qor_parse_file=qor_vpr_titan.txt
+
+# Pass requirements
+pass_requirements_file=pass_requirements_vpr_titan.txt
+
+script_params=-starting_stage vpr -track_memory_usage --route_chan_width 300 --max_router_iterations 400
```
AlexandreSinger, line 32: Is setting the max_router_iterations to 400 necessary here? I am just curious because I also put this into the Titan architecture tests.

amin1377 (approved):
Thanks, Amir!

libs/libarchfpga/src/device_grid.h
AmirhosseinPoolad: Yes. Added a comment making this more clear.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_interposer/titan_interposer/config/config.txt
AmirhosseinPoolad: No idea to be honest. I copied it from the nightly tests.

vpr/src/place/move_utils.cpp
```diff
@@ -508,6 +517,29 @@ bool is_legal_swap_to_location(ClusterBlockId blk,
...
+    int macro_id = place_macros.get_imacro_from_iblk(blk);
+    if (macro_id != UNDEFINED && device_ctx.grid.has_interposer_cuts()) {
+        const t_pl_macro& macro = place_macros.macros()[macro_id];
+
+        auto blk_in_macro = std::ranges::find(macro.members, blk,
+                                              [](const t_pl_macro_member& pl_mem) noexcept { return pl_mem.blk_index; });
+        VTR_ASSERT_SAFE(blk_in_macro != macro.members.end());
+        t_pl_loc head_to_pos = to - blk_in_macro->offset;
+
+        for (const t_pl_macro_member& macro_member : macro.members) {
```
AmirhosseinPoolad, line 531: Add comment: this works for any macro shape, but if CPU is affected could check only start and end.

AmirhosseinPoolad (commented):
Additional comments from Vaughn:

- Run QoR for 2D architectures.

PR #3425, opened 2026-02-11
[Pack][Router] Cleaned up the cluster router API

AlexandreSinger:
The cluster router is very old code in VPR (around 13 years old) and the current API makes it a bit hard to use.

The original API passed around a pointer to the router state, which the cluster router functions used to perform / update the route. I changed this API into a class since it would inherently maintain its own state while making it easier to track the state.

This update will make it easier to profile and work on the cluster router in the future.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated intra-cluster routing into a single encapsulated routing engine with clearer lifecycle management, improving packing reliability, stability and memory management.
  * Replaced procedural routing state with an object-oriented router interface and adjusted cluster lifecycle to use it.
  * Reduced public API surface and improved const-correctness for internal graph queries, simplifying usage and reducing possible side effects.

### DIFF
--- a/vpr/src/pack/cluster_router.h
+++ b/vpr/src/pack/cluster_router.h
@@ -127,8 +127,7 @@ class compare_expansion_node {
      *
      * TL;DR lowest cost is popped first.
      */
-    bool operator()(t_expansion_node& e1, t_expansion_node& e2)
-    {
+    bool operator()(t_expansion_node& e1, t_expansion_node& e2) {
         if (e1.cost > e2.cost) {
             return true;
         }


AlexandreSinger:
@coderabbitai review

vpr/src/pack/cluster_router.cpp
```diff
@@ -35,78 +37,24 @@
...
-
-static void expand_node(t_lb_router_data* router_data, t_expansion_node exp_node, reservable_pq<t_expansion_node, std::vector<t_expansion_node>, compare_expansion_node>& pq, int net_fanout);
-static void expand_node_all_modes(t_lb_router_data* router_data, t_expansion_node exp_node, reservable_pq<t_expansion_node, std::vector<t_expansion_node>, compare_expansion_node>& pq, int net_fanout);
-
-static bool add_to_rt(t_lb_trace* rt, int node_index, t_lb_router_data* router_data, int irt_net);
-static bool is_route_success(t_lb_router_data* router_data);
-static t_lb_trace* find_node_in_rt(t_lb_trace* rt, int rt_index);
-static void reset_explored_node_tb(t_lb_router_data* router_data);
-static void save_and_reset_lb_route(t_lb_router_data* router_data);
+static void reset_lb_net_rt(t_lb_trace& lb_trace);
```
vaughnbetz, line 55: Not sure if you want to comment these or they'll be refactored out of existence soon anyway ....

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
 
-/* Constructors/Destructors */
-t_lb_router_data* alloc_and_load_router_data(std::vector<t_lb_type_rr_node>* lb_type_graph, t_logical_block_type_ptr type);
-void free_router_data(t_lb_router_data* router_data);
-void free_intra_lb_nets(std::vector<t_intra_lb_net>* intra_lb_nets);
+/**************************************************************************
+ * Intra-Logic Block Router Data Structures
+ ***************************************************************************/
+
+/* Describes the status of a logic cluster_ctx.blocks routing resource node for a given logic cluster_ctx.blocks instance */
```
vaughnbetz: Suggest converting comments to doxygen unless this is slated for further  clean up / removal soon.

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+    bool is_mode_conflict = false;
+    bool try_expand_all_modes = false;
+    bool expand_all_modes = false;
+
+    inline bool is_mode_issue() const {
+        return is_mode_conflict || try_expand_all_modes;
+    }
+};
+
+// TODO: check if this hacky class memory reserve thing is still necessary, if not, then delete
```
vaughnbetz, line 179: Agreed; not sure we really need this given that a vector grows efficiently (depends how often we create this queue I guess).

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+        this->c.reserve(cur_cap);
+    }
+
+  private:
+    size_type cur_cap;
+};
+
+typedef reservable_pq<t_expansion_node, std::vector<t_expansion_node>, compare_expansion_node> ClusterRouterPriorityQueue;
+
+enum e_commit_remove { RT_COMMIT,
```
vaughnbetz, line 211: Comment

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+     *
+     * This is used in the packer to keep the cluster router object around
+     * without a hit to memory footprint. The router can no longer be used after
+     * this function is called.
+     */
+    void clean_router_data();
+
+    /**
+     * @brief Reset the routing state between calls to intra_lb_route for mode
+     *        changes.
```
vaughnbetz, line 309: Not sure if you can expand on what state is reset / what is safe to do?

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+     *
+     * This is used to clean state associated with mode conflicts in routing.
+     */
+    void reset_intra_lb_route();
+
+    /**
+     * @brief Creates an array [0..num_pb_graph_pins-1] for intra-logic block routing lookup.
+     * Given a pb_graph_pin ID for a CLB, this lookup returns t_pb_route corresponding to that
+     * pin.
```
vaughnbetz, line 321: What is a t_pb_route?  A routing tree rooted at that pin?  Do we have route trees rooted only at pins that are drivers (CLB input pins and atom / primitive output pins)?  Good to explain some of that here, and make sure t_pb_route explains the result in its comment.

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+     * @brief Expand all nodes for a given lb_net.
+     */
+    bool try_expand_nodes_(const t_intra_lb_net& lb_net,
+                           t_expansion_node* exp_node,
+                           int itarget,
+                           bool try_other_modes,
+                           int verbosity);
+
+    /**
+     * @brief Expand all nodes found in a route tree into the priority queue.
```
vaughnbetz: Explain the param. I think it's the node to start traversing the tree from (likely the root, but may not have to be I guess).

vaughnbetz: Or maybe the comment is wrong and this just puts one node in the priority_queue (name expand_node_ suggests that)

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+     */
+    void expand_node_(const t_expansion_node& exp_node,
+                      int net_fanout);
+
+    /**
+     * @brief Expand all nodes using all possible modes found in the route tree
+     *        into the priority queue.
+     */
+    void expand_node_all_modes_(const t_expansion_node& exp_node,
```
vaughnbetz, line 428: Explain the params. Also explain how all_modes differs from the normal expand_node_ if you can.

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+                                int net_fanout);
+
+    /**
+     * @brief Expand all edges of an expansion node.
+     */
```
vaughnbetz, line 439: Explain params if you can.

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+    std::vector<t_lb_type_rr_node>* lb_type_graph_;
+
+    // =========================================================================
+    // Logical Netlist Info
+    // =========================================================================
+
+    /// @brief Vector of intra logic cluster_ctx.blocks nets and their connections.
+    std::vector<t_intra_lb_net> intra_lb_nets_;
+
+    /// @brief Save vector of intra logic cluster_ctx.blocks nets and their connections.
```
vaughnbetz, line 495: Explain what it is/can be used for (save & restore in case you find a better routing?  Something else?)

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+    std::vector<t_intra_lb_net> saved_lb_nets_;
+
+    /// @brief Map that records which atoms are added to cluster router.
```
vaughnbetz: what does atoms added to cluster router mean?  I guess I think of adding atoms as a packing decision, rather than something the intra-cluster router needs to separately track.

vpr/src/pack/cluster_router.h
```diff
@@ -4,38 +4,452 @@
...
+
+    // =========================================================================
+    // Stored State Info During Pathfinder Iterative Routing
+    // =========================================================================
+
+    /// @brief [0..lb_type_graph->size()-1] Stores mode exploration and lb_traceback info for nodes.
+    std::vector<t_explored_node_tb> explored_node_tb_;
+
+    /// @brief Used in conjunction with explored_node_tb_ to determine whether or not a location has been explored.
+    ///        By using a unique identifier every route, I don't have to clear the previous route exploration.
```
vaughnbetz, line 522: Could explain this a bit more. Presumably a field in explored_node_tb must match this number to indicate it has already been explored for the current in-progress routing.

vpr/src/pack/cluster_router.h
AlexandreSinger: We actually do not even use the reserve feature anymore. The only reason I kept this class is because it makes clearing the priority queue easier. For some reason, STL does not have a method to clear a priority queue; I prefer to mess with this in isolation, but I think it is more than safe to remove or slightly clean up.

vpr/src/pack/cluster_router.h
AlexandreSinger: I will update the comment to make this more clear. These are the atoms which the router is aware of being added. The overall legalizer will do whatever it wants, but eventually it will tell the router that it would like to add an atom / remove an atom and the router keeps track of which atoms have been added in this set.

PR #3426, opened 2026-02-14
Interposer cut location formula

soheilshahrouz:
Currently, interposer cut locations are specified using fixed integers; formulas based on device height and width are not supported. Implementing formula support will enable dynamic cut locations for auto-sized devices and improve readability for fixed layouts.
[This paper](https://dl.acm.org/doi/abs/10.1145/2554688.2554776) argues that auto-sized devices maximize utilization, which increases the complexity of placement and routing for interposer-based FPGAs. High-utilization test cases are similar to real-world scenarios for interposer-based FPGAs, such as ASIC emulation with high logic demand.

This PR extends interposer and scatter-gather (SG) support so that connection counts and cut locations can be expressed as formulas (e.g. of channel width W and height H), and adds interposer-cut–aware auto layout.

### DIFF
--- a/libs/libarchfpga/src/interposer_types.h
+++ b/libs/libarchfpga/src/interposer_types.h
@@ -43,7 +43,7 @@ struct t_interdie_wire_inf {
 struct t_interposer_cut_inf {
     /// Axis of interposer cut location. The cut is perpendicular to this axis. This specifies the dimension of `loc`.
     e_interposer_cut_type dim;
-    /// Location of the cut on the grid. Locations start from zero and cuts will happen above or to the right of the tiles at location=loc.
+    /// Formula for the location of the cut on the grid. Cuts will happen above or to the right of the tiles at location=loc.
     std::string loc;
     /// Connectivity specification between the two sides of the cut.
     std::vector<t_interdie_wire_inf> interdie_wires;
--- a/libs/libarchfpga/src/grid_util.cpp
+++ b/libs/libarchfpga/src/grid_util.cpp
@@ -84,7 +84,7 @@ resolve_interposer_cut_locations(const DeviceGrid& grid,
     vtr::t_formula_data formula_data;
 
     std::vector<std::vector<int>> horizontal(num_layers);
-    std::vector<std::vector<int>> vertical(num_layers);    
+    std::vector<std::vector<int>> vertical(num_layers);
 
     for (size_t layer = 0; layer < num_layers; layer++) {
         const t_layer_def& layer_def = grid_def.layers[layer];
--- a/vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
+++ b/vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
@@ -358,7 +358,7 @@ void convert_interposer_cuts_to_sg_patterns(const std::vector<t_layer_def>& inte
 
         // Step 2: Process each interposer cut (vertical or horizontal) on this layer
         for (const t_interposer_cut_inf& cut_inf : interposer_inf[layer].interposer_cuts) {
-        
+
             const int cut_loc = adjust_interposer_cut_location(grid, layer, cut_inf.dim, cut_inf.loc,
                                                                formula_parser, formula_data, /*print_warning=*/false);
             e_interposer_cut_type cut_type = cut_inf.dim;


doc/src/arch/reference.rst
```diff
@@ -582,24 +582,25 @@ Grid Layout Example
...
     .. note:: Exactly one of the ``x`` or ``y`` attributes must be specified.
 
-    Defines an interposer cut for modelling 2.5D interposer-based architectures. An interposer cut will cut all connections at location 'loc' along the axis 'dim' Leaving the two sides completely unconnected.
-    To reconnect the two sides, this tag can have multiple <interdie_wire> tags as children to specify the connection between the two sides.
+    Defines an interposer cut for modelling 2.5D interposer-based architectures. An interposer cut will cut all connections at the given location along the specified axis, leaving the two sides completely unconnected.
+    To reconnect the two sides, this tag can have multiple ``<interdie_wire>`` tags as children to specify the connection between the two sides.
+    This tag may appear inside ``<auto_layout>`` or ``<fixed_layout>`` as a grid location tag.
 
-.. arch:tag:: <interdie_wire sg_name="string" sg_link="string" offset_start="expr" offset_end="expr" offset_increment="expr" num="int"/>
+.. arch:tag:: <interdie_wire sg_name="string" sg_link="string" offset_start="expr" offset_end="expr" offset_increment="expr" num="expr"/>
```
AmirhosseinPoolad: offset_start, end and increment should be int.

doc/src/arch/reference.rst
```diff
@@ -2881,7 +2882,7 @@ tag can be inserted under the following XML tags:
 * ``<pb_type>``
 * Any tag under ``<interconnect>`` (``<direct>``, ``<mux>``, etc).
 * ``<mode>``
-* Any grid location type (``<perimeter>``, ``<fill>``, ``<corners>``, ``<single>``, ``<col>``, ``<row>``, ``<region>``)
+* Any grid location type (``<perimeter>``, ``<fill>``, ``<corners>``, ``<single>``, ``<col>``, ``<row>``, ``<region>``, ``<interposer_cut>``)
```
AmirhosseinPoolad: I don't think we support that.

libs/libarchfpga/src/interposer_types.h
```diff
@@ -35,14 +34,17 @@ struct t_interdie_wire_inf {
...
 /**
  * @brief Struct containing information of an interposer cut
  */
 struct t_interposer_cut_inf {
-    e_interposer_cut_type dim;                       ///< Axis of interposer cut location. The cut is perpendicular to this axis. This specifies the dimension of `loc`.
-    int loc;                                         ///< Location of the cut on the grid. Locations start from zero and cuts will happen above or to the right of the tiles at location=loc.
-    std::vector<t_interdie_wire_inf> interdie_wires; ///< Connectivity specification between the two sides of the cut.
+    /// Axis of interposer cut location. The cut is perpendicular to this axis. This specifies the dimension of `loc`.
+    e_interposer_cut_type dim;
+    /// Location of the cut on the grid. Locations start from zero and cuts will happen above or to the right of the tiles at location=loc.
```
AmirhosseinPoolad: Formula for location of the...

libs/libarchfpga/src/read_xml_arch_file_interposer.cpp
```diff
@@ -62,11 +55,11 @@ t_interposer_cut_inf parse_interposer_cut_tag(pugi::xml_node interposer_cut_tag,
...
         }
 
-        interdie_wire.num = pugiutil::get_attribute(interdie_wire_tag, "num", loc_data).as_int();
+        interdie_wire.num = pugiutil::get_attribute(interdie_wire_tag, "num", loc_data).as_string();
 
-        if (interdie_wire.num <= 0) {
+        if (interdie_wire.num.empty()) {
             archfpga_throw(loc_data.filename_c_str(), loc_data.line(interdie_wire_tag),
-                           "The `num` attribute of an `interdie_wire` must be specified with a positive number.");
+                           "The `num` attribute of an `interdie_wire` must be specified (formula or positive integer).");
```
AmirhosseinPoolad, line 62: This never happens since "num" is not an optional attribute. Line 58 will crash the program if num is not specified.

vpr/src/base/setup_grid.cpp
```diff
@@ -337,8 +358,99 @@ static bool grid_satisfies_instance_counts(const DeviceGrid& grid, const std::ma
...
 
-///@brief Build the specified device grid
-static DeviceGrid build_device_grid(const t_grid_def& grid_def, size_t grid_width, size_t grid_height, bool warn_out_of_range, const std::vector<t_logical_block_type_ptr>& limiting_resources) {
+static void resolve_interposer_cut_locations(const vtr::NdMatrix<t_grid_tile, 3>& grid,
+                                             const t_grid_def& grid_def,
+                                             vtr::FormulaParser& p,
+                                             size_t grid_width,
+                                             size_t grid_height,
+                                             std::vector<std::vector<int>>& horizontal_interposer_cuts,
+                                             std::vector<std::vector<int>>& vertical_interposer_cuts) {
```
AmirhosseinPoolad: Remove output arguments, return them instead.

vpr/src/base/setup_grid.cpp
```diff
@@ -337,8 +358,99 @@ static bool grid_satisfies_instance_counts(const DeviceGrid& grid, const std::ma
...
+
+        for (const t_interposer_cut_inf& cut_inf : layer_def.interposer_cuts) {
+            vtr::t_formula_data cut_vars;
+            cut_vars.set_var_value("W", grid_width);
+            cut_vars.set_var_value("H", grid_height);
+            const int base_cut_loc = p.parse_formula(cut_inf.loc, cut_vars);
+
+            // Vertical cut at loc: locations to the right of the cut (column at loc+1) must be root.
+            // Horizontal cut at loc: locations above the cut (row at loc+1) must be root.
+            auto is_cut_through_roots_only = [&grid, layer](e_interposer_cut_type dim, int loc) {
```
AmirhosseinPoolad: Turn this into two lambdas (horz and vert)
Add comments for the indexing and general algorithm
You only need to check height offset for horz cut and width offset of vert cut (double check, might've made a mistake)

vpr/src/base/setup_grid.cpp
```diff
@@ -337,8 +358,99 @@ static bool grid_satisfies_instance_counts(const DeviceGrid& grid, const std::ma
...
+            int cut_loc = base_cut_loc;
+            for (int offset = 0;; offset++) {
+                int try_pos_plus = base_cut_loc + offset;
+                int try_pos_minus = base_cut_loc - offset;
+                if (offset == 0) {
+                    if (is_cut_through_roots_only(cut_inf.dim, try_pos_plus)) {
+                        cut_loc = try_pos_plus;
+                        break;
+                    }
+                } else {
```
AmirhosseinPoolad: Not sure if this optimization is worth the extra code. I think a better approach would be to check the offset = 0 case outside of the loop, then do:
plus_ok = ...
if (plus ok) { ...
}
minus_ok = ...
if (minus_ok) {...
}
with the loop starting from offset 1.

vpr/src/base/setup_grid.cpp
```diff
@@ -337,8 +358,99 @@ static bool grid_satisfies_instance_counts(const DeviceGrid& grid, const std::ma
...
+                                    cut_inf.loc.c_str(), base_cut_loc);
+                }
+            }
+
+            if (cut_loc != base_cut_loc) {
+                VTR_LOG_WARN(
+                    "Interposer cut moved to avoid cutting through block: W=%zu H=%zu formula='%s' evaluated to %d, resolved to %d (%s)\n",
+                    grid_width, grid_height, cut_inf.loc.c_str(), base_cut_loc, cut_loc,
+                    cut_inf.dim == e_interposer_cut_type::VERT ? "VERT" : "HORZ");
+            }
```
AmirhosseinPoolad: Let's keep this warning only for fixed size devices, then print cut locations after packing is done. This warning is going to happen a lot of times for auto sized devices and spam users.

vpr/src/base/setup_grid.cpp
```diff
@@ -337,8 +358,99 @@ static bool grid_satisfies_instance_counts(const DeviceGrid& grid, const std::ma
...
+                    }
+                }
+                if (offset > static_cast<int>(std::max(grid_width, grid_height))) {
+                    VPR_FATAL_ERROR(VPR_ERROR_ARCH,
+                                    "Interposer cut (dim=%s, formula=%s -> %d) does not cross root locations only; "
+                                    "no valid offset found within grid bounds.",
+                                    cut_inf.dim == e_interposer_cut_type::VERT ? "VERT" : "HORZ",
+                                    cut_inf.loc.c_str(), base_cut_loc);
+                }
+            }
```
AmirhosseinPoolad: Change this to an assert safe since it should not happen under normal circumstances (device edges are always valid for cutting).

vtr_flow/arch/interposer/titan_interposer/stratixiv_arch.timing_interposer.xml
```diff
@@ -4866,9 +4866,9 @@
       <col type="M144K" startx="33" starty="2" repeatx="43" priority="100"/>
       <col type="EMPTY" startx="33" repeatx="43" starty="1" priority="99"/>
 
-      <interposer_cut y="81">
```
AmirhosseinPoolad, line 4869: add auto sized titan

libs/libarchfpga/src/read_xml_arch_file_interposer.cpp
soheilshahrouz: num is parsed as string
unlike int, string can be empty.

AmirhosseinPoolad:
LGTM, will merge after tests pass.

PR #3427, opened 2026-02-17
[Pack] Adding priority queue for primitive cost calculation to make pack runtime more scalable with bigger complex blocks

haydar-c:
This PR replaces the existing sequential primitive root selection (get_next_primitive_list) with a priority-queue-based candidate selection.

Previously, we were calculating the cost of each possible primitive a molecule can be placed at each time we try to place that molecule inside a cluster. Instead, this PR suggests to calculate the cost of all possible primitives for a molecule once and use the priority queue to select the next primitive each time. 

The QoR is expected to be identical due to deterministic priority key definition that is matching the old behaviour. It is identical in the experiment below for VTR Largest and for Titan benchmarks.

The pack runtime improvement after this change becomes bigger as the number of fails increase in the molecule placement. 

### VTR Largest Pack Runtime Scaling with fle Number:

| # fle | Pack Time (Old) | Pack Time (With PQ) | Normalized (PQ / Old) |
|-------------------|-----------------|----------------------|------------------------|
| 10                | 33.08          | 31.43               | **0.95**                   |
| 20                | 108.00         | 100.55              | **0.93**                   |
| 40                | 531.38         | 407.86              | **0.77**                   |

Pack Time Old and Pack Time with PQ (1)

### VTR Largest Results  (Average of 3 Seeds, Normalized to Master = 1.00)

| Metric                   | Master | With PQ | Δ (%) |
|--------------------------|--------|---------|-------|
| vtr_flow_elapsed_time    | 1.00   | 0.98    | -2%   |
| num_post_packed_blocks   | 1.00   | 1.00    | 0%    |
| **pack_time**                | **1.00**   | **0.96**    | **-4%**   |
| place_time               | 1.00   | 1.00    | 0%    |
| routed_wirelength        | 1.00   | 1.00    | 0%    |
| critical_path_delay      | 1.00   | 1.00    | 0%    |

### Titan Results (Normalized to Master = 1.00)

| Metric                   | Master | With PQ | Δ (%) |
|--------------------------|--------|---------|-------|
| vtr_flow_elapsed_time    | 1.00   | 0.99    | -1%   |
| max_vpr_mem              | 1.00   | 0.99    | -1%   |
| num_post_packed_blocks   | 1.00   | 1.00    | 0%    |
| **pack_time**                | **1.00**   | **0.96**    | **-4%**   |
| place_time               | 1.00   | 1.02    | +2%   |
| routed_wirelength        | 1.00   | 1.00    | 0%    |
| critical_path_delay      | 1.00   | 1.00    | 0%    |


vaughnbetz:
Looks very good. Please also run on the 7-series and the simplified and complex S10 archs.

vpr/src/pack/cluster_placement.h
```diff
@@ -155,29 +156,75 @@ t_intra_cluster_placement_stats* alloc_and_load_cluster_placement_stats(t_logica
 void free_cluster_placement_stats(t_intra_cluster_placement_stats* cluster_placement_stats);
 
 /**
- * get next list of primitives for list of atom blocks
+ * @brief Builds a priority queue of feasible primitive root candidates for a molecule.
```
vaughnbetz, line 159: Say what the queue is sorted on, at a high level (decreasing order of desirability to be placed at a given location within the cluster).
Maybe also explain what a 'primitive root candidate is' (location in the cluster where the primitive that is the start of the currently being considered molecule could go?)

vpr/src/pack/cluster_placement.h
```diff
@@ -155,29 +156,75 @@ t_intra_cluster_placement_stats* alloc_and_load_cluster_placement_stats(t_logica
...
- *                   Expects an allocated array of primitives ptrs as inputs.
- *                   This function loads the array with the lowest cost primitives that implement molecule
- * force_site - optional user-specified primitive site on which to place the molecule; if a force_site
+ * @param cluster_placement_stats
+ *              Placement statistics and primitive state information for the current cluster.
+ *              This structure maintains valid, tried, and in-flight primitive sets.
+ * @param molecule_id
+ *              Identifier of the molecule to be evaluated for placement.
+ * @param primitives_list
+ *              A list of primitives indexed to match atom_block_ids of molecule.
```
vaughnbetz: Is this an ordered list of the primitives (atom_block_ids) that form the molecule being considered?  If so, could probably write this a bit more explicitly.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1234,16 +1236,24 @@ e_block_pack_status ClusterLegalizer::try_pack_molecule(PackMoleculeId molecule_
...
-                                     molecule_id,
-                                     primitives_list,
-                                     prepacker_)) {
+        if (primitives_alive.empty()) {
             VTR_LOGV(log_verbosity_ > 3, "\t\tFAILED No candidate primitives available\n");
             block_pack_status = e_block_pack_status::BLK_FAILED_FEASIBLE;
             break; /* no more candidate primitives available, this molecule will not pack, return fail */
         }
 
+        std::pair<t_pb_graph_node*, std::tuple<float, int, int>> primitive = primitives_alive.pop();
```
AlexandreSinger, line 1251: I understand why you are using this tuple here, it is for sorting by the float first, then the second int in the case of a tie, and then the third int in the case of a tie. Its a bit strange, so I recommend adding a comment here. I would also add a TODO to explore other cost schemes. I wonder if using some weighted sum of these terms could be beneficial.

AlexandreSinger (approved):
This looks good to me! I left a minor point that we should comment the priority in the code somewhere clear and add a TODO to explore a single float cost scheme (not for this PR but in the future).

vpr/src/pack/cluster_legalizer.cpp
haydar-c: Added the explanation and TODO to cluster_placement.h, build_primitive_candidate_queue function doxygen comment,

vpr/src/pack/cluster_placement.h
haydar-c: Added what is a primitive root candidate and explained how do queue is sorted.

Issue #3428, opened 2026-02-17
Fan-in functions in RRGraphView silently fail without warning/erroring out


AmirhosseinPoolad:
Some functions in RRGraphView expect some additional data structures that are not initialized for normal RR Graphs. These functions silently return wrong values. The node fan-in data structure should be moved out of the main RR Graph classes and into their own class to avoid having a minefield API where you're never sure which functions you're allowed or not allowed to use.

PR #3429, opened 2026-02-17
[RR Graph] Adding RRGraphInEdges Class

amin1377:
This PR addresses Issue #3428.


vpr/src/route/rr_graph_generation/tileable_rr_graph/tileable_rr_graph_utils.h
```diff
@@ -12,6 +12,8 @@
 #include "rr_graph_obj.h"
 #include "rr_graph_view.h"
 
+class RRGraphInEdges;
+
```
AmirhosseinPoolad: I don't think this class exists anymore, right?

PR #3431, opened 2026-02-19
Introduce data structure to memoize cluster packing patterns, improving runtime.

miloliebster:
#### Description

This PR includes three commits, with the second of these commits being the primary contribution.

The first commit fixes a niche bug that for specific architectures can cause the speculative packing step to give up trying to route a cluster and fall back to detailed packing when the cluster it has packed is legal, and is the exact same legal cluster that detailed packing will produce. Previously, this bug has not impacted packer functionality, but with the addition of memoization we saw clusters with the same key reporting to be simultaneously legal and illegal, which lead to us identifying this issue.

The second commit introduces a new class called the PackingSignatureTree, which is stored as a member of ClusterLegalizer. This class stores a tree of nodes representing packed primitives and the connections between them (in the order that they were packed), representing decisions that the packer has made while populating clusters, and storing the results of legality checks as values that can be looked up by traversing the tree. When a familiar cluster arrangement is encountered during detailed packing, the packer can skip rerunning the routing step and instead lookup the legality status in the tree, saving significantly on runtime. packing_signature_tree.h goes more in depth on the theory and implementation of the packing signature tree; please let me know if there are parts you find could use further clarification.

The existing Stratix 10 architecture capture included with VTR simplified the LAB architecture to reduce the runtime burden on the packing step. In order to stress test the effectiveness of packing memoization, we modified this architecture to be more faithful to the real Stratix 10 LABs, adding sparse crossbars and more constrained connectivity to flip-flops. Our tests showed that with memoization enabled, the runtimes have been cut significantly to durations that are now reasonable to evaluate, so in the third commit we are contributing this modified architecture description.

#### Motivation and Context

For architectures with complex intracluster connectivity (such as the VTR 7-Series capture, and the modified Stratix 10 capture added in the third commit of this PR), we've measured that over half of the average VPR runtime is spent in the packing stage doing detailed routing. For large circuits on complex architectures, runtimes can grow to unreasonable lengths, as was the case with the original Stratix 10 architecture capture, leading to the LAB model being simplified. Memoizing previous work done during packing directly addresses these runtime issues for architectures that speculative packing is not as effective on, making evaluation of modern FPGA logic blocks more practical in VPR.

Tested circuits included VTR and Koios benchmarks (over 10,000 LEs) evaluated on `7series_BRAM_DSP_carry` and `k6_frac_N10_frac_chain_mem32K_40nm`, as well as a subset of Titan and Titanium benchmarks on the modified Stratix 10 architecture capture. Runtime and memory results for the VTR and Koios circuits are provided in the attached files, showing speedup and %memory increase for the packing stage and complete VPR flow with memoization enabled and disabled. These were run through `run_vtr_task` with `-j 32` on an Intel Xeon w9-3495 CPU with 1 TB of RAM.

The attached tables show how this optimization targets architectures such as the 7-Series where speculative packing is far less successful due to more complicated logic block architectures. On the other hand, the VTR flagship architecture, which has a full crossbar and is therefore easy to route successfully with speculative packing, sees effectively no speedup; however the packing signature is structured in a way where the memory overhead in the packing stage is negligible if detailed packing is rarely required, as is the case for flagship. Even on complex architectures where the packing signature tree gets larger, packing never exceeds the peak memory reached by later VPR stages for any of the circuits we've tested, so overall peak memory is generally unaffected...

I don't have a compelling explanation—besides maybe allocator fragmentation?—for the handful of 7-Series circuits where the peak VPR memory noticeably increases. The packing signature tree gets fully deallocated at the end of the packing stage, and I've run `lenet` through valgrind and confirmed I haven't leaked nodes from the tree. It's consistent behaviour on these circuits, but the mechanism eludes me.

To verify these changes only impact runtime, we have checked the diffs of QoR reports between runs with and without memoization enabled, as well as against the results from the master branch. For the same seeds, these files have all been completely identical.

(The Stratix 10 runs for the current build are still underway and I will add them here when the remaining runs finish in one or two days. Previous tests we've done showed a 6.9x geomean speedup to packing, and 5.31x for the complete flow, bringing the longest VPR flow time from 76.8 hours down to 6.8 hours.)

[7series_BRAM_DSP_carry.txt](https://github.com/user-attachments/files/25405344/7series_BRAM_DSP_carry.txt)
[k6_frac_N10_frac_chain_mem32K_40nm.txt](https://github.com/user-attachments/files/25405345/k6_frac_N10_frac_chain_mem32K_40nm.txt)

#### Types of changes

#### Checklist:

Regression tests will be created and added this week. I'm beginning the PR process now so we can get some eyes on the code in the meantime.


doc/src/vpr/command_line_usage.rst
```diff
@@ -766,6 +766,14 @@ For people not working on CAD, you can probably leave all the options to their d
 
     **Default:** ``30``
 
+.. option:: --memoize_cluster_packings {on | off}
+
+    Enables memoization of previously seen clusters during packing.
+
+    This can significantly reduce runtime for certain architectures by skipping redundant intracluster routing calls made to test for cluster legality
+
+    **Default:** ``on``
```
AlexandreSinger: I recommend setting this to false in this first PR and adding two testcases: One that tests this feature with this option on, and one that tests this feature with this option off. This will just lock in the functionality of this option.

Then in a future PR, you can just turn this on and update the strong tests (and check QoR). This just makes it easier to separate your functional changes from your resulting quality! The aim is so we can verify that the code behaves like normal without this option, but then only changes when we add this option.

vpr/src/base/read_options.cpp
```diff
@@ -2288,6 +2288,15 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("30")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    pack_grp.add_argument<bool, ParseOnOff>(args.memoize_cluster_packings, "--memoize_cluster_packings")
+        .help("Enables memoization of previously seen clusters during packing.\n"
+              "\n"
+              "This can significantly reduce runtime for certain architectures\n"
+              "by skipping redundant intracluster routing calls made to test\n"
+              "for cluster legality.")
+        .default_value("on")
```
AlexandreSinger: See comment above about making this off by default for just this PR, and then making it on by default in a second PR.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
...
+#include <tuple>
+#include <unordered_map>
+#include <vector>
+
+#include "atom_netlist.h"
+#include "physical_types.h"
+#include "vtr_strong_id.h"
+
+// Define to avoid circular dependancy from including cluster_legalizer.h
+typedef vtr::StrongId<struct legalization_cluster_id_tag, size_t> LegalizationClusterId;
```
AlexandreSinger: I am not sure why this is necessary since the `#pragma once` should be protecting against this. Its not really great practice to redefine a typedef like this. It will work but its not super stable.

The classic way of handling this is to instead create a `cluster_legalizer_fwd.h` file and put this declaration in there. Then `cluster_legalizer.h` and this file can include just that header file independently. This was something I should have done when creating this `cluster_legalizer.h` file, but I thought it was silly to just put this typedef in its own file if only one header file needs it. It looks like that has come back to bite us!

An example of how we do this is in VTR can be found here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d06d4ea726e2ef65284ec4e89f47547f8cc46861/vpr/src/base/atom_netlist_fwd.h#L1

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1281,6 +1282,7 @@ e_block_pack_status ClusterLegalizer::try_pack_molecule(PackMoleculeId molecule_
             }
         }
 
+        packing_signature_tree_.set_checkpoint();
```
AlexandreSinger: I like what you have done here! I recommend adding some high-level comments to help the reader understand why we are setting checkpoints, adding locations, checking for legality, and rolling back when we are.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1598,7 +1617,26 @@ bool ClusterLegalizer::check_cluster_legality(LegalizationClusterId cluster_id)
     // route on the cluster. If it succeeds, the cluster is fully legal.
     t_mode_selection_status mode_status;
     LegalizationCluster& cluster = legalization_clusters_[cluster_id];
-    return cluster.cluster_router.try_intra_lb_route(log_verbosity_, &mode_status);
+    bool routed;
```
AlexandreSinger, line 1673: This is very interesting! This was the original function of the router to only do a final route, but you are correct that it would make a lot more sense to try different modes through routing. This should probably be isolated from your PR since this can have an affect on QoR (for the better!). I am very interested on what the affect of this is on the packer.

Why did you need to change this though? Did you find something was wrong? I am genuinely curious on this!

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1598,7 +1617,26 @@ bool ClusterLegalizer::check_cluster_legality(LegalizationClusterId cluster_id)
...
+    do {
+        cluster.cluster_router.reset_intra_lb_route();
+        routed = cluster.cluster_router.try_intra_lb_route(log_verbosity_, &mode_status);
+    } while (mode_status.is_mode_issue());
+    return routed;
+}
+
+bool ClusterLegalizer::ensure_legal_final_routing(LegalizationClusterId cluster_id) {
```
AlexandreSinger, line 1681: I am concerned on the name of this function. The name implies that this is just checking that the final routing is legal, but it looks like the packing signature tree is being updated. I recommend changing to `ensure_legal_final_routing_and_update`.

Also I recommend adding some high-level comments here to make it more clear what is happening in this function.

vtr_flow/arch/titan/stratix10_arch.timing.depop50.complex_alm.xml
AlexandreSinger, line 1: In the comment at the top of this architecture, you should add a comment explaining what makes this architecture different from the original S10 architecture.

We should also add at least one strong test (on a small design) with this architecture to ensure that it works correctly.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
+#pragma once
+/**
+ * @file
+ * @author  Milo Liebster
+ * @date    January 2026
+ * @brief   The declaration of the PackingSignatureTree class.
+ *
+ * A structure for memoizing the results of legality checks during
+ * packing to significantly improve packing runtimes for complex
+ * logic block architectures.
```
AmirhosseinPoolad: I think you should add to the file-level comment here a definition of the LCN and ECN and PST. You have them in the comments for the PST class but it's a bit confusing to read through the file without knowing what these terms mean.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
...
+        size_t external_sinks_count;
+    };
+
+    /// @brief Bookkeeping of all nets that drive pins within the open cluster
+    ///        as well as the pins that the nets drive. Used for generating
+    ///        the cluster_inputs field in ECNs.
+    ///
+    /// If a net in this map is also present in output_nets_, then it is known
+    /// to be driven internally and is excluded from generated ECNs.
+    std::map<AtomNetId, std::vector<PstPinId>> input_nets_;
```
AmirhosseinPoolad: For this and the other std::maps, any particular reason that you need a binary search tree instead of a simple hashmap? As in, why not std::unordered_map?

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
...
+    /// With the PST, it is possible for the packer to reach a final cluster
+    /// packing without ever routing the cluster. If this value is false when
+    /// the packer is finalizing a cluster, it will run the cluster router so
+    /// that a routing exists for later VPR stages.
+    bool routed;
+
+  private:
+    /// @brief If false, then class methods exit immediately and no PST is
+    ///        constructed.
+    const bool enabled_;
```
AmirhosseinPoolad: Kind of disagree with this being part of the class. Why not make the instantiation be in a std::optional and do the checking from the outside?

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
...
+
+  public:
+    /// @brief Used for the packer to track whether it has a valid routing for
+    ///        a finalized cluster.
+    ///
+    /// With the PST, it is possible for the packer to reach a final cluster
+    /// packing without ever routing the cluster. If this value is false when
+    /// the packer is finalizing a cluster, it will run the cluster router so
+    /// that a routing exists for later VPR stages.
+    bool routed;
```
AmirhosseinPoolad: I don't think this should be a public field. If the packer code does "pst.routed = true" it would break this class's invariants/assumptions.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
...
+/// this is only once the cluster is finalized; for detailed packing, this
+/// means there is one ECN for the last atom of every molecule inserted. The
+/// PackingSignatureTree class also records what changes have been made to its
+/// external connectivity bookkeeping structures since the previous legal
+/// molecule was inserted so that it can rollback the state of its bookkeeping
+/// if the current molecule is illegal and needs to be unpacked. The LCN(s)
+/// and ECN added for the illegal molecule stay in the PST so that future
+/// clusters can recognize the illegal state and skip performing legalization
+/// by routing.
+class PackingSignatureTree {
```
AmirhosseinPoolad, line 382: If I understood the API correctly, this is a stateful API. You essentially have a "cursor" at the latest thing added to the tree and and check_legality() gives you the legality state of where the cursor is at. Add an explanation of this and the general intended way to correctly use this class to the class-level comment here.

doc/src/vpr/command_line_usage.rst
amin1377: I also think it would be better not to enable new features by default. Once more people start using the feature and the code has been tested more thoroughly, we can enable it by default.

amin1377: I’d also suggest briefly mentioning the type of architecture that benefits most from this update (namely, architectures with sparser and more irregular crossbar structures, rather than architectures with full crossbars where intra-cluster routing is relatively straightforward)

vpr/src/pack/packing_signature_tree.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    checkpoint_lcn_ = at_lcn_;
+    checkpoint_input_nets_.clear();
+    checkpoint_new_output_nets_.clear();
+    checkpoint_decremented_output_nets_.clear();
+    checkpoint_new_atoms_.clear();
+}
+
+void PackingSignatureTree::rollback_to_checkpoint() {
+    at_lcn_ = checkpoint_lcn_;
```
amin1377: checkpoint_lcn_ is never initialized in the constructor (unlike at_lcn_ which is set to nullptr). When memoize_cluster_packings is off, rollback_to_checkpoint is still called from try_pack_molecule on every failed molecule pack, it will read an uninitialized pointer, which is undefined behavior. I'd suggest adding 
`if (!enabled_)  return;` at the top.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
...
+  public:
+    PackingSignatureTree() = delete;
+    PackingSignatureTree(bool enabled)
+        : routed(false)
+        , enabled_(enabled)
+        , at_lcn_(nullptr) {}
```
amin1377: Similar to my earlier comment, I’d suggest initializing checkpoint_lcn_ to nullptr in the constructor.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,341 @@
...
+/// This is quite a large tuple (>32 bytes), and pin identifier copies make up
+/// most of the packing signature tree's memory footprint. To save on memory
+/// this type gets computed and then used as a key to look up a derived 4-byte
+/// pin ID (PstPinPd) that gets stored in the tree instead.
+typedef std::tuple<t_primitive_num, std::string, BitIndex> PstPin;
+
+/// @brief Hash function definition for PstPin.
+struct PstPinHash {
+    size_t operator()(PstPin pin) const noexcept {
+        return std::hash<t_primitive_num>{}(std::get<0>(pin)) ^ std::hash<std::string>{}(std::get<1>(pin)) ^ std::hash<int>{}((int)std::get<2>(pin));
```
amin1377: XOR-based hash combination is a well-known pitfall. I’d suggest using VTR’s hash utility function instead, since it performs a more robust combine step (`shift+ add`):

```cpp
struct PstPinHash {
    size_t operator()(PstPin pin) const noexcept {
        size_t seed = 0;
        vtr::hash_combine(seed, std::get<0>(pin));
        vtr::hash_combine(seed, std::get<1>(pin));
        vtr::hash_combine(seed, static_cast<int>(std::get<2>(pin)));
        return seed;
    }
};
```

vpr/src/pack/packing_signature_tree.cpp
```diff
@@ -0,0 +1,263 @@
...
+    std::sort(new_ecn->cluster_outputs.begin(), new_ecn->cluster_outputs.end());
+
+    return new_ecn;
+}
+
+e_ecn_legality PackingSignatureTree::check_legality() {
+    // Return immediately if packing memoization is disabled
+    if (!enabled_) return e_ecn_legality::UNKNOWN;
+
+    ExternalConnectivityNode* ecn = this->create_ecn();
```
amin1377: This is called on every molecule during detailed packing (the hot path). It heap-allocates and populates a full ExternalConnectivityNode just to compare it against existing children, then immediately deletes it. A stack-local ExternalConnectivityNode would be a more efficient choice I think.

vpr/src/pack/packing_signature_tree.cpp
miloliebster: Good catch! There was a suggestion to move the `enabled_` check outside of the PST class, so it can remain uninstantiated if the feature is disabled, so I will probably do that, but `checkpoint_lcn_` should definitely be initialized to something.

vpr/src/pack/packing_signature_tree.h
miloliebster: Cool, will do

vpr/src/pack/packing_signature_tree.cpp
miloliebster: My thinking is that the common case here is just the allocator invalidating the pointer to the ECN in an allocated page, thus freeing space that the next allocated node will reuse (so very few extra syscalls needed). But I can change this.

vpr/src/pack/packing_signature_tree.h
miloliebster: I do a lot of iterating over maps, and at some point somehow convinced myself that the binary trees were better for that, but that doesn't really make sense. These should probably be hashmaps like you say.

vpr/src/pack/packing_signature_tree.h
miloliebster: It is the packer's job to make sure this flag is accurate and then decide whether to run a final routing step based on this flag, so I'm moving this to be a private variable owned by ClusterLegalizer. I think this makes more sense rather than having the awkward public variable.

vpr/src/pack/cluster_legalizer.cpp
miloliebster: I find the "_and_update" to be even less specific. What I mean to communicate is "check if the cluster has a routing; if it doesn't, route it". This method is a bit awkward in general, but necessitated by encapsulation since this step is initiated by the greedy clusterer, but the cluster legalizer owns the PST (I am not a fan of strict OOP for these sorts of scenarios, but such is life). Anyways, I will think more on what an appropriate name would be.

doc/src/vpr/command_line_usage.rst
```diff
@@ -766,6 +766,14 @@ For people not working on CAD, you can probably leave all the options to their d
 
     **Default:** ``30``
 
+.. option:: --memoize_cluster_packings {on | off}
+
+    Enables memoization of previously seen clusters during packing.
+
+    This can significantly reduce runtime for certain architectures by skipping redundant intracluster routing calls made to test for cluster legality
```
vaughnbetz: I'd expand this a bit.  
'for architectures with complex intra-cluster interconnect by skipping redundant ...'
'It generally reduces clustering runtime, and does not affect quality metrics like routed wirelength or critical path delay.'

(Or something like that).

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -985,6 +985,7 @@ void FlatRecon::legalize(const PartialPlacement& p_placement) {
         high_fanout_thresholds,
         ClusterLegalizationStrategy::SKIP_INTRA_LB_ROUTE,
         vpr_setup_.PackerOpts.enable_pin_feasibility_filter,
+        false, // --memoize_cluster_packings is ineffective with AP, so force off.
```
vaughnbetz: Not sure this comment is true anymore. I'm OK with forcing it off, but if you no longer force it off by the time the PR is merged this comment should be updated.

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -1020,6 +1021,7 @@ void NaiveFullLegalizer::create_clusters(const PartialPlacement& p_placement) {
                                        high_fanout_thresholds,
                                        ClusterLegalizationStrategy::FULL,
                                        vpr_setup_.PackerOpts.enable_pin_feasibility_filter,
+                                       false, // --memoize_cluster_packings is ineffective with AP, so force off.
```
vaughnbetz: Same comment as above.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1321,17 +1327,43 @@ e_block_pack_status ClusterLegalizer::try_pack_molecule(PackMoleculeId molecule_
...
             t_mode_selection_status mode_status;
-            bool is_routed = false;
+            e_ecn_legality legality = e_ecn_legality::UNKNOWN;
             bool do_detailed_routing_stage = (cluster_legalization_strategy_ == ClusterLegalizationStrategy::FULL);
             if (do_detailed_routing_stage) {
-                do {
-                    cluster.cluster_router.reset_intra_lb_route();
-                    is_routed = cluster.cluster_router.try_intra_lb_route(log_verbosity_, &mode_status);
-                } while (do_detailed_routing_stage && mode_status.is_mode_issue());
+                if (packing_signature_tree_) {
```
vaughnbetz, line 1358: Probably a good idea to comment what the overall idea of the below code is.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1321,17 +1327,43 @@ e_block_pack_status ClusterLegalizer::try_pack_molecule(PackMoleculeId molecule_
              *
              * expand_all_modes is used to enable the expansion of all the nodes using all the possible modes.
              */
+
```
vaughnbetz, line 1329: Briefly comment what the overall goal / operation of the code below is.

vpr/src/pack/cluster_legalizer.h
```diff
@@ -395,6 +396,23 @@ class ClusterLegalizer {
      */
     bool check_cluster_legality(LegalizationClusterId cluster_id);
 
+    /*
+     * @brief Ensure that the cluster has a legal final routing.
+     *
+     * Checks whether the active routing strategy has produced a legal final
+     * routing. For SKIP_INTRA_LB_ROUTE packing, this is the first time routing
```
vaughnbetz: I'd update this a bit.
I'd change the 'Checks whether' sentence to 
'Checks that the cluster is routable, and if it is routable, ensures the routing is stored in the appropriate clustering data structures. '

With that comment (or something similar), I'm fine with the current routine name; we don't have a standard name for this kind of interface.

vpr/src/pack/cluster_legalizer_fwd.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+// A special ID to identify the legalization clusters. This is separate from the
+// ClusterBlockId since this legalizer is not necessarily tied to the Clustered
+// netlist, but is used as a sub-routine to it.
```
vaughnbetz: not sure what 'is used as a subroutine to it' means here. Can you clarify?

vpr/src/pack/packing_signature_tree.h
vaughnbetz: Should add a reference to the FCCM paper for further background.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+
+/// @brief Type alias for primitive_num from t_pb_graph_node.
+typedef int t_primitive_num;
+
+/// @brief Unique cluster-relative pin identifier assembled from accessible VPR
+///        types.
+///
+/// The packing signature tree requires cluster-relative pin IDs, but the VPR
+/// packing datatypes only implement global pin identifiers. As a workaround:
+/// the primitive_num, port name, and bit index of a pin uniquely identify
```
vaughnbetz: I'd explain that these are all identifiers related to the cluster architecture, not the design (e.g. primitive_num is the number of the primitive within the cluster architecture), which is important in matching repeated use of the same pattern (signature) in design clusters.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+        return hash;
+    }
+};
+
+/// @brief Derived 4-byte cluster-relative pin identifier.
+typedef vtr::StrongId<struct pst_pin_id_tag> PstPinId;
+
+/// @brief Pair of pins representing a connection in the packing signature tree.
+struct PstConnection {
+    PstPinId source;
```
vaughnbetz: Should comment every data member, even source and sink (driving pin, receiving pin). Be very explicit; this is a connection between two pins in the pb structures, or two pins in the primitive netlist?

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+    /// @brief Pins that are driven by a source outside the cluster.
+    ///
+    /// Pins that share a source external to the cluster are grouped together.
+    std::vector<std::vector<PstPinId>> cluster_inputs;
+
+    /// @brief Pins that drive one or more sinks outside the cluster.
+    std::vector<PstPinId> cluster_outputs;
+
+    /// @brief Legality of the packing signature that this node terminates
+    ///        (whether a routing solution exist for this configuration).
```
vaughnbetz: exist -> exists

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+    /// @brief List of input connections to the atom represented by this node
+    ///        that are driven by atoms that have been packed up until this
+    ///        point.
+    std::vector<PstConnection> intracluster_sources_to_primitive_inputs;
+
+    /// @brief List of output connections from the atom represented by this node
+    ///        that drive atoms that have been packed up until this point.
+    std::vector<PstConnection> intracluster_sinks_of_primitive_outputs;
+
+    /// @brief Child nodes representing atoms packed after the one represented
```
vaughnbetz, line 346: lcn hasn't been defined yet; good to spell out here or add an earlier definition.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+    ///        by this node.
+    std::vector<LocationAndConnectivityNode*> child_lcn;
+
+    /// @brief Child leaf nodes that encode distinct external connection
+    ///        variations for the packing signature that is described by the
+    ///        LCNs that have been traversed (from the root, up to and including
+    ///        this one).
+    std::vector<ExternalConnectivityNode*> child_ecn;
+
+    LocationAndConnectivityNode()
```
vaughnbetz, line 355: Explain why -1 is a reasonable sentinel value. primitive_nums start at 0 and are contiguous?

vpr/src/pack/packing_signature_tree.h
vaughnbetz: Agreed -- that would be very helpful.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+
+    /// @brief Record the external connections and legality outcome of a
+    ///        cluster packing pattern in the PST.
+    void add_ecn(e_ecn_legality legality);
+
+    /// @brief Get legality status of current cluster packing pattern, if known.
+    e_ecn_legality check_legality();
+
+  private:
+    /// @brief Generate an LCN from the current PST state and provided arguments.
```
vaughnbetz, line 425: I'd comment exactly what the parameters encode.

vpr/src/pack/packing_signature_tree.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    return lcn;
+}
+
+void PackingSignatureTree::add_ecn(e_ecn_legality legality) {
+    ExternalConnectivityNode* ecn = new ExternalConnectivityNode(legality);
+    this->populate_ecn(ecn);
+
+    for (ssize_t i = at_lcn_->child_ecn.size() - 1; i >= 0; i--) {
```
vaughnbetz: Comment what you're doing here. Avoiding duplicate ecns?

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+///
+/// Theory:
+/// Because the VPR packer is greedy and deterministic, repeated structures that
+/// show up in circuits will tend to be processed identically by the packer and
+/// produce matching clusters. Tracking seen cluster patterns in a decision tree
+/// lets us memoize and reuse cluster legalization outcomes.
+///
+/// The PackingSignatureTree (PST) is made up of LocationAndConnectivityNodes
+/// (LCNs) and ExternalConnectivityNodes (ECNs), as shown here:
+///
```
vaughnbetz: Good overview comment!

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+    /// @brief Get legality status of current cluster packing pattern, if known.
+    e_ecn_legality check_legality();
+
+  private:
+    /// @brief Generate an LCN from the current PST state and provided arguments.
+    LocationAndConnectivityNode* create_lcn(const t_pb_graph_node* primitive_pb_graph_node, const AtomBlockId atom_block_id);
+
+    /// @brief Generate an ECN from the current PST state and record its legality,
+    ///        if provided.
+    void populate_ecn(ExternalConnectivityNode* ecn);
```
vaughnbetz, line 437: Explain parameter. ecn node that will be filled in (must point to an ecn node, whose data will be overwritten).

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+    /// type corresponding to each index can be found at the same index in
+    /// branch_logical_block_types_.
+    std::vector<LocationAndConnectivityNode*> branches_;
+
+    /// @brief List of logical block types where the indices correspond to the
+    ///        respective branch in branches_ for that logical block type.
+    std::vector<const t_logical_block_type*> branch_logical_block_types_;
+
+    /// @brief Current active node for the signature being constructed.
+    LocationAndConnectivityNode* at_lcn_;
```
vaughnbetz: This is a pretty subtle concept; suggest expanding on it ('cursor' where the node indicates the state of the cluster being constructed?)

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+            PstPinId pin_id = PstPinId(pin_mappings_.size());
+            pin_mappings_[pin] = pin_id;
+            return pin_id;
+        }
+        return got->second;
+    }
+
+    /// @brief Tracks current number of external sinks for a given source pin.
+    struct ExternalSinksRecord {
+        PstPinId pin;
```
vaughnbetz: Comment data members.  pin must be a source pin?  Might want to rename it to src_pin in that case.

vpr/src/pack/packing_signature_tree.h
```diff
@@ -0,0 +1,326 @@
...
+    ///        cluster as well as a count of sinks that the net drives outside
+    ///        of the cluster.
+    ///
+    /// The external sinks count gets initialized to the total number of sinks
+    /// driven by this net and decremented for every pin this net drives inside
+    /// the cluster. If this count is zero when an ECN is generated, then this
+    /// net has no external sinks and is excluded from the ECN.
+    std::unordered_map<AtomNetId, ExternalSinksRecord> output_nets_;
+
+    /// @brief Value of at_lcn_ for most recent checkpoint.
```
vaughnbetz: Explain when you need / use checkpoints.

vpr/src/pack/packing_signature_tree.h
vaughnbetz: I suggest giving a bit more of an overview of the file and an example in psuedocode/brief of how the class would be used. You have a good comment later on the PST tree, but it comes later and still doesn't give a typical usage case of the class.

See the netlist Doxygen (on the web) for a good example: it comments the purpose and an example of how to use the class before diving into its detailed implementation.

miloliebster:
The flat-recon changes touch most of the PST methods, so that is probably best left to a later PR so there's not a need to re-review everything. I think for now I will leave memoization disabled for flat-recon.

vpr/src/pack/packing_signature_tree.cpp
```diff
@@ -0,0 +1,242 @@
...
+e_ecn_legality PackingSignatureTree::check_legality() {
+    ExternalConnectivityNode ecn = ExternalConnectivityNode(e_ecn_legality::UNKNOWN);
+    this->populate_ecn(&ecn);
+
+    for (ssize_t i = at_lcn_->child_ecn.size() - 1; i >= 0; i--) {
+        if (*at_lcn_->child_ecn[i] == ecn) {
+            return at_lcn_->child_ecn[i]->legality;
+        }
+    }
+    return e_ecn_legality::UNKNOWN;
```
vaughnbetz, line 247: Comment this.  Not found in tree --> don't know if this packing signature is legal or not.

vaughnbetz (changes requested):
Very nice code overall! Some requests for extra commenting below.

AlexandreSinger (approved):
LGTM. Thanks for adding the regression tests! I want to ensure any future changes to packing do not break your code. It was a good idea to add these to both the strong and weekly tests.

I am ok with the downsampled murax; the strong test for this architecture, to me, should really just be ensuring that the architecture (and your changes) are at least passing code through. Its not fun when on Sunday we find that the weekly tests are failing and we have to find the needle in the haystack of commits that broke something.

AlexandreSinger:
Once you update the golden results, it looks like you will need to update your branch with Master in order for this to merge. We may want to also run the weekly test github action to ensure everything is green.

miloliebster:
@AlexandreSinger I pushed the golden result for the weekly test earlier today and have just rebased onto master. Should be set to run the weekly test action, but if I have permissions to kick that off that myself, I am not sure how.

vaughnbetz (approved):
LGTM. Thanks for adding the extra tests.

PR #3433, opened 2026-02-20
Fix interposer cut location warning condition in grid_util.cpp

soheilshahrouz:
Replaced `==` with `!=`.


AmirhosseinPoolad (approved):
LGTM

PR #3443, opened 2026-03-02
[Pack] Hot-Starting the Cluster Router

AlexandreSinger:
When the packer is doing full intra-lb routing for each molecule added to the cluster, it completely restarts the routing from scratch; however, we can use the prior routing solution to "hot-start" the cluster router.

During each intra-lb route, we take the previous routing solution and use that as the starting point. We mark any nets that have changed since the last route as dirty. Then, only new and dirty nets are routed.

This should not change feasibility since we already know the previous solution was correct and if any new nets interract with the old nets, they will negotiate using Pathfinder.

I also tuned the pres-fac and historical-fac of the cluster router to make it try to resolve congestion faster. I found that this speeds up the cluster router considerably.

This improves the run-time for full intra-lb routing by 5-6x on local experiments.


AlexandreSinger:
I have redone how the hot-starting worked. After some experimentation, I actually found that marking the dirty nets was actually take more time than it was worth. I was also running into issues where nets were being marked dirty when they didn't need to due to how the packer is set up. Making the algorithm fully-lazy (having it check just before routing which nets are dirty) ended up being a much better approach.

I also found an issue with the memoization approach which was breaking this change. It was a similar fix: we needed a function to check if the saved routing is valid, instead of relying on what was the last result of intra-lb routing. This is also a much safer approach.

Since it is not clear if this is an overall improvement on normal architectures, I have left this off by default so we can test on more architectures later. I am collecting results on an internal architecture (with very large cluster sizes) now.

AlexandreSinger:
These are the results from on MCNC on the architecture with large clusters:

  | pack_time | vtr_flow_elapsed_time | max_vpr_mem | min_chan_width | routed_wirelength | critical_path_delay
Baseline | 39.24105446 | 60.72424501 | 89423.44118 | 41.4959958 | 21301.3211 | 4.756205698
Hot Starting | 17.36916544 | 35.98197967 | 89618.66208 | 41.76690063 | 21541.49125 | 4.699331317
Memoization | 32.78999343 | 54.55754016 | 96615.38221 | 41.4959958 | 21301.3211 | 4.756205698
Hot Starting + Memoization | 14.99001563 | 32.31904832 | 96034.92231 | 41.71673751 | 21421.65712 | 4.757737943

These are the same results, but normalized.
  | pack_time | vtr_flow_elapsed_time | max_vpr_mem | min_chan_width | routed_wirelength | critical_path_delay
Baseline | 1.00 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00
Hot Starting | 0.44 | 0.59 | 1.00 | 1.01 | 1.01 | 0.99
Memoization | 0.84 | 0.90 | 1.08 | 1.00 | 1.00 | 1.00
Hot Starting + Memoization | 0.38 | 0.53 | 1.07 | 1.01 | 1.01 | 1.00

Hot starting is a clear win in this case, and the improvement combines with memoization; however, they do eat into each other's gains. I am a bit concerned with how these two algorithms will interract since memoization may think that a routing is possible when it may not be able to be found. I am not sure what would happen in that case.

vaughnbetz (approved):
LGTM, thanks.

vaughnbetz:
I suspect it's worth running Titan with this change too (on and off) to confirm QoR isn't affected when it is off at least (and hopefully nothing horrible happens when it is on too).

AlexandreSinger:
> I suspect it's worth running Titan with this change too (on and off) to confirm QoR isn't affected when it is off at least (and hopefully nothing horrible happens when it is on too).

I think that is a good idea! It should only be an improvement, but its worth a double check! Running now.

AlexandreSinger:
@miloliebster FYI, this PR modifies part of your code. There was a boolean flag being set if the last intra-lb route was successful; but it was ignoring the case when the last intra-lb route was not successful and it was the last molecule to be tried. This may actually improve your results since, on cases when memoization was used, it was likely performing the intra-lb route twice. It was conflicting with this change (and a future one that I am working on), which is why I needed to fix it.

miloliebster:
@AlexandreSinger The changes make sense with what you've added.

One thing that I'm a bit worried about with how these changes might interact: PathFinder can give different outcomes, or even fail when a solution exists, if seeded differently. Is there not a scenario here where an initial cluster reaches a legal solution iteratively with hot-starting, but then a future matching cluster fails to produce a legal routing for the final state since the PST has let it skip all routing up to this point? It would be routing the entire cluster from scratch rather than starting from an existing state, so would possibly give a different solution.

[edit] reading through the comments, you've mentioned this scenario already. Interested to see how the approaches compare across more architectures and benchmarks if we ultimately deem that they are fully incompatible.

AlexandreSinger:
> One thing that I'm a bit worried about with how these changes might interact: PathFinder can give different outcomes, or even fail when a solution exists, if seeded differently. Is there not a scenario here where an initial cluster reaches a legal solution iteratively with hot-starting, but then a future matching cluster fails to produce a legal routing for the final state since the PST has let it skip all routing up to this point? It would be routing the entire cluster from scratch rather than starting from an existing state, so would possible give a different solution.

@miloliebster I share this concern 100%. It absolutely occurs as I have observed this in my testing. I am not sure if these options are completely mutually exclusive yet (I have only tested this on a very simple architecture). I do know that using the memoization causes a different solution to occur; but I am not sure if there may be a hidden crash or not.

My hypothesis is the following:
1. Incremental intra-lb routing is used (with hot starting).
2. The solution is memoized using your approach.
3. The memoization kicks in.
4. The same solution comes back, memoization matches the solution, and a final route is performed. If the routing succeeds, the solution will be different from what is cached but the quality will be similar (which is why using both produces a different solution). If the routing fails, I am not sure what happens here, my hope is that full intra-lb routing kicks back in; but its possible that it hits a hard crash... There is a check which crashes if a cluster fails full-intra-lb routing since it is assumed to be possible at each step.

AlexandreSinger:
To provide more context on the boolean flag. The issue I am hitting in another branch is that I am building multiple clusters at the same time; but that boolean flag is shared amongst all clusters, which was causing issues. We are moving to create multiple clusters in parallel since the current greedy packer is not as good as we would like.

doc/src/vpr/command_line_usage.rst
```diff
@@ -778,6 +778,17 @@ For people not working on CAD, you can probably leave all the options to their d
...
     **Default:** ``off``
 
+.. option:: --cluster_router_hot_start {on | off}
+
+    Enables hot-starting of the intra-cluster router during packing.
+
+    When enabled, each call to the intra-cluster router seeds unchanged nets from the previous successful route before running pathfinder.
+    Nets whose terminals are unchanged and whose route trees are still valid under the current mode assignments are committed upfront, allowing pathfinder to skip them on its first iteration.
+    This can reduce router runtime when many candidate molecules are tried and rejected: after a failed molecule is removed, the cluster returns to a known-good state without re-routing nets that did not change.
+    Enabling this option does not affect circuit quality metrics like routed wirelength or critical path delay.
```
haydar-c: As i understand, the packing solution can be slightly different even though the wirelength or critical path delay is not affected significantly. I don't know if we need to mention it here but just noting it down to be cautious.

miloliebster:
@AlexandreSinger With the current control flow iirc this scenario would occur at the end of detailed packing where the assumption is there must be a legal solution by now, so I think we're looking at a hard crash. I think to converge in this (relatively rare, but certainly possible) case you would need to fall back to _another_ detailed attempt with hot-starting on and memoization off.

The most elegant fix that would make these play well together would be having flyweighted routing solutions that get stored in the PST so that the final routing for matching clusters isn't needed, but that is not an insignificant amount of work. On the other hand, what I want to explore over this term is some ideas I have to try to formulate cluster legality/routing as an allocation problem rather than invoking PathFinder, which I think would have a similar iterative approach to your hot-starting but use far cheaper and more cache friendly operations. I'd actually be interested in syncing up to discuss or potentially collaborating on that if you're open to it.

miloliebster:
> To provide more context on the boolean flag. The issue I am hitting in another branch is that I am building multiple clusters at the same time; but that boolean flag is shared amongst all clusters, which was causing issues. We are moving to create multiple clusters in parallel since the current greedy packer is not as good as we would like.

I have a branch from a while back where I was working on allowing multiple cluster states to exist simultaneously in the PST so that it would be able to be used with one of the AP algorithms that requires that. I can move that change up in my queue if you think that would be helpful?

vpr/src/base/read_options.cpp
```diff
@@ -2305,6 +2305,21 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
...
+              "\n"
+              "When enabled, the router seeds each new routing attempt with the\n"
+              "route trees from the previous successful route. Nets whose terminals\n"
+              "are unchanged and whose route trees are still valid under the current\n"
+              "mode assignments are committed before pathfinder begins, allowing them\n"
+              "to be skipped in the first iteration. This can reduce router runtime\n"
+              "when many molecules are tried and rejected, since the cluster returns\n"
+              "to a known-good state without re-routing nets that did not change.\n"
+              "\n"
+              "Enabling this option does not affect circuit quality metrics.\n")
```
haydar-c: Same comment about QoR effect here.

haydar-c (commented):
LGTM @AlexandreSinger, I just added a small comment about the documentation. You can correct me if I am wrong there.

AlexandreSinger:
> The most elegant fix that would make these play well together would be having flyweighted routing solutions that get stored in the PST so that the final routing for matching clusters isn't needed, but that is not an insignificant amount of work.

A simpler idea would be to make this a legalization stage in the cluster legalizer. Currently we have two:

1. SKIP_INTRA_LB_ROUTING (terrible name, it just means we only do pin counting)
2. FULL_LEGALIZATION (this does routing at each added molecules and gaurentees that the solution is legal).

I am thinking of doing the following:

1. SKIP_INTRA_LB_ROUTING (we need to change this name...)
2. MEMOIZATION (basically what we have now for memoization + full-legalization)
3. FULL_LEGALIZATION (basically what we had before but with no memoization).

In truth, that last stage will not be hit, but it just acts as a fallback if order dependence causes an issue. This would fix the case we have now, and potentially any future ones. I do not think its needed for this PR, but if the two need to work together, this would be an approach.

> On the other hand, what I want to explore over this term is some ideas I have to try to formulate cluster legality/routing as an allocation problem rather than invoking PathFinder, which I think would have a similar iterative approach to your hot-starting but use far cheaper and more cache friendly operations. I'd actually be interested in syncing up to discuss or potentially collaborating on that if you're open to it.

I would be happy to discuss! I hate the cluster router in its current form; I am not sure how we would avoid the pathfinder though, but it has been something on my radar for some time to elide the pathfinder (by proving that a solution exists somehow) and then let Amin's flat router do the actual pathfinder which is more intuitive.

> I have a branch from a while back where I was working on allowing multiple cluster states to exist simultaneously in the PST so that it would be able to be used with one of the AP algorithms that requires that. I can move that change up in my queue if you think that would be helpful?

We can discuss. I have spent a very long time trying clean up the packer to allow multiple clusters to be built at the same time, it would be nice to maintain that feature if we can. But it all depends on the tradeoffs. Currently, we do not have a solution that runs in parallel.

doc/src/vpr/command_line_usage.rst
AlexandreSinger: Thats a very good point! Thanks Haydar! I will update.

AlexandreSinger:
Results on Titan Quick QOR:

  | baseline.txt | hot_start_off.txt | hot_start_on.txt
-- | -- | -- | --
vtr_flow_elapsed_time | 1.00 | **1.01** | **1.03**
num_LAB | 1.00 | 1.00 | 1.00
num_DSP | 1.00 | 1.00 | 1.00
num_M9K | 1.00 | 1.00 | 1.00
num_M144K | 1.00 | 1.00 | 1.00
max_vpr_mem | 1.00 | 1.00 | 1.00
num_pre_packed_blocks | 1.00 | 1.00 | 1.00
num_post_packed_blocks | 1.00 | 1.00 | 1.00
device_grid_tiles | 1.00 | 1.00 | 1.00
pack_time | 1.00 | **1.02** | **1.05**
placed_wirelength_est | 1.00 | 1.00 | 1.00
place_time | 1.00 | 1.00 | **1.04**
placed_CPD_est | 1.00 | 1.00 | 1.00
routed_wirelength | 1.00 | 1.00 | 1.00
critical_path_delay | 1.00 | 1.00 | 1.00
geomean_nonvirtual_intradomain_critical_path_delay | 1.00 | 1.00 | 1.00
crit_path_route_time | 1.00 | **1.02** | **1.03**

The baseline here is Master without any changes, hot_start_off is the default of this branch, and hot_start_on is turning on hot starting for the titan circuits.

There appears to be a 2% loss in packer performance on Titan. The cause may be because whenever a cluster finishes, we are checking the pins and nets in the intra-lb route to ensure that it is still valid. The original code just assumed that the last route must have been valid between calls (between adding and removing molecules). We could speed this up easily by adding a fast path to the code (which returns true if FULL_LEGALIZATION is on), since that is what it was doing before. But this was actually an unsafe assumption since someone may want to change the legalization strategy between adding molucules. For example, suppose someone used FULL_LEGALIZATION for a while and then switch to only doing the pin counting legalization, this may still return true!

Turning hot-starting on appeared to have no affect on result quality, meaning the solution has not changed. What this implies to me is that the clusters may be small enough that it ended up not mattering whether hot starting is used or not. This is expected. I made the hot-starter specifically for larger clusters.

@vaughnbetz What do you think about these results? I believe that the 2% increase in run time may be worth it to have a safer check at the end of clustering (having a true verification that the route is legal before continuing); but I concede to you on this. I could poke around and find some data structures that give us 2% more performance :)

miloliebster:
> I am thinking of doing the following:
> 
> 1. SKIP_INTRA_LB_ROUTING (we need to change this name...)
> 2. MEMOIZATION (basically what we have now for memoization + full-legalization)
> 3. FULL_LEGALIZATION (basically what we had before but with no memoization).

I agree that this would be a sufficient fallback to handle the case we're concerned about. It's a bit clunky, but from my observations dealing with similar quirks for the PST, the cases where you have these pathfinder discrepancies are like a couple clusters on _some_ benchmarks on _some_ architectures, so it seems reasonable as a fallback.

> In truth, that last stage will not be hit, but it just acts as a fallback if order dependence causes an issue. This would fix the case we have now, and potentially any future ones. I do not think its needed for this PR, but if the two need to work together, this would be an approach.

I'm a bit confused by this. The cases where the third stage will be hit are very infrequent (and architecture dependant) but there will be cases where it would get hit. Is your plan for this PR to make these mutually exclusive options?

> I would be happy to discuss!

We can move that discussion out of this PR since it's a bit off topic. I'll ping you on LinkedIn and we can set something up!

vaughnbetz:
The 2% year increase in pack time when the hot start is off looks like machine load, since nice routing also slowed down 2%. Seems like maybe 2% increase when it is turned on.VaughnOn May 24, 2026, at 10:04 AM, Milo Liebster ***@***.***> wrote:﻿miloliebster left a comment (verilog-to-routing/vtr-verilog-to-routing#3443)

I am thinking of doing the following:

SKIP_INTRA_LB_ROUTING (we need to change this name...)
MEMOIZATION (basically what we have now for memoization + full-legalization)
FULL_LEGALIZATION (basically what we had before but with no memoization).


I agree that this would be a sufficient fallback to handle the case we're concerned about. It's a bit clunky, but from my observations dealing with similar quirks for the PST, the cases where you have these pathfinder discrepancies are like a couple clusters on some benchmarks on some architectures, so it seems reasonable as a fallback.

In truth, that last stage will not be hit, but it just acts as a fallback if order dependence causes an issue. This would fix the case we have now, and potentially any future ones. I do not think its needed for this PR, but if the two need to work together, this would be an approach.

I'm a bit confused by this. The cases where the third stage will be hit are very infrequent (and architecture dependant) but there will be cases where it would get hit. Is your plan for this PR to make these mutually exclusive options?

I would be happy to discuss!

We can move that discussion out of this PR since it's a bit off topic. I'll ping you on LinkedIn and we can set something up!

Issue #3432, opened 2026-02-19
Support native Windows build on CI


tangxifan:
#### Proposed Behaviour
Currently, VTR project supports Linux platform w.r.t code compilation and regression tests.
The Linux support has been continuously validated on Github actions.
In fact, Windows platform is a popular choice for FPGA developers to use CAD tools.
CAD tools from AMD, Altera, Lattice and other FPGA vendors all provide a windows build.
To popularize VTR project and advocates VTR's integration to commercial FPGA vendors, supporting Windows platform is necessary.

#### Current Behaviour

Currently, VTR project only supports Linux platform w.r.t code compilation and regression tests.

#### Context

Thanks to Github actions (windows machine available) and the advancements in eco-system on Windows platform (vcpkg, powershell etc.), cross-platform for VTR project is much easier than before. 

Note that the Windows supports targets the code compilation by MSVC compilers, rather than GCC on Windows (which are commonly seen in Msys/MinGW)
The CMake has good supports on MSVC to generate nmake-compatible Makefile.
Most of the challenges are seen in installing the dependency on Windows (which are commonly easy in Linux).
Fortunately, after looking into the dependencies required by VTR, most of them have prebuilt binaries on Windows already, including

- pkgconf (by vcpkg)
- zlib (by vcpkg)
- flex and bison (by chocolatey)
- tcl (by official Github release: https://github.com/teclabat/tcltk-binaries/releases)
- swig (by chocolatey)

#### Current status

I have started a try on the OpenFPGA project: https://github.com/lnis-uofu/OpenFPGA/pull/2386
where I have developed the CI workflow to generate Makefile for MSVC. 
Locally, I have tried ``nmake`` to build but run into errors due to some VTR codes are only for Linux. See details [here](https://github.com/lnis-uofu/OpenFPGA/issues/2382#issuecomment-3907757551)

#### Remaining challenges

- Yosys has supported native windows build through Clang. 
As Yosys is not based on CMake system, it relies a number of scripts to convert the Makefile to a MSVC project file. See details [here](https://github.com/YosysHQ/yosys/blob/main/misc/create_vcxsrc.sh)
As Yosys is treated as a custom target in current CMake build system, it is better to follow its official windows support by integrating its CI flow to the CMakeList.txt (consider to encapsulate as a module)

[...]

PR #3476, opened 2026-04-06
Modifications for windows support with OpenFPGA

oscarcheng105:
#### Related Issue
Change is based on the issue : #3432
As well as this issue from OpenFPGA: https://github.com/lnis-uofu/OpenFPGA/issues/2382

#### Motivation and Context

To ensure vtr builds correctly on windows (MSVC).

#### Types of changes

#### Checklist:


libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -23,6 +23,10 @@
 #include <string>
 #include <vector>
 
+#ifdef _MSC_VER
```
vaughnbetz: Comment why -- no prefetch on windows (and was only a performance optimization so OK to turn off).

libs/librrgraph/src/base/rr_node_types.h
```diff
@@ -75,7 +75,8 @@ inline e_rr_type get_rr_type(const std::string& type_name) {
     } else if (type_name == "MUX") {
         return e_rr_type::MUX;
     } else {
-        VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Invalid RR type name: %s\n", type_name.c_str());
+        //VPR_FATAL_ERROR(VPR_ERROR_ROUTE, "Invalid RR type name: %s\n", type_name.c_str());
```
vaughnbetz: Why are you removing the VPR_FATAL_ERROR macro (which is a wrapper around vtr_throw that is a bit easier to read)?  I think the prior code was more idiomatic in vpr, and easier to read.

libs/librrgraph/src/io/gen/rr_graph_uxsdcxx.h
```diff
@@ -203,10 +205,20 @@ inline void write_rr_graph_xml(T &in, Context &context, std::ostream &os){
 }
 
 
+#ifdef _MSC_VER
```
vaughnbetz, line 206: I'd add a comment // Windows to make it clear what the #ifdef is for.

libs/libvtrcapnproto/mmap_file.cpp
```diff
@@ -7,7 +7,12 @@
 #ifndef _WIN32
     #include <sys/mman.h>
 #endif
+
+#ifdef _WIN32
```
vaughnbetz: Not sure why we're checking _WIN32 sometime and _MSC_VER others.  If there's a reason to check a specific one (WIN32 but not WIN64?) comment it; if not I suggest picking on check for windows.

libs/libvtrutil/src/vpr_error.h
```diff
@@ -104,11 +104,17 @@ void vpr_throw_opt(enum e_vpr_error type, const char* psz_func_pretty_name, cons
...
-#define VPR_FATAL_ERROR(...)    \
-    do {                        \
-        VPR_THROW(__VA_ARGS__); \
+
+#define VPR_FATAL_ERROR(type, ...)                        \
+    do {                                                  \
+        vpr_throw(type, __FILE__, __LINE__, __VA_ARGS__); \
     } while (false)
 
+//#define VPR_FATAL_ERROR(...)    \
```
vaughnbetz: Why take this out?  Seems unnecessary for windows (we should be able to make the macro work), and we use VPR_FATAL_ERROR a lot.

libs/libvtrutil/src/vtr_path.cpp
```diff
@@ -4,7 +4,11 @@
 
 //TODO: currently this file assumes unix-like
```
vaughnbetz: Should remove this TODO if we support windows now (and check if there are any file issues first).

libs/libvtrutil/src/vtr_util.h
```diff
@@ -7,6 +7,11 @@
 #include <cstdarg>
 #include <array>
 
+#ifdef _MSC_VER
+#define __attribute__(x)
+#define __builtin_prefetch(...) ((void)0)
```
vaughnbetz: Comment why it's OK to take out builtin_prefetch (just an optimization hint).
Comment why we're using _MSC_VER here (not WIN32)

vpr/src/analytical_place/ap_argparse_utils.cpp
```diff
@@ -81,8 +81,8 @@ parse_key_val_arg(const std::string& arg, unsigned expected_num_vals_per_key) {
     if (colon_count != 1) {
         VTR_LOG_ERROR("Invalid argument string: %s\n",
                       arg.c_str());
-        VPR_FATAL_ERROR(VPR_ERROR_PACK,
-                        "Error when parsing argument string");
+        vpr_throw(VPR_ERROR_PACK, __FILE__, __LINE__,
```
vaughnbetz: Prefer to leave the VPR_FATAL_ERROR macro in.

vpr/src/analytical_place/ap_argparse_utils.cpp
```diff
@@ -100,8 +100,8 @@ parse_key_val_arg(const std::string& arg, unsigned expected_num_vals_per_key) {
     if (comma_count != expected_num_vals_per_key - 1) {
         VTR_LOG_ERROR("Invalid argument string (too many commas): %s\n",
                       arg.c_str());
-        VPR_FATAL_ERROR(VPR_ERROR_PACK,
-                        "Error when parsing argument string");
+        vpr_throw(VPR_ERROR_PACK, __FILE__, __LINE__,
```
vaughnbetz: VPR_FATAL_ERROR

vpr/src/base/netlist.tpp
```diff
@@ -627,23 +629,26 @@ bool Netlist<BlockId, PortId, PinId, NetId>::verify_lookups() const {
         auto port_id = pin_port(pin_id);
         auto bit = pin_port_bit(pin_id);
         if (find_pin(port_id, bit) != pin_id) {
-            VPR_FATAL_ERROR(VPR_ERROR_NETLIST, "Pin lookup by name mismatch");
+            // VPR_FATAL_ERROR(VPR_ERROR_NETLIST, "Pin lookup by name mismatch");
+            vpr_throw(VPR_ERROR_NETLIST, __FILE__, __LINE__, "Pin lookup by name mismatch");
```
vaughnbetz: Prefer VPR_FATAL_ERROR () ...

vaughnbetz (changes requested):

libs/librrgraph/src/io/gen/rr_graph_uxsdcxx.h
```diff
@@ -9,6 +9,8 @@
  * md5sum of input file: 45774433f1b54981c349fecadf578b11
  */
 
+// NOTES CHECK WITH VTR ON GENERATION 4/3/2026
+
```
AmirhosseinPoolad: This file is auto generated from a schema. The fix should be done in the uxsdcxx project. See the comment above this line.

libs/libvtrutil/src/vtr_path.cpp
```diff
@@ -65,7 +69,13 @@ std::string getcwd() {
...
     char buf[BUF_SIZE];
 
+#ifdef _WIN32 // Windows
+    // Windows uses _getcwd() instead of POSIX getcwd().
+    if (::_getcwd(buf, static_cast<int>(BUF_SIZE))) {
+#else
+    // POSIX getcwd() returns the current working directory.
     if (::getcwd(buf, BUF_SIZE)) {
+#endif
         return std::string(buf);
```
AmirhosseinPoolad: We shouldn't be dealing with low level OS routines. lets use std::filesystem::current_path() instead.

AmirhosseinPoolad: You can remove both unistd.h and direct.h after that.

libs/libvtrutil/src/vtr_strong_id_range.h
```diff
@@ -3,6 +3,13 @@
 #include <iterator>
 #include "vtr_assert.h"
 
+#ifdef _WIN32 // Windows
+// Windows does not define the POSIX type ssize_t.
+// Use SSIZE_T from BaseTsd.h as a compatible replacement.
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
```
AmirhosseinPoolad: Instead of doing this, change all ssize_ts to ptrdiff_t. It's standard and cross platform.

libs/libvtrcapnproto/serdes_utils.cpp
```diff
@@ -2,7 +2,14 @@
...
 #include "capnp/serialize.h"
 #include <fcntl.h>
+
+#ifdef _WIN32 // Windows
+// Windows does not provide POSIX unistd.h; use io.h instead for low-level I/O.
+#include <io.h>
+#else
+// POSIX systems provide unistd.h for low-level file operations.
 #include <unistd.h>
+#endif
```
AmirhosseinPoolad: fcnt.h and unistd.h are not needed for linux builds and we were unnecessarily including them here. Please remove them and double check if io.h is actually needed for windows builds.

libs/libvtrcapnproto/mmap_file.cpp
```diff
@@ -4,10 +4,17 @@
...
+
+#ifdef _WIN32 // Windows
+// Windows does not provide POSIX mmap() or unistd.h.
+// Use Windows-specific headers instead. Note that actual file mapping is
+// handled via the KJ filesystem abstraction.
+#include <io.h>
+#else
+// POSIX systems provide mmap() and unistd.h for file access and mapping.
+#include <sys/mman.h>
 #include <unistd.h>
```
AmirhosseinPoolad: unistd.h, mman.h, fcntl.h, and sys/stat.h are not needed here for linux builds. Please remove them and check if io.h is actually needed for windows builds.

utils/fasm/src/fasm.cpp
```diff
@@ -18,6 +18,16 @@
...
+#ifdef _MSC_VER // MSVC
+// MSVC does not provide GCC-style __builtin_popcount.
+// Map it to the equivalent MSVC intrinsic __popcnt from <intrin.h>.
+#include <intrin.h>
+
+inline int __builtin_popcount(unsigned int x) {
+    return __popcnt(x);
+}
+#endif
+
```
AmirhosseinPoolad: Since we're on C++20 now, we can use std::popcount or even std::has_single_bit instead.

AmirhosseinPoolad: This line:
> if(__builtin_popcount(lut_parts.size()) != 1) {

Can be changed to this:

> if(std::popcount(lut_parts.size()) != 1) {

or

> if(!std::has_single_bit(lut_parts.size())) {

Don't have a strong preference for either one but doing this would remove the need for ifdefs.

utils/fasm/src/lut.cpp
```diff
@@ -6,6 +6,16 @@
...
+#ifdef _MSC_VER // MSVC
+// MSVC does not provide GCC builtins like __builtin_popcount and __builtin_ctz.
+// Map them to equivalent MSVC intrinsics from <intrin.h>.
+#include <intrin.h>
+
+#define __builtin_popcount(x) __popcnt(x)
+// _tzcnt_u32 counts trailing zeros (equivalent to __builtin_ctz for 32-bit values)
+#define __builtin_ctz(x) _tzcnt_u32(x)
+#endif
+
```
AmirhosseinPoolad: Same as above, change __builtin_popcount to std::popcount and __builtin_ctz to std::countr_zero. Removes the need for ifdefs.

vpr/src/base/gen/vpr_constraints_uxsdcxx.h
```diff
@@ -119,10 +119,25 @@ inline void write_vpr_constraints_xml(T& in, Context& context, std::ostream& os)
     in.finish_write();
 }
 
+#ifdef _MSC_VER // MSVC
+// MSVC uses __declspec(align(...)) instead of GCC-style
+// __attribute__((aligned(...))) for aligned type declarations.
+typedef __declspec(align(1)) const uint32_t triehash_uu32;
+typedef __declspec(align(1)) const uint64_t triehash_uu64;
+#else
```
AmirhosseinPoolad: This is also auto generated from a schema.

AmirhosseinPoolad (changes requested):
Thanks for the PR Oscar. Had some comments about how to avoid doing some of the ifdefs. There's also an issue with the auto generated files.

I also found that some of the OS includes in libvtrcapnproto weren't needed for linux builds. Please check if windows builds work without those includes and remove them if so.

libs/librrgraph/src/base/rr_graph_storage.h
```diff
@@ -23,6 +23,14 @@
...
 #include <vector>
 
+#ifdef _MSC_VER // MSVC
+// MSVC does not support GCC __builtin_prefetch.
+// Without this guard, GCC-specific constructs can cause syntax errors
+// (e.g., in multi_queue_d_ary_heap.tpp). Since prefetch is only a
+// performance hint, it is safe to make it a no-op and ignore attributes.
+#define __builtin_prefetch(...) ((void)0)
+#endif
+
```
AmirhosseinPoolad, line 33: This is defined in the wrong place. We have a macro VTR_PREFETCH defined in libs/libvtrutil/src/vtr_memory.h. Do an ifdef for that macro instead.

libs/libvtrutil/src/vtr_util.h
```diff
@@ -7,6 +7,15 @@
 #include <cstdarg>
 #include <array>
 
+#ifdef _MSC_VER // MSVC
+// MSVC does not support GCC __attribute__ or __builtin_prefetch.
+// Without this guard, GCC-specific constructs can cause syntax errors
+// (e.g., in multi_queue_d_ary_heap.tpp). Since prefetch is only a
+// performance hint, it is safe to make it a no-op and ignore attributes.
+#define __attribute__(x)
+#define __builtin_prefetch(...) ((void)0)
```
AmirhosseinPoolad, line 16: #define __builtin_prefetch can be removed from here when you address the previous comment. I think the \_\_attribute\_\_ fix only works by accident, since vtr_util.h is included in all files that actually use \_\_attribute\_\_. I don't think that's  maintainable and we would see MSVC break a lot with future changes.

Check place_delay_model.h. In there we define a custom macro "ALWAYS_INLINE" with relevant ifdefs to make it work with GCC, Clang and MSVC. I think if you did something similar in multi_queue_d_ary_heap.tpp it would fix the issue in a more maintainable way. multi_queue_d_ary_heap uses two attributes, 'noinline' and 'aligned' for both performance and correctness reasons.

libs/libvtrcapnproto/CMakeLists.txt
```diff
@@ -44,8 +44,9 @@ if (VPR_ENABLE_INTERCHANGE)
         OUTPUT ${JAVA_SCHEMA}
         COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/schema/capnp/
         COMMAND ${WGET}
+            -L
             https://raw.githubusercontent.com/capnproto/capnproto-java/master/compiler/src/main/schema/capnp/java.capnp
-            -O ${JAVA_SCHEMA}
```
tangxifan, line 48: Is this the root of cause on JAVA failures in CI?

libs/libvtrcapnproto/mmap_file.cpp
```diff
@@ -1,14 +1,6 @@
 #include "mmap_file.h"
 #include "vtr_error.h"
 #include "vtr_util.h"
-
-#include <fcntl.h>
```
tangxifan, line 5: Can you check why these codes are removed?

libs/libvtrcapnproto/serdes_utils.cpp
```diff
@@ -2,8 +2,6 @@
 
 #include "capnp/serialize.h"
 #include <fcntl.h>
-#include <unistd.h>
```
tangxifan, line 5: Can you double check why this is removed?

libs/libvqm/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ add_library(libvqm STATIC
              ${LIB_SOURCES}
              ${FLEX_VqmLexer_OUTPUTS} 
              ${BISON_VqmParser_OUTPUT_SOURCE})
+target_compile_definitions(libvqm PRIVATE YY_NO_UNISTD_H)
+target_compile_definitions(libvqm PRIVATE isatty=_isatty)
```
tangxifan: You may branch here for WIN32. This causes VQM build errors on Linux.

libs/libvtrcapnproto/mmap_file.cpp
AmirhosseinPoolad: For this and serdes_utils, that was my review comment. Those includes weren't actually needed and were included by mistake. Removing them makes the code cross-platform by default and removes the need for ifdefs.

libs/libvtrutil/src/vtr_util.h
```diff
@@ -7,6 +7,15 @@
...
+#endif
```
tangxifan: @oscarcheng105 Can you check if these lines are still required? If the already exists in the vtr_util.h, they can be removed.

tangxifan (commented):
@oscarcheng105 Can you wrap up the PR? CI is green.

libs/librrgraph/src/io/gen/rr_graph_uxsdcxx.h
oscarcheng105, line 1: This fix is temprorary for windows build. Requires a fix on the uxsdcxx project.

vpr/src/base/gen/vpr_constraints_uxsdcxx.h
oscarcheng105, line 1: This fix is temprorary for windows build. Requires a fix on the uxsdcxx project.

oscarcheng105: need to make a PR on https://github.com/SymbiFlow/uxsdcxx

abc/CMakeLists.txt
```diff
@@ -78,10 +78,11 @@ function(extract_var SEPARATOR DEST_VARIABLE MAKE_OUTPUT)
   set(${DEST_VARIABLE} ${TMP} PARENT_SCOPE)
 endfunction()
 
-extract_var(SEPARATOR_SRC ABC_SRC ${MAKE_OUTPUT})
-extract_var(SEPARATOR_LIBS ABC_LIBS ${MAKE_OUTPUT})
-extract_var(SEPARATOR_CFLAGS ABC_CFLAGS ${MAKE_OUTPUT})
-extract_var(SEPARATOR_CXXFLAGS ABC_CXXFLAGS ${MAKE_OUTPUT})
+message(STATUS "MAKE_OUTPUT=${MAKE_OUTPUT}")
```
AlexandreSinger: abc is not a part of the VTR project. It is actually a sub-tree. As such, we want to avoid making any modifications to these files unless we have to.

Why were these changes necessary? Even on the abc side they do not need the quotation marks:
https://github.com/berkeley-abc/abc/blob/8762d6c66727455f82e7581097b8c81b72ac99e2/CMakeLists.txt#L87-L90

The message being produced here is also not necessary, and it should be removed.

libs/libvqm/vqm_parser.l
```diff
@@ -22,6 +22,15 @@
...
 #include "vqm_common.h"
 
+#ifdef _WIN32 // Windows
+// Windows does not provide POSIX isatty() in <unistd.h>.
+// Map isatty to the Windows equivalent _isatty() from <io.h>.
+#include <io.h>
+#define isatty _isatty
+#else
+#include <unistd.h>
+#endif
```
AlexandreSinger: I do not like this solution. It is very dangerous to make a define replace like this. There is a well defined way of fixing this exact issue in Lex by using `%option never-interactive`.

See the following examples of when we did this exact thing before:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/263b990fb4cd0b690cf6f3567b9aa5f54693c360/libs/EXTERNAL/libblifparse/src/blif_lexer.l#L32-L33

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/263b990fb4cd0b690cf6f3567b9aa5f54693c360/libs/EXTERNAL/libtatum/libtatumparse/tatumparse/tatumparse.l#L28-L29

AlexandreSinger: The fix you made would only be necessary if this was interactive; but I do not think VQM2Blif is ever interactive.

libs/libvtrcapnproto/CMakeLists.txt
```diff
@@ -40,13 +40,23 @@ if (VPR_ENABLE_INTERCHANGE)
...
+    if(WIN32)
+        add_custom_command(
+            OUTPUT ${JAVA_SCHEMA}
+            COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/schema/capnp/
+            COMMAND ${WGET}
+                https://raw.githubusercontent.com/capnproto/capnproto-java/master/compiler/src/main/schema/capnp/java.capnp
+                -o ${JAVA_SCHEMA}
+        )
+    else()
+        add_custom_command(
```
AlexandreSinger: What is the difference between the above custom command and below? The only difference I can see is `-o` instead of `-O`, but from my understanding the lower-case `-o` is correct. Have you tried just replacing the `-O` with the lower-case version and seeing if it works?

If it does not work, we should investigate why.

libs/libvtrutil/src/vpr_error.h
```diff
@@ -104,10 +104,21 @@ void vpr_throw_opt(enum e_vpr_error type, const char* psz_func_pretty_name, cons
...
+#define VPR_FATAL_ERROR(type, ...)                        \
+    do {                                                  \
+        vpr_throw(type, __FILE__, __LINE__, __VA_ARGS__); \
+    } while (false)
+#else
 #define VPR_FATAL_ERROR(...)    \
     do {                        \
         VPR_THROW(__VA_ARGS__); \
     } while (false)
+#endif
```
AlexandreSinger: I think it would just be cleaner to use the version of this that you defined so both MSVC and Linux can use the same define. I think nested variadic can be confusing in general.

I recommend just making it:
```cpp
#define VPR_FATAL_ERROR(type, ...)                        \
    do {                                                  \
        vpr_throw(type, __FILE__, __LINE__, __VA_ARGS__); \
    } while (false)
```

And then update the comment above this define.

AlexandreSinger (changes requested):
I quickly read through this PR and found some issues that should be resolved.

libs/libvtrcapnproto/CMakeLists.txt
oscarcheng105: Checked `-o` build with linux.
Using the `-o` flag causes errors in the linux build. The original CMakeLists.txt uses `-O` for linux, need to investigate why

oscarcheng105:
A PR has been created in the uxsdcxx repo: https://github.com/SymbiFlow/uxsdcxx/pull/55
Resolves issues regarding GCC `__attribute__` incompatibility with MSVC guards in generated files such as `vpr_constraints_uxsdcxx.h` and `rr_graph_uxsdcxx.h`.

tangxifan:
@vaughnbetz @AlexandreSinger @AmirhosseinPoolad This PR is ready for your review again.

- We have create a PR on the uxsdcxx repo and waiting for it to be merged. After that, the generator for rr_graph should include the MSVC guard, which is manually added in this PR.
- The libsdcparse modification has been merged. @AlexandreSinger Please confirm if the submodule is updated to the latest or not. If not, we can do it in this PR.

If the PR on the uxsdcxx repo takes too long to be merged, I suggest that we merge this PR as is, and start plan B as discussed in the weekly meeting.

AlexandreSinger:
>The libsdcparse modification has been merged. @AlexandreSinger Please confirm if the submodule is updated to the latest or not. If not, we can do it in this PR.

@tangxifan The current version of LibSDCParse in VTR is behind the changes that were merged in yesterday. We should update it in this PR using a command like `git submodule update --remote libs/EXTERNAL/libsdcparse`

oscarcheng105:
Screenshot 2026-04-26 095430

@AlexandreSinger an error is encountered when VTR tries to build with the latest version of libsdcparse. I did a quick fix on `vpr/src/timing/read_sdc.cpp`.

tangxifan:
@vaughnbetz Our PR on the github repo uxsdcxx has been merged to main branch. So the modification on the generated rr_graph parser is not an issue anymore. 

@AlexandreSinger The libsdcparse has been updated to the latest in this PR. @oscarcheng105 has fixed the mismatches on APIs. Please kindly check and comment.

@vaughnbetz If nothing significant shows up, I suggest to merge this PR as all the major concerns have been resolved.

tangxifan (approved):

AmirhosseinPoolad (approved):
I think the PR is in good shape now. Please use the updated uxsd to generate the parsers, just to make sure the uxsd changes work correctly. Could be merged afterwards.

AlexandreSinger (approved):
LGTM

Thank you for updating the LibSDCParse APIs. I did not realize that the changes I made to the SetIODelays required VTR changes. The changes you made look correct.

libs/libvtrcapnproto/CMakeLists.txt
vaughnbetz: Could this be made more concise with a variable for -o / -O, rather than duplicating several lines like this to have the -o / -O differentiation?

vaughnbetz (approved):
LGTM. I have one suggestion to check out to see if we can simplify make the cmake commands a bit more common.

Also, please run one benchmark set on Linux to confirm no performance degradation before merging (e.g. VTR or Titan).  I don't expect any, but it is good to be paranoid.

It would also be good to immediately check the performance on Windows vs. Linux for a similar machine, on that same benchmark set (VTR or Titan). Please add those perf results.

oscarcheng105:
@vaughnbetz @tangxifan
The current build works on Windows (MSVC) and I’ve verified that `vpr.exe` runs correctly when invoked directly.

Currently I couldn't get `run_vtr_flow.py` to run properly on Windows. The Python-based flow (`run_vtr_flow.py` / regression scripts) fails to launch VPR on Windows with a `WinError 193`. This appears to be due to path handling in the Python wrappers (e.g., passing relative paths like `architecture.name` and `circuit.name` while executing in a different working directory). This breaks on Windows, and similar issues may exist in other wrappers. 

Also, ABC and Parmys/Yosys are disabled with the current build (`WITH_ABC=OFF`, `WITH_PARMYS=OFF`), so regression tasks like `vtr_reg_basic` (which includes `hdl_include_yosys`) cannot be fully executed.

tangxifan:
@vaughnbetz To get the regression tests QoR, the python scripts require an upgrade to support Windows platform. It will requires additional effort on VTR. I suggest to merge this PR as is and create a follow-up PR on the regression test upgrade. What do you think?

oscarcheng105:
Screenshot 2026-04-28 023034

 `run_reg_test.py vtr_reg_basic` result from linux(wsl) build.

vaughnbetz:
OK, looks good to me.  Just to be paranoid, please run one benchmark suite on Linux and confirm no overall compile time or QoR degradation with this change.  (Linux only, no windows, to avoid any problems).

PR #3485, opened 2026-04-09
[CI] Separated out the parallel regression tests

AlexandreSinger:
The parallel regression test may cause random failures due to using too many cores and running at the same time. To fix this, I have isolated these tests to their own test suites and fixed their number of cores to
2. Currently the CI runners have 4 cores; so this should be sufficient, but we can reduce this to 1 if needed.

Resolves #3483


AmirhosseinPoolad (commented):
Looks good.

vaughnbetz (approved):

Issue #3487, opened 2026-04-14
[flow] Unused `YYY` placeholder in synthesis flow during script string replacement


WZ-Tong:
The `YYY` placeholder in parmys.py is never substituted during the VTR flow. It remains in the generated synthesis command and produces stray tokens in the yosys invocation.

#### Expected Behaviour

During replacement, placeholder `YYY` is not replaced, and the content from comment `other args: [YYY]` is not passed to the function `init_script_file` (other args are even not available in callers)
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/vtr_flow/scripts/python_libs/vtr/parmys/parmys.py#L66-L75

#### Current Behaviour

As a template synthesis script, placeholder `YYY` is here
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/vtr_flow/misc/yosys/synthesis.tcl#L25
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/vtr_flow/misc/yosys/synthesis.tcl#L86

#### Possible Solution

Remove the placeholder in the template synthesis script

WZ-Tong:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/parmys/verify_parmys.sh#L1044

Seems like this placeholder is used for test and caused the check flow failed. This test reads args from `{test}/task.conf` file, adding time limit (maybe) to make some test fail.

Example (Failed test):
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/46670517b4541f2f6f8469b47fb5f67aee965032/parmys/regression_test/benchmark/task/mixing_optimization/mults_auto_half/task.conf#L31

So maybe the right choice is to replace placeholder `YYY` with an empty string in python side? Or more complicated, introduce an argparser option to bring options there?

AlexandreSinger:
I liked your current solution to this! To resolve the issues with that one task conf, we could update the script_params for the config files to allow this argument to pass through.

It may also be a good idea to default this to some good defaults; but thats would require a lot of testing and is not related to this issue.

PR #3488, opened 2026-04-14
[flow] Add -synthesis_params to replace YYY placeholder in parmys flow

WZ-Tong:
The `YYY` placeholder in parmys.py is never substituted during the VTR flow. It remains in the generated synthesis command and produces stray tokens in the yosys invocation.

#### Description

Remove the placeholder `YYY` in the template synthesis script

#### Related Issue

#3487

#### Motivation and Context

Accidentally found during debugging/writing my own script

#### Types of changes

### DIFF
--- a/doc/src/vtr/run_vtr_flow.rst
+++ b/doc/src/vtr/run_vtr_flow.rst
@@ -249,6 +249,13 @@ Detailed Command-line Options
     
     **Default:** default
 
+.. option:: -synthesis_params <SYNTHESIS_PARAMS>
+
+    Extra parameters passed to the ``parmys`` synthesis command in the Yosys script.
+    This replaces the ``YYY`` placeholder in ``synthesis.tcl``.
+
+    **Default:** "" (empty string)
+
 .. note::
 
     Yosys-Slang is a Yosys plugin that is built on top of the slang library to provide comprehensive SystemVerilog support.
--- a/doc/src/vtr/run_vtr_flow.rst
+++ b/doc/src/vtr/run_vtr_flow.rst
@@ -251,8 +251,8 @@ Detailed Command-line Options
 
 .. option:: -synthesis_params <SYNTHESIS_PARAMS>
 
-    Extra parameters passed to the ``parmys`` synthesis command in the Yosys script.
-    This replaces the ``YYY`` placeholder in ``synthesis.tcl``.
+    Specify additional synthesis parameters that are directly appended to the ``parmys`` command in the **default** synthesis flow. (e.g., ``-mults_ratio 0.5``, ``-exact_mults 0``).
+    For users using a **custom** Yosys script with ``-yosys_script``, this option acts as a dynamic substitute for the ``YYY`` placeholder (if present).
 
     **Default:** "" (empty string)
 
--- a/vtr_flow/scripts/run_vtr_flow.py
+++ b/vtr_flow/scripts/run_vtr_flow.py
@@ -373,7 +373,8 @@ def vtr_command_argparser(prog=None):
         "-synthesis_params",
         default=None,
         dest="synthesis_params",
-        help="Specify additional synthesis parameters directly appended to the parmys command in the default flow"
+        help="Specify additional synthesis parameters"
+        + " directly appended to the parmys command in the default flow"
         + " (e.g., -mults_ratio 0.5), or used to substitute the 'YYY' placeholder"
         + " in a custom Yosys script.",
     )
--- a/vtr_flow/scripts/run_vtr_flow.py
+++ b/vtr_flow/scripts/run_vtr_flow.py
@@ -373,8 +373,9 @@ def vtr_command_argparser(prog=None):
         "-synthesis_params",
         default="",
         dest="synthesis_params",
-        help="Extra parameters passed to the parmys synthesis command "
-        + "(replaces YYY placeholder in synthesis.tcl).",
+        help="Specify additional synthesis parameters directly appended to the parmys command in the default flow"
+        + " (e.g., -mults_ratio 0.5), or used to substitute the 'YYY' placeholder"
+        + " in a custom Yosys script.",
     )
     #
     # VPR arguments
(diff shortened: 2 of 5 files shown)


doc/src/vtr/run_vtr_flow.rst
```diff
@@ -249,6 +249,13 @@ Detailed Command-line Options
     
     **Default:** default
 
+.. option:: -synthesis_params <SYNTHESIS_PARAMS>
+
+    Extra parameters passed to the ``parmys`` synthesis command in the Yosys script.
+    This replaces the ``YYY`` placeholder in ``synthesis.tcl``.
+
+    **Default:** "" (empty string)
```
AlexandreSinger, line 257: The YYY placeholder in synthesis TCL script appears to be used to provide extra arguments to Parmys / Yosys. Is that correct?

If so, I think we should update this description to make it more clear what YYY actually is and why people may want to use this option.

vtr_flow/scripts/run_vtr_flow.py
```diff
@@ -369,6 +369,13 @@ def vtr_command_argparser(prog=None):
         + "system-verilog]. The script used the Yosys conventional Verilog"
         + " parser if this argument is not specified.",
     )
+    parmys.add_argument(
+        "-synthesis_params",
+        default="",
+        dest="synthesis_params",
+        help="Extra parameters passed to the parmys synthesis command "
+        + "(replaces YYY placeholder in synthesis.tcl).",
```
AlexandreSinger: Same comment down here. I think documenting what YYY is will make this interface more clear.

AlexandreSinger (changes requested):
Hi @WZ-Tong , thank you very much for these changes! I think they make sense to me, I just had some comments on documentation. Once the documentation is updated, I think this is good to merge.

doc/src/vtr/run_vtr_flow.rst
WZ-Tong: > The YYY placeholder in synthesis TCL script appears to be used to provide extra arguments to Parmys / Yosys. Is that correct?

Yes, currently it's in file:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/vtr_flow/misc/yosys/synthesis.tcl#L86

During invoking `run_vtr_flow.py`, flow will copy this default synth script (unless specified by `-yosys_script`) and replace the placeholder in python:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/vtr_flow/scripts/python_libs/vtr/parmys/parmys.py#L66-L75

vtr_flow/scripts/run_vtr_flow.py
WZ-Tong: > If so, I think we should update this description to make it more clear what YYY actually is and why people may want to use this option.

Yeah! Currently this placeholder is (may be) used when:
1. User uses **default synth script** wants to provide args to `parmys`
2. User writes his **own synth script** wants to reserve the flexibility to append args to `parmys`
3. [Running Test](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/46670517b4541f2f6f8469b47fb5f67aee965032/run_reg_test.py#L249) with specific [args](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/parmys/regression_test/benchmark/task/FIR/task.conf#L6) (the name of the new parser option is obtained here)

I'll modify the docs with these intent

doc/src/vtr/run_vtr_flow.rst
WZ-Tong: > The YYY placeholder in synthesis TCL script appears to be used to provide extra arguments to Parmys / Yosys. Is that correct?

According to the comment, the placeholder should be:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/vtr_flow/misc/yosys/synthesis.tcl#L23-L27

I thought it's simple to solve this because python code does not replace placeholder `YYY`, so I simply removed this placeholder, but found the github flow corrupted: [Parmys Basic Test Failed](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/24387249057/job/71224289138?pr=3488)

By reading the failed test, I found this placeholder is used and replaced by `sed`, the call chain is:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/46670517b4541f2f6f8469b47fb5f67aee965032/parmys/regression_test/benchmark/task/mixing_optimization/mults_auto_half/task.conf#L31
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/452ee1c467a3fa9362139f75c8f7ca11f3d379fe/parmys/verify_parmys.sh#L1044
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/46670517b4541f2f6f8469b47fb5f67aee965032/vtr_flow/scripts/python_libs/vtr/paths.py#L34
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/46670517b4541f2f6f8469b47fb5f67aee965032/run_reg_test.py#L249-L253

So, I added a parser option to provide this argument.

vtr_flow/scripts/run_vtr_flow.py
WZ-Tong: > 1. default synth script

In the default synth script, it provides `-a`, `-nopass`, `-c`. Options leftover: `-top`, `-exact_mults`, `-mults_ratio`, `-vtr_prim`, `-viz`.

WZ-Tong (commented):
Doc updated with default `None`.

Maybe the doc in the current version can make the usage of this parser argument clear

WZ-Tong (commented):
Make PyLint happy by split help lines

doc/src/vtr/run_vtr_flow.rst
AlexandreSinger: That makes sense. Thank you very much for looking into this!

AlexandreSinger (approved):
LGTM! Thanks again @WZ-Tong !

PR #3490, opened 2026-04-15
Directly construct NoOpLookahead in lookahead report

AmirhosseinPoolad:
Very small change. The lookahead report should not be using the overly complicated factory method to construct the NoOpLookahead with a lot of unneeded arguments. This should make changing the lookahead API a bit simpler and less annoying.


AlexandreSinger (approved):
LGTM, I completely agree with this change. Thanks @AmirhosseinPoolad !

PR #3491, opened 2026-04-15
[SDC] Basic create_generated_clock Support

AlexandreSinger:
Added basic support for the create_generated_clock SDC command.

This command allows the user to constrain a clock relative to another, master clock. This is mostly used in the cases where circuit elements within the design generate the clock signal. For example, a simple clock divider would be a flip-flop where D is tied to ~Q; Q would then be a clock 2x slower than the clock driving the flip-flop. This new constraint will allow the user specify the relationship between these two clocks.

There are currently two limitations of this implementation:

1. The generated clock is relative to the rising edge of the master clock. So the generated clock will always have 0 phase and 50% duty cycle.
2. The generated clock cannot have any phase, so the source clock must have 0 phase offset.

The second limitation is not a hard issue, but its something I wanted to investigate in isolation. I wanted to keep this PR introducing the constraint minimum viable product. What we have supported now should handle a large number of cases.


doc/src/vpr/sdc_commands.rst
```diff
@@ -63,6 +63,72 @@ If a virtual clock is assigned using a create_clock command, it must be referenc
 
 .. warning:: If a netlist clock is not specified with a :sdc:command:`create_clock` command, paths to and from that clock domain will not be analysed.
 
+create_generated_clock
+----------------------
+
+Defines a derived clock based on an existing source clock.
+
+Generated clocks are used to describe clock signals created by internal design logic (such as clock dividers or multipliers). The frequency of the generated clock is defined by scaling the frequency of a master clock.
```
vaughnbetz, line 71: I'd mention PLLs here too as they are the common case.

vpr/src/timing/read_sdc.cpp
```diff
@@ -259,9 +228,171 @@ class SdcParseCallback : public sdcparse::Callback {
...
+        }
+        //  Find the command that created this clock.
+        VTR_ASSERT(sdc_clocks_.contains(source_domain_id));
+        const sdcparse::CreateClock& source_clock_cmd = sdc_clocks_[source_domain_id];
+
+        // Get the generated clock period.
+        double source_clock_period = source_clock_cmd.period;
+        double generated_clock_period = sdcparse::UNINITIALIZED_FLOAT;
+        if (cmd.divide_by != sdcparse::UNINITIALIZED_INT) {
+            generated_clock_period = source_clock_period * static_cast<double>(cmd.divide_by);
```
vaughnbetz, line 307: Comment that dividing the frequency means multiplying the period.

vpr/src/timing/read_sdc.cpp
```diff
@@ -259,9 +228,171 @@ class SdcParseCallback : public sdcparse::Callback {
...
+            vpr_throw(VPR_ERROR_SDC, fname_.c_str(), lineno_,
+                      "named clocks with more than 1 unique target are not supported for create_generated_clock");
+        }
+
+        // Create a netlist clock for every target pin.
+        for (const auto& p : target_pins) {
+            AtomPinId clock_pin = p.first;
+            const std::string& object_name = p.second;
+
+            // Get the name of the clock. If no name is provided, the name of the object is used.
```
vaughnbetz, line 365: object (clock pin, port or net).

vpr/src/timing/read_sdc.cpp
```diff
@@ -259,9 +228,171 @@ class SdcParseCallback : public sdcparse::Callback {
...
+            } else {
+                VTR_ASSERT(target_pins.size() == 1);
+                clock_name = cmd.name;
+            }
+
+            // Get the clock source associated with this pin.
+            tatum::NodeId clock_source = get_clock_source(clock_pin);
+            VTR_ASSERT(clock_source.is_valid());
+
+            // Create an equivalent create_clock commands for this generated clock.
```
vaughnbetz, line 378: remove 'an'

vaughnbetz (approved):
LGTM, a few suggestions (like a PLL example).

PR #3492, opened 2026-04-15
[Pack] Add candidates directly from same physical ram group

haydar-c:
For the RAM clusters, we were filtering the candidates that are not in the same physical ram group. Instead of filtering the candidates later, we directly offer the candidates only from the same physical ram group now. This has a little improvement on the Koios benchmark (heavier on memory blocks) assuming due to not being limited anymore by the greedy packer candidate offering to fill up the RAM clusters.

The affect on the results is marginal. All collected on AP flow.

**Titan Result Summary (AP flow)**

  | Routed WL |Final CPD | Avg. Displacement | Total Runtime | # of LAB | # of M9K | # of M144K
PR / Master | 0.9988 | 1.0010 | 1.0000 | 0.9849 | 1.0000 | 1.0000 | 1.0000

**Koios Results Summary (AP flow)**

 | Routed WL |Final CPD | Avg. Displacement | Total Runtime | # of CLB | # of memory
PR / Master | 0.9998 | 0.9946 | 0.9970 | 0.9988 | 0.9999 | 0.9982


vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -702,6 +677,77 @@ PackMoleculeId GreedyCandidateSelector::get_next_candidate_for_cluster(
...
+    // RAM clusters offer only atoms from the same physical RAM group. All other
+    // clusters use net-traversal based candidate search (connectivity and timing,
+    // transitive, high-fanout, and attraction groups).
+    if (cluster_gain_stats.is_memory && has_ram_groups_) {
+        add_ram_cluster_molecule_candidates(cluster_gain_stats,
+                                            cluster_id,
+                                            cluster_legalizer,
+                                            attraction_groups);
+    } else {
+        add_cluster_molecule_candidates(cluster_gain_stats,
```
AlexandreSinger: Small nit: The names of these two functions are a bit similar. What do you think of renaming this second function to `add_general_cluster_mol_candidates`. This just makes it clear that its a general candidate proposal method. Feel free to push back though if you think its fine the way it is.

AlexandreSinger (commented):
LGTM, just one nit for you to think over.

The overall results are small, but that makes sense given these changes. The good thing is that it is a slight improvement instead of a slight regression.

vpr/src/pack/greedy_candidate_selector.cpp
haydar-c: I spend time to find a better name but could not end up with something better and I completely agree `add_general_cluster_mol_candidates` makes the distinction between two clearer. Thanks @AlexandreSinger !

vpr/src/pack/greedy_candidate_selector.cpp
```diff
@@ -914,6 +918,34 @@ void GreedyCandidateSelector::add_cluster_molecule_candidates_by_highfanout_conn
     cluster_gain_stats.tie_break_high_fanout_net = AtomNetId::INVALID(); /* Mark off that this high fanout net has been considered */
 }
 
+void GreedyCandidateSelector::add_cluster_molecule_candidates_by_ram_group(
+    ClusterGainStats& cluster_gain_stats,
+    LegalizationClusterId legalization_cluster_id,
+    const ClusterLegalizer& cluster_legalizer,
+    AttractionInfo& attraction_groups) {
+    // TODO: Atoms are offered in the order they appear in the physical RAM group,
+    //       which determines the order they are packed into the cluster. Note that
```
vaughnbetz: I doubt this would make a difference except maybe for big RAMs that span more than 1 row or col. There it might make a difference.

vpr/src/pack/greedy_candidate_selector.h
```diff
@@ -465,6 +465,28 @@ class GreedyCandidateSelector {
...
+     */
+    void add_general_cluster_molecule_candidates(
+        ClusterGainStats& cluster_gain_stats,
+        LegalizationClusterId legalization_cluster_id,
+        const ClusterLegalizer& cluster_legalizer,
+        AttractionInfo& attraction_groups);
+
+    /**
+     * @brief Populate the feasible blocks list for a RAM cluster by offering
+     *        only atoms from the same physical RAM group as the cluster seed.
```
vaughnbetz: does it 'offer' all the atoms from that physical RAM group, or is there a limit enforced by some parameter?

Also to discuss (not for this PR): how does this offer work?  Are we populating a list/heap in one step and them pulling thing out in another?  Possibly we could be faster for RAMs by just proposing on primitive at a time.

vpr/src/pack/greedy_candidate_selector.h
```diff
@@ -529,6 +551,17 @@ class GreedyCandidateSelector {
         const ClusterLegalizer& cluster_legalizer,
         AttractionInfo& attraction_groups);
 
+    /**
+     * @brief Add molecules from the same physical RAM group as candidates for
+     *        the given RAM cluster. Molecules in the same physical RAM group are
+     *        pre-determined by the RAM mapper to be packable together.
```
vaughnbetz: How does this function differ from the one above (add_ram_cluster_molecule_candiddates)?  They seem very similar ... maybe give a little more detail or an example use (or make the names more distinct.

vaughnbetz (approved):
I have a fewncomments (commenting / naming) but this is good to merge if you take a crack at addressing those.

vpr/src/pack/greedy_candidate_selector.h
haydar-c: Thanks for pointing out @vaughnbetz, I have added the limitation as todo.

vaughnbetz: I think we should ignore the feasible block limit when we have an entire physical RAM preselected.  OK with TODO for now, but please raise an issue.

PR #3497, opened 2026-04-21
[Pack] Pack RAM clusters first and set candidate propose limit to RAM group size

haydar-c:
For RAM clusters, we are only offering candidates from the physical RAM groups (as of #3492). Since we know what to be packed in these clusters, it makes sense to move seeds of type memory to front if we have RAM groups. That way types can have the transitive connectivity information while they are being packed. 

We also set the candidate propose limit to the number of atoms in the RAM group of that seed since we know the exact number of atoms to be packed in RAM clusters. Both changes makes the RAM packing more intuitive. The change in results is nearly noise to me.

Titan results normalized to master

| Metric        | PR/Master |
| ------------- | --------- |
| Routed WL     | 1.0048    |
| Final CPD     | 0.9989    |
| Avg. Disp.    | 0.9944    |
| GP Runtime    | 1.0157    |
| FL Runtime    | 1.0001    |
| DP Runtime    | 1.0254    |
| Route Runtime | 1.0489    |
| Total Runtime | 1.0137    |
| LAB Count     | 0.9988    |
| M9K Count     | 1.0000    |
| M144 Count    | 1.0000    |


AlexandreSinger (approved):
LGTM

PR #3498, opened 2026-04-21
CRR Pin Name

amin1377:
Update the CRR code to use pin names instead of pin PTC numbers to identify the corresponding nodes in the CSV files.


vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -301,15 +311,44 @@ CRRConnectionBuilder::SegmentInfo CRRConnectionBuilder::parse_segment_info(const
...
 }
 
+int CRRConnectionBuilder::resolve_pin_ptc(const SegmentInfo& info,
+                                          int x,
+                                          int y) const {
+    if (info.pin_name.empty()) {
+        VTR_LOG_ERROR("No pin name is specified for segment index %d at (%d,%d)\n", info.seg_index, x, y);
+    }
+
+    auto tile_type = g_vpr_ctx.device().grid.get_physical_type({x, y, 0});
```
AlexandreSinger, line 322: Magic number of 0. Although 3D is not supported for CRR, you may want to use an assert on the grid here to ensure that the grid has only 1 layer and use a variable for the layer (so if we choose to make this 3D in the future, it will be obvious how to fix this).

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -301,15 +311,44 @@ CRRConnectionBuilder::SegmentInfo CRRConnectionBuilder::parse_segment_info(const
...
+    if (tile_type == nullptr) {
+        VTR_LOG_ERROR("No tile type found while resolving pin '%s' at (%d,%d)\n",
+                      info.pin_name.c_str(), x, y);
+    }
+
+    for (int pin_ptc = 0; pin_ptc < tile_type->num_pins; ++pin_ptc) {
+        if (block_type_pin_index_to_name(tile_type, pin_ptc, false) == info.pin_name) {
```
AlexandreSinger, line 331: This is not really doing much parsing. So for example, suppose a port only had 1 pin. A user may write it as "port[0]", but I think the code above would not allow it since it would simplify the name to "port". This search is also a bit slow to do for every pin we read in.

Is it possible to somehow parse the pin_name and get the associated port and pin offset?

If not, we should at least document this somewhere in the documentation.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -301,15 +311,44 @@ CRRConnectionBuilder::SegmentInfo CRRConnectionBuilder::parse_segment_info(const
...
+            return pin_ptc;
+        }
+    }
+
+    VTR_LOGV(verbosity_ > 0,
+             "Failed to resolve %s pin '%s' on tile type '%s' at (%d,%d); falling back to CSV index %d\n",
+             rr_node_typename[(info.side == e_sw_template_dir::OPIN) ? e_rr_type::OPIN : e_rr_type::IPIN],
+             info.pin_name.c_str(),
+             tile_type->name.c_str(),
+             x, y, info.seg_index);
```
AlexandreSinger, line 341: Instead of just warning, should this not produce a crash? What is the affect of using the CSV index to the user?

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -301,15 +311,44 @@ CRRConnectionBuilder::SegmentInfo CRRConnectionBuilder::parse_segment_info(const
     return info;
 }
 
+int CRRConnectionBuilder::resolve_pin_ptc(const SegmentInfo& info,
+                                          int x,
+                                          int y) const {
+    if (info.pin_name.empty()) {
+        VTR_LOG_ERROR("No pin name is specified for segment index %d at (%d,%d)\n", info.seg_index, x, y);
```
AlexandreSinger: This will not crash the program, it will just produce the error. Should this be a thrown error?

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
```diff
@@ -128,6 +129,10 @@ class CRRConnectionBuilder {
                                     const std::unordered_map<NodeHash, RRNodeId, NodeHasher>& col_nodes,
                                     const std::unordered_map<NodeHash, RRNodeId, NodeHasher>& row_nodes) const;
 
+    int resolve_pin_ptc(const SegmentInfo& info,
```
AlexandreSinger, line 146: It may be useful to document what this function is doing, since it may not be obvious.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
```diff
@@ -105,6 +105,7 @@ class CRRConnectionBuilder {
         std::string seg_type;
         int seg_index;
         int tap;
+        std::string pin_name;
```
AlexandreSinger, line 110: I recommend adding a comment somewhere here. When the pin_name is provided, is the seg_index or tap used? If not, it may be useful to make this clear.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/data_frame_processor.h
```diff
@@ -42,15 +43,21 @@ struct Cell {
...
 
     bool is_number() const {
-        return is_integer();
+        if (is_integer()) return true;
+        if (auto p = std::get_if<std::string>(&value)) {
+            const char* start = p->c_str();
+            char* end = nullptr;
+            double d = std::strtod(start, &end);
+            return end == start + p->size() && std::isfinite(d) && std::floor(d) == d;
+        }
```
AlexandreSinger, line 57: This if statement is not obvious what it is doing. You should add a comment explaining what it is doing and why.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/data_frame_processor.h
```diff
@@ -42,15 +43,21 @@ struct Cell {
...
 
     int64_t as_int() const {
-        if (auto p = std::get_if<int64_t>(&value)) {
-            return *p;
-        }
-        if (auto p = std::get_if<std::string>(&value)) {
-            return std::stoll(*p);
+        if (auto p = std::get_if<int64_t>(&value)) return *p;
+        if (is_number()) {
+            if (auto p = std::get_if<std::string>(&value))
```
AlexandreSinger, line 64: Shouldn't this be an assert? I suggest crashing somehow if the value cannot be cast to a string. In the current implementation, this case will not crash and will just return 0.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/data_frame_processor.h
```diff
@@ -42,15 +43,21 @@ struct Cell {
...
+                return static_cast<int64_t>(std::strtod(p->c_str(), nullptr));
         }
         return 0;
```
AlexandreSinger: Why is this returning 0? Why not crash? Or at least using a weird number like -1 or -INT_MAX.

AlexandreSinger (commented):
Hi @amin1377 , these changes look good! Its nice that we are using the pin names directly. I left some comments below.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
amin1377: Good point. I added the assertion in the parent function that calls `get_tile_connections`, since only `x` and `y` are passed to the downstream functions.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
amin1377: I didn’t quite understand what you mean here. info.pin_name is something like io.in[2], and the function I called would return such a name based on the tile type and pin PTC number.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
amin1377: This is basically for backward compatibility. If the pin name in the CSV file is not valid, we issue a warning but fall back to the PTC number specified in the CSV file.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
amin1377: I changed it to VPR_FATAL_ERROR. I do not think this is the first time I have made this mistake :)

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/data_frame_processor.h
amin1377: Added this:
```
// strtod sets `end` to the first character it couldn't parse.
// The three conditions together check: the entire string was consumed
// (no trailing garbage), the value is not inf/nan, and it has no
// fractional part (i.e. "3.0" qualifies but "3.5" does not).
```

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/data_frame_processor.h
amin1377: Changed it to the following:
```
VTR_ASSERT_MSG(false, "as_int() called on a non-numeric Cell");
return 0; // unreachable
```

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
AlexandreSinger: Using your example, suppose we had an IO tile with a single input pin (silly, I know, but some tile may have a special port). Then a user may write `io.in[0]`; however, I do not think this will match. The user would have to write `io.in`. The reason is that `block_type_pin_index_to_name` simplifies the pin names automatically; but I am not sure that is what you want in this case.

I think it would make more sense to write a function which takes the string `io.in[0]` and turns it into {tile = io, port = in, pin_index = 0}. Then you can just quickly lookup that the given port and pin exist for the given tile name. This is how I think I did it in the rust visualizer.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
AlexandreSinger: As long as that works for you guys it should be fine!

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/data_frame_processor.h
AlexandreSinger: Just to confirm: you are using an assert here; that means that the end-user would not crash (since we assume the end-user is using a lower assert level for performance). So you are saying that this code should never be hit, and you only want to crash when we are developing.

If it is convenient, it may be cleaner just to raise a VTR error. Then I do not think the return is even necessary since the compiler should recognize a throw as terminating a function block.

amin1377: Yes, that's my intention. I'll change it to VTR error.

amin1377:
@AlexandreSinger: Please let me know if you have any other concerns. Thanks!

AlexandreSinger (approved):
LGTM

PR #3499, opened 2026-04-22
avoid to create infinite loop while macro formation

saaramahmoudi:
#### Description

I recently hit a testcase that we had a net driving the same block and VPR tries to form macro with it. My direct connection was kind of irregular in the architecture that I was testing which is not something that VPR user would try. Anyway, I thought maybe we should check this condition and avoid the loop goes forever in case somebody define a directlist like mine. 

#### Types of changes

#### Checklist:


vpr/src/place/place_macro.cpp
```diff
@@ -224,7 +224,7 @@ int PlaceMacros::find_all_the_macro_(const ClusteredNetlist& clb_nlist,
                     //
                     // The output SOURCE (from_pin) of a true head macro will:
                     //  * drive another block with the same direct connection
-                    if (from_src_or_sink == e_pin_type::DRIVER && to_idirect == from_idirect && from_net_id != ClusterNetId::INVALID()) {
+                    if (from_src_or_sink == e_pin_type::DRIVER && to_idirect == from_idirect && from_net_id != ClusterNetId::INVALID() && is_net_direct_connection(from_net_id, to_idirect, clb_nlist)) {
```
amin1377, line 227: I think the above commit needs to be augmented with an explanation of why is_net_direct_connection is needed here.

vpr/src/place/place_macro.cpp
```diff
@@ -679,6 +679,11 @@ bool PlaceMacros::is_net_direct_connection(ClusterNetId clb_net, int idirect, co
     ClusterPinId net_sink = clb_nlist.net_pin(clb_net, 1);
     ClusterBlockId sink_block_id = clb_nlist.pin_block(net_sink);
 
+    if (sink_block_id == block_id) {
+        //net is connected back to the same block, should not be counted as a direct connection
+        return false;
```
amin1377, line 687: I think this is a strange case. Shouldn’t we warn the user about the existence of such a case?

amin1377 (commented):
Thanks, Sara. Just to be cautious, could you please run the VTR and Titan benchmarks and confirm that the QoR has not changed at all?

vpr/src/place/place_macro.cpp
saaramahmoudi: It might be the packer trying to exploit the direct connection to get back to the same block, meaning that there is nothing wrong inherently with the design to notify the user about. With the previous code, it will end up in the segmentation fault, and is hard to debug. But if we avoid to enter the infinite loop, the routing would fail with a correct error message that this can't be done which is very clear error message.
Do you still think this should turn to warning?

amin1377: I guess I do not have a strong opinion on this. The main reason I think the user should be warned about such a case is that we have not encountered it before, and this is the first time it has come up.

saaramahmoudi: I don't have a strong opinion either just don't know what should be printed as a warning to user as the design is not inherently wrong. If you have any suggestion for the warning message, I would add that in.

saaramahmoudi:
> Thanks, Sara. Just to be cautious, could you please run the VTR and Titan benchmarks and confirm that the QoR has not changed at all?

@amin1377 @vaughnbetz  Attached the results for both benchmarks, I don't see any quality changes at all (As we expected). There are some timing changes which I think I ran the benchmarks in a loaded machine. 

Link to results:
[Titan](https://docs.google.com/spreadsheets/d/1fKGxhBbt8O_7b255YFMAlTF2hkprTby1sgwxwQNuuCg/edit?usp=sharing)
[VTR](https://docs.google.com/spreadsheets/d/108x2sRORdXF_X1H6Pjx4kTyIK85TqcC6y5nczcjjER4/edit?usp=sharing)

vpr/src/place/place_macro.cpp
saaramahmoudi: The comment above is exactly what you are asking for: "drive another block with the same direct connection"

amin1377 (approved):

saaramahmoudi:
@vaughnbetz can you please merge this after the QoR has been completed?

vaughnbetz (approved):
LGTM

Issue #3434, opened 2026-02-23
[WIP] VPR constraints can force the mapping of an atom block to a fixed location of clustering


tangxifan:

treelin611:
My intuition is to add the constraint at the prepacker stage, specifically in the following code:

<img width="857" height="547" alt="Image" src="https://github.com/user-attachments/assets/8c41ccd8-7170-4338-ba70-414509f16020" />
Do you have any suggestions?

vaughnbetz:
This is fixed in PR #3501

PR #3501, opened 2026-04-27
Constrain atom block mapping location on exact pb_type

treelin611:
address the issue: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3434

1. Parse logical block location constraints from vpr_constraint.xml.
2. Apply these constraints during clustering by rejecting illegal primitive locations.


vpr/src/pack/logic_block_location_util.h
```diff
@@ -0,0 +1,23 @@
+﻿#pragma once
+
+#include <string>
+#include <unordered_set>
+#include <vector>
+
+#include "atom_netlist_fwd.h"
+
+struct t_pb_type;
```
tangxifan: You can include the ``physical_types.h" header file instead. No need to repeat the declaration.

vpr/src/pack/greedy_clusterer.cpp
```diff
@@ -1,40 +1,40 @@
...
+  *  For example:
+  *      Consider an architecture where CLBs contain 4 BLEs, and each BLE is a LUT + FF pair.
+  *      We wish to map a netlist of 400 LUTs and 400 FFs.
+  *
+  *      A BLE corresponds to one t_pb_type (which has num_pb = 4).
+  *
+  *      Each of the 4 BLE positions corresponds to a t_pb_graph_node (each of which references the BLE t_pb_type).
+  *
+  *      The output of clustering is 400 t_pb of type BLE which represent the clustered user netlist.
+  *      Each of the 400 t_pb will reference one of the 4 BLE-type t_pb_graph_nodes.
```
tangxifan: Please restore unnecessary changes on the code comments.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1,18 +1,17 @@
...
- * @author  Alex Singer
- * @date    September 2024
- * @brief   The implementation of the Cluster Legalizer class.
- *
- * Most of the code in this file was originally part of cluster_util.cpp and was
- * highly integrated with the clusterer in VPR. All code that was used for
- * legalizing the clusters was moved into this file and all the functionality
- * was moved into the ClusterLegalizer class.
- */
+﻿/**
```
tangxifan: Please restore unnecessary changes on the code comments.

vpr/src/pack/prepack.h
```diff
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
```
tangxifan: Please restore unnecessary changes on the code comments.

vpr/src/pack/prepack.cpp
```diff
@@ -1,14 +1,14 @@
...
- * This gives hints to the packer on what groups of blocks to keep together during packing.
- * Primary purpose:
- *    1) "Forced" packs (eg LUT+FF pair)
- *    2) Carry-chains
- * Duties: Find pack patterns in architecture, find pack patterns in netlist.
- *
- * Author: Jason Luu
- * March 12, 2012
- */
+﻿/*
```
tangxifan: Please restore unnecessary changes on the code comments.

vpr/src/base/vpr_types.h
```diff
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
```
tangxifan: Please restore unnecessary changes on the code comments.

vpr/src/base/vpr_constraints_writer.cpp
```diff
@@ -1,8 +1,8 @@
-/*
- * vpr_constraints_writer.cpp
- *
- *      Author: khalid88
- */
+﻿/*
```
tangxifan: Please restore unnecessary changes on the code comments.

vpr/src/base/user_place_constraints.h
```diff
@@ -129,6 +130,26 @@ class UserPlaceConstraints {
...
+     * matches.
+     */
+    void set_atom_logical_block_location(AtomBlockId blk_id, const std::string& logical_block_location);
+
+    /**
+     * @brief Return an atom's logical_block_location constraint, if one exists.
+     *
+     * Returns empty string when no location constraint was provided.
+     */
+    const std::string get_atom_logical_block_location(AtomBlockId blk_id) const;
```
tangxifan: return ``std::string`` is good enough. I do not see why we should ensure it is a constant.

vpr/src/base/gen/vpr_constraints_uxsdcxx.h
```diff
@@ -206,6 +207,39 @@ inline atok_t_add_atom lex_attr_t_add_atom(const char* in, const std::function<v
                     break;
             }
             break;
+        case 22:
```
tangxifan, line 210: Is this auto generated or you manually added?

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -38,7 +37,25 @@
...
-#include "cluster_placement.h"
+#include "logic_block_location_util.h"
+
+static bool check_logical_block_location_constraint(const AtomBlockId blk_id, const t_pb* pb, int verbosity) {
+    const auto& constraints = g_vpr_ctx.floorplanning().constraints;
+    const std::string logical_block_location = constraints.get_atom_logical_block_location(blk_id);
+    if (logical_block_location.empty()) {
+        return true;
+    }
+    if (logical_block_location_matches_hierarchical_type(logical_block_location, pb->hierarchical_type_name())) {
```
tangxifan: Can you provide an example string of the ``pb->hierarchical_type_name()`` ?  If we can support the same string, it can save your effort when building the parser in the ``logic_block_location_util.h``

vpr/src/pack/logic_block_location_util.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include "pack_patterns.h"
+#include "physical_types.h"
+#include "vpr_utils.h"
+
+enum class e_token_format {
+    LOGICAL_LOCATION,
+    HIERARCHICAL_TYPE,
+};
+
+static int try_parse_int(const std::string& value) {
```
tangxifan: If failed, you should throw error messages in addition to return an invalid value.

tangxifan (commented):
@treelin611 Good work in general. Please check details. The parser looks complicated in my view. We can do simply string comparison, it will speed up packer. If not, we should try to parse the string once instead decoding in each iteration of packing an atom block.

Also, we should check if each logical_block_location is a valid one in the existing ``pb_graph``, before packer starts. Early exit is preferred here.

vpr/src/base/gen/vpr_constraints_uxsdcxx.h
treelin611: This file is auto generated from vpr_constraints.xsd

vpr/src/pack/cluster_legalizer.cpp
treelin611: Image

`clb[0][default]/fle[1][n1_lut4]/ble4[0][default]/ff[0]` is the` pb->hierarchical_type_name()`

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -39,7 +38,25 @@
...
+        return true;
+    }
+
+    VTR_LOGV(verbosity > 3,
+             "\t\t\tFAILED logical_block_location constraint: atom '%s' expected '%s' but candidate '%s'\n",
+             g_vpr_ctx.atom().netlist().block_name(blk_id).c_str(),
+             logical_block_location.c_str(),
+             pb->hierarchical_type_name().c_str());
```
treelin611, line 86: The log is printed out from this line: 

Image

treelin611: This part of the code works like a quality checker. If a logical_block_location constraint is specified, it verifies the mapping result produced by the clustering algorithm. The mapping will only pass the check when the atom is placed at the exact logical block location we specified.

vpr/src/pack/logic_block_location_util.cpp
treelin611: try_parse_int returns -1 on purpose for hierarchical-type tokens: non-numeric bracket content is a mode name (e.g. fle[3][n1_lut4]), not a parse failure. I will add VPR_FATAL_ERROR at the LOGICAL_LOCATION parse path when [...] is non-empty but not a valid index, so invalid constraints are reported separately from the hierarchical mode case.

treelin611: please refer to line 58

vpr/src/pack/logic_block_location_util.cpp
```diff
@@ -0,0 +1,151 @@
...
+                            "Invalid logical_block_location token '%s': missing ']' after '['",
+                            token.c_str());
+        }
+        return parsed;
+    }
+
+    if (format == e_token_format::LOGICAL_LOCATION) {
+        const std::string bracket_content = token.substr(lbr + 1, rbr - lbr - 1);
+        if (!bracket_content.empty()) {
+            const int idx = try_parse_int(bracket_content);
```
treelin611: Try to error out if the index is invalid.

vpr/src/pack/logic_block_location_util.cpp
```diff
@@ -0,0 +1,151 @@
+﻿#include "logic_block_location_util.h"
+
+#include <algorithm>
+#include <cctype>
+#include <utility>
+
+#include "vpr_error.h"
+
+enum class e_token_format {
```
tangxifan: Can you convert the C functions to a class?
It can be 

class LbHierPathParser {
   LbHierPathParser(const std::string&);
   int parse();
   t_pb_graph_node* target_lb() const;
};

vpr/src/pack/cluster_legalizer.cpp
tangxifan: O.k. To avoid code complexity, we can follow the hierarchical_type_name() convention. it can simplify our codes

vpr/src/base/user_place_constraints.cpp
```diff
@@ -93,6 +93,22 @@ const std::unordered_set<t_logical_block_type_ptr>& UserPlaceConstraints::get_pa
     return it->second;
 }
 
+void UserPlaceConstraints::set_atom_logical_block_location(AtomBlockId blk_id, const std::string& logical_block_location) {
+    atom_logical_block_locations_[blk_id] = logical_block_location;
```
tangxifan, line 103: When you add the string, is it possible to apply sanity checks ? Ensure the path is valid (match one of the pb_type / pb_graph_node, which you can get from t_logical_block_type). Here you can use the parser in logic_block_location_util.h

tangxifan (changes requested):
@treelin611 Thanks for cleaning up. It is now getting into a good shape. Can you try to simplify and make it robust? See my detailed comments.

tangxifan:
@treelin611 Can you update documentation? Thanks!

tangxifan:
Please remove the tag ``[WIP]`` from the title. This PR is ready for review.

tangxifan:
@treelin611 Per discussion with @vaughnbetz , before merging this PR, QoR check should be run a nightly task to ensure no significant runtime, peak memory usage and QoR degradation.

https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/vtr_flow/tasks/regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain_depop/config

I suggest you run the test locally on the current master and then on this feature branch

vpr/src/base/user_place_constraints.h
```diff
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
```
AlexandreSinger: I think this line change is indicative of a byte-order mark that was added to this and a few other files. This can happen if you are using Microsoft Visual Studio to work with VTR. In VTR we require our source files to be UTF-8 encoded for portability reason. as per our editor config:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/919090cd5bea96865593da0a0842c29deb66dafe/.editorconfig#L4-L5

Please remove the byte-order mark from this file and ensure the other files doe not have one either.

vpr/src/pack/logic_block_location_util.cpp
```diff
@@ -0,0 +1,195 @@
...
+            parsed.index = idx;
+        } else if (!first_bracket.empty()) {
+            parsed.mode = first_bracket;
+        }
+
+        size_t lbr2 = token.find('[', rbr + 1);
+        if (lbr2 != std::string::npos) {
+            size_t rbr2 = token.find(']', lbr2 + 1);
+            if (rbr2 != std::string::npos && rbr2 > lbr2 + 1) {
+                parsed.mode = token.substr(lbr2 + 1, rbr2 - lbr2 - 1);
```
AlexandreSinger, line 106: I thought modes were supposed to be wrapped in curly braces "{ / }"? At lines 163 and 165 you are parsing for square braces. Is something wrong here?

I think all of the functions in this file needs some comments to make it easier to review.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/vpr_constraint.xml
AlexandreSinger, line 1: This testcase is good. I think we may need a a bit more for other important logical block locations to lock this feature in well.

I would really like to see a testcase which locks the IOs to a specific logical location. This testcase also always  locks to the same mode (n1_lut4). I have seen the packer can have some order dependence, so I think we should also have a testcase which constrains the primitives to another mode (maybe LUT5 mode if it exists).

vpr/src/pack/logic_block_location_util.h
```diff
@@ -0,0 +1,63 @@
...
+     */
+    void validate_logical_block_types(const std::vector<t_logical_block_type>& logical_block_types);
+
+    enum class TokenFormat {
+        LOGICAL_LOCATION,
+        HIERARCHICAL_TYPE,
+    };
+
+    static bool token_matches(const t_logical_location_token& want, const t_logical_location_token& got);
+    static std::vector<t_logical_location_token> parse_segmented_tokens_impl(const std::string& input,
```
AlexandreSinger: Both of these static methods are only used by methods within logic_block_location_util.cpp. I think they can be moved to static methods within that file instead of being exposed by this class' API.

They also both need doxygen comments.

vpr/src/pack/logic_block_location_util.h
```diff
@@ -0,0 +1,63 @@
...
+                                                                             char separator,
+                                                                             TokenFormat format);
+
+  private:
+    static int try_parse_int_impl(const std::string& value);
+    static t_logical_location_token parse_single_token_impl(const std::string& token, TokenFormat format);
```
AlexandreSinger: Same thing with these static methods.

vpr/src/pack/logic_block_location_util.h
```diff
@@ -0,0 +1,63 @@
...
+
+    std::string logical_block_location_;
+    std::vector<t_logical_location_token> want_tokens_;
```
AlexandreSinger, line 78: These two member variables need documentation. Especially `want_tokens_` as it appears to be very important to this parser.

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -234,7 +243,6 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
     virtual inline void set_add_logical_block_name_pattern(const char* name_pattern, void*& /*ctx*/) final {
         lb_type_name_pattern_ = name_pattern;
     }
-
```
AlexandreSinger, line 237: I do not think this line was meant to be removed.

vpr/src/pack/logic_block_location_util.cpp
```diff
@@ -0,0 +1,195 @@
...
+    for (size_t start = 0; start + want_tokens_.size() <= got_tokens.size(); ++start) {
+        bool all_match = true;
+        for (size_t i = 0; i < want_tokens_.size(); ++i) {
+            if (!token_matches(want_tokens_[i], got_tokens[start + i])) {
+                all_match = false;
+                break;
+            }
+        }
+        if (all_match) {
+            return true;
```
AlexandreSinger: This uses a sliding window to match for a string. This may match any substring that contains the `want_tokens_`. Is this intentional?

If it is, I think we need to document that substrings may be matched and add a regression test for this; but I am not sure if we want this behavior or not.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/vpr_constraint.xml
```diff
@@ -0,0 +1,12 @@
+<vpr_constraints tool_name="vpr">
+   <partition_list>
+       <partition name="fix_pb_mapping_loc">
+            <add_atom name_pattern="d" logical_block_location="clb[0].fle[0]{n1_lut4}.ble4[0].ff[0]"/>
+            <add_region x_low="2" y_low="1" x_high="2" y_high="1"/>
+       </partition>
+       <partition name="fix_pb_mapping_loc">
```
AlexandreSinger: The two partitions in this constraint XML has the same name. I think they may need to be unique (double check this though!).

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/config/config.txt
```diff
@@ -0,0 +1,27 @@
...
+parse_file=vpr_standard.txt
+
+# How to parse QoR info
+qor_parse_file=qor_standard.txt
+
+# Pass requirements
+pass_requirements_file=pass_requirements.txt
+
+# Script parameters
+script_params=-starting_stage vpr -track_memory_usage --analytical_place --route --max_router_iterations 150 --pack_verbosity 4 --read_vpr_constraints ../../../../vpr_constraint.xml
```
AlexandreSinger: I do not think we should set the pack_verbosity in this testcase. It generates tons of log messages which can slow down the CI's parsers. We use the verbosity parameters for debugging.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -39,7 +39,27 @@
 #include "vtr_vector.h"
 #include "vtr_vector_map.h"
 #include "lazy_pop_unique_priority_queue.h"
-#include "cluster_placement.h"
+#include "logic_block_location_util.h"
+
+static bool check_logical_block_location_constraint(const AtomBlockId blk_id, const t_pb* pb, int verbosity) {
```
AlexandreSinger, line 70: This function needs doxygen comments. It should be clear what it is checking for and why.

AlexandreSinger (changes requested):
I like this feature a lot! I am excited to have it in. I have several comments below.

This feature should also be documented in this file: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/doc/src/vpr/placement_constraints.rst

vpr/src/pack/logic_block_location_util.cpp
treelin611: This is intentional, not a bug. 

LbHierPathParser supports two different syntaxes: for `LbHierPathParser::TokenFormat::LOGICAL_LOCATION`, mode is encoded with `{} `(e.g.` clb[0].fle[0]{n1_lut4}.ble4[0].ff[0]`); for `LbHierPathParser::TokenFormat::HIERARCHICAL_TYPE`, we parse` t_pb::hierarchical_type_name()`, whose mode encoding uses `[] `(e.g.` clb[0][default]/fle[3][n1_lut4]/ble4[0][default]/ff[0]`). So parsing [] in that branch is expected behavior.

I agree on readability, and I will add function comments in this file to clearly document the two formats and why they differ.

vpr/src/pack/logic_block_location_util.cpp
treelin611: Yes, the sliding-window loop is intentional. We tokenize both sides (`want_tokens_ `from `logical_block_location`, `got_tokens` from` pb->hierarchical_type_name()`) and match a contiguous token subsequence (name/index/mode).

The constraint path and the candidate path use different syntax and are often different lengths. For example:

Constraint:` clb[0].fle[1]{n1_lut4}.ble4[0].lut4[0]`
Candidate: `clb[0][default]/fle[1][n1_lut4]/ble4[0][default]/lut4[0][lut4]/lut[0]`
The candidate is usually longer because it is the full primitive hierarchy (e.g. extra `lut[0]` under` lut4[0]`). The sliding window lets the shorter constraint match the longer internal path.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/vpr_constraint.xml
treelin611: Thanks for catching that — duplicate partition names were a typo. I'll rename them.

treelin611:
Thank you for your comment @tangxifan . Attached/Provided is the comparison of qor_result.txt between the master branch and the constrain_atom branch for https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/vtr_flow/tasks/regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain_depop/config.

img_v3_02129_1a6b4627-1d3a-4c75-8352-f1902a3cee5g

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/vpr_constraint.xml
tangxifan: @AlexandreSinger If you need to assign I/O to dedicated location, it is done by the ``.place file`` through the command-line option ``--fix_clusters``. 
VPR constraints focuses on the micro placements on the pb_types. 
I think you have raised a good question, which one should have the highest priority. VPR constraints or .place file, when conflicts are seen.

I will provide a testcase on the multi-mode logic block for @treelin611 to add

tangxifan:
@treelin611 Please provide the spreadsheet file instead of a screenshot. All details should be revealed. There is no secret to hide. Thank you.

> Thank you for your comment @tangxifan . Attached/Provided is the comparison of qor_result.txt between the master branch and the constrain_atom branch for https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/vtr_flow/tasks/regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain_depop/config.
> 
> img_v3_02129_1a6b4627-1d3a-4c75-8352-f1902a3cee5g

doc/src/vpr/placement_constraints.rst
```diff
@@ -83,6 +83,16 @@ Partitions, Atoms, Regions, and Logical Block Types
 			A boolean value indicating whether the ``name_pattern`` should be treated as a regular expression.
 			**Default:** ``false``
 
+		:opt_param logical_block_location:
+			Constrains the atom to a specified location inside a logic block (e.g. a particular FF or LUT
+			slice in a CLB) during packing. Path syntax:
+			``pb_type[index]`` at each level, levels separated by ``.``, and ``{mode_name}`` after a level when
+			that mode must be chosen (names come from the architecture file).
+
+			**Example:** ``logical_block_location="clb[0].fle[0]{n1_lut4}.ble4[0].ff[0]"`` places the atom in
```
tangxifan: [constrain_atom_block_src_slide.pptx](https://github.com/user-attachments/files/28567846/constrain_atom_block_src_slide.pptx)

Please add figure to better illustrate the idea. You may reuse the figure in the attached slide.

vaughnbetz:
You can use 1x1 regions (and a subtile specification too if you wish) in the vpr_constraints file to lock down I/Os (or anything else) to a specific location. Going forward, that is the preferred approach, as it constraints both the packer and placer, which is more robust than making a .place file that assumes a certain packing. For I/Os the packing is often trivial (one I/O per cluster), in which case the .place file will work too, but even then using vpr_constraints unifies everything.

vtr_flow/benchmarks/microbenchmarks/and2_latch.blif
```diff
@@ -0,0 +1,8 @@
+.model and2_latch
```
tangxifan, line 1: Can you move the blif file to ``vtr_flow/benchmarks/microbenchmarks``?

This is where all the benchmarks are placed.

vtr_flow/tasks/regression_tests/vtr_reg_strong/task_list.txt
```diff
@@ -137,6 +137,7 @@ regression_tests/vtr_reg_strong/strong_direct_super_subtile_neg_z_offset
 regression_tests/vtr_reg_strong/strong_noc
 regression_tests/vtr_reg_strong/strong_flat_router
 regression_tests/vtr_reg_strong/strong_routing_constraints
+regression_tests/vtr_reg_strong/strong_constraint_atom
```
tangxifan, line 140: Please add another testcase based on the architecture file in  

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/timing/k4_frac_N4_fracff_localClkGen_40nm.xml

You may force to map the LUT in the ``n2_lut3`` mode and ``n1_lut4`` mode as two separated testcases.
Also, try to map FF to the two modes.
I suggest you to try 

- lut in ``n2_lut3`` mode of fle[0], ff in ``n1_lut4`` mode of fle[1]
- - lut in ``n1_lut4`` mode of fle[2], ff in ``n2_lut3`` mode of fle[3]

tangxifan (commented):
@treelin611 See detailed comments.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/vpr_constraint.xml
AlexandreSinger: Sounds good @tangxifan , I forgot that this architecture only has one IO primitive per IO tile. I was mainly thinking about the Stratix-IV and Stratix-10 architecture which have more interesting IO tiles. But if the testcase can be improved to more modes that will help!

treelin611:
Attached is the complete comparison summary between the master branch (`run002`) and the constrain_atom branch (`run001`). 
[comparison_summary.zip](https://github.com/user-attachments/files/28593913/comparison_summary.zip)

haydar-c (approved):
I haven no further comment beyond the above ones regarding the changes. The vtr benchmark runtime seems to be increased by around 2.6% but this might be a bit noisy.  Only suggestion might be to report the Titan benchmark runtime results as well to be documented.

P.S. The vtr top-8 benchmark set (which is generally less noisy) appears to show no runtime degradation (1.00) based on the results you posted.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/config/config.txt
AlexandreSinger, line 1: This and the other tasks are missing golden results. You can see our developer docs to see how to generate these automatically. When running the task, you can add "-create_golden" to generate the golden results.

AlexandreSinger (changes requested):
This is looking good to me! We are just missing some golden results in the tasks.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/config/config.txt
AlexandreSinger: image

vpr/src/base/user_place_constraints.h
```diff
@@ -129,6 +129,26 @@ class UserPlaceConstraints {
...
+     * @brief Associate an atom with its logical_block_location constraint.
+     *
+     * The path is parsed and validated against the architecture pb graphs when set.
+     * The cluster legalizer later matches this location against candidate pb hierarchies.
+     */
+    void set_atom_logical_block_location(AtomBlockId blk_id, const std::string& logical_block_location);
+
+    /**
+     * @brief Return an atom's logical_block_location constraint, if one exists.
+     *
```
vaughnbetz, line 142: I think you should briefly describe how the logical_block_location is encoded in a string (an example would be helpful too).

vpr/src/base/user_place_constraints.h
```diff
@@ -140,6 +160,11 @@ class UserPlaceConstraints {
      */
     std::unordered_map<AtomBlockId, PartitionId> constrained_atoms;
 
+    /**
+     * Store optional per-atom logical_block_location strings.
+     */
```
vaughnbetz, line 173: Maybe this is the best place to define the format of the string, and give an example.

vpr/src/base/vpr_constraints_serializer.h
```diff
@@ -146,6 +147,14 @@ class VprConstraintsSerializer final : public uxsd::VprConstraintsBase<VprConstr
         name_pattern_ = name_pattern;
     }
 
+    virtual inline const char* get_add_atom_logical_block_location(AtomBlockId& /*blk_id*/) final {
```
vaughnbetz, line 150: Any reason to use C strings here instead of C++ strings?  I generally prefer C++ strings for memory allocation safety and ease of use.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/vpr_constraint.xml
```diff
@@ -0,0 +1,12 @@
+<vpr_constraints tool_name="vpr">
+   <partition_list>
+       <partition name="fix_pb_mapping_loc_d">
+            <add_atom name_pattern="d" logical_block_location="clb[0].fle[0]{n1_lut4}.ble4[0].ff[0]"/>
+            <add_region x_low="2" y_low="1" x_high="2" y_high="1"/>
+       </partition>
+       <partition name="fix_pb_mapping_loc_c">
+            <add_atom name_pattern="c" logical_block_location="clb[0].fle[3]{n1_lut4}.ble4[0].lut4[0]"/>
```
vaughnbetz, line 8: I think we can also use this logical_block_location with no region (or, if the parser requires a region we could use a really big region).  It would be good to add a test case for that case.

Allowing the add_region to be optional if there is a logical_block_location would also be a good upgrade (lets you control how things are packed while leaving their placement unspecified) but if that's too much work it's OK.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -39,7 +39,52 @@
...
+ * mode). LbHierPathParser parses the constraint with '.' separators and '{}' for mode,
+ * parses hierarchical_type_name() with '/' separators and '[]' for mode, then compares
+ * token-by-token (see matches_hierarchical_type()).
+ *
+ * Example (atom 'd' packing into a CLB), logged when verbosity > 3:
+ * - expected (constraint):  `clb[0].fle[0]{n1_lut4}.ble4[0].ff[0]`
+ * - candidate (pb path):    `clb[0][default]/fle[3][n1_lut4]/ble4[0][default]/ff[0]`
+ *   -> fails because fle index 3 != constrained fle index 0 (same for fle[2], fle[1]).
+ * - a passing candidate must match every specified index/mode, e.g. fle[0] and {n1_lut4}.
+ *
```
vaughnbetz, line 63: I think you should also explain 'default'. Are those one mode hierarchy levels, so the 'default' word can be omitted?

vpr/src/base/user_place_constraints.h
vaughnbetz: Could alternatively point the reader at the example in cluster_legalizer.cpp, but I think directly giving an example of the format here is best.

vpr/src/pack/logic_block_location_util.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include <vector>
+
+#include "atom_netlist_fwd.h"
+#include "physical_types.h"
+
+struct t_logical_location_token {
```
vaughnbetz, line 20: Should have a Doxygen comment for what the fields mean and what the struct is for.

vpr/src/pack/logic_block_location_util.h
```diff
@@ -0,0 +1,62 @@
...
+     */
+    const t_pb_graph_node* find_matching_pb_graph_node(const t_pb_graph_node* root) const;
+
+    /**
+     * @brief Validate that this logical_block_location matches at least one pb graph
+     *        among the provided logical_block_types.
+     */
+    void validate_logical_block_types(const std::vector<t_logical_block_type>& logical_block_types);
+
+    enum class TokenFormat {
```
vaughnbetz, line 62: Add Doxygen comment.

vpr/src/pack/logic_block_location_util.cpp
```diff
@@ -0,0 +1,239 @@
...
+ * @brief Parse one logical location/hierarchical-type token.
+ *
+ * Depending on @p format, bracket fields are interpreted as either
+ * logical-block index/mode constraints or hierarchical type annotations.
+ *
+ * Example full paths:
+ * - `TokenFormat::LOGICAL_LOCATION`:
+ *   `clb[0].fle[0]{n1_lut4}.ble4[0].ff[0]`
+ * - `TokenFormat::HIERARCHICAL_TYPE`:
+ *   `clb[0][default]/fle[3][n1_lut4]/ble4[0][default]/ff[0]`
```
vaughnbetz, line 43: Again, explain meaning of default.

vaughnbetz (changes requested):
This looks very good; just suggesting some commenting changes.
If it's not too much work, you could also consider allowing this constraint to be specified without a region, as it should work in that case too (and would be useful in that case, as it lets you control the packing).

vpr/src/base/vpr_constraints_serializer.h
treelin611: The return type is fixed by the uxsdcxx-generated VprConstraintsBase interface. The same pattern is used throughout this file. The following screenshot shows another example: 
image

vpr/src/base/user_place_constraints.h
treelin611: Thanks for your comment. I've just added documentation on the logical_block_location string format at line 164 in user_place_constraints.h.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_constraint_atom/vpr_constraint.xml
treelin611: Thank you for the thoughtful comment. I've addressed this by updating the verification logic and adding a regression test,` strong_constraint_atom_pack_only`, which confirms that logical_block_location is still enforced during packing even when no add_region is specified.

treelin611:
Please see the attached QoR results: run001 corresponds to the constrain_atom branch, and run002 to master.
[compare_result.zip](https://github.com/user-attachments/files/28701347/compare_result.zip)

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -57,6 +57,8 @@
  * Example (atom 'd' packing into a CLB), logged when verbosity > 3:
  * - expected (constraint):  `clb[0].fle[0]{n1_lut4}.ble4[0].ff[0]`
  * - candidate (pb path):    `clb[0][default]/fle[3][n1_lut4]/ble4[0][default]/ff[0]`
+ *   `[default]` is the implicit pb_mode at levels with no explicit architecture mode; constraints
+ *   need not write `{default}` there—omit `{...}` to match any mode, including default.
```
vaughnbetz: there-omit --> they can omit (?)

vaughnbetz (approved):
One minor typo / confusion in the documentation remains. Other than that it looks good to me to merge once the conflicts are resolved.

vaughnbetz:
@tangxifan @AlexandreSinger : does this address all your comments now?

vpr/src/pack/cluster_legalizer.cpp
treelin611: Yes. Thanks for your comment. Typo fixed.

tangxifan:
> @tangxifan @AlexandreSinger : does this address all your comments now?

Yes. I suggest to merge. If you see any issue, we are happy to address it.

AlexandreSinger (approved):
LGTM

AlexandreSinger:
It sounds like everyone is ok to merge. Merging.

vaughnbetz:
@treelin611 : Thanks, this looks like a very useful new feature.

Since I believe the vpr constraints now allow a packing within cluster constraint to be specified without a region (the region is optional if a packing within cluster constraint is specified) the documentation should be updated to note that (but first I'd test to make sure it really goes all the way through the parsers etc. if there is no region).

PR #3503, opened 2026-04-28
Add fc check for intra-cluster routing

amin1377:
In this PR, the intra-cluster router is updated to accommodate the following case:

Assume we have a connection whose source and sink are both in the same cluster. Due to limitations in the routing resources inside the cluster, the connection may need to be routed outside the cluster and then back into the cluster, i.e., as a feedback connection using general routing resources.

However, the intra-cluster router currently does not check whether the top-level output pin it routes the connection to is connected to routing wires, i.e., whether `fc > 0`. In this PR, we add this check to ensure the connection is not routed to a pin without any connection to routing wires.


libs/libarchfpga/src/physical_types.h
```diff
@@ -1336,12 +1336,36 @@ class t_pb_graph_node {
...
+     * pins). Pins with Fc_out == 0 cannot drive any wire and so cannot
+     * physically carry a signal that exits this cluster and returns to it.
+     *
+     * Only meaningful on the pb_graph_head of a logical block type; for
+     * non-root pb_graph_nodes this set is left empty. Populated once per
+     * type by populate_valid_feedback_pins() (declared in vpr_utils.h)
+     * during architecture setup. Queried by the intra-cluster router via
+     * is_valid_feedback_pin().
+     */
+    std::unordered_set<int> valid_feedback_pin_indices;
```
AlexandreSinger: Why add this to the t_pb_graph_node data structure? This information is specific to the cluster router. I think this would be better placed in either the packer or the cluster router. I find that this data structure is overfilled with algorithm-specific information, when it is supposed to just be a graph datastructure.

Eventually we would want to give unique IDs to logical block types, but you can create a vector that takes a logical block type and returns this set of valid feedback pins. This datastructure would be cleaner and would not create an unordered set for every pb_graph_node.

vpr/src/base/setup_vpr.cpp
```diff
@@ -354,6 +354,16 @@ void SetupVPR(const t_options* options,
...
 
+    /* Populate, once per logical block type, the set of top-level cluster
+     * output pins that may be used as the "out" leg of a feedback path
+     * (driver and sink both inside the same cluster, signal goes out
+     * through a top-level output pin and re-enters through a top-level
+     * input pin). The intra-cluster router queries this set via
+     * pb_graph_head->is_valid_feedback_pin() during edge expansion. */
+    for (t_logical_block_type& lb_type : device_ctx.logical_block_types) {
+        populate_valid_feedback_pins(lb_type);
+    }
```
AlexandreSinger: See my comment above. If performance is a concern, we may consider putting this in the global structures for packing (since this data has the same lifetime as the pb_graph), but I would ask that we please avoid that if we can.

vpr/src/util/vpr_utils.cpp
```diff
@@ -374,6 +374,58 @@ bool is_empty_type(t_logical_block_type_ptr type) {
...
+    for (int iport = 0; iport < pb_graph_head->num_output_ports; iport++) {
+        for (int ipin = 0; ipin < pb_graph_head->num_output_pins[iport]; ipin++) {
+            const t_pb_graph_pin* pb_pin = &pb_graph_head->output_pins[iport][ipin];
+            int physical_pin_num = get_physical_pin(tile_type, &lb_type, pb_pin->pin_count_in_cluster);
+            if (non_zero_fc_physical_pins.count(physical_pin_num)) {
+                valid_set.insert(pb_pin->pin_count_in_cluster);
+            }
+        }
+    }
+}
```
AlexandreSinger: If you agree that we should move the data structure, I would really like to move this out of vpr_utils. I actually really do not like this file, its a collection of a bunch of disjoint functions. I think this belongs somewhere in the packer.

AlexandreSinger (changes requested):
Hi @amin1377 ,

Thanks for the changes! I can see why this may be useful. I did not look too deeply into the implementation yet, but I had some structural issues that should be addressed.

I have been trying to clean up the pb_graph (it has been put on a bit of a hold), but I have been trying to remove algorithm-specific code from the data structure. We should keep this out and put it in the packer or router.

AmirhosseinPoolad (commented):
I really think this should be a boolean field in t_pb_graph_pin. Also I'm guessing the use case here is for direct connections since it wouldn't really make sense to have an output pin with Fc = 0 otherwise. How did VPR handle those before and why doesn't that approach work for you?

libs/libarchfpga/src/physical_types.h
amin1377: I agree. I moved the initialization of valid_feedback_pins_by_type_ to the cluster legalizer.

vpr/src/util/vpr_utils.cpp
amin1377: Done

amin1377:
@AmirhosseinPoolad: As discussed before, the reason I did not do that is that I think it would be cleaner to keep only the list of top-level pins with `fc > 0`.
What you are suggesting would require a fairly large data structure where the value for most pins would be zero, since most of them are inside logical blocks.

amin1377:
@AlexandreSinger: I think this PR is ready to be merged. Please let me know if you have any other comments.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1765,6 +1768,37 @@ ClusterLegalizer::ClusterLegalizer(const AtomNetlist& atom_netlist,
...
+        t_physical_tile_type_ptr tile = lb_type.equivalent_tiles[0];
+        for (int iport = 0; iport < pb_head->num_output_ports; iport++) {
+            for (int ipin = 0; ipin < pb_head->num_output_pins[iport]; ipin++) {
+                const t_pb_graph_pin* pb_pin = &pb_head->output_pins[iport][ipin];
+                int phys_pin = get_physical_pin(tile, &lb_type, pb_pin->pin_count_in_cluster);
+                if (non_zero_fc_pins.count(phys_pin)) {
+                    valid_set.insert(pb_pin->pin_count_in_cluster);
+                }
+            }
+        }
```
AlexandreSinger, line 1806: I want to keep this constructor clean if we can. Can you turn this into a helper function? I think it will make this code much cleaner and easier to follow.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1537,7 +1539,8 @@ ClusterLegalizer::start_new_cluster(PackMoleculeId molecule_id,
     const AtomNetlist& atom_nlist = g_vpr_ctx.atom().netlist();
 
     // Create the new cluster
-    LegalizationCluster new_cluster(cluster_type, cluster_mode, lb_type_rr_graphs_);
+    LegalizationCluster new_cluster(cluster_type, cluster_mode, lb_type_rr_graphs_,
+                                    valid_feedback_pins_by_type_.at(cluster_type));
```
AlexandreSinger: This can be a bit confusing if the cluster type cannot be found in the map for any reason. I think we should use an explicit assert here instead of rely on `at`.

vpr/src/pack/cluster_legalizer.h
```diff
@@ -624,6 +627,10 @@ class ClusterLegalizer {
     ///       meant to be a vector of vectors...
     std::vector<t_lb_type_rr_node>* lb_type_rr_graphs_ = nullptr;
 
+    /// @brief Per-type set of top-level output pin indices with Fc_out > 0.
+    ///        Built once at packing start; each ClusterRouter holds a pointer into this map.
+    std::unordered_map<const t_logical_block_type*, std::unordered_set<int>> valid_feedback_pins_by_type_;
```
AlexandreSinger: There is a t_logical_block_type_ptr typedef that would make this a bit cleaner; however, I recommend using a vector instead. Each logical block type has a unique ID. If it is easy, I recommend using that instead of hashing on the pointer.

vpr/src/pack/cluster_router.cpp
```diff
@@ -1082,7 +1124,8 @@ void ClusterRouter::expand_node_all_modes_(const t_expansion_node& exp_node, int
             continue;
         }
 
-        expand_edges_(mode, cur_inode, cur_cost, net_fanout);
+        /* Forward target_is_internal_sink for the same reason as expand_node_. */
```
AlexandreSinger: The comments over some of these functions seem strange. It only labels one of the arguments going into these methods. I think it would be better if they were removed. See the other expand edges call.

AlexandreSinger (commented):
Hi @amin1377 

Just some final comments and then I think this will be good to merge.

vpr/src/pack/cluster_legalizer.cpp
amin1377: I applied your comment and added a helper function to initialize this data structure. However, I am not quite clear on the motivation behind this change.
This initialization is something that should be done when the class is instantiated, and by definition, it seems to belong in the constructor. I am just curious to understand your reasoning behind this.

vpr/src/pack/cluster_router.cpp
AmirhosseinPoolad: // for non doxygen comments please!

vpr/src/pack/cluster_legalizer.cpp
AlexandreSinger: This is the constructor for the entire cluster legalizer class. As we continue to work on it, it will continue to get longer and longer as people add "just one more thing". The other lines in the constructor all use helper methods as well.

AlexandreSinger: A better way of doing this is having a class to manage these pin sets. So something like "ClusterFeedbackPinSet".

The constructor for this will populate the data structures internally and then the cluster router would use APIs to get the information it needs. I find this pattern to be the cleanest for something like this; but its not necessary. Helper methods should be clean enough.

vpr/src/analytical_place/full_legalizer.cpp
```diff
@@ -1058,6 +1059,7 @@ void NaiveFullLegalizer::create_clusters(const PartialPlacement& p_placement) {
                                        vpr_setup_.PackerOpts.memoize_cluster_packings,
                                        arch_.models,
                                        vpr_setup_.PackerOpts.pack_verbosity);
+    cluster_legalizer.init_feedback_pin_sets();
```
AlexandreSinger: Oh, I see what happened here. This is not what I meant by helper method.

In the constructor of the cluster legalizer, we should have a function call which initializes the feedback pin sets. I am sorry for the confusion!

vpr/src/pack/pack.cpp
```diff
@@ -324,6 +324,7 @@ bool try_pack(const t_packer_opts& packer_opts,
                                        packer_opts.memoize_cluster_packings,
                                        arch.models,
                                        packer_opts.pack_verbosity);
+    cluster_legalizer.init_feedback_pin_sets();
```
AlexandreSinger: Yeah, we really want to avoid this if we can. This will make the cluster legalizer way harder to work with since people need to remember to call this random function at all times.

AlexandreSinger (commented):
I am very sorry for the confusion Amin. This is my last comment. After this it can be merged.

vpr/src/analytical_place/full_legalizer.cpp
amin1377: Ah, I see. That makes sense.

AlexandreSinger (approved):
LGTM

PR #3504, opened 2026-04-28
CRR Switch ID

amin1377:
In this PR, instead of assigning a unique switch ID for each row-column cell of each CSV pattern(which can easily overflow the switch ID type, `short`, when the number of patterns or channel width exceeds a certain threshold) we get the minimum and maximum delay values specified in the switch pattern files and create new switch types within that range.

The reason we do not create new switches only for the unique delay values is to avoid generating a significantly different RR graph file when there is only a simple change in switch delay values.


vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_edge_builder.cpp
```diff
@@ -40,50 +39,45 @@ static t_arch_switch_inf create_crr_switch(const int delay_ps, const std::string
...
-    int found_sw_id = -1;
-
-    // Iterate over map entries (O(n), no accidental inserts)
-    for (const auto& [sw_id, sw_inf] : all_sw_inf) {
-        if (sw_inf.template_id == sw_template_id) {
-            found_sw_id = sw_id;
-            break;
-        }
+    if (min_delay_ps > max_delay_ps) {
+        return delay_to_switch_id;
```
AmirhosseinPoolad, line 56: Shouldn't this be an assertion?

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_edge_builder.cpp
```diff
@@ -40,50 +39,45 @@ static t_arch_switch_inf create_crr_switch(const int delay_ps, const std::string
...
-    if (found_sw_id == -1) {
-        t_arch_switch_inf new_arch_switch_inf = create_crr_switch(delay_ps, sw_template_id);
-
-        found_sw_id = static_cast<int>(all_sw_inf.size());
-        all_sw_inf.emplace(found_sw_id, std::move(new_arch_switch_inf));
+    delay_to_switch_id.reserve(static_cast<size_t>(max_delay_ps - min_delay_ps + 1));
+    for (int delay_ps = min_delay_ps; delay_ps <= max_delay_ps; ++delay_ps) {
+        int sw_id = static_cast<int>(all_sw_inf.size());
+        all_sw_inf.emplace(sw_id, create_crr_switch(delay_ps));
+        delay_to_switch_id.emplace(delay_ps, RRSwitchId(sw_id));
```
AmirhosseinPoolad, line 63: Why do you need to create switches for the entire range? What if I have two delays, one 30ps another 300ps?

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_edge_builder.cpp
```diff
@@ -95,20 +89,15 @@ void build_crr_gsb_edges(RRGraphBuilder& rr_graph_builder,
...
-            } else {
-                rr_switch_id = find_or_create_crr_switch_id(delay_ps,
-                                                            sw_template_id,
-                                                            verbosity);
-                template_id_cache[sw_template_id] = int(rr_switch_id);
-            }
+            auto it = delay_to_switch_id.find(delay_ps);
+            VTR_ASSERT_MSG(it != delay_to_switch_id.end(),
+                           "CRR connection delay was not pre-created in delay_to_switch_id map");
+            rr_switch_id = it->second;
```
AmirhosseinPoolad, line 103: Honestly you might as well not do the pre-creation and just create it here.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_edge_builder.cpp
amin1377: No, not really. If the delay range is invalid, it means the CSV pattern files are not annotated with delay number and only contain connectivity information. I added a comment to explain that.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_edge_builder.cpp
amin1377: I explained this in person, but just to write it down as well: we create switches for the whole range so that if the delay values for a few switches change in the future, the overall RR graph file structure does not change, including the switch list and the switch IDs assigned to each edge.

amin1377:
@AmirhosseinPoolad: Thanks for taking the time to review this PR. I addressed your concerns. Please let me know if anything else is required.

amin1377:
@AmirhosseinPoolad: I do not think the build failure is caused by the changes in this PR.

AmirhosseinPoolad:
Yeah, windows CI is broken.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_switch_block_manager.h
```diff
@@ -114,6 +132,15 @@ class SwitchBlockManager {
...
 
+    /**
+     * @brief Min/max switch delay (ps) accumulated across every loaded
+     *        template file when annotated_rr_graph is enabled.
+     *        Initialized to INT_MAX/INT_MIN so the first observed value
+     *        always wins.
+     */
+    int switch_delay_min_ps_ = INT_MAX;
+    int switch_delay_max_ps_ = INT_MIN;
+
```
AmirhosseinPoolad, line 143: Use std::numeric_limits instead. also remove #include \<climits\> after that.

AmirhosseinPoolad (commented):
One more comment (sorry) and then LGTM.

PR #3509, opened 2026-05-01
[CI] Updated Failing Weekly QoR Tests

AlexandreSinger:
The weekly tests have been down for the last month or so. Since then, the weekly tests now have some failing QoR tests. Most of them are caused by recent changes in the packer, where we are better selecting which RAM types to use. This had an overall positive change on the results. QoR was not changed, the major changes are device sizes.

There are two potential errors that need to be investigated.

vpr_tight_floorplan has an error that needs to be investigated.

ap_titan has an issue with the file parsing that needs to be investigated and fixed.

These two errors should be fixed in a separate PR.


AmirhosseinPoolad:
Looks good. Started another run of the tests on this branch to make sure nothing is left out.

AlexandreSinger:
Verified the weekly tests and found all QoRs updated. The only things left failing are the two tests I mentioned above:

- vtr_reg_nightly_test5/vpr_tight_floorplan
- vtr_reg_nightly_test7/ap_titan

I will merge this PR in so the QoR results are up to date.

PR #3511, opened 2026-05-01
Flatten pin feasibility function in packer

AmirhosseinPoolad:
This PR refactors try_update_lookahead_pins_used to not do any recursion. Instead of traversing the t_pb tree, we now go through the atoms in the molecule and get the primitive/leaf t_pb from AtomPBBimap directly. This makes the code cleaner and also the less deep stack frame should make profiling the packer easier.

### DIFF
--- a/vpr/src/pack/cluster_legalizer.cpp
+++ b/vpr/src/pack/cluster_legalizer.cpp
@@ -838,27 +838,26 @@ static void compute_and_mark_lookahead_pins_used(const AtomBlockId blk_id,
  * Runtime is actually not that bad for this.  It's worst case O(k^2) where k is the
  * number of pb_graph pins.  Can use hash tables or make incremental if becomes an issue.
  */
-static void try_update_lookahead_pins_used(t_pb* cur_pb,
+static void try_update_lookahead_pins_used(const LegalizationCluster& cluster,
+                                           const Prepacker& prepacker,
                                            const vtr::vector_map<AtomBlockId, LegalizationClusterId>& atom_cluster,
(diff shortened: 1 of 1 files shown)


AmirhosseinPoolad:
From my limited tests, QoR is unaffected. TODO is running Titan and Koios.

AmirhosseinPoolad:
Titan QoR results:

  | master_titan | flat_titan
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.998
max_vpr_mem | 1 | 1
pack_time | 1 | 0.987
place_time | 1 | 1.005
crit_path_route_time | 1 | 0.992
routed_wirelength | 1 | 1
critical_path_delay | 1 | 1

CPD and WL are identical (expected), and runtime is not meaningfully faster or slower.

vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -838,27 +838,26 @@ static void compute_and_mark_lookahead_pins_used(const AtomBlockId blk_id,
  * Runtime is actually not that bad for this.  It's worst case O(k^2) where k is the
  * number of pb_graph pins.  Can use hash tables or make incremental if becomes an issue.
  */
-static void try_update_lookahead_pins_used(t_pb* cur_pb,
+static void try_update_lookahead_pins_used(const LegalizationCluster& cluster,
```
amin1377, line 856: I think the Doxygen comment should be expanded a little bit.

amin1377 (approved):
Thanks, Amir!

PR #3512, opened 2026-05-01
Add net file constraint support to tasks configs

AmirhosseinPoolad:
Adds the ability to input a per-circuit .net file to the vtr task config. Useful for when you're running experiments that wouldn't affect packing and only place and route matters.


AlexandreSinger (approved):
LGTM

I am happy this interface is useful for other people beyond AP!

Issue #3076, opened 2025-05-22
set of windows (mingw64) build fixes


w0lek:
group of fixes to finalize build on windows (mingw64)

PR #3513, opened 2026-05-02
Add msys2 build CI workflow

tangxifan:
#### Description

Add a Msys2 build workflow to Github CI to validate the build compatibility of VTR project on Windows.

#### Related Issue

#3076 #3432 

#### Motivation and Context

As stated in #3432

- [x] A CI job is added where the VTR code base is built through Msys2 MINGW64 enviornment on a Windows machine.
- [x] Push button script to install all the required dependencies on Msys2 Mingw64 enviornment
- [x] CMake script changes to adapt Msys2 flags  

Know limitations:

The codes of ABC and Yosys (the version VTR is using and also the latest) do not support Msys2. Dirty patch can be applied but is not scalable in future. 

This results in

- Parmys is not included in the Msys2 build and Yosys+VPR flow is disabled. User can use the [prebuilt version of Yosys](https://packages.msys2.org/packages/mingw-w64-x86_64-yosys) in Msys2 but Parmys is not available.
- ABC is not included in the Msys2 build and power estimation flow is disabled.
- All the basic reg tests require Yosys or a Verilog frontend. Before Parmys is enabled, no basic reg tests can be enabled on CI.
- TBB is disabled as it includes <windows.h>, which results in build errors.

#### Types of changes

#### Checklist:


tangxifan:
@vaughnbetz This PR is ready for your review.

.github/scripts/build_dependency_win-msys2-MINGW64.sh
AlexandreSinger, line 1: I think this file is extremely useful to expose beyond the CI! We should provide this to users similar to the Linux install scripts.

Can we move this file to the root of the VTR directory and name it something like: `install_win_msys2_packages.sh`

.github/scripts/build_dependency_win-msys2-MINGW64.sh
```diff
@@ -0,0 +1,76 @@
...
+    mingw-w64-x86_64-lld \
+    mingw-w64-x86_64-clang \
+    mingw-w64-x86_64-yosys \
+    mingw-w64-x86_64-boost-libs \
+    libreadline-devel \
+    zlib-devel \
+    openssl-devel \
+    mingw-w64-x86_64-zlib \
+    mingw-w64-x86_64-graphviz \
+    mingw-w64-x86_64-openssl
```
AlexandreSinger, line 62: We are missing two packages in this list:

- eigen3
- tbb

Without these libraries, I am seeing the following CMAKE warnings.
Image

Without eigen3, AP will not work properly.

Without TBB, Tatum and the router cannot be run in parallel using the default algorithms.

If it is easy, could we add those libraries to this list? If these cause issues, it should not gate this PR; but we should investigate adding these packages in the near future.

.github/workflows/test.yml
```diff
@@ -775,3 +775,99 @@ jobs:
         export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
         make -j${{ steps.cpu-cores.outputs.count}}
 
+  Msys2Compatibility:
+    needs: [BuildVTR]
+    name: ${{ matrix.config.name }}
+    runs-on: windows-latest
+    # Note: dependencies are installed in the container. See details about dependency list in docker/Dockerfile.master
+    # Comment the line out when base image is built again
+    #container: ghcr.io/${{ needs.change_detect.outputs.docker_repo }}/openfpga-build-${{ matrix.config.cc}}
```
AlexandreSinger: Are these comments relevant to VTR? These seem to be OpenFPGA specific. Eventually it would be a good idea to build all of our dependencies into docker containers to smooth-out the CI; but the rest of the tests do not do this yet. More CI infrastructure is needed on VTR to support this.

.github/workflows/test.yml
```diff
@@ -775,3 +775,99 @@ jobs:
         export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
         make -j${{ steps.cpu-cores.outputs.count}}
 
+  Msys2Compatibility:
+    needs: [BuildVTR]
+    name: ${{ matrix.config.name }}
+    runs-on: windows-latest
```
AlexandreSinger: Instead of locking to `windows-latest`, can we lock it to `windows-2025`.

This is equivalent to `windows-latest`: https://github.com/actions/runner-images

The reasoning is that I would prefer us to lock ourselves to a specific version of windows so the CI does not crash randomly in the future when the windows runner image is updated.

.github/workflows/test.yml
```diff
@@ -775,3 +775,99 @@ jobs:
...
+    # Branch on different OS and settings
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+#          - name: "Build Compatibility (GCC, Windows Msys2-UCRT64)"
+#            cc: gcc
+#            cxx: g++
+#            os: UCRT64
+#            prefix: ucrt64
```
AlexandreSinger, line 792: Why was this commented out? If there is a reason, I do not think it should block this PR but we should have a comment explaining why its not being run.

.github/workflows/test.yml
```diff
@@ -775,3 +775,99 @@ jobs:
...
+        run: |
+          ./.github/scripts/build_dependency_win-msys2-${{ matrix.config.os}}.sh
+          cp /usr/include/FlexLexer.h /${{ matrix.config.prefix}}/include/
+          python3 -m venv myenv
+          source myenv/bin/activate
+          python3 -m pip install --upgrade pip
+          #python3 -m pip install -r requirements.txt
+          #echo "==== Checking python packages ===="
+          #python -c "import yaml; print('Installed')"      
+          #pip install numpy cython setuptools>=60.0.0
```
AlexandreSinger: These lines are commented out. Is this intended?

libs/EXTERNAL/libcatch2
AlexandreSinger, line 1: I think this may have been a mistake. LibCatch2 was updated on master a couple of days ago, I think you just need to run `git submodule update` to pull in these changes. This PR appears to be undoing the update by mistake.

AlexandreSinger (changes requested):
Hi @tangxifan this looks good!

I have left some comments below. Do note that this CI test that you added is only checking that VPR is built in windows, it is not actually running it. I recognize why this is the case, since it looks like Parmys cannot be built on Windows yet; but this is something we should add in the future.

.github/scripts/build_dependency_win-msys2-MINGW64.sh
tangxifan: Sure thing

.github/scripts/build_dependency_win-msys2-MINGW64.sh
tangxifan: @AlexandreSinger TBB has issues in build as stated in the description of this PR. Further investigation is required. I believe the problem is on the STA codes, which is not yet well ported to Windows platform when TBB is included.
I added the eigen3 to the dependency list now. See if the CI can pass or not.

.github/workflows/test.yml
tangxifan: Sure thing.

.github/workflows/test.yml
tangxifan: UCRT64 is another environment supported by Msys2. It has a different set of package names than the MINGW64 but the same as foundational libraries. If MINGW64 can pass, UCRT64 can pass mostly.
It is commented out because I do not want to increase CI runtime for now. If you think it is necessary, I can enable it.

.github/workflows/test.yml
tangxifan: I enable it now and see if it works. Msys2 is using the latest python 3.14 which is not yet stable for all the packages to be installed. It it works, I will keep it. If not, I will comment and disable it.

tangxifan:
@AlexandreSinger All your comments have been addressed. Please review and see if you are satisfied.

AlexandreSinger (approved):
LGTM, thanks @tangxifan !

It looks like the version of Eigen in the msys2 build is not compatible with our version of Eigen that we expect. I do not think that we should concern ourselves with that just yet. I am thinking of moving away from Eigen (or providing an alternative) soon since I do not like the dependency and it will only get more complicated.

Vaughn is away this week, but I think we can merge this PR now. I think it will be useful to lock in the msys build ASAP. There are a lot of warnings in the build, but those can be resolved in time.

I will continue to monitor the CI to ensure that the caching is working correctly. I want to try and keep the overall CI time down if we can.

Issue #3189, opened 2025-07-05
Exploration of Using a Different Memory Allocator for VTR Runtime Improvement


stefanpie:
Note, this is an independent exploration that I conducted, and I don't expect that VTR authors to immediately consider these changes in VTR. I merely want to provide some useful insight into VTR I gained from my own research to the authors of VTR to help make VTR better.
## Motivation

Recently, I've been looking at ways to speed up EDA tools without the need to make changes to the underlying code of the EDA tools themselves. For example, I want to try to make Vitis HLS or Vivado run faster, but I don't have access to the Vitis HLS or Vivado source code, and I can't change it unless I go work or intern for AMD/Xilinx. I want to explore approaches that would let me speed up commercial EDA tools in a "black box" manner. My original intent with this idea was that I could speed up EDA tools that interact with LLM agents, since the bottleneck there is the EDA tools (which can take >30s to minutes) and not LLM inference (which is usually <20s). But this is also a useful pursuit in general for large-scale EDA dataset collection and EDA benchmarking.

## Approach

One trick in this exploration was to use a different memory allocator than the default malloc implementation that was compiled or linked with the software. For my exploration, I was testing the use of mimalloc (https://github.com/microsoft/mimalloc), a custom memory allocator from Microsoft Research, which has many tunable and customizable parameters. Even better, you can override the default allocator used by any software that dynamically links malloc using LD_PRELOAD to swap in mimalloc instead.

The general idea here is that (assuming our system has lots of memory resources) we can tune the memory allocator to take advantage of memory resources to trade off faster program execution for a small amount of memory overhead. This can hopefully result in non-trivial speedups in runtime over many program executions. For example, if I have to place and route 1000 benchmarks and I can get a 0.9x-0.8x reduction in runtime with a better memory allocator, my benchmark evaluation can take 9-8 hours instead of 10 hours. However, this model of what I would expect speedups to be is not that realistic. Program execution and memory allocation in a tool may be complex. Running the program on small designs might yield a 0.99x reduction (virtually nothing) while running the program on larger designs might yield a 0.8x reduction. Some programs or phases of program execution (synth vs. pack vs. place vs. route) might also see different behaviors and runtime reductions as well.

I decided to also try to evaluate this approach on open-source EDA software like yosys (still testing) and VTR, and I was able to capture some simple initial results for VTR just by swapping in mimalloc with minimal tweaking of the allocator configuration.

## Evaluation

[...]

PR #3514, opened 2026-05-03
Use jemalloc where available for ~6% performance improvement

Majora320:
#### Description
jemalloc is an alternative allocator that provides better performance than the default glibc malloc in most situations, especially for memory-heavy applications. This PR uses jemalloc by default when available and adds it to the default install lists.

#### Motivation and Context
vtr is slow. Anything that shaves a few seconds off each iteration is useful.

I tested jemalloc locally on a 7800X3D along with tcmalloc and micalloc, two other popular allocator options. The following command was used:

```sh
/usr/bin/time -v vpr ../arch/titan/stratixiv_arch.timing.xml ./benchmarks/titan_blif/titan23/stratixiv/neuron_stratixiv_arch_timing.blif --sdc_file ./benchmarks/titan_blif/titan23/stratixiv/neuron_stratixiv_arch_timing.sdc --max_router_iterations 400     --router_lookahead map     --initial_pres_fac 1.0     --router_profiler_astar_fac 1.5     --seed 3     --route_chan_width 400 -j8
```

Here is a table of times and relative improvements.
  | malloc version | vpr time (s) |   | vpr max_rss (MiB) |   | user time (measured by `time`) |   | time rss (kbytes) |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
glibc malloc | 2.42 | 123.62 |   | 3160.3 |   | 158.05 |   | 3236160 |  
tcmalloc | 4.6.3 | 117.68 | -4.81% | 3072.8 | -2.77% | 149.14 | -5.64% | 3148120 | -2.72%
mimalloc | 2.1 | 118.06 | -4.50% | 3818.4 | 20.82% | 148.99 | -5.73% | 3910056 | 20.82%
jemalloc | 5.3.0 | 117.26 | -5.14% | 3042.9 | -3.71% | 148.32 | -6.16% | 3115924 | -3.72%

As you can see, jemalloc shaves about 6% off of total execution time and reduces peak memory usage by about 4%.

Please let me know if any documentation changes are required as a result of this change. Thanks :)

#### Types of changes

#### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ option(VTR_ENABLE_DEBUG_LOGGING "Enable debug logging" OFF)
 option(VTR_ENABLE_VERBOSE "Enable increased debug verbosity" OFF)
 option(SPEC_CPU "Enable SPEC CPU v8 support" OFF)
 
+option(VPR_USE_JEMALLOC "Use jemalloc for vpr instead of default glibc allocator when available" ON)
+
 #Allow the user to decide whether to compile the graphics library
 set(VPR_USE_EZGL "auto" CACHE STRING "Specify whether vpr uses the graphics library")
 set_property(CACHE VPR_USE_EZGL PROPERTY STRINGS auto off on)
--- a/install_apt_packages.sh
+++ b/install_apt_packages.sh
@@ -53,6 +53,11 @@ packages_to_install+=(
     zlib1g-dev
 )
 
+# Optional, but improves performance
+packages_to_install+=(
+    libjemalloc-dev
+)
+
 # Required for code formatting
 # NOTE: clang-format-18 may only be found on specific distributions. Only
 #       install it if the distribution has this version of clang format.
--- a/install_dnf_packages.sh
+++ b/install_dnf_packages.sh
@@ -49,3 +49,7 @@ sudo dnf install --refresh -y \
 # Required to run the analytical placement flow
 sudo dnf install --refresh -y \
     eigen3-devel
+
+# Optional, but improves performance
+sudo dnf install --refresh -y \
+    jemalloc-devel
--- a/vpr/CMakeLists.txt
+++ b/vpr/CMakeLists.txt
@@ -175,6 +175,16 @@ add_executable(vpr ${EXEC_SOURCES})
 
 target_link_libraries(vpr libvpr)
 
+# Use jemalloc for ~5% better performance if enabled and we find it
+if(VPR_USE_JEMALLOC)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(JEMALLOC jemalloc)
+
+    if(JEMALLOC_FOUND)
+        target_link_libraries(vpr ${JEMALLOC_LIBRARIES})
+    endif()
+endif()
+
 
 #
 # Profile Guilded Optimization Configuration


AmirhosseinPoolad:
>vtr is slow.

We're trying our best :'( Although I gotta say [the last time anyone checked](https://dl.acm.org/doi/full/10.1145/3734798), VtR was actually ~30% faster than Quartus. FPGA CAD is just that hard.

I personally like the implementation here, not super intrusive and nothing bad happens if jemalloc is not installed on user machines. Approved the CI tests to see if everything works as expected. Can you test the whole Titan benchmark suite to see gains on a more diverse set of circuits? You can run them using:
> vtr_flow/scripts/run_vtr_task.py regression_tests/vtr_reg_nightly_test2/titan_quick_qor -j[cores]

The run_vtr_task script outputs a file in vtr_flow/tasks/regression_tests/vtr_reg_nightly_test2/titan_quick_qor/run[num] called parse_results.txt or something. You can copy that into google sheets to get all the relevant numbers. Please also check memory usage. I don't expect meaningful changes in large circuits, but the new allocator probably gets memory pages from the OS in a different way and I'm seeing some CI failures in smaller tests because of higher memory usage.

They do take a while to run (sorry!) and I would probably not put in -j8 because of memory usage. I think dividing your RAM by 8 would give you a fairly safe number of benchmarks you can run at the same time.

AlexandreSinger:
Looks like one testcase uses more memory and is causing a QoR failure:
image

We put bounds on our QoR metrics to keep noise down, so it is possible that this is overall good for memory usage and runtime, and this one case is a fluke.

I agree with @AmirhosseinPoolad here that we need to see the results on more testcases to bring something like this in.

stefanpie:
I proposed something like this in #3189. I continued down this road a bit further but never finished the project at that time.

I wanted to note that, when expanding testing to different test cases, you should be sure to test different VTR modes. I found large runtime improvements when no routing channel size was specified and VTR searched for it automatically. However, when I set a fixed channel size and it ran one complete flow, the runtime improvements were greatly diminished. This is obvious in hindsight, since VTR is basically relaunching parts of the flow over and over again with some kind of reallocation each time. But I tripped over obvious things like this when I first went down this path.

I also was not sure if there is a clear tradeoff between runtime and memory usage that the VTR developers would prefer. For example, how much memory overhead is acceptable for a given runtime speedup? I assumed that runtime improvements far outweigh any memory overhead, but I have not used VTR in any industrial use cases where even the baseline memory usage is quite large. Any insight into this would definitely be interesting.

vaughnbetz:
The two regtest failures are not a real concern; they are small QoR changes (memory footprint) on the same tiny design. They can be resolved by updating the golden results.

They key thing we'd need to merge this would be a full QoR run as @AmirhosseinPoolad noted. If you don't have a machine that can run that, perhaps Ayan could help.

vaughnbetz (commented):
The code changes look minimally intrusive and hence fine.
A quick web search shows jemalloc isn't supposed to support valgrind; not sure how you've worked around this (turned off jemalloc for some builds?) as we do use valgrind, and wouldn't want to lose it. If this implicitly makes valgrind not work, we'd need to figure out a solution to that too.

So outstanding list is:
- Bigger QoR experiment
- valgrind status / interaction
- update golden results on the two failing tests

gigeresk:
We should also disable jemalloc when VTR_ENABLE_SANITIZE is set, as ASan also replaces malloc

gigeresk:
I ran valgrind on the jemalloc build and did not run into any issues; all the information seems to have been preserved. I'd need a +1 on this, though, since I've used ASan more than I've used valgrind...

PR #3516, opened 2026-05-04
[CI][HOTFIX] Commented Out Failing Step in Win Build

AlexandreSinger:
The CI is failing on Master. It is failing when dumping the tool version. Commenting out this step to try and fix this.

@tangxifan FYI.

### DIFF
--- a/install_win_msys2_mingw64_packages.sh
+++ b/install_win_msys2_mingw64_packages.sh
@@ -25,6 +25,7 @@ fi
 #pacman -Syu --noconfirm
 
 pacman -S --noconfirm --needed \
+    --ignore mingw-w64-x86_64-gcc,mingw-w64-x86_64-gcc-ada,mingw-w64-x86_64-gcc-libs \
     autoconf \
     automake \
     libtool \
--- a/install_win_msys2_mingw64_packages.sh
+++ b/install_win_msys2_mingw64_packages.sh
@@ -4,28 +4,11 @@
 # Converted from Ubuntu 22.04 apt-get script
 # Run this script inside an MSYS2 MINGW64 shell (or UCRT64)
 
-# Yosys is still based on GCC 15
-# Pin GCC to a fixed version before installing other packages so that the
-# MinGW tools are resolved against the correct gcc-libs ABI.
-pacman -U --noconfirm \
-    https://mirror.msys2.org/mingw/mingw64/mingw-w64-x86_64-gcc-15.2.0-14-any.pkg.tar.zst \
-    https://mirror.msys2.org/mingw/mingw64/mingw-w64-x86_64-gcc-ada-15.2.0-14-any.pkg.tar.zst \
-    https://mirror.msys2.org/mingw/mingw64/mingw-w64-x86_64-gcc-libs-15.2.0-14-any.pkg.tar.zst
-
-# Prevent pacman from upgrading GCC tools beyond 15.2.0-14
-# Add to /etc/pacman.conf IgnorePkg line if not already present
-if grep -q "^IgnorePkg" /etc/pacman.conf; then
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
Testing to see if this is the real issue. This is likely not the root cause. I think there may be something wrong with the `cmake` executable.

AlexandreSinger:
The failure is still present. It looks like the cmake executable is having issues. It is likely caused by the GCC downgrade that is happening.

I am not sure why the original PR build did not show this behaviour.

AlexandreSinger:
I am trying one more thing. If this does not work, I may remove the downgrade. I do not think we are even building Yosys right now. I think this downgrade may not be a great idea, it may be better to update Yosys if possible. The version of Yosys that we use may be a bit old. I have not ensured that VPR worked on gcc-16, but we will see.

AlexandreSinger:
It looks like VPR also has some issues with GCC-16 (since it is technically not in stable release yet for Ubuntu, we do not test on it.

I am not sure how best to fix this. @tangxifan Do you have any ideas what to do? I want to avoid reverting this change since it may only appear when on the Master branch. We can just ignore this test as we merge things in.

AmirhosseinPoolad:
Seems like GCC-16 requires you to explicitly include \<cstdint\> for fixed-size integer types. Ubuntu 26 with GCC-16 *is* actually released and is stable, but GitHub actions doesn't provide it yet. I guess we can just leave the failure be there for now. I'll upgrade my distro tomorrow and raise a PR fixing the compile errors.

tangxifan:
@AlexandreSinger Comment out the check tool version is o.k. This is an annoying aspect of Msys2, all the packages are keep rolling to its latest release without a support on a fixed version. That is why I fixed the version of GCC in the script here. The latest prebuilt Yosys on Msys2 is still compiled by GCC-15. When you use GCC-16 to compile Parmys, you will face the issue in lib linking (conflicts between GCC-16 and GCC-16 libs). 


@AmirhosseinPoolad I agree with the syntax errors. GCC-16 may force strict checks here while GCC-15 just ignores.

install_win_msys2_mingw64_packages.sh
```diff
@@ -60,18 +60,3 @@ pacman -S --noconfirm --needed \
     mingw-w64-x86_64-eigen3 \
     mingw-w64-x86_64-graphviz \
     mingw-w64-x86_64-openssl
-
```
tangxifan, line 63: @AlexandreSinger If you prefer GCC-16, you may add it to LINE62 explicitedly:

- mingw-w64-x86_64-gcc
- mingw-w64-x86_64-gcc-ada
- mingw-w64-x86_64-gcc-libs

tangxifan:
@AlexandreSinger I do not think it is a GCC problem. It is the Msys2 cmake issue. I have the same issue on OpenFPGA now. Look like someone in Msys2 has messed up cmake packages.

AmirhosseinPoolad:
So Msys2 is basically a rolling release? That's honestly a bit unsustainable to have in CI. We should switch to MSVC and NMAKE as soon as possible imo.

tangxifan:
> So Msys2 is basically a rolling release? That's honestly a bit unsustainable to have in CI. We should switch to MSVC and NMAKE as soon as possible imo.

True. When you run something like: 

```
pacman -S ming-w64-x86_64-gcc 
```
It will always install the latest stable gcc-version which is rolling. 
You cannot select which version to install like ``apt install``. Same to all the other packages.
This will cause CI not stable as it may crash anytime Msys2 updates their package list.

That being said, I think we can still build a CI but it requires a stable container image where
- All the packages are installed with a fixed version, just as I do for the GCC-15
- The container image can be built and hosted by a Github repo or other Cloud drive. We pull it each time.
As such, we avoid frequent maintenance on Msys2 setup

AlexandreSinger:
Thank @AmirhosseinPoolad for getting the CI back up and running.

@tangxifan We can discuss next steps in the VTR meeting. We can downgrade back to GCC-15 if it does not break the CI again; but it looks like us being at GCC-16 fixed the issue at least temporarily.

PR #3517, opened 2026-05-05
[AP] Handled Corner Case Testcases

AlexandreSinger:
Some test cases in the strong tests are non-AP specific. I made these tests explicitly non-AP and added AP tests for the parts that were explicitly supported.

I also added a better error for when the delta delay override is used since it uses pins as part of its delay model which does not make sense for a flat placement.


vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_fix_clusters/config/config.txt
```diff
@@ -15,13 +15,13 @@ circuit_list_add=apex2.blif
...
-qor_parse_file=qor_standard.txt
+qor_parse_file=qor_fixed_chan_width.txt
 
 # Pass requirements
-pass_requirements_file=pass_requirements.txt
+pass_requirements_file=pass_requirements_fixed_chan_width.txt
 
 # Script parameters
-script_params =-starting_stage vpr --fix_clusters ../../../../apex2_block_locations.place
+script_params =-starting_stage vpr --pack --place --route --route_chan_width 206  --fix_clusters ../../../../apex2_block_locations.place
```
AmirhosseinPoolad, line 27: Why fix channel width for this test?

AmirhosseinPoolad (commented):
Thanks Alex, this should improve test coverage.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_fix_clusters/config/config.txt
AlexandreSinger: This testcase took 3x longer because it was doing a min-channel width search. The purpose of this testcase was to test the `fix_clusters` feature; so burning the compute time doing min-channel width didn't make sense to me.

What do you think? I can revert this if you think it is needed.

AmirhosseinPoolad: Makes sense to me.

PR #3518, opened 2026-05-05
Fix gcc-16 compile errors

AmirhosseinPoolad:
This PR fixes compile errors with GCC 16.  I also incorporate #3516 into this. Merging this would also merge #3516.

### DIFF
--- a/vpr/src/pack/packing_signature_tree.cpp
+++ b/vpr/src/pack/packing_signature_tree.cpp
@@ -75,7 +75,7 @@ void PackingSignatureTree::add_lcn(const t_pb_graph_node* primitive_pb_graph_nod
     // Determine whether a path with this new LCN already exists.
     // Similar clusters are likely to get packed close to each other in seed-based flows,
     // so iterate over the list in reverse to take better advantage of this locality.
-    for (ptrdiff_t i = cursor_->child_lcn.size() - 1; i >= 0; i--) {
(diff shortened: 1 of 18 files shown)


vpr/src/pack/packing_signature_tree.cpp
```diff
@@ -237,7 +237,7 @@ e_ecn_legality PackingSignatureTree::check_legality() {
 
     // Similar clusters are likely to get packed close to each other in seed-based flows,
     // so iterate over the list in reverse to take better advantage of this locality.
-    for (ptrdiff_t i = cursor_->child_ecn.size() - 1; i >= 0; i--) {
+    for (size_t i = cursor_->child_ecn.size() - 1; i >= 0; i--) {
```
AlexandreSinger: hmmm... These should be integers I think...

AlexandreSinger: This should have caused an infinite loop if it was ever run. Thats weird.

AlexandreSinger: Ah! They all have branches inside them! We should still fix this.

AmirhosseinPoolad: Yep yep, fixed it already. Will push the commit soon.

AlexandreSinger (approved):
LGTM! Thanks Amir.

PR #3519, opened 2026-05-05
[WIP][Place] Fixed Issue Where Macros Were Not Being Created

AlexandreSinger:
For special architectures, there may exist direct connections from one sub-tile to another within the same tile. For blocks which use these direct connections, they must be placed relative to specific constraints.

Normally, to enforce these constraints during packing we use macros; however macros were not being created for this case. Although the tests were passing, they were passing due to random chance (since the initial placer happened to place them in the correct order and the annealer did not change them). Now that AP is coming in, this is actually causing an issue.

Fixed the macro creation code so it actually creates a macro for this case.


vpr/src/place/place_macro.cpp
```diff
@@ -305,6 +323,168 @@ int PlaceMacros::find_all_the_macro_(const ClusteredNetlist& clb_nlist,
         } // Finish going through all the pins for to_pins.
     } // Finish going through all blocks.
 
+    // Pure-source macro detection:
+    // Handles blocks that have DRIVER pins for a direct but no RECEIVER pins for that
+    // same direct (e.g. a multiplier driving memory blocks via sub-tile direct connections).
```
vaughnbetz, line 328: 1. This is a very long routine. Suggest adding some helper functions to break it up.
2. It's not clear to me what a 'pure-source' macro is (maybe spans two different types of blocks?), or what the overall algorithm is fo identifying them.  I suggest definiing the term, and giving an overview of the algorithm here,

AlexandreSinger: I agree, this was the PR I was discussing with you a week or two ago. It fixes the issue of the inter-sub-tile macros; however I am not confident in the quality of the solution (Claude had a heavy hand in creating it). I left this PR open for @amin1377 to look at if he was curious.

I did verify that it fixes the issue without breaking the rest of VPR (as far as I can tell); but the code is a bit magical.

vaughnbetz: Ok, to @amin1377 to see if he can make a better solution.

PR #3520, opened 2026-05-06
[CI] Fix Windows Build Caching

AlexandreSinger:
Currently the windows build is not using ccache. Looking at the build script, I beleive it is caused by CMAKE not using the ccache by default.

Normally we set ccache using the PATH variable on linux, but I am not sure if Windows can do the same. Instead we can set ccache in CMAKE by declaring it as our compiler launcher.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -857,8 +857,8 @@ jobs:
       - name: Build
         shell: msys2 {0}
         env:
-          CMAKE_PARAMS: "-DWITH_ABC=OFF -DWITH_PARMYS=OFF -DSLANG_SYSTEMVERILOG=OFF"
+          CMAKE_PARAMS: "-DWITH_ABC=OFF -DWITH_PARMYS=OFF -DSLANG_SYSTEMVERILOG=OFF -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
           BUILD_TYPE: release # Note: Found that release build caches way better than debug build for this test.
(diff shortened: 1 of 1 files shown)


AmirhosseinPoolad:
Windows build is still taking significantly longer than linux builds, so I don't think this approach works.

AlexandreSinger:
> Windows build is still taking significantly longer than linux builds, so I don't think this approach works.

Just checked it. The cache is still not being used. ccache was already built on the windows machine, and it was pointing to the pre-built one instead of the one used by our ccache action. Tried another fix now to see if it worked.

AlexandreSinger:
Note that it will be slow the first time we run this; but if it caches I can rerun the action and see if it restores the cache correctly.

AlexandreSinger:
Ok, that worked.
image

I am rerunning he job now to see if the cache is restored properly.

AlexandreSinger:
OK, caching confirmed working:
Screenshot from 2026-05-06 13-16-00

The build is now around 2x-4x faster; but I think we can do a LOT better.

Right now, this build has IPO turned on (the default for VTR). I am going to turn that off. There is no reason for us to use LTO for this build. We are just wasting time linking.

AlexandreSinger:
Now that's what I am talking about!

|Build | time (mins) |
| --------------------------- | ----------- |
| Original build time | 22 |
| new uncached build time (No IPO) | 16 |
| new cached build time (No IPO) | **1.5** |

15x faster for fully cached builds!

AmirhosseinPoolad (approved):
LGTM

AlexandreSinger:
@tangxifan FYI

tangxifan:
@AlexandreSinger LGTM Thanks!

PR #3521, opened 2026-05-06
Add clang tidy bugprone

AmirhosseinPoolad:
This PR adds a helper script to run clang-tidy, along with a config with some sensible checks enabled. There's also fixes for all the stuff that were flagged by clang-tidy. I don't want to add this to CI yet, but it would be nice at some point.

Here's some of the issues flagged by clang-tidy and fixed in this PR:

>VTR_ASSERT(bin_target_density_[new_bin_id] = target_density);

Assignment instead of comparision

> for (double pivot_x = min_pivot_x; pivot_x <= max_pivot_x; pivot_x++)

Floating point loop index

> block_id = std::atoi(block_id_string.c_str());
 
could simply use std::stoi instead. We had lots of this.

> static_cast<int>(0.5 + (float)first_move_lim / (float)placer_opts.inner_loop_recompute_divider);

Wrong way to round a float. For example static_cast<int>(0.5 + 0.499999975) results in 1 instead of zero.

> ss << "/c=" << m_compressor_id;

Because m_compressor_id is char, this actually prints a character instead the numerical id.

> const double noc_link_bandwidth = dist_2.max() / 5;

integer division into a float.

__________________

And a lot of use after moves. I overall think this is a good addition to our CI. Note that I think some of the static_cast changes are not needed anymore since I decided to exclude narrowing cast warnings. If you think they're excessive, I can remove them.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -30,7 +30,7 @@ Checks: >
   #-readability-redundant-inline-specifier,
   #-readability-isolate-declaration,
 
-WarningsAsErrors: ''
+WarningsAsErrors: '*'
 HeaderFilterRegex: '.*'
 ExcludeHeaderFilterRegex: '.*/libs/EXTERNAL/.*'
 FormatStyle: file
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,36 @@
+---
+Checks: >
+  -*,
+  bugprone-*,
+  -bugprone-easily-swappable-parameters,
+  -bugprone-narrowing-conversions,
+  -bugprone-derived-method-shadowing-base-method,
+  -bugprone-implicit-widening-of-multiplication-result,
+  -bugprone-branch-clone,
+  -bugprone-signed-char-misuse,
+  -bugprone-sizeof-expression,
+  -bugprone-lambda-function-name,
+  -bugprone-throwing-static-initialization,
+  -bugprone-random-generator-seed,
+  -bugprone-nondeterministic-pointer-iteration-order,
+  -bugprone-unchecked-optional-access,
+  -bugprone-suspicious-stringview-data-usage,
+  -bugprone-return-const-ref-from-parameter,
+  -bugprone-unchecked-string-to-number-conversion,
+  # performance-*,
+  # readability-*,
+  #-readability-function-cognitive-complexity,
+  #-readability-identifier-length,
+  #-readability-simplify-boolean-expr,
+  #-readability-redundant-access-specifiers,
+  #-readability-avoid-const-params-in-decls,
+  #-readability-else-after-return,
+  #-readability-uppercase-literal-suffix,
+  #-readability-use-anyofallof,
+  #-readability-redundant-inline-specifier,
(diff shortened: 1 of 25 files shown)


dev/run_linter.py
AlexandreSinger, line 1: I like having a file like this a lot. But I do not think it belongs in the root of the whole project. I recommend moving it to the `dev/` folder since it contains a lot of scripts like this one.

For example, I put the test suite verifier here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/tree/master/dev/vtr_test_suite_verifier

vpr/test/test_setup_noc.cpp
```diff
@@ -372,7 +372,7 @@ TEST_CASE("test_identify_and_store_noc_router_tile_positions", "[vpr_setup_noc]"
         t_grid_def grid_def;
         grid_def.name = device_grid_name;
         grid_def.layers.resize(1);
-        DeviceGrid test_device = DeviceGrid(grid_def, std::move(test_grid));
+        DeviceGrid test_device = DeviceGrid(grid_def, test_grid);
```
AlexandreSinger, line 375: Why remove these moves? The t_grid_def object is passed by value not by reference, so a move here would be correct I think if we want performance. I do recognize that this is part of the tests though.

.clang-tidy
```diff
@@ -0,0 +1,36 @@
...
+  #-readability-simplify-boolean-expr,
+  #-readability-redundant-access-specifiers,
+  #-readability-avoid-const-params-in-decls,
+  #-readability-else-after-return,
+  #-readability-uppercase-literal-suffix,
+  #-readability-use-anyofallof,
+  #-readability-redundant-inline-specifier,
+  #-readability-isolate-declaration,
+
+WarningsAsErrors: ''
```
AlexandreSinger: Once this becomes a CI test, this should probably be turned on; or we should learn how to turn this on from the script you added.

AlexandreSinger (commented):
Hi @AmirhosseinPoolad , just some comments.

I like this change a lot and I am looking forward to bringing it in! We should consider creating a CI test for this! Let me know if I can help in any way.

.clang-tidy
AmirhosseinPoolad: It's turned on in the script. I should probably turn it on in the config file instead.

vpr/test/test_setup_noc.cpp
AmirhosseinPoolad: We're moving test_grid, but then use again in the same scope. Not sure why it doesn't crash. I'll recheck it again to see what's going on, but I believe t_grid_def's move constructor doesn't actually invalidate the moved object.

AlexandreSinger: OH! That's very interesting! Good catch!

.clang-tidy
AlexandreSinger: Up to you! I did not see it in the script; but perhaps I was not looking too deeply.

AlexandreSinger (approved):
LGTM

PR #3522, opened 2026-05-06
[SDC] Added -invert and -duty_cycle support

AlexandreSinger:
Added support for the -invert and -duty_cycle arguments for the create_generated_clock command. This just required changing the rise and fall time calculations.

I also fixed the limitation where generated clock sources must have a rise time at 0ns. This came for free with verifying and implementing these changes.

Added tests for these new arguments. To make the tests clean, I added a new constraint list item for SDC files. This will make it easier to constrain SDC files to circuits directly in the task list.


vpr/src/timing/read_sdc.cpp
```diff
@@ -332,19 +336,29 @@ class SdcParseCallback : public sdcparse::Callback {
...
-            //       shift should be performed.
-            vpr_throw(VPR_ERROR_SDC, fname_.c_str(), lineno_,
-                      "Creating a generated clock on a source with a non-zero rising edge is currently unsupported in VPR.");
+        // Get the rise and fall edge for clock multipliers and dividers.
+        //  The source clock's rising edge may fall outside of the generated clock's
+        //  period (suppose this is a clock multiplier with a high multiple). Modulate
+        //  the rise edge. The generated rise-edge must occur at the same time as the source
+        //  at some point, so the modulo should be safe.
+        double generated_rise_edge = std::fmod(source_clock_cmd.rise_edge, generated_clock_period);
+        //  Since we can only multiply and divide the clock, the generated clock
```
vaughnbetz: Clarify comment. (If duty cycle not specified, SDC specifies it is 50%).
Then say if a duty cycle is specified (which is only allowed for multiplication and was previously checked) then use it.

vaughnbetz (approved):
Looks good. 
suggest one commenting update, and confirming you get the same launch and latch edges on these test cases as another tool (e.g. Quartus or OpenSTA).

AlexandreSinger:
@vaughnbetz FYI: I have been comparing the output of OpenSTA with Tatum to see how we are performing the timing analysis differently. I found a few differences that I was not expecting:

1. For paths that cross clock domains that are close together (for example they both rise at t = 0), Tatum reports a violation while OpenSTA reports meets timing. I dug into it and OpenSTA seems to always wait for a clock cycle. I think this is expected behaviour for Tatum since it is very pessimistic when it comes to clock domain crossings. This is true for any multi-clock circuit where the clock crossing arcs are not cut. I am leaving this as an expected issue. I think we should document this better somewhere if this is expected behaviour.
2. OpenSTA includes the time it takes for the Master clock to propagate to the driver of the generated clock. For example, if the generated clock is generated on the Q port of a FF (and the master clock is the clock on that FF), then the clock-to-Q delay is included in the propagation delay of the generated clock. I am not sure how we would support something like this, but for PLLs this may not be an issue. I think the best next step for this is to just document it as a quirk of our tool for now. Eventually this may be something worth supporting.
3. OpenSTA and Tatum report slightly different paths. OpenSTA is missing quite a few paths I would expect to see; but this may just be an issue with how I am calling OpenSTA since I was able to traverse the timing graph in OpenSTA and see the arcs myself. Tatum was missing a timing path: for a basic clock divider using a FF, it did not report the timing arc of the generated clock leaving the Q pin of the FF and being fed back into the driving FF (which is clocked by the master clock). I am not sure if this is expected behaviour or not.
4. OpenSTA actually does report the waveforms of generated clocks (fortunately), which allowed me to verify my clock division and multiplication. I found that they do not actually modulo their waveforms! This was interesting to me. It looks like they store their waveforms to be non-moduloed and then they must modulo them within their calculation functions. For example, OpenSTA allows a clock of period 4 to have a waveform {6.0, 8.0}, even though those rise and fall times fall outside of the period. Its quite interesting, and I think this is a behaviour worth supporting. The code is easy to update.

I will update the code according to (4) and I will try to update the documentation the best that I can for (2). (1) and (3) I am not sure what to do about since these may be known things in Tatum. They are likely not caused by the generated clocks, so it may be worth putting them in a separate issue.

vaughnbetz:
Thanks Alex. Let's talk on a whiteboard next week. For 1, I think Tatum is correct and would match Quartus, as STA/SDC generally defaults to conservative.

PR #3523, opened 2026-05-06
[WIP] Add clang tidy perfermance warnings

AmirhosseinPoolad:
Builds on top of #3521. Still nothing in the CI, but adds perf related flags to the clang-tidy config and the fixes that come with it. This flagged a lot of unnecessary copies and things like that.


AmirhosseinPoolad:
I've changed my mind, don't plan to do this in now. I'll do the readability checks next.

PR #3524, opened 2026-05-07
[AP] Fixed Incremental Timing Updates in GP

AlexandreSinger:
The global placer was not invalidating the timing edges, so when incremental timing updates were used, the results were incorrect.

Properly invalidated the edges.


vpr/src/analytical_place/global_placer.cpp
```diff
@@ -322,6 +322,18 @@ static void update_timing_info_with_gp_placement(PreClusterTimingManager& pre_cl
         pre_cluster_timing_manager.set_timing_arc_delay(atom_sink_pin_id, delay);
     }
 
+    // If the timing update type is incremental, we need to invalidate all edges which have changed.
+    // We assume here that all edge delays change in some way. We could do a more complicated
```
AmirhosseinPoolad, line 326: Isn't that non-incremental then? I might be mistaking something but how is it incremental if you invalidate everything every time?

AmirhosseinPoolad (commented):
Random drive-by comment. Sorry.

vpr/src/analytical_place/global_placer.cpp
AlexandreSinger: No problem, I appreciate the feedback! I actually looked into turning off incremental routing for Global Placement when requested (and use it only for detailed placement), but the code that manages the timing updates are shared between global placement and legalization and I did not want to lock the incremental feature from the legalizer. I chose this approach since it was the cleanest and simplest.

haydar-c (approved):
Looks good to me!

PR #3525, opened 2026-05-07
[AP] Fixed Bug in Extended Map Router Lookahead

AlexandreSinger:
The extended map router lookahead test was failing for AP due to an assert being hit when a cost was found to be negative infinity.

After debugging, I found that this was caused by a memory access out of bounds. There is a special case where a vector is not as large as expected and we were accessing a value outside of that vector. Handled that case to resolve the issue.

Added AP test cases to lock this in.


amin1377 (approved):
LGTM. Thanks, Alex!

PR #3526, opened 2026-05-07
Add MSVC compatibility testing workflow

tangxifan:
#### Description

Add CI workflow to validate build compatibility for MSVC compiler on windows platform

#### Related Issue

#3432 

#### Motivation and Context

As described in #3432 

- [x] Added a installer script for vcpkg which can be reused on windows platform
- [x] Added a workflow job to run msbuild 

**Note that**
- Due to Yosys has a prebuilt version on Windows and it requires a special workflow to reproduce.  This workflow is currently being tested on OpenFPGA. Will bring it once ready.
- ABC, Yosys are disabled in the MSVC build

#### Types of changes

#### Checklist:


tangxifan:
@vaughnbetz @AlexandreSinger This PR is ready for review. 
Same to Msys2 CI, the MSVC CI does not include regression tests due to missing Yosys and Parmys. 
In the next step, we will focus on
- the integration of prebuilt Yosys on Windows
- adapt regression tests to Windows platform
- Then we can track the QoR changes between platforms.

tangxifan:
Once this PR is merged, the OpenFPGA PR https://github.com/lnis-uofu/OpenFPGA/pull/2505 will be  merged.

vaughnbetz (approved):
Looks good to me.

vaughnbetz:
@AlexandreSinger or @AmirhosseinPoolad  : If you can take a quick look that would be good.

.github/workflows/test.yml
```diff
@@ -862,3 +862,75 @@ jobs:
...
+        with:
+          key: ${{ runner.os }}-msvc-${{ github.sha }}
+          restore-keys: ${{ runner.os }}-msvc-
+
+      - name: Configure and build with CMake
+        run: |
+          mkdir build
+          cd build
+          cmake ${{ steps.vcpkg.outputs.vcpkg-cmake-config }} -S .. -B . -G "Visual Studio 17 2022" -DWITH_PARMYS=OFF -DSLANG_SYSTEMVERILOG=OFF -DVTR_IPO_BUILD=OFF -DWITH_ABC=OFF -DWGET="${{ env.CURL_PATH }}" -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+          cmake --build .
```
AlexandreSinger, line 938: This cmake build is not being cached correctly:

Image

I think this is a similar issue to what I raised a PR about on Wednesday. I think you need to set the CCACHE_DIR enviornment variable.

We need this build to be cached or else it will become the long pole for the CI.

libs/libvtrutil/src/vtr_memory.h
```diff
@@ -126,6 +126,18 @@ struct aligned_allocator {
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
 
+    // Add rebind template for allocator conversion
+    template<typename U>
+    struct rebind {
+        using other = aligned_allocator<U>;
+    };
```
AlexandreSinger, line 135: The rebind template was depracted in C++17 I thought:
Image

Why was it needed here? I recognize that this class is basically our own version of std::allocator; but any downstream code should not be using rebind I think.

Is this struct necessary to fix some issue?

vcpkg.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "verilog-to-routing",
+  "version": "1.0.0",
```
AlexandreSinger: I am not sure what this version is used for, but I am wondering if this should be 9.0.0 to match VTR's current version?

AlexandreSinger (changes requested):
Hi @tangxifan ,

Looks good, I had some comments below.

libs/libvtrutil/src/vtr_memory.h
AmirhosseinPoolad: I think this at least should have a comment about why it's needed.

tangxifan: Currently yes. 
If removed, CI will failed. You can check CI logs in my commit history.
My guess: 
  It may be due to the MSVC header files (version 14) are still old while the compiler is new (version 19).  See the screenshot below.

image

vcpkg.json
tangxifan: Agree. Modified.
It is a tag here which will not impact anything package versions.
We need to update this file when version is bumped.

libs/libvtrutil/src/vtr_memory.h
tangxifan: > I think this at least should have a comment about why it's needed.

Yes. I added MSVC guard now and code comments.

.github/workflows/test.yml
tangxifan: Yes. Added ENV to MSVS build workflow. See if it works or not. I am not an expert in ccache.

AmirhosseinPoolad (approved):
Thanks Xifan! I think if CI passes this is good to go. Can you please re-run the MSVC build action after this one succeeds to make sure the cache works?

On a separate note, I'm pretty sure the entirety of vtr::aligned_allocator can actually be deleted and is not needed anymore. I'll raise a separate PR soon if my experiments don't show any performance degradation.

tangxifan:
> Thanks Xifan! I think if CI passes this is good to go. Can you please re-run the MSVC build action after this one succeeds to make sure the cache works?
> 
> On a separate note, I'm pretty sure the entirety of vtr::aligned_allocator can actually be deleted and is not needed anymore. I'll raise a separate PR soon if my experiments don't show any performance degradation.

@AmirhosseinPoolad Look like the ccache is not yet working on the MSVC build. May need to try a few ccache options after merging this PR to ensure it is working.

vaughnbetz (approved):
LGTM

tangxifan:
@AlexandreSinger @AmirhosseinPoolad Let me know if you have further comments. If not, I suggest to merge and then we follow-up ccache issue.

AlexandreSinger:
@tangxifan I would prefer not merging this in without caching if we can avoid it. The last windows test made it challenging to get new PRs in since it increased the end-to-end time of the CI. I would prefer to debug it here. I am testing moving to a new generator (Ninja) which I have used in the past. I think this has a good chance of working.

AlexandreSinger:
I have confirmed that the build is now caching successfully. When we have a 100% cache hit, the whole job takes under 5 minutes which is reasonable.

@tangxifan What do you think about this change? Are you ok with switching the generator in the CI test to Ninja? I have confirmed that the MSVC compiler is still being used:
image

tangxifan:
> I have confirmed that the build is now caching successfully. When we have a 100% cache hit, the whole job takes under 5 minutes which is reasonable.
> 
> @tangxifan What do you think about this change? Are you ok with switching the generator in the CI test to Ninja? I have confirmed that the MSVC compiler is still being used: image

Thanks @AlexandreSinger . We are good to go then? Ninja should be ok for me. 
Waiting for your approval and I will hit the merge the button.

AlexandreSinger (approved):
LGTM! I wanted to wait to approve until the Ninja generator was confirmed to be ok.

PR #3527, opened 2026-05-08
[Build] Resolved CLANG Warnings

AlexandreSinger:
Working on cleaning up some of the Clang warnings. Some of these warnings also appear in GCC-16 which will eventually need to be warning clean. Just getting ahead of the curve.


AmirhosseinPoolad (approved):
LGTM. Awesome.

PR #3529, opened 2026-05-08
Remove legacy memory management code

AmirhosseinPoolad:
There were some unused/mostly unused memory management code in vtr_memory.h. In this PR I removed the unused things and removed vtr::aligned_allocator which was used in RR Graph storage for performance critical stuff. This is not needed anymore in modern C++ and the default allocator respects alingas(). To be sure, I ran the Titan benchmarks:

  | Master | Branch
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.004046651
max_vpr_mem | 1 | 0.9999824426
pack_time | 1 | 1.013110634
place_time | 1 | 0.9995324822
crit_path_route_time | 1 | 0.9919467308
routed_wirelength | 1 | 1
critical_path_delay | 1 | 1
total_heap_pops | 1 | 1
total_swap | 1 | 1
accepted_swap | 1 | 1
placed_CPD_est | 1 | 1

We can see that QoR is completely unaffected. Runtime is not changed beyond noise and machine load. The allocator was actually the source of some problems in #3526. Removing that and unused code would make the code base simpler.

### DIFF
--- a/libs/librrgraph/src/base/rr_graph_storage.h
+++ b/libs/librrgraph/src/base/rr_graph_storage.h
@@ -1005,7 +1005,7 @@ class t_rr_graph_storage {
      * storage_ stores the core RR node data used by the router and is **very**
      * hot.
      */
-    vtr::vector<RRNodeId, t_rr_node_data, vtr::aligned_allocator<t_rr_node_data>> node_storage_;
+    vtr::vector<RRNodeId, t_rr_node_data> node_storage_;
 
     /** @brief
      * The PTC data is cold data, and is generally not used during the inner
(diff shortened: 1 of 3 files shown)


vaughnbetz (approved):
LGTM; nice simplification @AmirhosseinPoolad !
The chunk allocator stuff was written by me long ago in my PhD, before vpr was C++.  Looks like we've converted all uses of it so it's good to get rid of this blast from the past :).

vaughnbetz:
Looks like it needs some conflicts resolved; go ahead and merge it as soon as that's done.

AlexandreSinger (approved):
LGTM. I dug around and confirmed what you said about C++ 17. We should monitor the run time of routing and if we see random changes we should keep this PR in mind.

Aligned allocators can be nasty to debug since you get a lottery in performance if the alignment breaks. I think this should be safe.

PR #3532, opened 2026-05-12
[AP][Solver] Resolved potential issues with delays

AlexandreSinger:
I have found that the delay calculations were a bit unsafe at times. They worked properly for rock-solid, simple delay models; however some delay models may have holes in them (since some routes were not found) and some delay models do not rely on spatial symmetries that my old code was relying on.

Updated the delay code to check for missing delay placeholder values and handled them explicitly. I also added proper bounds checking to ensure that the delays returned are always valid.

Also added cleaner sentinel handling for the missing delay placeholder values. This will make future updates much cleaner.


AlexandreSinger:
@AmirhosseinPoolad FYI. I will test to ensure this fixes the issues in your branch. This refactoring should resolve the issues we discussed and then some.

AlexandreSinger:
> @AmirhosseinPoolad FYI. I will test to ensure this fixes the issues in your branch. This refactoring should resolve the issues we discussed and then some.

Results on the circuit we were discussing with assert-level 3:
image

The quality is surprisingly better. It could be a fluke, but more likely that this cleanup resolved a long-standing bug. I need to run some benchmarks to be sure that overall everything is ok.

AmirhosseinPoolad:
That's great! thanks Alex.

AlexandreSinger:
Results on Titan with fixed device sizes. All results are normalized to what is on Master relative to this change. Gaussian blur was omitted for time.

  | Result normalized to Master
-- | --
Post-GP WL | 0.96
Post-FL WL | 0.97
Post-DP WL | 1.00
Post-Route WL | 1.00
Post-GP CPD | 1.01
Post-FL CPD | 1.02
Post-DP CPD | 1.00
Post-Route CPD | 1.02

Overall the results post detailed placement are identical, but post routing the WL tied with the CPD increasing by 2%. Looking through the results some outliers are pulling up the average. I believe these results are fine. I much prefer this new code since I have more confidence that it is doing what we want.

AlexandreSinger:
@haydar-c When you have a moment, could you please review this PR?

vpr/src/analytical_place/global_placement/global_placer.cpp
```diff
@@ -314,6 +314,14 @@ static void update_timing_info_with_gp_placement(PreClusterTimingManager& pre_cl
...
                                               sink_block_loc,
                                               0 /*to_pin*/);
+        // TODO: The place delay model returns ROUTER_LOOKAHEAD_NO_PATH_SENTINEL
+        //       when it does not have a delay for the given driver/sink pair.
+        //       This makes the placement look artificially bad, but it may
+        //       simply be a gap in the delay model. For now we treat it as zero;
+        //       a better fix would be to estimate it from distance, e.g.
+        //       distance * delay_per_tile.
+        if (delay >= ROUTER_LOOKAHEAD_NO_PATH_SENTINEL)
+            delay = 0.0;
```
haydar-c: I am not sure how good zero here, but definitely agree with the TODO.

vpr/src/analytical_place/global_placement/analytical_solver.h
```diff
@@ -656,20 +687,20 @@ class B2BSolver : public AnalyticalSolver {
      * objective function to help guide the solver to trading off timing and
      * wirelength.
      *
+     * The delay derivative is computed on the legalized placement, not the
+     * currently solved placement. So this would be constant per iteration of
+     * the solver.
```
haydar-c: Would be nice to add a sentence explaining the tradeoff of why we choose this and how it affects the solver. Maybe more stable and intuitive.

haydar-c (approved):
Thanks @AlexandreSinger, LGTM. Left a small comment about the documentation.

PR #3533, opened 2026-05-12
[Agent] Added CLAUDE.md File

AlexandreSinger:
From discussions with different members of the community, we decided to add a CLAUDE.md file to direct AI agents on how to work with VTR more efficiently. This includes issues we have seen with Claude in the past where it has issues with building VPR, running tests, and finding files.

I also updated our contribution policy for VTR to reflect our opinions on AI agents being used for VTR development. We are ok with them being used as a tool, but we want to ensure that the developer maintains full control and understanding of what they are doing.

This file will continue to evolve as it is being used. Every so often, we should reflect on how Claude solved a task and think about how we can make this document better.


CLAUDE.md
```diff
@@ -0,0 +1,256 @@
...
+- `//` — all inline implementation comments; always include a space after `//`
+- Never use `/* */` block comments inside function bodies
+
+All non-trivial functions must have a Doxygen `@brief` comment in the header (or at the static declaration in `.cpp`). Document every member variable with `///<`.
+
+### Assertions
+
+Prefer `VTR_ASSERT` and `VTR_ASSERT_SAFE` over bare `assert()`:
+- `VTR_ASSERT(cond)` — inexpensive checks, always on including release builds
+- `VTR_ASSERT_SAFE(cond)` — expensive checks, disabled in release builds
```
haydar-c: This might be a personal thing but I like *_MSG variants when useful. We can encourage the use of it if the condition being checked is not obvious from the statement itself.

haydar-c (approved):
Thanks @AlexandreSinger for bringing this in. I agree with the document, just a small comment.

CLAUDE.md
```diff
@@ -0,0 +1,256 @@
...
+C/C++ formatting applies to: `vpr/`, `libs/libarchfpga`, `libs/libvtrutil`, `libs/libpugiutil`, `libs/liblog`, `libs/librtlnumber`, `odin_ii/`.
+
+## Python Environment
+
+Many scripts in `vtr_flow/scripts/` require the Python virtual environment. Set it up once and activate it in every new terminal before running flow scripts or regression tests:
+
+```shell
+# One-time setup
+make env
+pip install -r requirements.txt
```
AmirhosseinPoolad: This might run pip for every single session.

AmirhosseinPoolad: I think the agent should only ever do the 'source' command. The user should install all packages themselves.

CLAUDE.md
```diff
@@ -0,0 +1,256 @@
...
+./run_reg_test.py vtr_reg_nightly_test1   # through test7 in parallel on CI
+./run_reg_test.py vtr_reg_weekly
+```
+
+Download extra benchmarks when needed:
+```shell
+make get_titan_benchmarks
+make get_ispd_benchmarks
+make get_symbiflow_benchmarks
+```
```
AmirhosseinPoolad: Yeah I don't want claude to run/download titan for me. strong and basic tests suffice imo.

CLAUDE.md
```diff
@@ -0,0 +1,256 @@
...
+| `draw/` | OpenGL/EZGL graphical visualization |
+| `noc/` | Network-on-Chip placement support |
+| `analytical_place/` | Analytical placement (AP) engine |
+| `power/` | Power estimation |
+| `util/` | Shared VPR utilities (`vpr_utils.h`) used across multiple stages |
+| `server/` | VPR server mode for external tool integration |
+
+### External Subtrees (`libs/EXTERNAL/`)
+
+`libs/EXTERNAL/` contains code pulled in from upstream repositories via `git subtree`: `abc`, `libargparse`, `libblifparse`, `libsdcparse`, and `libtatum`. Do not modify files under `libs/EXTERNAL/` directly — changes must be made in the upstream repository and then synced into VTR using `dev/external_subtrees.py`.
```
AmirhosseinPoolad: abc is not in the libs/EXTERNAL directory.

AmirhosseinPoolad (commented):
In general I have a feeling that this \*might\* be too long, but any users can just monitor context usage and file issues if that turns out to be a problem.

CLAUDE.md
AmirhosseinPoolad: If you ask claude right now to "run the titan benchmarks" I'm pretty sure it would use grep and whatnot and look at the docs directly and do it. This file should contain things that are needed for \*every\* code change, no matter how big or small. This is why I don't think this should be here.

soheilshahrouz:
Every agentic coding tool except Calude Code uses AGENTS.md. I think it is better to rename this file and create a symbolic clink for claude file.

The file is too long. I think we should have a `.agents/skills/` directory and move most of this file's content to related skill files.

AlexandreSinger:
> Every agentic coding tool except Calude Code uses AGENTS.md. I think it is better to rename this file and create a symbolic clink for claude file.

I have been trying to poke around and verify that claim (that all agents use AGENTS.md), but I have been finding that each tool has its own file format. I was also considering the symbolic link stuff, but from looking around its not suggested since symbolic links do not work on windows and each agent file should be changed for their specific agent. I decided to stick with Claude since its what I use. Feel free to push back though!

> The file is too long. I think we should have a .agents/skills/ directory and move most of this file's content to related skill files.

I was worried about this too, but I wanted to keep this simple for the first push into VTR. Do you think this will be a big enough issue to fix this now? I really want to try out the CLAUDE.md file and see how it behaves before going down the skills route.

AlexandreSinger:
@soheilshahrouz I have made the updates you suggested. I think it should work well. Let me know what you think and if you have any further comments.

AlexandreSinger:
@vaughnbetz Let me know if you have any comments on this.

amin1377 (approved):
Looks good to me, Alex. I don’t have anything further to add beyond what others have already said, and I think it would be very useful. Thanks for doing this!

PR #3534, opened 2026-05-14
[MQ] Cleaned up target-specific code

AlexandreSinger:
The original MQ code was written to only target a specific implementation of threading using pthreads and targetted only linux; however, we have since moved the code to use std::threads instead. This allows us to use more idiomatic ways of working with the threads.

Cleaned up the code so it would work on different systems, not just linux; while keeping the code idiomatic.

I also removed the PERF define from the MQ file. This was originally used to profile the MQ file in its isolated repo (outside of VTR), but it does not work at all here in VTR. Much cleaner to just remove.

### DIFF
--- a/vpr/src/route/multi_queue_d_ary_heap.tpp
+++ b/vpr/src/route/multi_queue_d_ary_heap.tpp
@@ -27,36 +27,11 @@
 #include <algorithm>
 #include <atomic>
 #include <cassert>
+#include <thread>
 #include "d_ary_heap.tpp"
-
-#ifdef _WIN32 // Windows
-// Windows thread identification uses Win32 APIs from <windows.h>.
-#include <windows.h>
-#else
-// POSIX platforms use pthread APIs from <pthread.h>.
-#include <pthread.h>
-#endif
-
-// We use compiler-specific attributes for performance and correctness:
-// - MQ_NOINLINE: prevent inlining of selected functions (PERF builds)
-// - MQ_ALIGNAS(N): enforce alignment (e.g., cacheline alignment to avoid false sharing)
-//
-// GCC/Clang use __attribute__((...)), while MSVC uses __declspec(...).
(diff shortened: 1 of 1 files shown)


vpr/src/route/multi_queue_d_ary_heap.tpp
```diff
@@ -27,36 +27,11 @@
...
-#define MQ_ALIGNAS(N) __declspec(align(N))
-#elif defined(__GNUC__) || defined(__clang__)
-#define MQ_NOINLINE __attribute__((noinline))
-#define MQ_ALIGNAS(N) __attribute__((aligned(N)))
-#else
-#define MQ_NOINLINE
-#define MQ_ALIGNAS(N)
-#endif
-
 #define CACHELINE 64
```
AmirhosseinPoolad: Could we use std::hardware_destructive_interference_size as a portable way to do this instead of hard coding it?

AmirhosseinPoolad (commented):
Nice change!

vpr/src/route/multi_queue_d_ary_heap.tpp
AmirhosseinPoolad: Checking some modern CPUs, I could find any x86 CPU with either 32B or 128B cache line size ([source](https://www.agner.org/optimize/microarchitecture.pdf)) but Apple silicon (M1, M2, etc.) uses 128B cache lines. We don't officially support MacOS but I guess with the latest changes it should compile without any serious issues and using "std::hardware_destructive_interference_size" would further help with portability.

AlexandreSinger: This was a really good point! I have updated the code. It revealed a weird warning, which we both decided was fine to ignore for this file.

AmirhosseinPoolad (approved):

PR #3535, opened 2026-05-14
[CI] Made the MSVC build parallel

AlexandreSinger:
I found that the MSVC build was running serially which is just a waste on the GitHub runners. We may as well use the extra cores.

This should speed up both the cached and uncached MSVC builds.


AlexandreSinger:
This was SO worth it. The build time cut in half for non-cached builds. Sometimes the caches get removed if they are not used for sometime, so this will keep the CI running smoothely even in that case.

AmirhosseinPoolad (approved):
LGTM

PR #3539, opened 2026-05-20
Update index.rst under "doc/src/quickstart"

kevinxing-a12:
#### Description

Revised text in the quickstart page to explicitly tell the user to cd into "temp" before running VPR.
#### Motivation and Context

Since the VPR flow sometimes requires ".net" files in the "temp" directory to run, it would be better to explicitly tell the user to cd into that directory, or they may encounter a failed flow. 
I have checked the revised text in the preview mode and there is no display issue.

#### Types of changes

- a simple text revision

#### Checklist:


doc/src/quickstart/index.rst
AlexandreSinger, line 184: Above, the docs explicitly write out the command to change directory to a folder. To be consistent,  think we should do something similar:

Image

Something like: `cd $VTR_ROOT/temp`

AlexandreSinger: I am also wondering if it is worth adding a note below this code block explaining that it is reading the temporary files directly since `--analysis` is being provided. If it was not provided, the cd to temp is not necessary. @vaughnbetz Let me know what you think about that.

AlexandreSinger (commented):
Looks good @kevinxing-a12 !

I had a minor comment below.

doc/src/quickstart/index.rst
kevinxing-a12: Thanks Alex! That makes sense. Do you want me to make the change now (for the cd thing only)?

vaughnbetz: I think Alex is right; we should probably explain that we are cd'ing into temp so that vpr will read the prior packing, placement and routing output files witout us explicitly specifying them. Could say:

'To visualize this output, it is more convenient to cd into the directory (called temp in our case) that has the packing, placement and routing output files so we don't have to explicitly specify them.'

vaughnbetz (changes requested):

doc/src/quickstart/index.rst
```diff
@@ -176,12 +176,16 @@ which we can visualize with:
...
 .. code-block:: bash
 
+    > cd $VTR_FLOW_DIR/temp
     > $VTR_ROOT/vpr/vpr \
         $VTR_ROOT/vtr_flow/arch/timing/EArch.xml \
         blink --circuit_file $VTR_FLOW_DIR/temp/blink.pre-vpr.blif \
         --route_chan_width 100 \
         --analysis --disp on
 
+.. note:: When :option:`--analysis <vpr --analysis>` is provided, the VPR flow will attempt to read the temporary files under ``temp``, and hence we need to cd to that directory. If it is not provided, this step is optional.
```
vaughnbetz: temporary files -> prior optimization output files (.net, .place and .route)

vaughnbetz (changes requested):
One suggested tweak.

vaughnbetz (approved):
Looks good, thanks.

PR #3540, opened 2026-05-20
CRR Generation Speed Up

amin1377:
Implement two fixes to CRR generation:

- Support pin lookup by name in switch block CSVs (IPIN/OPIN cells may now contain a pin name string instead of a numeric PTC index). A per-tile-type name→PTC cache is built once at construction to keep repeated lookups O(1).
- Pre-compile regex patterns once at startup so the std::regex object is not reconstructed on every tile match.

The node search in process_opin_ipin_node is now strict: failing to find a node is a fatal error rather than a silent INVALID() return.


amin1377:
@AlexandreSinger: CI tests are passing now, and this PR is ready for your review.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -34,7 +34,16 @@ CRRConnectionBuilder::CRRConnectionBuilder(const RRGraphView& rr_graph,
     : rr_graph_(rr_graph)
     , node_lookup_(node_lookup)
     , sb_manager_(sb_manager)
-    , verbosity_(verbosity) {}
+    , verbosity_(verbosity) {
+    for (const t_physical_tile_type& tile_type : g_vpr_ctx.device().physical_tile_types) {
```
AlexandreSinger: Nit: We should probably pass in the physical types into the constructor, but I know that can be a pain.

Could we at least put the device context at the top of this function so it matches the rest of the code base.

Something like:
```cpp
const DeviceContext& device_ctx = g_vpr_ctx.device();
...
```

That way it will be cleaner and easier to move later.

AlexandreSinger: This is also up to you, but putting this logic in a helper function would allow it to be constructed in the initializer list which may be a bit cleaner. Something like:

```cpp
CRRConnectionBuilder::CRRConnectionBuilder(const RRGraphView& rr_graph,
                                           const NodeLookupManager& node_lookup,
                                           const SwitchBlockManager& sb_manager,
                                           const int verbosity)
    : rr_graph_(rr_graph)
    , node_lookup_(node_lookup)
    , sb_manager_(sb_manager)
    , pin_name_to_ptc_cache_(create_pin_name_to_ptc_cache(physical_tile_types)
    , verbosity_(verbosity) {}
```

This will probably make no performance difference, but its a clean way to do it. Not gating at all, it just came to mind.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
```diff
@@ -179,6 +181,10 @@ class CRRConnectionBuilder {
                                 RRNodeId source_node,
                                 RRNodeId sink_node,
                                 int segment_length = -1) const;
+
+    // Per-tile-type reverse map: pin_name -> pin_ptc.
+    // Built once at construction for all physical tile types.
+    std::unordered_map<const t_physical_tile_type*, std::unordered_map<std::string, int>> pin_name_to_ptc_cache_;
```
AlexandreSinger: t_physical_tile_type has an `index` member which can be used as an ID. So this can be made a bit faster (and cleaner) by using a vector, but up to you. Eventually, I would like to make strong IDs for this and logical block types.

vtr_flow/arch/timing/k6_frac_N10_frac_chain_mem32K_40nm_tileable_crr.xml
AlexandreSinger, line 1: Could you add to the comment in this file (at the top) the difference between this and the other copies of this architecture. We keep adding new ones and its hard to tell the difference quickly.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_pattern_matcher.h
```diff
@@ -22,10 +25,37 @@ namespace crrgenerator {
...
  */
 class CRRPatternMatcher {
+  public:
+    // Pre-compile a pattern and store it in the cache. Call once per pattern at startup.
+    void register_pattern(const std::string& pattern) {
+        cache_.emplace(pattern, compile(pattern));
+    }
+
+    // Returns true if name matches pattern. Uses the pre-compiled cache for registered
+    // patterns; falls back to on-the-fly compilation for unregistered ones.
```
AlexandreSinger: This comment may not be correct. For unregistered patterns, it raises a fatal error.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
AlexandreSinger: This is fine for now since the physical tiles do not change for the lifetime of this class; but its better to be safe.

AlexandreSinger (commented):
Hi @amin1377 

I have some minor comments and then this is good to merge.

amin1377:
@AlexandreSinger: I've addressed your comments. Please let me know if you have any other concerns.

AlexandreSinger (approved):
LGTM

PR #3541, opened 2026-05-20
[SDC] Fixed phase-related clock issues

AlexandreSinger:
Tatum was not analyzing phase-related clock domain crossings correctly. These changes fix this issue.

When the launch clock and capture clocks were phase related (meaning that at some point they align), Tatum was assuming that the period-based setup constraint was 0ns (meaning that the setup constraint was hitting the exact same clock edge for the launch and capture clock). Comparing this result with OpenSTA, we found that this was likely not the setup constraint we should be using. Resolving this by converting a >= in the code into a strict >.

Tatum was also not handling hold constraints between phase-related clocks correctly either. Tatum always assumed that the hold constraint was one capture period prior to the setup constraint. This does not hold true in general and it needs to be made relative to the actual rise times of the launch and capture edges that led to the setup constraint. I updated the hold constraint code to more explicitly search for the correct hold constraint.

A few strong tests had their golden results change due to this. I went through each one of them using OpenSTA and verified that this new version is correct (at least according to OpenSTA).

I also added 4 new strong tests (which I verified independently with OpenSTA) which verify phase-related clock crossings. These tests also test for multi-cycle paths (MCP), since the calculations I changed may affect their calculations.


AlexandreSinger:
Here is a more detailed example of what the issue was:
image

In this basic example, the setup constraint is the launch from FFA at t = 6.0 and the capture at FFB at t = 12.0, making the period-based setup constraint 6ns (12.0 - 6.0).

VTR was incorrectly setting the period-based setup constraint as 0ns. This was caused by VTR sweeping the rise edges on the launch and capture clocks and not using the pair which was STRICTLY larger than 0.

The second problem has to do with how the period-based hold constraint was computed. VPR originally just took the period-based setup constraint and subtracted 1 capture-clock period. So for this case it was using -6ns as the period-base hold constraint. This is incorrect for this case, the actual value should be 0ns. VPR was taking a bit of a lazy approach to computing this, which is only valid for cases where the launch and capture clocks had the same period. I fixed this by doing a proper sweep of launch and capture edges and calculating the hold delay directly.

I verified this and other similar testcases using OpenSTA to ensure Tatum and OpenSTA align on this.

AlexandreSinger:
@vaughnbetz This is the fix for the issues with the setup and hold calculations we discussed yesterday. From what I can find online, it looks like OpenSTA's behavior is the correct one. I have updated the code to be a strict inequality as we discussed.

After this change, I realized that the hold was still incorrect. After some digging, I found a bad assumption in the code which was causing the hold constraint to be incorrect. I have fixed this code and documented why it was necessary (in the code).

vaughnbetz (approved):
Looks great; nice commenting!

AlexandreSinger:
@vaughnbetz  Final note before merging. I was a bit paranoid because the comments in the code even said that the comparison should be strict. I was able to track down the exact commit that changed the comparison to include equality: `3a6b0d8578`

It was a refactoring of how clock periods were being calculated and adding support for hold overrides (set_min_delay). From my analysis, I have already verified the clock period calculation. Offline, I have just confirmed that the hold override is working correctly as well (matches OpenSTA). I think this change may have been needed for the structure of the timing analysis before and then regressed over time.

From this, I think this new PR is good (and is not undoing some other bug fix). Merging now.

@amin1377 If you have any comments on this from your side, please let me know!

PR #3542, opened 2026-05-21
Qt migration

w0lek:
**LIMITATION**: The doxygen doc will be corrected in separate PR after solving
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3563

libezgl PR
https://github.com/verilog-to-routing/ezgl/pull/25

Migrate VPR GUI from GTK3/Cairo to Qt 6.

#### Description
Replaces the entire GTK3/Cairo graphics stack with Qt 6 while keeping ezgl's public rendering API (renderer, draw_line, fill_rectangle, …) source-compatible, so VPR's drawing code in vpr/src/draw/ continues to call the same surface unchanged. The Qt port lives in the libezgl submodule and ships three interchangeable rendering backends:

- immediate — direct QPainter strokes on a QImage, closest in semantics to the legacy Cairo path.
- deferred — QPainter rendered from a per-frame batched command list (style-coalesced), trading higher latency for fewer state changes.
- rhi — GPU-accelerated via QRhi (D3D11 / Vulkan / Metal / Null backend), shader-rasterised lines, dashed lines, thick lines, fills, and arrowheads with per-tile spatial batching.

Selectable at runtime via the new --renderer {immediate,deferred,rhi} VPR flag (default rhi). The Glade .ui XML is reused as-is — a new QtGladeLoader (libezgl/src/qt/qtgladeloader.cpp) walks the XML and instantiates equivalent Qt widgets (GtkBox→QHBoxLayout/QVBoxLayout, GtkGrid→QGridLayout, GtkButton→QPushButton, GtkPopover→frameless QFrame, etc.).

The PR also lands a complete three-layer test suite for the new graphics path under vpr/test/gui/:

Layer 1 — Headless Smoke (run_headless_smoke.sh): VPR under QT_QPA_PLATFORM=offscreen with --disp off + --graphics_commands, validates pack/place/route reach the GUI subsystem without crashes.
Layer 3 — Catch2 Integration (test_vpr_gui binary): UI loader + widget-tree contract, draw-state and renderer fixtures, button/zoom/mouse/keyboard event coverage.
Layer 5 — Visual Regression (run_visual_regression.sh): per-renderer goldens at vpr/test/gui/golden/{rhi,immediate,deferred}/<case>.png; renders the matrix of cases × renderers, SSIM-compares each render to its renderer-specific baseline. Failures (default threshold 0.98) auto-emit [golden | current | 8×amplified-diff] triptychs

#### Motivation and Context
- Better cross-platform support.
- A GPU-accelerated rendering path (rhi) for large designs where the per-frame QPainter overhead in the immediate path becomes a bottleneck.

- manual running side by side GTK vs Qt
- new automated GUI test suite

#### Checklist:


w0lek:
@AlexandreSinger May I have help with the Strong test case failures? I suspect that, due to the Qt migration, a few test cases are failing because of an increase in ap_fl_max_rss caused by the transition from GTK to Qt.
```
vtr_reg_strong/strong_ap/flatrecon_full_legalizer...[Fail]
[Fail]
k6_frac_N10_40nm.xml/apex4.pre-vpr.blif/common ap_fl_max_rss relative value 1.3191214470284236 outside of range [0.8,1.2] and not equal to golden value: 77.4

vtr_reg_strong/strong_ap/flatrecon_full_legalizer...[Fail]
[Fail]
k6_frac_N10_40nm.xml/ex5p.pre-vpr.blif/common ap_fl_max_rss relative value 1.3823529411764706 outside of range [0.8,1.2] and not equal to golden value: 71.4

vtr_reg_strong/strong_ap/flatrecon_full_legalizer...[Fail]
[Fail]
k6_frac_N10_40nm.xml/frisc.pre-vpr.blif/common ap_fl_max_rss relative value 1.4039517014270035 outside of range [0.8,1.2] and not equal to golden value: 91.1
```
**Note:** we don't have such failures in the private fork, but we had similar ones that I fixed by updating the golden results. However, I think this stage requires approval.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/changes#diff-c743ef032cffbdac59bc2488cadd37b2130e68e2269ec6d0fe2c2fd78b22edd6

AlexandreSinger:
> @AlexandreSinger May I have help with the Strong test case failures? I suspect that, due to the Qt migration, a few test cases are failing because of an increase in ap_fl_max_rss caused by the transition from GTK to Qt.
> 
> ```
> vtr_reg_strong/strong_ap/flatrecon_full_legalizer...[Fail]
> [Fail]
> k6_frac_N10_40nm.xml/apex4.pre-vpr.blif/common ap_fl_max_rss relative value 1.3191214470284236 outside of range [0.8,1.2] and not equal to golden value: 77.4
> 
> vtr_reg_strong/strong_ap/flatrecon_full_legalizer...[Fail]
> [Fail]
> k6_frac_N10_40nm.xml/ex5p.pre-vpr.blif/common ap_fl_max_rss relative value 1.3823529411764706 outside of range [0.8,1.2] and not equal to golden value: 71.4
> 
> vtr_reg_strong/strong_ap/flatrecon_full_legalizer...[Fail]
> [Fail]
> k6_frac_N10_40nm.xml/frisc.pre-vpr.blif/common ap_fl_max_rss relative value 1.4039517014270035 outside of range [0.8,1.2] and not equal to golden value: 91.1
> ```
> 
> **Note:** we don't have such failures in the private fork, but we had similar ones that I fixed by updating the golden results. However, I think this stage requires approval. https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/changes#diff-c743ef032cffbdac59bc2488cadd37b2130e68e2269ec6d0fe2c2fd78b22edd6

@w0lek We can just update the golden results for these tests. This can be done locally by following the instructions found here: https://docs.verilogtorouting.org/en/latest/README.developers/#generating-new-qor-golden-result

Although we do build with graphics enabled, these tests do not request graphics. I am surprised that the memory changed at all. Before we merge this in, we may need to run this branch on Titan and measure if the memory increased. It would be strange to increase the memory of all runs for graphics. We can discuss.

w0lek:
@AlexandreSinger

> @w0lek We can just update the golden results for these tests. This can be done locally by following the instructions found here: https://docs.verilogtorouting.org/en/latest/README.developers/#generating-new-qor-golden-result

I run this instruction locally (ubuntu 24.04,gcc13,Qt6.9.3), but local values still 25% lower compare to values reported by CI. In order to not change the upper acceptable range to 1.3/1.4 i bump golden values manually. Hope this is ok.

> Although we do build with graphics enabled, these tests do not request graphics.  I am surprised that the memory changed at all. 

I agree for such kind of tests it's better to use not graphical mode. But i don't want to make such changes right now, to not update many other goldens in cascade if there will be such. Maybe we could log this as separate issue.

>Before we merge this in, we may need to run this branch on Titan and measure if the memory increased. It would be strange to increase the memory of all runs for graphics. We can discuss.

I run this test case (script is attached below), and got following numbers:
GTK: ~3100 MB, Qt: ~2925 MB. So Qt actually consumes less memory, but I still cannot explain these numbers. Also, I need someone other than myself to run this test to confirm that I am doing everything correctly.
Here is a scriptbody of running titan test i used.
[titan.sh](https://github.com/user-attachments/files/28261374/titan.sh)

Note: using rendering/graphics mode consumes more RAM compared to NO_GRAPHICS, even in headless mode. This is because, even without a visible GUI, we still link against Qt6Core.so, etc., and Qt additionally loads some plugins and related infrastructure.

Also, the more advanced renderers consume more RAM compared to the immediate renderer because they require additional buffers to sort and batch primitives.

AlexandreSinger:
@w0lek You may need to perform a clean build. In VTR, when you type `make` in the terminal it will use the last CMake arguments you used to build. So if you turned off drawing or used different assert levels, you can get different results.

I recommend running:
```sh
make distclean
make -j <num_cores>
```

This will hopefully put you in the same build environment as CI. Your machine matches our CI machines, so I am surprised to see such a large discrepancy. It is totally fine to bump the golden results manually in my opinion, but I do want to ensure that your environment is set up correctly so we can validate the results you collect! Unfortunately, this CMake caching stuff can make things a bit tricky.

AlexandreSinger:
I can help validate the memory overhead (by running your script on my end) once we get this PR closer to getting merged. Let's get the EZGL side merged first.

w0lek:
> I can help validate the memory overhead (by running your script on my end) once we get this PR closer to getting merged. Let's get the EZGL side merged first.

Thanks, sure. Please let me know if i need to do anything else about EZGL PR, i already updated it with respect of reviewer's comments.

AlexandreSinger:
> Thanks, sure. Please let me know if i need to do anything else about EZGL PR, i already updated it with respect of reviewer's comments.

I will review it right now!

w0lek:
> @w0lek You may need to perform a clean build. In VTR, when you type `make` in the terminal it will use the last CMake arguments you used to build. So if you turned off drawing or used different assert levels, you can get different results.
> 
> I recommend running:
> 
> ```shell
> make distclean
> make -j <num_cores>
> ```
> 
> This will hopefully put you in the same build environment as CI. Your machine matches our CI machines, so I am surprised to see such a large discrepancy. It is totally fine to bump the golden results manually in my opinion, but I do want to ensure that your environment is set up correctly so we can validate the results you collect! Unfortunately, this CMake caching stuff can make things a bit tricky.

I followed this instruction, but still have a gap in 25%. So for me currently no chance to fix it, except using manual values. 
Note: 
Manual values for such goldens are taken from the CI, which i also has a clean build. So i think environment is set correctly.

AlexandreSinger:
> I followed this instruction, but still have a gap in 25%. So for me currently no chance to fix it, except using manual values.
> Note:
> Manual values for such goldens are taken from the CI, which i also has a clean build. So i think environment is set correctly.

Thank you for trying it out! That is very strange. Yeah, manual update is fine for now as long as it came from the CI.

.github/workflows/containers.yml
```diff
@@ -39,8 +39,9 @@ jobs:
 
       - name: 🧰 Repository Checkout
         uses: actions/checkout@v6
-        with:
-          submodules: 'true'
+
+      - name: Fetch submodules
+        run: git submodule update --init --recursive
```
AlexandreSinger: Why was it necessary to recursively update the submodules? This is something we actually do not do on purpose since many submodules have submodules which we do not use.

This is also not the same steps as what we recommend in the quick start guide.

Is it possible to remove this for the CI tests?

.github/workflows/test.yml
```diff
@@ -65,7 +68,7 @@ jobs:
     - name: Pack Build
       shell: bash
       run: |
-        tar -czvf build.tar.gz --exclude='CMakeFiles' --exclude='*.a' --exclude='*.cmake' build/
+        tar -czvf build.tar.gz --exclude='CMakeFiles' --exclude='*.a' --exclude='cmake_install.cmake' build/
```
AlexandreSinger, line 76: What cmake files are being included here? The build directory is only copied over for the executables and their associated libraries; I am surprised that you would need cmake files in later tests.

.github/workflows/test.yml
```diff
@@ -32,15 +32,18 @@ concurrency:
 env:
   # default compiler for all non-compatibility tests
   MATRIX_EVAL: "CC=gcc-13 && CXX=g++-13"
+  # Qt6 install location — CMake reads CMAKE_PREFIX_PATH from the environment
+  CMAKE_PREFIX_PATH: /opt/qt6/6.9.3/gcc_64
```
AlexandreSinger: This concerns me a bit. The design of the CI is meant to mimic what a user might do when they first install VTR (following our quick-start guide). By adding this CMAKE_PREFIX_PATH, you are saying that in the quick-start, the user now needs to specify this path in their make command.

Can we make this automated using CMake? Or make it optional so people who need graphics can specify it?

.github/workflows/test.yml
```diff
@@ -225,9 +231,245 @@ jobs:
...
+        # case trips a pre-existing VTR_ASSERT_SAFE_MSG (level 3) in
+        # libs/libarchfpga/src/physical_types.cpp:419 grid_layout()
+        # ("Device layout not found.") — introduced upstream by
+        # f0b8ffdc4 ("call convert_interposer_cuts_to_sg_patterns() in
+        # RR graph generation"). The failure is unrelated to the GUI
+        # work in this PR and must be fixed in the libarchfpga grid
+        # layout lookup. The earlier qt_layer baseline silently passed
+        # this step because ctest reported "No tests were found!!!";
+        # our test-discovery fix in vpr/test/gui/CMakeLists.txt makes
+        # the latent assertion visible.
```
AlexandreSinger: Can we raise this as an issue so it can be resolved. We need FASM to be tested somehow and this is the only tests we have.

AlexandreSinger: If there is a way we could disable only part of the test, that would be ideal.

.github/workflows/test.yml
```diff
@@ -225,9 +231,245 @@ jobs:
...
+      env:
+        # libvpr links Qt6 when VPR_USE_EZGL=on; running test executables
+        # therefore needs the Qt 6.9.3 runtime that BuildVTR was linked
+        # against (installed via aqt to /opt/qt6/6.9.3/gcc_64). Without
+        # these the dynamic loader aborts before main() (test_fasm, etc.).
+        # CI=true is honoured by run_visual_regression.sh / test_save_graphics
+        # to skip image-comparison cases that require a real GL context.
+        LD_LIBRARY_PATH: /opt/qt6/6.9.3/gcc_64/lib
+        QT_PLUGIN_PATH: /opt/qt6/6.9.3/gcc_64/plugins
+        QT_QPA_PLATFORM: offscreen
```
AlexandreSinger: When a user wishes to use the graphics, do they always need to update their library path variable? What would this look like for an end-user of VPR?

.github/workflows/test.yml
```diff
@@ -225,9 +231,245 @@ jobs:
...
+  # ThickLineInstance, DashedLineInstance, Chunk) via std::construct_at
+  # because the standard library invokes _Tp(args...) rather than
+  # _Tp{args...}. That is a libezgl portability bug, tracked
+  # separately, and is unrelated to this test-stack PR. Linux is
+  # sufficient for the §7.9 gate; the macOS leg will be re-added once
+  # the libezgl ctors land. Each leg builds with -DVPR_USE_EZGL=on
+  # (the BuildVTR artifact is configured without ezgl, so it is not
+  # reused here) so the local and CI command lines are identical:
+  #
+  #     make CMAKE_PARAMS="-DVPR_USE_EZGL=on" gui-tests
```
AlexandreSinger, line 245: If the MacOs leg is currently unsupported, I think it would be best to remove its associated CI code until its needed (and is being tested). Having it in here and not used is confusing.

.github/workflows/test.yml
```diff
@@ -551,7 +796,7 @@ jobs:
         include: [
           {
             name: 'Basic',
-            params: '-DVTR_ASSERT_LEVEL=3 -DVTR_ENABLE_SANITIZE=on -DVTR_IPO_BUILD=off -DWITH_BLIFEXPLORER=on -DWITH_PARMYS=OFF -DWITH_ODIN=on',
+            params: '-DVTR_ASSERT_LEVEL=3 -DVTR_ENABLE_SANITIZE=on -DVTR_IPO_BUILD=off -DWITH_BLIFEXPLORER=on -DWITH_PARMYS=OFF -DWITH_ODIN=on -DVPR_USE_EZGL=off',
```
AlexandreSinger: Does graphics fail the sanitized builds? If it does, this needs to be tracked.

AlexandreSinger: We want our graphics code to also go under sanitized builds.

libs/EXTERNAL/CMakeLists.txt
```diff
@@ -8,6 +8,39 @@ add_subdirectory(libpugixml)
...
+# Patched in-place at configure time (idempotent). The submodule will
+# show as dirty in 'git status' — this is expected until upstream merges
+# the fix or we fork libsdcparse.
+set(_SDC_SWIG_FILE "${CMAKE_CURRENT_SOURCE_DIR}/libsdcparse/src/tcl/sdc_commands.i")
+file(READ "${_SDC_SWIG_FILE}" _sdc_swig_content)
+if(_sdc_swig_content MATCHES "\\$function;")
+    string(REPLACE "\$function;" "\$action;" _sdc_swig_content "${_sdc_swig_content}")
+    file(WRITE "${_SDC_SWIG_FILE}" "${_sdc_swig_content}")
+    message(STATUS "Patched libsdcparse sdc_commands.i: \$function -> \$action (SWIG 4.x compat)")
+endif()
```
AlexandreSinger: Why was this needed? LibSDCParse is 100% managed by us. If there is an issue with it we should fix this before this gets merged in. Also, I am not sure why this would be needed because our current version of SWIG in the CI is working.

We really do not want the dirty git status issues. It makes VTR a bit rough to work with. If there is a fix for LibSDCParse, lets get that in first.

libs/EXTERNAL/CMakeLists.txt
```diff
@@ -8,6 +8,39 @@ add_subdirectory(libpugixml)
...
+        string(REPLACE "${_old}" "${_new}" _sdc_cml_content "${_sdc_cml_content}")
+        file(WRITE "${_SDC_CML}" "${_sdc_cml_content}")
+        message(STATUS "Patched libsdcparse CMakeLists.txt: TCL/sdc_commands link order (MinGW single-pass ld)")
+    endif()
+    message(WARNING
+        "VTR is patching libsdcparse/CMakeLists.txt at configure time to fix "
+        "MinGW link order (sdc_commands must precede \${TCL_LIBRARY}). "
+        "TODO: upstream this to verilog-to-routing/libsdcparse and drop the "
+        "patch in libs/EXTERNAL/CMakeLists.txt.")
+endif()
```
AlexandreSinger: Same comment here.

.github/workflows/test.yml
```diff
@@ -225,9 +231,245 @@ jobs:
...
+        # against (installed via aqt to /opt/qt6/6.9.3/gcc_64). Without
+        # these the dynamic loader aborts before main() (test_fasm, etc.).
+        # CI=true is honoured by run_visual_regression.sh / test_save_graphics
+        # to skip image-comparison cases that require a real GL context.
+        LD_LIBRARY_PATH: /opt/qt6/6.9.3/gcc_64/lib
+        QT_PLUGIN_PATH: /opt/qt6/6.9.3/gcc_64/plugins
+        QT_QPA_PLATFORM: offscreen
       run: |
         cd build
-        make test -j${{ steps.cpu-cores.outputs.count }}
```
AlexandreSinger, line 235: make test is our standard way to run the Catch2 tests in VTR. If it is broken now, we need to fix it or update the documentation with the new way of running it.

.gitmodules
```diff
@@ -22,3 +23,4 @@
 [submodule "libs/EXTERNAL/libsdcparse"]
 	path = libs/EXTERNAL/libsdcparse
 	url = https://github.com/verilog-to-routing/libsdcparse.git
+	ignore = dirty
```
AlexandreSinger: As I mentioned in my above comment, we do not want to mark this as dirty.

.gitmodules
```diff
@@ -10,6 +10,7 @@
 [submodule "libs/EXTERNAL/libezgl"]
 	path = libs/EXTERNAL/libezgl
 	url = https://github.com/verilog-to-routing/ezgl.git
+	branch = qt_layer
```
AlexandreSinger: Reminder to update this to the default branch.

build_vpr_qt6.sh
AlexandreSinger, line 1: Instead of using a bash file for this, I am wondering if it would be cleaner to add this to our Makefile. People tend to build VPR using that Makefile, so adding a special make target for this may make sense. If not, we will need to document this bash file carefully.

BUILDING.md
```diff
@@ -30,10 +51,21 @@ From the top-level, run:
...
+    # macOS
+    make -j$(sysctl -n hw.ncpu)
+
+### Supported Platforms
 
-The complete VTR flow has been tested on 64-bit Linux systems.
-The flow should work in other platforms (32-bit Linux, Windows with cygwin) but this is untested.
+The complete VTR flow has been tested on:
+- 64-bit Linux (Debian, Ubuntu, Fedora)
+- macOS (Apple Silicon and Intel, with Homebrew)
```
AlexandreSinger: Has this been tested? Including Parmys?

install_apt_packages.sh
```diff
@@ -64,6 +97,110 @@ else
     echo "clang-format-18 not found in apt-cache. Skipping installation."
 fi
 
-sudo apt-get update
-sudo apt-get install -y "${packages_to_install[@]}"
+# Required for GUI test infrastructure (Layer 7 coverage tooling).
+packages_to_install+=(
+    gcovr
+)
```
AlexandreSinger: Whatever gets added to this file is added for all users of VTR. This gcovr is only used by a single CI test. If that is the case, I think that the CI test should install this package. It should not be in this file.

install_apt_packages.sh
```diff
@@ -64,6 +97,110 @@ else
...
+
+echo ""
+echo "Setting up Python virtualenv at ${VENV_DIR}..."
+if [ ! -d "${VENV_DIR}" ]; then
+    python3 -m venv "${VENV_DIR}"
+fi
+# shellcheck disable=SC1091
+source "${VENV_DIR}/bin/activate"
+pip install --upgrade pip
+pip install -r "${SCRIPT_DIR}/requirements.txt"
```
AlexandreSinger: You are changing how people normally build VTR. Normally this is done in an external step. I think it may be too much to make all of these changes in a single PR. I think it may be best not to bring this in now.

Also the packages that you add below are only used for the CI test as far as I can tell. They should be brought into the CI.

install_apt_packages.sh
```diff
@@ -64,6 +97,110 @@ else
...
+echo ""
+echo "  Qt 6.9.3 is at: ${QT_PREFIX}/6.9.3/gcc_64"
+echo ""
+echo "  Before building or running GUI tests in any new shell:"
+echo "    export QT6_HOME=\"${QT_PREFIX}/6.9.3/gcc_64\""
+echo "    export PATH=\"\$QT6_HOME/bin:\$PATH\""
+echo "    export LD_LIBRARY_PATH=\"\$QT6_HOME/lib:\${LD_LIBRARY_PATH:-}\""
+echo "    export CMAKE_PREFIX_PATH=\"\$QT6_HOME:\${CMAKE_PREFIX_PATH:-}\""
+echo "    export QT_PLUGIN_PATH=\"\$QT6_HOME/plugins\""
+echo "    source ${VENV_DIR}/bin/activate"
```
AlexandreSinger: Every user of VTR will be seeing this message. This is very likely to confuse them because they would not know what this means for using the graphics normally.

If this is only for running graphics tests, I think we can keep this in a developer doc.

install_brew_packages.sh
AlexandreSinger, line 1: This PR is bringing in an entirely new graphics system into VTR. It may be a lot to also bring in Mac support. I do not recall seeing a CI test for Mac support yet, so I am worried about adding this files in without proper test coverage.

I recommend separating out the Mac support from this PR if it is not working yet.

Makefile
```diff
@@ -54,6 +57,13 @@ endif
 # --output-sync target : For parallel compilation ensure output for each target is synchronized (make version >= 4.0)
 MAKEFLAGS := -s
 
+#Detect number of CPU cores for parallel build suggestions
+ifeq ($(UNAME_S),Darwin)
+    NUM_CORES := $(shell sysctl -n hw.ncpu 2>/dev/null || echo 1)
+else
+    NUM_CORES := $(shell nproc 2>/dev/null || echo 1)
+endif
```
AlexandreSinger: We do not want VTR to use multiple cores by default. VTR builds should be single core unless requested in my opinion. If this changes this behaviour, I do not think it should be added.

AlexandreSinger (changes requested):
Hi @w0lek ,

I looked specifically at the CI and the updates to the build scripts. I had a lot of comments that I think must be addressed before this can come in. Let me know if you have any questions and we can discuss.

You are making a very large change to VTR's graphics, and this PR also brings a lot of changes to how VTR is being built and many assumptions. For things like MacOS support and updates to how python libraries are instralled, they are probably better served in a separate PR so they can be discussed. I recommend bringing in the minimal number of changes needed to move to the new Qt backend, and anything that are "nice-to-haves" should probably move to another PR.

w0lek:
Hi @AlexandreSinger , thank you for reviewing this.
Give me time to minimize diff in CI. Also i will collect and review all changes for MacOS to decide what to do with this (move out to separate PR or so). These both parts were made not by me, so i need some time to analyze it. And ofcourse i would think about easy way of building using Qt6, using make target as u suggested, os something like this.
Let me process you comments first with more attention.

.github/workflows/containers.yml
w0lek: Alexandre the most CI changes originally was made not by me, so i couldn't exactly explain why this changes was made. My strategy will be revert CI changes to master, and check if gui testing framework still works.

.github/workflows/test.yml
w0lek: > What cmake files are being included here?

I believe it's for CTestTestfile.cmake, which is needed to keep test tree.
https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html
"CTest is run from within your build tree. It will run all the tests found in the current directory as well as any subdirectories listed in the CTestTestfile.cmake."

install_apt_packages.sh
w0lek: yes this extra step is only for CI tests, i will move them out

.github/workflows/containers.yml
w0lek: SOLVED
removed, behavior matches to upstream

.github/workflows/test.yml
w0lek: SOLVED
Usage of CMAKE_PREFIX_PATH has been removed.

If a custom Qt6 installation is used, its location is embedded into the binary via RPATH, allowing the executable to locate the required Qt libraries automatically.

If Qt6 is provided by the operating system, RPATH is not used, since system libraries are already discoverable through the OS's standard library search paths.

.github/workflows/test.yml
w0lek: not solved yet, by missing CTestTestfile.cmake might lost test cases to execute

.github/workflows/test.yml
w0lek: SOLVED
Usage of LD_LIBRARY_PATH/QT_PLUGIN_PATH/QT_QPA_PLATFORM for Qt has been removed.

If a custom Qt6 installation is used, its location is embedded into the binary via RPATH, allowing the executable to locate the required Qt libraries automatically.

If Qt6 is provided by the operating system, RPATH is not used, since system libraries are already discoverable through the OS's standard library search paths.

QT_PLUGIN_PATH is picked up using qt.conf in qt sdk for Qt6 custom.

QT_QPA_PLATFORM: offscreen is set programmatically if vpr runs with "--disp off"

.github/workflows/test.yml
w0lek: SOLVED.
macos placeholders for CI was removed.
will be re-added as separate issue/PR in future if needed

libs/EXTERNAL/CMakeLists.txt
w0lek: > Why was this needed? LibSDCParse is 100% managed by us. If there is an issue with it we should fix this before this gets merged in. Also, I am not sure why this would be needed because our current version of SWIG in the CI is working.
> 
> We really do not want the dirty git status issues. It makes VTR a bit rough to work with. If there is a fix for LibSDCParse, lets get that in first.

SOLVED.
removed. 
you are right, the changes in upstream. it looks when i was doing windows check and fixes i was used older master(vpr-fork-masker being used in Aurora), where this cause issue on windows build. I did this via configure patch, because wasn't sure where the changes must be settled, so i kept changes in vpr source tree.

libs/EXTERNAL/CMakeLists.txt
w0lek: SOLVED.
patch removed, since upstream already has that change.

.gitmodules
w0lek: RESOLVED.
ignore = dirty removed from libsdcparse

.gitmodules
w0lek: SOLVED
set master branch for ezgl

build_vpr_qt6.sh
w0lek: SOLVED.
```
make qt6sdk
```
will install qt6sdk into ${repo_root}/qt6.
CMakeLists.txt automatically search this path and pick suitable(latest + >6.9.3) version.
So user don't have to set any variables

BUILDING.md
w0lek: don't know if Parmys was tested.
i revert BUILDING.md, close to original, no mention about MacOS.
Also few blocks in CMakeLists.txt for macos was also removed.
MacOS support will be pushed later by other issue/PR.

w0lek: SOLVED

install_apt_packages.sh
w0lek: RESOLVED.

gcovr now is a part of dev/install_test_deps.sh, as the rest of deps which are needed only for tests

install_apt_packages.sh
w0lek: SOLVED

these deps now part of dev/install_test_deps.sh

install_apt_packages.sh
w0lek: SOLVED.

This information now is redundant and removed.

install_brew_packages.sh
w0lek: SOLVED. 

Agree. MacOS support is removed from configuration or docs. 
Will be part of separate issue/PR not connected to Qt migration.

Makefile
w0lek: SOLVED.
remove that section, var NUM_CORES wasn't in use, so it was dead code.

.github/workflows/test.yml
w0lek: SOLVED.

i am not really sure why this changes has been made, so i will revert it to match master.
so sanity check runs with ezgl included

.github/workflows/test.yml
w0lek: reverted to upstream

w0lek: SOLVED

w0lek:
@AlexandreSinger i fixed all you comments from last your review.
In case if you want have a look to diff code of changes based on you last review comments only (i did this in separate branch qt_layer_PR_fixes of public fork), you may check fork PR [qt_layer_PR_fixes -> qt_layer] with clean diff
https://github.com/w0lek/vtr-verilog-to-routing/pull/16/changes

.github/workflows/test.yml
```diff
@@ -32,22 +32,27 @@ concurrency:
...
   BuildVTR:
     name: 'B: Building VTR Release'
     runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v6
-      with:
-        submodules: 'true'
+
+    - name: Fetch submodules
+      run: git submodule update --init --recursive
```
AlexandreSinger: There are still these recursive submodule fetching throughout the CI. Are these necessary for the Qt upgrade?

.github/workflows/test.yml
```diff
@@ -235,6 +243,175 @@ jobs:
         make test -j${{ steps.cpu-cores.outputs.count }}
 
 
+  # GUITests — Plan §7.9 — runs the full GUI test pyramid on ubuntu-24.04.
+  # Builds with -DVPR_USE_EZGL=on so the local and CI command lines match:
+  #
+  #     make CMAKE_PARAMS="-DVPR_USE_EZGL=on" gui-tests
+  GUITests:
+    name: 'T: GUI Tests (Layers 1+3+4+5)'
```
AlexandreSinger, line 247: This GUI test is currently on the critical path of the CI:

Image

This and the other GUI test form a dependency chain which add up to around 2 hours. The overall CI takes around 30 mins (sometimes 45 mins if the caches were evicted).

The CI GUI runs are taking a while for different reasons. For the longer of the two, it looks like it is building twice for some reason:

Image

Even though the actual test itself takes under a minute:

Image

One solution to this problem is to use the packaged build artifact from the release build. That way you do not have to build the executable again for this test. Another alternative is you can cache the build using CCache similar to the other builds. It looks like this test tries to use caching, so you may need to try running the CI again and see if it is getting cached correctly. We would prefer to keep the CI runtime reasonable if we can. I would not want to block this feature due to this though, so we can move this to another PR if we do not want this to hold back the Qt upgrade.

@vaughnbetz Let me know if you have any comments on this.

dev/install_test_deps.sh
AlexandreSinger, line 1: This script appears to only be used to install the GUI test dependencies. Could you rename this file accordingly to keep it from getting confusing. Same with `test_requirements.txt` below.

AlexandreSinger (commented):
I went through the CI and CMake files, I have not gone through the code in its entirety yet. The interface for building VPR is looking a lot cleaner now, thank you very much for updating that! The most urgent comment below is the length of time it takes to run the GUI CI jobs.

kevinxing-a12:
Hi @w0lek, I cloned your "qt_layer" branch and explicitly followed the new quickstart page to build and run, and there are two issues I found. First, the quickstart does not tell the user to install the Qt6 packages. Amir and I believe that mentioning "install_build_deps.sh" would be nice (one trivial thing: on our machine, "install_build_deps.sh" requires sudo but "install_qt6_sdk.sh" does not. So maybe you could list both, just in case?). Second, I ran "install_qt6_sdk.sh" and did another "make", but I was shown this:

After the build I was able to run a successful VPR flow (great news!), but the GUI never showed up with "--disp on", which was expected.
So, in brief, using VPR in fully command-line style isn't a problem at all. It's just the GUI.
@vaughnbetz @AmirhosseinPoolad @AlexandreSinger FYI

w0lek:
Hi @kevinxing-a12 
Thanks for testing it and for feedback!

>First, the quickstart does not tell the user to install the Qt6 packages.

You are right! This is corrected now.

>Amir and I believe that mentioning "install_build_deps.sh" would be nice (one trivial thing: on our machine, "install_build_deps.sh" requires sudo but "install_qt6_sdk.sh" does not.

install_qt6_sdk.sh itself doesn't require root, but since it's called from install_build_deps.sh which also runs apt install  commands, this part requires the root.
Anyway, i had to delete install_build_deps.sh, because on BUILDING.md i found mention that we also have a script to install Fedora packages. So in order to not having many scripts from different distros, + ensure_qt6_sdk.sh (renamed from install_qt6_sdk.sh) is common for any distro, i descide to keep things splitted. So we need prepeare our env in two steps instead of 1.
1) install_apt_packages.sh
2) ensure_qt6_sdk.sh

But for simplification purpose, i added new Makefile targets, to explicitly build vpr in gui and headless mode.
```make ensure-gui```
```make ensure-headless```
because if we run simply 
```make```
we have not deterministic behaviour (gui is on only if qt6 will be found, and off automatically if it's not found)

```make ensure-gui```
Will automatically runs ./ensure_qt6_sdk.sh and fail if it will be some problem with qt6 installation.
So u will not get situation when you wanted to have gui app, but it silently dropped to headless mode (non-qt6 variant)

```make ensure-headless```
is added mainly based on tangxifan@gmail.com question on the meeting. By executing this command, even if qt6 is installed in our system, vpr will be build in headless mode.

All this information is reflected in BUILDING.md page as well. But please check it again, maybe you will have idea how to make it better.

>Second, I ran "install_qt6_sdk.sh" and did another "make", but I was shown this:  After the build I was able to run a successful VPR flow (great news!), but the GUI never showed up with "--disp on", which was expected. So, in brief, using VPR in fully command-line style isn't a problem at all.

I see, here i see two issues: 
1) mention about `make qt6sdk` is obsolete now and was removed.
2) if you run script ensure_qt6_sdk.sh (old install_qt6_sdk.sh), and qt6 in some reason is not installed, we should debug it. I slightly enhance the script, but if it still didn't install qt6, could you share you OS name/version and script output? Because for me and for CI the old script was working fine.

You may pull qt_layer branch, the changes should be there.
If you are willing to change the diff code since you last check here is full diff https://github.com/w0lek/vtr-verilog-to-routing/pull/17/changes

**Comment by w0lek** [comment]
@AlexandreSinger while i was fixing comments from Kevin i also did optimization in CI:
1) qt6 installation is cached
2) GUI Coverage job in test.yml is disabled (it was consume ~1h of runtime and not needed for PR).
But i didn't measure final time numbers.

@AlexandreSinger may you also please review carefully changes i made in BUILDING.md and Makefile.
There are two new make targets:
```
make ensure-gui           # force GUI ON, configuration fails if it didn't find proper qt6 version
make ensure-headless # force GUI OFF, even if we have qt6 sdk, we build vpr without gui
```
behavior of legacy ```make``` without target specifying remains non deterministic (gui is on if qt6 is found, otherwise it automatically off)

Note: i revert approach from prev changes to use generic install_build_deps.sh script, because i found that i have to add special script for fedora/rhat as well. 
+in case if somebody don't want to have qt6 gui, he could just not run ./ensure_qt6_sdk.sh. And it's easy to debug things, as seems Kevin has a problem with that script on his system.

Here is full changelog
https://github.com/w0lek/vtr-verilog-to-routing/pull/17/changes

**Comment by kevinxing-a12** [comment]
Hi @w0lek, I ran `ensure_qt6_sdk.sh` but it seemed that the installation still didn't take place. I also got a very weird and contradictory output:

> System Qt6 (none found) does not satisfy >= 6.9.3; provisioning via aqt...

> Qt 6.9.3 already present at home/user/vtr-verilog-to-routing/qt6/6.9.3/gcc_64 - nothing to do. 

As a result, `make ensure-gui` didn't succeed. I did try `make ensure-headless` and performed a successful VPR flow, so that one is good.
For your reference, our OS is Ubuntu 24.04.2 LTS.
One last kind reminder, you might want to check if the Qt6 installation info has been reflected in https://vtr--3542.org.readthedocs.build/en/3542/quickstart/ :)

**Comment by w0lek** [comment]
Hi @kevinxing-a12 
>I also got a very weird and contradictory output

These two doesn't have contradiction
> System Qt6 (none found) does not satisfy >= 6.9.3; provisioning via aqt...

This means Qt6 package shipped by ubuntu apt package manager (System qt6) are not found, and it goes to custom qt6 sdk (not system one), which is shipped by aqt. It's failback path of installing qt6 with proper version, if standart package manager of OS couldn't provide us proper version.

> Qt 6.9.3 already present at home/user/vtr-verilog-to-routing/qt6/6.9.3/gcc_64 - nothing to do.

And this message is probably when u run ./ensure_qt6_sdk.sh second time, when it detects that custom qt6 sdk is already installed and skip it's installation

Is it possible for me to have a look at full build log, when ```make ensure-gui``` is executed and fail?
This would help me detect and illuminate problem.

Also i will probably will add ```make ensure-gui-clean``` target in order to forcely reinstall custom qt6 sdk, if installation was interrupted by some reason. To handle the case when qt6 sdk might be presented but broken (interrupted installation procedure etc)

>One last kind reminder, you might want to check if the Qt6 installation info has been reflected in

Sorry for this, i did update on BUILDING.md, thought it would reflect automatically, i double check this later, and will update it in a proper place

**Comment by kevinxing-a12** [comment]
@w0lek Sure! Let me know if this contains all you need:

> System Qt6 (none found) does not satisfy >= 6.9.3; provisioning via aqt...
Qt 6.9.3 already present at /home/user/vtr-verilog-to-routing/qt6/6.9.3/gcc_64 - nothing to do.
Performing standard build...
cd build && /usr/bin/cmake -DCMAKE_BUILD_TYPE=release -G 'Unix Makefiles'  -DVPR_USE_EZGL=on /home/user/vtr-verilog-to-routing
-- CMAKE_BUILD_TYPE: release
-- Building with IPO: on (auto)
-- Performing Test CXX_COMPILER_SUPPORTS_-Wformat=2
-- Performing Test CXX_COMPILER_SUPPORTS_-Wformat=2 - Success
-- Performing Test CXX_COMPILER_SUPPORTS_-Wcatch-value=3
-- Performing Test CXX_COMPILER_SUPPORTS_-Wcatch-value=3 - Success
-- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=3
-- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=3 - Success
-- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=0
-- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=0 - Success
-- IPO: Suppressing known VTR warnings.
CMake Error at CMakeLists.txt:481 (message):
  VPR_USE_EZGL=on but Qt6 >= 6.9.3 was not found.  Run
  ./dev/ensure_qt6_sdk.sh to provision a repo-local Qt SDK, install a system
  Qt6 >= 6.9.3, or pass -DCMAKE_PREFIX_PATH=<qt-prefix>.
-- Configuring incomplete, errors occurred!
make[1]: *** [Makefile:110: vpr] Error 1
make: *** [Makefile:153: ensure-gui] Error 2

**Comment by AlexandreSinger** [comment]
> @AlexandreSinger while i was fixing comments from Kevin i also did optimization in CI:
> 
>     1. qt6 installation is cached
> 
>     2. GUI Coverage job in test.yml is disabled (it was consume ~1h of runtime and not needed for PR).
>        But i didn't measure final time numbers.
> 
> 
> @AlexandreSinger may you also please review carefully changes i made in BUILDING.md and Makefile. There are two new make targets:
> 
> ```
> make ensure-gui           # force GUI ON, configuration fails if it didn't find proper qt6 version
> make ensure-headless # force GUI OFF, even if we have qt6 sdk, we build vpr without gui
> ```
> 
> behavior of legacy `make` without target specifying remains non deterministic (gui is on if qt6 is found, otherwise it automatically off)
> 
> Note: i revert approach from prev changes to use generic install_build_deps.sh script, because i found that i have to add special script for fedora/rhat as well. +in case if somebody don't want to have qt6 gui, he could just not run ./ensure_qt6_sdk.sh. And it's easy to debug things, as seems Kevin has a problem with that script on his system.
> 
> Here is full changelog https://github.com/w0lek/vtr-verilog-to-routing/pull/17/changes

Hi @w0lek , I really like these changes! I agree with your comment about skipping the long test for now (I agree that the signal was low for the test time). We can decide to move it to the weekly test in another PR. The other graphics test is no longer taking the longest amount of time, which is excellent. It looks like it is having trouble caching for some reason, but that can be resolved in the future.

Regarding the new targets, I think it is a good idea separating the install deps and the gui install scripts. Once we move to Ubuntu-26.04 that second script will no longer be needed; which I think is a very clean way of handling this.

**Comment by AlexandreSinger** [comment]
Once the current in-flight comments are resolved (ensure that an end user can easily get this code working on their local machine and the install sub-module recursive calls), then I can give this PR a full review.

File: vpr/src/base/read_options.cpp [hunk 20]
```diff
@@ -1599,6 +1606,12 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .help("Save all graphical contents to PDF files")
         .default_value("off");
 
+    gfx_grp.add_argument(args.graphics_renderer, "--renderer")
```
vaughnbetz, line 1655: I'd give a little more information here.  I suggest adding another sentence saying: "rhi gives the highest performance but requires a GPU; deferred gives the next highest performance and will work without a GPU."

vpr/src/base/read_options.cpp
```diff
@@ -1583,7 +1583,14 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
...
     gfx_grp.add_argument<bool, ParseOnOff>(args.show_graphics, "--disp")
-        .help("Enable or disable interactive graphics")
+        .help(
+            "Enable or disable interactive graphics."
+            " When 'off' and QT_QPA_PLATFORM is not already set, VPR sets"
+            " QT_QPA_PLATFORM=offscreen so Qt does not try to connect to"
+            " an X11/Wayland display. Note that the offscreen platform"
+            " typically disables the RHI (GPU) renderer; rendering falls"
+            " back to the QPainter (immediate) path. To override, set"
+            " QT_QPA_PLATFORM in the environment before invoking VPR.")
```
vaughnbetz: Is it possible to give the user context on what offscreen will mean?  Will it still render any scripted graphics?  I'd  say that at the start of the second sentence, since otherwise the discussion of what exact method will be used may be baffling to the end user.

vpr/src/base/vpr_types.h
```diff
@@ -1619,6 +1619,7 @@ struct t_vpr_setup {
     int GraphPause;                      ///<user interactiveness graphics option
     bool SaveGraphics;                   ///<option to save graphical contents to pdf, png, or svg
     std::string GraphicsCommands;        ///<commands to control graphics settings
+    std::string RendererType;            ///<rendering backend: "immediate", "deferred", or "rhi"
```
vaughnbetz: Suggest adding: "rhi requires a GPU and is the highest performance; deferred is the next most performant and does not need a GPU."

vpr/src/draw/draw.cpp
```diff
@@ -131,15 +144,39 @@ const std::vector<ezgl::color> kelly_max_contrast_colors = {
...
+ezgl::application* application = nullptr;
 
 bool window_mode = false;
 bool window_point_1_collected = false;
 ezgl::point2d point_1(0, 0);
+ezgl::point2d window_preview_cursor(0, 0); // updated by act_on_mouse_move while
+                                           // window_point_1_collected is true; drawn
+                                           // as a dashed rect by the regular draw flow.
 ezgl::rectangle initial_world;
 std::string rr_highlight_message;
```
vaughnbetz, line 161: Suggest mentioning at the start of these two comments that these variables are used for scripted graphics (rendered to files).

vpr/src/draw/draw.cpp
```diff
@@ -1035,181 +1226,242 @@ ClusterBlockId get_cluster_block_id_from_xy_loc(double x, double y) {
...
+    application->update_message(draw_state->default_message);
+    application->refresh_drawing();
 }
 
-static void set_force_pause(GtkWidget* /*widget*/, gint /*response_id*/, gpointer /*data*/) {
+static void set_force_pause() {
     t_draw_state* draw_state = get_draw_state_vars();
 
     draw_state->forced_pause = true;
 }
```
vaughnbetz, line 1393: Should have some kind of comment saying these are values used in graphics scripts.

vpr/src/draw/draw.cpp
```diff
@@ -1035,181 +1226,242 @@ ClusterBlockId get_cluster_block_id_from_xy_loc(double x, double y) {
...
+    while (draw_state->graphics_cmd_cursor < draw_state->parsed_graphics_cmds.size()) {
+        auto& cmd = draw_state->parsed_graphics_cmds[draw_state->graphics_cmd_cursor];
         VTR_ASSERT_MSG(cmd.size() > 0, "Expect non-empty graphics commands");
 
         for (auto& item : cmd) {
             VTR_LOG("%s ", item.c_str());
         }
         VTR_LOG("\n");
 
+        if (cmd[0] == "wait_for_stage") {
```
vaughnbetz, line 1489: Long routine; consider if it could/should be broken up with some helper functions.

vpr/src/draw/draw_types.h
```diff
@@ -299,6 +299,20 @@ struct t_draw_state {
 
     std::string graphics_commands;
 
+    ///@brief Tokenized form of `graphics_commands` (one inner vector per
+    /// `;`-separated command). Built lazily by run_graphics_commands on the
+    /// first call (when this vector is still empty); `graphics_commands` is
+    /// set once at init and never mutated, so a single parse is sufficient.
+    std::vector<std::vector<std::string>> parsed_graphics_cmds;
+
+    ///@brief Cursor into `parsed_graphics_cmds`. Persists across
```
vaughnbetz: Explain what the exact cursor is; index into the outer vector (i.e. the command index)?

vpr/src/draw/draw_types.h
```diff
@@ -299,6 +299,20 @@ struct t_draw_state {
 
     std::string graphics_commands;
 
+    ///@brief Tokenized form of `graphics_commands` (one inner vector per
```
vaughnbetz: mention this is 'for graphics scripting'

vpr/test/gui/README.rst
```diff
@@ -0,0 +1,199 @@
...
+
+**Debug mode (write a diff PNG for every visual case):**
+
+.. code-block:: bash
+
+   ./vpr/test/gui/run_all_tests.sh --debug \
+       ./build/vpr/vpr \
+       ./vtr_flow/arch/timing \
+       ./vtr_flow/benchmarks/microbenchmarks
+
```
vaughnbetz, line 85: We might need a script option that lets us quickly update all the golden visual results when we make an intentional change to how we show placement etc. (for example). Don't need to do that now, but something to think about once this is merged.

vaughnbetz (changes requested):
Looks very good, thanks. I have a few suggested changes (all minor commenting suggestions / requests). Other than that, please update the vtr quick start documentation if you haven't already so it specifies how to build (easiest commands you can come up with) so users don't have to read the more detailed documentation.

w0lek:
> @w0lek Sure! Let me know if this contains all you need:
> 
> > System Qt6 (none found) does not satisfy >= 6.9.3; provisioning via aqt...
> > Qt 6.9.3 already present at /home/user/vtr-verilog-to-routing/qt6/6.9.3/gcc_64 - nothing to do.
> > Performing standard build...
> > cd build && /usr/bin/cmake -DCMAKE_BUILD_TYPE=release -G 'Unix Makefiles'  -DVPR_USE_EZGL=on /home/user/vtr-verilog-to-routing
> > -- CMAKE_BUILD_TYPE: release
> > -- Building with IPO: on (auto)
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wformat=2
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wformat=2 - Success
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wcatch-value=3
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wcatch-value=3 - Success
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=3
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=3 - Success
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=0
> > -- Performing Test CXX_COMPILER_SUPPORTS_-Wimplicit-fallthrough=0 - Success
> > -- IPO: Suppressing known VTR warnings.
> > CMake Error at CMakeLists.txt:481 (message):
> > VPR_USE_EZGL=on but Qt6 >= 6.9.3 was not found.  Run
> > ./dev/ensure_qt6_sdk.sh to provision a repo-local Qt SDK, install a system
> > Qt6 >= 6.9.3, or pass -DCMAKE_PREFIX_PATH=.
> > -- Configuring incomplete, errors occurred!
> > make[1]: *** [Makefile:110: vpr] Error 1
> > make: *** [Makefile:153: ensure-gui] Error 2

Hi @kevinxing-a12 thank you.
This doesn't give me more information, unfortunatelly, but anyway it give me point to think how to improve the qt6 sdk validation script. i will push latest updated script, which has some sanity check for local qt6sdk, and wipe it out if sanity check fail, in order to proceed with fresh new install.

w0lek:
@vaughnbetz @kevinxing-a12  @AlexandreSinger 
i push corrections into qt_layer branch, clean diff is here in case you want review changes were made since last review request:
https://github.com/w0lek/vtr-verilog-to-routing/pull/18/changes

FYI @kevinxing-a12 
If u still have a problem with building graphics in vtr, may you also please share:
1) cmake log output (like you already did before)
2) build/CMakeCache.txt
3) ensure_qt6_sdk.log

kevinxing-a12:
Hi @w0lek, so here is how I tested it again:
I ran "make ensure-gui", but still got a failure saying that no Qt was found.
[cmake log output (unsuccessful).txt](https://github.com/user-attachments/files/28885492/cmake.log.output.unsuccessful.txt)
I then ran "make distclean", assuming something in the cache might be playing a trick. After that I ran "make ensure-gui", and no failure about Qt was reported again!
However, two things to note: there was still a warning regarding Qt (doesn't seem to be a real issue, but just so you know):
```
CMake Warning at qt6/6.9.3/gcc_64/lib/cmake/Qt6/QtPublicDependencyHelpers.cmake:335 (message):
  This project is using headers of the GuiPrivate module and will therefore
  be tied to this specific Qt module build version.  Running this project
  against other versions of the Qt modules may crash at any arbitrary point.
  This is not a bug, but a result of using Qt internals.  You have been
  warned!

  You can disable this warning by setting QT_NO_PRIVATE_MODULE_WARNING to ON.
Call Stack (most recent call first):
  qt6/6.9.3/gcc_64/lib/cmake/Qt6/Qt6Config.cmake:230 (_qt_internal_show_private_module_warning)
  libs/EXTERNAL/libezgl/CMakeLists.txt:106 (find_package)
```
Also, the compilation still did not succeed, but the errors this time were about sdc. @AlexandreSinger Would you mind taking a look?
[cmake log output (half successful).txt](https://github.com/user-attachments/files/28885566/cmake.log.output.half.successful.txt)

kevinxing-a12:
Sorry, regarding the sdc failure, it was a submodule update mistake that I forgot to handle. I just got a successful compilation!

w0lek:
> Hi @w0lek, so here is how I tested it again: I ran "make ensure-gui", but still got a failure saying that no Qt was found. [cmake log output (unsuccessful).txt](https://github.com/user-attachments/files/28885492/cmake.log.output.unsuccessful.txt) I then ran "make distclean", assuming something in the cache might be playing a trick. After that I ran "make ensure-gui", and no failure about Qt was reported again! 

Hi @kevinxing-a12. Thank you for your efforts!
It looks like this is cache problem in CMake, due to we tried different approaches/scripts on same build dir. If i would have CMakeCache.txt i would say for sure what the problem was. But i reviewed part in CMake responsible for Qt6 finding and did some enhancements which should cover this stale cache problem in future, so we don't have to wipe out build dir in future.

>However, two things to note: there was still a warning regarding Qt (doesn't seem to be a real issue, but just so you know):

This is not critical, but i masked warning, so user won't see it.
I am going to push these changes when CI becomes green.

kevinxing-a12:
@w0lek The CMakeCache file for the first unsuccessful build is here:
[CMakeCache.txt](https://github.com/user-attachments/files/28887187/CMakeCache.txt)
Two more problems found as I followed the quickstart:
When I tried to run the VTR flow using the following command:
```
$VTR_ROOT/vtr_flow/scripts/run_vtr_flow.py \
    $VTR_ROOT/doc/src/quickstart/blink.v \
    --route_chan_width 100
```
it was reported that yosys couldn't be found.
I suspected that something might be overseen by `make ensure-gui`. So I did a regular `make` and ran the same thing, and the problem was gone.
Second, the VPR flow with `--disp on` didn't succeed, and the log is here:
[VPR log.txt](https://github.com/user-attachments/files/28887995/VPR.log.txt)
It seems that some plugins aren't configured properly.

kevinxing-a12:
@w0lek I also tried the `--renderer deferred` and `--renderer immediate` options. Neither worked. Also FYI, I was using VNC.

w0lek:
@kevinxing-a12 thanks for finding!

> @w0lek The CMakeCache file for the first unsuccessful build is here: [CMakeCache.txt](https://github.com/user-attachments/files/28887187/CMakeCache.txt) Two more problems found as I followed the quickstart: When I tried to run the VTR flow using the following command:
> 
> ```
> $VTR_ROOT/vtr_flow/scripts/run_vtr_flow.py \
>     $VTR_ROOT/doc/src/quickstart/blink.v \
>     $VTR_ROOT/vtr_flow/arch/timing/EArch.xml \
>     --route_chan_width 100
> ```
> 
> it was reported that yosys couldn't be found. I suspected that something might be overseen by `make ensure-gui`. 
>So I did a regular `make` and ran the same thing, and the problem was gone. 

this is bug, ensure-gui builds vpr target but should build all. i am fixing it.


>Second, the VPR flow with `--disp on` didn't succeed, and the log is here: [VPR log.txt](https://github.com/user->attachments/files/28887995/VPR.log.txt) It seems that some plugins aren't configured properly.

it seems we are missing some system dependencies required by Qt. xcb-cursor0 or libxcb-cursor0 . let me correct this

doc/src/quickstart/index.rst
```diff
@@ -91,6 +91,8 @@ On most unix-like systems you can run:
 
 For more details on building VTR on various operating systems/platforms see :doc:`Building VTR</BUILDING>`.
 
+.. note:: A plain ``make`` builds VPR headless unless a suitable Qt6 (>= 6.9.3) is present. To use the interactive GUI (``--disp on``, used in the visualization steps below) you must build VPR with graphics support — see :ref:`VPR Graphics <vpr_graphics>`.
```
vaughnbetz: Explain headless means no graphics

doc/src/vpr/graphics.rst
```diff
@@ -12,22 +12,47 @@ Enabling Graphics
 
 Compiling with Graphics Support
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The build system will attempt to build VPR with graphics support by default.
+VPR's graphics are built on `Qt6 <https://www.qt.io/>`_. A plain ``make`` builds
+the GUI **only if a suitable Qt6 (>= 6.9.3) is found**; otherwise VPR is built
+headless. The 6.9.3 floor exists because earlier Qt6 releases have QRhi
```
vaughnbetz, line 17: QRhi -> QRhi (GPU accelerated)

vaughnbetz (commented):
A couple of minor nits in the docs but looks good.

vaughnbetz:
Thanks @w0lek and @kevinxing-a12 .... getting close I think.  I think the only remaining key thing to resolve is the vpr error with --disp on

w0lek:
Hi @kevinxing-a12 
I just have push changes to here, here the diff
https://github.com/w0lek/vtr-verilog-to-routing/pull/19

Note: you have to re-run the install_apt_packages.sh
because the new dependency relevant for QT runtime has been added (package libxcb-cursor0)

kevinxing-a12:
Thank you @w0lek! I followed the steps you specified, but I still got the exact same error from the last VPR log file. I cannot tell if it's an issue with the script or our computer setup. @vaughnbetz @AlexandreSinger

w0lek:
> Thank you @w0lek! I followed the steps you specified, but I still got the exact same error from the last VPR log file. I cannot tell if it's an issue with the script or our computer setup. @vaughnbetz @AlexandreSinger

Hi Kevin.
Thank you for testing, and sorry for several failures u are exprieincing.
Not sure what was the reason. I was doing dev and test on ubuntu 20.04. The CI doesn't run true UI mode. 
So i need time to setup ubuntu 24.04 and try to reproduce the issue there to analyze and provide proper fix.
Up to know i have just guessing that it has to be conflict between system Qt6 version and the custom. The error u see is related to Qt runtime plugin loading. Give me some time to analyze the problem.

kevinxing-a12:
No worries at all!

vaughnbetz:
I think we may have sudo issues that are confusing things.  Kevin doesn't have sudo access on the server he is using, so the install_apt_packages.sh may not have installed all the dependencies properly.

I just tried pulling all these changes in on a different server (which I also don't have sudo access on, but hopefully has the right packages already).  I got to this point in compilation, and then had an error:

/nfs/ug/fast1/ece297s/dev/vaughn/vtr-verilog-to-routing2/vpr/src/server/gateio.h:19:10: fatal error: QTimer: No such file or directory
   19 | #include <QTimer>
      |          ^~~~~~~~

It may be due to me not running install_apt_packages.sh properly too though.

On the positive side, make ensure-gui appears to have installed a local Qt for me.

vaughnbetz:
@AmirhosseinPoolad : can you try on your local machine (where you have sudo) and report what happens?
We've also asked the sysadmin for the server to install all the install_apt_packages.sh dependencies, so Kevin can try again after that.

vaughnbetz:
I confirmed the machine I'm running on has all the graphics packages listed in install_apt_packages.sh. So the compile errors doesn't seem to be due to missing packages.
This is on debian12

vaughnbetz:
OK, my compiler error was a silly mistake: forgot to do a git submodule update after pulling in Oleksandr's changes on top of master (I think).  I did that today and it fixed the compile error. 
Now I get a successful build when I do 
make ensure-gui

But vpr still fails to initialize graphics.  I run the command:

vpr/vpr vtr_flow/arch/timing/EArch.xml doc/src/quickstart/blink.odin.blif --disp on

and get this output, which matches what Kevin had:

```
<vpr log snipped>
	Block Utilization: 1.00 Logical Block: clb

FPGA size limited by block type(s): clb

qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: wayland, offscreen, eglfs, minimal, vkkhrdisplay, minimalegl, wayland-egl, linuxfb, vnc, xcb.

Abort
```

I do have xcb-cursor0 installed as a package (on Debian 12):

libxcb-cursor-dev/oldstable,oldstable 0.1.4-1 i386
libxcb-cursor0/oldstable,oldstable 0.1.4-1 amd64


I've attached the full vpr.log output and the output from apt list (apt.log) and the ensure_qt6_sdk.log.

[vpr.log](https://github.com/user-attachments/files/28978988/vpr.log)
[ensure_qt6_sdk.log](https://github.com/user-attachments/files/28978997/ensure_qt6_sdk.log)
[apt.log](https://github.com/user-attachments/files/28979007/apt.log)

@w0lek : any ideas? @AlexandreSinger thought it might have something to do with the dynamic library load path. As far as I can tell, my LD_LIBRARY_PATH isn't set.

Specifying --renderer of rhi, deferred or immediate makes no difference: same error each time.
Specifying --disp off (turning off graphics) works fine: vpr runs to completion normally.

w0lek:
> @w0lek : any ideas?

I installed vanilla ubuntu24.04 using Virtual Box, run two scripts:
1) sudo ./install_apt_packages.sh
2) ./dev/ensure_qt6_sdk.sh
3) Set SLANG_SYSTEMVERILOG to OFF (in some reason i got, not sure why CI didn't detect that. (i also did git submodule update, it didn't solve the problem) 
```
/home/yeti/workspace/fork_vtr-verilog-to-routing/libs/EXTERNAL/yosys-slang/src/slang_frontend.h:9:10: fatal error: kernel/rtlil.h: No such file or directory
    9 | #include "kernel/rtlil.h"
```
i will get back to that issue later, want focus completely on Qt related issue, so now i disable SLANG_SYSTEMVERILOG to bypass error.
4) `make ensure-gui` builds fine, and i was able to run vpr gui using

```
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
$SCRIPT_DIR/build/vpr/vpr \
$SCRIPT_DIR/vpr/test/test_post_verilog_arch.xml \
$SCRIPT_DIR/vpr/test/unconnected.eblif \
--route_chan_width 100 \
--gen_post_synthesis_netlist on \
--post_synth_netlist_unconn_inputs unconnected \
--post_synth_netlist_unconn_outputs unconnected \
--disp on

```
so no problem on vanila ubuntu.

About the issue:
Qt has build time dependency and runtime dependency (which are loaded dynamically during app runtime).
Since we didn't see any build issues, the build time dependencies are satisfied. Looks like the problem in dynamically loaded dependencies, i am investigating issue.

w0lek:
i still cannot replicate that issue on ubuntu 24.04.
installing system qt6 along side with custom, didn't bring me that issue (i suspect conflict in mixing loading plugins during runtime, when system qt6 plugins attempted to be loaded with custom qt6 core).

w0lek:
Hi @kevinxing-a12 
I still cannot reproduce issue on a fresh ubuntu 24.04 installation. Could you help me please with collecting more sensitive information about this issue? This might help me to dig issue further.
I have a script, which dump relevant info related to env vars,qt6, qt6 plugins, vpr deps etc. It also runs vpr with enabled qt plugins env var, to print more useful info to log from qt side.

Steps are:
1) make sure vtr is build with gui [make ensure-gui]
2) put this file to root of vtr repo [run_vpr_to_debug_qt6.sh](https://github.com/user-attachments/files/29001764/run_vpr_to_debug_qt6.sh)
and run it
3) script will generate run_vpr_to_debug_qt6.log, could u share it here?

kevinxing-a12:
Hi @w0lek, the log is here: 
[run_vpr_to_debug_qt6.log](https://github.com/user-attachments/files/29021893/run_vpr_to_debug_qt6.log)

ueqri:
@kevinxing-a12 `libxcb-cursor0` has been installed on Wintermute (Ubuntu 24.04) in case you want to run it on the server, you can find the dynamic library at `/usr/lib/x86_64-linux-gnu/libxcb-cursor.so`.

`libxcb-cursor-dev` has also been installed in case of potential build-time dependencies.

vaughnbetz:
Thanks @ueqri . @kevinxing-a12 : can you try building and running again to see if this resolves the issue?

w0lek:
FYI:

@vaughnbetz @ueqri

> Hi @w0lek, the log is here: [run_vpr_to_debug_qt6.log](https://github.com/user-attachments/files/29021893/run_vpr_to_debug_qt6.log)

Thank you! More inputs for my side, on VM mine ubuntu 24.04 uses wayland session, which doesn't required libxcb-cursor.so.0, that's why for my case the app was runnable without libxcb-cursor.so.0 and i didn't capture this behaviour.

But if ubuntu 24.04 uses X session instead of wayland, the libxcb-cursor.so.0 is required.
According your log i see that runtime xcb plug were missing following dependencies:
```
libxcb-cursor.so.0 => not found
	libxcb-icccm.so.4 => not found
	libxcb-image.so.0 => not found
	libxcb-keysyms.so.1 => not found
	libxcb-render-util.so.0 => not found
	libxcb-cursor.so.0 => not found
	libxcb-icccm.so.4 => not found
	libxcb-image.so.0 => not found
	libxcb-keysyms.so.1 => not found
	libxcb-render-util.so.0 => not found
```

So i updated install_apt_packages.sh to add all of them

https://github.com/w0lek/vtr-verilog-to-routing/pull/21/changes

May you please check?
Note: you should re-run latest install_apt_packages.sh to install new packages (i added with last commit). Rebuilding vtr is not required, since it's only runtime dependencies.

kevinxing-a12:
Hi @ueqri, could you run the newly updated install_apt_packages script? Thanks!
[install_apt_packages.sh](https://github.com/user-attachments/files/29052261/install_apt_packages.sh)

vaughnbetz:
@ueqri has installed the new packages.  @kevinxing-a12 : Hang recommends you log out and then log in again to ensure the environment variables are all updated/correct. Then please run the test again and update this PR as soon as you can.

kevinxing-a12:
@w0lek @vaughnbetz @AmirhosseinPoolad @AlexandreSinger Great news! The VPR GUI with Qt is running for the first time on wintermute! This is how it looks like: 

I did spot a few drawing issues though. Will discuss them in detail in my next comment.
@ueqri Huge thanks again for running the package script that makes everything work! :)

kevinxing-a12:
Problems that I noticed:

Critical:
After global placement (newly added to the GUI by Alex) is complete, almost everything disappears.

The only thing that can be displayed are critical paths. While other UI stuff remain interactable, nothing related shows up.
I doubt if it's caused by Alex's AP feature, because we tested AP graphics on the code without QT the other day, and almost everything was properly drawn. @AlexandreSinger Did you make additional commits on this after that day?

Moderate:
The "Route" button is already visible when the GUI starts, and I can toggle the switch and checkboxes even though routing hasn't commenced yet.

It disappears after global placement is done (see the very first picture), and appears again after initial routing completes. Before this PR, the button should stay invisible until after the initial routing stage.

The UI component that changes block internal level is frozen somehow. I can't interact with it to enable the drawing of internals.

This one likely has something to do with AP graphics, because Alex and I already came across with this problem in the earlier run I mentioned above.

Trivial:
The dropdown panels are left aligned now (as can be seen in previous pics), whereas they were center aligned before.

@w0lek @AlexandreSinger @vaughnbetz @AmirhosseinPoolad FYI.

kevinxing-a12:
I will start diving into the code to see why most drawing functions are skipped. It might be a trivial fix.

vaughnbetz:
Great news!

w0lek:
hi @kevinxing-a12 

> I doubt if it's caused by Alex's AP feature, because we tested AP graphics on the code without QT the other day, and almost everything was properly drawn. @AlexandreSinger Did you make additional commits on this after that day?

It seems i was able to reproduce issue and it's scoped to rhi renderer.
If u try --renderer immediate, it renders fine?

> I will start diving into the code to see why most drawing functions are skipped. It might be a trivial fix.

hint: if visual bug is not reproduced with other renderers ( for instance --renderer immediate), this probably scoped to rhi bugs, which is not easy to fix. I am checking this right now. If all renderers shows skipping visual behaviour, the problem probably in upper graphic level (draw states, refresh draw call etc) which is easy to fix.

kevinxing-a12:
Hi @w0lek I've tried --renderer immediate and deferred, and both worked.

vaughnbetz:
It looks like the issue isolated to 
- AP is on
- rhi is used

If we turn AP off, the rhi renderer seems to work fine except block internal viewing.
If we use the immediate or deferred renderers, AP (and the old flow) both work fine.

I think at this point we can merge, and @w0lek and @AlexandreSinger can work to get AP working with rhi.  @AlexandreSinger had to change the initialization of some graphics data to make AP graphics work, as some data structures like the grid weren't open yet.  Not sure why that only breaks rhi rendering, but I think this will not be too bad to fix since the AP graphics updates/fixes to make AP display are very recent, and hence may have missed something.

w0lek:
> Hi @w0lek I've tried --renderer immediate and deferred, and both worked.

I have fix for that, so i am going to push it shortly.
Fix is about refreshing initial_world (happens after we leave AP to regular device grid world), but fix will not be relied on leaving AP word event and be generic.
immediate and deffered renderes are not depend on that initial_world, that's why bug wasn't seen, and rhi use this var in building projection matrix (to cull out primitives outside the projection). Note: these parts u saw (text + triangles from critical path) are part of overlay which is part of deffered renderer (overlay is not depend on initial_world var too).

FYI @vaughnbetz

vaughnbetz:
Excellent!

w0lek:
> Not sure why that only breaks rhi rendering

@vaughnbetz 
rhi uses stale initial_world initialized from AP grid inside the device_grid world.
this stale is used in projection matrix, which make all rhi primitives invisible (out of projections). Only overlayed items are drawn, because they used deffered render mechanism.

w0lek:
FYI
@kevinxing-a12 @vaughnbetz 
- rhi bug with AP is fixed [dd9c2f2](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/commits/dd9c2f2c41a66ce28a543c138663e40918255d70)
- routing button is visible during AP is fixed here [01a1db4](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/commits/01a1db48368c62f435386507ddce6e5a02af3a5b)
-  block internal level is frozen for AP is fixed [12f8db0](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/commits/12f8db00eabe05a15c318d15cc555540e93a9ac0)
all commit pushed to qt_layer, so if you pull latest changes u should have all fixes.

vaughnbetz (approved):
LGTM

kevinxing-a12:
Here are what I tested:

- rhi with AP
- rhi without AP
- immediate with AP
- immediate without AP
- deferred with AP
- deferred without AP

From what I saw, these bugs have all gone:

> FYI @kevinxing-a12 @vaughnbetz
> 
> * rhi bug with AP is fixed [dd9c2f2](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/commits/dd9c2f2c41a66ce28a543c138663e40918255d70)
> * routing button is visible during AP is fixed here [01a1db4](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/commits/01a1db48368c62f435386507ddce6e5a02af3a5b)
> * block internal level is frozen for AP is fixed [12f8db0](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3542/commits/12f8db00eabe05a15c318d15cc555540e93a9ac0)
>   all commit pushed to qt_layer, so if you pull latest changes u should have all fixes.

@w0lek Congratulations! And thank you so much for all the efficient fixes!
@vaughnbetz @AlexandreSinger @AmirhosseinPoolad FYI.

AlexandreSinger (approved):
The CMAKE and CI changes LGTM!

Thank you so much for all your work on this @w0lek ! Assuming everything works locally (according to Kevin's message), I think this is good to merge from my end.

AmirhosseinPoolad:
GitHub Actions is not quite working right now and the dependency installation is taking forever. I restarted the Clang 16 and Functional Simulation tests once because of this but it's still getting stuck. Unfortunate.

AlexandreSinger:
@vaughnbetz CI is green and it looks like all approvals are in.

vaughnbetz:
Thanks for all the great work @w0lek and for the quick and thorough testing @kevinxing-a12, and all the CI debugging and install etc. support @AlexandreSinger and @AmirhosseinPoolad !  This is a great upgrade -- it's merged!

PR #3546, opened 2026-05-22
Enable packing cluster memoization by default

miloliebster:
#### Description

These are some changes I spoke with Vaughn about at FCCM:

1. Packing memoization has been merged for a while now and would make for a sensible default for the vast majority of users. We'd like to enable this option by default so that more users can benefit from it without combing through all of the command line options.

2. Updating the name of the new, more accurate, Stratix 10 capture to more clearly indicate to users why they would likely prefer to use this variant.

I don't believe new tests are required but some golden values may need to be updated for runtimes? The memoization _on_ test that I added with the original feature PR might also be redundant now if there was a similar existing test that used the previous default with it off. Need to look into these points more, but I'm starting the PR so that I can see what CI does with the current state of things, and so that we can open a larger discussion about these changes.


AmirhosseinPoolad:
Awesome! I believe you need to update the architecture's name in the test configs to resolve the current CI failure. There might also be some random QoR failures here and there but I don't expect anything actually important.

> The memoization on test that I added with the original feature PR might also be redundant now if there was a similar existing test that used the previous default with it off.

I think you can just change that test to be the memoization off test.

vaughnbetz (commented):
This would be good .... what do we need to do to be confident we can turn this on?  Should we get AP on by default first and then stack this on top?  Or vice versa?
@AlexandreSinger

AlexandreSinger:
> This would be good .... what do we need to do to be confident we can turn this on? Should we get AP on by default first and then stack this on top? Or vice versa? @AlexandreSinger

I would much prefer AP first; but I am obviously biased here. AP has a lot of moving parts and I am going through all the test cases to ensure that it is interracting well with all the different features. We know that memoization works with AP without issue (APPack is practically the same packing legalization approach), but I am a bit worried about blocking AP.

From my current timeline, I would expect AP to become default by the end of this week or early next week. The issues that are left are very small and I am just being very careful.

vaughnbetz:
Ok with me.

AlexandreSinger:
AP is now the default flow. It should no longer be blocking this PR.

vaughnbetz:
Yaay! @miloliebster : if you have the time it would be great to get CI green and land this.

miloliebster:
@vaughnbetz Apologies for the late reply—time is not an abundant resource for me at the moment. Let me get through midterms this week and I'll try to take a look when I am able. I believe the failing cases should be pretty trivial fixes.

vaughnbetz:
No worries @miloliebster  -- totally understood.

The CI failure might be a bad path in a new test:

scripts/run_vtr_task.py -l /home/runner/work/vtr-verilog-to-routing/vtr-verilog-to-routing/vtr_flow/tasks/regression_tests/vtr_reg_strong/task_list.txt -j 4 -script run_vtr_flow.py -show_failures -short_task_names 

Error: Failed to resolve VTR source file stratix10_arch.timing.depop50.complex_alm.xml
Error: Failed to resolve VTR source file stratix10_arch.timing.depop50.complex_alm.xml

Test 'vtr_reg_strong' had -1 qor test failures

miloliebster:
Finally escaped the chaos of this last term enough to sort out those regression tests, strong tests are all passing now!

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_titan_s10_non_simplified/config/config.txt
```diff
@@ -28,5 +27,4 @@ qor_parse_file=qor_vpr_titan.txt
 pass_requirements_file=pass_requirements_vpr_titan_s10.txt
 
 script_params_common=-starting_stage vpr -track_memory_usage --route_chan_width 400 --max_router_iterations 400 --router_lookahead map
-script_params_list_add =
```
AlexandreSinger, line 31: Why was AP turned off for this testcase? Was there an issue?

By removing this line, you are only running this test with the lower params (pack, place, route).

AlexandreSinger (changes requested):
Thanks @miloliebster 

I had a comment on one of the config files that was changed.

I was also curious if the memoization algorithm had any bounds on its memory usage? I have noticed in some local testcases that the memory usage increased when memoization was on. This is ok so long as the memory usage is bounded by something reasonable.

AlexandreSinger:
I have launched a weekly test just to be safe:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/31353860241/job/93349682197

I had to create a temporary branch on VTR master to run the action; we can delete it after the test concludes.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_titan_s10_non_simplified/config/config.txt
miloliebster: My mistake. Had removed that when troubleshooting and forgot to restore it. I've put it back now.

miloliebster:
@AlexandreSinger re: memory bound.

I'm curious if you could share more details on the nature of the memory increases you've seen. [Figure 7](https://arxiv.org/pdf/2604.24649) shows the memory behaviour we saw for our test benchmarks, where the memory use during the packing stage was always less than the peak memory usage from the baseline runs. There were some cases with the 7-series where even though the packing stage memory was still below peak, the peak memory increased by ~10% due to mysterious forces, perhaps fragmentation. Does that align at all with what you've seen, or is the memory during packing exceeding the baseline peak memory? How big is the increase you're seeing? What types of architectures have you seen this on, and what style of benchmark?

I'm assuming the tests you've seen this on used an AP mode, since that's now the default flow? Fundamentally, the PST was designed with seed-based packers in mind based on the principle that similar structures in the unplaced netlist will get packed in the same order and see a lot of overlap in the resulting tree. I don't truthfully have a good intuition for how this translates to AP, where those similar structures might get packed in a variety of different ways based on placement proximity, instead of from their abstract graph form. Initial tests still showed meaningful speedups with AP enabled, but I can't precisely explain the mechanism for repeated clusters forming with AP and if that will reliably occur. Do you see a correlation between higher memory, and worse speedup? Especially high memory would indicate to me that there are many failing cluster attempts, but very low hit-rate for seeing repeated clusters.

AlexandreSinger:
Hi @miloliebster , this data is not the cleanest; I am not directly trying to measure Deja Vu's memory usage; but what I have been doing is running a 7nm interposer architecture, similar to Aman's 22nm architecture (which we often call the Koios architecture). I ran it on a large number of the Koios benchmarks. Here is a scatter plot of the packer memory increase (note these are both AP and Traditional flow pack runs):

chart

I am noticing a peak memory increase when memoization is used with AP. Overall the speedup is reasonably the same with and without AP (but there is a slight difference). For the traditional flow, memoization speeds up the pack time by 32%; while for AP, memoization speeds up the pack time by 26%. So still an improvement, but likely the ordering changes from AP is hurting slightly.

To pull out a small example, for dla_like.large (the largest design in our suite), turning on memoization increased the peak memory for the traditional flow by 200 MB (11.28 GiB -> 11.54 GiB), but increased the peak memory for the AP flow by 1.3 GB (14.24 GiB -> 15.49 GiB). Many smaller designs are not changing in memory usage.

The reason the memory usage is more of an issue for AP is due to the AP flow generating the RR graph before packing. Before, the RR graph was generated after the packer (since the packer originally did not need). AP uses the RR graph to more effectively predict delays. I am actually not worried about this, since it is OK to use more memory to get performance. We can also unload the RR graph if we really need to save size.

My bigger concern is if the memory usage is unbounded. AP uses gain attenuation which will change the order that primitives are packed into clusters (even if topologically they form the same structure). Many other packing algorithms do something similar (such as LSC, and other algorithms which use positional data as input). Based on my understanding of your algorithm (which may be wrong so please correct me), the memory usage of your algorithm is proportional to the number of intra-lb route attempts (successful or unsuccessful, since both are cached). This can be a lot since the packer tries many candidates before terminating for a given cluster. I am honestly not sure if that is even bounded by anything since it tries candidates again everytime a new primitive is successfully added.

Is it possible to put a cap on the number of cached entries in the memoization structure? For example, can we stop memoizing when a certain limit is reached? I think a limit proportional to the number of atoms / molecules would be reasonable. This would just protect us from future strange algorithms / architectures increasing the memory in strange ways. What do you think?

miloliebster:
Hi @AlexandreSinger, thanks for sharing this context. Good to see that the memory increases you've noted in your tests are not horrifying, and the speedups are not wiped out by AP. Still, I completely agree with your concern about packing ordering with AP enabled--I don't believe the algorithm itself is bounded, and I'm sure a pathological case could exist where the memory usage gets out of control.

I'm on board with setting an artificial bound. If the memory passes a certain point then it indicates that the PST is likely not doing much good for runtime in that case anyway. I'll need to run some tests to get a count for what a reasonable coefficient w.r.t. the number of molecules would be. Is it possible for you to share the task configs for the tests you observed the memory increase on so that I could include those in my sweep?

AlexandreSinger:
> Is it possible for you to share the task configs for the tests you observed the memory increase on so that I could include those in my sweep?

The architecture I tested this on was a bit weird and I am currently working on fixing it. It was also out of branch. I used a script similar to this one: https://github.com/AlexandreSinger/fpga-analytical-placement/blob/main/testing/multi_die/hecate_2d_synth_verification/config/config.txt

However, the architecture I used was this one: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/fc50691626bcc2b46761884a87198eb96bd70f49/vtr_flow/arch/multi_die/hecate/2d/hecate_2d_k6FracN10LB_mem20K_complexDSP_customSB_7nm.xml#L1

I currently have a PR open that is going to completely change this architecture, so if you want the exact test you would run the task list above on this architecture.

This architecture does not have hardened adders or complex DSP hard blocks, so you would need to comment out this line:
image

The new architecture has both of these features.

Frankly, I do not know how much you would gain from using this architecture (it also takes days to synthesize some of these circuits, so if you can avoid it I recommend it). I think you will get reasonable bounds when you just run the normal COFFE architecture that is in this task list (commenting out the hecate architecture).

Let me know if you want the raw data for anything, I can send it over.

vaughnbetz:
I like the idea of bounding to some multiple of cluster count or atom count or molecule count.  I'm not too concerned of what the exact constant is though.  

I suggest printing out the memory footprint increase and the memoization instances per molecule (or per atom etc.) for a few of our favourite architectures.  From that we can probably just take 2x (or some such) the worst one (assuming the memory increase is moderate, which it seems to be) and set that as a bound.

miloliebster:
I've been running the 3 architectures from the original memoization tests (full-crossbar flagship, 7-series, s10) and the COFFE architecture from Alex's config using AP; getting ratios of `num_lcn/total_molecules` and `num_ecn/total_molecules`. Still will let all the results come in, but the worst case outlier result I've seen was on s10 where the num_ecn was 21x the number of molecules, although the other circuits cap out at around 10x. The `num_lcn` ratios are roughly the same with a bit of variance.

Since it would be quite messy to get the PST to stop adding new LCNs while still maintaining valid state with traversals, checkpoint rollbacks, and everything, my proposal is the following:

If the number of ECNs reaches 25x the number of molecules in the netlist, then we block the PST from being able to create new ECNs. This is non invasive to the algorithm and still allows for lookups into the existing tree up to this point. LCNs will still be allowed to be created, but since these are much much smaller than ECNs, the rate of memory growth will drop sharply.

Then If the number of LCNs continues to grow to 50x or perhaps 40x the number of molecules, we conclude that this architecture/circuit combo is exceptionally irregular and almost certainly not benefiting from the memoization anyway. In this case we free the PST and continue as though the `--memoize_cluster_packings` flag was never enabled to begin with.

vaughnbetz:
That sounds good to me.  If it's simpler, I'd also be OK with just turning off memoization if the number of LCNs or ECNs hits some multiple (e.g. 50x) the number of molecules.  Your call.

AlexandreSinger:
@miloliebster I also like this solution. I like the idea of stopping the recording once we go above a certain amount, and then stopping entirely if we go above a high threshold; but I agree with Vaughn that as long as we have some limit that is also sufficient.

miloliebster:
I've introduced the memory bounds in the way I'd proposed. Keeps the logic very uncomplicated.

AlexandreSinger (approved):
LGTM, thanks @miloliebster for resolving my concern on memory. I think as long as we have this limit, we should be in good shape.

I have no other issues with this PR. @vaughnbetz If you are OK with this PR, I think it is ready to merge.

vaughnbetz (approved):
LGTM, thanks!

PR #3547, opened 2026-05-23
[CI] Added No Assert Test

AlexandreSinger:
I have been finding some bad anti-patterns in the code where people are using asserts to throw an error from a function. This is always a bad idea and will cause undefined behavior in release builds where assert levels are turned down.

The default VTR assert level is 2, which is hiding a lot of this behavior, but it does concern me since I sometimes run at assert level 1 to collect results.

Added this CI test to at least keep an eye on this. The warnings are actually quite good at catching this. We should decide if we should make this test error on warning or not.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -278,6 +278,18 @@ jobs:
         make -j${{ steps.cpu-cores.outputs.count}}
         ./run_reg_test.py vtr_reg_basic -show_failures -j${{ steps.cpu-cores.outputs.count}}
 
+    - name: 'Test with VTR_ASSERT_LEVEL 0'
+      timeout-minutes: 120
+      if: success() || failure()
+      env:
+        CMAKE_PARAMS: "${{ env.COMMON_CMAKE_PARAMS }} -DVTR_ASSERT_LEVEL=0"
+        NUM_PROC: ${{ steps.cpu-cores.outputs.count }}
+      run: |
+        rm -f build/CMakeCache.txt
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        make -j${{ steps.cpu-cores.outputs.count}}
+        ./run_reg_test.py vtr_reg_basic -show_failures -j${{ steps.cpu-cores.outputs.count}}
+
     - name: 'Test with NO_GRAPHICS'
       timeout-minutes: 60
       if: success() || failure()
--- a/libs/libarchfpga/src/read_fpga_interchange_arch.cpp
+++ b/libs/libarchfpga/src/read_fpga_interchange_arch.cpp
@@ -380,7 +380,7 @@ struct ArchReader {
         for (auto bel : site.getBels())
             if (str(bel.getName()) == bel_name)
                 return bel;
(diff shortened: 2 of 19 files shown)


AlexandreSinger:
@AmirhosseinPoolad What is your opinion on this change? I do not want to make people's lives misreable with this change, but at the same time its really bad to have code safety checked behind asserts.

AlexandreSinger:
Oh, well that is unnexpected. Setting assert level to 0 actually failed basic test... That's legitimately a problem...

AlexandreSinger:
And this is why we need a test like this:
image

This is very very scary...

AmirhosseinPoolad (commented):
I like having a no assert CI run to catch side effect-having asserts like the one in route_common, but some of the checks are optional assertions by design to avoid performance hit in situations where the assertion really should never happen no matter what the user's input is. Essentially my guideline with asserts and fatal error is:

1- Dev fault: assert
2- User's fault: fatal error

And I'm not really sure if all of the changes here would be because of user error.

You should at least compare runtime to master to make sure this doesn't affect performance.

AlexandreSinger:
@AmirhosseinPoolad I agree with your points completely. This is why I am on the fence about this PR. We necessarily need the check, but not all asserts are a problem. I agree with your point that we should check that we did not get any runtime hit; but the issue is that we usually test at assert level 2, so by definition there would be no hit.

I can run at assert level 0 before and after this change (I may have to separate out the asserts that actually cause errors), to see if there are runtime hits; but for the future we would need to decide what to do. In general this would not always flag asserts that should be removed, this will just catch the ones which are likely to have side-effects. We may need to bring this up in a meeting.

AlexandreSinger:
Runtime results on Titan before and after this PR on ASSERT_LEVEL 1:

  | baseline.txt | no_assert_update.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.003781
num_LAB | 1 | 1
num_DSP | 1 | 1
num_M9K | 1 | 1
num_M144K | 1 | 1
max_vpr_mem | 1 | 1.000024
num_pre_packed_blocks | 1 | 1
device_grid_tiles | 1 | 1
pack_time | 1 | 1.00268
placed_wirelength_est | 1 | 1
place_time | 1 | 1.00267
placed_CPD_est | 1 | 1
routed_wirelength | 1 | 1
critical_path_delay | 1 | 1
crit_path_route_time | 1 | 1.011092

Fortunately the quality has stayed exactly the same, the assert routing issues I have flagged may have been min-channel width specific. The runtime did not increase by a large enough amount to worry. I think this change should be safe to merge.

@AmirhosseinPoolad What do you think? @vaughnbetz What do you think about adding a test like this?

vaughnbetz (approved):
LGTM. I agree anything a user can enter should result in an error, not an assert.  Developer mistakes can be asserts.

AlexandreSinger:
Ill merge this for now. Regarding the annoyance, we can play it by ear. If it becomes a pain we can turn off the error on warning part of this so we at least check for functionality. I think the nullptr things should be caught anyways for safety; but we can decide then. I think it should be rare enough that it wont be that bad.

PR #3549, opened 2026-05-25
Model TSV holes for Power Delivery

soheilshahrouz:
- Skip inter-layer SG links when either endpoint is an `EMPTY` tile, modeling TSV holes reserved for power delivery
- Fix per-side OPIN --> CHANZ edge creation when an adjacent switch block is `EMPTY`
- Cap scatter-gather fanin/fanout warnings at 10 per run
- Document `EMPTY` tile TSV-hole behavior in the architecture reference
- Update 3D simple test arch layouts, `basic_3d`/`basic_3d_ap` golden results, and `mm9a` IO constraints


soheilshahrouz:
TODO:
- use a custom tile instead of EMPTY 
- make sure that wires don't cross this new tile

doc/src/arch/reference.rst
```diff
@@ -221,8 +221,15 @@ Empty grid locations can be specified using the special block type ``EMPTY``.
 
 .. note:: All grid locations default to ``EMPTY`` unless otherwise specified.
 
-In multi-die devices, ``EMPTY`` tiles can also be used to model through-silicon via (TSV) holes reserved for power delivery.
-At these grid locations, VPR does not create inter-layer routing connections.
+TSV Hole Grid Locations
+~~~~~~~~~~~~~~~~~~~~~~~
+In multi-die devices, through-silicon via (TSV) holes reserved for power delivery are modeled using a tile type named ``tsv_hole``.
+The tile must be defined in the architecture file (in the ``<tiles>`` section and corresponding ``<pb_type>``) and placed in the device layout at grid locations corresponding to TSV holes.
+The ``<switchblock_locations>`` tag on the ``tsv_hole`` sub-tile should use ``pattern="external"`` so that switch blocks are created only outside the tile (see :ref:`fig_sb_locations`).
```
vaughnbetz: Couldn't it also be external_full_internal_straight?

vtr_flow/arch/multi_die/simple_arch/3d_k4_N4_90nm_opin_per_block.xml
```diff
@@ -61,23 +61,25 @@
   <layout>
     <auto_layout aspect_ratio="1.000000">
       <layer die="1">
-        <perimeter type="EMPTY" priority="101"/>
+        <region type="EMPTY" priority="101" startx="2" endx="W-3" starty="2" endy="H-3" />
```
vaughnbetz, line 83: Are we changing some /all of the EMPTY ones to tsv_hole?

doc/src/arch/reference.rst
soheilshahrouz: Maybe. Where is that connection created? Doesn't the driver wire go back to lower silicon layer to drive a buffer? If TSV holes destroys the silicon layer in that region, are straight connections realistic?

vaughnbetz: Straight wire to wire shorts are done in the metal stack; no transistors needed.

soheilshahrouz: I think we should set switchblock_locations to none. Even with internal straight, Fs is still larger than 1. In addition, For a 2x2 tile, straight connections only apply to one of SBs. The other three use full SBs

doc/src/arch/reference.rst
```diff
@@ -221,6 +221,16 @@ Empty grid locations can be specified using the special block type ``EMPTY``.
 
 .. note:: All grid locations default to ``EMPTY`` unless otherwise specified.
 
+TSV Hole Grid Locations
+~~~~~~~~~~~~~~~~~~~~~~~
+In multi-die devices, through-silicon via (TSV) holes reserved for power delivery are modeled using a tile type named ``tsv_hole``.
+The tile must be defined in the architecture file (in the ``<tiles>`` section and corresponding ``<pb_type>``) and placed in the device layout at grid locations corresponding to TSV holes.
+The ``<switchblock_locations>`` tag on the ``tsv_hole`` sub-tile should use ``pattern="external"`` so that switch blocks are created only outside the tile (see :ref:`fig_sb_locations`).
+
+.. note:: VPR recognizes TSV holes solely by the physical tile name ``tsv_hole``. This name must match exactly.
```
vaughnbetz: Is this still true?  I thought we were going to not do scatter-gather at spots with no switch blocks, so this special case would go away.  In that case we could document the use of this tile type as an example, but not as a special string.

doc/src/arch/reference.rst
```diff
@@ -2867,7 +2877,13 @@ The number of additional wires or muxes created by scatter-gather specifications
     Overview of how scatter-gather patterns work. First, connections from a switchblock location are selected according to the specification.
     These selected connection are then muxed and passed through the scatter-gather node, which is typically a wire segment. The scatter-gather node then fans out or scatters in another switchblock location.
 
-When instantiated, a scatter-gather pattern gathers connections from a switchblock and passes the connection through a multiplexer and the scatter-gather node which is typically a wire segment, then scatters or fans out somewhere else in the device. These patterns can be used to define 3D switchblocks. An example is shown below:
+When instantiated, a scatter-gather pattern gathers connections from a switchblock and passes the connection through a multiplexer and the scatter-gather node which is typically a wire segment, then scatters or fans out somewhere else in the device. These patterns can be used to define 3D switchblocks.
+
+.. note:: For inter-layer scatter-gather links (``z_offset`` != 0), VPR skips instantiation when either the gather or scatter endpoint is a ``tsv_hole`` tile.
```
vaughnbetz: Still true?  Or can we tie them to whether there is a switch block or not?

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -120,6 +120,8 @@ static void collect_sg_wire_candidates(const t_wireconn_inf& gather_pattern,
  */
 static t_wireconn_inf mirror_sg_pattern(const t_wireconn_inf& sg_pattern, const t_sg_link& sg_link);
 
+static constexpr int MAX_SG_FANIN_FANOUT_WARNINGS = 10;
```
vaughnbetz: Comment what this is and why we have it.

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
```diff
@@ -389,6 +393,20 @@ std::vector<t_bottleneck_link> alloc_and_load_scatter_gather_connections(const s
                     continue;
                 }
 
+                // TSV hole tiles model power-delivery openings; no CHANZ connection there.
+                if (sg_link.z_offset != 0
```
vaughnbetz, line 407: Same question ... can we tie this to the lack of a switch block instead?

vpr/src/route/rr_graph_generation/rr_graph_sg.cpp
```diff
@@ -58,8 +58,20 @@ static std::pair<RRNodeId, short> next_chanz_node_for_seg(const std::array<t_phy
                                                           std::vector<std::pair<RRNodeId, short>>& nodes0,
                                                           std::vector<std::pair<RRNodeId, short>>& nodes1) {
 
-    VTR_ASSERT_SAFE(!nodes0.empty());
-    VTR_ASSERT_SAFE(!nodes1.empty());
+    VTR_ASSERT_SAFE(!nodes0.empty() || !nodes1.empty());
+
+    // If an adjacent switch block is tsv_hole, no CHANZ is created there; use the other block.
```
vaughnbetz: I'm a bit lost as to how this is checking for a tsv_hole .... can you explain more?

vtr_flow/arch/multi_die/simple_arch/3d_k4_N4_90nm_opin_per_block.xml
```diff
@@ -51,27 +51,42 @@
...
-                     layer 1 ((layer + layer_offset) % num_of_avail_layers ((0 + 1) % 2 = 1))
-                  2) For clbs that are located at layer 1: their pins can be connected to layer 1 itself, and layer
-                     0 ((layer + layer_offset) % num_of_avail_layers ((1 + 1) % 2 = 0))
-               -->
           <loc side="bottom">clb.O</loc>
         </pinlocations>
       </sub_tile>
     </tile>
+    <tile name="tsv_hole" height="2" width="2" area="0">
+      <switchblock_locations pattern="external_full_internal_straight"/>
```
vaughnbetz: Looks like you're doing the external_full_internal_straight style after all?

vtr_flow/arch/multi_die/simple_arch/3d_k4_N4_90nm_opin_per_block.xml
```diff
@@ -51,27 +51,42 @@
...
+        <region type="EMPTY" priority="101" startx="2" endx="W-3" starty="2" endy="H-3" />
+        <region type="io" priority="100" startx="1" endx="W-2" starty="1" endy="H-2" />
+        <fill type="EMPTY" priority="99"/>
+      </layer>
+      <layer die="0">
+        <perimeter type="EMPTY" priority="101"/>
+        <!--fill with 'clb'-->
+        <fill type="clb" priority="100"/>
+
+        <!-- PW -->
```
vaughnbetz, line 92: Good to comment why we have tsv holes only on layer 0 (connection to pcb).

vaughnbetz (changes requested):
Not sure we need documentation / extra code paths for the special tsv_hole type.
I'm also a bit confused about the switch block locations to use (comments in one place say no switch blocks, arch seems to say straight through wires OK).

doc/src/arch/reference.rst
soheilshahrouz: Fixed this.

soheilshahrouz:
Even with straight switchblock connections, custom switchblock patterns could produce fan-ins greater than one. I changed RR graph generation to create only a single connection to each wire that starts at a TSV tile. That wire is driven by the wire with the same PTC number that ends in the previous channel segment

vpr/src/route/rr_graph_generation/build_scatter_gathers.cpp
soheilshahrouz: Fixed this.

vpr/src/route/rr_graph_generation/rr_graph_sg.cpp
soheilshahrouz: In per-side OPIN-CHANZ connectivity, we select CHANZ nodes from two SBs adjacent to the source OPIN. It resorts to the other SB if one doesn't have any CHANZ nodes.

vpr/src/route/rr_graph_generation/rr_graph_chan_chan_edges.cpp
```diff
@@ -129,11 +130,22 @@ static int get_track_to_ipins(RRGraphBuilder& rr_graph_builder,
 //  from_seg_coord: The horizontal or vertical location along the channel (i.e. y-coord for CHANY, x-coord for CHANX)
 //  from_chan_type: The from channel type
 //  to_chan_type: The to channel type
-static int should_create_switchblock(int layer_num,
-                                     int from_chan_coord,
-                                     int from_seg_coord,
-                                     e_rr_type from_chan_type,
-                                     e_rr_type to_chan_type);
+//  Returns {switch_override, is_short_connection}
```
vaughnbetz: Explain what these mean (e.g. I think NO_SWITCH for switch_override means don't create a switchblock at all). What does is_short_connection mean in terms of creating a switchblock).

vpr/src/route/rr_graph_generation/rr_graph_sg.cpp
vaughnbetz: Is this tsv hole comment too specific?  It seems like a tsv_hole is maybe a special case of the logic here, which your comment above says is trying to connect an OPIN to CHANZ nodes, and is checking two possible switch blocks. Suggest updating the comment.

vtr_flow/arch/multi_die/simple_arch/3d_k4_N4_90nm_opin_per_block.xml
```diff
@@ -51,27 +51,48 @@
...
-                  2) For clbs that are located at layer 1: their pins can be connected to layer 1 itself, and layer
-                     0 ((layer + layer_offset) % num_of_avail_layers ((1 + 1) % 2 = 0))
-               -->
           <loc side="bottom">clb.O</loc>
         </pinlocations>
       </sub_tile>
     </tile>
+    <tile name="tsv_hole" height="2" width="2" area="0">
+      <switchblock_locations pattern="custom">
+        <!-- electrical shorts -->
```
vaughnbetz: Add a comment saying in the area where the TSVs are, we can't have transistors but we can have wiring. Hence routing wires pass through, but there aren't regular switch blocks (which would include transistors).

vtr_flow/arch/multi_die/simple_arch/3d_k4_N4_90nm_opin_per_side.xml
```diff
@@ -51,13 +51,6 @@
         </fc>
         <pinlocations pattern="custom">
```
vaughnbetz, line 52: I guess this arch is not supposed to have tsv holes?

vtr_flow/arch/multi_die/stratixiv_3d/3d_bidir_SB_inter_die_stratixiv_arch.timing.xml
```diff
@@ -4736,6 +4736,23 @@
                 </pinlocations>
             </sub_tile>
         </tile>
+        <tile name="tsv_hole" height="2" width="2" area="0">
+            <switchblock_locations pattern="custom">
+                <!-- electrical shorts -->
```
vaughnbetz: I'd copy the same comment I suggested above to this arch here as well.

vtr_flow/arch/multi_die/stratixiv_3d/3d_SB_inter_die_stratixiv_arch.timing.xml
```diff
@@ -4736,6 +4736,23 @@
                 </pinlocations>
             </sub_tile>
         </tile>
+        <tile name="tsv_hole" height="2" width="2" area="0">
+            <switchblock_locations pattern="custom">
+                <!-- electrical shorts -->
+                <sb_loc type="straight_short" xoffset="0" yoffset="0"/>
```
vaughnbetz, line 4744: Same comment as above.

vaughnbetz (changes requested):
I like this technique -- more general. 
Added a few comments to address though.

vtr_flow/arch/multi_die/simple_arch/3d_k4_N4_90nm_opin_per_side.xml
soheilshahrouz: I didn't change this one, so we can test both cade paths with and without TSV modeling.

soheilshahrouz:
@vaughnbetz I applied your comments. If you think this is ready, can we merge it?

vaughnbetz (approved):

vaughnbetz:
LGTM. Go ahead and merge once you resolve the conflicts.

PR #3552, opened 2026-05-26
Fix linter issues in LibRRGraph

AmirhosseinPoolad:
Ran clang-tidy on librrgraph, these were the issues we had. Mostly just some unused variables.

PR #3551, opened 2026-05-26
Fix LibArchFPGA linter issues

AmirhosseinPoolad:
Ran clang-tidy on libarchfpga, these were the issues we had. Nothing serious thankfully.

PR #3553, opened 2026-05-26
[WIP] Add clang-tidy checks to CI

AmirhosseinPoolad:
Adds automatic linter runs for vpr, libarchfpga and librrgraph to CI. Should wait until #3551 and #3552 are merger before this.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -17,18 +17,6 @@ Checks: >
   -bugprone-suspicious-stringview-data-usage,
   -bugprone-return-const-ref-from-parameter,
   -bugprone-unchecked-string-to-number-conversion,
-  # performance-*,
-  # readability-*,
-  #-readability-function-cognitive-complexity,
-  #-readability-identifier-length,
-  #-readability-simplify-boolean-expr,
-  #-readability-redundant-access-specifiers,
-  #-readability-avoid-const-params-in-decls,
-  #-readability-else-after-return,
-  #-readability-uppercase-literal-suffix,
-  #-readability-use-anyofallof,
-  #-readability-redundant-inline-specifier,
-  #-readability-isolate-declaration,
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
A lot of the warnings and errors I see in the lint test I think are resolved in this PR that I was working on: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3527

Let me know if you have any comments.

AmirhosseinPoolad:
It takes \*way\* too long to run this in the CI. Not sure if I like it.

AlexandreSinger:
> It takes *way* too long to run this in the CI. Not sure if I like it.

My guess is that clang-tidy is actually running clang under the hood to do this (actually I am almost certain it is). I wonder if we can cache that build like we do the other builds.

hzeller:
For caching, I have written a clang-tidy runner that runs clang-tidy in parallel and also builds a directory where it caches results; so next time around it only has to process changed files or files that depend on changed files.
Using the github cache, you can use that to restore the cache, then next time it will be fast.

https://github.com/hzeller/dev-tools/blob/main/run-clang-tidy-cached.cc

Might be interesting to play with.

AmirhosseinPoolad:
> For caching, I have written a clang-tidy runner that runs clang-tidy in parallel and also builds a directory where it caches results; so next time around it only has to process changed files or files that depend on changed files. Using the github cache, you can use that to restore the cache, then next time it will be fast.
> 
> https://github.com/hzeller/dev-tools/blob/main/run-clang-tidy-cached.cc
> 
> Might be interesting to play with.

Thanks a lot! This should be very helpful.

vaughnbetz:
No rush, but are you still working to get this in @AmirhosseinPoolad ?

PR #3554, opened 2026-05-26
Fix --route to not also run analysis stage

amin1377:
When VPR is invoked with only --route (and no --analysis), the analysis stage runs anyway whenever routing succeeds. In this PR, this is fixed so that --analysis is not run by default after routing.


amin1377:
@AlexandreSinger The CI tests are passing now. Could you please take a look at this PR when you get a chance?

Thanks!

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_ap/annealer_detailed_placer/config/config.txt
```diff
@@ -50,5 +50,5 @@ qor_parse_file=qor_ap_fixed_chan_width.txt
 pass_requirements_file=pass_requirements_ap_fixed_chan_width.txt
 
 # Pass the script params while writing the vpr constraints.
-script_params=-starting_stage vpr -track_memory_usage --analytical_place --ap_detailed_placer annealer --route
+script_params=-starting_stage vpr -track_memory_usage --analytical_place --ap_detailed_placer annealer --route --analysis --analysis
```
AlexandreSinger: Duplicate analysis here.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_place_delay_calc_method/config/golden_results.txt
AlexandreSinger, line 1: Why are these and the other golden results changing? Shouldn't they get the exact same solution?

AlexandreSinger (commented):
A couple of minor comments. I am not directly against this PR, but I think we need another opinion (from @vaughnbetz perhaps). This is directly changing the API for any non-standard flows.

Right now, AP is not the default, so the only way to activate the stage is by saying `--analytical_place`. Due to VPR's current syntax, that means ONLY run the AP stage (not route), so we had to add `--route` as well. Now this is making it so we have to add `--analysis` as well. Its annoying, but its frankly more correct since if someone specifies `--route` they ONLY want route as per our own syntax.

In the future, I am making AP the default, so if someone want to use the old flow they would need to write `--pack --place --route --analysis`; so its going to be a bit annoying (and perhaps non-obvious since why wouldn't I want analysis if I am running the whole flow).

This same issue will happen for any other non-standard flow.

I think this interface may need to a face-lift to make me happy (which is independent of this PR; but its why I want Vaughn's opinion on this).

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_place_delay_calc_method/config/golden_results.txt
AlexandreSinger: Oh is it because the golden test get screwed because of the folder naming convention coming from the different args?

AlexandreSinger:
As an aside. I wonder if we should add a "--flow <flow_name>" flag which selects pre-set flows that we care about. It could be something like:
```sh
# This is the current default flow
vpr --flow pack_place_and_route ...

# This is the AP flow
vpr --flow ap_and_route
```

And then by default it would be whichever of the two flows. Then, when people specify single-stages, it disables this argument and does what we see above.

I just want to make this flow interface cleaner so changes like this do not break non-standard flows.

vaughnbetz:
Let's brainstorm on this Friday morning in person.  @amin1377 @AlexandreSinger : both around? Alex and I are meeting at 10:30, so that could work.

amin1377:
@vaughnbetz: I think setting a time after noon would work better for me. However, if no other time works for you, please let me know, and I can move my other meeting.

vaughnbetz:
Sure. Could meet right after the vtr meeting, or at 3 pm.VaughnOn May 28, 2026, at 8:50 AM, Amin Mohaghegh ***@***.***> wrote:﻿amin1377 left a comment (verilog-to-routing/vtr-verilog-to-routing#3554)
@vaughnbetz: I think setting a time after noon would work better for me. However, if no other time works for you, please let me know, and I can move my other meeting.

amin1377:
> Sure. Could meet right after the vtr meeting, or at 3 pm.VaughnOn May 28, 2026, at 8:50 AM, Amin Mohaghegh ***@***.***> wrote:﻿amin1377 left a comment ([verilog-to-routing/vtr-verilog-to-routing#3554](https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3554)) @vaughnbetz: I think setting a time after noon would work better for me. However, if no other time works for you, please let me know, and I can move my other meeting. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

After the VTR meeting works perfectly for me. Thanks!

amin1377:
@AlexandreSinger Based on the discussion we had yesterday, I think this PR is now ready to be merged.

AlexandreSinger (approved):
LGTM

PR #3555, opened 2026-05-26
InterposerCostHandler Class

soheilshahrouz:
Extract interposer placement cost logic from `NetCostHandler` into a dedicated `InterposerCostHandler` class.


soheilshahrouz:
titan_quick_qor 

| Metric | master | this PR |
| --- | ---: | ---: |
| vtr_flow_elapsed_time | 1.000 | 0.994 |
| num_LAB | 1.000 | 1.000 |
| num_DSP | 1.000 | 1.000 |
| num_M9K | 1.000 | 1.000 |
| num_M144K | 1.000 | 1.000 |
| max_vpr_mem | 1.000 | 1.000 |
| num_pre_packed_blocks | 1.000 | 1.000 |
| num_post_packed_blocks | 1.000 | 1.000 |
| device_grid_tiles | 1.000 | 1.000 |
| pack_time | 1.000 | 1.009 |
| placed_wirelength_est | 1.000 | 1.000 |
| place_time | 1.000 | 0.991 |
| placed_CPD_est | 1.000 | 1.000 |
| routed_wirelength | 1.000 | 1.000 |
| critical_path_delay | 1.000 | 1.000 |
| geomean_nonvirtual_intradomain_critical_path_delay | 1.000 | 1.000 |
| crit_path_route_time | 1.000 | 0.982 |

soheilshahrouz:
koios results for hecate_25d_L5_int_5um_bump_fanin_24

| Metric | master.txt | feat.txt |
|--------|------------|----------|
| vtr_flow_elapsed_time | 1 | 1.015073915 |
| max_vpr_mem | 1 | 0.9998549314 |
| num_pre_packed_blocks | 1 | 1 |
| num_post_packed_blocks | 1 | 1 |
| device_grid_tiles | 1 | 1 |
| pack_time | 1 | 1.029742101 |
| placed_wirelength_est | 1 | 1 |
| place_time | 1 | 1.000361784 |
| placed_CPD_est | 1 | 1 |
| routed_wirelength | 1 | 1 |
| critical_path_delay | 1 | 1 |
| geomean_nonvirtual_intradomain_critical_path_delay | 1 | 1 |
| crit_path_route_time | 1 | 0.9977270031 |

vpr/src/place/net_cost_handler.cpp
```diff
@@ -1851,152 +1677,8 @@ double NetCostHandler::get_total_wirelength_estimate() const {
...
-                continue;
-            }
-
-            net_interposer_cong_cost_[net_id] = get_net_cube_interposer_cong_cost_(net_id, /*use_ts=*/false);
-            total_interposer_cong_cost += net_interposer_cong_cost_[net_id];
-        }
-    }
-    return total_interposer_cong_cost;
+const std::vector<ClusterNetId>& NetCostHandler::affected_nets() const {
+    return ts_nets_to_update_;
```
AmirhosseinPoolad, line 1681: Are there any limitations on when you can call affected_nets? If, for example, you can only call it after find_affected_nets_and_update_costs is called then this should be documented or ideally have an assertion that checks it.

AmirhosseinPoolad (commented):
Thanks Soheil. Had one comment.

I think after this change we could very easily set non-zero default values for interposer cost terms without changing the default flow's QoR. Let's do that in a future PR though. I'll raise an issue to keep track.

soheilshahrouz:
@vaughnbetz Do you think this is ready to merge?

vpr/src/place/interposer_cost_handler.cpp
```diff
@@ -0,0 +1,367 @@
...
+    // Precompute 1/width and 1/height so interposer crossing cost can normalize BB spans with
+    // multiplies instead of a divide per net (see get_net_interposer_cost_).
+    inv_device_grid_width_ = 1.0 / static_cast<double>(grid.width());
+    inv_device_grid_height_ = 1.0 / static_cast<double>(grid.height());
+
+    const size_t num_layers = grid.get_num_layers();
+    const size_t num_nets = g_vpr_ctx.clustering().clb_nlist.nets().size();
+
+    if (interposer_cost_enabled_ || interposer_cong_threshold_ > 0) {
+        VTR_ASSERT(num_layers == 1);
```
vaughnbetz: Probably good to comment; no allowing combinations of 3D and interposer devices for now.

vpr/src/place/interposer_cost_handler.cpp
```diff
@@ -0,0 +1,367 @@
...
+        VTR_ASSERT(std::ranges::is_sorted(grid.get_vertical_interposer_cuts()));
+    }
+
+    if (interposer_cong_threshold_ > 0) {
+        net_interposer_cong_cost_.resize(num_nets, -1.);
+        proposed_net_interposer_cong_cost_.resize(num_nets, -1.);
+
+        // Interposer cut locations can vary by layer, so allocate using the maximum cut count across layers.
+        size_t max_h_cuts = 0, max_v_cuts = 0;
+        for (size_t layer = 0; layer < num_layers; layer++) {
```
vaughnbetz, line 52: We're currently allowing only one layer, so probably want to comment why you have the assert above since at least some of the code is looking at multi-layer devices.

vpr/src/place/interposer_cost_handler.cpp
```diff
@@ -0,0 +1,367 @@
...
+    return interposer_cost_enabled_ || interposer_cong_modeling_started_;
+}
+
+std::pair<double, double> InterposerCostHandler::recompute_costs() {
+    const ClusteredNetlist& clb_nlist = g_vpr_ctx.clustering().clb_nlist;
+
+    std::pair<double, double> cost_terms;
+    for (ClusterNetId net_id : clb_nlist.nets()) {
+        if (!clb_nlist.net_is_ignored(net_id)) {
+            if (interposer_cost_enabled_) {
```
vaughnbetz: I think this if is because interposer costs may be turned on part way through placement; is that right?  Should comment that. Otherwise you could remove the if and have the caller avoid the call.

vpr/src/place/interposer_cost_handler.cpp
```diff
@@ -0,0 +1,367 @@
...
+    // If we support devices with many dice, consider a refactor (e.g. interval queries)
+    // so cost updates do not scale linearly with cut count.
+
+    const t_bb& bb = get_net_bb_(net_id, use_ts);
+    const auto [num_horizontal_crossings, num_vertical_crossings] = count_bb_interposer_cut_crossings_(bb);
+
+    // Weight crossings by the normalized BB span orthogonal to the cut direction:
+    // - a horizontal cut spans X, so we scale by BB height / grid height
+    // - a vertical cut spans Y, so we scale by BB width / grid width
+    const double bb_width_factor = double(bb.xmax - bb.xmin + 1) * inv_device_grid_width_;
```
vaughnbetz, line 182: Should comment the intuition behind this ... shrink the bb so it may not cross the interposer in a later move.  If you've tested this works better than a fixed interposer cost mention that; if not, put a TODO in to try a constant interposer crossing cost.

vaughnbetz (approved):
I suggest a few commenting changes and I had a few questions. After that, good to merge.

vpr/src/place/interposer_cost_handler.cpp
soheilshahrouz: Command line options allow the use to enable interposer cost terms independently. Congestion modeling starts when range limits falls below a threshold.

vpr/src/place/interposer_cost_handler.cpp
soheilshahrouz: I added this limitation when I was developing the class. At this point, I think this class supports 3d devices. I never tested it with with a 3d device with interposer cuts though. 

I can add TODO to test it with a multi-layer interposer based FPGA, and then remove the assertion on the number of layers.

AlexandreSinger (approved):
LGTM

PR #3556, opened 2026-05-26
[CI] Added Functional Simulation Regression Test Suite

AlexandreSinger:
Updated the VTR task syntax to allow for functional simulation using verilator. This is used with the same config syntax as the rest of the VTR infrastructure.

Added a new vtr_reg_func_sim regression suite which is run by the CI. This currently verifies a basic adder circuit on different architectures. I plan to extend this to more testcases soon.


AlexandreSinger:
@ayan20985 FYI, this is the CI infrastructure I was talking about in the team meeting. We will want to add more tests to this that try things like RAMs, flip-flops that hold state, DSPs, mutli-clock, etc.

AlexandreSinger:
@vaughnbetz The VTR Task infrastructure and CI test for functionally testing the post-implementation netlist has been added by this PR and it is ready to review when you have a moment. The updates to the task and the new testing script I added can be found in the docs here:
- https://vtr--3556.org.readthedocs.build/en/3556/vtr/run_func_sim_flow/
- https://vtr--3556.org.readthedocs.build/en/3556/vtr/tasks/#functional-simulation-tasks

This PR looks long, but a lot of it was just implementing the testing infrastructure boilerplate, so adding benchmark circuits and config files. Let me know what you think. Once this is merged, we can start to investigate testing gaps and try to fill them. This interface should make it very easy to add new tests.

vtr_flow/benchmarks/func_sim/adder_4bit/adder_4bit_carry_adder.blif
```diff
@@ -0,0 +1,84 @@
+# Benchmark "top" written by ABC on Tue May 26 15:03:08 2026 
+
+
+.model top
```
vaughnbetz, line 20: What do we use this .blif for?  I assumed we were comparing the adder_4bit.v to the tb.

vtr_flow/benchmarks/func_sim/adder_4bit/adder_4bit_luts.blif
```diff
@@ -0,0 +1,80 @@
+# Benchmark "top" written by ABC on Tue May 26 15:03:08 2026 
+
```
vaughnbetz, line 18: Same question. Why the .blif?  Is it so you can verify vpr with no synthesis involved? If so I suggest commenting that at the top of the blif or some such. We'd also want some tests that go all the way through synthesis + p&r, so there should be some .v -> post_impl.v tests too.

vaughnbetz (approved):
Looks great. My only thought was that explaining somewhere that why we have .blif format tests (I assume they skip yosys and maybe abc in the flow scripts and hence are more focused test?) would be good.  Maybe that explaination should be in the overall documentation and/or the relevant config file?

vtr_flow/benchmarks/func_sim/adder_4bit/adder_4bit_carry_adder.blif
AlexandreSinger: We are doing both. These are smoke tests to ensure that verilator is working with standard LUTs and Adders irrespective of the front-end being used.

All four files that you see in this folder (both verilog and blif) are independently being verified against the one adder testbench to ensure that they are getting exactly the same result after P&R. I will update the comments to reflect this.

AlexandreSinger: The Task infrastructure I set up is also capable of functionally verifying blif files through the verilator flow as well, and I needed tests to lock-in that feature. I predict that we will want to verify one of the Titan or Titanium benchmarks in the future, and I think we would want to verify they blif versions to ensure everything is good.

vtr_flow/benchmarks/func_sim/adder_4bit/adder_4bit_luts.blif
AlexandreSinger: Yep, and we also have tests for the full VTR flow as well. There are two different task list which are running.

Issue #3543, opened 2026-05-21
Legacy Linked List in t_model Data Structure


AmirhosseinPoolad:
In the t_model struct, a type-unsafe linked list is used:
```
struct t_model {
    char* name = nullptr;                   ///< name of this logic model
    t_model_ports* inputs = nullptr;        ///< linked list of input/clock ports
    t_model_ports* outputs = nullptr;       ///< linked list of output ports
    void* instances = nullptr;              ///< TODO: Remove this. This is only used in the Parmys plugin and should be moved into there.
    int used = 0;                           ///< TODO: Remove this. This is only used in the Parmys plugin and should be moved into there.
    vtr::t_linked_vptr* pb_types = nullptr; ///< Physical block types that implement this model
    bool never_prune = false;               ///< Don't remove from the netlist even if a block of this type has no output ports used and, therefore, unconnected to the rest of the netlist
};
```

See pb_types. This is unsafe legacy code that is hard to use and maintain. pb_types can be changed into a std::vector<t_pb_type*> easily and is not used much so the changes shouldn't be too hard.

From what I can see:

1- The functions doing allocation and deallocation would be greatly simplified (no need to deallocate anything, RAII would do it automatically)
2- Other users of t_model::pb_types mostly just loop on it. The loops can be simplified into range for loops.

PR #3559, opened 2026-05-27
Turned the type-unsafe "pb_type" linked list into a vector

kevinxing-a12:
#### Description

I turned the type-unsafe "pb_type" linked list into a vector and I casted the old linked list data type from "void*" to "t_pb_type*". I also made changes to files where syntax difference occurred due to this change in data structure. 
#### Related Issue

Legacy Linked List in t_model Data Structure #3543

#### Motivation and Context

This change fixes a legacy linked list that is error-prone, hard to maintain, and rarely used.

I run the strong VPR test and everything passed

In arch_util.cpp, the old "pb_type" linked list was extended through head insertion, whereas now I use emplace_back() to extend the vector. A possible issue would be the discrepancy in traversal difference, but I didn't get any errors from the strong VTR test.
#### Types of changes

#### Checklist:


libs/libarchfpga/src/arch_util.cpp
```diff
@@ -782,10 +782,7 @@ void SyncModelsPbTypes_rec(t_arch* arch,
         t_model& model_match_prim = arch->models.get_model(model_match_prim_id);
 
         pb_type->model_id = model_match_prim_id;
-        vtr::t_linked_vptr* old = model_match_prim.pb_types;
-        model_match_prim.pb_types = new vtr::t_linked_vptr;
-        model_match_prim.pb_types->next = old;
-        model_match_prim.pb_types->data_vptr = pb_type;
+        model_match_prim.pb_types.emplace_back(pb_type);
```
AmirhosseinPoolad, line 785: The old code adds the pb_type to the start of the list. I'm not sure if it's actually important or not but to be safe I think we should do the same and add a TODO to investigate this later.

libs/libarchfpga/src/logic_types.cpp
```diff
@@ -144,12 +144,7 @@ void LogicalModels::free_model_data(t_model& model) {
     free_arch_model_ports(model.inputs);
     free_arch_model_ports(model.outputs);
 
-    vtr::t_linked_vptr* vptr = model.pb_types;
-    while (vptr) {
-        vtr::t_linked_vptr* vptr_prev = vptr;
-        vptr = vptr->next;
-        delete vptr_prev;
-    }
+    std::vector<t_pb_type*>().swap(model.pb_types);
```
AmirhosseinPoolad, line 147: The old code is deleting the vtr::t_linked_vptr structure. Swapping with an empty vector works but I believe you can also just... not do anything. When the model goes out of scope the vector would automatically call the destructor.

AmirhosseinPoolad (commented):
Thanks a lot Kevin! This is some very tricky legacy code. Had some minor comments.

libs/libarchfpga/src/arch_util.cpp
kevinxing-a12: Yes, I realized the same thing! When the vector is traversed, the order is opposite to the old linked list. I was discussing this with Alex and we thought if running the strong test does not throw any error then we can temporarily keep it this way to save some runtime.

AmirhosseinPoolad:
You can fix the formatting issue by running `make format`

kevinxing-a12:
Wait I'm not sure why there's a check failing. Let me take a look real quick.

libs/libarchfpga/src/logic_types.cpp
kevinxing-a12: Yes! I did this because I wasn't sure if the intention was to free up space as early as possible.

kevinxing-a12:
Ok I have run "make format" and re-run the strong test.

libs/libarchfpga/src/arch_util.cpp
vaughnbetz: I'd do the more natural / efficient thing (which you've done) and see if it causes any problems (which it shouldn't).  I don't think the code really preferred one order over the other.

To be safe, after CI passes you should run one QoR test (either the VTR designs or Titan) and confirm no degradation on average in runtime or quality.  The developer guide in the VTR documentation says how to do that, and Amir or Alex can help if you aren't sure how to run it.

libs/libarchfpga/src/logic_types.cpp
```diff
@@ -144,12 +144,7 @@ void LogicalModels::free_model_data(t_model& model) {
     free_arch_model_ports(model.inputs);
```
vaughnbetz, line 144: Looks like these model ports are also custom linked lists; maybe a good follow-up issue to keep cleaning this up by turning those into vectors too.

libs/libarchfpga/src/logic_types.h
```diff
@@ -67,13 +70,13 @@ struct t_model_ports {
...
  */
 struct t_model {
-    char* name = nullptr;                   ///< name of this logic model
-    t_model_ports* inputs = nullptr;        ///< linked list of input/clock ports
-    t_model_ports* outputs = nullptr;       ///< linked list of output ports
-    void* instances = nullptr;              ///< TODO: Remove this. This is only used in the Parmys plugin and should be moved into there.
-    int used = 0;                           ///< TODO: Remove this. This is only used in the Parmys plugin and should be moved into there.
-    vtr::t_linked_vptr* pb_types = nullptr; ///< Physical block types that implement this model
-    bool never_prune = false;               ///< Don't remove from the netlist even if a block of this type has no output ports used and, therefore, unconnected to the rest of the netlist
+    char* name = nullptr;             ///< name of this logic model
```
vaughnbetz, line 73: More cleanups (for later):
- name --> std::string
- inputs --> make a vector instead of a linked_list
- outputs --> make a vector instead of a linked list

vaughnbetz (approved):
Looks good. 
I have a few suggestions for future work (getting rid of a couple more linked lists).

Another possible future work (but not sure if it's worthwhile): only Odin-II and parmys (I think) use t_linked_vptr now. So we could possibly hide it a bit by moving it out of lib_vtr_util and into parmys or Odin II (so long as they can both call it there).

vaughnbetz:
Should do one QoR check before merging, since this does make a small change in functionality (list order).

libs/libarchfpga/src/logic_types.h
kevinxing-a12: Got it!

kevinxing-a12:
Comparsion is made between commit #0b2aff2a74f48408fed3c0cece65f23ebcc2c4e4 and #89c4b185e7797697e5a9b1d7a7d7fff73edc7dc8 (the first commit in this PR).
[comparison.xlsx](https://github.com/user-attachments/files/28373731/comparison.xlsx)
It seems that "parmys_synth_time" and "crit_path_route_time" become a bit longer, yet I'm not sure if this is still within the acceptance threshold.

AmirhosseinPoolad:
QoR is identical, runtime is probably just machine load/noise since you didn't change routing or anything routing related but route time got longer. LGTM.

AmirhosseinPoolad (approved):

vaughnbetz:
LGTM, thanks.

PR #3560, opened 2026-05-28
Enable windows regression tests

oscarcheng105:
#### Related Issue

Change is based on the issue : https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3432
As well as this issue from OpenFPGA: https://github.com/lnis-uofu/OpenFPGA/issues/2382

#### Motivation and Context

Ensure reg tests run on windows(MSVC) builds.

#### Types of changes

#### Checklist:


AmirhosseinPoolad:
BTW you can use the `vtr_flow/tasks/regression_tests/vtr_reg_nightly_test1/vpr_reg_mcnc` test if you want to avoid synthesis. It's a very small set of circuits but I think it would be okay for regression testing. You can also make a modified version of the `vtr_flow/tasks/regression_tests/vtr_reg_nightly_test2/titan_other` test (Only do 1 or 2 of the smaller circuits to not bloat the CI runtime too much) if you want better coverage.

oscarcheng105:
Screenshot 2026-05-30 213339
mcnc benchmarks on both wins and wsl

vtr_flow/scripts/python_libs/vtr/paths.py
```diff
@@ -45,9 +46,13 @@
 ace_exe_path = ace_path / "ace"
 ace_extract_clk_from_blif_script_path = ace_path / "scripts" / "extract_clk_from_blif.py"
 
+# VPR paths
 # VPR paths
 vpr_path = root_path / "vpr"
-vpr_exe_path = vpr_path / "vpr"
+if os.name == "nt":
+    vpr_exe_path = root_path / "build" / "vpr" / "vpr.exe"
+else:
```
tangxifan: LINE54-55 is redundant.

tangxifan (commented):
@oscarcheng105 Can you enable the MCNC test on the windows build CI job? Thanks!

vaughnbetz (approved):
LGTM

AmirhosseinPoolad:
Do you plan to add MCNC regression testing in this PR or a future PR?

oscarcheng105:
> Do you plan to add MCNC regression testing in this PR or a future PR?

MCNC reg test for CI will be added in this pr

AlexandreSinger (approved):
LGTM, I will merge this.

A couple of non-blocking things:

1. I noticed that the run_vtr_task script was set to 1 thread; that is fine for this test since it takes under 2.5 minutes, but is the winodws build capable of running multiple designs at once? If so, we may want to consider upping the number of cores in this test since people tend to copy-paste things when they create new testcases. That can come in another PR, I would rather get these regression tests in to ensure Windows is supported through VPR.
2. Now that the windows flow is being tested end to end in CI, it may be worth adding documentation on how to run VPR on windows. I noticed that you had a step for setting the DLL paths, it would be nice to have this cleanly documented for end users. This is not blocking, since I think the documentation may change once we support Parmys / Yosys as well for Windows.
3. This test is not checking if the golden results match. Right now this test is only checking that VPR ran without explicit error, it is not checking if the solution is reasonable. I recognize that Windows may get slightly different golden results from the Linux builds, which may cause the tests to fail; but eventually we should add golden results checking. Without this, I am not sure if the CI will actually fail; it may fail and still pass CI.

PR #3561, opened 2026-05-28
[SDC] Added -edges and -edge_shift support

AlexandreSinger:
-multiply_by and -divide_by are not sufficient to describe all generated clocks. -edges and -edge_shift enables more generated clocks to be supported.

These two options define the generated clock based on the rising and falling edges of the source clock.

Added strong tests to lock in these options.

Updated the docs with these new options


vaughnbetz (approved):
Looks good, thanks!  Great documentation.

PR #3562, opened 2026-05-28
Packing legality check for direct connections

amin1377:
During packing, the intra-cluster router may choose any top-level output pin as a net’s exit point. However, if it selects a pin with Fc_out = 0, that pin has no access to general routing resources. In that case, the net can only reach its sink if there is a matching <direct> connection.

Previously, this case was not checked, which meant a net could be assigned to an Fc_out = 0 output pin even when its sink was not reachable through a direct connection. This would later cause routing to fail.

This PR adds a sanity check after packing, as part of check_netlist, to detect this situation. The check flags any net that uses an Fc_out = 0 top-level output pin when its sink is not reachable through a matching <direct> connection.

This PR only adds the detection/checking mechanism. The follow-up PR will address how to automatically prevent or fix this issue during routing/packing.


vpr/src/base/check_netlist.h
```diff
@@ -1,3 +1,5 @@
 #pragma once
 
-void check_netlist(int verbosity);
+struct t_arch;
+
+void check_netlist(int verbosity, const t_arch* arch);
```
vaughnbetz: Suggest a const ref instead of a pointer.  Looks nicer :).

libs/libarchfpga/src/physical_types_util.h
```diff
@@ -325,6 +325,12 @@ int get_tile_class_max_ptc(t_physical_tile_type_ptr tile, bool is_flat);
 /** get information given pin physical number **/
 std::tuple<const t_sub_tile*, int> get_sub_tile_from_pin_physical_num(t_physical_tile_type_ptr physical_tile, int physical_num);
 
+/**
+ * @brief Returns the sub-tile (within 'physical_tile') that owns the given
+ * capacity location, or nullptr if no sub-tile covers that location.
```
AlexandreSinger, line 330: I do not think "capacity location" is defined elsewhere in the code. I know what you mean though, if a tile has multiple sub-tiles with different capacities, each individual physical subtile is indexed by a number which lies within the range of the sub-tile. I think adding a small comment here about what capacity location means would be helpful.

vpr/src/base/check_netlist.h
AlexandreSinger: Small nit: Can this be pass by const reference? Or was there a reason you wanted to use a pointer here?

vpr/src/base/direct_connection_legality.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+#include <set>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "physical_types.h"
+
+class DirectConnectionLegality {
```
AlexandreSinger, line 21: Can you add a Doxygen comment to this class? That way the intellisence will have the documentation.

vpr/src/base/direct_connection_legality.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+} // namespace
+
+DirectConnectionLegality::DirectConnectionLegality(const std::vector<t_direct_inf>& directs) {
+    const int delayless_switch = g_vpr_ctx.device().delayless_switch_idx;
+    std::vector<t_clb_to_clb_directs> clb_to_clb_directs =
+        alloc_and_load_clb_to_clb_directs(directs, delayless_switch);
+
+    forward_by_from_lb_.resize(g_vpr_ctx.device().logical_block_types.size());
```
AlexandreSinger: Small nit: the g_vpr_ctx is only used in the constructor, I am wondering if the necessary information could be passed into the class instead. So either pass in the delayless_switch_idx and number of logical block types, or just pass in the DeviceContext. It makes the interface clear about what information is needed and can make this function easier to test in the future.

vpr/src/base/direct_connection_legality.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    for (const t_clb_to_clb_directs& cd : clb_to_clb_directs) {
+        if (cd.from_clb_type == nullptr || cd.to_clb_type == nullptr) continue;
+
+        // Pin ranges can be given in either direction; step = -1 means reversed.
+        const int from_size = std::abs(cd.from_clb_pin_end_index - cd.from_clb_pin_start_index) + 1;
+        const int to_size = std::abs(cd.to_clb_pin_end_index - cd.to_clb_pin_start_index) + 1;
```
AlexandreSinger, line 46: These absolute values are strange to me. Why are they needed?

AlexandreSinger: Is it because the end index may be lower than the start index at times?

vpr/src/base/direct_connection_legality.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+bool DirectConnectionLegality::is_direct_legal(t_logical_block_type_ptr from_lb, int from_top_pin, t_logical_block_type_ptr to_lb, int to_top_pin) const {
+    if (from_lb == nullptr || from_lb->index < 0
+        || static_cast<size_t>(from_lb->index) >= forward_by_from_lb_.size()) {
+        return false;
+    }
+    const auto& pin_map = forward_by_from_lb_[from_lb->index];
+    auto pin_it = pin_map.find(from_top_pin);
+    if (pin_it == pin_map.end()) return false;
+    return pin_it->second.count({to_lb->index, to_top_pin}) != 0;
```
AlexandreSinger: Can we use the `contains(...)` method here? I think it would be cleaner than count not equal to 0.

vpr/src/base/check_netlist.cpp
```diff
@@ -82,6 +116,10 @@ void check_netlist(int verbosity) {
 
     error += check_for_duplicated_names();
 
+    if (arch != nullptr) {
```
AlexandreSinger: Why is this if statement needed? This is related to making arch a reference since as far as I can tell all usages of the check_netlist method dereference arch to get the pointer.

vpr/src/base/check_netlist.cpp
```diff
@@ -220,6 +258,169 @@ static int check_clb_internal_nets(ClusterBlockId iblk, const IntraLbPbPinLookup
     return error;
 }
 
+static int check_external_directs_legality(const t_arch& arch) {
+    auto& atom_ctx = g_vpr_ctx.atom();
+    auto& cluster_ctx = g_vpr_ctx.clustering();
```
AlexandreSinger: Small nit: Replace auto with type.

vpr/src/base/check_netlist.cpp
```diff
@@ -31,9 +37,37 @@ static int check_clb_conn(ClusterBlockId iblk, int num_conn);
...
+    ClusterBlockId cluster_id;      ///< Cluster driving the Fc_out = 0 OUT pin.
+    const t_pb_graph_pin* out_pin;  ///< The Fc_out = 0 top-level OUT pin.
+    ClusterBlockId sink_cluster_id; ///< Cluster of the sink that couldn't be reached legally.
+    const t_pb_graph_pin* sink_pin; ///< The sink pin within sink_cluster_id.
+};
+
 /*********************** Subroutine definitions *****************************/
 
-void check_netlist(int verbosity) {
+void check_netlist(int verbosity, const t_arch* arch) {
```
AlexandreSinger: Why add this check to check_netlist instead of verify_clustering? TBH this can really go in either spot. I think of this as a clustering issue, but I can see why this can be thought of as a netlist issue.

vpr/src/base/check_netlist.cpp
```diff
@@ -220,6 +258,169 @@ static int check_clb_internal_nets(ClusterBlockId iblk, const IntraLbPbPinLookup
...
+}
+
+static int find_top_level_input_pin_of_sink(const t_pb_routes& pb_route,
+                                            const t_pb_graph_node* pb_head,
+                                            int sink_primitive_pin_id) {
+    int current_pin_id = sink_primitive_pin_id;
+
+    // Follow driver links upstream. The step count bounds the walk in case a
+    // malformed pb_route contains a cycle.
+    for (size_t remaining_steps = pb_route.size(); remaining_steps > 0; --remaining_steps) {
```
AlexandreSinger, line 393: Why is this loop iterating in reverse? Why not count up? This is counting pb_route.size() - 1 times, instead of pb_route.size() times which may be the intention. The issue is that decrementing on size_t values can be dangerous due to underflow.

vpr/src/base/check_netlist.cpp
```diff
@@ -220,6 +258,169 @@ static int check_clb_internal_nets(ClusterBlockId iblk, const IntraLbPbPinLookup
...
+        current_pin_id = entry.driver_pb_pin_id;
+    }
+
+    return -1;
+}
+
+static bool is_sink_routed_within_cluster(const t_pb_routes& pb_route,
+                                          const t_pb_graph_node* pb_head,
+                                          int sink_primitive_pin_id) {
+    return find_top_level_input_pin_of_sink(pb_route, pb_head, sink_primitive_pin_id) < 0;
```
AlexandreSinger, line 422: I think a small comment here would be helpful.

AlexandreSinger (commented):
Thank @amin1377 , this is an interesting PR!

I like having more checks in VPR, and this one is a tricky one. We will need to keep an eye on this since there are some architectures that we do not check in our strong tests. We will find out quickly if there is an issue.

I think you may need to at least collect some run time numbers on VTR or something. I am a bit worried about the run time of this; however, it is run only once during the flow so it should be ok. I am just a bit paranoid of the loops of loops that I am seeing.

libs/libarchfpga/src/physical_types_util.h
amin1377: I tried to use the same vocabulary that is used throughout this file. Here, capacity refers to each subtile instance. I added this clarification to the comment.

vpr/src/base/direct_connection_legality.cpp
amin1377: Other parts of the code use it that way as well.

vpr/src/base/check_netlist.cpp
amin1377: I just wanted to show that the traversal is moving from inside the cluster toward the top-level pin, which is why I’m counting downward. Also, it iterates from N = size() down to 1, so I don’t think there is a risk of underflow.

vpr/src/base/check_netlist.cpp
amin1377: Added the following:
```
// find_top_level_input_pin_of_sink returns >= 0 only when the sink's driver chain
// reaches a top-level input pin (i.e. the signal enters from outside). A negative
// result means the chain ends on an internal driver, so the sink is fed within the cluster.
```

vpr/src/base/check_netlist.cpp
amin1377: I actually originally put it there. However, I wanted this check to run even when we are reading a netlist, so I moved it to check_netlist.

AlexandreSinger: That makes sense! In the future we should combine the two since we should always be checking everything whenever a clustered netlist needs to be verified.

AlexandreSinger (commented):
LGTM

Once the run time results are in this can be merged.

amin1377:
@AlexandreSinger [Results for VTR large benchmarks](https://docs.google.com/spreadsheets/d/13e7wgObUrpNuRAcdAG7ssIxMiVVvjIHdWJOEjeiVsZU/edit?usp=sharing)

QoR is exactly the same, and the runtime is in the same ballpark, with less than a 2% difference.

AlexandreSinger (approved):
LGTM

Issue #3564, opened 2026-05-28
Failing Tight Floorplanning Weekly Test


AlexandreSinger:
The following weekly test is failing on the CI:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/tasks/regression_tests/vtr_reg_nightly_test5/vpr_tight_floorplan/config/config.txt

This is the last failing weekly test. What appears to be happening is that RAMs are now struggling to fulfill the user-region constraints:

<img width="1144" height="436" alt="Image" src="https://github.com/user-attachments/assets/89a2156f-af4b-467a-862f-c930a2d8d800" />

It attempts to re-pack with stronger attraction groups, but it never resolves on a solution. It is likely that a change recently came into master (within the last month) which is causing RAMs to struggle to adhere to region constraints.

AlexandreSinger:
@haydar-c Do you think that your RAM pre-packer may be causing this? How does your RAM pre-packer work with placement constraints? If two RAMs are in different attraction groups, does it know to break them apart?

AlexandreSinger:
As I mentioned in PR #3566, the packer should also use the attraction groups somehow to decide if it should follow the RAM mapper's decisions or not. I can imagine a case where there is such tight floorplanning constraints that the RAM mapping solution could cause no solution to be found.

This may be extremely rare though, so the current RAM mapper solution is more than sufficient.

AlexandreSinger:
> the packer should also use the attraction groups somehow to decide if it should follow the RAM mapper's decisions or not

I left this issue open to track this idea regarding the RAM mapper not following attraction groups. Should I open it as a separate issue and assign it to @haydar-c ?

haydar-c:
I can open a new one mentioning about the attraction groups and considering falling back to the greedy packer for RAMs in the very last iteration of packing if we fail to pack in earlier iterations due to tight floorplanning.

PR #3566, opened 2026-05-29
[Pack] Check partition regions while creating physical rams in ram pre-packer

haydar-c:
#### Description
RAM pre-packer was not checking the partition regions while assigning the RAM slices to physical blocks. That was causing problems when we have tight floorplan constraints. This PR adds partition region aware physical RAM mapping.
#### Related Issue
https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3564


vpr/src/pack/logical_ram_infer.cpp
```diff
@@ -257,29 +258,70 @@ vtr::vector<PhysicalRamGroupId, PhysicalRamGroup> RamMapper::create_physical_ram
...
+                    }
+                }
+
+                // Skip this atom if not partition-compatible
+                if (!pr_compatible) {
+                    remaining_atoms.push_back(atom_id);
+                    continue;
+                }
+
+                // Stop exploring for this group after its full.
```
vaughnbetz: its -> it is

vaughnbetz (approved):
LGTM. Just suggest fixing one tiny typo.

haydar-c:
Thanks for reviewing @vaughnbetz . The nightly tests are also passing: [NightlyTestManual](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/26639668092).

AlexandreSinger (approved):
LGTM! Thank you so much with your help on this Haydar! The weekly tests are passing now for the first time in months!

I think the packer could still improve for region constraints and the RAM mapping, since we should probably completely ignore the RAM mapping decision if the packer is in peril trying to create a solution; however, that case may be sufficiently rare that it may not matter. Something that we may need to explore in the future. I can add it to the issue so we can track it.

Merging now so the weekly tests can finally go green!

PR #3567, opened 2026-05-30
Changed the C style strings in t_model and t_model_ports to std::string

kevinxing-a12:
#### Description

Changed the C style strings in t_model and t_model_ports to std::string and made pertinent and non-logical changes to affected files. (I made a stupid mistake in my commit message by forgetting to mention the change in t_model_ports so don't refer to that one)
#### Motivation and Context

The old C style strings are inconsistent with other parts of VTR where std::string is usually used, and they are hard to maintain. This commit brings the t_model and t_model_ports structs closer to the modern C++ fashion.

I had a successful build, and run three strong regression tests named "strong_timing", "strong_fpu_hard_block_arch" and "strong_soft_multipliers". No issues were found. I am not super sure if more tests need to be run, given that the files I changed span quite many directories.

Environment: pchenry + virtual python + -j16. Test names are mentioned above.

A lot of archfpga_throw() calls and fprintf statements require a conversion back to C string using .c_str(), but I believe I have taken good care of them.

#### Types of changes

#### Checklist:

### DIFF
--- a/libs/libarchfpga/src/arch_check.cpp
+++ b/libs/libarchfpga/src/arch_check.cpp
@@ -24,7 +24,7 @@ bool check_model_clocks(const t_model& model, const char* file, uint32_t line) {
             if (!port->clock.empty() && !clocks.count(port->clock)) {
                 archfpga_throw(file, line,
                                "No matching clock port '%s' on model '%s', required for port '%s'",
-                               port->clock.c_str(), model.name, port->name);
+                               port->clock.c_str(), model.name.c_str(), port->name.c_str());
             }
         }
     }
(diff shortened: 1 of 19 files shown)


kevinxing-a12:
Ok it seems that I failed to catch all the strcmp() nuisance in Odin (maybe the regular "make" doesn't cover it?). Will work on that tomorrow.

AlexandreSinger:
> Ok it seems that I failed to catch all the strcmp() nuisance in Odin (maybe the regular "make" doesn't cover it?). Will work on that tomorrow.

Odin is the old Synthesis tool we used to use in VTR. We have since moved to Parmys. By default. VTR builds with Parmys, but our CI does still build Odin just to be safe and runs a few tests with it to ensure it does not regress. CI is doing its job!

It should be pretty simple to support Odin; but if you run into really bad issues do let me know. Odin is currently deprecated, meaning that we are not actively improving on it and we intend to drop it in the future.

PR #3569, opened 2026-05-31
[SDC] Added support for the -add option for create_clock

AlexandreSinger:
This required a small modification to Tatum to allow multiple clocks to be defined on a single clock pin.

I also needed to expand the syntax for set_clock_groups to allow for different types of clock groups that -add option is used for. It honestly does not matter what clock group you use in VPR, but its nice to be accurate.

This now allows external clock muxing to be supported in VPR; this is when you have a single port which can be one of two possible clocks.

Added strong tests to lock in the new features.

After this PR is merged, I plan to add support for the -add option for create_generated_clock as well; but that requires syntax updates to allow for the -master_clock option. I did not expect to need it, but it is important when you have pins with multiple clocks.


AlexandreSinger:
@vaughnbetz This PR is ready for review when you have a moment. Some explanations of the changes:

Previously, the only clock group relationship we could define in VPR was -asynchronous (previously -exclusive):
image

With this PR, we now can explicitly write -logically_exclusive clock groups, which can occur with internal clock muxing within circuits:
image
VTR was capable of doing clock muxing before, but these clocks were always labeled as -asynchronous. For VPR, this makes literally no difference; but technically this relationship is logically exclusive.

With this PR, we now have the support for the last clock group relationship which is -physically_exclusive:
image
This allows us to define multiple clocks on the same pin. This can occur when the clocks are muxed externally to the FPGA device. Since they do not physically exist on the device at the same time, this relationship is physically_exclusive.

The timing analysis for the circuit above is correct from Tatum as far as I can tell:
image
Notice how there now exists two independent timing paths from FFA to FFB which are both being timing analyzed.

From digging around, it appears that -asynchronous, -logically_exclusive, and -physically_exclusive are all synonymous in the context of STA in VPR; so they are all trivially supported. However, its good to be able to be precise about the timing considerations of circuits.

libs/EXTERNAL/libtatum/libtatum/tatum/graph_visitors/CommonAnalysisVisitor.hpp
```diff
@@ -167,54 +167,58 @@ bool CommonAnalysisVisitor<AnalysisOps>::do_arrival_pre_traverse_node(const Timi
...
+                //Note that we enumerate all pairs of valid launch domains for the current domain
+                //(which is now treated as the capture domain), since each pair may have different constraints
+                for(DomainId launch_domain_id : tc.clock_domains()) {
+                    if(tc.should_analyze(launch_domain_id, domain_id)) {
+
+                        //Initialize the clock capture tag with the constraint, including the effect of any source latency
+                        //
+                        //Note: We assume that this period constraint has been resolved by edge counting for this
+                        //domain pair. Note that it does not include the effect of clock uncertainty, which is handled
+                        //when the caputre tag is converted into a data-arrival tag.
```
vaughnbetz: caputre -> capture

vaughnbetz (approved):
Looks good, thanks!
Found one typo, but that's it.  

Not sure the example of physically exclusive vs. logically exclusive is quite correct (I think the two cases you mention are at least mostly both physically exclusive) but it doesn't matter much since as you note SI is not modeled in Tatum/VPR.

Issue #3528, opened 2026-05-08
[Placement] Macros Not Being Created for Sub-tiles


AlexandreSinger:
In the strong tests, there are testcases with the naming patterns:

- `strong_direct_*`
- `strong_tileable_rr_graph_direct_*`

These testcase place and route onto an architecture with direct connects between sub-tiles: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vtr_flow/arch/timing/k4_frac_N8_reset_softadder_register_scan_chain_dsp8_mem1K_nonLR_caravel_io_skywater130nm.xml

This architecture contains a tile with three sub-tiles:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/b48c6f5f38ce61a6b70122f098d5530b12742fb1/vtr_flow/arch/timing/k4_frac_N8_reset_softadder_register_scan_chain_dsp8_mem1K_nonLR_caravel_io_skywater130nm.xml#L232-L277

At sub-tile 0 it has an 8-bit multiplier, which means it has a 16 bit output. At sub-tiles 1 and 2 it has RAM blocks. The idea of this tile is that the lower 8 bits of the multiplier output are directly connected to one RAM, and the upper 8 bits of the multiplier output are directly connected to the other RAM. This is done through direct connects:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/b48c6f5f38ce61a6b70122f098d5530b12742fb1/vtr_flow/arch/timing/k4_frac_N8_reset_softadder_register_scan_chain_dsp8_mem1K_nonLR_caravel_io_skywater130nm.xml#L388-L394

Normally this would create a macro in placement since there are hard placement constraints between the multiplier and RAM blocks. However these macros are not being created.

I believe in the default flow, the clusters just so happen to be created and placed in the correct order that these constraints are met and the annealer is not messing them up; however, this is not a stable solution.

When this testcase is passed through the AP flow, AP is randomizing the placement of these clusters some and this testcase is not routing. A stable solution to this problem requires that the macro generator be able to recognize this pattern.

AlexandreSinger:
@vaughnbetz FYI

PR #3570, opened 2026-06-01
[VPR][Flow] Set AP as the default flow

AlexandreSinger:
Analytical Placement has been shown to improve wirelength on all the benchmarks by around 10%, with similar CPD and wirelength. It has also been shown to work on every VTR-compatible architecture and circuit.

Setting AP as the default flow in VPR. The old flow can still be used by setting the stages to --pack --place --route --analysis


AlexandreSinger:
Raising this as a PR so I can track what changes are needed in the CI by setting the default flow to AP.

The weekly tests are currently failing due to an issue with device size estimation causing the test machine to run out of memory (device sizes are set too large). We are working on resolving this before we run the weekly test to avoid destroying our CI runner.

AlexandreSinger:
@amin1377 FYI

AlexandreSinger:
Here are the issues that are currently being tracked:

- [x] Functional simulation has an issue with the architecture not routing for specific channel widths. Resolved in #3572 
- [x] S: Basic has an issue where the RR graph is not being regenerated after the device is resized for the AP flow when timing analysis is disabled. Resolved in #3575 
- [x] R: Strong is failing on 14 cases:
  - [x] Strong verify rr graph testcases are failing due to the device size not matching when an RR graph is being read in. PR: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3585
  - [x] Direct connects between sub-tiles has random failures even in the default flow. The feature is not rigourously locked in as far as I can tell. Issue has been raised: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3528 . PR has been raised locking this test to the non-AP flow: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3599
  - [x] Issue with the tracking of molecules within clusters when pin_feasibility_filter is off. PR raised here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3581
- [x] R: Strong_odin is failing on 2 cases. This is strong_verify_rr_graph{_bin} which is also failing in regular strong. It is a known issue where reading in the RR graph may change the device size and cause a crash. We need the device size estimator to read the RR graph and fix the size to the RR graph's size.
- [x] NightlyTest7 is using way too much RAM due to the device size estimator creating too large of devices. PR: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3577
- [x] Still need to run the full weekly tests for failures. Weekly tests launched: https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/26817710901
  - [x] NightlyTest2 has three failures where the RR graph is being read in. Same error as the strong tests.
  - [x] NightlyTest5 has three failures for vpr_tight_floorplan_3d. Need to investigate. These failures are strange. They are placement constraint failure issues when the macros are formed. Something weird is going on that is likely not AP-specific.
- [x] All golden results need to be regenerated. This must be done at the last moment and each one should be checked for weird behaviors or regressions.
  - [x] We will need to revisit how the parse_results are getting populated when the AP flow is on. I think some arguments like pack_time may not be matched anymore and should be fixed to match the packer run time (APPack or old packer). Other options to look out for: `device_name`, `device_grid_tiles`, `place_time`. PR: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3583
  - [x] We will need to go through all of the failing QoR tests and ensure that they are "within reason". We expect the device size to increase a small amount due to AP usually preferring a lower density than the current default. We want to look for any bad behavior that may need to be investigated.
  - [x] We will then need to regenerate the golden results for all failing tests, including the weekly tests.
- [x] AP docs are out of date and need to be updated. PR:  https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3584
- [x] Add a few basic and strong tests which explicitly run the --pack --place --route flow instead of AP. This will lock in the old flow and ensure it does not break. PR: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3592
- [x] Clear documentation for how to run the "legacy" flow.

AlexandreSinger:
@haydar-c FYI. I am using this PR to track the progress of getting AP into default.

amin1377:
@AlexandreSinger As discussed in the VTR meeting, once you are done with all the changes, could you please post the results here comparing the master branch vs. AP for large VTR, Titan, and Koios?

Thanks!

AlexandreSinger:
> As discussed in the VTR meeting, once you are done with all the changes, could you please post the results here comparing the master branch vs. AP for large VTR, Titan, and Koios?

Hi @amin1377 , yes, I am planning on collecting the fixed-sized devices tonight (when Wintermute is less busy) and then I will run the non-fixed-sized devices tomorrow night. I will put all of the results in this thread.

AlexandreSinger:
Here are the results across my fixed-device size test suite. Wirelength and CPD are post-route. Results are normalized as AP_FLOW / Traditional Flow. A timeout of 12 hours was used (any circuit that was timed out was removed from the geomeans). AP always completed within the time limit, the traditional flow did not complete on 5 circuits which is shown in the table below:

test_suite | total_wirelength | crit_path_delay | total_runtime | Baseline Unroutes
-- | -- | -- | -- | --
titanium_s10 | 0.88 | 0.95 | 0.95 | 2
titan | 0.90 | 1.01 | 1.07 | 2
koios | 0.92 | 0.97 | 1.04 | 1
artix_7 | 0.90 | 1.02 | 0.88 | 0
vtr_largest | 0.95 | 0.99 | 1.07 | 0

Raw (per-circuit) results can be found here: 
[june_10_full_suite.xlsx](https://github.com/user-attachments/files/28813300/june_10_full_suite.xlsx)

On this suite, AP improves wirelength by 8% across all benchmarks with slightly better CPD and similar overall run time. It also was able to complete 5 more circuits than the original flow. This matches the last run I did, plus a small amount of noise.

I will collect the non-fixed device runs now (from the usual NightlyTest tasks we run for QoR) and will post here after that is done.

@amin1377 FYI.

AlexandreSinger:
Results on titan_quick_qor task script from nigthly test 2. This is the exact script, I just moved from master branch to this branch:

  | titan_default.txt | titan_ap.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.18
num_LAB | 1 | 1.13
num_DSP | 1 | 1.24
num_M9K | 1 | 1.00
num_M144K | 1 | 1.00
max_vpr_mem | 1 | 1.15
num_post_packed_blocks | 1 | 1.12
device_grid_tiles | 1 | 1.10
pack_time | 1 | 0.90
placed_wirelength_est | 1 | 0.86
place_time | 1 | 0.89
placed_CPD_est | 1 | 0.99
routed_wirelength | 1 | 0.87
critical_path_delay | 1 | 1.03
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 0.98
crit_path_route_time | 1 | 0.93

Raw data: 
[comparison_output.xlsx](https://github.com/user-attachments/files/28846271/comparison_output.xlsx)

The key thing to notice is that the number of post-packed blocks increase by 12%. This makes sense since AP tends to reduce the density of the circuit by recognizing that it can use the space more efficiently. This leads to the number of device grid tiles (i.e. the device width and height) increasing by 10% which leads to around a 15% increase in memory and runtime.

The wirelength is improved by 13% (which is slightly better than my results on the fixed sized architecture) and the CPD is worse by around 3%. It is worth noting that the placed CPD estimate is tied if not a little better, so it may just be a bit noisy here (or the results of some tuning). If you look at the per-circuit results, you will find that many circuits are improving CPD by a lot, only 5 are actually worse and I think a couple of outliers bring up the average. Something for me to explore in the future. I have some ideas what is causing this.

@amin1377 FYI

AlexandreSinger:
These are the results of the koios_medium task from nightly test 4:

  | koios_medium_default.txt | koios_medium_ap.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.11
odin_synth_time |   |  
parmys_synth_time | 1 | 1.01
abc_depth | 1 | 1.00
abc_synth_time | 1 | 1.02
num_clb | 1 | 1.07
num_memories | 1 | 1.00
num_mult |   |  
max_vpr_mem | 1 | 1.03
num_post_packed_blocks | 1 | 1.04
device_grid_tiles | 1 | 1.02
pack_time | 1 | 0.99
placed_wirelength_est | 1 | 0.93
place_time | 1 | 0.76
placed_CPD_est | 1 | 1.01
routed_wirelength | 1 | 0.94
critical_path_delay | 1 | 1.02
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 1.01
crit_path_route_time | 1 | 1.08

Raw data:
[koios_medium_ap.xlsx](https://github.com/user-attachments/files/28846530/koios_medium_ap.xlsx)

These results are much more modest. The device sizes did not change by very much. Wirelength is still improved and CPD is very slightly worse. Koios medium contains only 11 circuits and they are all very small (less than 7k LUTs) so this makes sense.

AlexandreSinger:
Here is the vtr_reg_chain task results:

  | chain_vtr_default.txt | chain_vtr_ap.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.16
odin_synth_time |   |  
parmys_synth_time | 1 | 1.00
abc_depth | 1 | 1.00
abc_synth_time | 1 | 1.00
num_clb | 1 | 1.03
num_memories | 1 | 1.00
num_mult | 1 | 1.03
max_vpr_mem | 1 | 1.11
num_post_packed_blocks | 1 | 1.02
device_grid_tiles | 1 | 1.03
pack_time | 1 | 0.91
placed_wirelength_est | 1 | 0.95
place_time | 1 | 0.70
placed_CPD_est | 1 | 1.00
min_chan_width | 1 | 0.95
routed_wirelength | 1 | 0.96
min_chan_width_route_time | 1 | 1.52
crit_path_routed_wirelength | 1 | 0.96
critical_path_delay | 1 | 1.00
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 1.00
crit_path_route_time | 1 | 0.96

This one is doing a min-channel width sweep, so that can have an affect on the run time and memory usage due to which widths are tried. It looks like the device size is around the same. The minimum channel width was reduced by around 5%, with a 4% reduction in WL and the same CPD.

vaughnbetz (approved):
Looks great!  Documentation and changes look good, and testing looks good.

Tuning ideas for the future:
- We're using more DSP blocks -- is this a good idea? Could use the compressed device grid in range tests to have more dense hard block packing.
- What is result quality like on the smallest possible devices that both flows can fit on (and how different are those smallest possible devices)?  With desperation mode / unrelated clustering and no pin utilization limits, analytical placement may be significantly better than the traditional packer. How much denser is that limit than the default device sizes / auto selection?  How do ap vs. legacy compare at that point.

amin1377:
@vaughnbetz I was thinking that, once AP becomes the default flow with this PR, we may no longer regularly run VTR large, Titan, Koios, etc. with the current default flow (`pack+place+route`).
Should we duplicate some tests, maybe Titan and a few of the stronger tests, so the current default flow is still tested regularly?

vaughnbetz:
That's a good idea.  @AlexandreSinger : can you add that to your to do list?

AlexandreSinger:
Sure! I have already added some smoke tests to ensure that the old flow still works, but I can add some stronger tests as well.

AlexandreSinger:
A strong test sneaked in with failures on AP. It is the same interposer cost term update issue we saw in a previous issue. One bug went down and another popped up:
vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_interposer/vtr_flagship_interposer/config/config.txt

I am locking this task onto the original pack, place, and route flow while I transition to AP. I will raise an issue.

AlexandreSinger:
All golden results are updated (included Weekly tests) and CI is green. Vaughn has approved the PR including the changes to documentation. Merging now.

PR #3572, opened 2026-06-01
[CI] Locked the chanW of the functional verif tests

AlexandreSinger:
I found that the architecture these run on have weird unroutes at specific channel widths. This is not for the functional verification tests to mess with so I fixed the channel width. This also speeds up these tests very slightly.

This was revealed by making AP default, but these unroutes also appear in the current pack, place, and route flow.


haydar-c (approved):
LGTM @AlexandreSinger, thanks!

PR #3573, opened 2026-06-01
Add APPack interposer gain multiplier

AmirhosseinPoolad:
Adds a new option --appack_inter_die_gain_multiplier with a default value of 0.1 that penalizes packing together molecules that are placed on different dice.

inter-die gain multiplier | 1 | 0.75 | 0.5 | 0.25 | 0.1
-- | -- | -- | -- | -- | --
vtr_flow_elapsed_time | 1 | 1.003303723 | 0.9920243437 | 1.005093034 | 0.9985976003
max_vpr_mem | 1 | 0.9999967307 | 0.9999687007 | 1.000255661 | 1.000107037
place_time | 1 | 1.006803593 | 0.9923678069 | 1.006045928 | 0.9983451511
routed_wirelength | 1 | 0.9983856128 | 0.9970706835 | 0.9996980128 | 1.00021068
critical_path_delay | 1 | 1.022312659 | 1.007497817 | 0.9858393346 | 0.9789470556
crit_path_route_time | 1 | 0.990572735 | 1.002302073 | 1.007940993 | 1.007240622

Results (Koios) show CPD gains without any run-time cost.

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -2132,7 +2132,9 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .help(
             "Multiplier applied to APPack candidate gains when the candidate's "
             "flat placement location is on a different die than the current "
-            "cluster location.")
+            "cluster location in an interposer-based architecture. This does "
+            "not apply to candidates on a different layer in a 3D architecture "
+            "without interposer cuts.")
         .default_value("0.1")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -2128,6 +2128,14 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value({"auto"})
         .show_in(argparse::ShowIn::HELP_ONLY);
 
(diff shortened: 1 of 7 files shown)


vpr/src/base/read_options.cpp
```diff
@@ -2128,6 +2128,14 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value({"auto"})
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    ap_grp.add_argument(args.appack_inter_die_gain_multiplier, "--appack_inter_die_gain_multiplier")
+        .help(
+            "Multiplier applied to APPack candidate gains when the candidate's "
+            "flat placement location is on a different die than the current "
+            "cluster location.")
```
AlexandreSinger: This is called "inter_die", does it also apply for 3D dies? It does not appear to according to the code. I think it may be worth mentioning in this description.

Also, if you think its useful to put in the docs, could you add this to: doc/src/vpr/command_line_usage.rst

AlexandreSinger (commented):
Looks great @AmirhosseinPoolad ! Thanks for looking into this, the results look great! Just one comment on documentation.

As we discussed in-person, I would love to see if, past a certain distance, it may be worth just removing candidates if they are on different dies. This may increase the number of clusters, but based on these results its usually a bad idea to cross die boundaries. But that is something that can be in another PR.

AlexandreSinger (approved):
LGTM

PR #3575, opened 2026-06-01
[AP][Legalizer] Cleaned Up RR Graph Regeneration

AlexandreSinger:
I found that when timing analysis is turned off, the RR graph was being generated prior to global placement. Without timing analysis the RR graph is not needed, so it was a waste.

I also found an independent issue where if the RR graph existed before packing and the device was resized, the RR graph was not regenerated. This caused issues in packing since some data structures were the wrong size.


vpr/src/analytical_place/full_legalization/full_legalizer.cpp
```diff
@@ -1313,14 +1299,7 @@ void APPack::legalize(const PartialPlacement& p_placement) {
...
 
-    // If auto device sizing is used, recreate the device grid after final
-    // clustering and before initial placement. The earlier grid was based on
-    // a pre-clustering estimate, but the required device size can change once
-    // the final clustering is known. Rebuilding the grid also invalidates the
-    // RR graph generated for the estimated device size as required.
-    if (vpr_setup_.PackerOpts.device_layout == "auto") {
-        vpr_create_device_grid(vpr_setup_, arch_);
-    }
+    recreate_device_if_needed();
```
haydar-c, line 1302: As we talked, calling this function after packing and before initial placement is a good solution for now. I verify that this version is working for all 4 combinations of over/under-estimate and timing_analysis on/off (also with the minimum number of RRG build instances). However, when the RRG is build by which part of the code is still implicit. We can make a more explicit handler in a future PR (not for performance or functionality as this pr resolves them but) for better understandability.

haydar-c (approved):
Thanks @AlexandreSinger! LGTM, the RRG generation path in the auto device size for AP is simpler now. Adding a small comment below for future reference.

PR #3577, opened 2026-06-02
[Pack] Use pb_graph connectivity to select block type in device size estimator

haydar-c:
The device size estimator previously selected the first candidate block type for each atom (except for RAMs, which are handled by the RAM mapper). This caused over-estimation when the first candidate type was not the one the packer would actually choose; particularly when that type is much scarcer on the device grid than the alternatives, inflating the estimated resource demand and causing the estimator to select a larger device than necessary.

An example is inpad atoms on Stratix 10. They can be implemented by both io_cell and OCT complex blocks, but since OCT appears first in the architecture, all inpad atoms were being assigned to it. OCT blocks are far scarcer on the Stratix 10 grid than io_cell blocks, so this led to the inflation of the estimated device size incorrectly.

To fix this, instead of always taking the first candidate, we now prioritize the candidate type whose representative primitive has a forward path to the root block's external output pins. A primitive that determines block occupancy should drive signals that exit the block; primitives whose outputs are routed only to other internal primitives are a weaker basis for resource estimation, since their resource demand is already captured by the primitives they feed. If no candidate passes the check as is the case for outpad atoms (which have no pb_graph output pins), we fall back to the first candidate, preserving the original behavior.

The Stratix 4 architecture results are unchanged, since each non-RAM atom maps to exactly one candidate block type there except for RAMs. The Stratix 10 results improve due to correct assignment of inpad atoms. The mean absolute relative error for Stratix10 and Stratix4 are 4% and 5%, respectively for device width.

**Stratix10 Summary**
| Metric | Master | PR |
|---------|---------:|---------:|
| Mean Absolute Error (Estimated Width − Actual Width) | 603.90 | **7.24** |
| Mean Absolute Relative Error ((Estimated Width − Actual Width) / Actual Width) | 1.83 | **0.04** |

**Stratix4 Summary**
| Metric | Master | PR |
|---------|---------:|---------:|
| Mean Absolute Error (Estimated Width − Actual Width) | 9.32 | 9.32 |
| Mean Absolute Relative Error ((Estimated Width − Actual Width) / Actual Width) | 0.05 | 0.05 |


vpr/src/pack/device_size_estimate.cpp
```diff
@@ -229,9 +270,24 @@ std::map<t_logical_block_type_ptr, size_t> DeviceSizeEstimator::estimate_resourc
...
+        for (t_logical_block_type_ptr cand : primitive_candidate_block_types[root_model_id]) {
+            std::vector<t_logical_block_type> candidate_logical_type = {*cand};
+            const t_pb_graph_node* current_type_prim = prepacker.get_expected_lowest_cost_primitive_for_atom_block(root_atom, candidate_logical_type);
+            if (current_type_prim && primitive_has_external_output(current_type_prim)) {
+                mapped_type = cand;
+                break;
+            }
+        }
+        if (!mapped_type)
+            mapped_type = primitive_candidate_block_types[root_model_id][0];
```
AlexandreSinger, line 288: Instead of picking the first one, why not select the one that is most common on the device? Or the one that is the most "average". I am just thinking of ways to prevent this from happening in the future.

AlexandreSinger (approved):
LGTM, I just left a small note for future work.

PR #3581, opened 2026-06-02
[Pack] Fixed Molecule Tracking Bug

AlexandreSinger:
When the pin_feasibility filter was disabled, the molecules were not being tracked as being part of the cluster. This is not intuitive and was causing division by zero in the AP flow.

Resolved the tracking issue in the packer.

### DIFF
--- a/vpr/src/pack/cluster_legalizer.cpp
+++ b/vpr/src/pack/cluster_legalizer.cpp
@@ -1288,23 +1288,24 @@ e_block_pack_status ClusterLegalizer::try_pack_molecule(PackMoleculeId molecule_
                                                          mutable_atom_pb_lookup());
         }
 
-        // If we're using the pin feasibility filter, we need to add the candidate molecule to
-        // cluster.molecules. We use this flag to control the cleanup in case of failure.
+        // This flag controls the pop_back cleanup of cluster.molecules in case of subsequent failure.
         bool candidate_molecule_added_to_cluster = false;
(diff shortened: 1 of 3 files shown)


vpr/src/pack/cluster_legalizer.cpp
```diff
@@ -1288,23 +1288,24 @@ e_block_pack_status ClusterLegalizer::try_pack_molecule(PackMoleculeId molecule_
...
-        // If we're using the pin feasibility filter, we need to add the candidate molecule to
-        // cluster.molecules. We use this flag to control the cleanup in case of failure.
+        // This flag controls the pop_back cleanup of cluster.molecules in case of subsequent failure.
         bool candidate_molecule_added_to_cluster = false;
 
-        if (enable_pin_feasibility_filter_ && block_pack_status == e_block_pack_status::BLK_PASSED) {
-            // try_update_lookahead_pins_used needs the candidate molecule to be in cluster.molecules.
+        if (block_pack_status == e_block_pack_status::BLK_PASSED) {
             cluster.molecules.push_back(molecule_id);
             candidate_molecule_added_to_cluster = true;
```
haydar-c, line 1296: This is more intuitive to me as well. Adding successful cluster molecules should not be gated by pin feasibility if user didn't enable it.

haydar-c (approved):
LGTM, I agree this version is more intuitive.

PR #3583, opened 2026-06-02
[AP] Updated Task Parsing

AlexandreSinger:
The task parsing between the AP flow and non-AP flows were a bit strange. We should be able to still collect the same data regardless of flow, we just need the analog. For example, pack_time in the non-AP flow is the time it takes to pack; in the AP flow we can still collect this time.

This resolves a lot of golden result failures when AP is activated by default.


AlexandreSinger:
@haydar-c When you have a mement, can you please review this PR?

vpr/src/base/setup_grid.h
```diff
@@ -30,3 +30,12 @@ DeviceGrid create_device_grid(const std::string& layout_name,
  *        (size of the bounding box of non-empty grid tiles)
  */
 size_t count_grid_tiles(const DeviceGrid& grid);
+
+/**
+ * @brief Logs a one-line summary of the device grid dimensions and tile count.
+ *
+ * The format of this log line is parsed by VTR's task parser to track device
+ * size as a QoR metric across regression runs. Do not change the format without
+ * also updating the task parser's regex patterns.
```
haydar-c, line 39: We should comment these more, just taking some future notes for myself as well (the logs used in the parser's regex patterns).

haydar-c (approved):
LGTM, @AlexandreSinger. Thanks for the simplification in logging code as well.

vpr/src/base/setup_grid.h
AlexandreSinger: I agree. I hope to make a cleaner logging interface for information people care about. The pattern matcher is a bit messy at times.

PR #3585, opened 2026-06-03
Use the device size from provided RR graph

haydar-c:
When vpr reads an rr graph in, we should use the device size implied by the rr graph. If user also provides a fixed layout at the same time with the rr graph, the grid generated by both should match. This PR ensures these two conditions are always met through DeviceSizeEstimator for both xml and bin rr graphs.

Previously, we were not using the device size implied by the rr graph provided and that was causing problems when the estimated device size is not matching the rr graph device size.


haydar-c:
When we read in the rr graph with auto device size, we set the device grid to device size implied by rr graph. However, the packer can expand or shrink the device size after clustering finishes. Then rr graph parser complains about this mismatch as expected even though the CI is green (not testing the case mentioned). Working on a fix for this discrepancy so making the PR WIP for now.

AlexandreSinger:
> When we read in the rr graph with auto device size, we set the device grid to device size implied by rr graph. However, the packer can expand or shrink the device size after clustering finishes. Then rr graph parser complains about this mismatch as expected even though the CI is green (not testing the case mentioned). Working on a fix for this discrepancy so making the PR WIP for now.

Ah! I was about to put that in my review :)

That makes sense. I think we should consider making this more formal somehow. I find it a bit strange that your code is reading the RR graph independently. I wonder if it would be easier to just force the RR graph to load when we do device estimation. I recognize that this may lead to cases where the RR graph is loaded when it is not needed (which is bad for RAM), so I will let you decide on that.

I also noticed that you get the device width and height by parsing the tiles in the RR graph; does the RR graph not contain specific metadata about the grid dimension? Perhaps this is something that should be added to the RR graph data structure. @amin1377 Let me know what you think about this.

I think if we can fix the device width and height when the RR graph is read in (so it does not change after or during packing), this PR can be merged; but we should think on the issues above to make this better supported.

haydar-c:
Thanks for the comments @AlexandreSinger.

>  I wonder if it would be easier to just force the RR graph to load when we do device estimation. I recognize that this may lead to cases where the RR graph is loaded when it is not needed (which is bad for RAM), so I will let you decide on that.

I think about the same approach but on top of loading the RR graph fully, that path requires a device grid to load the RR graph unfortunately.

> Perhaps this is something that should be added to the RR graph data structure

I agree with this completely. I will document that in the parsing function for future.

amin1377:
@AlexandreSinger Grid width and height are not explicitly stored in the rr_graph file. I agree that it would be good to store them so we don’t have to infer the grid size from node locations, etc.

AlexandreSinger (approved):
LGTM. Thanks @haydar-c .

Can you also raise an issue to track the RR graph thing. I think this may be important for long term support. Not super urgent right now though.

PR #3588, opened 2026-06-03
[CI][HOTFIX] Fixed FASM Unit Test Failure

AlexandreSinger:
The VTR unit tests were not running in CI due to missing CMAKE files in
the build folder. After fixing this, we found that the FASM tests were
failing. The fix was basic.
    
The issue is that --device must be provided when no `auto` device
exists. We need to discuss to see if this is expected and if so we
should provide a better error than just crashing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,10 @@ jobs:
         export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
         make -j${{ steps.cpu-cores.outputs.count}}
 
+    # Here we pack the build to store it as a build artifact. We exclude specific files
+    # to keep the build artifact reasonably low (to speed up file download times in the
+    # CI). Beyond the executables, some files need to be kept such as CTestTestFile.cmake
+    # files which are used by the unit tests CI run.
     - name: Pack Build
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


AlexandreSinger:
Unit tests are now running and not failing:
image

.github/workflows/test.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     - name: Pack Build
       shell: bash
       run: |
-        tar -czvf build.tar.gz --exclude='CMakeFiles' --exclude='*.a' --exclude='*.cmake' build/
+        tar -czvf build.tar.gz --exclude='CMakeFiles' --exclude='*.a' --exclude='cmake_install.cmake' build/
```
AmirhosseinPoolad, line 72: Does this change the zipped file's size in anyway? I also don't really get why it's there, would be worth to add comment.

AlexandreSinger: It would increase it slightly to include a few specific CMAKE files. The CI prints which files are being compressed:
image

There was no specific reason I excluded all CMAKE files originally beyond me just thinking that they were not necessary.

I agree that a comment will be useful. I will add that now.

AlexandreSinger:
Just to have this written down somewhere, here is the change is size for the build artifact:

- Before: 72641145 bytes
- After: 72656705 bytes

Which is only around a 0.02% increase.

PR #3589, opened 2026-06-03
[CRR] Remove preserve pin

amin1377:
To simplify the CRR code, the option to preserve IPIN/OPIN connections created by the tileable RR graph generator has been removed.

Now, in CRR patterns, both routing-wire connections and IPIN/OPIN connections are generated by the CRR generator.


vpr/src/route/rr_graph_generation/tileable_rr_graph/tileable_rr_graph_edge_builder.cpp
```diff
@@ -373,20 +371,11 @@ void build_rr_graph_regular_edges(const RRGraphView& rr_graph,
 
             t_track2pin_map track2ipin_map; /* [0..track_gsb_side][0..num_tracks][ipin_indices] */
             t_pin2track_map opin2track_map; /* [0..gsb_side][0..num_opin_node][track_indices] */
+            t_track2track_map sb_conn;      /* [0..from_gsb_side][0..chan_width-1][track_indices] */
```
AlexandreSinger: Minor nit: These three variables are only used in the else condition. I think they can be moved into that scope.

AlexandreSinger (commented):
LGTM, just one minor nit.

The strong_crr test is failing with a much higher CPD and slack values. But that is likely due to the CRR architecture changing so much. Just flagging.

AlexandreSinger (approved):
LGTM

PR #3590, opened 2026-06-03
Update dependency install script to use --dev 

ayan20985:
#### Description

Updated both the apt and dnf files to use --dev for installing dependencies like verilator, etc. Had to update the docker file to use the new --dev instead of parsing the file directly.

#### Related Issue

Discussed with Alex. 

#### Motivation and Context

Useful for the future.

CI should pass, will fix if it doesn't.

#### Types of changes

#### Checklist:

- [X] My change requires a change to the documentation


vaughnbetz (commented):
Looks OK to me. Can you give more context on the reason for the change than 'useful for the future'?
I assume we're installing verilator if you say you want to develop in VTR, rather than just running it (--dev)?

ayan20985:
> Looks OK to me. Can you give more context on the reason for the change than 'useful for the future'? I assume we're installing verilator if you say you want to develop in VTR, rather than just running it (--dev)?

I think the main reason here is that people working on VTR can use the --dev when building it so all the required dependencies (i.e. verilator) are installed without bloating the end-user build. As it stands verilator is only used in the CI and so it should remain there unless it becomes a checker the end-users can also run down the line.

Dockerfile
```diff
@@ -10,10 +10,9 @@ COPY . ${WORKSPACE}
 ENV PIP_BREAK_SYSTEM_PACKAGES=1
 # Install and cleanup is done in one command to minimize the build cache size
 RUN apt-get update -qq \
-# Extract package names from install_apt_packages.sh
-    && sed '/sudo/d' install_apt_packages.sh | sed '/#/d' | sed '/packages_to_install/d' | sed '/)/d' | sed '/if\s.*then$/d' | sed '/else$/d' | sed '/fi$/d' | sed '/echo\s/d' | sed 's/ \\//g' | sed '/^$/d' | sed '/^[[:space:]]*$/d' | sed 's/\s//g' \
-# Install packages
-    | xargs apt-get -y install --no-install-recommends \
+# install_apt_packages.sh uses bash control flow (--dev) instead of sed parsing
```
vaughnbetz: Update comment -- I think you want to say that you're creating a docker image that includes the packages needed by developers (--dev), not just those needed by people who want to run vtr

install_apt_packages.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+install_dev=false
```
vaughnbetz, line 5: Add a comment: if --dev is specified, include packages that developers of vtr would need; otherwise only install the packages needed to run vtr.

install_dnf_packages.sh
```diff
@@ -1,3 +1,12 @@
+#!/bin/bash
+
+install_dev=false
```
vaughnbetz, line 5: Add comment on what --dev means (same comment as above)

vaughnbetz (changes requested):
A few comments to change / add

Dockerfile
```diff
@@ -10,10 +10,9 @@ COPY . ${WORKSPACE}
...
+    && apt-get install -y --no-install-recommends sudo \
+    && bash install_apt_packages.sh --dev \
```
AlexandreSinger: As we discussed in-person, this may be an ok change; but we would need to measure the build cache size before and after this change (since that was the reason this originally was written this way).

An alternative to this is to add another flag to the bash file which prints out all of the APT packeges as a raw string which can be used here. This may be the cleanest approach.

AlexandreSinger (changes requested):
I like this change, but I have a small note on removing the SED commands. When I first started working on this file, I saw the SED commands and left them since the comment said it was an optimization to reduce cache size. From a quick search I found that to actually be good practice, but I never verified it. If we are to remove the SED commands, we would want to ensure there is not a cache hit. My worry is that whoever did this did it to resolve some issue.

vaughnbetz (approved):
LGTM

vaughnbetz:
This looks good @ayan20985 . Can you resolve the conflicts and merge?

Dockerfile
ayan20985: I will just revert to SED parsing as my changes can be consolidated into a more robust docker CI testing scheme down the road given that it doesn't really contribute much as it stands.

AlexandreSinger (approved):
LGTM. I also went to the Qt branch to ensure the two dont conflict. We should be fine as far as I could tell. Thanks for adding the --dev option! I think it will be helpful!

PR #3592, opened 2026-06-03
[CI] Add Pack Place and Route Flow Strong Test

AlexandreSinger:
Now that we are trying to move the default flow from the traditional pack, place, and route flow to the AP flow, I was worried that we would no longer be testing the old flow anymore once we move (since everyone assumed that the old flow was always running). I added a strong test which explicitly tests the old flow on some small circuits as a smoke test to ensure the old flow is working.


AmirhosseinPoolad:
I strongly think this should be part of the other PR. As of now this would be a redundant test.

AlexandreSinger:
> I strongly think this should be part of the other PR. As of now this would be a redundant test.

Let's discuss. I am happy to hold onto this until AP is just about to go in. I want to keep that other PR very lightweight since it will change tons of golden results and I want to ensure everything is reviewed completely.

AmirhosseinPoolad:
After further discussion, this is good to go in. Even in current VPR it's good to test for explicit testing of manually setting the flow options.

PR #3596, opened 2026-06-05
Zoom-based decluttering for channel nodes, switch box edges and connection box edges

kevinxing-a12:
#### Description

Implement zoom-based decluttering for channel nodes, switch box edges and connection box edges. Furthermore, if a node is selected when it's visible, and the user zooms out, the node still remains visible while all the others are shut off. Invisible nodes cannot be selected. In the first commit there was a new UI button that toggles this decluttering mode, but it's temporarily deleted to prevent potential conflict with QT. Therefore, the decluttering mode is always on now.

#### Motivation and Context

This PR improves users' visual experience when they use the VPR GUI, and also makes rendering speed faster when decluttering is in effect.

I have run the VPR flow with the GUI on several different architectures.

#### Types of changes

#### Checklist:


vpr/src/draw/draw.cpp
```diff
@@ -84,6 +84,8 @@ static float get_router_expansion_cost(const t_rr_node_route_inf& node_inf,
                                        e_draw_router_expansion_cost draw_router_expansion_cost);
 static void draw_router_expansion_costs(ezgl::renderer* g);
 
+static double get_zoom_scale(ezgl::renderer* g);
```
AmirhosseinPoolad: Add doxygen comments for what "zoom scale" means and what this function returns.

vpr/src/draw/draw.cpp
```diff
@@ -177,6 +179,13 @@ void init_graphics_state(bool show_graphics_val,
 }
 
 #ifndef NO_GRAPHICS
+// Helper for calculating draw_state->zoom_scale
```
AmirhosseinPoolad: Why do doxygen above declarations.

See https://docs.verilogtorouting.org/en/latest/dev/coding_style/

vpr/src/draw/draw_types.h
```diff
@@ -183,6 +183,18 @@ struct t_draw_state {
...
+
+    ///@brief The ratio between pixels and world units spanning the screen width.
+    double zoom_scale = 1.0;
+
+    ///@brief Currently always true, but will be wired to a UI button later on
+    bool enable_decluttering = true;
+
+    ///@brief Whether to declutter RR nodes
+    bool declutter_rr = false;
+
```
AmirhosseinPoolad: I suggest you explain these two variables in more details.

'enable_decluttering':  controls whether level of detail/decluttering is turned on (also explain what that means e.g. stops drawing wires/nets when zoomed out) + add a TODO for wiring it to the UI.

'declutter_rr': This is dynamically set if decluttering is enabled based on the current zoom level. tbh I would prefer not having this here (Less global state) and calculate this in draw_rr.

vpr/src/draw/search_bar.cpp
```diff
@@ -203,6 +203,7 @@ void search_and_highlight(GtkWidget* /*widget*/, ezgl::application* app) {
 
 bool highlight_rr_nodes(RRNodeId hit_node) {
     t_draw_state* draw_state = get_draw_state_vars();
+    const RRGraphView& rr_graph = g_vpr_ctx.device().rr_graph;
 
     //TODO: fixed sized char array may cause overflow.
     char message[250] = "";
```
AmirhosseinPoolad: Can you please refactor this to use std::string? Would also simplify the code a lot.

vpr/src/draw/draw_types.h
```diff
@@ -183,6 +183,18 @@ struct t_draw_state {
     ///@brief What to draw on the screen (ROUTING, PLACEMENT, NO_PICTURE)
     e_pic_type pic_on_screen = e_pic_type::NO_PICTURE;
 
+    ///@brief 1 pixel per channel node is the threshold at which the nodes blend into a solid color. 1.5 is a more relaxed bar.
+    static constexpr double min_pixel_per_node = 1.5;
+
+    ///@brief The ratio between pixels and world units spanning the screen width.
+    double zoom_scale = 1.0;
+
```
AmirhosseinPoolad: Do these need to be part of draw state? They're only used in draw_main_canvas, might as well just calculate everything there and avoid having another global variable.

vpr/src/draw/draw.cpp
```diff
@@ -177,6 +179,13 @@ void init_graphics_state(bool show_graphics_val,
...
+static double get_zoom_scale(ezgl::renderer* g) {
+    double world_width = g->get_visible_world().width();
+    double screen_width = g->get_visible_screen().width();
+    return screen_width / world_width;
```
AmirhosseinPoolad, line 194: Since this doesn't care about channel width, add a comment explaining why this works with all combinations of channel width and device size.

AmirhosseinPoolad: I think what's going on is that if a tile is 20 units wide, the channel would be also ~20 units wide so whatever number you would get here is approximately two times the number of pixels per RR wire.

This is very dependent on how we set up the coordinates. If in the future we do that in a different way the de-cluttering would break. This might be fine since we don't really touch that code much, if at all. @AlexandreSinger What do you think?

vpr/src/draw/search_bar.cpp
```diff
@@ -214,6 +215,19 @@ bool highlight_rr_nodes(RRNodeId hit_node) {
...
+        // Return true here to indicate that we processed the click, even though we didn't highlight anything
+        return true;
+    }
+
+    // The user cannot select channel nodes if they are turned off or decluttered
+    if ((!draw_state->draw_channel_nodes || draw_state->declutter_rr) && (rr_graph.node_type(hit_node) == e_rr_type::CHANX || rr_graph.node_type(hit_node) == e_rr_type::CHANY)) {
+        application.refresh_drawing();
+        return true;
+    }
+
```
AmirhosseinPoolad: Can you please check if it's possible to move these checks to highlight_rr_nodes(float x, float y) in draw_rr.cpp?

The main benefit is avoiding looping through the entire RR Graph when clicking anywhere.

vpr/src/draw/search_bar.cpp
AmirhosseinPoolad, line 237: Cool, we have hit_nodes in there already. You can use this in a following PR to remove the 'draw_rr_node' array completely.

AmirhosseinPoolad (commented):
Thank Kevin! Had some comments about commenting and moving some code around.

vpr/src/draw/draw_types.h
kevinxing-a12: I think it's reasonable to remove `min_pixel_per_node` and `zoom_scale` from `draw_state`, but I still want to keep `declutter_rr` there because otherwise I would need to declare another `zoom_scale` and call `get_zoom_scale()` in `draw_rr.cpp`.

vpr/src/draw/draw.cpp
```diff
@@ -177,6 +179,13 @@ void init_graphics_state(bool show_graphics_val,
 }
 
 #ifndef NO_GRAPHICS
+// Helper for calculating draw_state->zoom_scale
+static double get_zoom_scale(ezgl::renderer* g) {
```
vaughnbetz: I think you could give this function a more meaningful name:
e.g.
draw_pixels_per_world_unit ();

vpr/src/draw/draw_rr.cpp
```diff
@@ -78,8 +80,8 @@ void draw_rr(ezgl::renderer* g) {
         }
 
         if (!node_highlighted) {
-            // Draw channel nodes if enabled
-            if ((node_type == e_rr_type::CHANX || node_type == e_rr_type::CHANY) && !draw_state->draw_channel_nodes) {
+            // Draw channel nodes if enabled and not clutterd
```
vaughnbetz: cluttered

vpr/src/draw/draw_types.h
```diff
@@ -183,6 +183,18 @@ struct t_draw_state {
     ///@brief What to draw on the screen (ROUTING, PLACEMENT, NO_PICTURE)
     e_pic_type pic_on_screen = e_pic_type::NO_PICTURE;
 
+    ///@brief 1 pixel per channel node is the threshold at which the nodes blend into a solid color. 1.5 is a more relaxed bar.
+    static constexpr double min_pixel_per_node = 1.5;
```
vaughnbetz: I'd change this to "min_pixel_per_rr_node" as node is a pretty generic term in  a CAD program.

vpr/src/draw/draw_types.h
```diff
@@ -183,6 +183,18 @@ struct t_draw_state {
...
+
+    ///@brief The ratio between pixels and world units spanning the screen width.
+    double zoom_scale = 1.0;
```
vaughnbetz: I'd change the name of this variable. pixel_per_world_unit would be a better (more informative) name.

vpr/src/draw/draw_types.h
```diff
@@ -183,6 +183,18 @@ struct t_draw_state {
...
+
+    ///@brief Currently always true, but will be wired to a UI button later on
```
vaughnbetz: Comment should explain what this does.
 
'If true, use level of detail modeling to deside when to not draw various chip elements (rr_nodes) etc. to avoid excessive clutter when zoomed out.'
You can put the current comment after that, as it is less important and more temporary.

vpr/src/draw/draw_types.h
```diff
@@ -183,6 +183,18 @@ struct t_draw_state {
...
+    bool enable_decluttering = true;
+
+    ///@brief Whether to declutter RR nodes
```
vaughnbetz: Explain what this does in the comment. 
'When true, do not draw RR nodes to avoid excessive clutter at the current zoom level'

vpr/src/draw/search_bar.cpp
```diff
@@ -214,6 +215,19 @@ bool highlight_rr_nodes(RRNodeId hit_node) {
...
     }
 
+    // The user cannot select any nodes if RR drawing is turned off
+    if (!draw_state->show_rr) {
+        application.refresh_drawing();
+        // Return true here to indicate that we processed the click, even though we didn't highlight anything
+        return true;
+    }
+
+    // The user cannot select channel nodes if they are turned off or decluttered
```
vaughnbetz: Users can also click on the input and output pins of blocks (as these also turn into rr_nodes) when the rr_nodes are visible. Check this code also turns off selecting those IPIN and OPIN rr_nodes when they are not drawn. 
I think the CHANX and CHANY tests may be preventing that (if so they should be removed); also the comment talks only about channel nodes so I think it is not quite accurate (it should be talking about rr_nodes I think).

vaughnbetz (changes requested):
Looks good overall. Some things to address in my comments though.

vpr/src/draw/search_bar.cpp
kevinxing-a12: These pins are not involved in decluttering (since they are too small anyway), but I just ran the visualizer and found that they can still be selected when not drawn. Is this in line with the original intention, or should I make a change?

vaughnbetz:
No, they shouldn't be selectable when not visible.
Also, we shouldn't draw them when things will be too cluttered (we can turn them off alongside the rr channel nodes) as they will still slow down rendering speed, even if they don't clutter the screen as much because they are too small to be seen (still wastes cpu)

vpr/src/draw/draw.cpp
kevinxing-a12: Do you mind explaining why it's approximately two times?
Also, I thought the channel width corresponds to max_chan_ptc_num? There might be gaps in my understanding.
> for (size_t i = 0; i < grid.width() - 1; i++) {
        draw_coords->tile_x[i] = i * draw_coords->get_tile_width() + j;
        j += max_chany_ptc_nums[i] + 2; 
    }

kevinxing-a12:
Understood!

vpr/src/draw/draw.cpp
vaughnbetz: Channel width should correspond to the maximum ptc_num.  I'm not too worried about the decluttering breaking if we change our coordinate system; lots of stuff will break if we do that, so we'd have to go through the code anyway.

vpr/src/draw/search_bar.cpp
kevinxing-a12: I have tried something similar previously (not checking what the mouse actually clicks on if rr is off), but there are possible defects that are relevant to the code snippet below:
```
void act_on_mouse_press(ezgl::application* app, GdkEventButton* event, double x, double y) {
            ......
            ......
            //Check if we hit an rr node
            // Note that we check this before checking for a block, since pins and routing may appear overtop of a multi-width/height block
            if (highlight_rr_nodes(x, y)) {
                return; //Selected an rr node
            }

            highlight_blocks(x, y);
        }
    }
}
```
Under the case where rr is off, if we return true from highlight_rr_nodes(x, y), the user can never select a block. But if we return false, we will always call highlight_blocks(x, y) even when the mouse actually clicked on an RR node (unnecessary processing). Since the current code for selection feels instant already, I might just keep this part as such. But please let me know if you have other ideas!

vpr/src/draw/draw.cpp
AmirhosseinPoolad: >Do you mind explaining why it's approximately two times?

I see now, you're simply checking if 1 world unit is more than 1.5 screen pixels. Yeah that works.

vpr/src/draw/draw_types.h
AmirhosseinPoolad: You're basically kind of doing that anyway. draw_rr() is called immediately after you calculate declutter_rr in draw_main_canvas() and declutter_rr is only ever used in draw_rr() (And I don't think anyone else should use it! Encapsulation is good). Might as well just have min_pixel_per_node, zoom_scale, and declutter_rr as local variables in draw_rr()

vpr/src/draw/search_bar.cpp
AmirhosseinPoolad: I think that's fine. Let's say the user clicks on some empty space between blocks while RR nodes are not being shown.

Old code:

1. Go through all RR Nodes and highlight the selected node (but since it's not being drawn then nothing happens)
2. Don't check the blocks since we've already highlighted a node
3. Nothing happens from a user's PoV

If you change it to return false:

1. We don't draw rr nodes for whatever reason -> Don't go through RR nodes, return false
2. Check if a block was clicked, go through at most 90000 blocks (vs 100 million RR nodes) and see that nothing was clicked on
3. Nothing happens from a user's PoV

I might have missed something though, please correct me in that case.

vpr/src/draw/draw_types.h
kevinxing-a12: declutter_rr is also used in search_bar.cpp in the function highlight_rr_node to prevent the user from selecting rr nodes when they are decluttered. Even if I move the check to highlight_rr_node(x, y) (which is in the same file as draw_rr()), declutter_rr cannot be used there if it's only a local variable in draw_rr().

vpr/src/draw/search_bar.cpp
kevinxing-a12: This makes sense! I will change the code.

vpr/src/draw/draw_types.h
AmirhosseinPoolad: That's fair.

vpr/src/draw/draw_rr.cpp
```diff
@@ -274,21 +278,17 @@ void draw_rr_edges(RRNodeId inode, ezgl::renderer* g) {
...
+        } else if ((edge_type == e_edge_type::PIN_TO_OPIN || edge_type == e_edge_type::PIN_TO_IPIN) && !draw_state->draw_intra_cluster_edges) {
             draw_edge = false;
-        }
-
-        if (edge_type == e_edge_type::CHAN_TO_CHAN && !draw_state->draw_switch_box_edges) {
+        } else if ((edge_type == e_edge_type::OPIN_TO_CHAN || edge_type == e_edge_type::CHAN_TO_IPIN) && !draw_state->draw_connection_box_edges) {
+            draw_edge = false;
+        } else if (edge_type == e_edge_type::CHAN_TO_CHAN && !draw_state->draw_switch_box_edges) {
             draw_edge = false;
         }
```
AmirhosseinPoolad, line 297: Can you add some comments explaining what each case means and why we don't draw edges in that case?

vpr/src/draw/draw_searchbar.cpp
```diff
@@ -129,7 +129,7 @@ void draw_highlight_blocks_color(t_logical_block_type_ptr type,
 /* If an rr_node has been clicked on, it will be highlighted in MAGENTA.
  * If so, and toggle nets is selected, highlight the whole net in that colour.
  */
-void highlight_nets(char* message, RRNodeId hit_node) {
+void highlight_nets(std::string& message, RRNodeId hit_node) {
```
AmirhosseinPoolad, line 132: Thanks for doing this refactoring!

AmirhosseinPoolad (approved):
I have one more comment then LGTM! Thanks again for this PR Kevin.

AmirhosseinPoolad (approved):
LGTM

vaughnbetz (approved):
LGTM

vpr/src/draw/search_bar.cpp
vaughnbetz: You should also check that blocks are properly de-highlighted if the user clicks on the background (blank canvas region).

kevinxing-a12: I tested it just now, and yes they are!

PR #3600, opened 2026-06-08
[CI] Updated Weekly Test Golden

AlexandreSinger:
Recent changes to how tasks are parsed caused one golden result to be incorrect and need to be regenerated.

The PR changing task parsing can be found here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3583

It looks like one of the ISPD circuits was not being parsed correctly. My PR appears to have fixed the parsing, so now the golden results just need to be regenerated:
image


AmirhosseinPoolad:
LGTM.

I'm not sure if I'm misreading but it seems like we've gained a bit of runtime since the last golden results. Probably the SA initial cost changes. Cool!

AlexandreSinger:
> LGTM.
> 
> I'm not sure if I'm misreading but it seems like we've gained a bit of runtime since the last golden results. Probably the SA initial cost changes. Cool!

These golden results are updated very rarely; I think they were just old, but good to keep an eye on!

PR #3601, opened 2026-06-08
GSB V2

amin1377:
This PR adds support for GSB V2. In this version of GSB, the connections described in the switch block pattern CSV files are interpreted differently.

For sink connections, it assumes that the first part of the routing wires, `x_low`/`y_low` for `INC` and `x_high`/`y_high` for `DEC`, belongs to the GSB currently being considered.

For source connections, the tap number is interpreted as the index of the blocks passed by the wire.

### DIFF
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -10,7 +10,6 @@
 #include "vtr_log.h"
 #include "vtr_path.h"
 #include "vtr_util.h"
-#include <filesystem>
 #include <string>
 
 using argparse::ConvertedValue;
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -1514,9 +1514,12 @@ struct ParseGsbVersion {
         ConvertedValue<std::string> conv_value;
         if (val == e_gsb_version::GSB_V1)
             conv_value.set_value("1");
-        else {
-            VTR_ASSERT(val == e_gsb_version::GSB_V2);
+        else if (val == e_gsb_version::GSB_V2)
             conv_value.set_value("2");
+        else {
+            std::stringstream msg;
+            msg << "Unrecognized e_gsb_version value: " << static_cast<int>(val);
+            conv_value.set_error(msg.str());
         }
         return conv_value;
     }
--- a/vpr/src/base/read_options.cpp
+++ b/vpr/src/base/read_options.cpp
@@ -10,6 +10,7 @@
 #include "vtr_log.h"
 #include "vtr_path.h"
 #include "vtr_util.h"
+#include <filesystem>
 #include <string>
 
 using argparse::ConvertedValue;
@@ -1492,6 +1493,37 @@ struct ParsePostSynthNetlistUnconnOutputHandling {
 
     std::vector<std::string> default_choices() {
         return {"unconnected", "nets"};
+    }
+};
+
+struct ParseGsbVersion {
+    ConvertedValue<e_gsb_version> from_str(const std::string& str) {
+        ConvertedValue<e_gsb_version> conv_value;
+        if (str == "1")
+            conv_value.set_value(e_gsb_version::GSB_V1);
+        else if (str == "2")
+            conv_value.set_value(e_gsb_version::GSB_V2);
+        else {
+            std::stringstream msg;
+            msg << "Invalid conversion from '" << str << "' to e_gsb_version (expected one of: " << argparse::join(default_choices(), ", ") << ")";
+            conv_value.set_error(msg.str());
+        }
+        return conv_value;
+    }
+
+    ConvertedValue<std::string> to_str(e_gsb_version val) {
+        ConvertedValue<std::string> conv_value;
+        if (val == e_gsb_version::GSB_V1)
+            conv_value.set_value("1");
+        else {
+            VTR_ASSERT(val == e_gsb_version::GSB_V2);
+            conv_value.set_value("2");
+        }
+        return conv_value;
+    }
+
+    std::vector<std::string> default_choices() {
+        return {"1", "2"};
     }
 };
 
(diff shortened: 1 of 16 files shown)


vpr/src/base/read_options.cpp
```diff
@@ -1495,6 +1495,37 @@ struct ParsePostSynthNetlistUnconnOutputHandling {
...
+        }
+        return conv_value;
+    }
+
+    ConvertedValue<std::string> to_str(e_gsb_version val) {
+        ConvertedValue<std::string> conv_value;
+        if (val == e_gsb_version::GSB_V1)
+            conv_value.set_value("1");
+        else {
+            VTR_ASSERT(val == e_gsb_version::GSB_V2);
```
AlexandreSinger: Nit: I think it would be cleaner to use `conv_value.set_error("...")` instead of an assert when an unrecognized GSB version is passed in. This is minor though, so if you are just matching the rest of the file I understand. I do want to move away from using ASSERTs to guard code like this.

vpr/src/base/read_options.cpp
```diff
@@ -3457,6 +3488,11 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
         .default_value("")
         .show_in(argparse::ShowIn::HELP_ONLY);
 
+    crr_grp.add_argument<e_gsb_version, ParseGsbVersion>(args.gsb_version, "--gsb_version")
+        .help("Specifies which GSB version should be used for CRR switch block templates. Valid values are 1 or 2.")
```
AlexandreSinger, line 3495: I think this description needs to be improved. It should clearly write the differences between version 1 and 2. A random user would not know the difference without tracing the code.

This is also up to you, I do not know the full context of this, but this argument could also be added to the docs. Right now this only shows up in the help menu.

vpr/src/base/vpr_types.h
```diff
@@ -1409,13 +1409,19 @@ struct t_analysis_opts {
     bool skip_sync_clustering_and_routing_results;
 };
 
+enum class e_gsb_version {
+    GSB_V1 = 1,
+    GSB_V2 = 2,
```
AlexandreSinger: You give both of these enums default values, but they are never used. You could just let compiler use whatever default values it wants.

I do not think each member needs a comment, but I recommend adding a doxygen comment above this enum class.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -468,8 +470,13 @@ void CRRConnectionBuilder::calculate_segment_coordinates(const SegmentInfo& info
...
             case e_sw_template_dir::LEFT:
-                x_high = x + (seg_length - tap);
-                x_low = x - (tap - 1);
+                if (gsb_version_ == e_gsb_version::GSB_V2) {
+                    x_high = x + (seg_length - tap - 1);
+                    x_low = x - tap;
+                } else {
+                    x_high = x + (seg_length - tap);
+                    x_low = x - (tap - 1);
+                }
```
AlexandreSinger: Just a longevity thing. If future versions are ever going to happen, I assume that they will follow V2 (not V1). It may be a wise idea to make this:
```cpp
if (gsb_version_ == V1) {
  // Do the old way.
} else {
  // Do the new way.
}
```

Or if this is not clear, we should error out if an unrecognized version is passed in. These issues can become nasty when doing versioning, so it is better to be strict in my opinion.

AlexandreSinger: Small aside for you to think on, it looks like the only difference here is that `tap = tap - 1`. Why not at the start of the function updating the tap variable accordingly? Something like:
```
int tap_idx = gsp_version_ == V1 ? tap : tap  - 1;
```

May just be much cleaner to support into the long term.

AlexandreSinger: Same comment for the rest of this file.

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_graph_view_util.cpp
```diff
@@ -115,11 +114,8 @@ std::vector<RRNodeId> find_rr_graph_grid_nodes(const RRGraphView& rr_graph,
 
     for (int pin = 0; pin < device_grid.get_physical_type(tile_loc)->num_pins; ++pin) {
         /* Skip those pins have been ignored during rr_graph build-up */
-        if (true == device_grid.get_physical_type(tile_loc)->is_ignored_pin[pin]) {
-            /* If specified, force to include all the clock pins */
-            if (!include_clock || std::find(device_grid.get_physical_type(tile_loc)->get_clock_pins_indices().begin(), device_grid.get_physical_type(tile_loc)->get_clock_pins_indices().end(), pin) == device_grid.get_physical_type(tile_loc)->get_clock_pins_indices().end()) {
-                continue;
-            }
```
AlexandreSinger, line 122: Why was this removed in this PR? It looks like include_clock was always false, but I am not sure of the context of this original code. It may be wise to raise this as a separate PR.

AlexandreSinger: But this is up to you since this is a tileable RR graph function call.

AlexandreSinger (commented):
Looks good @amin1377 , just some minor comments. My main concerns are on longevity; versioning can be tricky if not done cleanly. I recommend documenting this carefully and isolating the differences between versions the best that you can.

There are also no testcases, but we can wait on that as you continue to develop this feature.

vpr/src/base/read_options.cpp
amin1377: Once the feature is fully in place, I’ll update the document to explain the differences between the versions in detail.

vpr/src/route/rr_graph_generation/tileable_rr_graph/rr_graph_view_util.cpp
amin1377: Good point. I only checked the VTR codebase and saw that it is always set to `false`, which is why I removed it. However, I just checked the OpenFPGA code and realized there are some function calls where this parameter is set to `true`.

AlexandreSinger (approved):
The offset calculations are much cleaner now! Thank you. LGTM.

PR #3602, opened 2026-06-08
Make scatter-gather warnings more verbose

AmirhosseinPoolad:
The sg code prints warnings when there is no fan-in or fan-out candidates. We actually expect this to happen on device edges and we were printing a thousand of these warnings, so this PR makes them happen only on route_verbosity>1 (default is 1)

### DIFF
--- a/vpr/src/route/rr_graph_generation/build_scatter_gathers.h
+++ b/vpr/src/route/rr_graph_generation/build_scatter_gathers.h
@@ -57,7 +57,10 @@ struct t_bottleneck_link {
  * @param nodes_per_chan Channel width data.
  * @param rng Random number generator used to shuffle wire candidates.
  * @param interdie_3d_links Output: matrix storing inter-die (3D) bottleneck links.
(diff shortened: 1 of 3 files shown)


soheilshahrouz:
In another PR, I only print the first 10 SG-related warnings. If you want to keep this PR open, I can revert my code so it doesn't confilict with yours.

vpr/src/route/rr_graph_generation/build_scatter_gathers.h
```diff
@@ -57,6 +57,7 @@ struct t_bottleneck_link {
  * @param nodes_per_chan Channel width data.
  * @param rng Random number generator used to shuffle wire candidates.
  * @param interdie_3d_links Output: matrix storing inter-die (3D) bottleneck links.
+ * @param route_verbosity Controls verbose routing diagnostics.
```
soheilshahrouz: Explain what warnings are printed at each verbosity level. This sets a precedent, forcing others to keep this comment up to date when they add new warnings.

PR #3603, opened 2026-06-09
Zoom-based decluttering for block internals

kevinxing-a12:
#### Description

I implemented zoom-based decluttering for block internals and they can now be drawn in different levels of details. I also fixed the text display issue related to them. During the implementation I realized that get_pixels_per_world_unit() cannot be a static function in draw.cpp only, because it was going to be needed by other files, and as a result I reorganized that code part.

#### Motivation and Context

Make the block internals drawn in a more smooth and user-friendly way.

I ran the GUI using several different architectures, and the level of details seemed to be quite decent.

#### Types of changes

#### Checklist:


kevinxing-a12:
Demo:
https://github.com/user-attachments/assets/48654216-a3ad-4fe6-9aed-bdab06886b1b

vpr/src/draw/draw_rr.cpp
```diff
@@ -49,6 +49,18 @@ void draw_rr(ezgl::renderer* g) {
         return;
     }
 
+    // The ratio between pixels and world units spanning the screen width. Used to determine when decluttering should occur.
+    double pixel_per_world_unit = get_pixels_per_world_unit(g);
+    if (draw_state->enable_decluttering) {
+        // 1 pixel per channel node is the threshold at which the nodes are blended into a solid color. 1.5 is a more relaxed bar.
+        constexpr double min_pixel_per_chan_node = 1.5;
```
AlexandreSinger: I recommend pulling this up as a static constexpr at the top of the file for easy access with clear documentation of what it is used for. This seems like something we may want to revisit tuning in the future.

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -400,6 +400,18 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
...
 
+    // The ratio between pixels and world units spanning the screen width.
+    double pixels_per_world_unit = get_pixels_per_world_unit(g);
+    // Get the number of pixels spanned by the width and height of the bounding box, respectively.
+    double abs_bbox_width_pixels = abs_bbox.width() * pixels_per_world_unit;
+    double abs_bbox_height_pixels = abs_bbox.height() * pixels_per_world_unit;
+    // If the area of the block's bounding box is less than 800 pixels, don't draw the block.
+    // This is a method of decluttering, and 800 was chosen as a threshold through experimentation.
+    // Note: since this function handles sub-blocks only, the drawing of top-level blocks will not be affected.
+    if (abs_bbox_width_pixels * abs_bbox_height_pixels < 800) {
```
AlexandreSinger: Magic number, I recommend making this 800 a variable and similar to my other comment, this can be made as a static constexpr and put at the top of this file (or in a more organized place).

AlexandreSinger (commented):
Thanks @kevinxing-a12 , the demo looks really good!

I had a couple of small comments.

AmirhosseinPoolad (approved):
Looks good. Try running a device with channel width = 2 or 4 to see how well the decluttering works in edge cases. I don't see any reasons why it shouldn't work but it's good to know any possible limitations.

kevinxing-a12:
Yes, and no drawing errors found!

AmirhosseinPoolad (approved):
LGTM

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -468,12 +483,14 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
...
         if (draw_state->draw_block_text) {
+            // draw_coords->get_tile_height() * FRACTION_TEXT_PADDING represents the vertical padding from the block's ceiling.
+            // We double this value when calculating the height of the text box, because the same amount of padding is needed to reach the block's floor from the text.
             g->draw_text(
                 ezgl::point2d(abs_bbox.center_x(),
                               abs_bbox.top() - draw_coords->get_tile_height() * FRACTION_TEXT_PADDING),
                 pb_display_text.c_str(),
                 abs_bbox.width(),
-                abs_bbox.height());
+                draw_coords->get_tile_height() * FRACTION_TEXT_PADDING * 2);
```
vaughnbetz, line 493: I don't really understand this test.  abs_bbox.height should be the height of the block internal we are drawing. We shouldn't draw if we don't fit within that height (current test).  The new test seems to check against a padding dimension that doesn't seem directly relevant.  Maybe our height test is wrong because fitting in the block internal box isn't enough; e.g. we also need to know if another block internal drawn inside this box will cut off the text.  But if that's the case I still think you need a more robust test and some additional commenting.

vaughnbetz (changes requested):
See my comment on the vertical text clipping. Otherwise looks good.

vpr/src/draw/intra_logic_block.cpp
kevinxing-a12: The code here corresponds to the case where text is drawn at the top of the block:

The argument in the same function call that specifies the location of the text is `abs_bbox.top() - draw_coords->get_tile_height() * FRACTION_TEXT_PADDING`. It basically sets the text right on the top, and then does some padding to bring it inside the block. The same amount of padding is needed to reach the bottom of the text, and that is where the floor of the text bounding box should be.

vaughnbetz: OK, thanks.

vaughnbetz (approved):
LGTM

kevinxing-a12:
Is this good to be merged?

vaughnbetz:
Yup, done!

PR #3605, opened 2026-06-09
[AP] Fixed Unrelated Clustering Bug

AlexandreSinger:
I found a bug in the unrelated clustering code where the width and height of the device were confused with the number of layers.

Part of this code pre-dates bringing AP to 3D, so the 0th dim was width and the 1st dim was height. I noticed a strange access out of bounds and found that this was no longer the case (num layers is the 0th dim). I fixed the code.

This code path is a very rare fall-back, but this bug certainly made that path much worse.


haydar-c (approved):
Thanks @AlexandreSinger, LGTM.

PR #3606, opened 2026-06-09
Interposer zero place cost fix

AmirhosseinPoolad:
This PR does two things:

1- Adds logic that handles small numerical errors for when interposer cost becomes zero (Possible when you can fit the circuit on a single die)
2- Adds AP flow regression testing for AP flow.


AmirhosseinPoolad:
AP was previously failing the hecate strong test. This PR fixes that.

vpr/src/place/placer.cpp
```diff
@@ -285,6 +285,7 @@ int Placer::check_placement_costs_() {
     int error = 0;
 
     constexpr double MIN_EXPECTED_CONG_COST = 1.e-6;
+    constexpr double MIN_EXPECTED_INT_COST = 1.e-6;
```
vaughnbetz: I'd expand to interposer (INT is vague).
A comment also might be good

vpr/src/place/placer.cpp
```diff
@@ -316,7 +317,12 @@ int Placer::check_placement_costs_() {
         error++;
     }
 
-    if (!vtr::isclose(cost_terms_check.interposer_cost, costs_.interposer_cost, PL_INCREMENTAL_COST_TOLERANCE, 0.)) {
+    // Similar logic to congestion cost. In case of a small circuit on a large device,
+    // it's possible to have zero interposer cost so we need to handle
+    // tiny numerical differences.
+    if (!((std::fabs(cost_terms_check.interposer_cost) < MIN_EXPECTED_INT_COST
+           && std::fabs(costs_.interposer_cost) < MIN_EXPECTED_INT_COST)
+          || vtr::isclose(cost_terms_check.interposer_cost, costs_.interposer_cost, PL_INCREMENTAL_COST_TOLERANCE, 0.))) {
```
vaughnbetz: couldn't you just use MIN_EXPECTED_INT_COST as the absolute tolerance (Replace the final 0) in the isClose call, and get rid of the if statements comparing interposer_cost an MIN_EXPECTED_INT_COST?

vpr/src/place/placer.cpp
soheilshahrouz: Make sure this value is small enough. I guess the minimum non-zero interposer crossing cost is 1/H or 1/W. You can also add a comment about this.

soheilshahrouz (commented):
Thanks Amir!

vpr/src/place/placer.cpp
AmirhosseinPoolad: The logic is slightly different. isclose uses the maximum of relative tolerance and absolute tolerance to compare, here we say that if the cost is less than this min value treat it as a zero. I'm pretty sure it would work fine but in cases where 1% (current relative tolerance) of a cost is smaller than 10^-6 it would result in a larger than intended tolerance.

I've implemented this change, but I'll revert it if CI fails or if you think it's unsafe to do so.

AlexandreSinger (approved):
LGTM

vaughnbetz (approved):
LGTM

PR #3607, opened 2026-06-09
Add GSB version to device context

amin1377:
This PR adds a field to the VPR context to store the GSB version. This is mainly required by downstream repositories, such as OpenFPGA, so they can access the GSB version used to create the RR graph.


vpr/src/base/setup_vpr.cpp
```diff
@@ -776,6 +776,9 @@ static void setup_crr_opts(const t_options& Options, t_crr_opts& crr_opts) {
     crr_opts.remove_dangling_nodes = Options.remove_dangling_nodes;
     crr_opts.sb_count_dir = Options.sb_count_dir;
     crr_opts.gsb_version = Options.gsb_version;
+
+    DeviceContext& device_ctx = g_vpr_ctx.mutable_device();
+    device_ctx.gsb_version = Options.gsb_version;
```
AlexandreSinger: I see that this file does modify the device context in the SetupVpr function. This is something I am hoping to pull out in the future. You can leave this here, but could you pass in the device context by mutable reference into the setup_crr_opts function? Or update the device context in the SetupVpr function instead?

I want to isolate where the device context gets modified into a single location if possible. Knowing when the device context gets populated helps a lot.

AlexandreSinger (commented):
LGTM, minor comment.

I am assuming that more features are coming in which will use this variable.

AmirhosseinPoolad:
I don't want to block this, but since this variable is added to all of VPR, it would be good if you could add a "NOT_CRR" value to e_gsb_version. That would be the default value for all architectures and CRR architectures would then change that to V1 or V2.

amin1377:
> I don't want to block this, but since this variable is added to all of VPR, it would be good if you could add a "NOT_CRR" value to e_gsb_version. That would be the default value for all architectures and CRR architectures would then change that to V1 or V2.

I don't have a strong opinion on this, and if you think it would be beneficial, I can certainly make the change.
My only question is: since this variable is only used by the tileable RR graph generator, there doesn't seem to be any ambiguity in its usage. Why do you think having a `NOT_CRR` value is necessary?

AmirhosseinPoolad:
> since this variable is only used by the tileable RR graph generator, there doesn't seem to be any ambiguity in its usage. Why do you think having a NOT_CRR value is necessary?

Mostly for future-proofing. It's also the principle of not having invalid data in these global data structures, even when it's clear that anything you read would be invalid.

AmirhosseinPoolad (approved):

PR #3608, opened 2026-06-09
[Test] Locked a CI test to traditional flow

AlexandreSinger:
I found that one of the CI tests was checking that the post-implementation netlist matched exactly a golden results file from a prior test. So long as the packing is exactly the same the golden results would not change. Since AP uses flat placement to inform packing, it would make this test very flakey so we are keeping it on the traditional flow.

The goal of this test is to just ensure that the post-implementation netlist is being generated properly; so making this change will not modify the original intent of this test.


AlexandreSinger:
@AmirhosseinPoolad @amin1377 This is blocking getting AP into default. Can one of you review this when you have a moment?

amin1377:
@AlexandreSinger I’m a bit confused here. I assume AP produces the same results each time it runs, and if that’s the case, shouldn’t updating the golden results make the test pass?

AlexandreSinger:
> I assume AP produces the same results each time it runs, and if that’s the case, shouldn’t updating the golden results make the test pass?

You are correct; however, this golden result is really not clean or easy to update. This test is also not trying to test anything in the AP flow.

I am worried about updating this to the AP flow since the packing is not as "stable": small changes to the AP algorithm will change the clustering; which is not the same for the original pack flow.

amin1377:
This was also the case with the traditional flow; updating packing would require updating the golden results for this test.

I can think of two possible solutions:

1. Update the golden results. Since AP should be stable now, this should hopefully not cause too many issues.
2. Store the `.net` file, read it back, and generate the post-implementation netlist based on that.

amin1377:
If you don’t like either of the solutions above, that’s also fine with me. I honestly don’t have a strong opinion on this issue.

amin1377 (approved):

AlexandreSinger:
> 2\. Store the `.net` file, read it back, and generate the post-implementation netlist based on that.

I like this idea. This will be the fix. Let me not block AP with this. I will raise an issue and merge, but I will fix the issue as soon as I can.

AlexandreSinger:
Issue raised: https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/3623

PR #3609, opened 2026-06-09
Clarified quickstart package installation

vaughnbetz:
#### Description
Made the description of how to install packages on different Linux systems a bit more clear, and moved the clang-format stuff later so it doesn't distract those who don't need it. Also documented the --dev option and who should use it for the package set up scripts.

### DIFF
--- a/doc/src/quickstart/index.rst
+++ b/doc/src/quickstart/index.rst
@@ -24,10 +24,33 @@ If you cloned the repository, you will need to set up the git submodules (if you
     > git submodule init
     > git submodule update
 
-VTR requires several system and Python packages to build and run the flow. Ubuntu users can install the required system packages using the provided script or the command below. This setup works on Ubuntu 18.04, 20.04, 22.04, and 24.04, but note that some packages (such as ``clang-format-18``) are only available by default on Ubuntu 24.04. On older versions, this package will not be installed unless you manually add the appropriate LLVM APT repository.
+VTR requires several system and Python packages to build and run the flow.
+On Ubuntu and Debian systems, install the packages required by VTR using the command below. 
 
-To install ``clang-format-18`` on older Ubuntu versions (e.g., 20.04 or 22.04), you must add the LLVM repository manually. Note that this tool is only required if you want to run ``make format`` to automatically fix formatting issues in the code. It is not necessary for building or running VPR.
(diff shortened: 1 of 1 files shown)


vaughnbetz:
@ayan20985 : I think we should condition code formatting package installation on --dev too (just move the clang-format-18 lines inside the 'install_dev' block).

doc/src/quickstart/index.rst
```diff
@@ -24,40 +24,45 @@ If you cloned the repository, you will need to set up the git submodules (if you
...
 
 .. code-block:: bash
 
-    > ./install_apt_packages.sh
+    > make env                          # optional: install python virtual environment
+    > source .venv/bin/activate         # optional: activate python virtual environment
+    > pip install -r requirements.txt   # install python packages (in virtual environment if prior commands run, system wide otherwise)
 
-Fedora and RHEL users may use the following command to install the required system packages.
+.. note:: Developers who wish to modify VTR and are working on older Ubuntu versions (e.g. 20.04 or 22.04) must add  ``clang-format-18`` manually using the commands below. Note that this tool is only required if you want to run ``make format`` to automatically fix formatting issues in the code; it is not necessary for building or running VPR.
```
kevinxing-a12, line 50: Is this package also applicable to Fedora and RHEL users?

kevinxing-a12 (commented):
LGTM! One very brief question below:)

doc/src/quickstart/index.rst
AmirhosseinPoolad: Fedora is generally more up to date so it gets a newer version of clang by default. We also don't really officially support it other than "we have this one script that should probably hopefully work, good luck" since it's not tested in CI.

AlexandreSinger:
> @ayan20985 : I think we should condition code formatting package installation on --dev too (just move the clang-format-18 lines inside the 'install_dev' block).

I agree with this.

doc/src/quickstart/index.rst
AlexandreSinger: I agree with Amir here.

AlexandreSinger (approved):
LGTM, thanks @vaughnbetz . I think tieing the --dev option into the quickstart docs is good so we can start moving packages into there!

PR #3610, opened 2026-06-10
[SDC] Generated Clock -add, -master_clock, -phase, -offset support

AlexandreSinger:
Added support for the final few options for creating generated clocks.

-add and -master_clock give support for defining multiple generated clocks on the same pin and distinguishing which of multiple clocks to select as the source of the generated clock.

-phase and -offset are not commonly supported options by some open source tools; however, they are very easy to support and I think they can be useful.

The SDC documentation was updated to reflect these changes.

Strong tests were added to lock in these new options.

I am making this PR WIP as I am verifying the STA to ensure that these options are working as expected.


AlexandreSinger:
@vaughnbetz Please review when you have a moment. These should be the last options to add for create_generated_clock (there are no other options that exist as far as I am aware). I am working on getting the clock delays correct next.

Verified this locally on the strong tests. I verified that the phase and offsets are being respected in the timing reports. I also verified that the -add options was adding generated clocks correctly. The timing report is correctly enumerating all the clocks on each timing edge and the setup and hold constraints appear to be correct.

vaughnbetz (approved):
LGTM

PR #3612, opened 2026-06-10
Fix router bounding box heuristic for interposer architectures

AmirhosseinPoolad:
Interposer wires are not compatible with the router's bounding box heuristic. This would make routing bonding boxes to be at least as big as the area of the device that has interposer wires. Doing this avoids situations where the routing bounding box is smaller than the interposer wire that must be crossed, resulting in first iteration routing failures.

QoR changes for 2.5D arch:

  | Master | Feature
-- | -- | --
routed_wirelength | 1 | 0.999660872
critical_path_delay | 1 | 1.001979415
crit_path_route_time | 1 | 0.8099027725
total_heap_pops | 1 | 0.7089621547
total_swap | 1 | 1
accepted_swap | 1 | 1
placed_CPD_est | 1 | 1

Free 30% reduction of heap pops.

### DIFF
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -282,6 +282,10 @@ struct DeviceContext : public Context {
      * Indexed as `[layer][cut_idx]`, where `cut_idx` refers to the entry in
      * `DeviceGrid::get_horizontal_interposer_cuts()[layer]`.
      * The pair stores `{min_y, max_y}` of all `CHANY` rr_nodes that cross the cut.
+     * i.e. the smallest y of any rr_node that crosses a cut, and the largest y of
+     * any rr node that crosses the cut. These values will normally come from different wires;
+     * e.g. with length 10 interposer wires you'd expect to get min_y to be 9 below the cut
+     * and max_y to be 9 avove the cut.
      */
     vtr::NdMatrix<std::pair<int, int>, 2> horz_interposer_cut_bounds_;
 
@@ -291,6 +295,7 @@ struct DeviceContext : public Context {
      * Indexed as `[layer][cut_idx]`, where `cut_idx` refers to the entry in
      * `DeviceGrid::get_vertical_interposer_cuts()[layer]`.
      * The pair stores `{min_x, max_x}` of all `CHANX` rr_nodes that cross the cut.
+     * This is similar to horz_interposer_cut_bounds_, but for vertical interposer cuts and wires.
      */
     vtr::NdMatrix<std::pair<int, int>, 2> vert_interposer_cut_bounds_;
 
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -285,7 +285,7 @@ struct DeviceContext : public Context {
      * i.e. the smallest y of any rr_node that crosses a cut, and the largest y of
      * any rr node that crosses the cut. These values will normally come from different wires;
      * e.g. with length 10 interposer wires you'd expect to get min_y to be 9 below the cut
-     * and max_y to be 9 avove the cut.
+     * and max_y to be 9 above the cut.
      */
     vtr::NdMatrix<std::pair<int, int>, 2> horz_interposer_cut_bounds_;
 
--- a/vpr/src/base/vpr_context.h
+++ b/vpr/src/base/vpr_context.h
@@ -4,6 +4,7 @@
 #include <memory>
 #include <vector>
 #include <mutex>
+#include <utility>
 
 #include "flat_placement_types.h"
 #include "physical_types.h"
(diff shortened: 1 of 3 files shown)


vpr/src/base/vpr_context.h
```diff
@@ -275,6 +276,24 @@ struct DeviceContext : public Context {
      */
     vtr::NdMatrix<int, 3> vert_interposer_capacity_;
 
+    /**
+     * @brief Min/max y coordinate of all routing wires that cross each horizontal interposer cut.
+     *
+     * Indexed as `[layer][cut_idx]`, where `cut_idx` refers to the entry in
+     * `DeviceGrid::get_horizontal_interposer_cuts()[layer]`.
+     * The pair stores `{min_y, max_y}` of all `CHANY` rr_nodes that cross the cut.
```
vaughnbetz, line 284: I'd explain what this is for a little and the reasoning.
I think you're getting the smallest y of any rr_node that crosses a cut, and the largest y of any rr node that crosses the cut. These values will normally come from different wires; e.g. with length 10 interposer wires you'd expect to get min_y to be 9 below the cut and max_y to be 9 avove the cut.
This can be useful for expanding bounding boxes for routes that must cross the interposer.

vaughnbetz (changes requested):
Suggest a small commenting update; otherwise looks good.
Note that I think you're expanding bounding boxes by a lot, which I think is speeding up because often getting off an interposer wire would be outside the net bounding box and we'd fail to route it (and then iteratively expand the bounding box in later iterations).

vpr/src/route/rr_graph_generation/rr_graph.cpp
```diff
@@ -1066,6 +1074,10 @@ static void init_interposer_capacity_from_rr_graph(DeviceContext& device_ctx,
     horz_interposer_capacity.fill(0);
     vert_interposer_capacity.fill(0);
 
+    const std::pair<int, int> default_bounds = {std::numeric_limits<int>::max(), std::numeric_limits<int>::min()};
+    horz_interposer_cut_bounds.resize({num_layers, max_h_cuts}, default_bounds);
+    vert_interposer_cut_bounds.resize({num_layers, max_v_cuts}, default_bounds);
```
soheilshahrouz, line 1079: For capacity, I call `fill()` after resizing vectors. In that case, min W search calls RR graph generation multiple times, and I thought the number of crossing wires may depend on W. I might apply to this case too, but I am not sure.

vpr/src/route/route_common.cpp
```diff
@@ -806,17 +857,22 @@ t_bb load_net_route_bb(const Netlist<>& net_list,
...
     t_bb bb;
-
-    bb.xmin = std::max<int>(xmin - bb_factor, 0);
-    bb.xmax = std::min<int>(xmax + bb_factor, device_ctx.grid.width() - 1);
-    bb.ymin = std::max<int>(ymin - bb_factor, 0);
-    bb.ymax = std::min<int>(ymax + bb_factor, device_ctx.grid.height() - 1);
+    bb.xmin = xmin;
+    bb.xmax = xmax;
+    bb.ymin = ymin;
+    bb.ymax = ymax;
```
soheilshahrouz: You can remove int variables and use bb inside the loop.

AmirhosseinPoolad:
> Note that I think you're expanding bounding boxes by a lot, which I think is speeding up because often getting off an interposer wire would be outside the net bounding box and we'd fail to route it (and then iteratively expand the bounding box in later iterations).

That's exactly what happened. Previously pretty much all die-crossing nets would fail routing and then would be subsequently routed with a full-device (Not 100% sure about this) bounding box. This bounding box is probably way larger than it could be, but I think it's the safest option.

vpr/src/route/route_common.cpp
```diff
@@ -691,6 +701,47 @@ static bool classes_in_same_block(ParentBlockId blk_id, int first_class_ptc_num,
...
+            // bounding box crosses vertical_cuts[layer][cut_idx] at x = cut_x
+            if (cut_x >= original_bb.xmin && cut_x < original_bb.xmax) {
+                const std::pair<int, int>& cut_bounds = vertical_cut_bounds[layer][cut_idx];
+                bb.xmin = std::min(bb.xmin, cut_bounds.first);
+                bb.xmax = std::max(bb.xmax, cut_bounds.second);
+            }
+        }
+    }
+}
+
```
soheilshahrouz, line 744: For archs with long interposer wires and nets with high fanout, this might enlarge bb too much. I think connections that don't cross a cut should still use the old bounding box.

soheilshahrouz (commented):
Thanks Amir!

vpr/src/route/rr_graph_generation/rr_graph.cpp
AmirhosseinPoolad: I'm passing a value to vtr::NdMatrix::resize, which automatically calls fill inside.

```
    /**
     * @brief Resize the matrix to the specified dimension ranges
     *
     * If 'value' is specified all elements will be initialized to it,
     * otherwise they will be default constructed.
     */
    void resize(std::array<size_t, N> dim_sizes, T value = T())
```

vpr/src/route/route_common.cpp
AmirhosseinPoolad: I'm not sure if there's an easy way to have per-connection bounding boxes without changing a lot of functions. It's a good idea to investigate this later though.

vpr/src/route/rr_graph_generation/rr_graph.cpp
soheilshahrouz: You are right. I thought resize() only initializes new elements like std::vector.

vaughnbetz (approved):
LGTM

PR #3613, opened 2026-06-10
Fix a drawing issue related to block internal text

kevinxing-a12:
#### Description

In the previous code, if a block internal has valid children blocks, its name will be drawn at the top of the block (versus in the middle) to avoid being overdrawn by the children blocks. However, it doesn't check if the drawing of children blocks actually took place or not. Therefore, even if the children are hidden, the mother's name is still moved to the top. Due to the limited space at the top, the text won't be drawn until we zoom in. The pictures below show the difference before and after this PR:

To fix this issue, I had to change the logic inside the recursive function draw_internal_pb(). Previously, the order was:
```
draw box -> draw text -> recursion
```
Now, it is changed to:
```
draw box -> recursion -> draw text
```
I also changed the function's return type from void to bool, and the boolean is used to indicate if the drawing of children was successful or not.

Furthermore, I made a few other changes to draw_place() in draw_basic.cpp to improve consistency with draw_internal_pb(). Details are explained in the text drawing section in draw_internal_pb().

#### Motivation and Context

Improve drawing correctness and clarity

Tested with several different architectures, and checked for the expected visual result.

#### Types of changes

#### Checklist:


vpr/src/draw/draw_basic.cpp
```diff
@@ -155,19 +155,19 @@ void draw_place(ezgl::renderer* g) {
...
+                                         name, abs_clb_bbox.width(), abs_clb_bbox.height());
                         }
 
                         // Draw text for block type so that user knows what block
                         std::string block_type_loc = type->name;
                         block_type_loc += vtr::string_fmt(" (%d,%d)", i, j);
-
-                        g->draw_text(center - ezgl::point2d(0, abs_clb_bbox.height() / 4),
-                                     block_type_loc, abs_clb_bbox.width(), abs_clb_bbox.height());
+                        g->draw_text(center, block_type_loc, abs_clb_bbox.width(), abs_clb_bbox.height());
```
AmirhosseinPoolad, line 172: Add a comment explaining why you adjust center for one of the draw_test calls and not for the other

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -84,7 +85,23 @@ void collect_pb_atoms_recurr(const t_pb* pb, std::vector<AtomBlockId>& atoms);
 t_pb* highlight_sub_block_helper(const ClusterBlockId clb_index, t_pb* pb, const ezgl::point2d& local_pt, int max_depth);
 
 #ifndef NO_GRAPHICS
-static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezgl::rectangle& parent_bbox, const t_logical_block_type_ptr type, ezgl::renderer* g);
+/**
+ * @brief Helper subroutine to recursively draw sub-blocks.
+ *
+ * This function traverses through the pb_graph which a netlist block can map to,
+ * and draws each sub-block inside its parent block. A dynamic level of detail check is also implemented to
+ * determine whether the drawing should actually take place. The mother block recursively calls this function on
```
AmirhosseinPoolad: I think we use the term "root" in the rest of the code base.

AmirhosseinPoolad: Also it's not just the root pb that calls it recursively, everyone does. I think saying "each pb recursively..." would be more accurate here. Although I might've misinterpreted "mother block" as "root pb."

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -445,11 +457,60 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
         g->draw_rectangle(abs_bbox);
     }
 
+    // Now recurse on the child pbs.
+    // Note: we make the recursive function calls before drawing text,
+    // because the text position depends on whether or not the children blocks were drawn.
+    // So, the overall drawing order is: mother-box -> child-box -> (recursion) -> child-text -> mother-text.
```
AmirhosseinPoolad: Wouldn't the child box and text be contained inside the recursion?

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -445,11 +457,60 @@ static void draw_internal_pb(const ClusterBlockId clb_index, t_pb* pb, const ezg
...
     // Display text for each physical block.
     std::string pb_display_text(pb_type->name);
     std::string pb_type_name(pb_type->name);
 
-    if (!pb->is_primitive()) {
+    // This is a special case: blocks that correspond to pb_type->depth == 0 are essentially the top-level clustered blocks,
+    // and in each drawing cycle they are already drawn in draw_place() in draw_basic.cpp. The reason why they appear here again is that,
+    // the naming convention for clustered blocks has been historically different between the two functions, and we want to
+    // use the one in this function to overdraw the other when level of detail is on. To ensure consistency,
+    // when level of detail is off (!at_least_one_child_pb_drawn), we will still use the naming convention in draw_place().
```
AmirhosseinPoolad: I was a bit confused when reading about "level of detail" being on or off. Don't have any specific suggestion but if you could make it a bit more clear (LoD on -> we're not showing any child pbs of the root pb vs LoD off) it would be nice.

AmirhosseinPoolad (commented):
Thanks Kevin! The code looks fine to me, had a bunch of commenting nitpicks.

AmirhosseinPoolad:
Regarding the highlighting issue that we talked about, I believe if you change highlight_sub_block_helper() to take into account the minimum screen area ratio it would fix that too. Take a look to see if it would be simple enough to add to this PR. If not, let's do it in the next PR.

vpr/src/draw/intra_logic_block.cpp
kevinxing-a12: Sorry about the confusion! My intention was that any pb owing children blocks can be called a mother block.

kevinxing-a12: I will use a different word and phrase it better.

vpr/src/draw/intra_logic_block.cpp
kevinxing-a12: My bad. I was trying to say "further recursion". Will change the wording

vpr/src/draw/intra_logic_block.cpp
kevinxing-a12: I think "parent" should be good, since it's also used in other parts of the code.

kevinxing-a12:
The highlight thing it's possibly harder than what we expected; a global variable cannot handle the case where two clustered blocks inherently have different block internal depths. Maybe a vector holding the LoD depth of each clustered block is needed?

vpr/src/draw/draw_basic.cpp
kevinxing-a12: Added a comment for the draw_text call above

vaughnbetz (approved):
LGTM, thanks!

vpr/src/draw/intra_logic_block.cpp
```diff
@@ -832,9 +881,11 @@ t_pb* highlight_sub_block_helper(const ClusterBlockId clb_index, t_pb* pb, const
             // get the bbox for this child
             const ezgl::rectangle& bbox = draw_coords->get_pb_bbox(clb_index, *pb_child_node);
 
-            // If child block is being used, check if it intersects
-            if (child_pb->name != nullptr && bbox.contains(local_pt)) {
-                // check farther down the graph, see if we can find
+            ezgl::renderer* g = application.get_renderer();
+            // If child block is being used, check if it intersects. Check also if it is visible (drawn) on screen,
+            // because otherwise it would be unavailable for selection.
+            if (child_pb->name != nullptr && bbox.contains(local_pt) && check_if_draw_on_screen(bbox, g)) {
```
vaughnbetz: minor nit: instead of check_if_draw_on_screen, I'd prefer a name that indicates what true means.

E.g.: 
large_enough_to_draw (bbox, g)

would tell the reader more than check_if_draw_on_screen.

vaughnbetz:
@AmirhosseinPoolad : are all your comments addressed now?

AmirhosseinPoolad (approved):

kevinxing-a12:
Is this PR good to merge? :) @AmirhosseinPoolad @vaughnbetz

vaughnbetz (approved):
LGTM

vaughnbetz:
You need to resolve some git conflicts, and then it can be merged.

kevinxing-a12:
Is it because of the conflict from rebase? I'm working on a sub-branch on this branch now, so if possible can we do a direct merge?

vaughnbetz:
Looks like we can do a merge but not a rebase. I'll do that.  @AmirhosseinPoolad : I think you're OK with merging this now?

AmirhosseinPoolad:
Yes the PR looks fine, LGTM

AmirhosseinPoolad:
Let's merge this instead of rebase since Kevin's other branch is based on this. We would probably mess with his git too much if we rebased this into master. I'm waiting for the QT PR to be merged and would then merge this after.

vaughnbetz:
Updating branch to trigger CI again since we had a huge Qt merge that I want to make sure this was tested with. Then will merge if CI is OK.

PR #3615, opened 2026-06-11
[CI] Locked Tight Floorplanning to Traditional Flow

AlexandreSinger:
The original tight floorplanning testcases were designed based off the original pack, place, and route flow; they were built to just barely be able to legalize with the original packing algorithm.

The analytical placement flow is struggling to find the exact solution that the original flow found due to issues in how the region constraints are handled. I locked these tests to the traditional flow while I investigate improved handling of region constraints for AP.


vaughnbetz (approved):
LGTM

PR #3617, opened 2026-06-12
[VTR][Geometry] Updated Rect Contains Syntax

### LINKED ISSUE #2868: [LibVTRUtil] Confusing API for vtr::rect
While working on the initial placer in VTR, I found a bug when the code is performing centroid placement and is determining if a centroid placement is legal or not:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/848d1e77d8ea222e6f84716172c12b0eb84d8347/vpr/src/place/initial_placement.cpp#L302-L333

The issue is that the code creates a partition region that is the size of the device for unconstrained blocks. For example, for an 11x11 architecture it creates a rectangle from (0, 0) to (10, 10). The problem is that this code uses the `contains` method from vtr::rect. This does not include the top-right edges; so any centroids which were placed on the top-right edge of the device were being wrongfully rejected. 

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/848d1e77d8ea222e6f84716172c12b0eb84d8347/libs/libvtrutil/src/vtr_geometry.h#L171-L181

An easy fix for this problem is to use the `coincident` method instead; however, there may be other misunderstandings throughout VTR which need to be investigated.

In reality, this interface is a little confusing, and these methods should be renamed or even removed if they are not being used.

We will need to look through the code for cases where these methods are used and ensure they are correct.

AlexandreSinger:
I found the difference between contains and coincident confusing and found an example of a bug where this confusion can cause issues.

I updated the documentation and function names in vtr::Rect to try and alleviate confusion. I also documented the uses of these methods in the code to verify that the correct method was used in each case.

Resolves https://github.com/verilog-to-routing/vtr-verilog-to-routing/issues/2868

### DIFF
--- a/libs/libvtrutil/src/vtr_geometry.h
+++ b/libs/libvtrutil/src/vtr_geometry.h
@@ -166,14 +166,23 @@ class Rect {
     ///@brief Return the rectangle height
     T height() const;
 
-    ///@brief Returns true if the point is fully contained within the rectangle (excluding the top-right edges)
+    /// @brief Returns true if the point is within the rectangle using half-open bounds:
+    ///        xmin <= x < xmax  AND  ymin <= y < ymax  (top and right edges are excluded).
+    ///
+    /// Use this when the rectangle represents an index range where xmax/ymax is one past
+    /// the last valid value (e.g., Rect(0, 0, width, height) for a width x height matrix).
     bool contains(Point<T> point) const;
 
-    ///@brief Returns true if the point is strictly contained within the region (excluding all edges)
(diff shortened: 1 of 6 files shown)


AmirhosseinPoolad (approved):
I think this makes sense. To be clear, the bug you found and fixed was in a past PR, not this one, right?

AlexandreSinger:
> To be clear, the bug you found and fixed was in a past PR, not this one, right?

Correct. The past issue that I fixed was the reason I raised the issue originally. I went through each usage and ensured that they were correct.

PR #3618, opened 2026-06-12
router_opt_choke_points rst doc

amin1377:
Add documentation for `router_opt_choke_points` in the RST file and fix the choking point test.

### DIFF
--- a/doc/src/vpr/command_line_usage.rst
+++ b/doc/src/vpr/command_line_usage.rst
@@ -1589,6 +1589,13 @@ VPR uses a negotiated congestion algorithm (based on Pathfinder) to perform rout
 
     **Default:** ``off``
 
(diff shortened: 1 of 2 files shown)


doc/src/vpr/command_line_usage.rst
AmirhosseinPoolad, line 1624: Not related to this PR, but I think "Values of 1000" is wrong and should be changed. I think the factor is a number between 0 and 1.

vaughnbetz (approved):
Thanks!

vaughnbetz:
@amin1377 : any comment on Amir's comment above?

amin1377:
@vaughnbetz Amir is right that saying values of `1000` are reasonable is not correct. However, the `pres_fac` does not have to be between `0` and `1`. I’m making another PR to fix the documentation.

PR #3620, opened 2026-06-12
[AP][Graphics] Fixed Issues with AP Graphics

AlexandreSinger:
Now that AP is becoming the default flow, we need to ensure that its graphics are working. I found that the graphics were not working at all for any stage in the flow.

I fixed this issue and cleaned up AP's graphics a bit to make it clean enough to be the default.


vaughnbetz (approved):
LGTM

PR #3626, opened 2026-06-13
[CI] Added Strong Tests for Traditional Flow

AlexandreSinger:
A comment on the PR making AP the default flow was a concern that we may regress the old pack, place, and route flow. This is a valid concern, so I added more testing to the strong tests for the pack, place, and route flow for each of the benchmark suites.


AlexandreSinger:
@amin1377 FYI, this is regarding your comment on https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3570

AlexandreSinger:
I am worried about adding duplicate tests in the Weekly tests for the traditional flow since the weekly tests are getting a bit bloated. We may need to discuss how we can improve that situation since we are nearing 24 hour weekly test run time.

amin1377 (approved):
Thanks, Alex!

AlexandreSinger:
I am going to wait on merging this until after the graphics changes. I am taking no chances!

PR #3627, opened 2026-06-14
[CI] Added Ubuntu 26.04 Compatibility Test

AlexandreSinger:
Ubuntu 26.04 was released as a public preview for the GitHub Action Runners. The General Availability of these runners is coming at the end of the summer. To get prepared, I have added a compatibility test to the CI. Once the GA runners are available for 26.04, we will switch the whole CI over.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -797,6 +797,37 @@ jobs:
         export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
         make -j${{ steps.cpu-cores.outputs.count}}
 
+  ResoluteCompatibility:
+    name: 'Ubuntu Resolute - 26.04 Compatibility Test'
+    runs-on: ubuntu-26.04
+    needs: [BuildVTR]
+    env:
+      CC: gcc-15
+      CXX: g++-15
+    steps:
+    - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
Looks like we are getting a warning in CapNProto:
image

I recall this being an issue with our version of CapNProto that we currently have.

There are a couple of ways we can fix this. One of them is that we can upgrade our version of CapNProto. I think for now, just to get this test in, we can turn off the error on warnings for this test. That way this test can come in and we can slowly remove the warnings.

AlexandreSinger:
The build was successful. There are three classes of warnings left that can be resolved in future PRs:

- CapNProto extra semi-colons as shown above.
- A few potential null-ptr dereferences.
- A very interesting one in argparse: 
image

The CapNProto one will likely just be updating our version of CapNProto; the potential null-ptr dereferences are usually pretty easy to fix; and I am not sure what is wrong with the last one, my guess is that this file is so large that it is overflowing the compiler's global common subexpression elimination (GCSE) buffer. We may want to just increase this since this file is supposed to be very large.

AmirhosseinPoolad:
Finally GH released this. LGTM.

AmirhosseinPoolad:
I've been seeing these warnings for a while now lol. Time to fix'em soon.

PR #3629, opened 2026-06-14
--device_width command line option

soheilshahrouz:
Adds a `--device_width` VPR command-line option so users can fix the FPGA grid width while keeping `--device auto`. When set, VPR derives grid height from the `<auto_layout>` aspect ratio in the architecture file instead of auto-sizing the device to fit the circuit.


soheilshahrouz:
@vaughnbetz @AmirhosseinPoolad 
This PR is ready for review. I'd appreciate if you could take a look at it.

vpr/src/base/setup_grid.cpp
```diff
@@ -67,30 +68,62 @@ static void set_grid_block_type(int priority,
...
+        return vtr::nint(width / aspect_ratio.value());
+    }
+    VPR_FATAL_ERROR(VPR_ERROR_ARCH,
+                    "Cannot compute device height from width: architecture has no <auto_layout> to define aspect ratio.\n");
+    return 0; // Unreachable
+}
+
+bool has_fixed_device_size(const std::string& device_layout, size_t device_width) {
+    return device_layout != "auto" || device_width > 0;
+}
```
AmirhosseinPoolad: It''s not clear that the inputs to this function are command line args. I'm not entirely sure how to fix that but one possible solution is to rename "device_layout" to "device_layout_name" and "device_width" to "explicit_device_width" to make it more obvious. If you can think of better solutions/names I'm open to that.

AmirhosseinPoolad: I'm nitpicking to be honest but I can see someone misusing this function in the future.

soheilshahrouz: I'll pass `t_vpr_setup` instead.

AmirhosseinPoolad (approved):

vpr/src/base/CheckSetup.cpp
```diff
@@ -5,26 +5,40 @@
...
         VPR_FATAL_ERROR(VPR_ERROR_OTHER,
                         "Packing cannot be timing driven without timing analysis enabled\n");
     }
 
+    if (device_width > 0 && device_layout != "auto") {
+        VPR_FATAL_ERROR(VPR_ERROR_OTHER,
+                        "--device_width is only valid when --device is 'auto'\n");
+    }
+
+    if ((int)device_width < 0) {
```
vaughnbetz: Should this be <= 0?  A device_width of 0 doesn't seem to make sense (and then the message would be device_width must be positive).

vpr/src/base/setup_grid.h
```diff
@@ -8,16 +8,28 @@
...
 
 #include <vector>
+#include <optional>
 #include "physical_types.h"
 
 class DeviceGrid;
+struct t_vpr_setup;
 
-///@brief Find the device satisfying the specified minimum resources
+/// @brief Returns aspect ratio from the single auto_layout, or std::nullopt if none
```
vaughnbetz: if none seems vague; suggest 'if there is no auto_layout defined for the target architecture.'

vpr/src/base/vpr_types.h
```diff
@@ -1631,6 +1631,8 @@ struct t_vpr_setup {
     std::string GraphicsCommands;        ///<commands to control graphics settings
     t_power_opts PowerOpts;
     std::string device_layout;
+    /// When > 0 and device_layout is "auto", use this fixed grid width (height from aspect ratio).
```
vaughnbetz: change to 'and a height from the auto layout's aspect ratio'

vaughnbetz (changes requested):
A few nits to fix (minor).

vpr/src/base/CheckSetup.cpp
soheilshahrouz: `deive_width=0`means that auto resizing should determine the device size. I changed the condition to `if (device_width <= 0 && device_layout == "auto")`

soheilshahrouz:
@vaughnbetz I applied your comments. If you have no further comments, I think this is ready to merge.

vaughnbetz (approved):
LGTM

PR #3630, opened 2026-06-15
Add initial windowed bipartite matching detailed placer

AthavanBalakumar:
Initial implementation of windowed_bi_matching detailed placer. Provides a scaffold for future work on windowed bipartite matching.

Currently, the device grid is scanned in 2x2 windows. Windows containing empty tiles are skipped, and windows where sub-tile locations don't contain placed blocks are skipped. For suitable windows, the top left and top right blocks are swapped. This verifies that the algorithm is able to access and modify the clustered placement.

The number of swaps is logged as a debugging measure.


AthavanBalakumar:
@AlexandreSinger 

Results from before and after the bounding box cost based swapping is implemented.

| Circuit | Before routed_wirelength | After routed_wirelength | New/old ratio routed_wirelength | Before routed_wiresegment | After routed_wiresegment | New/old ratio routed_wiresegment |
|---|---:|---:|---:|---:|---:|---:|
| apex4 | 9594 | 9537 | 0.994059 | 2463 | 2449 | 0.994316 |
| des | 9836 | 9878 | 1.004270 | 2692 | 2694 | 1.000743 |
| seq | 10150 | 10005 | 0.985714 | 2626 | 2578 | 0.981721 |
| ex1010 | 43838 | 42237 | 0.963479 | 11803 | 11285 | 0.956113 |

Detailed placer stats, after cost based swapping is implemented.

| Circuit | Initial BB wirelength estimate | Final BB wirelength estimate | New/old ratio | Swaps committed |
|---|---:|---:|---:|---:|
| apex4 | 6317.15 | 6316.64 | 0.999919 | 3 |
| des | 7121.74 | 7121.74 | 1.000000 | 0 |
| seq | 6636.58 | 6617.84 | 1.000000 | 4 |
| ex1010 | 26627.3 | 26580.4 | 0.998239 | 14 |

vpr/src/analytical_place/detailed_placement/windowed_bi_matching_dp.cpp
```diff
@@ -1,9 +1,225 @@
...
+}
+
+t_pl_loc WindowedBiMatchingDetailedPlacer::make_pl_loc(int x, int y, int sub_tile, int layer) {
+    t_pl_loc loc;
+    loc.x = x;
+    loc.y = y;
+    loc.sub_tile = sub_tile;
+    loc.layer = layer;
+    return loc;
+}
```
AlexandreSinger, line 67: This member function, as well as many others before which do not use anything in the placer state, should probably be made into static methods. They are more efficient and it also keeps the function declarations out of the header file where they are not needed. Its usually good style to use static methods when you can.

It would look something like this:
```cpp
// At the top of the file (forward declaration)
// Add the function comment here.
static t_pl_loc make_pl_loc(...);

// At this point in the file:
static t_pl_loc make_pl_loc(...) {
  ...
}
```

vpr/src/analytical_place/detailed_placement/windowed_bi_matching_dp.cpp
```diff
@@ -1,9 +1,225 @@
...
+
+bool WindowedBiMatchingDetailedPlacer::is_non_empty_physical_tile(const DeviceGrid& grid, int x, int y, int layer) {
```
AlexandreSinger: For this function signature and a few of the others, you pass in x, y, and layer independently. This is very error prone since you may accidently write the order as layer, y, x. Since they are all integers, the compiler would not complain.

I recommend passing in a struct when appropriate. For example, for this one I would use `t_physical_loc` since it is used by the `get_physical_type` function I think.

vpr/src/analytical_place/detailed_placement/windowed_bi_matching_dp.cpp
```diff
@@ -1,9 +1,225 @@
...
+        return false;
+    }
+    const vtr::vector_map<ClusterBlockId, t_block_loc>& block_locs = blk_loc_registry.block_locs();
+    if (block_locs[block_a].is_fixed || block_locs[block_b].is_fixed) {
+        return false;
+    }
+    if (clb_nlist.block_type(block_a) != clb_nlist.block_type(block_b)) {
+        return false;
+    }
+    return true;
```
AlexandreSinger, line 111: I recommend adding some very short functions over each of the checks here. It may be helpful in the future.

Something like:
```cpp
// Check if either block is fixed. Cant swap fixed blocks.
// Check if the blocks have the same type. Cant swap blocks that are different types.
```

AlexandreSinger: I recognize that this code is still WIP, but you should add a TODO to this code somewhere. Two blocks that are different types may still be swappable. Two blocks are swappable so long as both clusters can exist in the other location.

vpr/src/analytical_place/detailed_placement/windowed_bi_matching_dp.cpp
```diff
@@ -1,9 +1,225 @@
...
+        net_cost_handler_.update_move_nets();
+        blk_loc_registry.commit_move_blocks(blocks_affected);
+        blocks_affected.clear_move_blocks();
+        return true;
+    }
+
+    net_cost_handler_.reset_move_nets();
+    blk_loc_registry.revert_move_blocks(blocks_affected);
+    blocks_affected.clear_move_blocks();
+    return false;
```
AlexandreSinger, line 166: I would add comments to this function as well. It is quite important to what you are doing.

vpr/src/analytical_place/detailed_placement/windowed_bi_matching_dp.cpp
```diff
@@ -1,9 +1,225 @@
...
+            // Temporarily evaluate swaps between the top-left and top-right blocks
+            t_pl_loc left_loc = make_pl_loc(x, y, placement_sub_tile_, placement_layer_);
+            t_pl_loc right_loc = make_pl_loc(x + 1, y, placement_sub_tile_, placement_layer_);
+
+            ClusterBlockId left_blk = grid_blocks.block_at_location(left_loc);
+            ClusterBlockId right_blk = grid_blocks.block_at_location(right_loc);
+
+            if (!blocks_are_swappable(blk_loc_registry, clb_nlist, left_blk, right_blk)) {
+                continue;
+            }
```
AlexandreSinger, line 207: I recognize that you made this very simple (thats a good idea for this PR). As a note for a future PR, I think this code can actually be simplified.

You check if two blocks are swappable; however the try_swap_blocks method below checks if the two blocks can be swapped implicitly (e_block_move_result::VALID I think is telling you this information). You can probably drop the if statement.

You are also ensuring that the window is full of placeable blocks and have valid tiles. You may not need to do that. You can have try_swap swap locations instead of blocks. If a source or target location is an empty tile it just does not do the swap. The reason I point this out is that you may have tiles that have no clusters in them that you are never considering; you are only swapping already placed blocks. It would be a great upgrade to allow blocks to move to valid, empty (no other blocks are there) locations.

vpr/src/analytical_place/detailed_placement/windowed_bi_matching_dp.cpp
```diff
@@ -1,9 +1,225 @@
...
+    placement_ctx.mutable_blk_loc_registry() = placer_state_.blk_loc_registry();
+    if (num_swaps > 0) {
+        post_place_sync();
+    }
+    // Verify that the placement remains valid after the detailed-placement swaps.
+    unsigned num_placement_errors = verify_placement(g_vpr_ctx);
+    if (num_placement_errors == 0) {
+        VTR_LOG("Completed placement consistency check successfully.\n");
+    } else {
+        VPR_ERROR(VPR_ERROR_AP,
```
AlexandreSinger: I think this should be VPR_ERROR_PLACE

vpr/src/analytical_place/detailed_placement/windowed_bi_matching_dp.h
```diff
@@ -17,7 +19,58 @@
...
+                                     const t_placer_opts& placer_opts);
 
     void optimize_placement() final;
+
+  private:
+    PlacerState placer_state_;
+    NetCostHandler net_cost_handler_;
+    int window_size_ = 2;        ///< Size of local placement window, forced to 2 temporarily
+    int placement_layer_ = 0;    ///< Device layer processed by placer, forced to 0 temporarily
+    int placement_sub_tile_ = 0; ///< Sub-tile processed by placer.
```
AlexandreSinger: I would also say forced to 0 temporarily. This is another thing that will eventually need to done by your algorithm. IO tiles, for example, usually have multiple sub_tiles in them for our architectures.

AlexandreSinger (changes requested):
Looks great @AthavanBalakumar 

Some comment below.

AthavanBalakumar:
@AlexandreSinger I believe I have appropriately addressed all the comments for this PR, please let me know if there's anything I've missed.

AlexandreSinger (approved):
LGTM

PR #3631, opened 2026-06-16
[WIP][AP][PartialLegalizer] Made Partial Legalizer Criticality-Aware

AlexandreSinger:
After looking at the drawing for AP, I noticed that the critical path produced during full legalization was way wider than expected. This implied that the partial legalizer was not handling critical paths correctly. After some thought, I realized that the solver was probably not provided enough pull to keep the critical path together. The best way to fix that is to inform the partial legalizer of the criticality of blocks when partitioning so that critical blocks (the blocks with critical pins) get priority when deciding which partition to move to.


AlexandreSinger:
I am collecting some results now on Titan, but local results showed a strong improvement in CPD.

AlexandreSinger:
Results on `titan_quick_qor`:

  | titan_ap_master.txt | parse_results.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.07
num_LAB | 1 | 1.03
num_DSP | 1 | 1.00
num_M9K | 1 | 1.00
num_M144K | 1 | 1.00
max_vpr_mem | 1 | 1.00
num_post_packed_blocks | 1 | 1.03
device_grid_tiles | 1 | 1.01
pack_time | 1 | 1.02
placed_wirelength_est | 1 | 0.99
place_time | 1 | 1.15
placed_CPD_est | 1 | 0.97
routed_wirelength | 1 | 0.99
critical_path_delay | 1 | 0.96
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 0.96
crit_path_route_time | 1 | 1.01

With this change, the CPD improved by 4% and the routed WL improved by 1%. The place time increased by 15%, which is not completely clear why.

I am thinking about updating our QoR compare script to always show the post-GP, post-FL, and post-DP WL/CPD; but that may become a bit messy. I like this script being a short summary. We can discuss.

Some more data on the per-circuit CPD for those interested:
image

Raw data: 
[titan_ap_crit_pl.xlsx](https://github.com/user-attachments/files/29000123/titan_ap_crit_pl.xlsx)

@amin1377 FYI

AlexandreSinger:
This was clearly a good idea, but since AP is so early in the flow, it is causing many golden results to change.

Not a problem, but a bit annoying to update.

There are some more updates that I am considering which I think I will add to this PR as well:

1. Long-chains (pre-packed chains of primitives that span multiple molecules) are not being constrained to move together in this flow. I have a basic commit that combines these together into single APBlocks and moves them together. I think the interaction with this change is important.
2. The global placer saves the best result towards the last iteration. What it considers as the best is dependent only on wirelength and not CPD. It should really be a mix of the two, plus a small bonus for being at a later iteration (later iterations may imply more legal). I may need to experiment to see if we should even be saving the best result, since I have found that the legality can improve with later iterations.
3. Clean up the generic parsing scripts so the post-GP, FL, and DP WL/CPD are always parsed to make it easier to track these metrics.

I may want to try both of these in this PR to limit the changes to Master.

AlexandreSinger:
@vaughnbetz FYI.

vaughnbetz:
That all makes sense to me.  Adding the additional parse metrics to the QoR script is fine with me.
Making a script that updates all golden results (or failing golden results) is also OK with me.  It is a bit annoying for changes like this to have to update manually.
Not sure why the runtime went up .... hopefully we can claw back.  Maybe fewer iterations?

AlexandreSinger:
Just noting down my next steps so I do not forget. I want to sweep how much the partial legalizer should focus on criticality since this is not obvious if it should be 50% or not. To do that I need to update the task parsing script to include the AP parameters I care about.

PR #3633, opened 2026-06-16
Choke point documentation

amin1377:
Add to the choke point CLI documentation that this option only affects the flat router.


vaughnbetz (approved):
LGTM

PR #3636, opened 2026-06-18
Adding bus attribute to mux tag

amin1377:
Add a `bus` attribute to the mux tag, set to `false` by default. If a bus-based mux is needed, it should be specified explicitly. This would reduce the chance of inadvertent mistakes and make bus-based muxes easier to detect. This detection is essential for downstream tools when assigning configuration bits and generating the bitstream.


doc/src/arch/reference.rst
```diff
@@ -1686,6 +1689,26 @@ The following describes the tags that are accepted in the ``<interconnect>`` tag
...
+        <mux name="a2a" input="one_mult_27x27.a one_mult_27x27.b" output="mult_27x27.a" bus="true"/>
+
+    A data line may also be a bit-slice of a wider port, or a concatenation of
+    pins using ``{...}``; in every case the width of each data line must match
+    the output width. Here each data line is a 27-bit slice of the 54-bit
+    ``mult_27.datain`` bus:
+
+    .. code-block:: xml
+
+        <mux name="datain2a" input="mult_27.datain[26:0] mult_27.datain[53:27]" output="one_mult_27x27.a" bus="true"/>
```
AlexandreSinger, line 1710: Up to you if you want to document this or not, but I am wondering if it may be a good idea to add a warning that currently this is just instantiating a set of muxes, instead of being a true multi-bit bus as of right now. This would just make it less magical for external people working with VTR who may not know this.

AlexandreSinger (approved):
LGTM, I just had a small comment on documentation regarding what we talked about earlier.

libs/libarchfpga/src/read_xml_arch_file.cpp
```diff
@@ -1847,17 +1847,26 @@ static void process_interconnect(vtr::string_internment& strings,
     for (auto child_name : {"complete", "direct", "mux"}) {
         pugi::xml_node cur = get_first_child(Parent, child_name, loc_data, ReqOpt::OPTIONAL);
 
+        const bool is_mux = (0 == strcmp(child_name, "mux"));
```
AmirhosseinPoolad: I think we can use std::string comparison. [Seems like](https://godbolt.org/z/a8E1dM5EE) the C++ type system does the right thing here.

AmirhosseinPoolad (approved):

PR #3638, opened 2026-06-18
Change floating RR Node warning to verbose mode

AmirhosseinPoolad:
This PR changes the RR Graph warnings regarding floating RR Nodes (nodes with no out-going edges) to happen in detail only in verbose mode. With default route_verbosity, we now only print the number of floating RR Nodes and not every single one.

I believe floating wires are expected in device edges and we have a lot of them in interposer-based architectures. Here's how the warnings look like for an interposer architecture now:

> Warning 1433: Found 11654 nodes with no out-going switches. Run VTR with route_verbosity > 1 to see details.

### DIFF
--- a/libs/librrgraph/src/io/rr_graph_reader.cpp
+++ b/libs/librrgraph/src/io/rr_graph_reader.cpp
@@ -127,6 +127,7 @@ void load_rr_file(RRGraphBuilder* rr_graph_builder,
         &arch->strings,
         schema_file_id,
         is_flat,
+        route_verbosity,
         device_model_warnings);
 
     if (vtr::check_file_name_extension(read_rr_graph_name, ".xml")) {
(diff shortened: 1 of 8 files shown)


soheilshahrouz (commented):
I think passing route_verbosity as an argument to almost every single function is a mistake. It should a be global variable or integrated into a logger class.

AlexandreSinger:
Although I agree with you @soheilshahrouz that a global logger class is the ideal solution; I am not sure if that is on this PR to fix.

I think that we should remove the warnings with this PR (since they are just noise at this point and they slow down the parsing scripts by a huge amount), and then we can raise an issue to add a logger to VPR.

What do you think?

soheilshahrouz:
I agree.

AlexandreSinger (approved):
LGTM

griffinandrew:
Great Work!!!

PR #3640, opened 2026-06-18
Add 2D Hecate (7nm) architecture

AmirhosseinPoolad:
Adds a 2D 7nm architecture which is same as our 2.5D and 3D architectures but it's 2D.


AmirhosseinPoolad:
@AlexandreSinger I think I should add a strong test for this (which is why this PR is WIP now), but I can't find a good place to add the config file. Do you have any suggestions?

AlexandreSinger:
> I think I should add a strong test for this (which is why this PR is WIP now), but I can't find a good place to add the config file. Do you have any suggestions?

I think adding it in `vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_interposer/hecate_interposer/config/config.txt` would make the most sense to me as another architecture in the list to sweep. I think we can add a comment there saying that we are adding a non-interposer version of the architecture to ensure that the general fabric is working as expected.

Other than that we can add another strong test just for it; but I do like combining task scripts that are relavent to each other.

AmirhosseinPoolad:
Couldn't add it to the previous tasks since they were in separate folders (tried doing `../[some_path]` but it didn't work). Turned the koios_no_hb strong test into two tests, one for the 22nm arch and another for the 7nm arch. The DSP in the 7nm arch is slightly different so we can't use the hard blocks for this architecture.

AlexandreSinger:
You should be able to have multiple architectures that are in separate folders. I am not sure why the relative paths are not working for `..`, but what I normally do is use the last common directory and then use relative paths from there.

For example:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/a5b86e0c796b6ed6648291a8241d492232a7463e/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_pack_place_route/config/config.txt

AmirhosseinPoolad:
>  but what I normally do is use the last common directory and then use relative paths from there.

Oh yeah that's a very obvious solution that I missed. Hopefully everything works now!

AlexandreSinger (approved):
LGTM.

One golden result looks like it was clobbered by another PR. Once that it resolved, I think this can be merged.

vtr_flow/tasks/hecate_25D/config/config.txt
AlexandreSinger, line 1: Was it intentional to push this config file in?

AlexandreSinger (commented):
I think an extra config file was pushed. I do not mind having config files like this in VTR, but this particular one looks to be a development config file.

vtr_flow/tasks/hecate_25D/config/config.txt
AmirhosseinPoolad: Oops, good catch.

AlexandreSinger (approved):
LGTM

Issue #3634, opened 2026-06-18
[AP] Add Nesterov Placer Option


WilliamZhang20:
#### Proposed Behaviour
Add an optional Nesterov-based nonlinear global placer to the AP flow that minimizes a smooth log-sum-exp wirelength model plus a bin-density penalty, optimized with a preconditioned Nesterov accelerated gradient.

#### Current Behaviour
The AP global placer (QPHybridSolver / B2BSolver in analytical_place/global_placement/) minimizes a quadratic wirelength surrogate by solving a sparse SPD system with conjugate gradient, and leans on the partial legalizer plus anchor pseudo-nets to handle overlap and legality outside the objective. That quadratic model over-penalizes long nets relative to true HPWL, and there is currently no nonlinear or accelerated-gradient solver option in the flow.

#### Possible Solution
Draft implementation at this branch: https://github.com/WilliamZhang20/vtr-verilog-to-routing/tree/nesterov-placer

Benchmarked QoR is still mixed compared to B2B and QP (i.e. runtime and circuit timing are better, but wirelength is not), so some fixes and adjustments are needed.

#### Context
Hope to incorporate some modern nonlinear approaches, i.e. RePlAce/DREAMPlace on the ASIC side, elfPlace for FPGAs. Those methods optimize a smooth wirelength model with overlap handled as a density penalty within the objective, rather than VTR's approach of minimizing a quadratic surrogate and pushing legality out to a separate legalizer and anchor step

AlexandreSinger:
Hi @WilliamZhang20 ,

This is a very interesting proposal! This was something I was looking into as well when I was originally developing the AP flow. I originally wanted to use this as my global placer, but went down the quadratic route since it was more stable (at least for me) across many different heterogeneous architectures. The biggest issue I was running into was the model of density; however, that analysis pre-dates the concept of primitive mass which I added to the partial legalizer. I was planning to revisit this later after the AP flow became more stable (which it is, since it is in master). I am very interested in how you model density in your model!

I would be interested to meet and discuss sometime soon if you are interested.

Best,
Alex

amin1377:
Hi @WilliamZhang20,

Interesting work! Could you please share more details about the benchmarks you ran? It would be great if you could also share the data, including circuit sizes, the architecture used, and any other relevant setup details.

PR #3641, opened 2026-06-19
[AP] Add Nesterov Placer Option

WilliamZhang20:
#### Description

Adds a Nesterov-style nonlinear analytical global placer for AP, selectable with:

```sh
--ap_global_placer nonlinear-nesterov
```

(alongside the existing `simpl` global placer). The placer optimizes a smooth nonlinear objective over movable AP block coordinates:

$$
\min_{x,y}\; W_\gamma(x,y) + \lambda\ D(x,y)
$$

where $W_\gamma$ is a smooth weighted-average wirelength approximation and $D$ is a continuous, primitive-vector-aware electrostatic bin-density penalty. The optimizer:

- seeds itself from a short B2B/QP warm-start solve so it starts from a wirelength-aware placement,
- runs accelerated first-order **Nesterov** updates with restart/step-shrink behavior over a small number of density-continuation epochs, partially legalizing between epochs to anchor the result, then
- hands the placement to the existing AP partial legalizer.

It supports timing-aware smooth wirelength weighting:

$$
w_n = \alpha \cdot \mathrm{crit}(n) + (1 - \alpha)
$$

where $\alpha =$ `--ap_timing_tradeoff` and $\mathrm{crit}(n)$ is the pre-cluster setup criticality of the atom net. For designs with long direct I/O-chain nets, it additionally applies a light chain-cohesion term to keep pad/obuf/OCT/termination chains coherent through the AP-to-APPack handoff; this term is gated off automatically for designs without such chains.

#### Related Issue

Issue #3634

#### Motivation and Context

This provides a nonlinear first-order alternative to the existing QP/B2B analytical solvers, reusing the existing AP netlist, density manager, and legalization infrastructure. It explores a placer structure closer to modern smooth-wirelength + electrostatic-density formulations while remaining a drop-in `--ap_global_placer` option.

Full paired head-to-head vs the default `simpl` + `lp-b2b` AP flow over 69 circuits (VTR, Koios, Titan, Titanium suites, fixed channel widths). Geomean nonlinear-nesterov / b2b (< 1 = nesterov better):

| metric | ratio |
|---|---|
| routed wirelength | 0.986 |
| critical-path delay | 0.993 |
| total flow runtime | 0.964 |

Nesterov is at or ahead of b2b on all three axes at the geomean; the residual per-circuit CPD outliers are dominated by a single virtual-I/O-clock timing artifact (LU_Network), where core-logic timing is already at parity.

#### Types of changes

#### Checklist:


doc/src/vpr/command_line_usage.rst
```diff
@@ -1339,7 +1339,7 @@ When using a pre-computed flat placement file with the ``flat-recon`` full legal
     ``--analysis`` must be specified explicitly to run post-route timing analysis.
     It is not implied by ``--route``.
 
-.. option:: --ap_analytical_solver {identity | qp-hybrid | lp-b2b}
+.. option:: --ap_analytical_solver {identity | qp-hybrid | lp-b2b | nesterov}
```
AlexandreSinger: This option that you are modifying is to control the analytical solver used; you appear to be proposing a global_placer. I think that this is the wrong option to use. We do not appear to have an --ap_global_placer option yet (since we have never needed one before) but for this feature to be added I think you will need to add this option to select between the global placers.

I think the name `nesterov` is a bit generic and not clear about what this feature really is. Nesterov's method is an optimization method used to perform a form of gradient descent. I think what really differentiates this from the other analytical global placers is that it is non-linear. So maybe including non-linear in the name to something like `non-linear-nesterov` would be more accurate.

vpr/src/analytical_place/global_placement/global_placer.cpp
```diff
@@ -48,6 +49,24 @@ std::unique_ptr<GlobalPlacer> make_global_placer(e_ap_analytical_solver analytic
                                                  const std::vector<std::string>& target_density_arg_strs,
                                                  unsigned num_threads,
                                                  int log_verbosity) {
+    if (analytical_solver_type == e_ap_analytical_solver::Nesterov) {
```
AlexandreSinger: Same comment here, this should be the global placer type. Not solver.

doc/src/vpr/command_line_usage.rst
```diff
@@ -1363,6 +1363,10 @@ When using a pre-computed flat placement file with the ``flat-recon`` full legal
       Uses the legalized solution as anchor-points to pull the solution to a
       more legal solution (similar to the approach from SimPL :cite:`Kim2013_SimPL`).
 
+    * ``nesterov`` Runs an experimental global placer that applies accelerated
+      first-order updates to a smooth wirelength and density objective before
+      invoking the selected partial legalizer.
```
AlexandreSinger: I think this comment can be improved some. I think that once this is moved into an --ap_global_placer option it would make more sense; it is not clear why this is different from other global placers?

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
```diff
@@ -0,0 +1,649 @@
+/**
+ * @file
+ * @author  William Zhang
+ * @date    June 2026
+ * @brief   Implementation of a Nesterov-style analytical global placer.
```
AlexandreSinger: This file has a lot of different moving parts. I recommend a high-level comment at the top of this file to explain what the algorithm is doing and what it is optimizing for.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+ */
+constexpr double kMinStepSize = 1e-6;
+
+/**
+ * @brief Maximum fraction of the device span a block should move in one step.
+ */
+constexpr double kInitialStepSpanFraction = 0.02;
+
+/**
+ * @brief Smooth wirelength gamma as a fraction of the larger device dimension.
```
AlexandreSinger, line 110: When the gamma value is large, the wirelength estimate is quite poor, but the function is quite smooth. What I have seen different works do is that they iteratively increase gamma during the solve to approach a more accurate wirelength estimate. This gamma appears to be fixed. I am interested to see how this affects the wirelength on larger circuits.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+/**
+ * @brief Device-bound epsilon to keep floor-based bin lookup inside the grid.
+ */
+constexpr double kDeviceBoundaryEpsilon = 1e-6;
+
+/**
+ * @brief Numerically stable log(sum(exp(value / gamma))).
+ */
+static double stable_log_sum_exp(const std::vector<double>& values, double gamma, bool negate) {
```
AlexandreSinger: This is a static method inside of the anonymous namespace. I think it does not need to be static. The static is redundant.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+    VTR_ASSERT(!values.empty());
+
+    double max_scaled = -std::numeric_limits<double>::infinity();
```
AlexandreSinger: Negative infinities make me a bit nervous. Since you are asserting that the values are non-empty, why not use the first element in values?

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+ */
+constexpr double kEpsilon = 1e-9;
+
+/**
+ * @brief Device-bound epsilon to keep floor-based bin lookup inside the grid.
+ */
+constexpr double kDeviceBoundaryEpsilon = 1e-6;
+
+/**
+ * @brief Numerically stable log(sum(exp(value / gamma))).
```
AlexandreSinger: I think it may be worth explaining what each parameter is and what it means. You and I know what gamma means in this function, but a regular passer-by may not.

Also why is negate needed? I think this function needs more context.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+    : GlobalPlacer(ap_netlist, log_verbosity)
+    , atom_netlist_(atom_netlist)
+    , pre_cluster_timing_manager_(pre_cluster_timing_manager)
+    , place_delay_model_(place_delay_model)
+    , net_weights_(ap_netlist.nets().size(), 1.0)
+    , device_grid_width_(device_grid.width())
+    , device_grid_height_(device_grid.height())
+    , device_grid_num_layers_(device_grid.get_num_layers())
+    , ap_timing_tradeoff_(ap_timing_tradeoff) {
+    vtr::ScopedStartFinishTimer nesterov_placer_building_timer("Constructing Nesterov Global Placer");
```
AlexandreSinger: Yeah, I still think Nesterov is not a very clear name. I think the word Non-Linear is needed somewhere in here.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+        density_manager_->generate_mass_report();
+
+    partial_legalizer_ = make_partial_legalizer(partial_legalizer_type,
+                                                ap_netlist_,
+                                                density_manager_,
+                                                prepacker,
+                                                models,
+                                                log_verbosity_);
+
+    optimizable_blocks_.reserve(ap_netlist_.blocks().size());
```
AlexandreSinger: I think a comment is needed to explain what this chunk of code is doing. What does it mean to be optimizable in this context? Why do we need to pre-compute these blocks.

It looks like you are just keeping track of the moveable blocks similar to what I did in the solver. Good comments can help make this clear.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+    for (APBlockId blk_id : ap_netlist_.blocks()) {
+        if (block_is_optimizable_(blk_id))
+            optimizable_blocks_.push_back(blk_id);
+    }
+}
+
+bool NesterovGlobalPlacer::block_is_optimizable_(APBlockId blk_id) const {
+    return ap_netlist_.block_mobility(blk_id) == APBlockMobility::MOVEABLE;
```
AlexandreSinger, line 563: I think a small comment saying why moveable blocks are optimizable in this context can be helpful. If this is the case though, why not just call them `moveable_blocks`?

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+}
+
+PartialPlacement NesterovGlobalPlacer::initialize_placement_() const {
+    PartialPlacement p_placement(ap_netlist_);
+
+    if (optimizable_blocks_.empty()) {
```
AlexandreSinger: What is this if statement doing? If there are no moveable blocks, that just means that all the blocks are fixed right?

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+        size_t layer = (idx / num_blocks_per_layer) % num_layers;
+        size_t layer_idx = idx % num_blocks_per_layer;
+        size_t row = layer_idx / cols;
+        size_t col = layer_idx % cols;
+
+        p_placement.block_x_locs[blk_id] = std::min<double>((col + 0.5) * gap, device_grid_width_ - kDeviceBoundaryEpsilon);
+        p_placement.block_y_locs[blk_id] = std::min<double>((row + 0.5) * gap, device_grid_height_ - kDeviceBoundaryEpsilon);
+        p_placement.block_layer_nums[blk_id] = layer;
+        p_placement.block_sub_tiles[blk_id] = 0;
+    }
```
AlexandreSinger: This looks very similar to the random uniform initial placement from the solver. If they are exactly the same, it may be a good idea to reuse that code. If not, there should be a clear comment explaining what this code is doing and how is it different from the other implementation.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+    project_placement_(p_placement);
```
AlexandreSinger: I am a bit confused why this projection is needed here? I think a comment is needed.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+    }
+
+    if (pre_cluster_timing_manager_.get_timing_update_type() == e_timing_update_type::INCREMENTAL) {
+        for (tatum::EdgeId edge : pre_cluster_timing_manager_.get_timing_info().timing_graph()->edges()) {
+            pre_cluster_timing_manager_.get_timing_info_ptr()->invalidate_delay(edge);
+        }
+    }
+
+    pre_cluster_timing_manager_.update_timing_info();
+    pre_cluster_timing_manager_.get_timing_info_ptr()->set_warn_unconstrained(false);
```
AlexandreSinger: This method that is updating the delay information appears to be directly copied from the SimPL global placer. We want to avoid duplicated code if we can, especially for this function since I am still working on exactly how this works.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+    VTR_LOG("Nesterov Global Placer Statistics:\n");
+    VTR_LOG("\tSmooth placement HPWL before partial legalization: %g\n", current.get_hpwl(ap_netlist_));
+    VTR_LOG("\tFinal density weight: %g\n", density_weight);
+
+    partial_legalizer_->legalize(current);
+    VTR_LOG("\tPlacement HPWL after partial legalization: %g\n", current.get_hpwl(ap_netlist_));
+    partial_legalizer_->print_statistics();
+
+    return current;
```
AlexandreSinger: This function desperately needs comments. Its a bit hard to follow.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+class DeviceGrid;
+class LogicalModels;
+class PlaceDelayModel;
+class PreClusterTimingManager;
+class Prepacker;
+struct t_logical_block_type;
+struct t_physical_tile_type;
+
+/**
+ * @brief Analytical global placer using accelerated first-order updates.
```
AlexandreSinger, line 38: As I mentioned before, this comment needs to be updated with more information that differentiates it from other global placers.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+                         PreClusterTimingManager& pre_cluster_timing_manager,
+                         std::shared_ptr<PlaceDelayModel> place_delay_model,
+                         float ap_timing_tradeoff,
+                         bool generate_mass_report,
+                         const std::vector<std::string>& target_density_arg_strs,
+                         e_ap_partial_legalizer partial_legalizer_type,
+                         int log_verbosity);
+
+    /**
+     * @brief Run accelerated smooth global placement and final partial legalization.
```
AlexandreSinger: What does accelerated mean in this context? It is not very clear. To me, it sounds like you are saying accelerated gradient descent; but the way it is written makes it sound like the placement itself is accelerated, which sounds odd.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+     * @brief Initialize all block locations before first-order optimization.
+     */
+    PartialPlacement initialize_placement_() const;
+
+    /**
+     * @brief Evaluate the smooth objective, optionally accumulating gradients.
+     */
+    ObjectiveValue evaluate_objective_(const PartialPlacement& p_placement,
+                                       double density_weight,
+                                       PlacementGradient* grad) const;
```
AlexandreSinger: We are trying to move away from pointers in VTR since they can be very dangerous to work with. Can you please use std::optional instead?

AlexandreSinger: Same comment for other usages of these types of pointers.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @brief Add smooth weighted-average wirelength value and gradient.
+     */
+    double add_wirelength_gradient_(const PartialPlacement& p_placement,
+                                    PlacementGradient* grad) const;
+
+    /**
+     * @brief Update smooth wirelength net weights from pre-cluster timing criticalities.
```
AlexandreSinger: This description I think can be improved for clarity. What is being updated here and why?

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+     */
+    void update_timing_net_weights_();
+
+    /**
+     * @brief Update pre-cluster timing delays and criticalities from a placement.
```
AlexandreSinger: Same comment here. How timing interracts with these flows is usually very important.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+     */
+    void update_timing_info_with_placement_(const PartialPlacement& p_placement);
+
+    /**
+     * @brief Estimate one tile of delay for missing delay-model entries.
+     */
+    float delay_per_tile_() const;
+
+    /**
+     * @brief Add smooth bin-density value and gradient.
```
AlexandreSinger: This is an incredibly important function that needs a clear comment explaining what it is doing. The most important part of any global placer is its model of density. This needs to be very clearly defined.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @brief Project all block locations into device bounds and restore fixed blocks.
+     */
+    void project_placement_(PartialPlacement& p_placement) const;
+
+    /**
+     * @brief Copy a placement.
+     */
+    void copy_placement_(const PartialPlacement& src, PartialPlacement& dst) const;
```
AlexandreSinger: Why is this method needed? The partial placer should have a copy constructor?

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+     * @brief Compute the squared norm of a placement gradient.
+     */
+    double gradient_norm_squared_(const PlacementGradient& grad) const;
+
+    /**
+     * @brief Return true if the block should be moved by the continuous optimizer.
+     */
+    bool block_is_optimizable_(APBlockId blk_id) const;
+
+    std::shared_ptr<FlatPlacementDensityManager> density_manager_; ///< Architecture-aware density metadata.
```
AlexandreSinger: I think this value comment could be improved. It should be clear what the density manager object is doing and why it is in this class.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+    std::unique_ptr<PartialLegalizer> partial_legalizer_;          ///< Existing partial legalizer used after smooth optimization.
```
AlexandreSinger: I would avoid the word "existing" here. It is a partial legalizer. You should make it clear why it is used. Why do we need the partial legalizer at the end?

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
```diff
@@ -0,0 +1,175 @@
...
+    const AtomNetlist& atom_netlist_;                              ///< Atom netlist used for timing criticality lookup.
+    PreClusterTimingManager& pre_cluster_timing_manager_;          ///< Timing manager used to weight smooth wirelength nets.
```
AlexandreSinger: What is "smooth wirelength nets", the term is not well defined.

AlexandreSinger: The term "smooth wirelength" is used throughout this file, but I am not quite sure what that means. I think this needs to be better defined or a better word used.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+    double nesterov_t = 1.0;
+    ObjectiveValue current_obj = evaluate_objective_(current, density_weight, nullptr);
+
+    if (log_verbosity_ >= 1) {
+        VTR_LOG("----  ------------  --------------  ------------  ------------  -----------\n");
+        VTR_LOG("Iter  Smooth Obj    Smooth WL       Density       HPWL          Step\n");
+        VTR_LOG("----  ------------  --------------  ------------  ------------  -----------\n");
+    }
+
+    for (size_t iter = 0; iter < kMaxNesterovIterations; iter++) {
```
AlexandreSinger: This max number of iterations is the only exit condition for the optimization loop. This is a bit concerning to me since we run VPR on a large range of circuits. Some are a small as 10 elements, and some are as large as 1 billion elements. I would be surprised if 200 is always the correct value. I think this loop needs an exit condition that "exits once the steps are small enough"; basically, stop when we are not improving the circuit anymore.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+double NesterovGlobalPlacer::add_wirelength_gradient_(const PartialPlacement& p_placement,
+                                                      PlacementGradient* grad) const {
+    double gamma = std::max(1.0, std::max<double>(device_grid_width_, device_grid_height_) * kWirelengthGammaFraction);
+    double smooth_wirelength = 0.;
+
+    std::vector<double> x_locs;
+    std::vector<double> y_locs;
+    x_locs.reserve(32);
+    y_locs.reserve(32);
```
AlexandreSinger: Magic number. What does 32 mean in this context? Why are we reserving these vectors like this?

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+    ObjectiveValue value;
+    value.wirelength = add_wirelength_gradient_(p_placement, grad);
+    value.density = add_density_gradient_(p_placement, density_weight, grad);
+    value.total = value.wirelength + density_weight * value.density;
+    return value;
+}
+
+double NesterovGlobalPlacer::add_wirelength_gradient_(const PartialPlacement& p_placement,
+                                                      PlacementGradient* grad) const {
+    double gamma = std::max(1.0, std::max<double>(device_grid_width_, device_grid_height_) * kWirelengthGammaFraction);
```
AlexandreSinger: What is the relationship between gamma and the device dimensions? That should be clearly defined. I can see a relationship, but this should be made clear.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
AlexandreSinger: Also why are these variables hoisted out of this function? The curse of early optimizations is bugs. The compiler is smarter than you think. I do not recommend hoisting these vectors unless performance is found to be an issue (and you profiled to show that this is the problem). We want nice clean code.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+                double pos_y_grad = std::exp(y_locs[pin_idx] / gamma - log_pos_y) / sum_pos_y;
+                double neg_y_grad = std::exp(-y_locs[pin_idx] / gamma - log_neg_y) / sum_neg_y;
+                grad->dx[blk_id] += net_weight * (pos_x_grad - neg_x_grad);
+                grad->dy[blk_id] += net_weight * (pos_y_grad - neg_y_grad);
+            }
+            pin_idx++;
+        }
+    }
+
+    return smooth_wirelength;
```
AlexandreSinger, line 1370: This method needs function comments. It is hard to follow.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+}
+
+void NesterovGlobalPlacer::update_timing_net_weights_() {
+    std::fill(net_weights_.begin(), net_weights_.end(), 1.0);
+
+    if (ap_timing_tradeoff_ == 0.f)
+        return;
+
+    if (!pre_cluster_timing_manager_.is_valid()) {
+        VTR_LOG_WARN("Nesterov analytical placement requested timing tradeoff %g, but pre-cluster timing is unavailable; using unit net weights.\n",
```
AlexandreSinger: I do not think this warning is needed. Some architectures do not have timing data, and the timing tradeoff is defaulted to 0.5.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+    for (APNetId net_id : ap_netlist_.nets()) {
+        if (ap_netlist_.net_is_ignored(net_id))
+            continue;
+
+        AtomNetId atom_net_id = ap_netlist_.net_atom_net(net_id);
+        VTR_ASSERT_SAFE(atom_net_id.is_valid());
+
+        double crit = pre_cluster_timing_manager_.calc_net_setup_criticality(atom_net_id, atom_netlist_);
+        double weight = ap_timing_tradeoff_ * crit + (1.0 - ap_timing_tradeoff_);
+        net_weights_[net_id] = weight;
```
AlexandreSinger, line 1400: This should be commented. This looks similar to my net weighting from the QP_Hybrid analytical solver. It is not obvious why this weight should be used.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+                density_derivative += bin_util_derivative[contribution.bin_id].get_dim_val(dim)
+                                      * contribution.block_mass.get_dim_val(dim);
+            }
+            density_derivative *= density_weight;
+            grad->dx[contribution.blk_id] += density_derivative * contribution.dweight_dx;
+            grad->dy[contribution.blk_id] += density_derivative * contribution.dweight_dy;
+        }
+    }
+
+    return density_penalty;
```
AlexandreSinger: This method needs comments. This is the most important function in this whole algorithm. This is what makes or breaks AP global placers.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
```diff
@@ -0,0 +1,649 @@
...
+            wx[1] = 0.0;
+            dwx[0] = 0.0;
+            dwx[1] = 0.0;
+        }
+        if (y0 == y1) {
+            wy[0] = 1.0;
+            wy[1] = 0.0;
+            dwy[0] = 0.0;
+            dwy[1] = 0.0;
+        }
```
AlexandreSinger: I am not quite following this formulation. This would work fine for homogenous architectures (where everything is exactly the same tile type); but I think this would fail to work properly on heterogenous architectures.

vpr/src/analytical_place/ap_flow_enums.h
```diff
@@ -16,7 +16,8 @@
 enum class e_ap_analytical_solver {
     Identity,  ///< The identity analytical solver, which does not perform any optimizations on the placement. Used as a placeholder when solving should not be used.
     QP_Hybrid, ///< Analytical Solver which uses the hybrid net model to optimize the quadratic HPWL objective.
-    LP_B2B     ///< Analytical Solver which uses the B2B net model to optimize the linear HPWL objective.
+    LP_B2B,    ///< Analytical Solver which uses the B2B net model to optimize the linear HPWL objective.
+    Nesterov   ///< Global placer which uses accelerated first-order updates on smooth wirelength and density objectives.
```
AlexandreSinger: Same comment as above. This is not a solver. This is a global placer. It should have its own enum.

vpr/src/base/read_options.cpp
```diff
@@ -2058,7 +2063,8 @@ argparse::ArgumentParser create_arg_parser(const std::string& prog_name, t_optio
             "Controls which Analytical Solver the Global Placer will use in the AP Flow.\n"
             " * identity: Does not formulate any equations and just passes the last legalized solution through. This solver is only used for testing and debugging.\n"
             " * qp-hybrid: Solves for a placement that minimizes the quadratic HPWL of the flat placement using a hybrid clique/star net model.\n"
-            " * lp-b2b: Solves for a placement that minimizes the linear HPWL of theflat placement using the Bound2Bound net model.")
+            " * lp-b2b: Solves for a placement that minimizes the linear HPWL of the flat placement using the Bound2Bound net model.\n"
+            " * nesterov: Uses accelerated first-order updates to optimize smooth wirelength and density objectives.")
```
AlexandreSinger: Same comment here.

AlexandreSinger (changes requested):
Looks very interesting @WilliamZhang20 !

There are a few things I want to see in order to bring this in.

The most important thing is documentation. I am really not following the algorithm as it is written. I can see and understand the log-sum-exp model used to estimate the smooth-wirelength object, but the way you are modeling density is not clear to me. The overall optimization algorithm is also not very clear to me. I am familiar with Nesterov's method, but the exit condition for the overall algorithm is not very well defined. Most importantly, I would like to see clear documentation on the algorithm that you are using for modeling density. Modelling density in heterogenous FPGAs (which is the common target of VTR) is the most difficult part of AP. This needs to be documented so it can continue to be built on in the future.

The next important thing I need to see is if this algorithm is working and is scaling on more-modern heterogenous FPGAs and benchmarks. The results you mentioned in your comment are from 4 very small circuits from MCNC (which is a benchmark we do not use for quality of results since they are too simple). I am interested to see how the results look on the VTR benchmarks: https://docs.verilogtorouting.org/en/latest/vtr/benchmarks/#vtr-benchmarks

A task list for the VTR benchmarks can be found here: https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/dcaa711dcc615ae999ef37003637d4b3f4148616/vtr_flow/tasks/regression_tests/vtr_reg_nightly_test3/vtr_reg_qor_chain/config/config.txt

I would like to see the wirelength, CPD, and runtime for these circuits and how they compare to the default flow on your machine. This will give us a better understanding of the algorithm.

Feel free to reach out if you have any questions.

AlexandreSinger:
@vaughnbetz FYI

doc/src/vpr/command_line_usage.rst
WilliamZhang20: Thanks, changed.

vpr/src/analytical_place/ap_flow_enums.h
WilliamZhang20: Thanks, fixed

vpr/src/base/read_options.cpp
WilliamZhang20: Fixed, thanks.

vaughnbetz:
Thanks @WilliamZhang20 . To build on what Alex said, I'd want to see more extensive QoR data before merging into master. In addition to the VTR designs, you should run on the Titan designs so you are checking on more than one architecture, and because the Titan designs are larger.  Another important suite is Koios, but results on two large suites would be enough to get a good sense of where the algorithm stands and to tune it for now.

You can find the steps to run Titan and check QoR at https://docs.verilogtorouting.org/en/latest/README.developers/#example-titan-benchmarks-qor-measurement . Reading the whole QoR section (just before that) would be a good idea too, to get a sense of the data we watch closely.

WilliamZhang20:
Latest algorithmic changes include:
- Electrostatic density formulation + DCT Poisson Solve
- Dynamic Fillers + DC Removal
- B2B Warm Start Cycles

QoR results as of July 3, 2026 (TLDR: WL slightly better on all benchmarks; CPD and runtime are the same or slightly worse)
[geomean_ratios.csv](https://github.com/user-attachments/files/29637556/geomean_ratios.csv)
[ap_headtohead_vtr.xlsx](https://github.com/user-attachments/files/29637558/ap_headtohead_vtr.xlsx)
[ap_headtohead_titanium.xlsx](https://github.com/user-attachments/files/29637559/ap_headtohead_titanium.xlsx)
[ap_headtohead_titan.xlsx](https://github.com/user-attachments/files/29637560/ap_headtohead_titan.xlsx)
[ap_headtohead_koios.xlsx](https://github.com/user-attachments/files/29637561/ap_headtohead_koios.xlsx)
[ap_headtohead_ALL.xlsx](https://github.com/user-attachments/files/29637564/ap_headtohead_ALL.xlsx)
[per_circuit_ratios.csv](https://github.com/user-attachments/files/29637568/per_circuit_ratios.csv)

vaughnbetz:
Thanks @WilliamZhang20 . Are crit_path_delay and total_wirelength both post routing?

doc/src/vpr/command_line_usage.rst
WilliamZhang20: Addressed — the `--ap_global_placer` doc now explicitly contrasts `nonlinear-nesterov` (accelerated first-order updates on a smooth wirelength+density objective) with `simpl`'s solver/legalizer alternation. See doc/src/vpr/command_line_usage.rst:1390-1401.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
WilliamZhang20: The file was renamed to nonlinear_nesterov_placer.cpp/.h and the class now carries a Doxygen block describing the algorithm (weighted-average wirelength + electrostatic density penalty, followed by partial legalization) — see nonlinear_nesterov_placer.h:29-36.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed — gamma is no longer fixed; current_gamma_fraction_ now anneals from kGammaStartFraction to kGammaEndFraction across epochs (gamma continuation).

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Fixed — the static qualifier was dropped from the anonymous-namespace helper.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Fixed — the running max is now seeded from values.front() instead of -infinity, per your suggestion.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed — the function now has a full Doxygen block explaining values, gamma, and negate (including why negation is needed to get a smooth min out of the same max-formula).

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
WilliamZhang20: Renamed — the class/file are now NonlinearNesterovPlacer / nonlinear_nesterov_placer.{cpp,h}, and the CLI option value is nonlinear-nesterov.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: block_is_optimizable_ now has a doc comment ("Return true if the block should be moved by the continuous optimizer") and optimizable_blocks_ has a member comment in the header explaining it holds the movable AP blocks touched by the optimizer.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: This duplicate path is gone — initialize_placement_() now always warm-starts from the B2B/QP solver (with a sparsity-gated deepening loop); there's no separate random-uniform placement branch anymore.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed — project_placement_/project_fillers_ now have Doxygen comments explaining they clamp block/filler locations into device bounds and restore fixed blocks.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed — this was factored out into a shared update_timing_info_with_partial_placement() helper in global_placer.cpp/.h, used by both the SimPL flow and this placer, so the logic isn't copy-pasted anymore.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed as part of the broader documentation pass — the file now has Doxygen/inline comments throughout the density, wirelength, and gradient functions.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Addressed — the class comment and the --ap_global_placer CLI doc now both explain how this differs from other global placers (accelerated first-order smooth optimization vs. SimPL's solver/legalizer alternation).

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Addressed — updated to "Update smooth wirelength net weights from pre-cluster timing criticalities," which states both what's updated and why.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Addressed — ap_timing_tradeoff_/effective_timing_tradeoff_ now have header comments, and the size-based adaptive-tradeoff logic (kAdaptiveTimingMinBlocks/MaxBlocks/Tradeoff) is documented where it's used in the .cpp.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Addressed in commit d00ce21e2a ("document electrostatic density formulation") — add_density_gradient_ now has inline comments for the utilization/target-capacity/potential/field arrays, the Poisson-Jacobi solve, and the charge-deposition step.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: This method no longer exists — there's no custom copy path for PartialPlacement in the current code, it looks like it was removed.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
WilliamZhang20: Addressed — the epoch loop now has a displacement-based convergence exit (convergence_displacement, gated by kMinNesterovIterationsPerEpoch) in addition to the iteration cap, so it can stop early once steps are small enough. The cap itself was also lowered from 200 to kMaxNesterovIterations = 80.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed — there's no reserve(32) or similar magic-number reservation in the current file, and the working vectors in add_wirelength_gradient_ are declared locally per-net rather than hoisted out of the function.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed — kWirelengthGammaFraction's doc comment and the inline comment in add_wirelength_gradient_ now state that gamma scales with the device span so smoothing stays grid-relative.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Addressed — add_wirelength_gradient_ now has a full Doxygen/inline comment block explaining the weighted-average wirelength formulation.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
WilliamZhang20: Addressed — optimize_from_seed_ (the epoch loop) now has an inline comment describing the accelerated-gradient-descent / density-continuation structure.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Yes, has been documented. Thanks.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Annotated with Nesterov Accelerated Gradient, thanks.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Thanks, clarified it to say it owns the per-tile capacity/mass model.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Thanks, added a detail to say the partial legalizer resolves discrete architecture-legality violations after iterations. Having a partial legalizer at the end adds another layer of guarantee of solution validity before packing.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Thanks, fixed.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.h
WilliamZhang20: Clarified smoothness for wire length and other terms.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Renamed, thanks.

vpr/src/analytical_place/global_placement/nesterov_global_placer.cpp
WilliamZhang20: Removed it, thanks.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: The formulation has been modified to account for heterogeneity.

vpr/src/analytical_place/global_placement/nonlinear_nesterov_placer.cpp
WilliamZhang20: Thanks, commented at lines 1307–1315.

WilliamZhang20:
All of Alex's comments have been resolved. Some algorithmic adjustments have resulted in better-than-baseline QoR and runtime geometric mean across all benchmarks tested. 

However, some outliers are likely an obstacle to further improvement without extending the scope, `LU_Network` being one of the worst (apparently due to I/O Chains). However, for those outliers and other examples, the WL and CPD are actually better than B2B after full legalization, but the gains get eroded by packing and/or detailed placement.

Overall Table (ratios wrt B2B): - updated 7/10/2026
| Benchmark suite | Circuits | Routed WL | Critical-path delay | Total VPR runtime | AP runtime | Pack + place runtime | Routing runtime |
|-----------------|:--------:|:---------:|:-------------------:|:-----------------:|:----------:|:------------:|:-------:|
| VTR             | 8        | 0.990     | 1.009               | 0.980             | 0.969      | 1.088        | 1.065   |
| Koios           | 29       | 0.994     | 0.987               | 0.975             | 0.968      | 1.022        | 1.028   |
| Titan           | 22       | 0.993     | 1.020               | 0.947             | 0.941      | 1.031        | 0.999   |
| Titanium        | 16       | 0.985     | 0.990               | 0.966             | 0.963      | 1.020        | 0.978   |
| **All**         | **75**   | **0.991** | **0.999**           | **0.965**         | **0.959**  | **1.031**    | **1.012** |

Omitted Benchmark Circuits:
- From Titan:
  - gaussianblur
- Koios:
  - bwave_like.fixed.large (apparently hit a VPR error on both B2B and nesterov)
 
Post Full-Legalization Ratios: (out of date, from 7/9/2026)
[titan_cpd_erosion.xlsx](https://github.com/user-attachments/files/29852250/titan_cpd_erosion.xlsx)

Raw Data: (out of date, from 7/9/2026)
[ap_headtohead_vtr.xlsx](https://github.com/user-attachments/files/29852241/ap_headtohead_vtr.xlsx)
[ap_headtohead_titanium.xlsx](https://github.com/user-attachments/files/29852242/ap_headtohead_titanium.xlsx)
[ap_headtohead_titan.xlsx](https://github.com/user-attachments/files/29852243/ap_headtohead_titan.xlsx)
[ap_headtohead_koios.xlsx](https://github.com/user-attachments/files/29852244/ap_headtohead_koios.xlsx)
[ap_headtohead_ALL.xlsx](https://github.com/user-attachments/files/29852245/ap_headtohead_ALL.xlsx)

vaughnbetz:
Thanks William. Can you also add the runtime ratios to the summary (e.g. total flow, ap, total pack + place).VaughnOn Jul 9, 2026, at 10:24 AM, William Zhang ***@***.***> wrote:﻿WilliamZhang20 left a comment (verilog-to-routing/vtr-verilog-to-routing#3641)
All of Alex's comments have been resolved. Some algorithmic adjustments have resulted in better than baseline QoR and runtime geometric mean across all benchmarks tested.
However, some outliers are likely an obstacle to further improvement without extending the scope, LU_Network being one of the worst (apparently due to I/O Chains). However, for those outliers and other examples, the WL and CPD are actually better than B2B after full legalization, but the gains get eroded by packing and/or detailed placement.
Overall Table:



Benchmark suite
Circuits
Routed WL
Critical-path delay




VTR
8
0.992
1.029


Koios
24
0.979
0.963


Titan
21
0.993
1.018


Titanium
16
0.985
0.990


All
69
0.986
0.993



Post Full-Legalization Ratios:
titan_cpd_erosion.xlsx
Raw Data:
ap_headtohead_vtr.xlsx
ap_headtohead_titanium.xlsx
ap_headtohead_titan.xlsx
ap_headtohead_koios.xlsx
ap_headtohead_ALL.xlsx

AlexandreSinger:
@WilliamZhang20 Why were those circuits ommited? Many of those are some of the larger circuits. You ommited 6 of the around 30 koios benchmarks. Ommiting sparcT1_chip2 seems a bit random to drop out?

Also in you major summary, I am interested to see the post-GP WL and CPD estimates for the different benchmarks. I want to know if the gains are coming from the global placer or if they are coming from the interaction with the full-legalizer and detailed placer.

Edit: Can you also report the device width. This is just a curiosity on my part. If you are not fixing the device widths, you may get different device widths; making the two algorithms less comparable. The device width is calculated from the packing (full legalization solution), so it can change based on the global placement.

WilliamZhang20:
Added runtime ratios to the previous comment above. A table showing where the gains come from is below via ratios wrt. B2B. For CPD (the weak point), I see that gains come from full legalization and are weakened after detailed placement & routing.
Fixing the device width ended up changing VTR, as reflected below. The table below also accounts for sparcT1_chip2.

### Wirelength (HPWL est. through post_dp, routed WL at final)

| Suite | N | post_gp | post_fl | post_dp | final (routed) |
|---|:--:|:--:|:--:|:--:|:--:|
| VTR | 8 | 1.010 | 0.957 | 0.990 | 0.990 |
| Koios | 24 | 1.000 | 0.982 | 0.993 | 0.994 |
| Titan | 22 | 1.041 | 0.987 | 0.991 | 0.993 |
| Titanium | 16 | 1.038 | 0.988 | 0.984 | 0.984 |
| **All** | 70 | 1.023 | 0.982 | 0.990 | 0.991 |

### Critical-path delay (estimated through post_dp, routed at final)

| Suite | N | post_gp | post_fl | post_dp | final (routed) |
|---|:--:|:--:|:--:|:--:|:--:|
| VTR | 8 | 0.978 | 1.024 | 1.006 | 1.009 |
| Koios | 24 | 1.001 | 0.972 | 0.982 | 0.979 |
| Titan | 22 | 1.001 | 0.985 | 1.016 | 1.020 |
| Titanium | 16 | 1.021 | 0.958 | 0.965 | 0.990 |
| **All** | 70 | 1.003 | 0.979 | 0.991 | 0.997 |

vaughnbetz:
Thanks. Looks like it might be a 5% faster AP stage with roughly the same QoR.

WilliamZhang20:
@AlexandreSinger @vaughnbetz Thank you. What would be the conditions for merging this PR? I believe that modifying later stages to account for the downstream degradation of QoR would be done in a separate PR.

WilliamZhang20:
Update: recently implemented [ADMM](https://stanford.edu/~boyd/admm.html), which I believe enables better constraint handling with the descent algorithm. 
Results are described below. Also implemented testing to make sure the gradient/objective was exact, hence the large line diff. Let me know if they should be deleted.

Data (from 4 seeds):
[full75_4seed_20260806.csv](https://github.com/user-attachments/files/30809877/full75_4seed_20260806.csv)

TLDR table:
Metric | 4-seed Geomean | 95% Confidence Interval | Circuits Better
-- | -- | -- | --
Routed wirelength | 0.9910 | [0.9842, 0.9977] | 48 / 75
Routed critical-path delay | 0.9991 | [0.9838, 1.0180] | 37 / 75
Flow runtime | 0.9474 | [0.9340, 0.9605] | 59 / 75

CI fixes are pending

vaughnbetz:
Capturing our discussion from today:
@WilliamZhang20 has some new QoR improvements, which he believes are mostly from increasing the number of global solve - partial legalize iterations.  He will post the results as time permits.
One question I have: could we get those improvements by increasing the number of solve - legalize iterations in the current analytical formulation?  That would be less new code, so preferred if it works. @WilliamZhang20 : if you can compare to that it would be good.

Code size / quality: to merge, we'll need a significant (to be defined, but multiple % in WL or CPD or both) QoR improvement. Against the QoR we'll evaluate the code maintainability. You don't need to worry about that while experimenting, but to merge code it has to be commented, architecture-adaptive, well understood and ideally as small as possible.  So the experiments / features that don't work out would be deleted, and any claude experiments etc. would have to be removed from the code or made fully readable.

WilliamZhang20:
Update:
- Regarding the push for more solve + legalize iterations on the b2b analytical formulation, it failed to obtain meaningful improvement.
- The improvement that is driven by this PR is a combination of increased iterations and smarter iterations, e.g. Barzilai-Borwein / inverse-Lipschitz during the descent. The ADMM (Alternating Direction Method of Multipliers) is a way to break down a large optimization problem into smaller, easier problems, and can be considered a mathematical descendant of Dual Ascent, Operator Splitting, and Augmented Lagrangian. Justification is very proof/theory-heavy, but that's for another day.
- Unfortunately, as past research (e.g. elfPlace) has shown regarding electrostatic global placement, CPD cannot be driven very far without sacrificing our WL and runtime improvements (one attempt got both WL and CPD better but at 2-3x runtime by repeated flow rerun on a different seed), and that has once again been observed despite my attempts to push on it. Packing/DP are simply the proper methods to find the real critical path and optimize for it.

**Results** table from 4 seeds (75 VTR (8), Koios, Titan, and Titanium - same as above):
| metric | nesterov / b2b | median | better on |
|---|---:|---:|---:|
| routed wirelength | **0.96102** | 0.97320 | 62/75 |
| routed CPD | 0.99756 | 1.00667 | 34/75 |
| flow runtime | **0.94428** | 0.94231 | 64/75 |

**Caveat.** `--ap_timing_tradeoff 0` is a command-line flag, not the default. Without it the same
build returns routed wirelength **0.96582**; the 0.48% difference is that flag alone.

Raw data from 4 seeds:
[shipping_4seed_vs_b2b.csv](https://github.com/user-attachments/files/31164848/shipping_4seed_vs_b2b.csv)

Please feel free to give feedback on diff shortening/maintainability. There are no more experiment scaffolds/dead code present in this PR. Every line diff is fully dedicated to the placer. So far, I have attempted to enhance readability by breaking it down into multiple files, let me know if that should be undone.

WilliamZhang20:
Reduced code diff from 4500 to 3900. This was in part due to running numerous ablation tests that allowed me to cut out functionally redundant components, keeping only what is load-bearing. QoR results remain roughly the same: 0.95 WL, parity CPD, runtime 0.938

Raw data below:
[comparison_09_04.csv](https://github.com/user-attachments/files/31871794/comparison_09_04.csv)

PR #3644, opened 2026-06-19
Create virtual python environment in docker image

AmirhosseinPoolad:
The current docker image tries to install python packages system wide. Currently this is resulting in a CI error. This PR changes the dockerfile to create a venv, install the packages in the venv and add the venv activation command to interactive shell startup.


AlexandreSinger:
I like this change a lot! Thanks Amir!

AmirhosseinPoolad:
CI is passing now.

AlexandreSinger (approved):
LGTM, I like this solution. I agree with using a virtual enviornment for this, and putting it in bashrc was a good idea.

PR #3645, opened 2026-06-20
[Scripts] Fixed Missing sys.exit in run_vtr_task

### LINKED ISSUE #3598: run_vtr_task.py not using sys.exit properly
While reviewing a PR, I noticed that run_vtr_task.py was not exiting on failure like you would expect it to. From looking at the file, I found the following.

When this script is run as the target script (i.e. `python3 run_vtr_task.py ...`), it runs the following:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/29a4c9721c4d6fcf7ce3b7f6f749c73073ade4ac/vtr_flow/scripts/run_vtr_task.py#L509-L510

Within the `vtr_command_main` function, I found the following if statement:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/29a4c9721c4d6fcf7ce3b7f6f749c73073ade4ac/vtr_flow/scripts/run_vtr_task.py#L289-L291

It looks like there may have been a mistake when this script was being made.

For context, `run_vtr_flow.py` has the following, which looks more correct to me:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/29a4c9721c4d6fcf7ce3b7f6f749c73073ade4ac/vtr_flow/scripts/run_vtr_flow.py#L853-L854

I think the `run_vtr_task.py` script should be updated to be similar to the `run_vtr_flow.py` script.

AlexandreSinger:
run_vtr_task.py had a malformed if statement in its main function where it was supposed to exit the program with an exit code if the script was executed directly, but it was never doing so. Resolved it by rewriting it to match run_vtr_flow.py

Resolves #3598

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -6,6 +6,7 @@
     "readline",
     "openssl",
     "pkgconf",
-    "capnproto"
+    "capnproto",
+    "eigen3"
   ]
 }
--- a/vtr_flow/scripts/run_vtr_task.py
+++ b/vtr_flow/scripts/run_vtr_task.py
@@ -286,8 +286,6 @@ def vtr_command_main(arg_list, prog=None) -> int:
             print("\tfile: ", exception.filename)
     except VtrError as exception:
         print("Error:", exception.msg)
-    if __name__ == "main":
-        sys.exit(num_failed)
     return num_failed
 
 
@@ -507,4 +505,4 @@ def run_vtr_flow_process(queue, run_dirs, job, script) -> None:
 
 
 if __name__ == "__main__":
-    vtr_command_main(sys.argv[1:])
+    sys.exit(vtr_command_main(sys.argv[1:], prog=sys.argv[0]))


AlexandreSinger:
Oh boy, it looks like it is failing the exact test I made this change for...

It looks like the MCNC testcases are failing on Windows.

vaughnbetz (approved):
LGTM.

vaughnbetz:
Any idea what causes the failure?  It looks like it happens right after AP?  Not sure if this is one for @tangxifan

tangxifan:
@vaughnbetz If you see the code changes in https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3560/changes, it does not impact the exit mechanism of regression tests

tangxifan:
It looks like the AP flow is failing. To validate if the bug is related to the python scripts, we can revert to the classical flow for Windows MCNC and see if the problem insists.

AlexandreSinger:
The issue appears to be due to the Eigen library being missing from the Windows build:
image

We can see in the build that it was not found:
image

I agree with Xifan that we should fix this immediately by making the windows build do --pack --place --route --analysis; however we should get the Windows build to work with Eigen (or if that is impossible, we I investigate removing the reliance on Eigen in the solver by using a hand-made solver); but I think Eigen is cross-platform.

AlexandreSinger:
Actually, it looks like it will be a bit difficult to make this test not use AP since the MSVC test is reusing the test from the nightly tests. I will investigate this tomorrow. I think ideally we should just get Eigen working on the windows build.

vaughnbetz:
Agreed, getting Eigen working on windows would be the best solution.

AmirhosseinPoolad:
Hopefully windows package managers all have Eigen and the fix is simple, but if they don't I think another option would be to add Eigen as a submodule. It would also have the additional benefit of making all platforms use the same Eigen version and make VPR more consistent between different platforms.

IMO we should merge this ASAP and raise an issue for the windows regression tests. This is a serious bug, the run_vtr_task script is broken right now.

AlexandreSinger:
@AmirhosseinPoolad I agree with you regarding making Eigen a sub-module; but TBH I would rather move away from Eigen entirely. I have heard from many people that spinning your own solver may actually be faster; however, Eigen also provides a sparse matrix library that I like a lot and am unsure how to move away from...

I think making Eigen a submodule or not is a discussion we should have on Thursday. For now, I added Eigen to the MSVC vcpkg config. If that works then we can merge this PR immediately. If it does not work, I can disable the MCNC test from the MSVC build and we can raise an issue.

AmirhosseinPoolad (approved):

PR #3646, opened 2026-06-20
[CI] Update External Actions

AlexandreSinger:
I always assumed that GitHub actions followed NPM-style semver matching such that when you select version 1.2, for example, you would implicitly be getting version 1.2.23 (or whatever the latest is). We are currently receiving warnings from the ccache action since we are on the release titled 1.2, but we want to be on 1.2.23 which is the latest.

Due to this, Dependabot was not updating these releases correctly. I went through our external actions and ensured that they were all locked to versions which Dependabot can see and maintain. This should prevent issues like this from coming in the future.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: 🛰️ Push container image
         if: github.event_name != 'pull_request'
-        uses: pyTooling/Actions/with-post-step@r0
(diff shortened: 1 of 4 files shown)


AlexandreSinger:
This change resolves these annoying warnings in CI:
image

AmirhosseinPoolad (approved):
LGTM

PR #3647, opened 2026-06-20
[Dependabot] Added Explicit Pip Dependency

AlexandreSinger:
I found an error in dependabot where the pass requirements were being misidentified as pip manifest files and causing some errors. I updated the dependabot config to explicitly target pip manifests that we care about.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,3 +18,24 @@ updates:
     # Check for updates once a week
     schedule:
       interval: "weekly"
+
+  # Pip dependency updates. Directories are listed explicitly to prevent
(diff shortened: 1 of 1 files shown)


AmirhosseinPoolad:
That's a funny bug. How would we test something like this? I'm fine with merging and seeing what happens tbh.

AlexandreSinger:
> How would we test something like this?

I am going to refresh the dependency graph after this is merged:
image

Long-term this is the right solution. TBH when I enabled this, I thought we only had one requirements.txt file in VTR; it turns out we have several for different sub-projects. Its actually quite interesting to see!

AlexandreSinger:
image

This is actually HILARIOUS now that I am looking at our dependency graph... Man...

PR #3662, opened 2026-06-20
[Dependabot] Cleaned Up Pip Dependency Updates

AlexandreSinger:
I found that my last PR did not fix the problem. I am trying just completely ignoring the parse repository. Eventually, we need to consider changing the parse files to potentially having a different extension. I hate pip...

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,23 +19,12 @@ updates:
     schedule:
       interval: "weekly"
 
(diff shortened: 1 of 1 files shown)


AmirhosseinPoolad:
> Eventually, we need to consider changing the parse files to potentially having a different extension.

I would argue that this is more of a dependabot bug than our issue. This workaround should work well hopefully.

PR #3666, opened 2026-06-21
[AP][Timing] Generate Timing Report After GP

AlexandreSinger:
I found that the pre-cluster timing report was being generated before global placement. This was using a delay model where all delays between primitives was estimated as a somewhat large constant delay (representing the inter-tile delay).

AP improves on this by updating the timing graph using the current flat placement. It is much more useful for users of AP to see the timing graph after global placement, so I moved the generation of the pre-pack timing report to after global placement and before full legalization.

Trying this on a basic testcase, the values look reasonable (much more reasonable than the prior timing report); however the delay of staying within a tile still appears quite large. Something to investigate.

### DIFF
--- a/vpr/src/analytical_place/analytical_placement_flow.cpp
+++ b/vpr/src/analytical_place/analytical_placement_flow.cpp
@@ -310,6 +310,13 @@ void run_analytical_placement_flow(t_vpr_setup& vpr_setup) {
                                   device_height,
                                   device_ctx.grid.get_num_layers()));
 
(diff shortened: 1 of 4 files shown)


AlexandreSinger:
@vaughnbetz Please review when you have a moment. This small PR updates the timing report generated pre-packing to include the updated delays from global placement. This will help me debug the critical path delays returned by GP.

vaughnbetz (approved):
LGTM.

PR #3668, opened 2026-06-22
[WIP][FlatRouting] Fixed Memory Corruption in Sync Netlist

### LINKED ISSUE #3667: [CI] Flat Router Weekly Test Memory Corruption
Currently the weekly tests are failing with a memory leak in many testcases that use flat routing:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/27899339282

The issues appear in vpr.out (not vpr_stdout.txt) and all appear to look like this:

Image

This crash happens when free is called on an invalid memory address. It appears that there is a memory leak somewhere in our datastructures which is cause it to crash when the data structures are destroyed.

AlexandreSinger:
Currently, several tests are crashing during deallocation of the major datastructures. This often means memory corruption. From looking at Valgrind, I traced the issue down to the sync_netlists_to_routing_flat method. I believe the issue is that we walk the routing data structure, but we are not checking if the tree exists. It is possible for trees to not exist when the net is global for example (i.e. clocks are not routed). Fixed this by adding a safety check.

Resolves #3667

### DIFF
--- a/vpr/src/pack/sync_netlists_to_routing_flat.cpp
+++ b/vpr/src/pack/sync_netlists_to_routing_flat.cpp
@@ -318,6 +318,8 @@ static void sync_clustered_netlist_to_routing(void) {
      * Add the associated net and port too if they don't exist */
     for (auto parent_net_id : atom_ctx.netlist().nets()) {
         auto& tree = route_ctx.route_trees[parent_net_id];
+        if (!tree)
+            continue;
         AtomNetId atom_net_id = convert_to_atom_net_id(parent_net_id);
 
         ClusterNetId clb_net_id;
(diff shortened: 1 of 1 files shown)


AlexandreSinger:
Made this WIP while I run the weekly tests to ensure that this fixes the issue: https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/27969292469

AlexandreSinger:
Weekly test still failed with this fix unfortunately. Back to the drawing board. I still like this last fix though so worth keeping.

AlexandreSinger:
Superseded by https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/3681

PR #3669, opened 2026-06-23
[Python][Lint] Fixed Linting Error in CI

AlexandreSinger:
It looks like the version of prettytable was updated on Pip and our code was not quite version agnostic, leading to an error in CI. Updated the prettytable printer to use a simpler method to avoid the lint error.


AlexandreSinger:
This resolves an issue in the CI causing it to fail on Master.

PR #3670, opened 2026-06-23
[Timing] Improved Pre-Cluster Delay Calc with Mol Info

### LINKED ISSUE #2972: [Pack][Timing] Pre-Cluster Timing Analysis May Not Be Aware of Molecules
The PreClusterDelayCalculator computes the max edge delay of a timing connection in the following function:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d6fe4b509ef5d3f24891e3b88da301d0e011e176/vpr/src/timing/PreClusterDelayCalculator.h#L30-L46

All nets that connect atom blocks together are labelled as INTERCONNECT in the timing_graph_builder here:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d6fe4b509ef5d3f24891e3b88da301d0e011e176/vpr/src/timing/timing_graph_builder.cpp#L297-L302

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/d6fe4b509ef5d3f24891e3b88da301d0e011e176/vpr/src/timing/timing_graph_builder.cpp#L670-L690

The timing graph is created before clustering. The prepacker then occurs after creating the timing graph, but before clustering. The prepacker will pre-cluster atoms together. These molecules will be observed by the packer (i.e. molecules will not get broken). Since the timing graph was created before the prepacker and it does not get updated after prepacking (as far as I am aware), this implies that the max edge delay computed by this calculator is not taking the prepacked molecules into account.

If a timing graph edge is between two atoms within the same molecule, the internal delay of that implemented molecule should be used instead of the inter_cluster_net_delay value.

Proposed next steps:

1. Verify that the assumptions above is true by adding an assertion within the max_edge_delay. The PreClusterDelayCalculator has access to the prepacker, it can check if two atoms are in the same molecule.
2. If we verify that this is correct, we should add a case to this method which will check if the edge is absorbed into a molecule and use a more accurate delay.
3. Verify the pre-clustering timing report to ensure that the molecule delays are being observed.

See https://github.com/verilog-to-routing/vtr-verilog-to-routing/pull/2965#discussion_r2036367221 for context

@vaughnbetz Let me know if I am missing anything here.

AlexandreSinger:
The pre-cluster delay calculator was unaware of molecules, so it was
assuming that the delay between atoms was a very high value (an estimate
of inter-tile delay). AP was likely never fixing this since it does not
change the delays within molecules. This can make molecules appear more
expensive than they actually are.

We now check if a timing arc is within a molecule and if it is we
traverse the pb_graph to find its delay. If we cannot find a path, we
default it to 0.0, meaning that the delay between two atoms in the same
molecule is likely very short.

Resolves #2972


AlexandreSinger:
Results on titan_quick_qor:

  | baseline_parse_results.txt | parse_results.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 1.07
num_LAB | 1 | 0.99
num_DSP | 1 | 0.99
num_M9K | 1 | 1.01
num_M144K | 1 | 0.98
max_vpr_mem | 1 | 1.00
num_post_packed_blocks | 1 | 0.99
device_grid_tiles | 1 | 1.00
pack_time | 1 | 1.05
placed_wirelength_est | 1 | 0.99
place_time | 1 | 1.06
placed_CPD_est | 1 | 1.02
routed_wirelength | 1 | 0.99
critical_path_delay | 1 | 1.01
geomean_nonvirtual_intradomain_critical_path_delay | 1 | 1.00
crit_path_route_time | 1 | 1.04

Overall, the results did not change by much. The machine was a bit loaded, so the run time is not a concern for me. The CPD went up slightly and the WL went down slightly, telling me this is a wash.

I think this is still worth merging in since it does make the pre-clustering timing analysis more accurate. I think this will interact well with my future upgrades to the timing in AP.

I will need to update some golden results after this, but I am ok with that.

@vaughnbetz FYI

AlexandreSinger:
@vaughnbetz This PR is ready for review. This is the PR bringing in the updates to the pre-cluster delay calculator. I originally just did the intra-molecule delay; however, you recommended adding the delays between molecules that are part of carry chains. I have since added carry-chains.

You also recommended caching, and I realized that you were absolutely right. Caching is an absolute necessity for this since this function is part of Tatum's STA process (so it is called frequently). I found that this may have caused the Valgrind test to explode in run time. I have added caching, using the post-cluster delay calculator as an example.

After your review, I can update the golden results.

vaughnbetz:
Failures seem to be QoR improvements (higher density packing) from the ones I've looked at, so OK.

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -177,6 +178,23 @@ static int compute_flat_index_for_child_node(int num_children_of_type,
...
+/**
+ * @brief BFS through the pb_graph, accumulating delay_max, until either
+ *        target is reached (if non-null) or a root-block pin is reached
+ *        (if target is nullptr).
+ *
+ * @param start    Pin to search from.
+ * @param forward  If true, traverse output_edges/output_pins (search
+ *                 downstream of start). If false, traverse input_edges/
+ *                 input_pins (search upstream of start).
+ * @param target   If non-null, search stops at this specific pin. If
```
vaughnbetz, line 186: Will the reader know what a 'root block pin' is?  I don't (maybe it's the top of the pb_graph, i.e. a pin on the cluster-level physical block)?  Good to say what this is in brackets.

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -1930,6 +1948,63 @@ const t_pb_graph_edge* get_edge_between_pins(const t_pb_graph_pin* driver_pin, c
     return nullptr;
 }
 
+// target == nullptr means "stop at the nearest root-block pin" instead of
```
vaughnbetz: Better to make a proper doxygen comment.

vpr/src/pack/pb_type_graph.cpp
```diff
@@ -1930,6 +1948,63 @@ const t_pb_graph_edge* get_edge_between_pins(const t_pb_graph_pin* driver_pin, c
...
+        size_t num_edges = forward ? cur_pin->output_edges.size() : cur_pin->input_edges.size();
+        for (size_t ie = 0; ie < num_edges; ++ie) {
+            const t_pb_graph_edge* edge = forward ? cur_pin->output_edges[ie] : cur_pin->input_edges[ie];
+            int num_pins = forward ? edge->num_output_pins : edge->num_input_pins;
+            for (int ip = 0; ip < num_pins; ++ip) {
+                const t_pb_graph_pin* next = forward ? edge->output_pins[ip] : edge->input_pins[ip];
+                if (is_search_target(next, target)) {
+                    return cur_delay + edge->delay_max;
+                }
+                if (visited.count(next) == 0) {
```
vaughnbetz, line 1975: I think this bfs will not find the longest path to a pin if there are multiple paths with different delays. It will do a bfs and make the delay of each node with the delay found when it is first reached.

Is this OK?  Normally for timing analysis you want to find the longest delay, which requires a levelized search (e.g. put into the queue only once all the pins / path to that node are found).  (there are other ways too if a levelized search isn't convenient.

vaughnbetz: I guess if you are searching for an interconnect path within a cluster, maybe you want the shortest path, which this bfs will usually give (it gives the fewest levels anyway).  Should comment what you're doing here though and what you want and whether any limitations (doesn't search all possible paths) are not important.

vpr/src/timing/PreClusterDelayCalculator.h
```diff
@@ -181,4 +260,12 @@ class PreClusterDelayCalculator : public tatum::DelayCalculator {
     const LogicalModels& models_;
     const vtr::vector<AtomPinId, float>& timing_arc_delays_;
     const Prepacker& prepacker_;
+
+    // Delay caches indexed by tatum::EdgeId, NaN-initialised. Lazily populated
```
vaughnbetz, line 289: I assume this data is deleted after the full legalizer (or regular clusterer) so no it doesn't impact the peak memory. We could use an unordered_map here, it would probably be slower so not worth it if wasting a bit of memory is not an issue.

vpr/src/timing/PreClusterDelayCalculator.h
```diff
@@ -162,6 +204,43 @@ class PreClusterDelayCalculator : public tatum::DelayCalculator {
...
+        const t_pb_graph_pin* sink_gpin = find_pb_graph_pin(sink_pin);
+
+        float delay = calc_pb_graph_path_delay(src_gpin, sink_gpin);
+
+        // If a valid path was found through the pb_graph hierarchy, use it.
+        // Otherwise we assume that it is a very low delay, close to 0.
+        return tatum::Time(delay >= 0.0f ? delay : 0.0f);
+    }
+
+    tatum::Time calc_inter_molecule_chain_delay(AtomPinId src_pin, AtomPinId sink_pin) const {
```
vaughnbetz, line 243: Good to add a brief doxygen comment.

vpr/src/timing/PreClusterDelayCalculator.h
```diff
@@ -162,6 +204,43 @@ class PreClusterDelayCalculator : public tatum::DelayCalculator {
         return gpin;
     }
 
+    tatum::Time calc_intra_molecule_delay(AtomPinId src_pin, AtomPinId sink_pin) const {
```
vaughnbetz, line 218: Good to add a brief doxygen comment.

vaughnbetz (commented):
Looks mostly OK but I have a question about the bfs / delay semantics (max, min, close enough?) that you want.

vaughnbetz:
@AlexandreSinger : when you get a chance let me know what you think about the bfs etc. comment.

vpr/src/pack/pb_type_graph.cpp
AlexandreSinger: This method is being used for delay estimation; basically given two atoms within the same molecule, what will be the delay of the path taken between these two atoms will be? It is trying to predict the routing delay that the intra-lb router will have.

BFS was purposeful here since that would return the most likely path that the intra-lb router will return (assuming no congestion).

vpr/src/timing/PreClusterDelayCalculator.h
AlexandreSinger: For the traditional flow, you are correct that this structure is freed after packing completes. For the AP flow it happens to be kept alive during the annealing flow (since the object just happens to live on). For both flows, this structure is freed before routing.

We cannot use an unordered_map here I think since this object is used inside a parallel_for loop in Tatum. I recall trying to use an unordered_map and it caused issues.

AlexandreSinger:
@vaughnbetz I have resolved your feedback. I expect QoR to fail for many tests, which I will resolve.

What do you think about my comments above? I think BFS is correct here since we want to estimate the delay of wires post intra-lb routing.

If you agree, I can update the comments to reflect this and then update the golden results.

vpr/src/pack/pb_type_graph.cpp
vaughnbetz: This sounds fine, but please add a comment to the code saying why bfs is a good choice here (matches the non-timing-driven intra-cluster router, and minimizes resource use even if we later use flat routing).  We could switch to a minimum delay (Djikstra) search but it would probably make little difference in timing estimates for these short routes.

vaughnbetz (approved):
Fine to merge when you are ready, but I suggest one comment documenting the bfs (embedded suggestion).

PR #3671, opened 2026-06-23
Remove -fPIC flag for MSVC

AmirhosseinPoolad:
MSVC does not support -fPIC (and the concept apparently doesn't really exist in Windows anyway) and the CI run was full of "cl : Command line warning D9002 : ignoring unknown option '-fPIC' ". This PR changes the Cmake config so that -fPIC is only added when the compiler is not MSVC.


AlexandreSinger:
Lol I also noticed this. We should merge this in.

After this change, the other major warnings are coming from "potential loss of data" in type conversions. I think it would be a royal pain to do explicit casts everywhere in VPR, so these we may want to explicitly disable in the MSVC build. Something for the future.

PR #3672, opened 2026-06-24
Update old_traceback

amin1377:
When creating the routing tree from `t_trace`, the current implementation uses a recursive function, where a new function call is made for each node and the calls remain on the stack until the last `SINK` node is reached. This is normally not an issue; however, for high-fanout nets on large architectures, it can cause a stack overflow. In this PR, the code is updated to use a loop instead of a recursive function.


AlexandreSinger (approved):
LGTM, thanks Amin!

Yeah, there is no good reason to make this function a recursive function; the while loop version is just cleaner. Normally when converting a recursive function to the non-recursive one you need a queue.

One comment: The inverse function `traceback_from_route_tree` is still a recursive function. It should not have the same fanout issues as the function you fixed, but if you fixed this for a specific reason you may want to fix that one too. Up to you though, it does not need to block this PR.

amin1377:
The case for `traceback_from_route_tree_recurr` is a bit different, since it removes function calls from the stack once it reaches a `SINK` node. The function I fixed here was a major problem because it would keep track of the whole net.

Overall, I agree with your comment though. I happen to dislike recursive functions regardless of whether they cause a problem or not :) I’ll fix the function you mentioned in a separate PR.

PR #3673, opened 2026-06-24
Created separate files for critical path drawing, and implemented a dynamic decluttering algorithm for delay labels

kevinxing-a12:
#### Description

I created two files named draw_crit_path.h and draw_crit_path.cpp and moved everything related to critical path drawing from draw_basic.h and .cpp there. The more important part about this PR is that I implemented an algorithm that tries different position candidates when placing a delay label, and chooses one that gives zero overlap, or hides the label if conflict is inevitable. Doing so minimizes cluttering among different labels.

An example:

Before:
before

After:
after

#### Related Issue

Not really an issue, but a big struggle that I had when writing the code:
Amir and Alex friendly suggested to me that having a class for the decluttering algorithm would be nice, because there are quite a few helper functions. I tried it (you can see the class structure in earlier commits), but it was found out later that removing implementation helpers and struct definitions from the class definition would make the header file cleaner. This was where I started struggling: if the class is only left with a public function and private variable (a vector holding info of every delay label), isn't it more straightforward to remove it? Also, the class object is constructed whenever delay labels are drawn, and destructed right after the drawing happens. So as for now, I return to using the helper functions directly, but I doubt if this is the ultimate solution.

I would very appreciate if people can share their views on this!!!

#### Motivation and Context

Regarding the new files:
Previously, there were too many things crammed all together in draw_basic.h and .cpp. The issue became more significant after I implemented the delay label decluttering algorithm, so I believe having dedicated critical path drawing files is nice.

Regarding the decluttering algorithm:
Previously, all delay labels were drawn at the same time at their corresponding flyline center, which resulted in massive cluttering and made the labels impossible to read.

I've tried it on the GUI using multiple different testbenches from the regression tests.

The decluttering algorithm partially works on the rhi renderer -- the algorithm itself is not broken, but the rhi renderer doesn't perform a full redraw for zooming and panning. If you change a setting or click on the canvas (which force a redraw) you will then be able to see the algorithm working.

The algorithm works well on the immediate and deferred renderer 

#### Types of changes

#### Checklist:


AmirhosseinPoolad:
> if the class is only left with a public function and private variable (a vector holding info of every delay label), isn't it more straightforward to remove it?

Makes sense in this case. OOP is nice but not everything needs OOP!

AmirhosseinPoolad:
I see that in some of the helper functions you're taking a vector by reference, modifying it and using it as the function's output. It's not "wrong" to do per-se (although I personally don't like output arguments), but it's actually not necessary for proper optimization!

Check out this snippet.
https://godbolt.org/z/e55fY1q9c

The assembly is a mess but the important part here is that `operator new` is only called once* in `main`. All the argument passing and returning is done by moves. Check out 'return value optimization' and 'move semantics' if you want to see why.

*There are two more 'new's in `a` and `b`, but that's just the compiler inlining `std::vector::push_back`.

vaughnbetz:
We should get the rhi redraws correct with lod / dynamic decluttering like this. Since this drawing is low complexity and changes often it should probably be an overlay.

kevinxing-a12:
Yes, will bring this change in the next PR!

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+ * @param path Timing path whose consecutive node pairs define the timing edges to place delay labels.
+ * @param pixels_per_world_unit The ratio between pixels and world units spanning the screen width.
+ * Used to perform screen-to-world conversions for label bounding boxes that primarily use pixels.
+ * @return Per-edge delay label drawing information that does not yet tell where each label will be drawn.
+ */
+static std::vector<t_label_drawing_info> calculate_basic_label_drawing_info(const tatum::TimingPath& path,
+                                                                            double pixels_per_world_unit);
+
+/**
+ * @brief Chooses label positions that result in the least number of overlaps between visible labels.
```
AmirhosseinPoolad: Is it actually the least number of overlaps? Very pedantic but I think saying "greedily tries to minimize overlaps" or something similar would be more accurate.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+
+    draw_labels(final_label_drawing_info, g);
+}
+
+static std::vector<t_label_drawing_info> calculate_basic_label_drawing_info(const tatum::TimingPath& path,
+                                                                            double pixels_per_world_unit) {
+
+    std::vector<t_label_drawing_info> basic_label_drawing_info;
+    // Subtract 1 to get the number of edges instead of nodes.
+    // The callers of this function have ensured that path is not empty.
```
AmirhosseinPoolad: Add a VTR_ASSERT_SAFE for this check. It wouldn't be run on default builds (no performance hit) but CI runs would check it and we'll get a nice error message if another caller forgets to ensure this.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+        prev_arr_time = arr_time;
+    }
+    return basic_label_drawing_info;
+}
+
+static std::vector<t_label_drawing_info> calculate_least_cluttered_label_pos(std::vector<t_label_drawing_info> basic_label_drawing_info,
+                                                                             double pixels_per_world_unit) {
+
+    // The label position candidates are ordered in an implied priority which the decluttering algorithm will follow.
+    std::vector<e_label_relative_pos> label_pos_candidates = {e_label_relative_pos::CENTER_ABOVE,
```
AmirhosseinPoolad: This should be a constexpr std::array instead.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+        // At this stage, hidden labels are due to their corresponding flylines being invisible. Therefore, we skip them.
+        if (drawing_info.hide_label) {
+            continue;
+        }
+
+        // The position candidate to start with is CENTER_ABOVE, which aligns with the order in label_pos_candidates.
+        e_label_relative_pos candidate_with_least_overlaps = e_label_relative_pos::CENTER_ABOVE;
+
+        // This is a good upper bound for the maximum number of overlaps a label can have.
+        int least_num_overlaps = basic_label_drawing_info.size();
```
AmirhosseinPoolad: I'm not sure why you're calculating an upper bound instead of using numeric_limts::max() or something similar.

Doesn't really matter but it's just more cognitive load and things to think about when reading the code.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+        t_label_drawing_info& drawing_info = post_decluttering_label_drawing_info[edge_idx];
+
+        // Skip if the label is already hidden.
+        if (drawing_info.hide_label) {
+            continue;
+        }
+
+        // As long as there is one overlap associated with this label, we hide it.
+        // Note: For every label we completely redo the overlap calculation, therefore hiding a label might
+        // change the fate of subsequent labels.
```
AmirhosseinPoolad: I think this might've been left over from a previous version of the code.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+                                                              e_label_relative_pos label_relative_pos,
+                                                              double pixels_per_world_unit) {
+    double edge_length = label_to_update.edge_length;
+    // The unit length for a scalable edge-direction offset in world coordinates.
+    double edge_offset_unit = edge_length * EDGE_OFFSET_PERCENT;
+
+    // Convert MAX_EDGE_OFFSET_UNIT to world coordinates.
+    if (edge_offset_unit > MAX_EDGE_OFFSET_UNIT / pixels_per_world_unit) {
+        edge_offset_unit = MAX_EDGE_OFFSET_UNIT / pixels_per_world_unit;
+    }
```
AmirhosseinPoolad: Why is the offset unit clamped to 40 pixels?

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+            perpendicular_offset = LABEL_HEIGHT / pixels_per_world_unit;
+            edge_offset = edge_offset_unit * 2;
+            break;
+        case e_label_relative_pos::FAR_RIGHT_BELOW:
+            perpendicular_offset = -LABEL_HEIGHT / pixels_per_world_unit;
+            edge_offset = edge_offset_unit * 2;
+            break;
+        default:
+            perpendicular_offset = LABEL_HEIGHT / pixels_per_world_unit;
+    }
```
AmirhosseinPoolad, line 737: Consider breaking this into two switch statements, one calculating perpendicular_offset and the other calculating edge_offset. This would fix the code duplication here.

```
switch (label_relative_pos) {
case CENTER_ABOVE:
case LEFT_ABOVE:
// and the other above cases
    perpendicular_offset = LABEL_HEIGHT / pixels_per_world_unit;
    break;
// same thing for below cases
}
```

The other switch would group the above and below cases for each distance.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+    // Apply the edge offset.
+    x_offset += edge_offset * cos(rotation_angle_in_deg);
+    y_offset += edge_offset * sin(rotation_angle_in_deg);
+
+    // Calculate the label bounding box by applying the 2d offset to the virtual bounding box sitting at edge center.
+    ezgl::rectangle label_bbox = label_to_update.virtual_centered_label_bbox + ezgl::point2d(x_offset, y_offset);
+    return label_bbox;
+}
+
+static bool check_if_bboxes_overlap(const ezgl::rectangle& bbox1, const ezgl::rectangle& bbox2) {
```
AmirhosseinPoolad, line 780: this is a static function so you're your only user, but you should still add a safe assertion to check if bbox1 and bbox2 are aligned with the X and Y axes. Helps with making the code self-documenting.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,839 @@
...
+ * @param label_to_update Single Label drawing information whose bounding box waits to be updated.
+ * @param label_relative_pos Candidate position to apply relative to the timing-edge flyline.
+ * @param pixels_per_world_unit The ratio between pixels and world units spanning the screen width.
+ * Used to convert values specified in pixels that determine the offset of label bounding boxes.
+ * @return The rectangle used to replace the current bounding box associated with label_to_update.
+ */
+static ezgl::rectangle calculate_label_bbox_from_relative_pos(t_label_drawing_info& label_to_update, e_label_relative_pos label_relative_pos, double pixels_per_world_unit);
+
+/**
+ * @brief Returns true if two label bounding boxes overlap.
```
AmirhosseinPoolad, line 247: Mention that bbox1 and bbox2 must be aligned with the X and Y axes.

vpr/src/draw/draw_types.h
```diff
@@ -125,6 +125,19 @@ enum class e_edge_type {
     NUM_EDGE_TYPES
 };
 
+enum class e_delay_text_relative_pos {
```
AmirhosseinPoolad: This is a duplicate of e_label_relative_pos. I would keep only one of them.

AmirhosseinPoolad (commented):
Great PR Kevin, thanks. Had some minor thoughts.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: I believe an ezgl::rectangle is always aligned with X and Y by itself?

vpr/src/draw/draw_types.h
kevinxing-a12: This is old code that should have been deleted. Very sorry.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: I've added a clarification comment. Basically, for super long edges, if we still use percentage for edge offset, labels can jump very drastically when changing their positions. Therefore we want to cap the offset amount.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: It's supposed to stay with the current code, but I may have done a poor job in explaining this, sorry! I have made a clearer explanation in the new commit.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: I see. In terms of absolute minimum, saying least num of overlaps can be inaccurate. I will change the comment.

vpr/src/draw/draw_crit_path.cpp
AmirhosseinPoolad: You're right!

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,874 @@
...
+    // Convert MAX_EDGE_OFFSET_UNIT (defined in pixels) to world coordinates.
+    if (edge_offset_unit > MAX_EDGE_OFFSET_UNIT / pixels_per_world_unit) {
+        edge_offset_unit = MAX_EDGE_OFFSET_UNIT / pixels_per_world_unit;
+    }
+
+    double perpendicular_offset = 0;
+    double edge_offset = 0.0;
+
+    // Apply perpendicular offset associated with the specified relative position.
+    switch (label_relative_pos) {
```
AmirhosseinPoolad, line 716: Not quite what I meant, sorry! In C/C++ you can have multiple cases for each piece of code. This switch statement would become:

```
    switch (label_relative_pos) {
        case e_label_relative_pos::CENTER_ABOVE:
        case e_label_relative_pos::LEFT_ABOVE:
        case e_label_relative_pos::RIGHT_ABOVE:
        case e_label_relative_pos::FAR_LEFT_ABOVE:
        case e_label_relative_pos::FAR_RIGHT_ABOVE:
            perpendicular_offset = LABEL_HEIGHT / pixels_per_world_unit;
            break;
        case e_label_relative_pos::CENTER_BELOW:
        case e_label_relative_pos::LEFT_BELOW:
        case e_label_relative_pos::RIGHT_BELOW:
        case e_label_relative_pos::FAR_LEFT_BELOW:
        case e_label_relative_pos::FAR_RIGHT_BELOW:
            perpendicular_offset = -LABEL_HEIGHT / pixels_per_world_unit;
            break;
        default:
            perpendicular_offset = LABEL_HEIGHT / pixels_per_world_unit;
    }
```

Same concept for the other switch statement.

I would also add a VTR_ASSERT(false) to the default case since if it happens it means that someone added a case to the enum while forgetting to update this code, in which case we would want to error out.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+#include "vpr_utils.h"
+#include "vtr_assert.h"
+
+/**
+ * @brief A scaling factor applied to DEFAULT_ARROW_SIZE.
+ */
+static constexpr int TIMING_EDGE_ARROW_SCALE = 30;
+
+/**
+ * @brief Indicates the relative position on a timing edge. 0 and 1 represent the two endpoints.
```
vaughnbetz: 0 is start and 1 is end?  Should say.

vpr/src/draw/draw_crit_path.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#ifndef NO_GRAPHICS
+
```
vaughnbetz, line 17: Put a brief comment in saying what this file is for. E.g.:


/**
--
* @file draw_crit_path.h
*
* This file has drawing routines related to visualizing the critical path in a placed and/or routed design.
*/

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+ */
+static constexpr float EDGE_CENTER = 0.5;
+
+/** 
+ * @brief The width of a critical path time delay label that has exactly three decimals (e.g 1.500), in pixels (screen coordinates).
+ * 
+ * TODO: Create a function to expose ezgl's internal Cairo function that returns the width and height
```
vaughnbetz: Should do this.  Note that I think after the Qt port we are not using Cairo anymore either.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+ */
+static constexpr int LABEL_WIDTH = 40;
+
+/**
+ * @brief The height of a time delay label in pixels.  
+ */
+static constexpr int LABEL_HEIGHT = 13;
+
+/**
+ * @brief The percentage of the total edge length that a delay label can be offset in the edge direction.
```
vaughnbetz: Not really sure what this means .... 0.1 percent is a very small offset too.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+ */
+static constexpr double EDGE_OFFSET_PERCENT = 0.1;
+
+/**
+ * @brief The maximum unit distance that a label can be offset along the edge in pixels.
```
vaughnbetz: Is offset along an edge or perpendicular to it?

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+ */
+static constexpr int MAX_EDGE_OFFSET_UNIT = 40;
+
```
vaughnbetz, line 41: Add comment.
Highly contrasting colours that are useful for visualization.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+    ezgl::color(220, 211, 0),   //greenish yellow
+    ezgl::color(136, 45, 23),   //redish brown
+    ezgl::color(141, 182, 0),   //yellow green
+    ezgl::color(101, 69, 34),   //yellowish brown
+    ezgl::color(226, 88, 34),   //reddish orange
+    ezgl::color(43, 61, 38)     //olive green
+};
+
+/**
+ * @brief Candidate positions for placing a delay label relative to its timing-edge flyline.
```
vaughnbetz, line 71: I'd explain this a bit more.  Above and below: what do these mean in general (above and below a line in the horizontal dimension? what do they mean for a vertical line?)

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+    }
+}
+
+static void draw_routed_connections_between_nodes(tatum::NodeId src_tnode, tatum::NodeId sink_tnode, ezgl::color color, ezgl::renderer* g) {
+    const AtomContext& atom_ctx = g_vpr_ctx.atom();
+    const ClusteringContext& cluster_ctx = g_vpr_ctx.clustering();
+    const TimingContext& timing_ctx = g_vpr_ctx.timing();
+
+    AtomPinId atom_src_pin = atom_ctx.lookup().tnode_atom_pin(src_tnode);
+    AtomPinId atom_sink_pin = atom_ctx.lookup().tnode_atom_pin(sink_tnode);
```
vaughnbetz, line 372: Check if this code works with flat routing on ... it seems like it would be broken or at least partly incorrect in that case.  For flat routing we can walk entirely through the routing.  Don't have to fix in this PR but worth investigating.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+            //      .................
+            //      .        ////// .
+            //      .      //////   .
+            //      .    //////     .
+            //      .  //////       .
+            //      .//////         .
+            //      .................
+
+            // LABEL_WIDTH and LABEL_HEIGHT correspond to a string that has exactly three decimals (e.g 1.500). Unit is pixel.
+            double label_bbox_width = (LABEL_WIDTH * cos(rotation_angle * (std::numbers::pi / 180))
```
vaughnbetz: We should instead expose the internal ezgl bounding box coordinates and use them -- refactor to have the bbox coordinates calculated in one place in ezgl and you can call that routine from both the drawing routine and the calculate routine.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -0,0 +1,853 @@
...
+
+static void draw_labels(std::vector<t_label_drawing_info>& final_label_drawing_info, ezgl::renderer* g) {
+    g->set_font_size(16);
+
+    for (std::size_t edge_idx = 0; edge_idx < final_label_drawing_info.size(); edge_idx++) {
+        t_label_drawing_info& drawing_info = final_label_drawing_info[edge_idx];
+
+        // In draw_timing_edge_flylines(), flylines share the same edge_idx with the labels here,
+        // so they are also paired with the same color.
+        ezgl::color color = kelly_max_contrast_colors[edge_idx % kelly_max_contrast_colors.size()];
```
vaughnbetz: A bit dangerous to tie the colour logic in two places together like this. Could instead put the colour in the drawing_info or make a small helper function to compute the timing_edge colour from an edge index (so at least everyone is calling the same function).

vaughnbetz (changes requested):
A few suggestions, mostly minor.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: Sorry, the choice of word is wrong here! Should have used "proportion", so 0.1 refers to 10% of the total edge length.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: Can be both. For this snippet, offset is along an edge. My intention was to use "EDGE" to indicate that the offset discussed here is in the edge direction (along the edge), but maybe I should make the name even clearer?

vaughnbetz: I'd just have the comment say it is offset from the center, along the edge (if that is what it is).
Change percent to fraction in the name.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: I tried `/vtr-verilog-to-routing/vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_flat_router` and confirmed that `--flat_routing on` is provided. I think the result is quite plausible?

Only routed timing connections:


Routed timing connections + flylines:

vaughnbetz: The flylines look reasonable. The routed ones seem to be missing some of the routed inter-cluster connections, and the colours don't match the flylines (I think) which makes things harder to interpret.  

Turning on block internals to see what is shown inside a cluster may also help show what is and isn't being drawn.

@amin1377 : does flat routing try to visualize the within-cluster routing?  That would be a useful upgrade if not. Does the amount of inter-cluster routing shown make sense to you, given the flyline critical path?  If you two could chat about this (given that you're in the same room) it would be good.

kevinxing-a12: I have identified the bug in the current code. Basically, the code expects a cluster net id to be passed to the route tree, but in flat routing it should be an atom net id. The current implementation doesn't crash likely because the cluster net id coincidentally doesn't exceed the range of the route tree which is now indexed by atom net id, but hence you see the mismatch between flylines and routed connections. I'm fixing it (it's not that bad) and will push the changes to this PR.

vaughnbetz: Great, thanks! This will be an important upgrade.

kevinxing-a12: This problem has been fixed! The routed connections in the flat router regression test now seem consistent with the flylines. The picture below is to contrast with the one I got yesterday (see above).



I also checked the non flat router tests just in case, and I didn't spot any changes in the routed connections!


In the commit that I am going to push very shortly, I made two functions that handle the flat router case and the non flat case respectively. They are called inside draw_routed_connections_between_nodes().

Though this problem is gone, I saw discrepancies in endpoint locations between flylines and routed connections:


I believe we have talked about this problem earlier in a different context, and it will be eventually on the agenda, though probably in a different PR. But just so people know:).

vaughnbetz:
Great, looks good.  I agree, aligning the endpoints between flylines and routed connections would be good, but can wait for a future PR.
Here's another enhancement idea for a future PR: show the endpoints of the timing path (start of the very first net and end of the very last one) somehow, e.g. with diamonds.

kevinxing-a12:
@vaughnbetz This PR should be good to merge after the CI!

vaughnbetz (approved):
LGTM

PR #3675, opened 2026-06-26
[Pack] Refactor and document pin class inference

haydar-c:
This PR simplifies the pin class assignment code and adds documentation of how pin class assignment is performed for a given complex block.


vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -39,50 +39,33 @@
...
 static void unmark_fanout_intermediate_nodes(t_pb_graph_pin* current_pb_graph_pin,
                                              std::unordered_set<t_pb_graph_pin*>& seen_pins);
-static void expand_pb_graph_node_and_load_pin_class_by_depth(t_pb_graph_pin* current_pb_graph_pin,
-                                                             const t_pb_graph_pin* reference_pb_graph_pin,
-                                                             const int depth,
-                                                             int* input_count,
-                                                             int* output_count,
-                                                             std::unordered_map<t_pb_graph_pin*, int>& pin_marker);
 static void sum_pin_class(t_pb_graph_node* pb_graph_node);
+static void assign_pin_class_in_subtree(t_pb_graph_pin* seed_pin,
```
vaughnbetz: Doxygen comments should go in declaration.

vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -39,50 +39,33 @@
...
+                                        t_pb_graph_node* pb_graph_node,
+                                        const int class_id);
+static void load_pin_classes_in_pb_graph_node(t_pb_graph_node* pb_graph_node);
```
vaughnbetz, line 132: Doxygen at declaration point.

vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -485,6 +261,184 @@ static void sum_pin_class(t_pb_graph_node* pb_graph_node) {
...
+ *        within pb_graph_node and its subtree.
+ *
+ * All reachable primitive pins of the same type as seed_pin are assigned class_id via
+ * parent_pin_class[depth]. All reachable boundary pins of pb_graph_node of the same type
+ * are assigned class_id via pin_class. Intermediate pins are traversed but not assigned.
+ *
+ * @param seed_pin      A primitive pin that seeds the BFS.
+ * @param pb_graph_node The pb_graph node whose subtree the BFS is confined to.
+ * @param class_id      The class ID to assign to all pins in this equivalence group.
+ */
```
vaughnbetz: Move to declaration.

vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -485,6 +261,184 @@ static void sum_pin_class(t_pb_graph_node* pb_graph_node) {
     }
 }
 
+/**
+ * @brief Assigns class_id to all same-type pins reachable from seed_pin
+ *        within pb_graph_node and its subtree.
+ *
+ * All reachable primitive pins of the same type as seed_pin are assigned class_id via
```
vaughnbetz: explain 'same type' -- e.g. could say:
same type (input, output or clock, and in the same pb as the seed_pin)

vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -485,6 +261,184 @@ static void sum_pin_class(t_pb_graph_node* pb_graph_node) {
...
+            }
+            for (int edge_idx = 0; edge_idx < pin->num_output_edges; edge_idx++) {
+                VTR_ASSERT(pin->output_edges[edge_idx]->num_output_pins == 1);
+                t_pb_graph_pin* neighbor_pin = pin->output_edges[edge_idx]->output_pins[0];
+                if (!visited.count(neighbor_pin)) {
+                    visited.insert(neighbor_pin);
+                    queue.push(neighbor_pin);
+                }
+            }
+        } else if (pin->parent_node == pb_graph_node) {
```
vaughnbetz, line 347: Not sure if you can explain this any further .... not 100% clear to me.  But clearly an improvement already so if you can't describe further its OK.

vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -485,6 +261,184 @@ static void sum_pin_class(t_pb_graph_node* pb_graph_node) {
...
+ *
+ * A pin class is a group of pins on a pb_graph node that are mutually reachable
+ * through the interconnect within that node's subtree (ignoring edge directionality)
+ * and of the same type (input/clock or output). Each class gets a unique ID.
+ *
+ * The function recurses into children so that pin classes are computed independently
+ * at every level of the hierarchy. Each level records:
+ *  - pin_class on its own boundary pins (which class they belong to)
+ *  - parent_pin_class[depth] on primitive pins (which class they belong to as seen
+ *    from the ancestor pb_graph node at this depth)
```
vaughnbetz: If you can give an example, that would help (e.g. a crossbar between a bunch of inputs at some level would lead to ...).

vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -485,6 +261,184 @@ static void sum_pin_class(t_pb_graph_node* pb_graph_node) {
...
+                if (prim->output_pins[port_idx][pin_idx].parent_pin_class[node_depth] == UNDEFINED) {
+                    assign_pin_class_in_subtree(&prim->output_pins[port_idx][pin_idx], pb_graph_node, output_class);
+                    output_class++;
+                }
+            }
+        }
+    }
+
+    // Allocate class size arrays.
+    pb_graph_node->num_input_pin_class = input_class;
```
vaughnbetz: 1. Could change to vectors ... better.
2. Could then (optionally) remove the count and load passes and just do one pass with .push_back().

vaughnbetz: But these can be deferred to a later PR.

vaughnbetz (changes requested):
A few suggestions.

vpr/src/pack/cluster_feasibility_filter.cpp
```diff
@@ -485,6 +261,184 @@ static void sum_pin_class(t_pb_graph_node* pb_graph_node) {
...
+                    assign_pin_class_in_subtree(&prim->input_pins[port_idx][pin_idx], pb_graph_node, input_class);
+                    input_class++;
+                }
+            }
+        }
+        for (int port_idx = 0; port_idx < prim->num_clock_ports; port_idx++) {
+            for (int pin_idx = 0; pin_idx < prim->num_clock_pins[port_idx]; pin_idx++) {
+                if (prim->clock_pins[port_idx][pin_idx].parent_pin_class[node_depth] == UNDEFINED) {
+                    assign_pin_class_in_subtree(&prim->clock_pins[port_idx][pin_idx], pb_graph_node, input_class);
+                    input_class++;
```
AlexandreSinger, line 435: A small note: clock pins are being treated as input classes; however, a clock may not necessarily be an input. TBH VTR's architecture spec is really not clear on this topic, but I think we may want to put clocks in their own category. I am worried (but I have not verified) that the clock pins are counted in the num_used_external_pins filter, which is weird since we would not want to limit the number of clocks entering a block.

This is not for this PR to fix, but could you add a TODO somewhere clear so we can deal with this. I have a feeling I will be looking into this soon.

vpr/src/pack/cluster_feasibility_filter.cpp
AlexandreSinger: I went through the code, this is not the code that I thought it was. This code is built when the pb_graph is created and this information is stored within the pb_graph. I still think that it is not great to store algorithmic information with architectural information (it makes making algorithmic changes more difficult); but thats not for this PR to fix.

I agree with Vaughn that if it is reasonable we could convert these to vectors; but it should not block this PR.

AlexandreSinger (commented):
Looks great @haydar-c 

I went through this code the best that I could, it looks like this code is pre-computed for each pb_graph in the forest (once when the pb_graphs are generated). This is a very valuable cleanup since it sets the stage for what these pin classes actually mean. I see this as a pure improvement in code legibility.

I left a couple of minor comments. I agree with Vaughn's other comments.

haydar-c:
I have added an example of the pin class assignment to the doxygen comment of the `load_pin_classes_in_pb_graph_node` function using a simple pb graph figure. I also changed the old count and load pass to vectors.

haydar-c:
Adding geomean VTR Chain Large results. The QoR is identical to master as expected, and the runtime differences seem like noise to me.


  | PR / Master
-- | --
vtr_flow_elapsed_time | 0.99
max_vpr_mem | 1.00
pack_time | 0.99
place_time | 0.93
crit_path_route_time | 1.02
num_post_packed_blocks | 1.00
routed_wirelength | 1.00
critical_path_delay | 1.00

vaughnbetz (approved):
Great example; thanks!  LGTM.
QoR results look good too.

AlexandreSinger (approved):
LGTM

PR #3679, opened 2026-06-29
Streamline windows build command and update documentation

tangxifan:
#### Description

This PR streamlines the build process in CI flow by modifying the top-level Makefile to
- Automatically set CMake parameters when detecting Windows and MSYS2 system. As a result, in windows, Parmys, ABC and SystemVerilog is automatically disabled

#### Motivation and Context

This is to simplify the build process on Windows and improve user experience to build source codes.

#### Types of changes

#### Checklist:


tangxifan:
@vaughnbetz This PR is ready for your review.

.github/workflows/test.yml
```diff
@@ -1191,12 +1191,10 @@ jobs:
...
       - name: Configure and build with CMake
         env:
+          CMAKE_PARAMS: "-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
           CCACHE_DIR: ${{ github.workspace }}/.ccache
         run: |
-          mkdir build
-          cd build
-          cmake ${{ steps.vcpkg.outputs.vcpkg-cmake-config }} -S .. -B . -G Ninja -DCMAKE_BUILD_TYPE=Release -DWITH_PARMYS=OFF -DSLANG_SYSTEMVERILOG=OFF -DVTR_IPO_BUILD=OFF -DWITH_ABC=OFF -DWGET="${{ env.CURL_PATH }}" -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-          cmake --build . --parallel ${{ steps.cpu-cores.outputs.count }}
+          make CURL_PATH="${{ env.CURL_PATH }}" VCPKG_PATH="${{ github.workspace }}/vcpkg" -j${{ steps.cpu-cores.outputs.count}}
```
AlexandreSinger: I am not as familiar with windows command-line, but why is the `CURL_PATH="${{ env.CURL_PATH }}"` necessary? At least in bash, I think this would be redundant.

BUILDING.md
```diff
@@ -41,6 +70,8 @@ From the top-level, run:
 
 which will build all the required tools.
 
+Note that on Windows platform, all the required tools will be built except ABC, Parmys and Slang.
```
AlexandreSinger, line 73: In the make command in the CI, the make command looks like it takes a few extra arguments. Is this specific to the CI, or is it worth documenting here if a user needs to do it too?

Makefile
```diff
@@ -38,9 +38,33 @@ CMAKE_BUILD_TYPE := $(shell echo $(BUILD_TYPE) | sed 's/_\?pgo//' | sed 's/_\?st
...
+endif
+endif
+
 #Allows users to pass parameters to cmake
 #  e.g. make CMAKE_PARAMS="-DVTR_ENABLE_SANITIZE=true"
-override CMAKE_PARAMS := -DCMAKE_BUILD_TYPE=$(CMAKE_BUILD_TYPE) -G 'Unix Makefiles' ${CMAKE_PARAMS}
+override CMAKE_PARAMS := -DCMAKE_BUILD_TYPE=$(CMAKE_BUILD_TYPE) -G '${CMAKE_GEN}' ${CMAKE_PARAMS}
+
+# The curl path should be defined by user. Try to get one from system
+CURL_PATH:=$(shell where curl.exe 2>nul | head -n 1)
```
AlexandreSinger, line 57: This command is windows-specific. It is looking for curl.exe and I think linux uses `/dev/null` not `nul`, so this would I think create a stray file in root. I think this should be gaurded with the ifeq windows.

Makefile
```diff
@@ -38,9 +38,33 @@ CMAKE_BUILD_TYPE := $(shell echo $(BUILD_TYPE) | sed 's/_\?pgo//' | sed 's/_\?st
...
+# VCPKG root is a system variable env:VCPKG in power shell. User can override by using VCPKG_PATH when calling the makefile
+VCPKG_CMAKE_PATH:=$(subst \,/,$(VCPKG_PATH))
+ifeq ($(OS),Windows_NT)
+# Msys2 can still use Linux gcc
+ifeq ($(MSYSTEM),MINGW64)
+	override CMAKE_PARAMS := ${CMAKE_PARAMS} -DWITH_PARMYS=OFF -DSLANG_SYSTEMVERILOG=OFF -DVTR_IPO_BUILD=OFF -DWITH_ABC=OFF
+else
+	override CMAKE_PARAMS := ${CMAKE_PARAMS} -DWGET="${CURL_PATH}" -DCMAKE_TOOLCHAIN_FILE="${VCPKG_CMAKE_PATH}/scripts/buildsystems/vcpkg.cmake" -DVCPKG_TARGET_TRIPLET=x64-windows-release -DVCPKG_MANIFEST_MODE=OFF
+	override CMAKE_PARAMS := ${CMAKE_PARAMS} -DWITH_PARMYS=OFF -DSLANG_SYSTEMVERILOG=OFF -DVTR_IPO_BUILD=OFF -DWITH_ABC=OFF
```
AlexandreSinger: This line is the same as line 62 above. I think it would be cleaner to combine these at line 60 or between 66 and 67 if it is possible. This will just make this easier to maintain as we improve the support of these features.

AlexandreSinger (changes requested):
Thanks @tangxifan !

These changes certainly make the build easier for MSVC. I had some minor comments below.

vaughnbetz (approved):
Thanks @tangxifan . LGTM.

BUILDING.md
tangxifan: This is general. Currently, ABC and Yosys have not yet supported Windows well.
This is the idea of this PR where the makefile automatically disables them instead of doing it  in CI. User just need to type ``make`` as they do on Linux

Makefile
tangxifan: Done.

Makefile
tangxifan: Done

.github/workflows/test.yml
tangxifan: Done

tangxifan:
@AlexandreSinger All the comments addressed. CI is passing. It is ready for your review.

AlexandreSinger (approved):
LGTM, thanks @tangxifan !

PR #3680, opened 2026-06-29
Fix convert switch index

amin1377:
Fixes a segfault in print_switch_usage() (--full_stats on) when flat routing or the tileable (crr) rr-graph generator is used. convert_switch_index() now iterates over all switches in switch_fanin_remap instead of only the architecture switches, so internal/appended switches are found instead of returning an out-of-range index.


vpr/src/util/vpr_utils.cpp
AlexandreSinger: I do not think this should be a log error. Instead I think this should be a full `VPR_FATAL_ERROR`. If this line is hit, {-1, -1} is returned and the continue you added below would just ignore it. The logs will be filled with the errors but execution would not end.

If this is intentional, perhaps this should be a warning?

vpr/src/util/vpr_utils.cpp
```diff
@@ -1489,6 +1491,10 @@ void print_switch_usage() {
         for (const RRSwitchId rr_switch_id : inward_switch_inf[rr_id] | std::views::keys) {
             float Tdel = rr_graph.rr_switch_inf(rr_switch_id).Tdel;
             const auto [arch_switch_id, fanin] = convert_switch_index(rr_switch_id);
+            // Skip unmappable switches rather than indexing switch_fanin_count[-1].
+            if (arch_switch_id < 0 || arch_switch_id >= (int)switch_fanin_count.size()) {
```
AlexandreSinger: Should this not be an assert? When would it be possible for convert_switch_index to return a value larger than the switch_fanin_count or negative? Or perhaps this should be a stronger error?

vpr/src/util/vpr_utils.cpp
AlexandreSinger, line 1473: Since this fix relies on switch_fanin_count being less than or equal to in size to switch_fanin_remap, should we add an assert here to lock this in: something like `VTR_ASSERT(switch_fanin_remap.size() <= all_sw_inf.size())`.

AlexandreSinger (commented):
LGTM, just a couple of minor comments below.

vpr/src/util/vpr_utils.cpp
amin1377: It was `VTR_LOG_ERROR` before, and I think it should remain that way. We do not necessarily want to stop the flow if this happens, but at the same time, this is something the user should be notified about, and it is more serious than a simple warning.

The way I look at `VTR_LOG_ERROR` is that when something is more serious than a warning and may indicate that the reported results are incorrect, we should use `VTR_LOG_ERROR` rather than a warning.

vpr/src/util/vpr_utils.cpp
amin1377: I changed it to an assertion here and removed the error log mentioned in the comment above.

amin1377:
@AlexandreSinger: Thanks for reviewing this PR. I addressed your comment. One CI test is still failing, but I don’t think it is related to the code changes in this PR.

AlexandreSinger (approved):
LGTM.

I do not agree with having "non-fatal errors"; to me, all errors should be fatal eventually. But if you feel that removing the error is ok, then that's fine with me. I just do not think we should ever have error messages in the log files and have success in execution. Since you removed the error message, then this does not block this PR.

The failing test is irrelevant to this change. Merging anyways.

amin1377:
@AlexandreSinger: In that case, what is the point of having `VTR_LOG_ERROR`? Also, not all warnings are equal, and I think we still need a separate macro for more serious warnings. `VTR_LOG_ERROR` seems like a good candidate for that.

AlexandreSinger:
> In that case, what is the point of having `VTR_LOG_ERROR`?

From the documentation:
image

The reason this does not stop execution is for things like error checkers which go through the entire netlist, find all errors, and then stop execution at the end. Think of the `check_clustering`, `check_placement`, or `check_routing` functions. We want them to print all errors before crashing.

A warning is intended to not end execution, but show that there may be an issue that some users may care about. You are welcome to propose a new form of warning; but to me, errors should always halt, warnings should never halt. I have seen some programs have warning levels, and allow the user to "error on warnings" above a certain level; so it is up to the user to decide how much they care about the warnings.

amin1377:
I see. The definition of `VTR_LOG_ERROR` is clear to me now. Regarding having different levels of warnings, that makes sense to me. We can discuss it later to see how much of a need there is for that.

PR #3682, opened 2026-07-03
Fix interposer cost interaction with placement checkpoints

### LINKED ISSUE #3624: [Interposer] Cost Update During Placement Issue
While transitioning to the new AP flow, the following test was failing:
- vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_interposer/vtr_flagship_interposer/config/config.txt

To prevent it from blocking the upgrade, I locked it to the old flow for now; however this should be fixed. This is likely not due to AP itself, but just that the placement solution found by AP was exercising an untested code path.

There was a similar issue fixed for the hecate interposer test. This should be fixed similarly.

AmirhosseinPoolad:
Placement checkpoints did not work very well with interposer costs. Previously restoring a checkpoint would not update the net bounding boxes, so recomputing the bounding-box dependent costs would actually result in inconsistencies. This PR recomputes the bounding boxes and the interposer congestion cost when restoring a placement checkpoint to make sure the cost is correct and consistent.

The reason we have to recompute the congestion costs from scratch is that the underlying cached data for computing the congestion cost is only updated periodically. We can't just save the congestion cost when saving the checkpoint because when we recompute the cost in the consistency check we would not use the same cached data and get different answers.

Fixes #3624


vaughnbetz (approved):
LGTM.  I don't think this will have a noticable impact on run time, but probably a good idea to run the VTR designs anyway to confirm before merging. Other than that this looks good.

amin1377 (approved):

AmirhosseinPoolad:
QoR is identical between this branch and master:

  | master.txt | branch.txt
-- | -- | --
vtr_flow_elapsed_time | 1 | 0.996877628758161
odin_synth_time |   |  
parmys_synth_time | 1 | 0.986987876875893
abc_depth | 1 | 1
abc_synth_time | 1 | 0.978094984062629
num_clb | 1 | 1
num_memories | 1 | 1
num_mult | 1 | 1
max_vpr_mem | 1 | 1.00017564602258
num_pre_packed_blocks | 1 | 1
device_grid_tiles | 1 | 1
pack_time | 1 | 0.995355823867832
placed_wirelength_est | 1 | 1
place_time | 1 | 0.998056880687503
placed_CPD_est | 1 | 1
min_chan_width | 1 | 1
routed_wirelength | 1 | 1
min_chan_width_route_time | 1 | 0.992460528938062
critical_path_delay | 1 | 1
crit_path_route_time | 1 | 0.996081822011045

PR #3685, opened 2026-07-06
Fix a text drawing issue related to the deferred and rhi renderer

kevinxing-a12:
#### Description

Currently, there are two functions that draw blocks, draw_place() and draw_internal_draw_subblk().
draw_place() is called first during every redraw, and it draws empty and non-empty clustered blocks.
draw_internal_draw_subblk() draws only non-empty clustered blocks and their child blocks, and these clustered blocks overlap with the previous ones.

Previously, draw_place() always draws texts inside the clustered blocks, because the same clustered blocks in draw_internal_draw_subblk() can effectively hide them, and texts drawn in draw_internal_draw_subblk() are the only visible ones.

But this is no longer valid after the Qt upgrade, because in the deferred and rhi renderer, all texts are queued and unleashed together after geometries are drawn (the idea of overlay), meaning that the texts mentioned before cannot be hidden anymore, causing the scene below:

I added a few condition checks inside draw_place() so that texts are drawn only when needed.

Note: the change only stays in the VPR drawing code, and nothing in ezgl was touched.

I have tested this on a few regression tests, and all worked well except the 3d test. This is expected, because when multiple layers are shown together, clustered blocks that share the same 2d location can have overlapped text because of the same overlay reason. If only one layer is shown, there is no issue.

I personally want to fix this in the next PR.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -1679,7 +1679,7 @@ void highlight_moved_block_and_its_terminals(
     clear_colored_locations();
     set_draw_loc_color(blocks_affected.moved_blocks[0].old_loc,
                        OLD_BLK_LOC_COLOR);
-    set_draw_loc_color(blocks_affected.moved_blocks[0].old_loc,
+    set_draw_loc_color(blocks_affected.moved_blocks[0].new_loc,
                        NEW_BLK_LOC_COLOR);
 }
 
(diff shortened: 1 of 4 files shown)


vaughnbetz (approved):
Looks good to me.

AmirhosseinPoolad (approved):

PR #3687, opened 2026-07-06
Report CHANZ segment utilization and multi-layer net counts

soheilshahrouz:


AlexandreSinger (approved):
LGTM, thanks @soheilshahrouz this is exactly what I needed and then some!

PR #3689, opened 2026-07-06
[AP][PL] Added Special Partitioning for Interposers

AlexandreSinger:
The current partial legalizer algorithm uses successive bi-partitioning to quickly legalize the flat placement (using the vector mass abstraction to quickly check if a set of atoms can be contained within a window with a given vector capacity).

Due to this being a bi-partitioner, the first partitions have a large affect on the solution. When targeting interposers, the interposer cut-lines form a natural partition boundary. This PR makes the partial legalizer partition along interposer cut lines first (if they exist).


AlexandreSinger:
I still need to collect direct A/B results for this branch.

vpr/src/analytical_place/global_placement/partial_legalizer.cpp
```diff
@@ -1576,6 +1577,108 @@ PartitionedWindow BiPartitioningPartialLegalizer::partition_window(
     float best_score = -1.0f;
     const std::vector<PrimitiveVectorDim>& dims = dim_grouper_.get_dims_in_group(group_id);
 
+    // If the device has interposer die boundaries, always try to partition
+    // along them first. Even an unbalanced die-boundary cut is preferable to
```
AmirhosseinPoolad: We use the term "interposer cut" for this instead of "die boundary". Would be good to be consistent.

vpr/src/analytical_place/global_placement/partial_legalizer.cpp
```diff
@@ -1576,6 +1577,108 @@ PartitionedWindow BiPartitioningPartialLegalizer::partition_window(
...
+    // spreading blocks across an interposer; the recursive partitioner will
+    // re-balance each sub-window on subsequent splits. We pick the most
+    // balanced interposer cut when multiple boundaries fall in the window.
+    const DeviceGrid& device_grid = g_vpr_ctx.device().grid;
+    if (device_grid.has_interposer_cuts()) {
+        float best_interposer_score = -1.0f;
+        PartitionedWindow best_interposer_partition;
+
+        // Try vertical interposer cuts (die boundaries at constant x).
```
AmirhosseinPoolad: Explain what "try" means (find interposer cuts that are contained inside the window and see which one results in a more balanced partition)

vpr/src/analytical_place/global_placement/partial_legalizer.cpp
```diff
@@ -1576,6 +1577,108 @@ PartitionedWindow BiPartitioningPartialLegalizer::partition_window(
...
+        int min_pivot_x = (int)std::floor(window.region.xmin()) + 1;
+        int max_pivot_x = (int)std::ceil(window.region.xmax()) - 1;
+        std::unordered_set<int> vert_cut_xs;
```
AmirhosseinPoolad, line 1597: Why use a set instead of vector? cut locations are unique per layer.

AmirhosseinPoolad (commented):
Thanks Alex. Some minor comments.

AlexandreSinger:
Results are in and they are pretty good:

  | Num Unroutes | num_nets_crossing_interposer | interposer_used_wires_crossing | critical_path_delay | routed_wirelength | total_heap_pops
Traditional | 3 | 1.00 | 1.00 | 1.00 | 1.00 | 1.00
AP Master | 0 | 0.97 | 0.98 | 1.03 | 0.94 | 1.05
AP + PL Ordering | 1 | 0.88 | 0.88 | 0.97 | 0.93 | 0.89

These results were collected on `hecate_25d_L13_int_10um_bump_fanin_12.xml`, full results can be found here: https://docs.google.com/spreadsheets/d/1fSVeseBE2Nm3REB3WoPr4E6SxO8d9p9qMAoPj5KmCoc/edit?usp=sharing

You will notice that one circuit became unroutable with this change (on L13), which was dla_like.large. We can see why by looking at the per-circuit breakdown of nets crossing (note: the results are normalized to AP master, lower means that this branch reduced the number of nets crossing):

circuit | num_nets_crossing_interposer
-- | --
attention_layer.blif | 1.354
bnn.blif | 0.941
bwave_like.fixed.large.blif | 1.038
clstm_like.large.blif | 0.766
clstm_like.medium.blif | 0.700
clstm_like.small.blif | 0.929
conv_layer_hls.blif | 0.385
dla_like.large.pre-vpr.blif | 1.399
dla_like.medium.blif | 1.148
dla_like.small.blif | 0.954
dnnweaver.blif | 0.887
gemm_layer.blif | 0.920
lstm.blif | 1.045
proxy.5.blif | 1.373
proxy.7.blif | 1.066
tdarknet_like.large.blif | 0.684
tdarknet_like.small.blif | 0.599
tpu_like.large.os.blif | 1.204
tpu_like.large.ws.blif | 0.957

You can see that many of the circuits improved by a large degree, but some increased. Unfortunately dla_like.large was close to the edge and got pushed over. I plan to investigate how we can improve this in the future.

To compare all results (traditional, AP master, and this branch) I dropped the 3 circuits that they commonly failed. We can see that this branch improves the nets crossing the interposer by around 10%, improves the CPD by 3% (compared to traditional flow, even more compared to AP), and improves WL by 1%. Overall this is an improvement. I think the dla_like.large is a fluke that may represent some noise in the system to investigate.

vpr/src/analytical_place/global_placement/partial_legalizer.cpp
AlexandreSinger: This is because there are multiple layers. Suppose we had a 3D architecture with interposer cuts at x = 5. Since we iterate over each layer, we would get two cuts at x = 5; but the partial legalizer cannot do different cuts on each layer.

I recognize that we currently do not support this feature yet; but I do not think we are bleeding too much performance here by using a hash set instead of a vector. What do you think?

AmirhosseinPoolad: Let's leave it, I think it's fine.

AmirhosseinPoolad:
Wanted to see if there's any bias towards how the circuit utilizes the device. The limiting resource uses the baseline SA packing and AP results might be different. Can't really see any patterns myself but here's the data.

circuit | num_nets_crossing_interposer | device_limiting_resources
-- | -- | --
attention_layer.blif | 1.354 | memory
bnn.blif | 0.941 | clb
bwave_like.fixed.large.blif | 1.038 | dsp_top
clstm_like.large.blif | 0.766 | clb
clstm_like.medium.blif | 0.700 | clb
clstm_like.small.blif | 0.929 | clb
conv_layer_hls.blif | 0.385 | io
dla_like.large.pre-vpr.blif | 1.399 | clb
dla_like.medium.blif | 1.148 | clb
dla_like.small.blif | 0.954 | clb
dnnweaver.blif | 0.887 | memory
gemm_layer.blif | 0.920 | clb
lstm.blif | 1.045 | io clb
proxy.5.blif | 1.373 | dsp_top
proxy.7.blif | 1.066 | clb
tdarknet_like.large.blif | 0.684 | memory
tdarknet_like.small.blif | 0.599 | memory
tpu_like.large.os.blif | 1.204 | dsp_top
tpu_like.large.ws.blif | 0.957 | dsp_top

AlexandreSinger:
@AmirhosseinPoolad That was a good idea to check. I also checked the number of post-packed blocks, they all reduced by 1-4%; no real correlation as far as I can tell.

AlexandreSinger:
@vaughnbetz FYI

PR #3691, opened 2026-07-07
Remove unused vectors in median move generators

soheilshahrouz:
Removes two unused vectors.
Removes an if statements that is always false. The same check is done outside loop to skip ignored nets.


vaughnbetz (approved):
LGTM

PR #3698, opened 2026-07-08
Improve CRR Build Time

amin1377:
Improve CRR build time by:

* Removing `sb_count_dir`, since it was not being used and its string operations were causing significant runtime overhead.
* Removing the node lookup manager and replacing it with a compiled CRR template. This significantly reduces runtime because all operations are numerical, with no string operations involved during CRR construction.
* Making a few additional minor optimizations.

These changes reduce CRR build time by at least 3× on a 30×30 device, with even greater performance improvements on larger devices.


vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_compiled_template.h
```diff
@@ -0,0 +1,116 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "rr_node_types.h"
+#include "crr_common.h"
+#include "data_frame_processor.h"
+
```
vaughnbetz, line 37: I suggest adding a file scope comment here explaining what this file is for, and how it works. The 'compiled' template/dataframe etc. is a key concept here but I don't think it is every explained. What does compiled mean in this context and what is the overall optimization / idea?

vaughnbetz (approved):

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_compiled_template.cpp
```diff
@@ -0,0 +1,167 @@
...
+        }
+        for (size_t col = NUM_EMPTY_COLS; col < df.cols(); ++col) {
+            if (sink_index[col] < 0) {
+                continue;
+            }
+            const Cell& cell = df.at(row, col);
+            if (cell.is_empty()) {
+                continue;
+            }
+            cells_.push_back({source_index[row], sink_index[col], parse_cell_delay(cell, is_annotated)});
```
AlexandreSinger, line 52: If you are going for performance in this PR, I recommend reserving the size of cells_ before doing all of these push_backs. I think before the for loop you can reserve the size of cells_ to be `df.rows() * df.cols()` or something.

Also, cells_ contains a complex struct. I recommend that you use emplace back instead of push back here.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_compiled_template.cpp
```diff
@@ -0,0 +1,167 @@
...
+            ptc += (spec.direction == Direction::DEC) ? PTCS_PER_TRACK_PAIR * (spec.seg_len - 1) : 0;
+            spec.base_ptc = ptc;
+        }
+
+        prev_seg_type = seg_type;
+        prev_side = spec.side;
+
+        if (emit_spec) {
+            spec_index[i] = static_cast<int32_t>(specs.size());
+            specs.push_back(std::move(spec));
```
AlexandreSinger, line 148: Same comment here. You know how many things are getting pushed back ahead of time. But push_back + move should be good.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
```diff
@@ -171,15 +132,12 @@ class CRRConnectionBuilder {
                                        int& truncated,
                                        bool is_vertical) const;
 
-    Direction get_direction_for_side(e_sw_template_dir side, bool is_vertical) const;
-    std::string get_segment_type_label(e_sw_template_dir side) const;
+    /// Compiled templates, one per unique dataframe
+    std::unordered_map<const DataFrame*, std::unique_ptr<CompiledTemplate>> compiled_templates_;
```
AlexandreSinger, line 136: This is ok, but I normally try to avoid hashing on pointers if we can. Does the dataframe not have a unique identifier like an ID or name?

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
```diff
@@ -171,15 +132,12 @@ class CRRConnectionBuilder {
...
 
-    // Return the switch id of an edge between two nodes
-    int get_connection_delay_ps(const std::string& cell_value,
-                                const std::string& sink_node_type,
-                                RRNodeId source_node,
-                                RRNodeId sink_node,
-                                int segment_length = -1) const;
+    /// Compiled template per SB_MAPS pattern index (nullptr when the pattern
+    /// has no template file, i.e. no connections)
+    std::vector<const CompiledTemplate*> template_by_pattern_;
```
AlexandreSinger, line 140: If this is the same pointer that is in compiled_templates_, why not use shared pointers?

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.cpp
```diff
@@ -72,392 +54,236 @@ CRRConnectionBuilder::CRRConnectionBuilder(const RRGraphView& rr_graph,
...
+        const DataFrame* df = sb_manager_.get_dataframe_by_index(pattern_idx);
+        if (df == nullptr) {
+            template_by_pattern_.push_back(nullptr);
+            continue;
+        }
+        std::unique_ptr<CompiledTemplate>& compiled = compiled_templates_[df];
+        if (!compiled) {
+            compiled = std::make_unique<CompiledTemplate>(*df, is_annotated_);
+        }
+        template_by_pattern_.push_back(compiled.get());
```
AlexandreSinger, line 75: See comment above on shared pointers.

AlexandreSinger (commented):
LGTM @amin1377 , a couple of small comments below.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_compiled_template.h
amin1377: Add the following comment to the file:

```
/**
 * @file crr_compiled_template.h
 * @brief A switch template, parsed once up front so it can be applied cheaply
 *        at every tile.
 *
 * A switch block is described by a *switch template*: a CSV file read into a
 * DataFrame (see data_frame_processor.h) in which each row is a connection
 * source, each column a connection sink, and a non-empty cell says "connect
 * this source to this sink". The row and column headers are text - a side
 * (LEFT/TOP/...), a segment type such as "L4", a tap position or a pin name.
 *
 * The same template is applied at every grid location matching its pattern,
 * which on a large FPGA is hundreds of thousands of tiles. Most of the work of
 * reading it gives the same answer at every one of them: parsing those header
 * strings, working out each segment's base PTC, and dropping rows, columns and
 * cells that can never resolve to a node. Only the final step is genuinely
 * per-tile - clipping a segment at the device edge, and looking a pin name up
 * in the tile type actually present at (x, y).
 *
 * "Compiling" a template means doing the tile-independent part once and keeping
 * the result as plain numbers: a CompiledSegSpec per usable row and column, a
 * CompiledCell per connection. Each tile then only resolves those specs to
 * nodes and walks the cells, so no string is ever parsed twice.
 *
 * Built by CRRConnectionBuilder::initialize(), used by
 * CRRConnectionBuilder::get_tile_connections().
 */
```

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
amin1377: We don't have a unique ID for it, and we don't iterate over this data structure. I see your point about not hashing addresses, but given that we don't iterate over it, I think we can keep it as is.

vpr/src/route/rr_graph_generation/tileable_rr_graph/crr_generator/crr_connection_builder.h
amin1377: These pointers serve different purposes: `compiled_templates_` owns the templates, while `template_by_pattern_` is just a non-owning index. The lifetime is already safe due to declaration order, and the pointers never escape the class, so `shared_ptr` would only add unnecessary refcounting.

AlexandreSinger (approved):
LGTM

PR #3699, opened 2026-07-08
Small RR Graph Improvements

amin1377:
Fix some inefficient function calls during RR graph construction.

### DIFF
--- a/libs/librrgraph/src/base/check_rr_graph.cpp
+++ b/libs/librrgraph/src/base/check_rr_graph.cpp
@@ -121,20 +121,17 @@ void check_rr_graph(const RRGraphView& rr_graph,
             }
         } /* End for all edges of node. */
 
-        std::sort(edges.begin(), edges.end(), [](const std::pair<int, int>& lhs, const std::pair<int, int>& rhs) {
+        auto edge_dest_less = [](const std::pair<int, int>& lhs, const std::pair<int, int>& rhs) {
             return lhs.first < rhs.first;
-        });
+        };
+        std::sort(edges.begin(), edges.end(), edge_dest_less);
 
         // Check that multiple edges between the same from/to nodes make sense.
         // Walk the sorted edge list one group of same-destination edges at a time.
         for (auto group_first = edges.cbegin(); group_first != edges.cend();) {
             int to_node = group_first->first;
-            auto group_last = group_first + 1;
-            while (group_last != edges.cend() && group_last->first == to_node) {
-                ++group_last;
-            }
-            auto range = std::make_pair(group_first, group_last);
-            group_first = group_last;
+            auto range = std::equal_range(group_first, edges.cend(), *group_first, edge_dest_less);
+            group_first = range.second;
 
             size_t num_edges_to_node = std::distance(range.first, range.second);
 
--- a/libs/librrgraph/src/base/check_rr_graph.cpp
+++ b/libs/librrgraph/src/base/check_rr_graph.cpp
@@ -125,18 +125,20 @@ void check_rr_graph(const RRGraphView& rr_graph,
             return lhs.first < rhs.first;
         });
 
-        //Check that multiple edges between the same from/to nodes make sense
-        for (int iedge = 0; iedge < num_edges; iedge++) {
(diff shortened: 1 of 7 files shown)


libs/librrgraph/src/base/rr_switch.h
```diff
@@ -44,16 +44,16 @@ struct t_rr_switch_inf {
...
     /// Returns true if this switch type isolates its input and output into
     /// separate DC-connected subcircuits
-    bool buffered() const;
+    bool buffered() const { return switch_type_is_buffered(type()); }
 
     /// Returns true if this switch type is configurable (i.e. can the switch can be turned on or off)
     /// This allows modelling of non-optional switches (e.g. fixed buffers, or shorted connections)
     /// which must be used (e.g. expanded by the router) if a connected segment is used.
-    bool configurable() const;
+    bool configurable() const { return switch_type_is_configurable(type()); }
```
AmirhosseinPoolad, line 56: Should make these 'inline' to avoid linker issues.

libs/librrgraph/src/base/check_rr_graph.cpp
```diff
@@ -125,19 +125,21 @@ void check_rr_graph(const RRGraphView& rr_graph,
...
+        // Check that multiple edges between the same from/to nodes make sense.
+        // Walk the sorted edge list one group of same-destination edges at a time.
+        for (auto group_first = edges.cbegin(); group_first != edges.cend();) {
+            int to_node = group_first->first;
+            auto group_last = group_first + 1;
+            while (group_last != edges.cend() && group_last->first == to_node) {
+                ++group_last;
+            }
+            auto range = std::make_pair(group_first, group_last);
+            group_first = group_last;
```
AmirhosseinPoolad: I think it would be better if you kept the std::equal_range and used it for the manual loop indexing instead of linearly iterating the array.

vpr/src/route/rr_graph_generation/tileable_rr_graph/chan_node_details.cpp
```diff
@@ -33,56 +33,52 @@ ChanNodeDetails::ChanNodeDetails() {
...
  ***********************************************************************/
 size_t ChanNodeDetails::get_chan_width() const {
-    VTR_ASSERT(validate_chan_width());
+    VTR_ASSERT_SAFE(validate_chan_width());
     return track_node_ids_.size();
 }
 
 size_t ChanNodeDetails::get_track_node_id(const size_t& track_id) const {
-    VTR_ASSERT(validate_track_id(track_id));
+    VTR_ASSERT(track_id < track_node_ids_.size());
```
AmirhosseinPoolad, line 41: Wouldn't it be safer if you kept "validate_track_id(track_id)" but changed the assertion to be safe? Same comment for all the other getter methods.

vpr/src/route/rr_graph_generation/tileable_rr_graph/chan_node_details.cpp
```diff
@@ -33,56 +33,52 @@ ChanNodeDetails::ChanNodeDetails() {
...
 }
 
 /* Return a copy of vector */
 std::vector<size_t> ChanNodeDetails::get_track_node_ids() const {
-    std::vector<size_t> copy;
-    for (size_t inode = 0; inode < get_chan_width(); ++inode) {
-        copy.push_back(track_node_ids_[inode]);
-    }
-    return copy;
+    return track_node_ids_;
```
AmirhosseinPoolad, line 47: Any particular reason for not returning a const reference?

libs/librrgraph/src/base/rr_switch.h
amin1377: Since the definition is inside the class body, I don't think we need to add the inline keyword here

vpr/src/route/rr_graph_generation/tileable_rr_graph/chan_node_details.cpp
amin1377: I don’t think this is necessary. The `validate_chan_width` check is already performed in the `get_chan_width()` function. Elsewhere, we should only need to ensure that no out-of-bounds access occurs.

AlexandreSinger (approved):
LGTM

Issue #3696, opened 2026-07-08
[CI] ISPD Weekly Test Failure


AlexandreSinger:
The ISPD weekly tests (NightlyTest 5) are failing in CI:
https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/28734016991/job/85204862308

<img width="1251" height="630" alt="Image" src="https://github.com/user-attachments/assets/fc746e70-d40d-402a-afb5-874337cdab17" />

NOTE: NightlyTest 2 and 7 are failing for different reasons which should have been resolved already.

They all fail with similar packing failures:
<img width="840" height="310" alt="Image" src="https://github.com/user-attachments/assets/6d18aef9-db0b-4f1d-aec1-7277733abf6f" />

I think something changed in packing that is causing these molecules to not be packed anymore.

AlexandreSinger:
@haydar-c This was the issue I discussed in the team meeting. My guess is that your cleanups may have caused this. I am wondering if there was a bug in this architecture, or a quirk of the algorithm that changed. Could you take a look?

@vaughnbetz FYI.

AlexandreSinger:
NightlyTest5 is now passing in the weekly tests: https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/29184294994/job/86627511976

Thanks for your help @haydar-c !

PR #3700, opened 2026-07-09
[Pack] Fix pin class regression for primitive clock pins fed through <input> ports

haydar-c:
#### Problem
The pin class refactor (#3675) required is_clock to match when counting boundary pins into a pin class. For archs that feed a primitive `<clock>` pin through an intermediate `<input>` port (e.g. FF.C in ultrascale_ispd.xml), the clock's pin class got zero supply, so the pin feasibility filter rejected every candidate and packing failed as in #3696. 

#### Solution
For the primitive `<clock>` pins, instead of requiring `<clock>` port connections on the pb boundary, allow `<input>` connections as well. That matches the old behaviour.

I am not sure if we should require the primitive `<clock>` pins to be fed by ports declared as `<clock>` in the future. That would be more intuitive and enable us to create clock pins to their own classes (currently we have only input pin classes and output pin classes where clock pins are considered in the input pin classes, see [related TODO](https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/fe6553e9be347daedc1897859d48111807a578a2/vpr/src/pack/cluster_feasibility_filter.cpp#L426)). 

#### Related Issue
#3696

### DIFF
--- a/vpr/src/pack/cluster_feasibility_filter.cpp
+++ b/vpr/src/pack/cluster_feasibility_filter.cpp
@@ -56,6 +56,10 @@ static void unmark_fanout_intermediate_nodes(t_pb_graph_pin* current_pb_graph_pi
  * parent_pin_class[depth]. All reachable boundary pins of pb_graph_node of the same type
  * are assigned class_id via pin_class. Intermediate pins are traversed but not assigned.
  *
+ * Note: boundary pins are matched on port type only (is_clock is ignored), since we do
+ * not expect every connection to a primitive clock pin to be declared <clock> at all
+ * levels of the hierarchy.
+ *
  * @param seed_pin      A primitive pin that seeds the BFS.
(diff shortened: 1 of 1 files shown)


haydar-c:
You can see [this ](https://github.com/verilog-to-routing/vtr-verilog-to-routing/actions/runs/29021518309) regular test.

AlexandreSinger:
Thanks @haydar-c ! This makes sense to me why it would fail on only this architecture. I do know other architectures which also use a pattern similar to this, so I am happy that we caught it!

Can you please run the ultrascale_ispd tests locally to verify that this fixed those tests? Other than that I think this can be merged.

vaughnbetz (approved):
LGTM

haydar-c:
I have locally verified that the ultrascale_ispd tests are also passing with this change. @AlexandreSinger
ispd_img
ispd_test

AlexandreSinger (approved):
LGTM

PR #3704, opened 2026-07-12
Add critical path drawing (flylines and delay labels) to the AP stage.

kevinxing-a12:
#### Description

Make critical path drawing (flylines and delay labels) available in the analytical placement stage.

The flyline endpoints are centered on the AP blocks containing the atom blocks of the associated timing nodes. This is made possible by creating an atom block to AP block reverse lookup inside APNetlist. If two timing nodes collapse to the same AP blocks, the flyline and delay label won't be drawn.

The lookup is constructed in gen_ap_netlist_from_atoms(). To avoid a two-stage construction before and after compress() is called, the lookup is constructed in one go by looping through the final APNetlist and the molecules it contains after compress().

Previously, the function place_clusters() (member of NaiveFullLegalizer) constructed the exact same reverse lookup locally, but since this lookup has become part of the APNetlist, I tentatively deleted the local version and let the function use the new one. 

Finally, all flylines and delay labels are drawn as opaque in any 3d device, because the APBlocks they connect can reside at continuous layer levels. This also aligns with how APBlocks (the little black boxes) are currently drawn in a 3d context.
#### Motivation and Context

This update can hopefully aid a little help to the timing analysis in analytical placement.

Tested with multiple 3d and non-3d regression tests. The graphics worked fine.

#### Types of changes

#### Checklist:

## Summary by CodeRabbit

* **New Features**
  * Analytical placement views now show initial and final placement scenes with improved visualization controls.
  * Critical-path graphics and timing information are available during analytical placement.
  * Flylines and delay labels now use placement-aware coordinates and visibility rules.
* **Bug Fixes**
  * Improved rendering of cross-layer connections and routed critical-path graphics.
  * Prevented misleading flylines and labels when endpoints map to the same placement block.
* **Documentation**
  * Clarified that delay labels apply to edge flylines, not routed connections.


vpr/src/analytical_place/common/gen_ap_netlist_from_atoms.cpp
```diff
@@ -265,5 +265,8 @@ APNetlist gen_ap_netlist_from_atoms(const AtomNetlist& atom_netlist,
     // Verify that the netlist was created correctly.
     VTR_ASSERT(ap_netlist.verify());
 
+    // Build a reverse lookup from atom block Id to AP block Id. Used by the full legalizer and external drawing code.
+    ap_netlist.build_atom_block_ap_block(atom_netlist.blocks().size(), prepacker);
```
vaughnbetz: Probably should move this before the verify() call, and have the verify() method check that the lookup makes sense (the AP block that each atom block refers to exists, and it contains that atom block according to the AP netlist).

vpr/src/analytical_place/global_placement/global_placer.cpp
```diff
@@ -397,12 +397,16 @@ PartialPlacement SimPLGlobalPlacer::place() {
...
-
-    // Pause to show initial FPGA state before any solving begins.
-    draw_manager.pause("Analytical Placement: Starting Global Placement");
+    // Initialize graphics for analytical placement, setting the references in the draw state.
+    APDrawManager draw_manager(p_placement, ap_netlist_);
+
+    // Pause to show initial FPGA state before any solving begins. Also pass in the shared timing info pointer
+    // when timing analysis is on so that the drawing code can prepare for critical path drawing.
+    if (pre_cluster_timing_manager_.is_valid()) {
+        draw_manager.pause_at_initial_scene("Analytical Placement: Starting Global Placement", pre_cluster_timing_manager_.get_timing_info_ptr());
```
vaughnbetz: Could shorten this code if the .get_timing_info_ptr() call returned nullptr when the pre_cluster_timing_manager isn't valid.  Not sure if that would be a good or bad change though.  I think @AlexandreSinger wrote the pre-cluster timing manager so I'll defer to him.

vpr/src/analytical_place/ap_draw_manager.cpp
```diff
@@ -33,23 +33,39 @@ void init_ap_graphics(const t_vpr_setup& vpr_setup, const t_arch& arch) {
 #endif
 }
 
-APDrawManager::APDrawManager(const PartialPlacement& p_placement) {
+APDrawManager::APDrawManager(const PartialPlacement& p_placement, const APNetlist& ap_netlist) {
 #ifndef NO_GRAPHICS
     // Set the analytical placement reference in draw state
     get_draw_state_vars()->set_ap_partial_placement_ref(p_placement);
+    // Set the AP netlist reference in draw state
```
vaughnbetz: Probably better to make one slightly higher level comment here describing both routines:
// Set references in draw_state to analytical placement variables we will need for drawing

That comment could replace the two current comments, which are pretty close to just repeating the function name.

vpr/src/analytical_place/ap_draw_manager.cpp
```diff
@@ -33,23 +33,39 @@ void init_ap_graphics(const t_vpr_setup& vpr_setup, const t_arch& arch) {
...
+    get_draw_state_vars()->set_ap_netlist_ref(ap_netlist);
 #else
     (void)p_placement;
+    (void)ap_netlist;
 #endif
 }
 
 APDrawManager::~APDrawManager() {
 #ifndef NO_GRAPHICS
     // Clear the analytical placement reference in draw state
```
vaughnbetz: Same comment as above: make one higher level comment instead of these two low level ones.

// Clear references to AP data structures that graphics redraws use, since those data structures are being destroyed

vpr/src/draw/draw.cpp
```diff
@@ -307,6 +307,8 @@ static void draw_main_canvas(ezgl::renderer* g) {
         }
     } else {
         draw_analytical_place(g);
+        // Note: during the AP stage, only the critical path flylines and delay labels are available.
```
vaughnbetz: Remove commented out code.  Comment also seems out of place.

vpr/src/draw/draw_basic.h
```diff
@@ -112,4 +108,6 @@ void draw_reset_blk_colors();
 //Reset a specific block's colour.
 void draw_reset_blk_color(ClusterBlockId blk_id);
 
+ezgl::point2d get_ap_block_draw_coord(APBlockId ap_block);
```
vaughnbetz: Add a doxygen comment.

vpr/src/draw/draw_types.h
```diff
@@ -445,11 +446,18 @@ struct t_draw_state {
      */
     std::optional<std::reference_wrapper<const PartialPlacement>> ap_partial_placement_ref_;
 
+    std::optional<std::reference_wrapper<const APNetlist>> ap_netlist_ref_;
```
vaughnbetz: Add a doxygen comment for this variable

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -323,14 +370,31 @@ static void draw_timing_edge_flylines(const tatum::TimingPath& path, ezgl::rende
...
-            int sink_block_layer = get_timing_path_node_layer_num(node);
-            t_draw_layer_display flyline_visibility = get_element_visibility_and_transparency(src_block_layer, sink_block_layer);
+            t_draw_layer_display flyline_visibility = get_timing_flyline_visibility(prev_node, node);
 
             if (flyline_visibility.visible) {
                 g->set_color(color, flyline_visibility.alpha);
-                ezgl::point2d start = tnode_draw_coord(prev_node);
-                ezgl::point2d end = tnode_draw_coord(node);
+
+                // Calculate the physical locations of the two nodes.
```
vaughnbetz: I think this could (and probably should) be factored into a helper routine to shorten this routine.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -540,11 +605,29 @@ static std::vector<t_label_drawing_info> calculate_basic_label_drawing_info(cons
             drawing_info.delay_label_str = delay_label_str;
 
             // Calculate the physical locations of the two nodes.
-            ezgl::point2d start = tnode_draw_coord(prev_node);
-            ezgl::point2d end = tnode_draw_coord(node);
+            ezgl::point2d start, end;
```
vaughnbetz: This looks like mostly repeated code that the helper routine I recommended above would remove.

vaughnbetz (changes requested):
Looks good overall; some suggested changes.

vpr/src/analytical_place/common/ap_netlist_utils.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "atom_netlist_fwd.h"
+#include "ap_netlist_fwd.h"
+#include "vtr_vector.h"
+
+class Prepacker;
+
+
+class AtomBlockAPBlockLookup {
```
vaughnbetz, line 23: You should have more doxygen comments, fully documenting the class.
What is the class for?
What is verify for?
Comment the two data members.

vpr/src/analytical_place/ap_draw_manager.h
```diff
@@ -57,7 +61,19 @@ class APDrawManager {
...
+    void pause_at_initial_scene(const std::string& msg, PreClusterTimingManager& timing_manager);
+
+    /**
+     * @brief Pause and wait at the final global placement scene.
+     */
+    void pause_at_final_scene(const std::string& msg);
+
+  private:
+
+    AtomBlockAPBlockLookup atom_block_ap_block_lookup_;
```
vaughnbetz, line 81: Add a doxygen comment on the data member.

vpr/src/draw/draw_types.h
```diff
@@ -445,11 +452,18 @@ struct t_draw_state {
      */
     std::optional<std::reference_wrapper<const PartialPlacement>> ap_partial_placement_ref_;
 
+    std::optional<std::reference_wrapper<const AtomBlockAPBlockLookup>> atom_block_ap_block_lookup_ref_;
```
vaughnbetz: This should have a doxygen comment. Say what this lookup is for, and whyit is an optional and reference wrapper and const (I'm guessing you didn't want to use a pointer and this might not be set so this lets you check for nullptr for not set, but you should say that).

vaughnbetz (changes requested):
A few more comments that I think you should add.

kevinxing-a12:
Sorry! I have been working on the comments but forgot to add a reminder here.

vpr/src/analytical_place/common/ap_netlist_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+            const t_pack_molecule& block_mol = prepacker.get_molecule(block_mol_id);
+            for (AtomBlockId atom_block_id : block_mol.atom_block_ids) {
+                // See issue #2791, some of the atom_block_ids may be invalid. They
+                // can safely be ignored.
+                if (!atom_block_id.is_valid())
+                    continue;
+                // Safety check.
+                VTR_ASSERT(static_cast<std::size_t>(atom_block_id) < atom_block_ap_block_.size());
+                // Ensure that this block is not in any other AP block. That would be weird.
+                VTR_ASSERT(!atom_block_ap_block_[atom_block_id].is_valid());
```
AlexandreSinger: I would elevate these two asserts to ASSERT_SAFEs.

vpr/src/analytical_place/common/ap_netlist_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+                    atom_block_id_found = true;
+                    break;
+                }
+            }
+            // Early exit.
+            if (atom_block_id_found)
+                break;
+        }
+        // Something unexpected must have happened if we did not find the atom block id.
+        VTR_ASSERT(atom_block_id_found);
```
AlexandreSinger: Using asserts for error checks are not a great idea since people can turn off these asserts. I recommend using "VTR_LOG_ERROR("...")" here to say something went wrong (it wont halt the program), and then return a count of how many errors happened. Outside of the verify function, you can assert that the returned count is 0.

vpr/src/analytical_place/common/ap_netlist_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    verified_ = true;
+}
+
+APBlockId AtomBlockAPBlockLookup::get_ap_block(const AtomBlockId atom_block_id) const {
+    // We should not use this getter without having verified the lookup.
+    VTR_ASSERT_MSG(verified_, "Must verify the lookup using verify() first!");
```
AlexandreSinger: I would remove this assert. The reason you verify this structure is to check that the construction is not wrong, thats fine, but it is also to verify that the AtomNetlist and APNetlist have not invalidated it. By storing the boolean you are ignoring the second issue.

I would remove this, and instead just have an external assert on the verify method.

AlexandreSinger (commented):
Looks great @kevinxing-a12 ! I like the new class you have added.

I had one comment about the verify method. The style I recommended for the verify method you wrote matches what we do for "verify_clustering": https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/master/vpr/src/pack/verify_clustering.cpp

vpr/src/analytical_place/common/ap_netlist_utils.h
```diff
@@ -0,0 +1,62 @@
...
+#include "vtr_vector.h"
+
+// Forward declaration
+class Prepacker;
+
+/**
+ * @brief A lookup used to map an atom block id to the AP block id that contains it.
+ * 
+ * The class constructs and owns a vtr::vector storing (AtomBlockId, APBlockId) pairs by iterating through the APNetlist.
+ * This vector is where the lookup essentially happens. After construction, it is required to call the verify() method
```
vaughnbetz: Is this comment out of date?  

The code doesn't seem to enforce this, and I don't think it should ... we normally can call a verify function whenever we want for safety / debugging, but it isn't required to call it in specific points.

If you wanted to track that the mapping is valid, you would need to add a valid flag to the class and invalidate it whenever there was a change to the AP netlist that would make the mapping invalid.  I think that will be hard to ensure everywhere (need to check/instructment any routine that could possibly modify the AP netlist) though so I wouldn't bother with that.

vaughnbetz (changes requested):
I think there's an out of date comment.

vpr/src/analytical_place/common/ap_netlist_utils.h
kevinxing-a12: Yes! Sorry about that. I have updated it.

kevinxing-a12:
@vaughnbetz I think this PR is ready to merge.

vaughnbetz (approved):
LGTM

Issue #3694, opened 2026-07-07
Code cleanup: remove force_site and flat_site_index from pb_graph_node / cluster legalizer


vaughnbetz:
There is some dead code in the cluster legalizer and pb_graph.
https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/5d44932717cdb8ec8c5bbb0b9dbb45f8f2249f53/vpr/src/pack/cluster_placement.cpp#L245
 -- force_site is never used (always kept at default of -1, which is inactive).

https://github.com/verilog-to-routing/vtr-verilog-to-routing/blob/5d44932717cdb8ec8c5bbb0b9dbb45f8f2249f53/libs/libarchfpga/src/physical_types.h#L1281
  -- flat_site_index in t_pb_graph node is loaded but never used. It can be deleted, I think, along with functions that use it or look at it.

compute_flat_index_for_child_node() is also dead at that point.

#### Proposed Behaviour
Remove this dead code.  Take a quick look to see if any other related code becomes dead at that point too.

AlexandreSinger:
Resolved in #3707

PR #3707, opened 2026-07-14
[Pack] Code cleanup: remove force_site and flat_site_index from pb_graph_node / cluster legalizer

haydar-c:
See #3694 for the related issue. 

Removes the dead force_site / flat_site_index code from the packer

The flat_site_index was originally added for flat placement reconstruction. The new version of reconstruction flow (flat-recon) does not use the flat site index and flat site index information has already been dropped from the FlatPlacementInfo and  from [flat placement file format](https://docs.verilogtorouting.org/en/latest/vpr/file_formats/#flat-placement-file-format-fplace) as well. If a user wants to assign an atom to a specific primitive site, they can use the [logic block location option](https://docs.verilogtorouting.org/en/latest/vpr/placement_constraints/#logical-block-location).


AlexandreSinger (approved):
LGTM, thanks @haydar-c

PR #3708, opened 2026-07-14
Chain with fanout

amin1377:
In all current VTR architectures, the `cout` and `cin` pins are connected directly without any fanout. Although, in the actual SIV architecture, `cout` can drive both `cin` and a data output pin, this feature is currently disabled (lines 6564–6568 of the SIV architecture file).

In this PR, direct connections for which both endpoints have an `fc` value of zero are marked. During molecule creation, if a connection belongs to a pattern associated with one of these marked direct connections, the pattern is not dropped even when the connection has a fanout greater than one.


vpr/src/pack/pack_patterns.h
```diff
@@ -79,6 +89,17 @@ struct t_pack_pattern_connections {
...
+ *      chain_link_fc_zero : true if the only way this chain can hop from one cluster to the next
+ *                          is the dedicated chain wiring, because the cluster input pins that
+ *                          feed the chain (Ex. a CLB's cin pin) cannot reach general routing at
+ *                          all (Fc = 0). When true, the prepacker keeps chains together even if
+ *                          the chain net has extra fanout, regardless of what that fanout
+ *                          drives: the hop has to use the dedicated wire no matter what, so
+ *                          splitting the chain would not make the net any easier to route.
+ *                          Caveat: only the representative tile (pick_physical_type) of the
+ *                          block type the pattern was found in is checked (see
+ *                          pattern_chain_input_has_zero_fc in prepack.cpp).
```
AlexandreSinger, line 101: Is there a way that we can detect this pattern and warn / error on that case? This seems like something that can be hit and should be warned against in-case it causes a crash.

I propose that we detect if a chain_link_fc_zero molecule is found to exist in multiple tiles and warn to the user why that may be an issue and what they may expect to see as a result.

vtr_flow/tasks/regression_tests/vtr_reg_strong/strong_chain_cout_fanout/config/config.txt
```diff
@@ -0,0 +1,33 @@
...
+parse_file=vpr_fixed_chan_width.txt
+
+# How to parse QoR info
+qor_parse_file=qor_fixed_chan_width.txt
+
+# Pass requirements
+pass_requirements_file=pass_requirements_fixed_chan_width.txt
+
+# Script parameters
+script_params=-starting_stage vpr --route_chan_width 100 --pack_verbosity 2
```
AlexandreSinger, line 33: Small note: this channel width seems high for this small blif file. Was this intentional?

AlexandreSinger (commented):
Looks good @amin1377 ! I appreciate what you are trying to do here. You are using Fc=0 as an indicator that something MUST be made into a molecule in some way and forming molecules accordingly (in this special case). Eventually we should harden this in some way  since the prepacker should probably emit an error if some primitive has no route to the general fabric and no molecule has been created. I am not sure what that may look like though and what the implications of that may be. This is a step in the right direction though.

I had a small comment below to prevent failures on reasonable architectures.

Could you please run a larger circuit like MCML or LU32 on the Flagship architecture and check that the prepacker time has not changed too much? I am a little concerned about the run time implication of the BFS graph traversal you are doing for each root node; but I was not sure how often that would be called.

vpr/src/pack/pack_patterns.h
```diff
@@ -40,6 +40,13 @@ struct t_pack_pattern_block {
...
  *      to_block   : block driven by this connection
  *      to_pin     : specific pin in the to_block driven by this connection
+ *      is_fc0_chain_link : set on every chain connection (Ex. an adder's cout feeding the next
+ *                          adder's cin) of a pattern that has chain_link_fc_zero set (see
+ *                          t_pack_patterns below). The prepacker normally refuses to grow a
+ *                          molecule across a net that has more than one sink; on these
+ *                          connections it grows the molecule anyway, because the chain hop 
+ *                          has to use the dedicated wire no matter what. Set by
+ *                          mark_fc0_chain_link_connections in prepack.cpp.
  *      next       : next connection in the linked list
```
vaughnbetz, line 50: Probably better to move the member-by-member doxygen comments to come with the member definitions, below.

vaughnbetz (commented):
Not many comments on the code (just one tiny nit).
It seems we're kind of using the Fc = 0 between clusters as a side method to figure out this is a carry chain with limited connectivity?  The tests are actually packing adders that feed both a LUT or FF and another adder (next in the ripple) within the cluster, so they don't have much to do with the fact that the between CLB carry chains are dedicated routing. Or at least I think that's what they're doing.  Doesn't that require connectivity from cout within the cluster primitives to the LUT / FF logic. And wouldn't that then mean we'd want a check for within-cluster connectivity to decide if this pack pattern is alowed?

amin1377:
Closing this PR since, after discussing it with @vaughnbetz, we concluded that the solution implemented in PR #3722 also addresses this issue and is a cleaner approach.

PR #3710, opened 2026-07-15
Update physical pin cache when restoring checkpoint

amin1377:
#3682 added a call to comp_bb_cong_cost() inside restore_best_placement(), right after the checkpoint's block locations are restored. However, the cached physical pin indices (physical_pins_ in BlkLocRegistry) are only re-synced later, at the end of Placer::place(). So the cost recomputation derives pin coordinates from restored block locations combined with pin indices belonging to the pre-restore placement. If any block occupies a different sub-tile in the checkpoint than in the final placement (common for I/O blocks, where several share a tile), the stale pin index reads pin_width_offset out of bounds, producing garbage bounding-box coordinates, which then index the channel-width arrays far out of range and crash VPR with a segfault.

### DIFF
--- a/vpr/src/place/place_checkpoint.cpp
+++ b/vpr/src/place/place_checkpoint.cpp
@@ -70,6 +70,13 @@ void restore_best_placement(PlacerState& placer_state,
 
         costs = placement_checkpoint.restore_placement(placer_state.mutable_block_locs(), placer_state.mutable_grid_blocks());
 
+        // The restored placement may put blocks in different tiles/sub-tiles than
+        // the placement they replaced, so the cached physical pin indices must be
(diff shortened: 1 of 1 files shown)


AmirhosseinPoolad:
Looks good, thanks for the fix. Can you do a quick run of the VTR benchmarks just to be safe?

amin1377:
@AmirhosseinPoolad: I just ran vtr_reg_qor_chain_large, and it completed without any issues.

AmirhosseinPoolad (approved):

amin1377:
@AmirhosseinPoolad: QoR for vtr_reg_qor_large
image

PR #3714, opened 2026-07-20
[Hecate] Added Hard Adder Version of 2D Arch

AlexandreSinger:
I found that the Hecate arch was missing hardened adders. I created a version of the 2D Hecate architecture with hardened adders.

The adders are modelled after the Flagship architecture. The LUT5 mode has register feedback muxs within the flut5 that feeds back into the in_C. I chose to not include these in the arithmetic mode since I was not sure what affect this would have.

The flut5 complex block has two identical outputs, one that is fed back and one that goes to general routing. I kept this structure for the arithmetic mode since it was reasonably easy to add and matches the structure of the rest of the complex block.

The delays are just scaled-down versions (0.75x) of the 22nm COFFE architecture delays. Real delays need to be added.


AlexandreSinger:
Once I have some compute lying around, I want to run the Koios benchmarks on this architecture to see how the distribution of the block types change with this change.

AlexandreSinger:
image

@vaughnbetz FYI

vtr_flow/arch/multi_die/hecate/2d/variations/hecate_2d_with_adders.xml
```diff
@@ -606,38 +608,38 @@
                   <interconnect>
                     <!--Clock -->
                     <direct name="direct5" input="arithmetic.clk" output="ff.clk"/>
-                    <!-- CAUTION: THESE DELAYS WERE RIPPED FROM THE blut5 MODE. NEED TO BE RE-COLLECTED. -->
+                    <!-- CAUTION: THESE DELAYS ARE JUST A SCALED VERSION OF AMON'S ARCH. NEED TO BE RE-COLLECTED. -->
```
vaughnbetz: Aman's arch

vtr_flow/arch/multi_die/hecate/2d/variations/hecate_2d_with_adders.xml
```diff
@@ -606,38 +608,38 @@
...
                     <!-- NOTE: I removed the registered feedback mux for the arith mode. It may not make sense to feedback the registered sumout. -->
```
vaughnbetz: Probably does make sense to feed it back so I'd eventually add this back in if possible. Won't make a huge difference, but it's useful for counters in that mode.

vtr_flow/arch/multi_die/hecate/2d/variations/hecate_2d_with_adders.xml
vaughnbetz: Aman's arch

vtr_flow/arch/multi_die/hecate/2d/variations/hecate_2d_with_adders.xml
vaughnbetz: Might want to comment that we have no delay on the CLB output carry delay since we put all CLB to CLB carry chain delay on the cin side (or maybe also on a direct link?)

vaughnbetz: Comment this. We break the five LUT into two four LUTs, each of which feeds one input of a dedicated adder.

vtr_flow/arch/multi_die/hecate/2d/variations/hecate_2d_with_adders.xml
```diff
@@ -606,38 +608,38 @@
...
                     <!-- NOTE: I also removed the input selection mux. Was not clear how to connect that. -->
                     <direct name="direct0" input="arithmetic.in_A" output="lut4[0:0].in[0:0]">
-                      <delay_constant max="1.1287999999999999e-10" in_port="arithmetic.in_A" out_port="lut4[0:0].in[0:0]"/>
+                      <delay_constant max="88.25e-12" in_port="arithmetic.in_A" out_port="lut4[0:0].in[0:0]"/>
```
vaughnbetz: Delays of the LUT inputs will be different; you could use the four fastest LUT5 delays as a guess of what these are, since that constitutes a LUT4 (with the same sizing).

vtr_flow/arch/multi_die/hecate/2d/variations/hecate_2d_with_adders.xml
```diff
@@ -759,8 +761,8 @@
           <!-- Carry chain links -->
           <direct name="carry_in" input="clb.cin" output="fle[0:0].cin">
             <!-- Put all inter-block carry chain delay on this one edge -->
-            <!-- CAUTION! THIS IS WRONG! I just multiplied the 22nm delay by 0.75 -->
-            <delay_constant max="0.12e-9" in_port="clb.cin" out_port="fle[0:0].cin"/>
+            <!-- CAUTION! THIS IS WRONG! I just multiplied the 22nm delay by 0.67 -->
```
vaughnbetz: Don't we also need to add the CLB-level cin -> cout path to the directlist (cluster level direct interconnect) in the earlier part of the arch file?

vaughnbetz (changes requested):
Thanks!  Some suggestions and comments embedded.

AlexandreSinger:
@vaughnbetz A few things I would like your feedback on.

### The feedback muxes:
Below is the fle in the clb in the flut5 mode (this has not been modified by me):
hecate-clb-flut5 mode

The multiplexer feedbacks that I was confused on are labelled in the figure as `mux1` and `2`. These muxes are not found in Aman's architecture or in the flagship architecture as far as I am aware.

Here is what is presently in the architecture file for the arithmetic mode:
hecate-clb-arithmetic mode

I was not sure how to connect these two muxes (for example should `mux1` go to both LUTs or just 1, or should `2` feed into the FF or the adder?); so I did not model them. I am ok with adding them, but where exactly should I add them that would make sense?

### The carry chains

Here is a model of the clb when all the fle's are in arithmetic mode:
hecate-clb-carry-chain-delays

Here is the delay path for the carry chain that I chose to go with:

1. Adder has intrinsic delay from a/b -> cout (this is the start of the chain)
2. For the interconnect between adders, there is NO interconnect delays between adders within the same clb.
3. Adder has intrisic delay from cin -> cout
4. Rinse and repeat for each adder in the clb
5. IF and only if the carry chain goes between two clbs, the interconnect delay is added on the cin interconnect feeding into the first fle.
6. Rinse and repeat for each adder in the clb
7. When the chain ends, we have TWO delays: the cin->sumout delay AND the sumout to FF delay.

The way I modeled the carry chain delay matches how it is done in the flagship architecture and Aman's architecture, aside from the following discrepency:

- Flagship has no sumout to FF interconnect delay delay

Let me know if this is not how you want the carry chains to be modeled.

AlexandreSinger:
For completeness, here is the ble6 mode:
hecate-clb-ble6 mode

One small note: when in ble6 mode, one of the out_local signals is left disconnected. I am not sure if it was purposeful or not. I do not see a comment explaining why this is like this.

vaughnbetz:
Hi Alex,

The local feedback (called qfbk in Altera notation) would exist in
arithmetic mode. It would go to the LUTC of both 4-LUTs.  It would
basically be there no matter what, given that it is there for the LUT5
mode. It probably won't make much difference to the mapping solutions, but
adding it would make the architecture more self-consistent, and it may
enable a slightly faster or denser packing.

Adder: that all sounds good. You have more places to add delay than you
need, and what you describe is a typical way to distribute the delays.

vaughnbetz:
Looks reasonable. One minor improvement: it would be more typical to hook
the LUT + FF combination to both local routing muxes in LUT6 mode (i.e.
give two options for the BLE to drive two different local feedbacks).
Reason: the local feedback is there already, so driving it with another 2:1
mux improves routability and can make it a bit easier to use the LUT and FF
together when the LUT doesn't drive the FF. There's almost no cost to doing
so.

We could sync up with Andrew to see what COFFE assumes in terms of the
BLE -> local mux structure.  Even if it wasn't there before, an extra 2:1 mux (or more likely an extra input in an existing BLE-> local mux selector) output muxing is pretty small though, so
little changes are going to be pretty close to the noise floor in terms of
area.

AlexandreSinger:
@vaughnbetz I spent all morning on this and unfortunately the feedback muxes in the arithmetic mode did not work very well.

Starting with the FF feedback mux, in order to support it you must add a new pb to wrap the two lut4s (this is required since you cannot connect a single mux to two pb inputs:
image
When I tested this, the pack time was 3x slower and the quality stayed mostly the same on attention_layer.v. I also saw similar results in my small scale testing. I think what is going on here is that the packer is struggling to recognize that the C port is special and causes more intra-lb unroutes. Since the quality remained largely the same and we only lost pack time, I chose not to include this feedback mux.

Then there is FF input mux (allowing LUT_C to be fed through to the FF directly, bypassing the adder). This also increased the run time (by a bit more than 3x, but note that the two feedback muxes together was 3x slower). This did not change the overall number of clusters by much; looking at the pb usages, it looks like FFs that were used in LUT5 mode were moved into the arithmetic mode instead. I think there is very little reason to use these FFs in the arithmetic mode since adders must be placed into this mode, and any FF that goes into this mode may block the arithmetic mode from creating the adder.

I think what is happening here is that I modelled the old arithmetic mode after the Flagship and Aman's architecture which we know packs very well. Now that we are adding these muxes, we are diverging from these architectures and it is causing the pack time to explode. Unfortunately we are not getting any gains from having these in (which may be a consequence of the packer's algorithms not taking advantage of them); but I do not think it is worth the runtime increase.

NOTE: I did try turning on memoization; it did not solve the runtime issue. Although it did improve runtime by 30%; it improved the runtime for both versions pretty equally (so instead of 3x slower it was 2.6x slower on my test design) so it was still far slower with no difference in quality.

Edit: I moved the code changes into a branch on my fork in case we ever want it again in the future: https://github.com/AlexandreSinger/vtr-verilog-to-routing/tree/vault-hecate-2d-feedback-muxs

AlexandreSinger:
FYI: Just to be safe, I launched a test run on the complex arithmetic mode to see if the runtime and quality I was experiencing is consistent across the entire benchmark suite. The config file can be found here:
https://github.com/AlexandreSinger/fpga-analytical-placement/blob/main/testing/multi_die/hecate_2d_arith_muxes/config/config.txt

I will post the results as they come in.

vaughnbetz:
OK, thanks Alex. We can skip the extra feedback connection then.

AlexandreSinger:
> OK, thanks Alex. We can skip the extra feedback connection then.

Sounds good. For completeness, here are the results on Koios (traditional flow, auto-sized, memoization on, two dropped circuits due to unroutes):

  | vtr_flow_elapsed_time | pack_time | place_time | crit_path_route_time | num_clb | critical_path_delay | routed_wirelength
Normalized | 1.61 | 2.91 | 0.94 | 0.94 | 1.00 | 1.00 | 0.99

2.9x pack time increase, leading to a 1.6x flow time increase. The number of CLB did reduce, but not by a meaningful amount. CPD and WL is slightly better; but there are very close to a draw.

Full raw results can be found here: https://docs.google.com/spreadsheets/d/1KqJdhx6dPJAaDtYS6oThLzSyEMeXqDJGsYxaSaGF7GM/edit?gid=977653768#gid=977653768

I will leave the complex arithmetic mode on the branch I created for now; but I will proceed without it.

AlexandreSinger:
@vaughnbetz @AmirhosseinPoolad @soheilshahrouz Do you guys have any further comments on this PR bringing in the updated Hecate 2D architecture?

vaughnbetz (approved):
LGTM

PR #3716, opened 2026-07-21
Draw critical path endpoint markers + total delay time message

kevinxing-a12:
#### Description

Draw a star at the beginning and end of a critical path. Beginning is marked as green and end is marked as red, so it is always fast to keep track of the critical path direction while it's changing.

Inside the star, a number that represents the critical path index is drawn. Currently, we only draw the worst setup timing path, so this index is always 0. In the very next PR, I will modify the relevant functions so that multiple paths can be drawn at the same time, and having this little indicator will be very helpful.

This indicator is in fact also important for the second feature included in this PR: now, when the `Critical Path Delay` UI checkbox is toggled, a message that shows the total delay time of the associated critical path index is drawn fixed at the screen's top right corner. Once we have multiple paths, the index will help us distinguish them.

#### Miscellaneous
There is one part of the code that does not have comment, and it is intentional:
In draw_timing_edge_flylines(), draw_crit_path.cpp:

> if (edge_idx == 0) {
         draw_crit_path_endpoint(e_crit_path_endpoint_type::START, start, 0, g);
   }

The 0 in the function call is a magic number, and it is meant to represent the critical path index (which can only be 0 now). In the next PR, I will make the drawing of multiple critical paths available, where a few functions (including this one) will be changed, so I think it is fine to leave this intermediate code here. After all, it will be replaced very soon.

One small thing open to discussion: the stars are affiliated to the flylines (i.e. they are drawn together with the flylines), so if the first and/or last flyline segment is hidden due to the device layer setting, the corresponding stars will become invisible as well. Do people find this natural?

#### Motivation and Context

Enhance how the critical path is visually represented, and provide developers the option to see the total delay time.

Tested on different devices using multiple regression tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -1758,6 +1758,7 @@ bool rgb_is_same(ezgl::color color1, ezgl::color color2) {
     color2.alpha = 255;
     return (color1 == color2);
 }
+
 t_draw_layer_display get_element_visibility_and_transparency(int src_layer, int sink_layer) {
     t_draw_layer_display element_visibility;
     t_draw_state* draw_state = get_draw_state_vars();
@@ -1765,16 +1766,17 @@ t_draw_layer_display get_element_visibility_and_transparency(int src_layer, int 
     element_visibility.visible = true;
     bool cross_layer_enabled = draw_state->cross_layer_display.visible;
 
-    //To only show elements (net flylines,noc links,etc...) that are connected to currently active layers on the screen
-    if (!draw_state->draw_layer_display[sink_layer].visible || !draw_state->draw_layer_display[src_layer].visible || (!cross_layer_enabled && src_layer != sink_layer)) {
-        element_visibility.visible = false; /* Don't Draw */
-    }
-
     if (src_layer != sink_layer) {
-        //assign transparency from cross layer option if connection is between different layers
+        if (!cross_layer_enabled) {
+            element_visibility.visible = false; /* Don't Draw */
+        }
+        // Assign transparency from cross layer option if connection is between different layers.
         element_visibility.alpha = draw_state->cross_layer_display.alpha;
     } else {
-        //otherwise assign transparency of current layer
+        if (!draw_state->draw_layer_display[sink_layer].visible || !draw_state->draw_layer_display[src_layer].visible) {
+            element_visibility.visible = false; /* Don't Draw */
+        }
+        // Assign transparency of current layer.
         element_visibility.alpha = draw_state->draw_layer_display[src_layer].alpha;
     }
 
--- a/vpr/src/draw/draw.cpp
+++ b/vpr/src/draw/draw.cpp
@@ -1773,7 +1773,7 @@ t_draw_layer_display get_element_visibility_and_transparency(int src_layer, int 
         // Assign transparency from cross layer option if connection is between different layers.
         element_visibility.alpha = draw_state->cross_layer_display.alpha;
     } else {
-        if (!draw_state->draw_layer_display[sink_layer].visible || !draw_state->draw_layer_display[src_layer].visible) {
+        if (!draw_state->draw_layer_display[src_layer].visible) {
             element_visibility.visible = false; /* Don't Draw */
         }
         // Assign transparency of current layer.
--- a/vpr/src/draw/draw_basic.cpp
+++ b/vpr/src/draw/draw_basic.cpp
@@ -571,7 +571,10 @@ void draw_x(float x, float y, float size, ezgl::renderer* g) {
     g->draw_line({x - size, y - size}, {x + size, y + size});
 }
 
-void draw_star_fixed_px(double x, double y, double size, ezgl::renderer* g) {
+void draw_star_fixed_px(ezgl::point2d star_coords, double size, ezgl::renderer* g) {
+    double x = star_coords.x;
+    double y = star_coords.y;
+
     // The star size is defined in pixels, but drawing is performed in world coordinates.
     double size_in_world = size / get_pixels_per_world_unit(g);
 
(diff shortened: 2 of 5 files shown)


vpr/src/draw/draw.cpp
```diff
@@ -1758,23 +1758,25 @@ bool rgb_is_same(ezgl::color color1, ezgl::color color2) {
...
     if (src_layer != sink_layer) {
-        //assign transparency from cross layer option if connection is between different layers
+        if (!cross_layer_enabled) {
+            element_visibility.visible = false; /* Don't Draw */
+        }
+        // Assign transparency from cross layer option if connection is between different layers.
         element_visibility.alpha = draw_state->cross_layer_display.alpha;
     } else {
-        //otherwise assign transparency of current layer
+        if (!draw_state->draw_layer_display[sink_layer].visible || !draw_state->draw_layer_display[src_layer].visible) {
```
vaughnbetz: Isn't sink_layer == src_layer if we reach here?  Hence this code could be simplified.

vpr/src/draw/draw_basic.cpp
```diff
@@ -571,6 +571,40 @@ void draw_x(float x, float y, float size, ezgl::renderer* g) {
     g->draw_line({x - size, y - size}, {x + size, y + size});
 }
 
+void draw_star_fixed_px(ezgl::point2d star_coords, double size, ezgl::renderer* g) {
```
vaughnbetz: I'd rename the size argument to size_in_pixels so the variable becomes self documenting.

vpr/src/draw/draw_basic.h
```diff
@@ -56,6 +56,21 @@ void draw_routing_bb(ezgl::renderer* g);
 /* Draws an X centered at (x,y). The width and height of the X are each 2 * size. */
 void draw_x(float x, float y, float size, ezgl::renderer* g);
 
+/**
+ * @brief Draws a filled star centered at the specified world-coordinate location with a fixed pixel size.
+ *
+ * The star size is converted from pixels to world units using the current renderer zoom level, so it
```
vaughnbetz: Can't you just say the star size is in screen (pixel) coordinates so it is the same size at any zoom?  Don't need to give implementation details here.

vpr/src/draw/draw_basic.h
```diff
@@ -56,6 +56,21 @@ void draw_routing_bb(ezgl::renderer* g);
...
+ * appears with a consistent on-screen size while zooming. The star can be imagined as a square with
+ * four triangles (tip pointing outwards) affixed to its four sides. The triangle's base has the same length
```
vaughnbetz: Delete the std::vector comment. Don't describe the implementation here -- just the interface and what it does. Implementation comments go in the .cpp file where they are needed.
Rename size to size_in_pixels

vpr/src/draw/draw_basic.h
```diff
@@ -56,6 +56,21 @@ void draw_routing_bb(ezgl::renderer* g);
 /* Draws an X centered at (x,y). The width and height of the X are each 2 * size. */
 void draw_x(float x, float y, float size, ezgl::renderer* g);
 
+/**
+ * @brief Draws a filled star centered at the specified world-coordinate location with a fixed pixel size.
+ *
+ * The star size is converted from pixels to world units using the current renderer zoom level, so it
+ * appears with a consistent on-screen size while zooming. The star can be imagined as a square with
```
vaughnbetz: Quickly describe the square in terms of size_in_pixels. E.g. a four-pointed star with the distance to the center to a star point of size_in_pixels. The center of the star is filled in with a square of width size_in_pixels / 2.  (I think).

Then move detailed discussion of triangles etc. to the .cpp file.

vpr/src/draw/draw_crit_path.cpp
```diff
@@ -880,9 +961,38 @@ static void draw_labels(std::vector<t_label_drawing_info>& final_label_drawing_i
...
+
+    // The canvas has an inverted y axis. Therefore, bottom right appears to be top right on the screen, which is what we aim for.
+    ezgl::point2d bottom_right = g->get_visible_screen().bottom_right();
+    // String dimension in pixels.
+    ezgl::t_text_dimension str_dimension = g->get_text_dimension(total_delay_str);
+
+    // Use the screen (pixel) coordinates to draw the total delay string at a fixed screen location.
+    g->set_coordinate_system(ezgl::SCREEN);
+    // Due to the same inverted y axis reason, adding str_dimension.height to y actually lowers down the string from the screen top.
+    g->draw_text(ezgl::point2d{bottom_right.x - str_dimension.width, bottom_right.y + str_dimension.height}, total_delay_str);
```
vaughnbetz: This is pretty weird ... do we have something wrong with drawing text in screen coordinates?  We shouldn't have to ask for the bottom_right corner and use its coordinates in order to draw something in the top right in screen drawing mode.

vaughnbetz (approved):
Looks mostly good. A couple of things to check / improve.

vaughnbetz (changes requested):
Whoops, meant this needs some changes.

vpr/src/draw/draw_crit_path.cpp
kevinxing-a12: Are you referring to the inverted y axis thing, or that we can simply hardcode the top right corner?

kevinxing-a12: If the latter, I'm not sure if that would work well under different monitor resolution. I can do a quick experiment though.

vaughnbetz: The y-value of the screen top in screen coordinates should always be 0.  So just drawing at the x value you compute, str_dimension.height should work.  If it doesn't, then I think we have a bug in the screen coordinates which should be flagged to Oleksandr.

The .bottom_right.y should be some significant y value (e.g. 1000) in screen coordinates.  If it is instead 0, then I think that's a bug too (we didn't define bottom_right in a reasonable way).

Your code clearly draws correctly, but the screen ccoordinate logic is strange for the y dimension, which is making wonder if there is an underlying bug in the api.

kevinxing-a12: I see. You are absolutely right; I overcomplicated the y coordinate logic.

vaughnbetz (approved):
Looks good, thanks.

